PR #15, opened 2022-01-07
chore: replace funding GH username

griest024:


damienwebdev:
@prisis this can be closed, but we should decide what to do here. I think we should still target the OpenCollection to support Marak -- his contributions still have merit and deserve financial support.

damienwebdev:
Closing this as I think we will leave Marak for posterity and none the current maintainers are interested in receiving fiscal contributions from this repo. If another maintainer wants funding to accomplish additional features, we will revisit this topic at that time.

Shinigami92:
IMO we should not replace it but just remove it for now
We can mention Marak in other files like README and LICENSE
But the funding file actively adds the sponsor button to the repo!

PR #17, opened 2022-01-07
chore: remove bower/meteor junk

griest024:


clarkerican (approved):

prisis (approved):

damienwebdev:
I don't think we should remove these. I don't know the bower ecosystem well enough to say whether or not we will break something by doing this. We would need someone from that ecosystem to tell us whether or not we should remove this.

cc: @kkirsche for the last known modifier.

griest024:
I don't know that much about bower either but surely whatever we do to this repo won't affect marak's published packages, right?

damienwebdev:
@griest024 it will not. Maybe v6.0.0 is a breaking release in this regard. We remove a number of older systems that may break if they were to update to v6 for example.

Shinigami92:
https://bower.io/ itself is kinda deprecated and already has a banner on it to migrate to npm and other stuff.

IMO this is safe to be purged, we should just concentrate on npm package
And maybe a CDN version :thinking:

griest024:
Afaik [unpkg](https://unpkg.com/) hosts npm packages automatically.

damienwebdev (approved):
lgtm

PR #23, opened 2022-01-08
Convert Localised Codes in ReadME into tabular helpers with language description 

Devabubakar:


damienwebdev:
The README was previously auto-generated, and I think we want to stick to that I suspect that we want to move to using something like https://fakerjsdocs.netlify.app/localization.html#individual-localization-packages perhaps @JeffBeltran we can merge your docs site into here?

PR #12, opened 2022-01-07
fix(tests): fix failing tests

MateusDadalto:
fixes #7

PR #35, opened 2022-01-09
feat(github): add github actions

### LINKED ISSUE #1: ci: Add Github Actions pipeline

ericjeker:
Proposal for https://github.com/faker-js/faker.js/issues/1

I just added testing and linting for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: "Continuous Integration"
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: "Testing"
+    needs: lint
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+          cache: 'npm'
+      - run: npm install
+      - run: npm run test
+  lint:
+    runs-on: ubuntu-latest
+    name: "Linting"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+          cache: 'npm'
+      - run: npm install
+      - run: npm run lint
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: "Continuous Integration"
 
 on:
   push:
+    branches:
+      - main
   pull_request:
 
 jobs:
@@ -15,7 +17,7 @@ jobs:
         with:
           node-version: '16'
           cache: 'npm'
-      - run: npm install
+      - run: npm ci
       - run: npm run test
   lint:
     runs-on: ubuntu-latest
@@ -26,5 +28,5 @@ jobs:
         with:
           node-version: '16'
           cache: 'npm'
-      - run: npm install
+      - run: npm ci
       - run: npm run lint


Shinigami92:
It's a start and we can improve it later on :+1:

damienwebdev (approved):
lgtm

prisis:
What node versions should we support?

Shinigami92:
before end of march: 12, 14, 16 and 17 + mac 16 + win 16
after end of march we could drop 12

But do we have relevant stuff for differences between unix and win?

Shinigami92:
For references:
Vite: https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml
My project: https://github.com/prettier/plugin-pug/blob/main/.github/workflows/ci.yml

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
+name: "Continuous Integration"
+
+on:
+  push:
```
Shinigami92, line 4: **suggestion:** it could be better to perform a ci run only on push on main xor on pull_request, otherwise you will produce duplicate workflow runs

```suggestion
  push:
    branches:
      - main
```

damienwebdev: For reference for how we do this in Azure Pipelines (I assume the GA syntax is similar):

```yml
trigger:
  branches:
    include:
    - master
  paths:
    exclude:
    - docs/*
    - README.md

pr:
  branches:
    include:
    - master
  paths:
    exclude:
    - docs/*
    - docs/README.md
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: "Testing"
+    needs: lint
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
```
damienwebdev, line 16: We should likely leverage https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix for matrixing across NodeJS versions

damienwebdev (changes requested):

.github/workflows/ci.yml
Shinigami92: I would suggest to not exclude `docs/*` and `README.md`
Cause later on we could also check these are correctly formatted with `pnpm prettier --check .`

.github/workflows/ci.yml
Shinigami92: May we do this in an extra PR?
For now we could just bring a simple CI to live and enhance it later in additional PRs

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+          cache: 'npm'
+      - run: npm install
```
Shinigami92, line 18:
```suggestion
      - run: npm ci
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    name: "Linting"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+          cache: 'npm'
+      - run: npm install
```
Shinigami92, line 29:
```suggestion
      - run: npm ci
```

Shinigami92 (changes requested):

Shinigami92:
It's currently expected that CI test are failing!
We already have #12 for that, but want to see that it takes effect

damienwebdev (approved):

Shinigami92 (approved):

PR #40, opened 2022-01-09
chore: add prettier

Shinigami92:
Add and configure prettier

Please note that I did NOT run it for now.

First we should bring in the GH Actions #1 
Also we may want to fix the test #12 

This PR can already be merged, but I can create a second PR later to format all files that are needed to get formatted.
Also we could commit formatted files when we touched them.

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -1,0 +1,11 @@
+// @ts-check
+
+/**
+ * @type {import('prettier').Options}
+ */
+module.exports = {
+  plugins: [
+    /* require.resolve("prettier-plugin-organize-imports") */
+  ],
+  singleQuote: true,
+};
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,4 @@
+package-lock.json
+coverage/
+dist/
+doc/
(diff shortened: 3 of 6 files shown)


.prettierrc.cjs
```diff
@@ -0,0 +1,11 @@
+// @ts-check
+
+/**
+ * @type {import('prettier').Options}
+ */
+module.exports = {
+  plugins: [
+    /* require.resolve("prettier-plugin-organize-imports") */
```
Shinigami92, line 8: I used cjs and added this for later.
This way, it's possible to use package managers like `pnpm`

Shinigami92:
I prepared https://github.com/Shinigami92/faker.js/pull/1

damienwebdev:
I'm not too familiar with Prettier or its configuration. What benefits does it provide over a VSCode plugin like https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint?

Shinigami92:
prettier and eslint are two different things.
prettier is for formatting and eslint is for finding e.g. code smells and help improve the code

damienwebdev:
Hrm. I'm confused here then, I have used `eslint` in the past for both fixing formatting (`lint --fix`) as well as for code smell checking and other various linting features. I think prettier should likely be used for Markdown checking (as described in the current prettier config) but I'm not sure of its usage for `.js` files when this could be handled by `eslint`.

Shinigami92:
prettier is soooo much stronger in formatting then eslint ^^
it formats usually every file, not only md or js.

Take a look into my repo: https://github.com/prettier/plugin-pug
Or here for Vite: https://github.com/vitejs/vite/blob/8fb854fd62f0edc9b1d1bab79b7d9b11cd03cf90/package.json#L67-L77

prettier formats the code, and after that you run eslint to "fix"/enhance stuff
prettier does NOT transform code, it just formats it
This is really great for Open Source projects that are community driven, cause you don't need to argue in any way about code formatting-style but you just format it as it is configured for the project

There is even a eslint plugin to read in the prettier config and this will report if something was not formatted correctly and then you can see this in GitHub PR reviews

damienwebdev (approved):
lgtm

prisis (approved):

PR #41, opened 2022-01-09
chore(ci): automated issue labeling

Shinigami92:


.github/workflows/issue-labeling.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  reply-labeled:
+    runs-on: ubuntu-latest
+    steps:
+      - name: contribution welcome
+        if: github.event.label.name == 'help wanted'
+        uses: actions-cool/issues-helper@v2
+        with:
+          actions: "create-comment, remove-labels"
+          token: ${{ secrets.GITHUB_TOKEN }}
```
damienwebdev, line 16: Who's token is this?

damienwebdev (changes requested):

.github/workflows/issue-labeling.yml
Shinigami92: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

damienwebdev: Neat.

damienwebdev (approved):

.github/workflows/issue-labeling.yml
```diff
@@ -0,0 +1,29 @@
...
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            Hello @${{ github.event.issue.user.login }}. We like your proposal/feedback and would appreciate a contribution via a Pull Request by you or another community member. We thank you in advance for your contribution and are looking forward to reviewing it!
+          labels: "pending triage"
+
+      - name: remove pending
+        if: github.event.label.name == 'enhancement' || github.event.label.name == 'bug' || (contains(github.event.label.name, 'pending triage') == false)
```
griest024, line 23:
```
(contains(github.event.label.name, 'pending triage') == false)
```

Is this saying that if the label name does NOT contain `pending triage`, then remove the pending triage label? That seems wrong.

griest024 (changes requested):
I think each step deserves a brief comment saying the intention of the step.

.github/workflows/issue-labeling.yml
Shinigami92: No, it is a filter
Check only labels that does not already have it removed

This workflow is already working in Vite repo

Shinigami92:
> I think each step deserves a brief comment saying the intention of the step.

Are the name of the steps not already self-explanatory?

griest024:
> > I think each step deserves a brief comment saying the intention of the step.
> 
> Are the name of the steps not already self-explanatory?

I don't think the conditionals are self-explanatory.

Shinigami92:
> I don't think the conditionals are self-explanatory.

I mean

https://github.com/faker-js/faker.js/blob/46bc978163c1e1aab1c26fe6ec10822c9a214a09/.github/workflows/issue-labeling.yml#L11

and 

https://github.com/faker-js/faker.js/blob/46bc978163c1e1aab1c26fe6ec10822c9a214a09/.github/workflows/issue-labeling.yml#L22

griest024:
> > I don't think the conditionals are self-explanatory.
> 
> I mean
> 
> https://github.com/faker-js/faker.js/blob/46bc978163c1e1aab1c26fe6ec10822c9a214a09/.github/workflows/issue-labeling.yml#L11
> 
> and
> 
> https://github.com/faker-js/faker.js/blob/46bc978163c1e1aab1c26fe6ec10822c9a214a09/.github/workflows/issue-labeling.yml#L22

What I mean is, I don't think the names of the steps explain why the conditionals are what they are. From a high level, what is the step trying to do? Why even check the label names at all? These are the questions that I would want answered by a comment.

.github/workflows/issue-labeling.yml
griest024: I need a sanity check.

Consider a label `test`. Wouldn't that label cause `(contains(github.event.label.name, 'pending triage') == false)` to be `true` and therefore the entire `if` condition to be `true`, causing the step to run? But based on your comment, I don't think `test` should cause this step to run.

Shinigami92: I missed to add the issue templates, did this now

---

The remove pending step will only be fired if
enhancement or bug WILL be assigned and filters issues that don't have the pending triage label

I swear this works in Vite for month now
https://github.com/vitejs/vite/blob/8fb854fd62f0edc9b1d1bab79b7d9b11cd03cf90/.github/workflows/issue-labeled.yml#L23

griest024: okay I guess I don't understand this conditional syntax

damienwebdev (approved):
lgtm

Shinigami92:
@griest024 Can I receive your approval anyway? I added the comments :slightly_smiling_face: 
And still: if something is wrong, we can change it

griest024 (approved):

Shinigami92:
Test was done in #45

PR #48, opened 2022-01-10
feat(gitignore): add dist folder to .gitignore

### LINKED ISSUE #36: fix(dist): The dist folder is present in the repository, shouldn't we remove that?
Add `dist` to .gitignore?

ericjeker:
Solve issue: https://github.com/faker-js/faker/issues/36

I added the dist folder to .gitignore and also update the `.gitignore` to include better coverage.
I kept packer and microbundle so if we migrate to another builder it will be there already.

One thing I didn't remove are those two files that are created when running `npm run browser`:

```
  examples/browser/js/faker.js
  examples/browser/js/faker.min.js
```

Not really sure about the implication of removing them so I kept them for now.


prisis:
If we will start to use a build from ts/js -> to esm and so on than we can remove them :)

Shinigami92 (approved):

damienwebdev (approved):
LGTM

PR #44, opened 2022-01-09
docs: update Faker.js logo and refs

damienwebdev:

PR #50, opened 2022-01-10
Removes broken decorative image

schalkventer:


Shinigami92:
This will already be done with #44

PR #60, opened 2022-01-10
chore: format files with prettier

Shinigami92:


Shinigami92:
I will create several PRs for sub folders, so a review can be made individually without laggs

PR #69, opened 2022-01-10
🔥 remove FUNDING.yml

ctcpip:


damienwebdev:
As described here: https://github.com/faker-js/faker/discussions/56 we will not be progressing with this.

ctcpip:
ah, I looked for an existing issue, but didn't see any.  I haven't interacted much with the `discussions` feature.  👀   thanks

ctcpip:
@damienwebdev - perhaps we should add a comment to the file for future visitors, maybe include a link to that discussion?

damienwebdev:
I don't think we should litter the codebase with externalities, just like issues are for bugs, features, etc. If people want further information, they can look to previous issues/discussions.

damienwebdev:
cc: @ctcpip https://github.com/faker-js/faker/pull/110

PR #72, opened 2022-01-10
feat: migrate to typescript

Shinigami92:
# This is the base branch of the TS migration

# Please do not squash-merge this branch

## How to test this PR
Use `npm build` and then run `npx mocha test/<module>.unit.js` to test the specific module
`datatype` is already working as an example!

We will create distinct PRs for each module and base them on this PR so a review can be simply made
After that we merge all together and then look that all (untouched) tests are running

## List of PRs

- [x] address (#133)
- [x] animal (#105)
- [x] commerce (#106)
- [x] company (#132)
- [x] database (#89)
- [x] datatype (this PR, as first-class example)
- [x] date (#83)
- [x] fake (#79)
- [x] finance (#131)
- [x] git (#78)
- [x] hacker (#81)
- [x] helpers (#85)
- [x] iban (#131)
- [x] image (#92)
- [x] internet (#94)
- [ ] locales
- [x] lorem (#86)
- [x] mersenne (this PR, also part of random in #91)
- [x] music (#107)
- [x] name (#103)
- [x] phone_number (#127)
- [x] random (this PR + needs a PR for extending it (#91))
- [x] system (#90)
- [x] time (#74)
- [x] unique (#128)
- [x] vehicle (#130)
- [x] word (#102)

image

Legend:
- red = not tackled yet
- green = in base branch or ready to be reviewed
- blue = in draft PR

<details> 
<summary>GraphViz Source</summary>

```
digraph ModuleGraph {
  # Colors:
  # red = not tackled yet
  # green = in base branch or ready to be reviewed
  # blue = in draft PR

  address[color=red]
  animal[color=blue]
  commerce[color=blue]
  company[color=red]
  database[color=green]
  datatype[color=green]
  date[color=green]
  fake[color=blue]
  finance[color=red]
  git[color=blue]
  hacker[color=blue]
  helpers[color=blue]
  iban[color=red]
  image[color=blue]
  internet[color=blue]
  locales[color=red]
  lorem[color=blue]
  mersenne[color=green]
  music[color=blue]
  name[color=blue]
  phone_number[color=red]
  random[color=blue]
  system[color=blue]
  time[color=green]
  unique[color=red]
  vehicle[color=red]
  word[color=blue]
  "_definitions"[color=black]
    
  datatype -> mersenne
  datatype -> datatype
  datatype -> random
  time
  git -> hacker
  git -> datatype
  git -> date
  git -> fake
  git -> random
  fake -> "*"
  hacker -> random
  hacker -> "_definitions"
[...]

PR #74, opened 2022-01-10
feat: migrate time

Shinigami92:
This branch is based on #72

test with `npm run build && npx mocha test/time.unit.js`

### DIFF
--- a/src/time.ts
+++ b/src/time.ts
@@ -6,6 +6,8 @@ export class Time {
    * @param outputType 'abbr' || 'wide' || 'unix' (default choice)
    */
   recent(outputType: 'abbr' | 'wide' | 'unix' = 'unix'): string | number {
+    // TODO @Shinigami92 2022-01-11: This is not non-deterministic
+    // https://github.com/faker-js/faker/pull/74/files#r781579842
     let date: string | number | Date = new Date();
 
     switch (outputType) {
--- a/src/time.ts
+++ b/src/time.ts
@@ -1,0 +1,26 @@
+export class Time {
+  /**
+   * recent
+   *
+   * @method faker.time.recent
+   * @param outputType 'abbr' || 'wide' || 'unix' (default choice)
+   */
+  recent(outputType: 'abbr' | 'wide' | 'unix' = 'unix'): string | number {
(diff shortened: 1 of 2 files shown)


src/time.ts
```diff
@@ -0,0 +1,26 @@
+export class Time {
+  /**
+   * recent
+   *
+   * @method faker.time.recent
+   * @param outputType 'abbr' || 'wide' || 'unix' (default choice)
+   */
+  recent(outputType: 'abbr' | 'wide' | 'unix' = 'unix'): string | number {
+    let date: string | number | Date = new Date();
```
HonzaMac, line 9: This ignores faker seed value. It would became non-deterministic.

HonzaMac (changes requested):

src/time.ts
Shinigami92: I could add a TODO for later
We currently aim just for the migration and not for runtime changes

MateusDadalto:
I don't think the actions are validating the typescript changes here.
If you run `tsc src/index.ts --outDir lib` and then `npm run test` we get tests failing.

src/time.ts
HonzaMac: Why adding feature then?

Shinigami92: What do you mean?
This is just a rewrite of `lib/time.js`

prisis: There is no new feaure  
  
\`\`\`

self.recent = _function_ (outputType) {  
  _if_ (_typeof_ outputType === 'undefined') { // this moved to the function  
    outputType = 'unix';  
  }  
  
  _var_ date = _new_ Date();  
  _switch_ (outputType) {  
    _case_ 'abbr':  
      date = date.toLocaleTimeString();  
      _break_;  
    _case_ 'wide':  
      date = date.toTimeString();  
      _break_;  
    _case_ 'unix':  
      date = date.getTime();  
      _break_;  
  }  
  _return_ date;  
};  
\`\`\`

Shinigami92: Still don't understand, as you said there is no new feature right now in this PR
It's just a migration to TS and only touched compile time code

prisis (approved):

PR #77, opened 2022-01-11
fix(eslint): fix all the problems encountered with Prettier and eslint

ericjeker:
Probably a duplicate, I just saw that @Shinigami92 already did that. But this includes corrections from eslint as well. 
Maybe we can merge the changes from the other PR and then I will redo the eslint integration?

There's one mystery here though, in `helpers.js` there are some Japanese characters, not sure what they re but they will break eslint if ran with `--no-ignore`.


Shinigami92:
We tried to explicitly not create PRs that contains 1.7k changed files
No one can review that.
Let's do it step by step.

ericjeker:
Yes, I saw your conversation on Discord only too late. I will redo it only with the files in `lib/`.

PR #79, opened 2022-01-11
feat: migrate fake

Shinigami92:
This is based on top of other TS migrations like #72

This module seems to need some other stuff cause it dynamically requests stuff with e.g. `faker[parts[0]][parts[1]]`
Usually this will request the name module. (#84)

test with `npm run build && npx mocha test/date.unit.js`

### DIFF
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -4,7 +4,15 @@ import type { Faker } from '.';
  * Generator method for combining faker methods based on string input
  */
 export class Fake {
-  constructor(private readonly faker: Faker) {}
+  constructor(private readonly faker: Faker) {
+    // Bind `this` so namespaced is working correctly
+    for (const name of Object.getOwnPropertyNames(Fake.prototype)) {
+      if (name === 'constructor' || typeof this[name] !== 'function') {
+        continue;
+      }
+      this[name] = this[name].bind(this);
+    }
+  }
 
   /**
    * Generator method for combining faker methods based on string input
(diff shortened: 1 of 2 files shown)


src/fake.ts
```diff
@@ -0,0 +1,110 @@
...
+   * ```
+   *
+   * This will interpolate the format string with the value of methods
+   * [name.lastName]{@link faker.name.lastName}, [name.firstName]{@link faker.name.firstName},
+   * and [name.suffix]{@link faker.name.suffix}
+   *
+   * @method faker.fake
+   * @param str
+   */
+  fake(str: string) {
```
HonzaMac, line 34:
```suggestion
  fake(str: string): string {
```

src/fake.ts
```diff
@@ -0,0 +1,110 @@
...
+    // find first matching {{ and }}
+    const start = str.search('{{');
+    const end = str.search('}}');
+
+    // if no {{ and }} is found, we are done
+    if (start === -1 || end === -1) {
+      return str;
+    }
+
+    // console.log('attempting to parse', str);
```
prisis, line 52: This should be removed, this will not break anything

src/fake.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    // extract method name from between the {{ }} that we found
+    // for example: {{name.firstName}}
+    const token = str.substr(start + 2, end - start - 2);
+    let method = token.replace('}}', '').replace('{{', '');
+
+    // console.log('method', method)
```
prisis, line 59: This should be removed, this will not break anything

prisis (approved):

PR #81, opened 2022-01-11
feat: migrate hacker

Shinigami92:
This is based on top of other TS migrations like #72

We need to migrate `helpers` cause this module depends on it

Also this module needs the `definitions` structure :thinking: I think we should try to create this dynamically instead of like it is currently (#84)

test with `npm run build && npx mocha test/hacker.unit.js`

### DIFF
--- a/src/hacker.ts
+++ b/src/hacker.ts
@@ -1,0 +1,84 @@
+import type { Faker } from '.';
+
+export class Hacker {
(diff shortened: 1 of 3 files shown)


Shinigami92:
@HonzaMac I have not added return types yet, cause the bodies of them are not clear right now what they really return
I want it stronger than `any[]`

prisis (approved):

PR #82, opened 2022-01-11
chore: add typedoc

ericjeker:
Added typedoc.

TODO: add it in the build process in Github Actions.


damienwebdev:
@ericjeker now that the `migrate-to-typescript` is merged to main, could you rebase and target main?

ericjeker:
Not sure this still need to be merged as Jess already generated the docs with Vitepress and it probably needs to be integrated with Vitepress directly.

damienwebdev:
@ericjeker could you output a `json` document that is consumable by the docs? @JessicaSachs how would this work with Vitepress?

JessicaSachs:
My dream is to be able to write human documentation in Markdown surrounding auto-generated typedocs that we export in json.

To do this we need to: 
1. Import the raw json and create UI components from it
2. Import these docs UI components throughout pages like guide/address.md

To do this, we can create N components for the API surface area and register them globally within the docs/components file. When you register components there, they become available globally in all Markdown files

```js
// somewhere within docs
// setting up jsx is actually easy, but this is the only time we'll use raw render fn's so it's nbd
import { h } from 'vue'
import typedocs from './typedocs.json'
import CodePreview from './components/CodePreview.vue'

export const docComponents = {}

each(typedocs, (value, key) => {
  docComponents['Typegen' + key] = h(CodePreview, { doc: value })
})
```

Now write the CodePreview Vue component to create the correct `td`'s and `tr`'s with formatting.

Expected API:
```md
---
title: API Guide
---

### Addresses

### Address API <Badge text="v3.3.0" type="tip" />
The address API is best used for (etc)

<TypegenAddress/>

### City

<!-- this is my next goal -->
<Playground namespace="address" />
....

```

JessicaSachs:
My dream is to be able to write human documentation in Markdown surrounding auto-generated typedocs that we export in json.

To do this we need to: 
1. Import the raw json and create UI components from it
2. Import these docs UI components throughout pages like guide/address.md

To do this, we can create N components for the API surface area and register them globally within the docs/components file. When you register components there, they become available globally in all Markdown files

```js
// somewhere within docs
// setting up jsx is actually easy, but this is the only time we'll use raw render fn's so it's nbd
import { h } from 'vue'
import typedocs from './typedocs.json'
import CodePreview from './components/CodePreview.vue'

export const docComponents = {}

each(typedocs, (value, key) => {
  docComponents['Typegen' + key] = h(CodePreview, { doc: value })
})
```

Now write the CodePreview Vue component to create the correct `td`'s and `tr`'s with formatting.

Expected API:
```md
---
title: API Guide
---

### Addresses

### Address API <Badge text="v3.3.0" type="tip" />
The address API is best used for (etc)

<TypegenAddress/>

### City

<!-- this is my next goal -->
<Playground namespace="address" />
....

```

I think this is a good approach. @antfu could you give me a sanity check? Am I missing something obvious?

ericjeker:
Hi guys, 

I created another PR with a clean state as this PR was from an outdated branch: https://github.com/faker-js/faker/pull/207

PR #90, opened 2022-01-11
feat: migrate system

Shinigami92:
test with `npm run build && npx mocha test/system.unit.js`

### DIFF
--- a/src/system.ts
+++ b/src/system.ts
@@ -148,9 +148,9 @@ export class System {
    * @method faker.system.fileExt
    * @param mimeType
    */
-  fileExt(mimeType) {
+  fileExt(mimeType?: string): string {
     const typeSet = new Set<string>();
-    const extensionSet = new Set();
+    const extensionSet = new Set<string>();
     const mimeTypes = this.faker.definitions.system.mimeTypes;
 
     Object.keys(mimeTypes).forEach((m) => {
--- a/src/system.ts
+++ b/src/system.ts
@@ -46,7 +46,7 @@ export class System {
    * @method faker.system.fileName
    */
   fileName() {
-    var str = this.faker.random.words();
+    let str = this.faker.random.words();
     str =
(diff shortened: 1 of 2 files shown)


src/system.ts
```diff
@@ -0,0 +1,213 @@
...
+      typeSet.add(type);
+
+      if (mimeTypes[m].extensions instanceof Array) {
+        mimeTypes[m].extensions.forEach((ext) => {
+          extensionSet.add(ext);
+        });
+      }
+    });
+
+    const types = setToArray(typeSet);
```
prisis, line 90: This const is not used

src/system.ts
```diff
@@ -0,0 +1,213 @@
...
+    const extensions = setToArray(extensionSet);
```
prisis, line 91: This const is not used

src/system.ts
```diff
@@ -0,0 +1,213 @@
...
```
prisis, line 140: same for this

src/system.ts
```diff
@@ -0,0 +1,213 @@
...
+    const mimeTypeKeys = Object.keys(this.faker.definitions.system.mimeTypes);
```
prisis, line 141: and this

src/system.ts
```diff
@@ -0,0 +1,213 @@
...
+      typeSet.add(type);
+
+      if (mimeTypes[m].extensions instanceof Array) {
+        mimeTypes[m].extensions.forEach((ext) => {
+          extensionSet.add(ext);
+        });
+      }
+    });
+
+    const types = setToArray(typeSet);
```
prisis, line 168: this is not used too  
  
did they just copy past it...

src/system.ts
```diff
@@ -0,0 +1,213 @@
...
+    const extensions = setToArray(extensionSet);
+    const mimeTypeKeys = Object.keys(this.faker.definitions.system.mimeTypes);
```
prisis, line 170: and this

prisis (approved):

PR #97, opened 2022-01-12
feat: rewrite name module to typescript

MohdImran001:
@Shinigami92 Please review.


Shinigami92:
Please rebase on `origin/migrate-to-typescript` or `upstream/migrate-to-typescript`

(https://gist.github.com/CristinaSolana/1885435)

src/name.ts
```diff
@@ -0,0 +1,298 @@
+import type { Faker } from '.';
+
+export class Name {
+  constructor(private readonly faker: Faker) {
+    // Bind `this` so namespaced is working correctly
+    for (const name of Object.getOwnPropertyNames(Name.prototype)) {
+      if (name === 'constructor' || typeof this[name] !== 'function') {
+        continue;
+      }
+      this[name] = this[name].bind(this);
```
ericjeker, line 10: This seems unnecessary to me.

MohdImran001: I don't get it. Can you please elaborate?

ericjeker: I don't think binding this to local properties is necessary in Javascript classes. I suppose this was done in the original code because this was not bound to the function. Or did you add this for a specific reason?

Shinigami92: It's exactly the other way around 
We explicitly add this so `const name = faker.name` will work 
It was not in original code

MohdImran001:
Closing this PR because of unnecessary commits will open a new PR for the name module

PR #98, opened 2022-01-12
feat: added prettier actions

prisis:


prisis:
This is just a proposal

.prettierignore
```diff
@@ -4,3 +4,5 @@ doc/
 examples/browser/js/
 CHANGELOG.md
 package-lock.json
+node_modules/
+test/mocha.opts
```
Shinigami92, line 8: I think prettier as an option for ignore unknown
Maybe we can configure that with the action also

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,31 @@
...
+        file_pattern: '**/*.*'
+
+    # This step only runs if prettier finds errors causing the previous step to fail
+    # This steps lists the files where errors were found
+    - name: Prettier Output
+      if: ${{ failure() }}
+      shell: bash
+      run: |
+        echo "The following files are not formatted:"
+        echo "${{steps.prettier-run.outputs.prettier_output}}"
```
Shinigami92, line 31:
```suggestion
        echo "${{ steps.prettier-run.outputs.prettier_output }}"
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,31 @@
...
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: lint
+      id: prettier-run
+      uses: rutajdash/prettier-cli-action@v1.0.0
```
Shinigami92, line 19: I think we can use so, that we are always on v1 latest
```suggestion
      uses: rutajdash/prettier-cli-action@v1
```

.prettierignore
Shinigami92, line 8: Please format this file in asc order and folders at top and files bottom

Shinigami92: I also think node_modules is ignored by default :thinking:

Shinigami92 (changes requested):

.prettierignore
prisis: Not sure, just added it, will not hurt :D

prisis: i will check

prisis: i will check

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,31 @@
...
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: lint
+      id: prettier-run
+      uses: rutajdash/prettier-cli-action@v1
```
Shinigami92, line 19:
```node
Error: Unable to resolve action `rutajdash/prettier-cli-action@v1`, unable to find version `v1`
```

:cry:

Shinigami92: Don't know why it works for e.g. `actions/checkout@v2` and other actions I'm used to, but this one

.prettierignore
prisis: On what files should prettier run?

.github/workflows/prettier.yml
prisis: Ah, because you need to create a version v1 to use something like this

.prettierignore
Shinigami92: everything but ignore unknown https://prettier.io/docs/en/cli.html#--ignore-unknown

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: lint
```
Shinigami92, line 17: This should not be named "lint"

Shinigami92 (changes requested):

.github/workflows/prettier.yml
prisis: Than no name?

Shinigami92: or prettify or check-formatting or something like this
it "conflicts" with (es)lint

Shinigami92: image

this just confused me and I thought I'm in the wrong action

prisis:
@Shinigami92 i did make is simpler

damienwebdev:
@prisis I'm not sure I understand the intent of this PR. Would it ever fail? Wouldn't the lint step beforehand fail first?

prisis:
I think this can be closed, after the eslint, has a plugin for prettier, im correct @Shinigami92?

Shinigami92:
> I think this can be closed, after the eslint, has a plugin for prettier, im correct @Shinigami92?

I think yes

PR #95, opened 2022-01-11
fix(#28): changed fakercloud url to cloudflare-ipfs.com

prisis:
…ar endpoint

PR #100, opened 2022-01-12
feat: deprecate avatar generator

fncolon:
Since fakercloud.com is not active anymore, the `avatar` feature is broke.

I propose changes to remove it for now.

Also it's not related to fake data, much like kind of generator.


prisis:
Hey :), i fixed it in https://github.com/faker-js/faker/pull/95

MilosPaunovic:
This is already in the works, [https://github.com/faker-js/faker/pull/95](https://github.com/faker-js/faker/pull/95).

PR #102, opened 2022-01-12
feat: migrate word

MohdImran001:
Part of #72.
Test with `npm run build && npx mocha test/word.unit.js`

### DIFF
--- a/src/word.ts
+++ b/src/word.ts
@@ -16,17 +16,15 @@ export class Word {
    * If specified length is unresolvable, returns random adjective.
    *
    * @method faker.word.adjective
-   * @param {number} [length] - optional length of word to return
-   * @returns {string}          a random adjective
+   * @param optional length of word to return
+   * @returns a random adjective
    */
   adjective(length?: number): string {
     var wordList = this.faker.definitions.word.adjective;
     if (length) {
-      wordList = this.faker.definitions.word.adjective.filter(function (
-        word: string
-      ) {
-        return word.length == length;
-      });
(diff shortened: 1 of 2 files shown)


src/word.ts
```diff
@@ -0,0 +1,183 @@
...
+      this[name] = this[name].bind(this);
+    }
+  }
+
+  /**
+   * Returns an adjective of random or optionally specified length.
+   * If specified length is unresolvable, returns random adjective.
+   *
+   * @method faker.word.adjective
+   * @param {number} [length] - optional length of word to return
```
Shinigami92, line 19: Please remove the JSDoc types

like 
```suggestion
   * @param optional length of word to return
```

any many more below

src/word.ts
```diff
@@ -0,0 +1,183 @@
...
+   * @returns {string}          a random adjective
+   */
+  adjective(length?: number): string {
+    var wordList = this.faker.definitions.word.adjective;
+    if (length) {
+      wordList = this.faker.definitions.word.adjective.filter(function (
```
Shinigami92, line 25: convert to arrow functions to not mess with `this`

Shinigami92 (changes requested):

src/word.ts
MohdImran001: are you talking about `filter` function ?

Shinigami92: yes :)

prisis (approved):

Shinigami92 (approved):

PR #103, opened 2022-01-12
feat: migrate name

MohdImran001:
Part of #72.
Test with `npm run build && npx mocha test/name.unit.js`

### DIFF
--- a/src/name.ts
+++ b/src/name.ts
@@ -1,0 +1,298 @@
+import type { Faker } from '.';
+
+export class Name {
+  constructor(private readonly faker: Faker) {
+    // Bind `this` so namespaced is working correctly
+    for (const name of Object.getOwnPropertyNames(Name.prototype)) {
+      if (name === 'constructor' || typeof this[name] !== 'function') {
+        continue;
+      }
+      this[name] = this[name].bind(this);
+    }
+  }
+
+  /**
+   * firstName
+   *
+   * @method faker.name.firstName
+   * @param {mixed} gender
+   * @memberof faker.name
+   */
+  firstName(gender?: string | number): string {
+    if (
+      typeof this.faker.definitions.name.male_first_name !== 'undefined' &&
+      typeof this.faker.definitions.name.female_first_name !== 'undefined'
+    ) {
+      // some locale datasets ( like ru ) have first_name split by gender. since the name.first_name field does not exist in these datasets,
+      // we must randomly pick a name from either gender array so faker.name.firstName will return the correct locale data ( and not fallback )
+
+      if (typeof gender === 'string') {
+        if (gender.toLowerCase() === 'male') {
+          gender = 0;
+        } else if (gender.toLowerCase() === 'female') {
+          gender = 1;
+        }
+      }
+
+      if (typeof gender !== 'number') {
+        if (typeof this.faker.definitions.name.first_name === 'undefined') {
+          gender = this.faker.datatype.number(1);
+        } else {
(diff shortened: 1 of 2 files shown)


src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+      }
+      this[name] = this[name].bind(this);
+    }
+  }
+
+  /**
+   * firstName
+   *
+   * @method faker.name.firstName
+   * @param {mixed} gender
```
Shinigami92, line 18: Please remove the param JSDoc types

Shinigami92 (changes requested):

src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+        } else if (gender.toLowerCase() === 'female') {
+          gender = 1;
+        }
+      }
+
+      if (typeof gender !== 'number') {
+        if (typeof this.faker.definitions.name.first_name === 'undefined') {
+          gender = this.faker.datatype.number(1);
+        } else {
+          //Fall back to non-gendered names if they exist and gender wasn't specified
```
Shinigami92, line 41: Would be cool if we already format a bit such comments

```suggestion
          // Fall back to non-gendered names if they exist and gender wasn't specified
```

src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+      suffix: string = '';
+
+    // in particular locales first and last names split by gender,
+    // thus we keep consistency by passing 0 as male and 1 as female
+
+    if (typeof gender !== 'number') {
+      gender = this.faker.datatype.number(1);
+    }
+
+    firstName = firstName || this.faker.name.firstName(gender);
```
Shinigami92, line 152: I already improved such code with `||=` in the other PRs

src/name.ts
MohdImran001: we should keep only `@method faker.name.firstName` right?

Shinigami92: param is actually okay, we can add doc messages/descriptions later

src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+      }
+      this[name] = this[name].bind(this);
+    }
+  }
+
+  /**
+   * firstName
+   *
+   * @method faker.name.firstName
+   * @param  gender
```
Shinigami92, line 18:
```suggestion
   * @param gender
```

src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    // in particular locales first and last names split by gender,
+    // thus we keep consistency by passing 0 as male and 1 as female
+
+    if (typeof gender !== 'number') {
+      gender = this.faker.datatype.number(1);
+    }
+
+    firstName ||= this.faker.name.firstName(gender);
+    lastName ||= lastName || this.faker.name.lastName(gender);
```
Shinigami92, line 153:
```suggestion
    lastName ||= this.faker.name.lastName(gender);
```

Shinigami92 (approved):

prisis (approved):

src/name.ts
```diff
@@ -0,0 +1,298 @@
...
+    // thus we keep consistency by passing 0 as male and 1 as female
+
+    if (typeof gender !== 'number') {
+      gender = this.faker.datatype.number(1);
+    }
+
+    firstName ||= this.faker.name.firstName(gender);
+    lastName ||= this.faker.name.lastName(gender);
+
+    switch (r) {
```
prisis, line 155: Just a note, this should be later transformed to a if

prisis (approved):

PR #75, opened 2022-01-10
Remove GitHub Sponsors link to original maintainer

DuaneOBrien:
Prospective sponsors who come to the new repository will be directed to sponsor Marek unless this file is updated. This change should at least prevent this from happening until new funding links are set up. 

cf. #2 
cf. #44

PR #110, opened 2022-01-12
chore: swap funding to be sent through the open collective

JessicaSachs:
Deals with #75 and #15 .

Removing the maintainers field is common and is what lodash and a few other projects do. Better to let NPM and Github show who all has committed to the repo.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,8 +1,8 @@
 # These are supported funding model platforms
 
-github: marak
+github: # Replace with the open collective's name https://docs.opencollective.com/help/collectives/osc-verification/github-sponsors
 patreon: # Replace with a single Patreon username
-open_collective: # fakerjs
+open_collective: fakerjs
 ko_fi: # Replace with a single Ko-fi username
 tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
 community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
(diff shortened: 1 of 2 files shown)


.github/FUNDING.yml
```diff
@@ -1,8 +1,8 @@
 # These are supported funding model platforms
 
-github: marak
+github: # Replace with the open collective's name https://docs.opencollective.com/help/collectives/osc-verification/github-sponsors
```
damienwebdev, line 3: My opinion is that we keep this, but we would need to vote on this decision. See https://github.com/faker-js/faker/discussions/56

damienwebdev (changes requested):

.github/FUNDING.yml
JessicaSachs: Yeah, we need to vote on this. My stance is that people have no idea who he is and just see the pink heart at the top and want to support the maintainers. Clicking the pink heart would not, in fact, support the maintainers.

JessicaSachs: Where do you want that discussion to happen?

damienwebdev: I unlocked https://github.com/faker-js/faker/discussions/56

damienwebdev (approved):

Shinigami92 (approved):

damienwebdev:
Please see https://github.com/faker-js/faker/discussions/56 for the discussion and voting.

PR #114, opened 2022-01-12
feat: added eslint action

prisis:


Shinigami92:
I don't think this is needed that way
I will setup typescript-eslint later anyways and the we will have proper inline lint in GitHub 
So sorry but I will close this for now

PR #115, opened 2022-01-12
fix: adds regex to `random.word` to trim unwanted characters

prisis:


Shinigami92:
Please fix this in based on the main branch, it has nothing to do with the ts migration 
Also we might want to have a test

Shinigami92:
Also folks could just do it manually filter it in their code for now, so I do not see it that high prio

MilosPaunovic:
> Also folks could just do it manually filter it in their code for now, so I do not see it that high prio

Yeah, but, it's more elegant if we deal with it ourselves in the codebase 🙂 
Anyway, this was one of the opened issues at the original repo, so, someone definitely encountered it.

damienwebdev:
This needs a rebase against main.

damienwebdev:
I'm not sure that I'm confident in this implementation. This seems like we should be modifying the word dictionary of a locale to be more correct for that locale rather than relying on a regex.

MilosPaunovic:
@damienwebdev That approach might be problematic, as words are generated from multiple methods https://github.com/faker-js/faker/blob/a973ee1ce64d2b593f64b633e927382280da8f27/src/random.ts#L169

And some of which, like countries, just have the parenthesis, mostly seen here:
https://github.com/faker-js/faker/blob/main/lib/locales/en/address/country.js

We could go through all the locales, but removing unwanted characters might change meaning of the word or make it incomprehensible.

For the reasons above, my vote goes to the regex approach.

PR #118, opened 2022-01-13
feat: extended internet email with a new option useFreeProviders

prisis:


mmahalwy (approved):

Shinigami92 (changes requested):
Please rebase on `main` instead of `migrate-to-typescript`

We want to **not** squash-merge the `migrate-to-typescript` branch later on

prisis:
I did the feat based on this branch, if this is needed in the alpha release, i will create a new PR

src/internet.ts
```diff
@@ -37,9 +37,16 @@ export class Internet {
           description: 'The last name of the user',
         },
         provider: {
-          type: 'string',
+          type: 'string|string[]',
```
Shinigami92, line 40: Are you sure this is valid syntax?

src/internet.ts
```diff
@@ -199,17 +206,32 @@ export class Internet {
...
    */
-  email(firstName?: string, lastName?: string, provider?: string): string {
-    provider ||= this.faker.random.arrayElement(
-      this.faker.definitions.internet.free_email
-    );
+  email(
+    firstName?: string,
+    lastName?: string,
+    provider?: string | string[],
+    useFreeProviders: boolean = false
```
Shinigami92, line 215: Personally coming from Vue, I would suggest something like this should be passed as an `options` argument.
That way, we can pass further options in the future if needed.

Also IMO we should wait with this a bit, cause a feature like that should properly get documented.

src/internet.ts
```diff
@@ -199,17 +206,32 @@ export class Internet {
...
+  ): string {
+    let _provider = this.faker.internet.domainName();
```
Shinigami92, line 217: I would suggest a name without underscore, cause historically this indicates `private` fields in JS community.
So instead we could name this just `domainName`

Shinigami92 (changes requested):

PR #131, opened 2022-01-13
feat: migrate finance

Shinigami92:


src/finance.ts
```diff
@@ -0,0 +1,404 @@
...
+  // NOTE: this returns a string representation of the value, if you want a number use parseFloat and no symbol
+
+  /**
+   * amount
+   *
+   * @method faker.finance.amount
+   * @param min
+   * @param max
+   * @param dec
+   * @param symbol
```
prisis, line 112: can you add autoFormat

src/finance.ts
```diff
@@ -0,0 +1,404 @@
...
+  }
+
+  /**
+   * currencyCode
+   *
+   * @method faker.finance.currencyCode
+   */
+  currencyCode() {
+    return this.faker.random.objectElement(
+      this.faker.definitions.finance.currency
```
prisis, line 157: You can put undefined into it, to fix this error or change the 2 param in objectElement to be optional

prisis (approved):

prisis (changes requested):
comments

src/finance.ts
```diff
@@ -0,0 +1,404 @@
...
+   * currencySymbol
+   *
+   * @method faker.finance.currencySymbol
+   */
+  currencySymbol() {
+    let symbol;
+
+    while (!symbol) {
+      symbol = this.faker.random.objectElement(
+        this.faker.definitions.finance.currency
```
prisis, line 183: same as the last comment on this function

PR #133, opened 2022-01-13
feat: migrate address

Shinigami92:


prisis (approved):

prisis:
Can you rebase this to check if the test are pasing?

Shinigami92:
> Can you rebase this to check if the test are pasing?

Has something to do with the `fake` function, will look into that after I merged #79 
Seems same problem as in #132

PR #134, opened 2022-01-13
feat: Add easteregg

clairefro:
- Adds `"Swartz"` to `locales/en/name/last_name`, as a nod to the history behind this fork :) 
- Updates deterministic seed unit tests accordingly.

Note - running `npm run browser` changed the format of strings in `last_name` from double to single quote, as per the prettier config.


Shinigami92:
I the other maintainers also like that, it's fine for me.
Please wait until we found a nice strategy to load the locale files. We plan to convert them to `json`, but this needs a rollup setup after the ts-migration.

Shinigami92:
> Can I also recommend forcing a `prettier --write` on all the files in `main` so that any new PRs don't have a ton of prettier junk?

We have planned this, but we need to sort out some files that will be generated
But we did already for some selected ones

JessicaSachs:
I understand the intent of this PR is positive and its goal is to recognize the history of the project (as easter eggs often do). However, on its own, it seems out of place as a way to show respect considering how not-fun/serious the Aaron Swartz story is.

I don't _disagree_ with adding the name, but I would like to state publicly that an easter egg is not a solution to this very sensitive topic.

clairefro:
The intent wasn't a joke rather a memorial, but I understand how it could be taken the wrong way. How about dropping it if too sensitive

PR #137, opened 2022-01-14
feat(domain_suffix): add missing domain for id_ID

fzn0x:
add `id` as one of all domain_suffix

https://en.wikipedia.org/wiki/.id


MilosPaunovic (approved):

prisis:
Didn't check before, can you add a test if its possible

fncolon:
Will work on this sunday!

fncolon:
> Didn't check before, can you add a test if its possible

Since there is no system logic and algorithm changes, what i see is we didn't need unit tests, correct me if I'm wrong, thank you!

Shinigami92 (approved):

damienwebdev (approved):
lgtm

fncolon:
Thanks to make clear the commit name!

PR #143, opened 2022-01-14
chore: update .npmignore to prevent unwanted code in the package dist

damienwebdev:


prisis (approved):

.npmignore
```diff
@@ -1,13 +1,27 @@
+.nyc_output/
```
Shinigami92, line 1: Instead of defining a huge `.npmignore`, we may want to use the `package.json`'s `files` field

Shinigami92 (changes requested):

Shinigami92 (approved):
Do files property in later PR, so we can do a release now

PR #145, opened 2022-01-14
fix: fixed date methods not being deterministic

prisis:


damienwebdev:
Needs rebase.

test/date.unit.js
```diff
@@ -25,6 +25,21 @@ describe('date.js', function () {
...
       assert.ok(date < refDate && date > new Date()); // date should be before date given but after the current time
     });
+
+    it('should return deterministic results when seeded', function () {
+      var refDate = new Date();
+
+      faker.seed(100);
+      var firstDate = faker.date.recent(75, refDate.toJSON()).valueOf();
+
+      for (var i = 0; i < 100; i++) {
```
damienwebdev, line 35: This just checks 100 times. Wouldn't it be safe to test against a much larger probability space (last 10000 days) rather than run the test "n" times? 

This also applies to the other tests.

damienwebdev (changes requested):

PR #146, opened 2022-01-14
feat: extended internet email with a new option useFreeProviders

prisis:


test/internet.unit.js
```diff
@@ -41,6 +41,48 @@ describe('internet.js', function () {
...
+
+      freeEmailProviders.forEach((provider) => {
+        assert.notEqual(res, provider);
+      });
+    });
+  });
+
+  describe('onlyFreeEmailProvider', function () {
+    it('returns an email with one of the free providers', function () {
+      var email1 = faker.internet.email(undefined, undefined, undefined, true);
```
Shinigami92, line 61: We should try to prevent something like this by using `options` as one parameter :thinking:

src/internet.ts
```diff
@@ -199,17 +206,32 @@ export class Internet {
...
-    provider ||= this.faker.random.arrayElement(
-      this.faker.definitions.internet.free_email
-    );
+  email(
+    firstName?: string,
+    lastName?: string,
+    provider?: string | string[],
+    useFreeProviders: boolean = false
+  ): string {
+    let _provider = this.faker.internet.domainName();
```
Shinigami92, line 217: Replace `_provider` with `domainName`

`_something` indicates a private field in the legacy JS community

src/internet.ts
```diff
@@ -37,9 +37,16 @@ export class Internet {
           description: 'The last name of the user',
         },
         provider: {
-          type: 'string',
+          type: 'string|string[]',
```
Shinigami92, line 40: Are you sure this is correct?
I would assume that we need to provide two definitions, or is this not JSONSchema?

Shinigami92 (changes requested):

PR #151, opened 2022-01-15
chore: setup improved linting

Shinigami92:


prisis (dismissed):

prisis:
The found issues will be fixed than in a new commit?

Shinigami92:
We may use `continue-on-error` for the last step in CI.
But on the other side that would be like an ignore all the errors :thinking:

import-brain (dismissed):

.eslintrc.js
```diff
@@ -0,0 +1,73 @@
...
+const { defineConfig } = require('eslint-define-config');
+
+module.exports = defineConfig({
+  ignorePatterns: [
+    // Skip self linting
+    '.eslintrc.js',
+
+    // Skip linting generated content and such
+    'coverage/',
+    'dist/',
```
prisis, line 11: can we use a eslintignore file

prisis: Or we use [https://github.com/mysticatea/eslint-gitignore/tree/master](https://github.com/mysticatea/eslint-gitignore/tree/master) and add just the files that are not in gitignore

src/random.ts
```diff
@@ -123,11 +123,10 @@ export class Random {
    * @param object
    * @param field
    */
-  // TODO @Shinigami92 2022-01-11: Not sure if these generic types are correct
-  objectElement<T extends any, Key extends keyof T>(
+  objectElement<T, Key extends keyof T>(
     object: T = { foo: 'bar', too: 'car' } as unknown as T,
-    field: Key
-  ): T[Key] {
+    field: Key | 'key'
```
ST-DDT, line 128: This should probably be `'key' | 'value'`.

Shinigami92: So in reality it is `'key' | !'key'`
IMO we should refactor the function and make it more clear
Something like `takeKey: boolean = false`?
This would be its own issue

ST-DDT (dismissed):

griest024 (dismissed):

.eslintrc.js
```diff
@@ -0,0 +1,70 @@
...
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: ['./tsconfig.lint.json'],
+    sourceType: 'module',
+    warnOnUnsupportedTypeScriptVersion: false,
+  },
+  plugins: ['@typescript-eslint', 'prettier'],
+  rules: {
+    // We may want to use this in the future
+    'no-useless-escape': 'off',
```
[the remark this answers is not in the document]
Shinigami92: Sometimes it's more readable to be explicit, also I currently only toggled some rules off or warning from the recommended set from line 16-19

src/name.ts
```diff
@@ -158,6 +158,8 @@ export class Name {
         if (prefix) {
           return prefix + ' ' + firstName + ' ' + lastName;
         }
+      // TODO @Shinigami92 2022-01-21: Not sure if this fallthrough is wanted
```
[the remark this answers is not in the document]
Shinigami92: Will do in separate PR

scripts/verifyCommit.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/restrict-plus-operands */
```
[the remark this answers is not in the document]
Shinigami92: Will do this in separate PRs

damienwebdev (approved):
lgtm

ST-DDT (approved):

PR #152, opened 2022-01-15
feat!: support tree-shaking

Shinigami92:
BREAKING CHANGE: importing of faker changes

### DIFF
--- a/examples/node/generateMultiLevelMultiLocaleJSON.js
+++ b/examples/node/generateMultiLevelMultiLocaleJSON.js
@@ -2,30 +2,38 @@
 // including name, address, company, date and commerce namespaces, moustache expressions and random element production
 // Using the helper function arr, randomly sized collections of elements are produced in the document.
 
-var faker = require('../../index');
+var faker = require('../../lib').faker;
 var fs = require('fs');
-// produce array with random number of empty elements 
-const arr = (maxNumberOfElements) => new Array(faker.datatype.number({min: 1, max: maxNumberOfElements})).fill()
+// produce array with random number of empty elements
+const arr = (maxNumberOfElements) =>
+  new Array(faker.datatype.number({ min: 1, max: maxNumberOfElements })).fill();
 
-const locales = ["nl","es","de","fr","en_AU"]
-const company = 
-  { "name"        : faker.company.companyName()
-  , "country"     : faker.address.country()
-  , "departments" : arr(8).map(() =>  { faker.locale = faker.random.arrayElement(locales)
-                                        return { "name"     : faker.commerce.department()
-                                               , "location" : faker.fake("{{address.city}} ({{address.country}})")
-                                               , "employees": arr(20).map(() => {
-                                                                 return { "name"     : faker.fake("{{name.firstName}} {{name.lastName}}")
-                                                                        , "job"      : faker.name.jobTitle()
-                                                                        , "hiredate" : faker.date.past(12).toISOString().split('T')[0]
-                                                                        , "salary"   : faker.datatype.number(700, 9000)
-                                                                        }
-                                                               })
-                                               }
-                                     })
-  }                                        
+const locales = ['nl', 'es', 'de', 'fr', 'en_AU'];
+const company = {
+  name: faker.company.companyName(),
+  country: faker.address.country(),
+  departments: arr(8).map(() => {
+    faker.locale = faker.random.arrayElement(locales);
+    return {
+      name: faker.commerce.department(),
+      location: faker.fake('{{address.city}} ({{address.country}})'),
+      employees: arr(20).map(() => {
+        return {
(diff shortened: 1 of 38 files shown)


Shinigami92:
when we merge this, we need to write a migration guide or at least provide a RegExp replace pattern for out users

```ts
// old-js
const faker = require('faker')

// to
const { faker } = require('@faker-js/faker')
// or
const faker = require('@faker-js/faker').faker

// --------------------------------------------

// new-js
import faker from 'faker'
// or
import * as faker from 'faker'

// to
```

examples/node/generateMultiLevelMultiLocaleJSON.js
```diff
@@ -1,31 +1,39 @@
...
-  console.log(JSON.stringify(company))
-  fs.writeFile(__dirname + '/companyDataSet.json',  JSON.stringify(company), function() {
-    console.log("dataSet generated successfully!");
-  });
\ No newline at end of file
+// This example shows the generation of a multilevel object and JSON document using various faker.js features
+// including name, address, company, date and commerce namespaces, moustache expressions and random element production
+// Using the helper function arr, randomly sized collections of elements are produced in the document.
+
+var faker = require('../../lib').faker;
```
Shinigami92, line 5: This is the important line in this file.
I didn't touched the rest of this file, just auto-prettier on save.

examples/node/generateSet.js
```diff
@@ -1,19 +1,26 @@
 var fs = require('fs');
 
-var faker = require('../../index');
-
+var faker = require('../../lib').faker;
```
Shinigami92, line 3: This is the important line in this file.
I didn't touched the rest of this file, just auto-prettier on save.

package.json
```diff
@@ -22,7 +22,7 @@
     "url": "https://github.com/faker-js/faker.git"
   },
   "license": "MIT",
-  "main": "index.js",
+  "main": "lib/index.js",
```
Shinigami92, line 25: This is the most important line of this PR :slightly_smiling_face:

Shinigami92:
This could affect how the browser version is used :thinking: 
I'm not sure but do we need something to do like this in `src/index.ts`?:
```ts
if (typeof window === 'object') {
  window.faker = faker;
}
```

damienwebdev:
We need an integration test running in CI to validate both Node behavior and browser behavior of the legacy scenario before I think we can merge this. I also think that we move this to `v7` to give our dependents some space to focus on their work rather than a breaking change.

Shinigami92:
> We need an integration test running in CI to validate both Node behavior and browser behavior of the legacy scenario before I think we can merge this. I also think that we move this to `v7` to give our dependents some space to focus on their work rather than a breaking change.

To me it's not such a critical change + I would like to do it now, cause the folks that migrate to us need to rewrite the import statements anyways.
So instead of doing it twice, they just need it to do it once.

clarkerican (dismissed):
Looks good to me!

import-brain:
> We need an integration test running in CI to validate both Node behavior and browser behavior of the legacy scenario before I think we can merge this. I also think that we move this to `v7` to give our dependents some space to focus on their work rather than a breaking change.

By semantic versioning, this would be a "next major version" change anyway, because it would be a possible breaking change.

import-brain:
I'll mark this as ```needs tests``` because we should probably write tests for this to make sure it works before we merge it.

MilosPaunovic (dismissed):

Shinigami92:
> > We need an integration test running in CI to validate both Node behavior and browser behavior of the legacy scenario before I think we can merge this. I also think that we move this to `v7` to give our dependents some space to focus on their work rather than a breaking change.
> 
> By semantic versioning, this would be a "next major version" change anyway, because it would be a possible breaking change.

v6 is currently the next major change, we just released alphas, so nothing wrong with that from my POV

> I'll mark this as `needs tests` because we should probably write tests for this to make sure it works before we merge it.

Could you explain what you mean what test do we need? Currently all ~20k are passing and I'm very happy that we have so much.
This PR does not affect the runtime behavior at all, just improve how you import it.

I may provide a regex-pattern and/or a sed-command so folks can easily update.
But IMO we already do need these patterns right now, cause the import already changed from `faker` to `@faker-js/faker`.

src/index.ts
```diff
@@ -271,5 +271,7 @@ export class Faker {
   }
 }
 
-export default Faker;
-module.exports = Faker;
+// since we are requiring the top level of faker, load all locales by default
+export const faker: Faker = new Faker({
+  locales: require('./locales'),
+});
```
Shinigami92, line 277: We need to add `export default faker;` here

JessicaSachs (dismissed):
After the export default line is in, this lgtm

MilosPaunovic (approved):

JessicaSachs (approved):

prisis (approved):

PR #155, opened 2022-01-15
chore: delete old examples and doc folder

damienwebdev:
The new docs can be found in the `docs` folder.


Shinigami92:
Good idea to delete them
Even if we need them, we can just access git history :+1:

Shinigami92:
I see you also deleted the examples folder :thinking: 
We should look if there is somewhere a test or we need to write a test to test that it works in the browser.
Alternative we could say for now we don't support browser version :thinking: 

Did you also check that the `examples/browser/js/faker(.min).js` files are not generated anymore by scripts?
Maybe we can also remove some stuff from gitignore and such.

JessicaSachs:
I plan to write integration tests for parts of the docs at least. We can use the playground for that.

damienwebdev:
@Shinigami92 I did verify the removal of the examples from the gulpfile. https://github.com/faker-js/faker/pull/155/files#diff-5389b9382804b62e7e964159df7c0c792dffb0705a48429d96d1aacd830aef09L28

import-brain (approved):

import-brain (dismissed):

Shinigami92:
image

Let's rebase this PR after we merged #235

Shinigami92:
I additionally cleaned up some files, so this is not just deleting the `doc` and `examples` folder but also format not formatted files and removed the old `logo.png`

Shinigami92 (approved):

doughlass (approved):

prisis (approved):

PR #157, opened 2022-01-15
chore: update license

damienwebdev:
We have retained the MIT License and added explicit reference to the original codebase of Marak.


Shinigami92 (changes requested):
We should add a `LICENSE` file (without .md at the end)
This is the expected default file as GitHub want it
Examples: https://github.com/prettier/plugin-pug/blob/main/LICENSE https://github.com/vitejs/vite/blob/main/LICENSE

damienwebdev:
Agreed @Shinigami92, I have moved the file.

Shinigami92 (approved):

import-brain (approved):

griest024 (approved):

PR #169, opened 2022-01-16
fix: ts support for locales

Shinigami92:


Shinigami92:
Oh... I found out that there a `locale` (without `s` at the end) that exports for each locale one faker instance :eyes: 
But these are surely currently not typed or covered.

1. We should move them into src for now so we can migrate them also to TS
2. As already planned, we should think about a better strategy to load locales

Shinigami92:
You need to review this PR with the commits one by one, cause there is at least one commit that has 2x1700 file changes and is not reviewable

Shinigami92:
This is now rebased and can be approved

Shinigami92:
Needed to use curly brace imports within the locale files, cause the default import is now not the `Faker constructor` but the default `faker` instance
This affects only internal behavior :slightly_smiling_face: 
The external behavior will remain the same, especially after #217

JessicaSachs (approved):

prisis (approved):

PR #182, opened 2022-01-17
feat(person): add `zodiacSign()`

### LINKED ISSUE #175: Add Random Zodiac Signs

We can add random zodiac signs. We can also use the `date.past()` function to generate a random birthday. We can also get the zodiac sign based on birthdate in wherever random details are generated.

### Suggested solution

Do a new module for zodiac signs. The birthday module can be appended to the date module itself.

### Alternative

_No response_

### Additional context

_No response_

luciferreeves:
### New Functions:

- ~~`faker.date.birthdate()`~~ (Moved to a separate PR)
- `faker.zodiac.sign()`

#### `faker.date.birthdate()`:

```js
// generates a random birthdate of a person aged between 18 and 80
console.log(faker.date.birthdate()) // 1970-12-22T05:00:00.000Z

// generates a random birthdate of a person, between two years
console.log(faker.date.birthdate({ min: 1990, max: 2000 });) // 1992-03-06T05:00:00.000Z
```

you can also pass a mode to the birthdate object. mode can be` 'age'` or `'year'`

```js
// generates a random birthdate of a person, between two years
console.log(faker.date.birthdate({ min: 1990, max: 2000, mode: 'year' });) // 1992-03-06T05:00:00.000Z

// generates a random birthdate of a person, between two ages
console.log(faker.date.birthdate({ min: 4, max: 5, mode: 'age' });) // 2017-06-08T04:00:00.000Z
```

#### `faker.zodiac.sign()`:

```js
// returns a random zodiac sign
console.log(faker.zodiac.sign()) // Leo

// takes optional 'birthday' parameter, if provided returns the zodiac sign for the given birthday
console.log(faker.zodiac.sign('1997-12-20')) // Sagittarius
```

**Unit tests for both the functions added.**

**Sorry for the repeated commit found later. Its a refactor, but I have some `zsh` extensions which automatically complete commands in the terminal and it picked up on the last commit message.**

Closes #175


Shinigami92:
I will review this after 6.0.0 was released

pkuczynski:
I would expect `faker.date.birthday` to have a bit different signature:

```js
faker.date.birthday(age: number, refDate?: string | Date)
faker.date.birthday(options: { min: number, max: number }, refDate?: string | Date)
```

So instead of providing a year, I would rather expect to provide an age...

luciferreeves:
> I would expect `faker.date.birthday` to have a bit different signature:
> 
> ```js
> faker.date.birthday(age: number, refDate?: string | Date)
> faker.date.birthday(options: { min: number, max: number }, refDate?: string | Date)
> ```
> 
> So instead of providing a year, I would rather expect to provide an age...

I have added a `mode` options to the `faker.date.birthday` object. Please take a look at the updated description.

ST-DDT:
Please migrate all js files to ts.

pkuczynski:
> Please migrate all js files to ts.

I think it was an accidental commit of `lib/*` files...

luciferreeves:
I removed the older js files

ST-DDT:
Have the examples files been committed intentionally?

luciferreeves:
I didn't notice the example files were removed from the newer structure of the repo. I removed them from this PR too.

src/date.ts
```diff
@@ -204,4 +204,59 @@ export class _Date {
...
+      31,
+      30,
+      31,
+      30,
+      31,
+    ][month - 1];
+
+    const day = this.faker.datatype.number({ min: 1, max: maxDay });
+
+    return new Date(year, month - 1, day);
```
ST-DDT, line 260: The age mode may be off, if the day and month of it are in the future.

I wrote a test to confirm this:

````ts
 it('returns a random birthday between two ages', function () {
      const min = 4;
      const max = 5;
      const now = new Date();
      const earlyBound = new Date();
      earlyBound.setFullYear(now.getFullYear() - max);
      const lateBound = new Date();
      lateBound.setFullYear(now.getFullYear() - min);

        const birthday = faker.date.birthday({ min, max, mode: 'age' });

        // birthday is a date object
        expect(birthday).to.be.an.instanceof(Date);

        // Generated date is between min and max
        expect(birthday).to.be.at.least(earlyBound);
        expect(birthday).to.be.at.most(lateBound);
      }
    });
````

Please check it for correctness.

File: src/date.ts [hunk 2]
```diff
@@ -204,4 +204,59 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  /**
+   * birthday
+   *
+   * @method faker.date.birthday
+   * @param options
```
ST-DDT, line 212: Please write a proper description including a hint regarding default values.

See https://github.com/faker-js/faker/pull/230/files#diff-2809d7d9276cfe7f2007cf47a87c64572e3d0a733ca9988e7cb82275c2360a83R275-R279 (and others) for an example.

src/date.ts
```diff
@@ -204,4 +204,59 @@ export class _Date {
...
+
+    if (mode === 'age') {
+      max =
+        new Date().getFullYear() - options.min || new Date().getFullYear() - 80;
+      min =
+        new Date().getFullYear() - options.max || new Date().getFullYear() - 18;
+    } else {
+      min = options.min || new Date().getFullYear() - 80;
+      max = options.max || new Date().getFullYear() - 18;
+    }
```
ST-DDT, line 231: Please add a comment why you choose these values. If you want to generate only adult birthdays, you have to make sure that it remains true also for the `year` mode

test/date.spec.ts
```diff
@@ -232,4 +216,42 @@ describe('date', () => {
       faker.definitions.date.weekday.abbr_context = backup_abbr_context;
     });
   });
+  describe('birthday', function () {
+    it('returns a random birthday between two years', function () {
+      var min = 1990;
+      var max = 2000;
```
ST-DDT, line 222: Please use `let` or `const` instead of `var` everywhere.

src/zodiac.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * sign
+   *
+   * @method faker.zodiac.sign
+   * @param {birthday} birthday - The birthday of the person.
+   * @returns {string}
+   */
+  sign(birthday: string = undefined): string {
```
ST-DDT, line 40: Maybe support also `Date` input.

ST-DDT (changes requested):

src/date.ts
luciferreeves: Fixed the date function

src/date.ts
luciferreeves: Added the description

test/date.spec.ts
luciferreeves: replaced all `var` with `const`

src/zodiac.ts
luciferreeves: Supported `Date` input

ST-DDT (changes requested):

src/date.ts
```diff
@@ -204,4 +204,39 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  /**
+   * returns a random birthday
+   *
+   * @method faker.date.birthday
+   * @param options
```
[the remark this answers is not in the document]
pkuczynski: Don't forget about `@example` as for example in here https://github.com/faker-js/faker/blob/main/src/time.ts#L7

pkuczynski (changes requested):

test/date.spec.ts
```diff
@@ -232,4 +216,62 @@ describe('date', () => {
...
+
+      // Generated date is between min and max
+      expect(birthday.getFullYear()).to.be.at.least(min);
+      expect(birthday.getFullYear()).to.be.at.most(max);
+    });
+    it('returns a random birthday between two ages', function () {
+      const min = 4;
+      const max = 5;
+
+      const birthday = faker.date.birthday({ min: min, max: max, mode: 'age' });
```
pkuczynski, line 237:
```suggestion
      const birthday = faker.date.birthday({ min, max, mode: 'age' });
```

luciferreeves:
Everything that was suggested is added. Please check

src/date.ts
```diff
@@ -204,4 +204,40 @@ export class _Date {
...
+      max = new Date().getFullYear() - (options.min ?? 80);
+    } else {
+      min = options.min ?? new Date().getFullYear() - 80;
+      max = options.max ?? new Date().getFullYear() - 18;
+    }
+
+    const startDate = now.setFullYear(min);
+    const endDate = now.setFullYear(max);
+
+    return new Date(+startDate + Math.random() * (endDate - startDate));
```
ST-DDT, line 241: Please use `datatype.number` instead of random to support seeded random.
https://github.com/faker-js/faker/blob/32b483920450a5c1a084e83723f1c433db8ff34f/src/datatype.ts#L27-L29

ST-DDT (changes requested):
One last part, otherwise it looks good to me.
Thanks for your contribution.
Please note that this is scheduled for 6.1 so it might take a while before this gets merged.

luciferreeves (commented):
Okay, everything is fixed now. Instead of `datatype.number()` as suggested, I used `this.faker.datatype.float({
          min: 0,
          max: 1,
          precision: 0.0000000000000001
       })` to replicate the `Math.random()` behavior.

src/date.ts
luciferreeves: Okay, everything is fixed now. Instead of `datatype.number()` as suggested, I used `this.faker.datatype.float({
          min: 0,
          max: 1,
          precision: 0.0000000000000001
       })` to replicate the `Math.random()` behavior.

pkuczynski:
General comment: you should separate zodiac signs and birthday into two separate PR, as they are totally independent.

luciferreeves:
> General comment: you should separate zodiac signs and birthday into two separate PR, as they are totally independent.

@pkuczynski Maybe from next time, I will target smaller issues/PRs. Too many commits have gone in this PR and it would be hard to undo those changes, split branches and redo again.

pkuczynski:
> @pkuczynski Maybe from next time, I will target smaller issues/PRs. Too many commits have gone in this PR and it would be hard to undo those changes, split branches and redo again.

You don't need to split branches. Just create a new one and copy 3 files from this PR regarding date, create new PR from them, and delete in this PR. We don't need the whole commit/review history.

luciferreeves:
I see. Although, I still fail to see the usage of 2 PRs now since this once is almost completed review and is targeted for v6.2 and won't be merged anytime soon either. I will try to split the PRs if I get time.

src/date.ts
```diff
@@ -204,4 +204,48 @@ export class _Date {
...
+    const endDate = now.setFullYear(max);
+
+    return new Date(
+      +startDate +
+        this.faker.datatype.float({
+          min: 0,
+          max: 1,
+          precision: 0.0000000000000001,
+        }) *
+          (endDate - startDate)
```
ST-DDT, line 248: `this.faker.datatype.number(startDate, endDate);`

ST-DDT (changes requested):

ST-DDT:
> Okay, everything is fixed now. Instead of `datatype.number()` as suggested, I used `this.faker.datatype.float({ min: 0, max: 1, precision: 0.0000000000000001 })` to replicate the `Math.random()` behavior.

Can you explain why? AFAIK datatype.number should return a value between startDate and endDate.

ST-DDT:
I replaced your return statement with

````ts
return new Date(
      this.faker.datatype.number({
        min: startDate,
        max: endDate,
      })
    );
````

And the tests still pass.

**Comment by luciferreeves** [comment]
Okay, done. I used:

```js
return new Date(
  this.faker.datatype.number({
    min: startDate,
    max: endDate,
  })
);
```

ST-DDT (dismissed):

pkuczynski (changes requested):

src/date.ts
```diff
@@ -204,4 +204,45 @@ export class _Date {
...
+    } else {
+      min = options.min ?? new Date().getFullYear() - 80;
+      max = options.max ?? new Date().getFullYear() - 18;
+    }
+
+    const startDate = now.setFullYear(min);
+    const endDate = now.setFullYear(max);
+
+    return new Date(
+      this.faker.datatype.number({
```
pkuczynski, line 242: Why not faker.date.between?

pkuczynski (dismissed):

src/date.ts
luciferreeves: Tried, does not work with start and end dates. `faker.datatype.number()` as suggested by @ST-DDT works

src/zodiac.ts
```diff
@@ -0,0 +1,81 @@
...
+      } else if ((month === 11 && day > 21) || (month === 12 && day <= 21)) {
+        return this.faker.definitions.zodiac.sign[10];
+      } else if ((month === 12 && day > 21) || (month === 1 && day <= 20)) {
+        return this.faker.definitions.zodiac.sign[11];
+      } else {
+        throw new Error('Invalid date');
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
luciferreeves: I do not have any idea of those. Probably add it with a later PR?

src/date.ts
```diff
@@ -204,4 +204,48 @@ export class _Date {
...
+   *
+   * @param options The options to use to generate the birthday. If no options are set, then an age between 18 and 80 (inclusive) is generated.
+   *
+   * There are two modes available 'age' and 'year' (default).
+   * - 'age': The min and max options define the age of the person (e.g. 18 - 42)
+   * - 'year': The min and max options define the range the birthdate may be in. (e.g. 1900 - 2000)
+   */
+  birthday(options?: { min?: number; max?: number; mode?: 'age' | 'year' }) {
+    // Generate a random birthday date between two years, min and max
+    options = options || {};
```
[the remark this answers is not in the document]
luciferreeves: Just used the same syntax used at other places.

luciferreeves:
I have resolved all the notified issues, and added comments to the unresolved ones.

src/zodiac.ts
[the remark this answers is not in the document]
ST-DDT: So how do we disambiguate them?

westernZodiac, chineseZodiac, ...?

src/date.ts
ST-DDT: The between method currently requires strings. Support for dates will be added in https://github.com/faker-js/faker/pull/200/files#diff-cdc74ab5a2d5c1414321a9e339b60e2e79e1018a99e545d87653e3e0d972fed6R63

src/date.ts
```diff
@@ -204,4 +204,56 @@ export class _Date {
...
+   *
+   * There are two modes available 'age' and 'year' (default).
+   * - 'age': The min and max options define the age of the person (e.g. 18 - 42)
+   * - 'year': The min and max options define the range the birthdate may be in. (e.g. 1900 - 2000)
+   * @example
+   * faker.date.birthdate(); // 2016-08-12T03:24:00.000Z
+   * faker.date.birthdate({min: 18, max: 65, mode: 'age'}); // 1994-02-11T03:24:00.000Z
+   * faker.date.birthdate({min: 1900, max: 2000, mode: 'year'}); // 1995-02-11T03:24:00.000Z
+   */
+  birthday(options?: {
```
ST-DDT, line 224: You forgot a `birthday` here.

src/zodiac.ts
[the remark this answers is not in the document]
luciferreeves: I am no expert in astrology but I think there is no such thing as "western zodiac" - there's only one - the internationally accepted one. The chinese and others are the "other variants". So we can keep the zodiac as it is, and just add functions for chinese, sidereal, etc.

src/date.ts
luciferreeves: Renamed it.

luciferreeves:
Tests pass on my system but are failing here. I am looking into it. Please disregard the review request.

ST-DDT:
You might have to rebase.

luciferreeves:
Okay, I think this is now ready to be reviewed.

src/zodiac.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (!birthdate) {
+      return this.faker.random.arrayElement(this.faker.definitions.zodiac.sign);
+    } else {
+      // Birthday is a string, parse it as a date string
+      const date =
+        typeof birthdate === 'string' ? new Date(birthdate) : birthdate;
+
+      // Get the month and day of the month
+      const month = date.getMonth() + 1;
+      const day = date.getDate() + 1;
```
ST-DDT, line 37: Please check whether my change is correct:

```suggestion
      const day = date.getDate();
```

Code used for testing with (probably) expected values:

````ts
console.log("19/3", faker.zodiac.sign("0000-03-19T00:00:00Z")); // Pisces
console.log("20/3", faker.zodiac.sign("0000-03-20T00:00:00Z")); // Pisces
console.log("21/3", faker.zodiac.sign("0000-03-21T00:00:00Z")); // Aries

console.log("19/4", faker.zodiac.sign("0000-04-19T00:00:00Z")); // Aries
console.log("20/4", faker.zodiac.sign("0000-04-20T00:00:00Z")); // Aries
console.log("21/4", faker.zodiac.sign("0000-04-21T00:00:00Z")); // Taurus
````

File: src/date.ts [hunk 15]
```diff
@@ -256,4 +256,62 @@ export class _Date {
...
+    }
+
+    const startDate = now.setFullYear(min);
+    const endDate = now.setFullYear(max);
+
+    return new Date(
+      this.faker.datatype.number({
+        min: startDate,
+        max: endDate,
+      })
```
ST-DDT, line 314: This yields wrong values if `min === max` because then it will always return now/refDate instead of a one year range within the bounds. I assume the same applies for the actual date range since it will always miss the one year.

--------------

````ts
faker.date.birthdate({
    mode: "age",
    min: 5,
    max: 5,
    refDate: "2020-01-01T00:00:00.000Z",
  })
````

Actual result:
`2015-01-01T00:00:00.000Z`

Expected result range:
`2014-01-01T00:00:00.001Z` - `2015-01-01T00:00:00.000Z`
(for safety reasons we should probably use `2014-01-02` - ...)

--------------------

````ts
faker.date.birthdate({
    mode: "year",
    min: 2000,
    max: 2000,
    refDate: "2020-01-01T00:00:00.000Z",
  })
````

Actual result:
`2000-01-01T00:00:00.000Z`

Expected result range:
`2000-01-01T00:00:00.000Z` - `2000-12-12T59:59:59.999Z`

**Review by ST-DDT — CHANGES_REQUESTED** [review]

**Comment by ST-DDT** [comment]
Could you please also include the source for the zodiac sign dates above the switch case?
There seems to be different dates based on the page you visit (e.g. German wiki vs English wiki). 😱

**Comment by Shinigami92** [comment]

**Comment by ST-DDT** [comment]
I updated the branch and will fix remaining issues.

**Review by ST-DDT — DISMISSED** [review]
I applied all fixes, so it should be ready now. Please review.

File: src/locales/en/zodiac/sign.ts [hunk 16]
```diff
@@ -0,0 +1,14 @@
...
+  'Taurus',
+  'Gemini',
+  'Cancer',
+  'Leo',
+  'Virgo',
+  'Libra',
+  'Scorpio',
+  'Sagittarius',
+  'Capricorn',
+];
```
ST-DDT, line 14: Should we use a record here with sign name keys instead of index access?

src/date.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
...
+      min = new Date(refDate).setUTCFullYear(refYear - (options.max ?? 80) - 1);
+      max = new Date(refDate).setUTCFullYear(refYear - (options.min ?? 18));
+    } else {
+      // Avoid generating dates the first and last date of the year.
+      min = new Date(Date.UTC(0, 0, 2)).setUTCFullYear(
+        options.min ?? refYear - 80
+      );
+      max = new Date(Date.UTC(0, 11, 30)).setUTCFullYear(
+        options.max ?? refYear - 18
+      );
```
ST-DDT, line 305: Should we avoid the change of the year or not.
How should we handle timezones (local or UTC), if not?

src/modules/date/index.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
 
     return this.faker.helpers.arrayElement(source[type]);
   }
+
+  /**
+   * Returns a random birthdate.
+   *
+   * @param options The options to use to generate the birthdate. If no options are set, then an age between 18 and 80 (inclusive) is generated.
```
Shinigami92, line 261:
```suggestion
   * @param options The options to use to generate the birthdate. If no options are set, an age between 18 and 80 (inclusive) is generated.
```

src/modules/date/index.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
...
+   * @param options.min The minimum age or year to generate a birthdate.
+   * @param options.max The maximum age or year to generate a birthdate.
+   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `now`.
+   * @param options.mode The mode to generate the birthdate. Supported modes are 'year' and 'age'.
+   *
+   * There are two modes available `'age'` and `'year'`:
+   * - `'age'`: The min and max options define the age of the person (e.g. `18` - `42`).
+   * - `'year'`: The min and max options define the range the birthdate may be in (e.g. `1900` - `2000`).
```
Shinigami92, line 269: We should stick to one order:
- age or year
- year or age

src/modules/date/index.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
...
+   *
+   * Defaults to `year`.
+   *
+   * @example
+   * faker.date.birthdate(); // 2016-08-12T03:24:00.000Z
+   * faker.date.birthdate({min: 18, max: 65, mode: 'age'}); // 1994-02-11T03:24:00.000Z
+   * faker.date.birthdate({min: 1900, max: 2000, mode: 'year'}); // 1995-02-11T03:24:00.000Z
```
Shinigami92, line 276:
```suggestion
   * faker.date.birthdate({ min: 18, max: 65, mode: 'age' }); // 1994-02-11T03:24:00.000Z
   * faker.date.birthdate({ min: 1900, max: 2000, mode: 'year' }); // 1995-02-11T03:24:00.000Z
```

Shinigami92: We should add one example with given min/max but not mode.
Maybe also one with `refDate`?

src/modules/date/index.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
...
+    // If no min or max is specified, generate a random date between (now - 80) years and (now - 18) years respectively
+    // So that people can still be considered as adults in most cases
+
+    let min: number;
+    let max: number;
+    if (mode === 'age') {
+      min = new Date(refDate).setUTCFullYear(refYear - (options.max ?? 80) - 1);
+      max = new Date(refDate).setUTCFullYear(refYear - (options.min ?? 18));
+    } else {
+      // Avoid generating dates the first and last date of the year.
```
Shinigami92, line 299: Would be nice if this comments also point out WHY

src/modules/date/index.ts
```diff
@@ -253,4 +254,66 @@ export class _Date {
...
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be larger then min ${min}.`);
+    }
+
+    return new Date(
+      this.faker.datatype.number({
+        min,
+        max,
+      })
+    );
```
Shinigami92, line 317: This can be written in one line

```suggestion
    return new Date(this.faker.datatype.number({ min, max }));
```

Or do we need `new Date(Date.UTC(this.faker.datatype.number({ min, max }), 0, 1))` here?

src/modules/zodiac/index.ts
```diff
@@ -0,0 +1,95 @@
...
+        case 11:
+          return day > 21
+            ? this.faker.definitions.zodiac.sign[10]
+            : this.faker.definitions.zodiac.sign[9];
+        case 12:
+          return day > 21
+            ? this.faker.definitions.zodiac.sign[11]
+            : this.faker.definitions.zodiac.sign[10];
+        default:
+          throw new Error(`Invalid date: ${date.toLocaleDateString()}`);
```
Shinigami92, line 91: Not sure if `toLocaleDateString` is a good idea, because it can differ very much for different countries
In addition we should throw a `FakerError`

Shinigami92: We need a test to cover this line and check the thrown error

Shinigami92 (changes requested):

src/modules/date/index.ts
ST-DDT: > Or do we need `new Date(Date.UTC(this.faker.datatype.number({ min, max }), 0, 1))` here?

No

src/modules/zodiac/index.ts
```diff
@@ -0,0 +1,74 @@
...
+          return day > 22 ? sign.libra : sign.virgo;
+        case 10:
+          return day > 22 ? sign.scorpio : sign.libra;
+        case 11:
+          return day > 21 ? sign.sagittarius : sign.scorpio;
+        case 12:
+          return day > 21 ? sign.capricorn : sign.sagittarius;
+        default:
+          throw new FakerError(`Invalid date: ${date.toDateString()}`);
+      }
```
Shinigami92, line 71:
```suggestion
      const sign: keyof ZodiacDefinitions['sign'] = {
        1: () => (day > 20 ? 'aquarius' : 'capricorn'),
        2: () => (day > 19 ? 'pisces' : 'aquarius'),
        3: () => (day > 20 ? 'aries' : 'pisces'),
        4: () => (day > 20 ? 'taurus' : 'aries'),
        5: () => (day > 20 ? 'gemini' : 'taurus'),
        6: () => (day > 21 ? 'cancer' : 'gemini'),
        7: () => (day > 22 ? 'leo' : 'cancer'),
        8: () => (day > 22 ? 'virgo' : 'leo'),
        9: () => (day > 22 ? 'libra' : 'virgo'),
        10: () => (day > 22 ? 'scorpio' : 'libra'),
        11: () => (day > 21 ? 'sagittarius' : 'scorpio'),
        12: () => (day > 21 ? 'capricorn' : 'sagittarius'),
      }[month]();

      return this.faker.definitions.zodiac.sign[sign];
```

src/modules/zodiac/index.ts
```diff
@@ -0,0 +1,74 @@
+import type { Faker } from '../.../../..';
+import { FakerError } from '../.../../..';
```
Shinigami92, line 2:
```suggestion
import type { Faker } from '../..';
import { FakerError } from '../..';
```

ST-DDT:
<details>
<summary><strong>Docs Preview (Click to expand)</strong></summary>

grafik

grafik

</details>

Shinigami92:
> Please check whether that `FakerError` is generally unsafe to use or just from inside our tests.

at least in the current version (v6.3.1) it is possible 🤔 
image

if you like you could add an extra test that checks importing the error, but I _trust_ esbuild 🤷

src/modules/zodiac/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        7: () => (day > 22 ? 'leo' : 'cancer'),
+        8: () => (day > 22 ? 'virgo' : 'leo'),
+        9: () => (day > 22 ? 'libra' : 'virgo'),
+        10: () => (day > 22 ? 'scorpio' : 'libra'),
+        11: () => (day > 21 ? 'sagittarius' : 'scorpio'),
+        12: () => (day > 21 ? 'capricorn' : 'sagittarius'),
+      }[month]();
+
+      return this.faker.definitions.zodiac.sign[sign];
+    }
```
Shinigami92, line 57: Just a quick question: what's the Faker randomness here? Not sure if I overlook something but I think it will always return the exact same result for each `seed` with a given `birthdate`.
Is this even in scope of Faker?

test/zodiac.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+          const signFromDateObject = faker.zodiac.sign(dateAsDateObject);
+          expect(arrayOfSigns).toContain(signFromDateObject);
+          expect(signFromString).toBe(signFromDateObject);
+        });
+      });
+    }
+  });
+
+  describe('sign()', () => {
+    it('returns a random zodiac sign for a given birthday', () => {
```
Shinigami92, line 84: Looks like it's not random...

ST-DDT:
> if you like you could add an extra test that checks importing the error, but I _trust_ esbuild

It might be only an issue if we use it from the faker modules itself (circular reference: module -> error@index.ts -> faker -> module).

src/modules/zodiac/index.ts
Shinigami92: Ah I see, should we add that to the function docs?
Something like: use the param together with faker.date.birthdate.

ST-DDT: No, I dont think so.
We dont have it for emails either.

Shinigami92: > No, I dont think so.
> 
> We dont have it for emails either.

❓
Could you link the code you mean that doesn't generate random values with given params?

ST-DDT: https://github.com/faker-js/faker/blob/3ea64ceeecfe4cac2f2e6708f12a4bee288d1cb6/src/modules/internet/index.ts#L59-L73

The provider is taken as is, the first and last name are transmuted a bit.
Neither of the parameters claim, that you should use a faker method to generate them.

test/zodiac.spec.ts
ST-DDT: Added a "random" test in [`3b55f84` (#182)](https://github.com/faker-js/faker/pull/182/commits/3b55f848ef0f28efbd3179063e1abb6f1836b312)

ST-DDT:
@faker-js/members @faker-js/maintainers: As per @Shinigami92's request:
Should we have `faker.zodiac.sign(birthday)` as option or is that too much utility method.
(We would still have `faker.zodiac.sign()`)

import-brain:
> @faker-js/members @faker-js/maintainers: As per @Shinigami92's request: Should we have `faker.zodiac.sign(birthday)` as option or is that too much of a utility method. (We would still have `faker.zodiac.sign()`)

Hmm, I don't know. I'm leaning towards it being too much of a utility method, but let's see what the other members of the team think.

xDivisionByZerox:
> @faker-js/members @faker-js/maintainers: As per @Shinigami92's request: Should we have `faker.zodiac.sign(birthday)` as option or is that too much of a utility method. (We would still have `faker.zodiac.sign()`)

Wouldn't that negate the use case of your lib to generate random data? From my understanding providing a specific birthday would result in the same return value for every call (under the condition that the parameter is the same).  If that association is correct, I would suggest dropping the signature.

ST-DDT:
> From my understanding providing a specific birthday would result in the same return value for every call (under the condition that the parameter is the same). If that association is correct, I would suggest dropping the signature.

The assumption is correct, just keep in mind that you can use faker to generate the birthdate that you pass in there.

Shinigami92:
I immediately found several packages on npm helping to get zodiac signs
- https://www.npmjs.com/package/zodiac-signs
- https://www.npmjs.com/package/horoscope
- https://www.npmjs.com/package/zodiac-ts
- https://www.npmjs.com/package/get-zodiac-sign

Without looking to deep into them, I would vote that we remove the signature `sign(birthdate)` and only provide `sign()`

import-brain:
> I immediately found several packages on npm helping to get zodiac signs
> 
> * [npmjs.com/package/zodiac-signs](https://www.npmjs.com/package/zodiac-signs)
> * [npmjs.com/package/horoscope](https://www.npmjs.com/package/horoscope)
> * [npmjs.com/package/zodiac-ts](https://www.npmjs.com/package/zodiac-ts)
> * [npmjs.com/package/get-zodiac-sign](https://www.npmjs.com/package/get-zodiac-sign)
> 
> Without looking to deep into them, I would vote that we remove the signature `sign(birthdate)` and only provide `sign()`

Yeah, I vote for just `sign()` too.

pkuczynski:
It makes sense to have a bday param. It wouldn't generate ransom value, but it would let you easily hook into previously generated random date to produce valid data without using another package

xDivisionByZerox:
> It makes sense to have a bday param. It wouldn't generate ransom value, but it would let you easily hook into previously generated random date to produce valid data without using another package

From our About section
> Generate massive amounts of fake data in the browser and node.js

I don't see how this aligns with the libraries philosophy

pkuczynski:
It doesn't go against it either. Fake does not mean it has to be invalid. In my projects I use faker in a way that produced output is sensible. It's fake, but correct.

ST-DDT:
> In my projects I use faker in a way that produced output is sensible. It's fake, but correct.

This also aligns with my expectations. If you dont generate correct data, why distiguish between firstnames and lastnames or names in general? Why can you generate an email from them?

ST-DDT:
I have split the `birthdate` part in its own PR to simplify the review (and discussions): #962

ST-DDT:
The birthdate has been merged, so this is now only the zodiac signs.

ST-DDT (dismissed):

ejcheng (dismissed):

src/modules/zodiac/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Returns a random zodiac sign.
+   *
+   * @param birthdate The optional birthdate of the person.
+   *
+   * @example
+   * faker.zodiac.sign() // 'Pisces'
+   * faker.zodiac.sign('01/01/1980') // 'Aquarius'
+   */
+  sign(birthdate?: string | number | Date): string {
```
Shinigami92, line 27: Until we meet together this week for a team meeting, I want to block this.
I'm still the opinion that we should not allow to pass a birthdate and get a non seeded static value out if that. Reason: not in scope of faker -> use other libs.
(This is a ground-discussion / precedent case / what is the scope of faker)

Shinigami92 (changes requested):

ST-DDT:
**Team decision**

- We will soon rename the name module to person and then add the zodiac method there.
- The zodiac method won't provide parameters for dates, so it won't be birthdate related.

Shinigami92:
We might want to discuss in team if we want just `zodiacSign` that accepts `'western' | 'chinese'` or if we want two separate methods :thinking:

ejcheng:
> We might want to discuss in team if we want just `zodiacSign` that accepts `'western' | 'chinese'` or if we want two separate methods 🤔

I would prefer just having `zodiacSign` that accepts `'western' | 'chinese'` or `'western' | 'eastern'`.

Shinigami92:
@luciferreeves if you like to, you can already at least move the method to the person module now after a rebase

If you do not have time, please let me know and I can take this PR over

ST-DDT:
**Team decision**

We will start with `zodiacSign()`, then we are free to introduce parameters with the specific variant.
The locale will use `person.zodiac(sign).western` to ensure, future extensibility.

ST-DDT:
Done. Ready for Review.

I used `faker.definitions.person.western_zodiac_sign` as that has better fallback behavior over `faker.definitions.person.zodiac_sign.western`.

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (commented):
Could you at least add german locales as well? :slightly_smiling_face:

ST-DDT:
> Could you at least add german locales as well? 🙂

I will do that in a separate PR.

Shinigami92 (approved):

PR #183, opened 2022-01-17
refactor: Use datatype.number() function instead of random.number()

luciferreeves:
Use `datatype.number()` function to generate a number in [src/internet.ts](https://github.com/faker-js/faker/pull/183/files#diff-6ec5c644d528743506612d7d96511ba25dd360836c2139e73d772fe10a25be73), instead of the deprecated `random.number()` function.


import-brain (approved):

MilosPaunovic (approved):

Shinigami92:
Oh, at first I thought this would change runtime behavior, because the seed could be different.
But:
1. This function was changed anyway due to the image provider was changed to the IPFS one
2. `random.number` just forwards the call to `datetype.number`

:+1:

Shinigami92 (approved):

prisis (approved):

PR #186, opened 2022-01-17
feat: migrate mocha

ericjeker:


package.json
```diff
@@ -33,7 +33,8 @@
     "format": "prettier --write .",
     "jsdoc": "gulp jsdoc",
     "lint": "echo 'TODO eslint'",
-    "test": "mocha test/*.*.js"
+    "test": "mocha --require ts-node/register ./test/*.spec.ts",
+    "test:js": "mocha test/*.*.js"
```
clarkerican, line 37: I think you're also going to want to add this new command to the Github Testing CI Action (https://github.com/faker-js/faker/blob/main/.github/workflows/ci.yml) to ensure the tests still run during CI :)

ericjeker: I was thinking we could use this branch to migrate the tests in full. Not sure the other contributors wanted to move forward with that. But we can also run both the JS tests and TS tests, so we can PR on main instead, and just increase the number of tests?

clarkerican: It might be nice to just branch off main, that may prevent us from losing tests from other PRs that may concurrently add.

Shinigami92:
I will close this because we will directly migrate the tests to vitest #235

Issue #187, opened 2022-01-17
Lib folder now needs built before running tests


clarkerican:
It seems like now we need to build the lib folder when running tests. We should consider either:

1. Updating the README to include this in the Running Tests section 
2. Adding a "pretest" step that automatically builds the lib folder for the developer

Shinigami92:
I think we now documented this in contribution.md and it doesn't count that much on project "stability"
I will close this for now
Feel free to reopen if really needed and update the issue

PR #153, opened 2022-01-15
chore: switch to pnpm

Shinigami92:

PR #188, opened 2022-01-17
chore: auto-build lib folder before running tests

clarkerican:
Issue #187

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,6 @@ jobs:
           node-version: '16'
           cache: 'npm'
       - run: npm ci
-      - run: npm run build
       - run: npm run test
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,6 @@ jobs:
           node-version: '16'
           cache: 'npm'
       - run: npm ci
-      - run: npm run build
```
clarkerican, line 20: I could leave this in just to be safe and explicit.

Shinigami92 (changes requested):
In [#153](https://github.com/faker-js/faker/pull/153/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R32) I introduced a common pattern like `<task>:ci`, I think I will change that to `ci:<task>` today!

So instead of having a `pretest` that could mess-up stuff or just run stuff twice without wanted, we could simply have a `"ci:test": "run-s build test"` and link that in ci.yml

damienwebdev:
@clarkerican I think that the docs have been updated to cover the scenario you ran into. I'm going to close this PR for now unless the `build` before `test` problem becomes really unbearable.

PR #191, opened 2022-01-17
feat: Add random binary and octal number generation

### LINKED ISSUE #184: feat: add binary and octal random number generation

`faker.random.[...]` already generates random number, bigInt, float, hexadecimal, etc. But it lacks octal and binary numbers. This can be added. 

### Suggested solution

This is an easy task. Either use the [`faker.datatype.hexadecimal()`](https://github.com/faker-js/faker/blob/4a69d6dc295153e037c15e084e55aef1e662096b/src/datatype.ts#L179) method as a reference or generate a random number using [`faker.datatype.number()`](https://github.com/faker-js/faker/blob/4a69d6dc295153e037c15e084e55aef1e662096b/src/datatype.ts#L27) and convert it into binary or octal.

A simple way to convert decimal to binary is:

```js
console.log(Number((128).toString(2))) // 10000000
```

Similarly decimal to octal conversion is:

```js
console.log(Number((128).toString(8))) // 200
```

(this is just one method to do so)

### Alternative

_No response_

### Additional context

_No response_

mbokinala:
closes #184


src/datatype.ts
```diff
@@ -208,6 +208,45 @@ export class Datatype {
...
   }
 
+  /**
+   * Returns random octal string of specified length.
+   *
+   * @method faker.datatype.octal
+   * @param count defaults to 1
+   */
+  octal(count: number = 1): string {
+    let wholeString = '';
```
poyea, line 218:
```suggestion
    let octalString = '';
```

src/datatype.ts
```diff
@@ -208,6 +208,45 @@ export class Datatype {
...
+  }
+
+  /**
+   * Returns random binary string of specified length.
+   *
+   * @method faker.datatype.binary
+   * @param count defaults to 1
+   */
+  binary(count: number = 1): string {
+    let wholeString = '';
```
poyea, line 242:
```suggestion
    let binaryString = '';
```

test/datatype.unit.js
```diff
@@ -232,6 +232,34 @@ describe('datatype.js', function () {
...
+    var binary = faker.datatype.binary;
+
+    it('generates a single binary charcter when no additional argument is provided', function () {
+      var binaryNum = binary();
+      assert.ok(binaryNum.match(/^(0b)[01]{1}$/i));
+    });
+
+    it('generates a random binary string', function () {
+      var hex = binary(12);
+      assert.ok(hex.match(/^(0b)[01]+$/i));
```
poyea, line 245:
```suggestion
      assert.ok(hex.match(/^0b[01]{12}$/i));
```

test/datatype.unit.js
```diff
@@ -232,6 +232,34 @@ describe('datatype.js', function () {
     });
   });
 
+  describe('binary', function () {
+    var binary = faker.datatype.binary;
+
+    it('generates a single binary charcter when no additional argument is provided', function () {
+      var binaryNum = binary();
+      assert.ok(binaryNum.match(/^(0b)[01]{1}$/i));
```
poyea, line 240:
```suggestion
      assert.ok(binaryNum.match(/^0b[01]$/i));
```

src/datatype.ts
```diff
@@ -208,6 +208,45 @@ export class Datatype {
     return '0x' + wholeString;
   }
 
+  /**
+   * Returns random octal string of specified length.
+   *
+   * @method faker.datatype.octal
+   * @param count defaults to 1
+   */
+  octal(count: number = 1): string {
```
poyea, line 217: I'd suggest `length`, but it's OK for consistency with `hexaDecimal`.

docs/api/datatype.md
```diff
@@ -123,3 +123,37 @@ faker.datatype.hexaDecimal(); // '0xA' (string)
...
 ```
+
+## Binary
+
+Generates random binary (base-2) number.
+
+::: tip
+| Param | Type | Default |
+| ----- | ----- | :---------------: |
+| length | number | 1 |
```
poyea, line 134: and it's `length` here

src/datatype.ts
mbokinala: I agree, I'll update this

docs/api/datatype.md
mbokinala: I modeled the two new methods on the existing `hexaDecimal` method, but it looks like the `hexaDecimal` method uses the `count` parameter while its docs say `length`. I will update all three methods and docs to use `length`.

src/datatype.ts
```diff
@@ -174,11 +174,11 @@ export class Datatype {
    * hexaDecimal
    *
    * @method faker.datatype.hexaDecimal
-   * @param count defaults to 1
+   * @param length defaults to 1
    */
-  hexaDecimal(count: number = 1): string {
+  hexaDecimal(length: number = 1): string {
     let wholeString = '';
```
poyea, line 180: Maybe `hexaString` too? It's very minor though.

poyea (approved):
looks great!

src/datatype.ts
mbokinala: Ok, I'll edit it

Shinigami92:

petrzjunior:
Could we also provide a version without the `0b`, `0o`, `0x` prefix? Possibly via optional parameter.
In my usecase, I need to repeatedly write `.substring(2)` to remove it.

```js
Buffer.from(faker.datatype.hexadecimal(20).substring(2), 'hex')
```

Shinigami92:
> Could we also provide a version without the `0b`, `0o`, `0x` prefix? Possibly via optional parameter.
> 
> In my usecase, I need to repeatedly write `.substring(2)` to remove it.
> 
> 
> 
> ```js
> 
> Buffer.from(faker.datatype.hexadecimal(20).substring(2), 'hex')
> 
> ```

Don't have to much time yet to go into details and I have other prios right now, but yes I would like to not prefix it by default at all but have just the option to add the prefix on demand via an option argument. Consistent for all of these functions!

xDivisionByZerox:
I will close this PR since it's heavily outdated. I suggest opening another PR that at least has the 'new' test strategy and directory structure implemented.

If anyone really wants to keep this alive, feel free to reopen it.

PR #192, opened 2022-01-17
fix(types): fixed datetime types for optional argument

jared-fraser:
Datetime has an optional `options` argument  

source: https://github.com/faker-js/faker/blob/main/src/datatype.ts#L98
test: https://github.com/faker-js/faker/blob/d9b6e5d584f92da4ca059f2705530e609525265e/test/datatype.unit.js#L157

### DIFF
--- a/src/datatype.ts
+++ b/src/datatype.ts
@@ -95,7 +95,7 @@ export class Datatype {
    * @method faker.datatype.datetime
    * @param options pass min OR max as number of milliseconds since 1. Jan 1970 UTC
    */
-  datetime(options): Date {
+  datetime(options?: number | { min?: number; max?: number}): Date {
     if (typeof options === 'number') {
       options = {
         max: options,


src/datatype.ts
```diff
@@ -95,7 +95,7 @@ export class Datatype {
    * @method faker.datatype.datetime
    * @param options pass min OR max as number of milliseconds since 1. Jan 1970 UTC
    */
-  datetime(options): Date {
+  datetime(options?: number | { min?: number | undefined; max?: number | undefined }): Date {
```
Shinigami92, line 98: Optional values don't need an extra `undefined` and we also need the `precision`  cause we forward options to `faker.datatype.number`

```suggestion
  datetime(
    options?: number | { min?: number; max?: number; precision?: number }
  ): Date {
```

jared-fraser: thanks, updated

src/datatype.ts
```diff
@@ -95,7 +95,9 @@ export class Datatype {
    * @method faker.datatype.datetime
    * @param options pass min OR max as number of milliseconds since 1. Jan 1970 UTC
    */
-  datetime(options): Date {
+  datetime(
+    options?: number | { min?: number; max?: number; precision?: number }
```
prisis, line 99: I dont see the use of precision?

Shinigami92: Line 119

Shinigami92 (approved):

prisis (approved):

PR #200, opened 2022-01-17
fix: accept dates as params for Date methods

### LINKED ISSUE #198: Accept dates as params for Date methods
### Describe the bug

Old faker used to support dates for functions like `faker.date.past` or `faker.date.between`. New types do not allow such calls.

### Reproduction

```
const baseDate = new Date()

faker.past(1, baseDate)
```

### Additional Info

_No response_

pkuczynski:
Fixes #198


src/date.ts
```diff
@@ -18,10 +18,10 @@ export class _Date {
    * @param years
    * @param refDate
    */
-  past(years?: number, refDate?: string): Date {
+  past(years?: number, refDate?: string | Date): Date {
     let date = new Date();
     if (typeof refDate !== 'undefined') {
-      date = new Date(Date.parse(refDate));
+      date = refDate instanceof Date ? date : new Date(Date.parse(refDate));
```
Shinigami92, line 24: I assume you wanted to pass `refDate` instead of `date`, do you?

_applies to below too_

Shinigami92 (changes requested):

Shinigami92:
Due to this is a runtime change, I think we will do this in v6.1, because we try to focus on stability of the project itself first (CI, tests, docs etc)

src/date.ts
pkuczynski: Ah! Of course! Sorry! Was quickly editing on Github without running locally. Was hoping unit tests would pick it up :) Fixed.

pkuczynski:
> Due to this is a runtime change, I think we will do this in v6.1, because we try to focus on stability of the project itself first (CI, tests, docs etc)

I only got to this change, because my current code start failing after the upgrade. So this is not strictly a new feature, but restoring functionality working with previous version of faker.

Shinigami92:
Speaking about unit tests, could you add one?

pkuczynski:
> Speaking about unit tests, could you add one?

Done. If you would be using `jest` I could do `it.each` to run the same tests with different examples. Without it, I just had to duplicate some tests

pkuczynski:
Btw. good that you asked about tests, as I discovered a bug, where original date was incorrectly mutated...

LawrenceWooding:
> I only got to this change, because my current code start failing after the upgrade. So this is not strictly a new feature, but restoring functionality working with previous version of faker.

I had a similar experience to @pkuczynski  - attempting to replace the old faker-js package with this new one caused errors in previously working code. I am confident that most (if not all) of the Date functions used to support Date() fields. This is supported by the fact that in the [documentation for Dates all of the functions take Date objects instead of strings](https://fakerjs.dev/api/date.html) (or at least the ones that were already documented).

Speaking of which the docs should probably be updated in this branch to better reflect the type definitions (e.g `string | Date` as opposed to `Date`). I am happy to do this if I get a spare moment.

Thank you @pkuczynski for looking into this issue, it is the primary roadblock between my organisation switching our frontend repository to the community faker-js package.

LawrenceWooding:
I have made a pull request in @pkuczynski 's fork with the updated documentation. :)

pkuczynski:
> I have made a pull request in @pkuczynski 's fork with the updated documentation. :)

And I just merged it, thanks :)

pkuczynski:
@Shinigami92 Is there anything else you want me to do to get this PR accepted, merged and released?

src/date.ts
```diff
@@ -204,4 +187,12 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private createDate(date: string | Date) {
```
Shinigami92, line 191:
```suggestion
  private createDate(date?: string | Date): Date {
```

src/date.ts
```diff
@@ -204,4 +187,12 @@ export class _Date {
...
+    if (typeof date !== 'undefined') {
```
Shinigami92, line 192: For sake of simplicity, I would suggest following:

```suggestion
    if (date != null) {
```

src/date.ts
```diff
@@ -85,18 +76,18 @@ export class _Date {
...
     }
     const newDates: Date[] = [];
-    let fromMilli = Date.parse(from);
-    const dateOffset = (Date.parse(to) - fromMilli) / (num + 1);
-    let lastDate: string | Date = from;
+    const toMilli = to instanceof Date ? to.getTime() : Date.parse(to);
+    let fromMilli = from instanceof Date ? from.getTime() : Date.parse(from);
+    const dateOffset = (toMilli - fromMilli) / (num + 1);
+    let lastDate =
+      from instanceof Date ? new Date(from) : new Date(Date.parse(from));
```
Shinigami92, line 88:
```suggestion
    let lastDate = this.createDate(from);
```

src/date.ts
```diff
@@ -85,18 +76,18 @@ export class _Date {
...
+  betweens(from: string | Date, to: string | Date, num?: number): Date[] {
     if (typeof num == 'undefined') {
       num = 3;
     }
     const newDates: Date[] = [];
-    let fromMilli = Date.parse(from);
-    const dateOffset = (Date.parse(to) - fromMilli) / (num + 1);
-    let lastDate: string | Date = from;
+    const toMilli = to instanceof Date ? to.getTime() : Date.parse(to);
+    let fromMilli = from instanceof Date ? from.getTime() : Date.parse(from);
```
Shinigami92, line 85: You could extract these also to a private helper function

```suggestion
    const toMilli = this.toMillis(to);
    let fromMilli = this.toMillis(from);
```

It could make sense to rename `createDate` to `toDate`

src/date.ts
```diff
@@ -68,13 +60,12 @@ export class _Date {
    * @param from
    * @param to
    */
-  between(from: string, to: string): Date {
-    const fromMilli = Date.parse(from);
-    const dateOffset = this.faker.datatype.number(Date.parse(to) - fromMilli);
+  between(from: string | Date, to: string | Date): Date {
+    const fromMilli = from instanceof Date ? from.getTime() : Date.parse(from);
+    const toMilli = to instanceof Date ? to.getTime() : Date.parse(to);
```
Shinigami92, line 65: Also use helper here

Shinigami92 (changes requested):

Shinigami92:
> Done. If you would be using `jest` I could do `it.each` to run the same tests with different examples. Without it, I just had to duplicate some tests

You will love our next PR incoming :P
See #235

pkuczynski:
> You will love our next PR incoming :P See #235

Citing their docs: "It's not recommended to migrate your current testing setups yet" ;) So I am not sure :) Is it jest based?

Shinigami92:
> Citing their docs: "It's not recommended to migrate your current testing setups yet" ;) So I am not sure :) Is it jest based?

I'm one of Vite core members and in direct contact with Anthony Fu :slightly_smiling_face: 
Also yes, vitest is a somewhat swap-in-out replacement for jest. But written in more modern syntax and stripped out old legacy stuff. So it's jest but in fast ^^

src/date.ts
pkuczynski: Done x2

src/date.ts
pkuczynski: Done

pkuczynski:
@Shinigami92 all feedback applied. Ready for re-review.

pkuczynski:
> > Citing their docs: "It's not recommended to migrate your current testing setups yet" ;) So I am not sure :) Is it jest based?
> 
> I'm one of Vite core members and in direct contact with Anthony Fu 🙂 Also yes, vitest is a somewhat swap-in-out replacement for jest. But written in more modern syntax and stripped out old legacy stuff. So it's jest but in fast ^^

Happy to try it out :)

Shinigami92 (approved):

Shinigami92:
The windows test is not related and can be ignore

Shinigami92:
We merged the vitest branch :partying_face: 
Please re-implement your tests in the new test-suite

pkuczynski:
You are not making my life easy! ;)

Shinigami92:
> You are not making my life easy! ;)

I'm a bit sorry :sweat_smile: 
But we are/were in a process to make the project stable (like milestone for v6.0 is saying)
And all the mini PRs here and there doesn't make my life easy as well :sweat_smile:

pkuczynski:
> I'm a bit sorry 😅 But we are/were in a process to make the project stable (like milestone for v6.0 is saying) And all the mini PRs here and there doesn't make my life easy as well 😅

I can imagine :) Sorry to interrupt your valuable work. Just wanted to fix the issue encountered after upgrade. Didn't know you would be making such significant changes soon...

Anyway, I think I am done. Let me know if there is anything more I should do?

test/date.spec.ts
```diff
@@ -24,6 +26,16 @@ describe('date', () => {
...
+
+    it('accepts reference date given as Date object', () => {
+      const refDate = new Date(2120, 11, 9, 10, 0, 0, 0); // set the date beyond the usual calculation (to make sure this is working correctly)
+
+      const date = faker.date.past(75, refDate);
+
+      // date should be before date given but after the current time
+      expect(date).lessThan(refDate);
+      expect(date).greaterThan(new Date());
+    });
```
pkuczynski, line 38: By the way, I tried to remove test duplication using `it.each`, but it does not seem to be supported? Example test improvement based on jest syntax would be:

```js
    describe('returns a date N years before the date given', () => {
      const refDate = new Date(2120, 11, 9, 10, 0, 0, 0); // set the date beyond the usual calculation (to make sure this is working correctly)

      it.each([
        ['refDate as string', refDate.toJSON()]
        ['refDate as Date', refDate]
      ])('%s', (_, ref) => {
        const date = faker.date.past(75, ref);

        // date should be before date given but after the current time
        expect(date).lessThan(refDate);
        expect(date).greaterThan(new Date());
      })
    });
```

Shinigami92: Progress can be tracked here :slightly_smiling_face: 
https://github.com/vitest-dev/vitest/issues/120

test/date.spec.ts
```diff
@@ -24,6 +26,16 @@ describe('date', () => {
...
     });
+
+    it('accepts reference date given as Date object', () => {
+      const refDate = new Date(2120, 11, 9, 10, 0, 0, 0); // set the date beyond the usual calculation (to make sure this is working correctly)
+
+      const date = faker.date.past(75, refDate);
+
+      // date should be before date given but after the current time
+      expect(date).lessThan(refDate);
+      expect(date).greaterThan(new Date());
```
Shinigami92, line 37: This test would only pass for the next 45 years and then suddenly fail :thinking: 
Should we just hardcode this date?

test/date.spec.ts
```diff
@@ -61,25 +81,39 @@ describe('date', () => {
...
-        days,
-        // @ts-expect-error
-        refDate
-      );
-
       const lowerBound = new Date(
         refDate.getTime() - days * 24 * 60 * 60 * 1000
       );
 
+      const date = faker.date.recent(days, refDate.toJSON());
```
Shinigami92, line 88: **question:** why is toJSON needed here?

Shinigami92: Ah, I think it's better to use `toISOString` here so it's more understandable

test/date.spec.ts
```diff
@@ -93,17 +127,31 @@ describe('date', () => {
...
-
-      const date = faker.date.soon(
-        days,
-        // @ts-expect-error
-        refDate
+      const upperBound = new Date(
+        refDate.getTime() + days * 24 * 60 * 60 * 1000
       );
 
+      const date = faker.date.soon(days, refDate.toJSON());
```
Shinigami92, line 134: Also better use `toISOString` here

test/date.spec.ts
```diff
@@ -120,11 +168,17 @@ describe('date', () => {
       const from = new Date(1990, 5, 7, 9, 11, 0, 0);
       const to = new Date(2000, 6, 8, 10, 12, 0, 0);
 
-      const date = faker.date.between(
-        //@ts-expect-error
-        from,
-        to
-      );
+      const date = faker.date.between(from.toJSON(), to.toJSON());
```
Shinigami92, line 171: And here :wink:

test/date.spec.ts
```diff
@@ -136,11 +190,19 @@ describe('date', () => {
       const from = new Date(1990, 5, 7, 9, 11, 0, 0);
       const to = new Date(2000, 6, 8, 10, 12, 0, 0);
 
-      const dates = faker.date.betweens(
-        // @ts-expect-error
-        from,
-        to
-      );
+      const dates = faker.date.betweens(from.toJSON(), to.toJSON());
```
Shinigami92, line 193: You get the idea

Shinigami92 (changes requested):

test/date.spec.ts
ST-DDT: You can use a normal for loop to achieve the same.

pkuczynski: I know I could use the loop. It would look like this then:

```
    describe('returns a date N years before the date given', () => {
      const refDate = new Date(2120, 11, 9, 10, 0, 0, 0); // set the date beyond the usual calculation (to make sure this is working correctly)

      ([
        ['refDate as string', refDate.toJSON()],
        ['refDate as Date', refDate]
      ] as [string, string|Date][]).forEach(t => it(t[0], () => {
        const date = faker.date.past(75, t[1]);

        // date should be before date given but after the current time
        expect(date).lessThan(refDate);
        expect(date).greaterThan(new Date());
      }));
    });
```

@Shinigami92 is this something you want me to do?

test/date.spec.ts
pkuczynski: 45 years is a pretty long time for a js package ;) This test was like this before, so changing it would not be related to his PR, right?

Let me know what you want me to do about it...

test/date.spec.ts
pkuczynski:
```js
const date = new Date();
console.log(date.toJSON(), date.toISOString())
```

produces `2022-01-23T15:57:05.546Z 2022-01-23T15:57:05.546Z`. So its the same output. `toJSON` was used before in some other tests, so I followed. But I agree `toISOString` looks more readable, so I updated with your suggestion :)

test/date.spec.ts
Shinigami92: I would put the loop outside the `describe` block

pkuczynski: Why outside of the `describe`? This is a single test with two examples, so it makes sense to be wrapped into a single `describe` and loop with `it`.

test/date.spec.ts
Shinigami92: We don't want to hold old not useful code, but just improve the project at all and make it robust
So yeah, please use a all-time stable solution :slightly_smiling_face:

test/date.spec.ts
Shinigami92: yeah, please use it in all places in this file

test/date.spec.ts
Shinigami92: Ah nevermind, you are right

test/date.spec.ts
```diff
@@ -5,20 +5,32 @@ describe('date', () => {
...
 
     it('returns a date N years before the date given', () => {
       const refDate = new Date(2120, 11, 9, 10, 0, 0, 0); // set the date beyond the usual calculation (to make sure this is working correctly)
 
-      const date = faker.date.past(75, refDate.toJSON());
+      const date = faker.date.past(75, refDate.toISOString());
+
+      // date should be before date given but after the current time
+      expect(date).lessThan(refDate);
+      expect(date).greaterThan(new Date());
```
Shinigami92, line 27: Okay, only this needs to be static not and I think we are good to go with this PR :+1:

Shinigami92 (changes requested):

test/date.spec.ts
pkuczynski: I used loops now. Let me know if its any better?

test/date.spec.ts
pkuczynski: Sure. Done.

test/date.spec.ts
pkuczynski: Done!

Shinigami92 (approved):

src/date.ts
```diff
@@ -204,4 +191,20 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private toDate(date?: string | Date): Date {
+    if (date != null) {
+      return new Date(date instanceof Date ? date : Date.parse(date));
```
prisis, line 197: You new to use new Date(date);  
  
if you call Date.parse('2008/03/20') it returns this 1205967600000

src/date.ts
```diff
@@ -204,4 +191,20 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private toDate(date?: string | Date): Date {
+    if (date != null) {
```
prisis, line 196: pls use date !== null

src/date.ts
```diff
@@ -204,4 +191,20 @@ export class _Date {
...
+      return new Date(date instanceof Date ? date : Date.parse(date));
+    }
+
+    return new Date();
+  }
+
+  private toMilliseconds(date?: string | Date): number {
+    if (date != null) {
```
prisis, line 204: pls use date !== null

src/date.ts
prisis: So you can make it simpler return date instanceof Date ? date : new Date(date);

pkuczynski: > You new to use new Date(date);

I am not sure what do you mean? 

```js
const date = new Date();
const date2 = new Date(date);

console.log(date.toISOString(), date2.toISOString())
=> 2022-01-25T23:01:33.558Z 2022-01-25T23:01:33.558Z
```

> if you call Date.parse('2008/03/20') it returns this 1205967600000

Yes, and? 

```js
new Date(Date.parse('2008/03/20')).toISOString()
'2008-03-20T00:00:00.000Z'
```

(assuming you are on +0 TZ)

Shinigami92:
Maybe we should rewrite the the date tests first? #287 
I don't feel comfortable anymore cause of @prisis change requests

src/date.ts
pkuczynski: `!= null` is equivalent of `!== null && !== undefined`. Should I really change it?

pkuczynski:
> Maybe we should rewrite the the date tests first? #287
> I don't feel comfortable anymore cause of @prisis change requests

We can, but bear in mind, that those tests do not use any mocks, so not sure what kind of benefit rewrite would bring?

src/date.ts
prisis: The type says date?: string | Date -> date can be undefined|string|Date

src/date.ts
pkuczynski: I could also do `return date instanceof Date ? date : new Date(date);`, would work pretty much the same way, right?

src/date.ts
pkuczynski: Indeed! But your comment was a bit different initially ;) I can change it no problem :)

src/date.ts
pkuczynski: I actually can't, just checked. I need to clone the date, as some of the methods mutate it...

damienwebdev:
This PR is incomplete compared to DefinitelyTyped, see the original type constraints here:

```
date: {
            past(years?: number, refDate?: string | Date): Date;
            future(years?: number, refDate?: string | Date): Date;
            between(from: string | number | Date, to: string | Date): Date;
            recent(days?: number, refDate?: string | Date): Date;
            soon(days?: number, refDate?: string | Date): Date;
            month(options?: { abbr?: boolean; context?: boolean }): string;
            weekday(options?: { abbr?: boolean; context?: boolean }): string;
        };
``` 

Note, specifically `between` was one that I noted was incorrect.

pkuczynski:
> Note, specifically `between` was one that I noted was incorrect.

I don't think DT was correct. If `from` could be a number (that's what's missing), then `to` should also take `number`. We could possibly accept numbers if it would be a timestamp, as we pass it through `Date.parse`, but then I guess I should add some validations to ensure its really a timestamp and not some random number. Is this something you want me to do?

Shinigami92:
Please rebase this PR when #312 was merged

pkuczynski:
> Please rebase this PR when #312 was merged

I am aware. Waiting for it. :)

ST-DDT:
> We could possibly accept numbers if it would be a timestamp, as we pass it through `Date.parse`, but then I guess I should add some validations to ensure its really a timestamp and not some random number.

Accepting numbers for the dates sounds useful. See also here https://github.com/faker-js/faker/pull/182#discussion_r794980533

I would only validate Number.isInteger. Everything else can be considered date enough.

ST-DDT (approved):

Shinigami92:
I discussed with @pkuczynski that we will fix also the millisecond issue with `Date.parse` in this PR
See: https://discord.com/channels/929487054990110771/938146569381888010/938150659646177342

pkuczynski:
> Accepting numbers for the dates sounds useful. See also here [#182 (comment)](https://github.com/faker-js/faker/pull/182#discussion_r794980533)
> 
> I would only validate Number.isInteger as that can be considered date enough.

https://github.com/faker-js/faker/issues/394

pkuczynski:
This PR is now done and ready for review @ST-DDT @Shinigami92 @prisis

src/date.ts
```diff
@@ -85,21 +78,23 @@ export class _Date {
...
-      // TODO @Shinigami92 2022-01-11: It may be a bug that `lastDate` is passed to parse if it's a `Date` not a `string`
-      // @ts-expect-error
-      fromMilli = Date.parse(lastDate);
-      lastDate = new Date(fromMilli + dateOffset);
+      fromMilliseconds = lastDate.getTime();
+      lastDate = new Date(fromMilliseconds + dateOffset);
       newDates.push(lastDate);
     }
+
     return newDates;
```
Shinigami92, line 98: I thought this would be just something like

```ts
const dates = [];
for (let i = 0; i < num; i++) {
  dates.push(this.between(from, to));
}
return dates.sort();
```

test/date.spec.ts
```diff
@@ -5,42 +5,22 @@ const seededRuns = [
...
-        new Date('2021-07-08T10:07:32.670Z'),
-        new Date('2024-11-19T18:52:06.785Z'),
-        new Date('2024-11-19T18:52:06.074Z'),
-      ],
-      between: [
-        new Date('2021-03-15T19:30:57.091Z'),
-        new Date('2021-07-29T19:19:12.731Z'),
-      ],
+      past: {
+        default: new Date('2020-10-08T00:10:58.041Z'),
```
Shinigami92, line 9: We started to name these `noArgs`, would you like to rename them?

test/date.spec.ts
```diff
@@ -289,7 +229,7 @@ describe('date', () => {
             '2021-04-21T17:11:17.711Z'
```
Shinigami92, line 229: Please remove my todo 3 lines above this

test/date.spec.ts
```diff
@@ -301,7 +241,7 @@ describe('date', () => {
             new Date('2021-04-21T17:11:17.711Z')
```
Shinigami92, line 241: Please remove my todo 3 lines above this

Shinigami92 (changes requested):

test/date.spec.ts
pkuczynski: sure! done.

test/date.spec.ts
pkuczynski: done x2

src/date.ts
pkuczynski: Initially, I tried to make the least amount of changes, but sure, can be done as you suggested. I like it more too...

ST-DDT (approved):

Shinigami92:
@pkuczynski This PR needs a rebase

src/date.ts
```diff
@@ -204,4 +185,20 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private toDate(date?: string | Date): Date {
+    if (date !== undefined) {
```
Shinigami92, line 190: We want to use smart null checks, would be nice if you directly use `date != null` here so I do not need to check this line when merging #650

Shinigami92 (changes requested):

src/date.ts
```diff
@@ -204,4 +185,20 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private toDate(date?: string | Date): Date {
```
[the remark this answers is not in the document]
pkuczynski: OK, will have a look at this tomorrow :)

src/date.ts
pkuczynski: Done!

src/date.ts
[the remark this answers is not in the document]
pkuczynski: Done!

src/date.ts
```diff
@@ -1,6 +1,22 @@
 import type { Faker } from '.';
 import type { DateEntryDefinition } from './definitions';
 
+const toDate = (date?: string | Date): Date => {
+  if (date != null) {
+    return new Date(date instanceof Date ? date : Date.parse(date));
+  }
+
+  return new Date();
+};
```
Shinigami92, line 10: On other places we already used the code-style using functions instead of arrow functions, so lets do here too

```suggestion
function toDate(date?: string | Date): Date {
  if (date != null) {
    return new Date(date instanceof Date ? date : Date.parse(date));
  }

  return new Date();
}
```

src/date.ts
```diff
@@ -1,6 +1,22 @@
...
+
+const toMilliseconds = (date?: string | Date): number => {
+  if (date != null) {
+    return date instanceof Date ? date.getTime() : Date.parse(date);
+  }
+
+  return new Date().getTime();
+};
```
Shinigami92, line 18:
```suggestion
function toMilliseconds(date?: string | Date): number {
  if (date != null) {
    return date instanceof Date ? date.getTime() : Date.parse(date);
  }

  return new Date().getTime();
}
```

Shinigami92 (changes requested):

src/date.ts
pkuczynski: Not a major issue, but sure, why not...

pkuczynski:
Added missing jsdoc to these new functions, after you converted them form arrow. Anything else @Shinigami92?

Shinigami92:
> Anything else @Shinigami92?

I'm in bed now ^^
Will look tomorrow, thx for the JSDocs

ST-DDT (approved):

Shinigami92 (approved):

PR #207, opened 2022-01-18
feat: Generate JSON file with TypeDoc

ericjeker:
I rebased from main as the [other branch](https://github.com/faker-js/faker/pull/82) was too far behind and also because in the meantime I deleted it from my fork to have a clean fork.

Now, here's the catch:

TypeDoc generates documentation only from directly exported types from the entrypoint:
https://github.com/TypeStrong/typedoc/issues/1657

So our sub-classes, like `Address`, are not included in the documentation except if they are exported from `index.ts`. Not sure if this is desirable.

There are workarounds, but I did not try that yet: https://github.com/Gerrit0/typedoc-plugin-missing-exports

As in `Faker` we instantiate all those sub-classes:

```typescript 
readonly address: Address = new Address(this);
```

We could imagine exporting these instances, so ideally we could then access them doing:

```TypeScript
import { Address } from '@faker-js/faker';

const cityName = Address.city();
```

This is up for discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "lint": "echo 'TODO eslint'",
     "test": "vitest",
     "coverage": "vitest --coverage",
-    "typedoc": "typedoc"
+    "typedoc": "ts-node typedoc.ts"
   },
   "simple-git-hooks": {
     "pre-commit": "npx lint-staged --concurrent false",
@@ -79,6 +79,7 @@
     "prettier": "2.5.1",
     "simple-git-hooks": "~2.7.0",
     "through2": "2.0.0",
+    "ts-node": "^10.4.0",
     "typedoc": "^0.22.11",
     "typedoc-plugin-markdown": "^3.11.12",
     "typedoc-plugin-missing-exports": "^0.22.6",
--- a/package.json
+++ b/package.json
@@ -79,14 +79,17 @@
     "prettier": "2.5.1",
     "simple-git-hooks": "~2.7.0",
     "through2": "2.0.0",
-    "typedoc": "^0.22.10",
+    "typedoc": "^0.22.11",
+    "typedoc-plugin-markdown": "^3.11.12",
+    "typedoc-plugin-missing-exports": "^0.22.6",
     "typescript": "~4.5.4",
     "vinyl-buffer": "^1.0.1",
     "vinyl-source-stream": "^2.0.0",
     "vinyl-transform": "^1.0.0",
     "vite": "~2.7.13",
     "vitepress": "^0.21.4",
-    "vitest": "~0.1.24"
+    "vitest": "~0.1.24",
+    "vuepress-plugin-typedoc": "^0.10.2"
   },
   "packageManager": "pnpm@6.26.1",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "lint": "echo 'TODO eslint'",
     "test": "vitest",
     "coverage": "vitest --coverage",
-    "typedoc": "ts-node typedoc.ts"
+    "typedoc": "npx esno scripts/typedoc.ts"
   },
   "simple-git-hooks": {
     "pre-commit": "npx lint-staged --concurrent false",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "format": "prettier --write .",
     "lint": "echo 'TODO eslint'",
     "test": "vitest",
-    "coverage": "vitest --coverage"
+    "coverage": "vitest --coverage",
+    "typedoc": "typedoc"
   },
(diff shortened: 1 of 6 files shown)


Shinigami92:
```ts
import { Address } from '@faker-js/faker';

const cityName = Address.city();
```

It's sadly not that simple

https://github.com/faker-js/faker/blob/a973ee1ce64d2b593f64b633e927382280da8f27/src/address.ts#L97-L114

Here you can see that city is relying on definitions and the datetype module.
So you would need to always pass that from outside.

https://github.com/faker-js/faker/blob/a973ee1ce64d2b593f64b633e927382280da8f27/src/helpers.ts#L332

Also modules like e.g. `helpers` need e.g. this member function.

Beside that it would break the whole projects compatibility. So we cannot work on that before v7 or later.

---

For now we should try to find another way to generate docs out of the existing code.
Maybe there is just a configuration or we need a JSDoc "annotation".
Maybe it's also possible to export just the types at `index.ts` like `export type { Address } from './address'` :thinking: 

I will test this out a bit later this day.

ST-DDT:
We already export the global Faker instance.
Maybe we can define, that these sub instances also refer to that.
````js
export const faker: Faker = new Faker({
  locales: require('./locales'),
});

export const address = faker.address;
[...]
````

**Comment by ericjeker** [comment]
Added a plugin to generate the documentation in markdown. For now I generate it under `docs/api/typedoc`.

There is a VuePress [plugin](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/vuepress-plugin-typedoc) available to integrate all this with VuePress, but sadly VitePress doesn't support plugins the same way, they integrated plugins in "themes". 

I am really not familiar with those projects so can't really help with that.

So we have 2 solutions I guess:

1. We dump VitePress in favor of VuePress. Which seems reasonable as it works out of the box.
2. We develop a TypeDoc theme or plugin for VitePress, maybe contributing to @tgreyuk project.

**Comment by Shinigami92** [comment]
I (personally) would like to choose the hard / second way here, because we would like to bring and push VitePress and the Vite ecosystem forward.
VuePress 1 seems not in active development anymore and introduce (dev) audit vulnerability dependencies. VuePress 2 is also just in beta for more then half a year.
Knowing that in around one month there will be the new Vue 3 docs and on 7 February 2022 Vue 3 will become the new default, we should really stay with VitePress 😉

All the problems we now find will help potentially thousands other devs to not have these issues in the future.

File: typedoc.ts [hunk 1]
```diff
@@ -0,0 +1,31 @@
+import * as TypeDoc from 'typedoc';
```
Shinigami92, line 1: Please move this file into the scripts folder

package.json
```diff
@@ -45,7 +45,8 @@
     "format": "prettier --write .",
     "lint": "echo 'TODO eslint'",
     "test": "vitest",
-    "coverage": "vitest --coverage"
+    "coverage": "vitest --coverage",
+    "typedoc": "ts-node typedoc.ts"
```
Shinigami92, line 49: Just replace ts-node with esno 
And then remove the ts-node dependency because esno is just like a replacement for ts-node 🙂

Shinigami92 (changes requested):

ericjeker:
I already opened an issue [here](https://github.com/tgreyuk/typedoc-plugin-markdown/issues/287) to see if we can contribute to @tgreyuk repo so more people will profit from this. So far, I was able to integrate some of the generated `.md` with VitePress but some Markdown are not properly generated (e.g. the tables), or properly sanitized and VitePress doesn't like that.

typedoc.ts
ericjeker: Done

package.json
ericjeker: Done

Shinigami92 (changes requested):

package.json
```diff
@@ -78,13 +79,18 @@
...
+    "typedoc-plugin-missing-exports": "^0.22.6",
     "typescript": "~4.5.4",
     "vinyl-buffer": "^1.0.1",
     "vinyl-source-stream": "^2.0.0",
     "vinyl-transform": "^1.0.0",
     "vite": "~2.7.13",
     "vitepress": "^0.21.4",
-    "vitest": "~0.1.24"
+    "vitest": "~0.1.24",
+    "vuepress-plugin-typedoc": "^0.10.2"
```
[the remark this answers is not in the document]
ST-DDT: Sure, just tell me what needs to be automated.

ST-DDT:
I continued this PR in https://github.com/faker-js/faker/pull/289.

ericjeker:
This can be closed and will continue here: https://github.com/faker-js/faker/pull/289

PR #217, opened 2022-01-18
feat: bundle faker with microbundle

prisis:


Shinigami92:
I assume this is blocked by #152 and #169 ?

@faker-js/maintainers Could we have some reviews/approvals for these other PRs? Then we can proceed with this setup :rocket:

prisis:
Yeah, after both are merged this should work :)

JessicaSachs:
I'd like to block on this. I don't know why we would drop support for export default and make this a breaking change.

Shinigami92:
> I'd like to block on this. I don't know why we would drop support for export default and make this a breaking change.

We don't :eyes:

JessicaSachs:
Spoke with Shini and I understand what this is about now 👍🏻

Shinigami92:
@prisis try to at these lines to our workflow ci.yml

https://github.com/vitejs/vite/blob/56d84fb5ea454255ed07b9e01cf9f429eab11b33/.github/workflows/ci.yml#L3-L6

src/index.ts
```diff
@@ -23,6 +23,64 @@ import { Time } from './time';
...
+  id_ID,
+  nb_NO,
+  nl_BE,
+  pt_BR,
+  pt_PT,
+  zh_CN,
+  zh_TW,
+  zu_ZA,
+  af_ZA,
+} from './locales';
```
Shinigami92, line 83:
```suggestion
import * as allLocales from './locales';
```

src/index.ts
```diff
@@ -437,7 +495,64 @@ export class Faker {
...
+    id_ID,
+    nb_NO,
+    nl_BE,
+    pt_BR,
+    pt_PT,
+    zh_CN,
+    zh_TW,
+    zu_ZA,
+    af_ZA,
+  },
```
Shinigami92, line 555:
```suggestion
  locales: allLocales,
```

Shinigami92 (changes requested):

package.json
```diff
@@ -114,11 +110,23 @@
...
-  }
+  },
+  "files": [
+    "CHANGELOG.md",
+    "CHANGELOG_old.md",
+    "dist",
+    "index.d.ts",
+    "index.js",
+    "lib",
+    "locale",
```
Shinigami92, line 128: We don't have these files and folders anymore

```suggestion
```

Shinigami92 (changes requested):

Shinigami92:
We also need to add dist and remove lib to/from some ignore files

package.json
```diff
@@ -22,7 +22,14 @@
...
+  "exports": {
+    "require": "./dist/faker.cjs",
+    "default": "./dist/faker.modern.js"
+  },
+  "main": "./dist/faker.cjs",
+  "module": "./dist/faker.module.js",
+  "unpkg": "./dist/faker.umd.js",
   "files": [
     "CHANGELOG.md",
     "CHANGELOG_old.md",
```
Shinigami92, line 35: Here in files, we can remove 
```
    "index.d.ts",
    "index.js",
    "lib",
    "locale",
```

package.json
```diff
@@ -84,7 +91,8 @@
     "vinyl-transform": "^1.0.0",
     "vite": "~2.7.13",
     "vitepress": "^0.21.4",
-    "vitest": "~0.1.24"
+    "vitest": "~0.1.24",
+    "microbundle": "^0.14.2"
```
Shinigami92, line 95: Please move it up to sorted position
Also it would be good to use `~` instead of `^`

The use `rm -Rf node_modules pnpm-lock.yaml && pnpm install` and commit both files

Shinigami92 (changes requested):

package.json
```diff
@@ -84,7 +91,8 @@
     "vinyl-transform": "^1.0.0",
```
Shinigami92, line 91: We should remove some dependencies:

- browserify
- gulp
- gulp-gh-pages
- gulp-rename
- gulp-uglify
- optimist
- ink-docstrap
- through2
- vinyl-buffer
- vinyl-source-stream
- vinyl-transform

Shinigami92: I think we can also remove the `build` folder completely
(and all references to that, like in readme/docs and ignore files)

Shinigami92 (changes requested):

Shinigami92:
This PR may get stale by #257

Issue #113, opened 2022-01-12
`random.word()` can return undesirable non-alpha characters


griest024:
### Describe the bug

The pool of strings from which `random.words()` pulls values includes strings that have parentheses, e.g. `Cocos (Keeling) Islands` and `Killer Whale (Orca)`. This can result in `(Keeling)` and `(Orca)` being returned.

There's a larger problem here of non-alphabetic characters being returned in words. Obviously some non-alpha characters are desired in words, e.g. `-`, but some are not. We should decide on the set of acceptable characters that `word()` can return and regex the others out.

### Reproduction

Mock `faker.fake` with the above mentioned values and you can see the bad returns.

### Additional Info

pkuczynski:
I can work on this. Let me know if you want me to...

griest024:
> I can work on this. Let me know if you want me to...

Sure I can assign you. What do you think should be the allowed chars? How would you handle other locales?

pkuczynski:
I took a bit different approach. Have a look at #267

Shinigami92:
I will try to create a PR soon that just uses a `/\w+/` regex

PR #218, opened 2022-01-18
fix(#113): Adds regex to `random.word` to trim unwanted characters

prisis:
The rebase on #115 didnt work...

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -199,11 +199,15 @@ export class Random {
       'name.jobArea',
       'name.jobType',
     ];
+    // regex statement used to remove unwanted characters from beginning/end of words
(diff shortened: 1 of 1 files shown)


MilosPaunovic (dismissed):

damienwebdev:
I still don't like this implementation. It feels kludgy. I would prefer a separate dataset rather than a regex.

HonzaMac:
Would this change hit performance of generating random data? I am little bit worried of calling regex like this.

Shinigami92:
I think we have now 2 different (with me 3) that don't really like this PR :confused: 
Maybe we should just use that regex outside the random function call on demand

PR #231, opened 2022-01-19
docs: shorten TLDR; move to the permalink instead

fzn0x:
Reduce extra work when modifying this regex as soon as possible

### DIFF
--- a/.github/commit-convention.md
+++ b/.github/commit-convention.md
@@ -4,12 +4,7 @@
 
 #### TL;DR:
 
-Messages must be matched by the following regex:
-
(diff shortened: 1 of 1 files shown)


bl0cknumber:
Look like the test failing by timeout, do we need to increase it?

Shinigami92:
> Look like the test failing by timeout, do we need to increase the timeout?

No, we assume it's a windows <-> UTC time bug

Shinigami92 (approved):

ejcheng (approved):

damienwebdev (approved):
lgtm

PR #242, opened 2022-01-20
add chinese hyphenated name

LeToNode:
add chinese hyphenated name


Shinigami92:
Please rebase and do it in the ts file now

Issue #210, opened 2022-01-18
docs: document supported locales


bl0cknumber:
### Clear and concise description of the problem

Document supported locales in https://github.com/faker-js/faker/tree/main/locale

Avoid related issues and discussions (e.g https://github.com/faker-js/faker/discussions/209)

### Suggested solution

Write supported locales in https://github.com/faker-js/faker/tree/main/locale

### Alternative

-

### Additional context

- Using table
- Add navigation to locale README in main README.

Shinigami92:
We should do three things:

1. Add the supported locales in the VitePress docs
2. Add them in README.md (so that folks on the npm site can read it without the need of jump into docs but just searching with ctrl+f)
3. Add an enum/string literal that defines the locales and export it + type the functions in TS  
    But maybe it's possible to provide you own locale on demand from external, so it might need to be an `LiteralUnion` (see e.g. https://github.com/microsoft/TypeScript/issues/29729#issuecomment-471566609)

Shinigami92:
Please also note that there is currently a huge PR that needs to be merged first, to reduce rebase/merge mess #169

fncolon:
> We should do three things:
> 
> 1. Add the supported locales in the VitePress docs
> 2. Add them in README.md (so that folks on the npm site can read it without the need of jump into docs but just searching with ctrl+f)
> 3. Add an enum/string literal that defines the locales and export it + type the functions in TS
>    But maybe it's possible to provide you own locale on demand from external, so it might need to be an `LiteralUnion` (see e.g. [Literal String Union Autocomplete microsoft/TypeScript#29729 (comment)](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-471566609))

Cool! especially number 3

fncolon:
> Please also note that there is currently a huge PR that needs to be merged first, to reduce rebase/merge mess #169

Noted, thanks @Shinigami92

Shinigami92:
Update to my comment: https://github.com/faker-js/faker/issues/210#issuecomment-1015139942

We now already have: https://github.com/faker-js/faker#localization and https://fakerjs.dev/api/localization#localization
We don't have an automatisation for updating these.
Also I think we should link from the README.md to the official docs, so we don't need to keep the README.md up to date.

Also we want inline dev docs for the Faker constructor `options.locale` + LiteralUnion typed

ST-DDT:
I would like to do this.

ST-DDT:
[...]

PR #252, opened 2022-01-21
feat: auto generate/update locale files

ST-DDT:
This PR adds a script that automatically generates
- `src/locale/*.ts`
- `src/locales/index.ts`.
- and `docs/api/localization.md`

This should be merged after https://github.com/faker-js/faker/pull/248

The is the generation/update related part for #210

Currently this is not integrated in the build process.


scripts/generateLocales.ts
```diff
@@ -0,0 +1,53 @@
+import fs from 'fs';
+import { format } from 'prettier';
+import options from '../.prettierrc.cjs';
+
+const pathLocale = 'src/locale/';
+const pathLocales = 'src/locales/';
```
Shinigami92, line 6: You should use `resolve` for these

```ts
import { resolve } from 'node:path'

const pathLocale = resolve(__dirname, 'src', 'locale');
```

scripts/generateLocales.ts
```diff
@@ -0,0 +1,53 @@
+import fs from 'fs';
```
Shinigami92, line 1: We could use `import { readdirSync, writeFileSync } from 'node:fs';`

Shinigami92 (changes requested):

ST-DDT:
This has been split into multiple PRs and needs to be updated once #263 and #264 have been merged.

ST-DDT:
If the following change is considered breaking, then we could just temporarily remove/change it.
That way, we can still use the automation script.

````diff
- export = faker;
+ export default faker;
````

**Comment by Shinigami92** [comment]
I thing `export = faker` is currently the right way, so plain js node users can just use `const fakerDe = require('@faker-js/faker/locale/de')` instead of `const fakerDe = require('@faker-js/faker/locale/de').default`.
So indeed, for the `locale`s we have a special case here.

File: package.json [hunk 3]
```diff
@@ -55,6 +55,7 @@
     "build:code": "esno ./scripts/bundle.ts",
     "build:types": "tsc --emitDeclarationOnly --outDir dist/types",
     "build": "run-s build:clean build:code build:types",
+    "generate": "esno ./scripts/generateLocales.ts",
```
Shinigami92, line 58: Maybe we should be a bit more explizit here and name the script like `generate:locale`

src/locale/en.ts
```diff
@@ -9,4 +9,4 @@ const faker = new Faker({
   },
 });
 
-export = faker;
```
Shinigami92, line 12: I think you missed that one, is it not generated?

Shinigami92 (changes requested):

Shinigami92 (dismissed):
Nice work :rocket:

Shinigami92 (dismissed):

scripts/generateLocales.ts
```diff
@@ -0,0 +1,100 @@
...
+let localeIndexLocales = 'const locales: KnownLocales = {\n';
+
+let localizationLocales = '| Locale | Name |\n| :--- | :--- |\n';
+
+const autoGeneratedCommentHeader = `/*
+ * This file is automatically generated.
+ * Run '${scriptCommand}' to update.
+ */`;
+
+for (let locale of locales) {
```
Shinigami92, line 34:
```suggestion
for (const locale of locales) {
```

Shinigami92 (approved):

Shinigami92 (dismissed):

Shinigami92 (approved):

ST-DDT:
I need this in main in order to update it for #363 .
(There is no automation enabled to this yet.)

CONTRIBUTING.md
```diff
@@ -13,6 +13,7 @@ If you want to make `Faker` a better, please read the following contribution gui
   Make sure you **build** the project before running the docs, cause some files depend on `dist`.
   Use `pnpm run docs:dev` to edit them in live mode.
 - The tests are executing `vitest` against `test/**/*.spec.ts`
+- If you update the locales, make sure to run `pnpm run generate:locales` to generate/update the related files.
```
prisis, line 16: Can we create a ticket to add a workflow test for this? that is only running if something in the locales folder did change?

prisis (approved):

PR #254, opened 2022-01-21
chore: migrate vendor

Shinigami92:


import-brain (dismissed):

prisis (approved):

import-brain (dismissed):

src/vendor/mersenne.ts
```diff
@@ -7,89 +7,91 @@
...
+ */
 
-   Any feedback is very welcome.
-   http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt.html
-   email: m-mat @ math.sci.hiroshima-u.ac.jp (remove space)
-*/
+let dbg: number;
 
-function MersenneTwister19937() {
+function MersenneTwister19937(): void {
```
griest024, line 56: This seems like this should be defined as a class.

src/vendor/unique.ts
```diff
@@ -39,12 +40,21 @@ unique.errorMessage = function (now, code, opts) {
...
-  var now = new Date().getTime();
+}
+
+export function exec<Method extends Function, Args extends any[], Result>(
+  method: Method,
+  args: Args,
+  opts: {
+    maxTime?: number;
+    maxRetries?: number;
+    exclude?: any[];
```
griest024, line 51: Can we do better with `exclude`'s type? Something like `string | (string | Result)[]`? This might introduce some annoying casts later but it would at least be more accurate. On line 76 `opts.exclude` is specifically checked for a `string` value so that should at least be part of the type.

src/vendor/user-agent.ts
```diff
@@ -175,26 +177,26 @@ exports.generate = function generate(faker) {
...
+      safari: { win: 0.04, mac: 0.96 },
+      iexplorer: ['win'],
+    };
 
     return [browser, rnd(os[browser])];
   }
 
-  function randomProc(arch) {
-    var procs = {
+  function randomProc(arch: 'lin' | 'mac' | 'win') {
```
griest024, line 198: Maybe this should be an enum? Not a big deal either way.

prisis (approved):

griest024 (changes requested):

src/vendor/mersenne.ts
Shinigami92: I wont like to do that now, we could do in a separate PR later

griest024: Can you add a TODO or issue for it?

src/vendor/unique.ts
```diff
@@ -73,19 +83,25 @@ unique.exec = function (method, args, opts) {
...
       'Exceeded maxRetries:' + opts.maxRetries,
+      // @ts-expect-error: we know that opts.startTime is defined
       opts
     );
   }
 
-  // execute the provided method to find a potential satifised value
-  var result = method.apply(this, args);
+  // execute the provided method to find a potential satisfied value
+  const result = method.apply(this, args);
```
griest024, line 104: Can we type this var as `Result`? Maybe `method` can have an even better type so that the return is a `Result`? Like `Method extends (args: Args) => Result`.

griest024 (changes requested):

src/vendor/mersenne.ts
Shinigami92: https://github.com/faker-js/faker/issues/280

src/vendor/unique.ts
griest024: Although it looks like `Function#apply` explicitly has `any` as its return, thats disappointing. Explicitly typing this var as `Result` introduces some type errors elsewhere.

Can result ever not be a string? Does it need to be generic?

Shinigami92: We should try later to get rid of `.apply` and call it natively

griest024 (approved):

prisis (approved):

PR #257, opened 2022-01-22
build: bundle with esbuild

Shinigami92:


Shinigami92:
@knightly-bot build this

damienwebdev:
As I an unfamiliar with the various types of packaging mechanics in NodeJS, I found it helpful to ref this document when reviewing this

https://nodejs.org/api/packages.html#packages_package_entry_points

package.json
```diff
@@ -22,21 +21,41 @@
     "url": "https://github.com/faker-js/faker.git"
   },
   "license": "MIT",
-  "main": "lib/index.js",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "index.d.ts",
+  "typesVersions": {
+    ">=4.2": {
```
damienwebdev, line 28: How did you decide on this version?

damienwebdev:
This is missing a README update.

damienwebdev (changes requested):

damienwebdev:
This also breaks browser builds (or at least documentation for them). What should the migration path be? Is it possible to output to dist backwards compatibly?

damienwebdev:
This also breaks the docs.

damienwebdev:
@Shinigami92 I want to ensure we're clear on whether or not this drops support for es5 in browser builds.

Shinigami92:
@damienwebdev 

> As I am unfamiliar with the various types of packaging mechanics in NodeJS, I found it helpful to ref [this packaging document](https://nodejs.org/api/packages.html#packages_package_entry_points) when reviewing this

Oh yeah, I also did know nothing about this before this PR.

> This is missing a README update.

Oh yeah, you are right. I will work on that today or tomorrow :+1: 

> This also breaks browser builds (or at least any existing documentation for them). What should the migration path be? Is it possible to output to dist backward compatibly?

Potentially yes, it could break the browser support. That's why I would like to have a `6.0.0-alpha.4` release to test out if we can do something with https://unpkg.com
If this doesn't work out of the box, we may need to add a third bundle for `iife` (See https://esbuild.github.io/api/#format-iife)
But this would mean additional around 10 MiB, because the locales needs always to be duplicated for each bundle.

I'm thinking about to drop browser support for now and bring it back in later when we found a much better way of async loading in locales. (This may not be that much difficult.)

> This also breaks the docs.

Oh :astonished:, thanks I need to have a look into that :+1: 

> I want to ensure we're clear on whether or not this drops support for es5 in browser builds, do you know if it does?

I need also to find out. But if we really provide an iife bundle, we can just target `es5`. So really easy with that :+1:.

package.json
Shinigami92: It's based on what's doing `rxjs` :eyes: 



https://github.com/ReactiveX/rxjs/blob/123a0f256e0022e87dfa249c895670ee745def53/package.json#L10

Shinigami92:
@damienwebdev Currently we provide `sourcemaps` via `esbuild`
Do we need to provide these? Because these are really big and blow up the bundle size.

damienwebdev:
@Shinigami92 can you doc here the before and after size of the the bundles so that we can compare?

package.json
damienwebdev: I think we target the major version instead? 4.0 for safety?

Shinigami92:
@damienwebdev 

I checked three things:

1. esbuild with sourcemaps  
   package size: 3.4 MB | unpacked size: 16.6 MB | total files: 5835  
   [esbuild-with-sourcemap.log](https://github.com/faker-js/faker/files/7920156/esbuild-with-sourcemap.log)

2. esbuild without sourcemaps  
   package size: 1.6 MB | unpacked size: 7.0 MB | total files: 3871  
   [esbuild-without-sourcemap.log](https://github.com/faker-js/faker/files/7920157/esbuild-without-sourcemap.log)

3. main with tsc + gulp browser no sourcemap (current state)  
   package size: 3.1 MB | unpacked size: 14.5 MB | total files: 3803  
   [tsc-and-gulp-browser-without-sourcemap.log](https://github.com/faker-js/faker/files/7920158/tsc-and-gulp-browser-without-sourcemap.log)

---

Please note that the esbuild process includes cjs and esm but no browser and the main includes cjs and browser but no esm
Also note that the current main doesn't provide source maps. So we could really leave them out for now and then we would reduce the bundle size by half!

And in addition to that, I'm already thinking about to split manually the locales into it's own folder and fetch them on demand.
But I assume we need to convert them then into json files or yaml if possible to make them even smaller.

But this would be definitely something for another PR

CONTRIBUTING.md
```diff
@@ -1,23 +1,19 @@
...
+## Good to know
 
-`Faker` relies on [commonJS](http://www.commonjs.org/) standard and supports both node.js and the browsers. Keep this in mind, when modifying and/or extending the sources.
-
-## Automation
-
-- The project is being built by [gulp](http://gulpjs.com/) (see [gulpfile](build/gulpfile.js)), destination directory is [build/build](build/build)
-- The documentation is auto-generated, based on [build/src](build/src) markdown sources. If you modify the main [Readme.md](Readme.md) file, the Pull Request will be rejected, since it will be overwritten by the upcoming `gulp` execution
+- The project is being built by [esbuild](https://esbuild.github.io) (see [bundle.ts](scripts/bundle.ts))
+- The documentation is running via VitePress. Use `pnpm run docs:dev` to edit them in live mode.
```
prisis, line 12: Should we add that the pnpm run build should be run before the docs?

package.json
```diff
@@ -22,23 +21,43 @@
...
   },
   "license": "MIT",
-  "main": "lib/index.js",
+  "main": "lib/cjs/index.js",
+  "module": "lib/esm/index.js",
+  "types": "index.d.ts",
+  "typesVersions": {
+    ">=4.0": {
+      "*": [
+        "lib/types/*"
```
prisis, line 30: why use lib and not dist?

package.json
```diff
@@ -22,23 +21,43 @@
...
+      ]
+    }
+  },
+  "exports": {
+    ".": {
+      "node": "./lib/cjs/index.js",
```
prisis, line 36: can we not just have a different ending like index.cjs and index.js

prisis: or index.cjs.js and index.esm.js

test/address.spec.ts
```diff
@@ -1,6 +1,6 @@
 import type { JestMockCompat } from 'vitest';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { faker } from '../lib';
+import { faker } from '../lib/cjs';
```
prisis, line 3: so all of this changes are not needed if the ending is different

package.json
Shinigami92: Not possible right now, cause we have multiple entrypoints that are referencing each other

damienwebdev (approved):
lgtm

prisis (approved):

PR #258, opened 2022-01-22
fix: improve internet tests and fix bug in domain name generation

### LINKED ISSUE #256: faker.internet.url() returning invalid URLs with spaces
### Describe the bug

faker.internet.url() sometimes returns invalid URLs with spaces, e.g.

http://sick-dune buggy.name
https://hollow-dust storm.info

This is causing some of my tests which incorporate URL validation to intermittently fail. I had never seen this with the old "faker" but it started happening since moving to "@faker-js/faker" (6.0.0-alpha.3).

### Reproduction

``` TypeScript
  const url = faker.internet.url();
  if (url.includes(" ")) console.log(url);
  // Yields unexpected output of invalid URLs
}
```

### Additional Info

https://www.npmjs.com/package/@faker-js/faker/v/6.0.0-alpha.3

ST-DDT:
Fixes #256


ST-DDT:
@cduff Does your tooling consider `https://full-jalapeã±o.info` to be a valid url?
@maintainers: Should we return such values?

Shinigami92:
If `jalapeño` is not a valid value for a URL, then we would need to normalize the URL
Maybe this is a good idea in general :thinking:

test/internet.spec.ts
```diff
@@ -186,14 +197,35 @@ describe('internet.js', () => {
...
+      spy_internet_domainWord.mockRestore();
+      spy_internet_domainSuffix.mockRestore();
+    });
+
+    it('url to be valid', () => {
+      for (let i = 0; i < 10000; i++) {
+        const url = faker.internet.url();
+
+        expect(url).match(
+          /^https?:\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?$/,
```
damienwebdev, line 225: This appears to be regex magic. Perhaps we pull in a validation library? I don't trust myself well enough to know all the edge-case requirements of domain naming. https://github.com/validatorjs/validator.js/blob/master/src/lib/isFQDN.js#L12

damienwebdev (changes requested):

test/internet.spec.ts
ST-DDT: I will work on that tomorrow evening.

cduff:
> @cduff Does your tooling consider `https://full-jalapeã±o.info` to be a valid url? What about `https://full-jalapeño.info`?
> 
> Or more specifically should we return such values? (Other languages might have more of them)
> 
> https://github.com/faker-js/faker/blob/60c90028ba76b7e291fdb8152425b93c41b117c9/src/locales/en/word/noun.ts#L3252

I'm using https://www.npmjs.com/package/class-validator `isUrl` for the validation. It reports the two strings you specified above as valid.

test/internet.spec.ts
ST-DDT: Done.

ST-DDT:
I also added the verification checks to the other internet module tests.

test/internet.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest';
 import { faker } from '../lib';
+import validator from 'validator';
 
 describe('internet.js', () => {
```
Shinigami92, line 5: Oh I missed that one

```suggestion
describe('internet', () => {
```

test/internet.spec.ts
```diff
@@ -100,6 +122,7 @@ describe('internet.js', () => {
       const domain_name = faker.internet.domainName();
 
       expect(domain_name).toBe('bar.net');
+      expect(validator.isURL('https://' + domain_name)).toBeTruthy();
```
Shinigami92, line 125: Let's try out `isFQDN`

test/internet.spec.ts
```diff
@@ -233,24 +251,54 @@ describe('internet.js', () => {
...
   describe('ipv6()', () => {
     it('returns a random IPv6 address with eight parts', () => {
       const ip = faker.internet.ipv6();
       const parts = ip.split(':');
+
       expect(parts).toHaveLength(8);
+      expect(validator.isIP(ip)).toBeTruthy();
+    });
+
+    it('ip6 to be valid', () => {
```
Shinigami92, line 277: Let's call this IPv6 and the other IPv4

test/internet.spec.ts
```diff
@@ -272,7 +320,17 @@ describe('internet.js', () => {
   describe('color()', () => {
     it('returns a valid hex value (like #ffffff)', () => {
       const color = faker.internet.color(100, 100, 100);
+
       expect(color).match(/^#[a-f0-9]{6}$/);
+      expect(validator.isHexColor(color)).toBeTruthy();
```
Shinigami92, line 325: I think you missed to remove the regex variant

Shinigami92 (approved):

Shinigami92:
We should find a better title for the PR now :wink:

Shinigami92 (approved):

Shinigami92 (approved):

ST-DDT:
@damienwebdev Would you mind reviewing this again, so that we can merge this?
Some other PRs would benefit from the validator library added here as well.

prisis (dismissed):

Shinigami92:
@ST-DDT you can decide if you want to merge it or wait on @damienwebdev :slightly_smiling_face:

ST-DDT:
If we want to use it for #303 or a release, then we should merge it now, otherwise we can wait a day for damien.

Shinigami92 (approved):

test/internet.spec.ts
```diff
@@ -186,39 +230,88 @@ describe('internet.js', () => {
...
   });
 
   describe('port()', () => {
     it('returns a random port number', () => {
       const port = faker.internet.port();
+
       expect(Number.isInteger(port)).toBe(true);
       expect(port).greaterThanOrEqual(0);
       expect(port).lessThanOrEqual(65535);
+      expect(port + '').satisfy(validator.isPort);
```
prisis, line 306:
```suggestion
      expect(String(port)).satisfy(validator.isPort);
```

test/internet.spec.ts
```diff
@@ -186,39 +230,88 @@ describe('internet.js', () => {
...
+    });
+
+    it('port() to return valid values', () => {
+      for (let i = 0; i < 10000; i++) {
+        const port = faker.internet.port();
+
+        expect(port + '').satisfy(validator.isPort);
```
prisis, line 313:
```suggestion
        expect(String(port)).satisfy(validator.isPort);
```

Shinigami92 (approved):

prisis (approved):

PR #260, opened 2022-01-22
fix: date.between typing

MoSattler:
Return the typing to the [original typing](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/291c346588768b8ea13472ff95bdacd659f67790/types/faker/v3/index.d.ts#L58) and in accordance with the [current spec](https://github.com/faker-js/faker/blob/main/test/date.spec.ts#L119-L131).

### DIFF
--- a/src/date.ts
+++ b/src/date.ts
@@ -68,9 +68,10 @@ export class _Date {
    * @param from
    * @param to
    */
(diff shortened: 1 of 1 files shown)


Shinigami92:
Already done by #200

PR #267, opened 2022-01-23
chore: add .nvmrc for those who use automatic node switching

pkuczynski:
See https://github.com/wbyoung/avn


prisis (dismissed):

Shinigami92:
Is it possible to give a version range? Or like `>=16.9`

`16.9` first introduces `corepack` and this could be really helpful for using the compatibility npm to pnpm

pkuczynski:
> Is it possible to give a version range? Or like `>=16.9`
> 
> `16.9` first introduces `corepack` and this could be really helpful for using the compatibility npm to pnpm if you have activated it on your machine

No, its not possible to specify a version like this. But we can do `16.9` which I just did. This will force to use 16.9 for anybody using `avn` (or similar automatic node switching).

Shinigami92:
Okay, sorry than for that
I would vote then to force 16.13 (the current version) for now

pkuczynski:
We can use `lts/gallium` to express latest node 16 or `lts/*` for latest LTS. See https://github.com/nvm-sh/nvm#long-term-support. Let me know how should I adjust it?

prisis:
> Okay, sorry than for that I would vote then to force 16.13 (the current version) for now

Than we need to change it in the ci too

pkuczynski:
> Than we need to change or add 16.13 to the ci too

Should I do this as part of this PR?

Shinigami92:
> Than we need to change or add 16.13 to the ci too

I think CI don't use `nvm`, does it?

The `lts/gallium` sounds also good to me :thinking:

Shinigami92 (approved):

pkuczynski:
Than its ready to merge :)

ST-DDT (approved):

PR #275, opened 2022-01-23
chore: drop fake usage in vehicle.ts

pkuczynski:


doughlass:
Would we not want to map the vehicle makes to manufacturers? Otherwise a fake vehicle response may respond with `Tesla PT Cruiser`

pkuczynski:
> Would we not want to map the vehicle makes to manufacturers? Otherwise a fake vehicle response may respond with `Tesla PT Cruiser`

I was thinking about it, but we don't have such data. Manufacturers and models are collected independently. We could implement this in the future, but not with the current data set.

Shinigami92:
@pkuczynski This PR needs a rebase

ST-DDT (approved):

Shinigami92 (approved):

PR #276, opened 2022-01-23
fix: change random.word to be based on definitions.word.*

### LINKED ISSUE #113: `random.word()` can return undesirable non-alpha characters

pkuczynski:
Fixes #113
Fixes #218


prisis:
I would put into 6.1  @import-brain

import-brain:
> I would put into 6.1  @import-brain

Done

pkuczynski:
How about the review @prisis ?

prisis:
@pkuczynski hope it's fine with you if i wait with the review a bit.  
  
I would like to fix the tests more and remove all the mock function call, so we now that we break something.  
  
And this is a runtime change of the code, so we will not merge it into v6.0 for now. But im thankful for you spend time on this PR.

src/random.ts
```diff
@@ -159,53 +159,28 @@ export class Random {
     return this.faker.datatype.boolean();
   }
 
-  // TODO: have ability to return specific type of word? As in: noun, adjective, verb, etc
```
prisis, line 162: The question is if we want to have this feature too.

prisis (approved):

src/random.ts
pkuczynski: We already have. Users can use `faker.word.xyz` to do it... That's why I removed this comment.

pkuczynski:
The problem is, that the longer you wait with merging, the higher chance of conflict. I worked on this as it was listed as "help needed". This feature does not have any tests, so not sure what mocks do you want to remove?

Shinigami92:
> The problem is, that the longer you wait with merging, the higher chance of conflict. I worked on this as it was listed as "help needed". This feature does not have any tests, so not sure what mocks do you want to remove?

Sorry but this is irrelevant to the current situation of the project.

We have a roadmap https://github.com/orgs/faker-js/projects/2 with the milestone https://github.com/faker-js/faker/milestone/1
We are thanksful that you try to contribute, but if you want to prevent merge conflicts right now, than wait until we release 6.0.0 and stop opening PR for now

This said, I can somewhat promise you that `src/random.ts` will not change that much anymore. The project stability is more regarding to the test-suite, bundling-setup and documentation. Multiple times said: no runtime changes before 6.0.0 released

src/random.ts
```diff
@@ -159,53 +159,28 @@ export class Random {
...
-    const result = this.faker.fake('{{' + randomWordMethod + '}}');
-    return this.faker.random.arrayElement(result.split(' '));
+    const definition = this.faker.random.arrayElement(definitions);
+    let word = '';
+
+    do {
+      word = this.faker.random.arrayElement(definition);
+    } while (word.includes(' '));
+
+    return word;
```
griest024, line 183: I much prefer the `split` implementation over the `do...while`.

griest024 (changes requested):

src/random.ts
pkuczynski: The `do...while` is only used here to get next random word if the one currently selected contains a space. Not sure how split would help here? Can you show an example?

griest024: For example if `contact lens` was selected, `do..while` would skip that whereas `split` would return either `contact` or `lens`. This is assuming that we need to eliminate spaces (which maybe we should not?).

griest024: I vote remove this logic entirely and remove word examples with spaces instead. A regex search for `'.* .*'` in `src/locales/*/word` should get you all of them.

pkuczynski: Sure, let's do that! I will remove this logic here...

griest024 (approved):

ST-DDT (approved):

pkuczynski:
Possibly in the future we might consider to move this to faker.word.any() as this what happened with other methods from this module...

griest024:
> Possibly in the future we might consider to move this to faker.word.any() as this what happened with other methods from this module...
> 
> What do you think? I could record this idea as a GitHub issue for future consideration...

I think that is reasonable.

pkuczynski:
> I think that is reasonable.

Here you go https://github.com/faker-js/faker/issues/339 :)

Shinigami92:
@pkuczynski Could you rebase this PR so we can see how it behaves with all our new tests?

Shinigami92:
As this is changing runtime behavior and we already want to do #339 anyways, I will close this PR in favor of #339

pkuczynski:
@Shinigami92 I think this should be still merged, because it contains valid changes. Then the move in #339 can happen more easily...

Shinigami92:
@ST-DDT and I decided this together via screen-share some minutes ago
We do not see to merge this because it just changes behavior that we already want to deprecate anyways.
What you could do right now would be to just add a `console.warn` that logs that this function will be deprecated so we can already deprecate it in `v6.1`.

Let me also say that I do totally **not** expect that the move from initial to v6.0.0 will take same amount of time as from v6.0 to v6.1 🙂 
We want now to move on a bit faster

pkuczynski:
I mean this function after the changes done currently, can be with pretty much (almost) no change moved to `word.any` and `random.word` deprecated. It does slightly change the behaviour, that's true but not in a way which is dangerous in any way and it already does fix the issues reported in #113... So I still see this as perfectly mergeable...

pkuczynski:
@Shinigami92 @ST-DDT rebased and tests fixed

Shinigami92 (changes requested):
I'm more and more tempted to not accept this PR 😕 Sorry...

I re-read through #113 and I thing we should go with the suggested proposal to just regex them out if they generate _"invalid"_ words and then make a retry

With this PR (#276) we loose many possible words, especially hacker words, were you can see that they are not in generated system filenames anymore

I will block this from my side with a `request changes`, but if another maintainer is on my side, then I would vote for close this PR

pkuczynski:
Well, at least we get more sensible words with this pr comparing to before... but yeah, do what you wish...

ST-DDT:
> Well, at least we get more sensible words with this pr comparing to before... but yeah, do what you wish...

What do you consider a "sensible" word? IMO we have to decide first what we expect from this method.
AFAIK this method is used in file name generation. Should those contain Chinese letters in Chinese locales?

pkuczynski:
> > Well, at least we get more sensible words with this pr comparing to before... but yeah, do what you wish...
> 
> What do you consider a "sensible" word? IMO we have to decide first what we expect from this method. AFAIK this method is used in file name generation. Should those contain Chinese letters/words in Chinese locales? What about the opposite? Should it partially return English file names in Chinese locales

I wold consider noun, adjective, verb, ... a sensible word. While things like `streetSuffix` or `currencyName` does not seem something I would expect from `random.word` (or `word.any` in the future). For example would `as long as` returned at the moment could be considered as a word? Even if you split it into `as` and `long`. The second one seems ok, but `as`? Don't think so...

Looking at the wikipedia https://en.wikipedia.org/wiki/Word:

> In [linguistics](https://en.wikipedia.org/wiki/Linguistics), a word of a [spoken language](https://en.wikipedia.org/wiki/Spoken_language) can be defined as the smallest sequence of [phonemes](https://en.wikipedia.org/wiki/Phoneme) that can be uttered in isolation with [objective](https://en.wikipedia.org/wiki/Semantics) or [practical](https://en.wikipedia.org/wiki/Pragmatics) [meaning](https://en.wikipedia.org/wiki/Meaning_(linguistics)). In many languages, words also correspond to sequences of [graphemes](https://en.wikipedia.org/wiki/Grapheme) ("letters") in their standard [writing systems](https://en.wikipedia.org/wiki/Writing_system) that are delimited by spaces wider than the normal inter-letter space, or by other graphical conventions.[[1]](https://en.wikipedia.org/wiki/Word#cite_note-hasp2011-1) The concept of "word" is usually distinguished from that of a [morpheme](https://en.wikipedia.org/wiki/Morpheme), which is the smallest unit of word which has a meaning, even if it will not stand on its own together or in other small words.

pkuczynski:
For filename generation we could also use `lorem.slug()`.

In other languages, I would expect the filename is generated in the selected locale, unless there is no translation for it. Wouldn't you?

pkuczynski:
I am not doing anything with his PR until you make a decision, right?

Shinigami92:
In yesterdays maintainers-team-meeting we discussed that we want to have the `random.word` function just as it is right now, but add a `@deprecated` JSDoc comment to it and a warning for removal
We want to have the `word.any` function directly and it should be implemented like you are already doing in this PR

pkuczynski:
OK!

PR #285, opened 2022-01-24
test: remove mocks from address

Shinigami92:


test/address.spec.ts
```diff
@@ -1,782 +1,563 @@
...
-
-      expect(spy_name_firstName).toHaveBeenCalledOnce();
-      expect(spy_name_lastName).not.toHaveBeenCalled();
-      expect(spy_address_streetSuffix).toHaveBeenCalledOnce();
-
-      spy_datatype_number.mockRestore();
-    });
+const seededRuns = [
+  {
+    seed: 42,
```
pkuczynski, line 6: How did you get to these seed values? When `en` files will change would seed values remain the same?

prisis: Just random number that was choosen. The idea is too have something like a snapshot of the output for the current version.

Shinigami92: 42: https://simple.wikipedia.org/wiki/42_(answer)
1337: https://simple.wikipedia.org/wiki/Leet
1211: another number related to us :shrug:

Shinigami92:
I will add some missing cases

test/address.spec.ts
pkuczynski: That would tight the test with the dataset, right? Wouldn't it be better to mock `definitions` to avoid this?

griest024: So just to clarify, would these expectations need to change if the definition file contents were changed?

test/address.spec.ts
```diff
@@ -1,782 +1,625 @@
...
 
-  describe('ordinalDirection()', () => {
-    it('returns random ordinal direction', () => {
-      const spy_address_ordinalDirection = vi
-        .spyOn(faker.address, 'ordinalDirection')
-        .mockReturnValue('West');
+            expect(typeof latitude).toBe('string');
+            expect(
+              latitude.split('.')[1].length,
+              'The precision of latitude should be had of 4 digits'
```
griest024, line 414: Grammar: The precision of latitude should be ~~had of~~ 4 digits | The precision of latitude should *have* ~~be had of~~ 4 digits

same in other places

test/address.spec.ts
```diff
@@ -1,782 +1,625 @@
...
 
-      expect(zipCode).match(/^[A-Za-z]\d[A-Za-z]\s\d[A-Za-z]\d$/);
+          const address = faker.address.streetAddress();
+          const parts = address.split(' ');
 
-      // try another format
-      zipCode = faker.address.zipCode('###-###');
+          expect(
+            parts[0].length,
+            `The street number should had ${expectations.streetAddressDigits} digits`
```
griest024, line 147: grammar: The street number should ~~had~~ *have* ${expectations.streetAddressDigits} digits

same in other places

test/address.spec.ts
```diff
@@ -1,782 +1,625 @@
...
-          latitude.split('.')[1].length,
-          'The precision of latitude should be had of 4 digits'
-        ).toBe(4);
+          const direction = faker.address.direction();
+          const expected = expectations.direction;
 
-        const latitude_float = parseFloat(latitude);
+          expect(
+            direction,
+            'The random direction should be equals ' + expected
```
griest024, line 247: grammar: The random direction should be ~~equals~~ *equal to* | The random direction should ~~be equals~~ *equal*

same in other places

test/address.spec.ts
```diff
@@ -1,782 +1,625 @@
...
+
+          // test once with undefined radius
+          const coordinate = faker.address.nearbyGPSCoordinate(
+            [latFloat1, lonFloat1],
+            undefined,
+            isMetric
+          );
+          expect(coordinate.length).toBe(2);
+          expect(typeof coordinate[0]).toBe('string');
+          expect(typeof coordinate[1]).toBe('string');
```
griest024, line 620: I think this should be a separate spec: `When the specified radius is undefined`

griest024 (changes requested):

Shinigami92:
The `non metric` coords test is currently flaky

We also found that we maybe should set a random seed for the non-seed-based tests, cause otherwise it would take the last set seed sequence

test/address.spec.ts
```diff
@@ -244,7 +244,7 @@ describe('address', () => {
 
           expect(
             direction,
-            'The random direction should be equals ' + expected
+            `The random direction should be equal ${expected}`
```
griest024, line 247: `be equal to` or `equal`

test/address.spec.ts
```diff
@@ -535,12 +529,9 @@ describe('address', () => {
 
           expect(
             typeof cardinalDirection,
-            `${prefixErrorMessage} be had typeof equals ${expectedType}. Current is ${typeof cardinalDirection}`
+            `${prefixErrorMessage} be typeof equal to ${expectedType}. Current is ${typeof cardinalDirection}`
```
griest024, line 532: `${prefixErrorMessage} be of type ${expectedType}` like you have on line 500

griest024 (changes requested):

test/address.spec.ts
```diff
@@ -1,782 +1,661 @@
...
+
+          const distanceToTarget =
+            Math.pow(coordinate[0] - latitude, 2) +
+            Math.pow(coordinate[1] - longitude, 2);
+
+          console.log([latitude, longitude], coordinate, distanceToTarget);
+
+          expect(distanceToTarget).lessThanOrEqual(
+            100 * 0.002 * 1.6093444978925633 // 100 miles to km ~= 0.9 degrees, we take 2 degrees
+          );
```
Shinigami92, line 656: This is currently failing hard...
I don't know if the function is wrong or if my test is wrong.

2 degrees are like 200+ km and the factor 1.6 converts the 100km to miles
also `isMetric = false` should then use the miles version of the function
but the delta is very huge :thinking:

Shinigami92: @prisis what do you think I should do here?
Currently it blocks the PR to get merged, so maybe we should mark it with `it.todo` and look into that later? Maybe even create an issue to track it?
The benefit would be, that we would have now a much better test coverage for the rest of the address module.

ST-DDT: Mark it as todo for now and maybe add an issue so it doesn't get lost.

Shinigami92: https://github.com/faker-js/faker/issues/321

ST-DDT:
Maybe remove the `chore(deps): update vitest` commit from this PR?

ST-DDT (approved):

Shinigami92:
> Maybe remove the `chore(deps): update vitest` commit from this PR?

I will rebase if #316 was merged

ST-DDT (approved):

test/address.spec.ts
```diff
@@ -1,782 +1,658 @@
...
-        .spyOn(faker.datatype, 'number')
-        .mockImplementation(() => 1);
-
-      const address = faker.address.streetAddress();
-      const parts = address.split(' ');
-      const expected = 4;
-
-      expect(parts[0].length, errorExpectDigits(expected)).toBe(expected);
-      expect(spy_address_streetName).toHaveBeenCalled();
+  for (let { seed, expectations } of seededRuns) {
```
Shinigami92, line 101:
```suggestion
  for (const { seed, expectations } of seededRuns) {
```

prisis (approved):

PR #196, opened 2022-01-17
docs: add zh-CN translation

renbaoshuo:

PR #289, opened 2022-01-24
docs: automatically generate api docs from source

ST-DDT:
Automatically generates the api documentation using typedocs.

Based on PR https://github.com/faker-js/faker/pull/207 by @ericjeker

- [x] Update Sidenav (manually)
- [x] Add script to generate docs
- [x] Generate docs (automatically)
- [x] Verify that the generated pages looks good
- [x] Remove the old api docs and generate them only during ci builds


JessicaSachs (approved):
Pulled it down and tried it. I love this. I think we could make the parameters a little prettier, but this is so much better than what exists, I think we should merge it now.

I feel a little awkward about committing generated source code... I think we should probably have CI re-build and publish the docs instead of committing them.

BUT! Regardless, I believe it is a good first step ❤️

ericjeker:
> I feel a little awkward about committing generated source code... I think we should probably have CI re-build and publish the docs instead of committing them.

Yes, same. I think we shouldn't commit that, otherwise the CI will create changes in the repo files, which is not good.

ericjeker:
This looks great!

Can we have the sidenav automatically generated when looping on the modules maybe? IIRC we can use Frontmatter for that? Or simply generate a Javascript array that is then processed in the `.vitepress/config`?

ST-DDT:
> Can we have the sidenav automatically generated when looping on the modules maybe?

I will do that.

> I think we should probably have CI re-build and publish the docs instead of committing them.

My original idea was to attach it to a pre commit hook, but on the fly generation is probably best.

> I think we could make the parameters a little prettier

Any suggestion how it should look like?
E.g. a table?

prisis:
I would only generate the code in the ci if the docs are build, so there is always up to date version live and put the docs/api to gitignore

Shinigami92:
> I would only generate the code in the ci if the docs are build, so there is always up to date version live and put the docs/api to gitignore

Also we have to do it like that, otherwise it could be hard to setup a CI that especially bypasses admin privileges to push into main branch

JessicaSachs:
Yeah, build it on CI and .gitignore the local files.

@ST-DDT I was looking for inspo when I did this PR review earlier... I think Styleguidist has a decent UI

https://react-kawaii.vercel.app/#/Components?id=cat
Screen Shot 2022-01-25 at 2 53 35 AM

Shinigami92:
What are your thoughts about localization support for that? https://github.com/faker-js/faker/pull/196

ST-DDT:
@JessicaSachs Like this?

I changed the command to manually generate the files to `pnpm run docs:api`, 
but they are also generated during `pnpm run build`.

scripts/apidoc.ts
```diff
@@ -6,9 +6,48 @@ import options from '../.prettierrc.cjs';
...
 const pathOutputJson = resolve(pathOutputDir, 'typedoc.json');
 
+function toBlock(comment?: TypeDoc.Comment): string {
+  return (
+    (comment?.shortText.trim() || 'Missing') +
+    (comment?.text ? '\n\n' + comment.text : '')
+  );
+}
+
+function escape(value: string) {
```
Shinigami92, line 21:
```suggestion
function escape(value: string): string {
```

scripts/apidoc.ts
```diff
@@ -6,9 +6,48 @@ import options from '../.prettierrc.cjs';
...
+  name: string,
+  type?: string,
+  def?: string,
+  comment?: TypeDoc.Comment
+): string {
+  if (def) {
+    def = `\`${def}\``;
+  } else {
+    def = '';
+  }
```
Shinigami92, line 35:
```suggestion
  def = def ? '' : `\`${def}\``
```

scripts/apidoc.ts
```diff
@@ -35,29 +74,32 @@ async function build() {
     .getChildrenByKind(TypeDoc.ReflectionKind.Namespace)[0]
     .getChildrenByKind(TypeDoc.ReflectionKind.Class);
 
+  let modulesPages: { text: string; link: string }[] = [];
```
Shinigami92, line 77:
```suggestion
  const modulesPages: Array<{ text: string; link: string }> = [];
```

Shinigami92, line 77:
```suggestion
  const modulesPages: Array<{ text: string; link: string }> = [];
```

scripts/apidoc.ts
```diff
@@ -6,9 +6,48 @@ import options from '../.prettierrc.cjs';
...
+    def +
+    ' | ' +
+    escape(toBlock(comment))
+      .replace(/\n{2,}/, '<br />')
+      .replace(/\n/, '') +
+    '|\n'
+  );
+}
+
 async function build() {
```
Shinigami92, line 51:
```suggestion
async function build(): Promise<void> {
```

scripts/apidoc.ts
```diff
@@ -87,17 +132,20 @@ async function build() {
...
 
           signatureTypeParameters.push(parameterName);
-          content += `- \\<${parameterName}\\> - ${
-            parameter.comment?.shortText || 'N/A'
-          }\n`;
-          if (parameter.comment?.text) {
-            content += '  ' + parameter.comment.text.replace('\n', '\n  ');
-          }
+          content += parameterRow(
+            '<' + parameterName + '>',
```
Shinigami92, line 144:
```suggestion
            `<${parameterName}>`,
```

scripts/apidoc.ts
```diff
@@ -150,10 +195,6 @@ async function build() {
           if (example.length > 50) {
             example = example.substring(0, 47) + '...';
           }
-          // Avoid updating the time.md all the time
-          if (lowerModuleName === 'time' && methodName === 'recent') {
-            example = '1643067304379';
-          }
 
           content += `faker.${lowerModuleName}.${methodName}()`;
           content += (example ? ' // => ' + example : '') + '\n';
```
Shinigami92, line 200:
```suggestion
          content += (example ? ` // => ${example}` : '') + '\n';
```

scripts/apidoc.ts
```diff
@@ -163,6 +204,15 @@ async function build() {
           );
         }
       }
+      let examples =
+        signature?.comment?.tags
+          .filter((tag) => tag.tagName === 'example')
+          .map((tag) => tag.text.trimEnd()) || [];
+
+      if (examples.length != 0) {
```
Shinigami92, line 212:
```suggestion
      if (examples.length !== 0) {
```

Shinigami92 (changes requested):

package.json
```diff
@@ -54,11 +54,12 @@
     "build:clean": "rimraf dist",
     "build:code": "esno ./scripts/bundle.ts",
     "build:types": "tsc --emitDeclarationOnly --outDir dist/types",
-    "build": "run-s build:clean build:code build:types",
+    "build": "run-s build:clean build:code build:types docs:api",
```
Shinigami92, line 57: build is running on all nodes in the CI matrix
but the docs only need to be build on one node version / one runner
therefore we need to organize this in a different way :thinking:

Shinigami92 (changes requested):

package.json
```diff
@@ -54,11 +54,12 @@
...
     "docs:build": "vitepress build docs",
     "docs:build:ci": "run-s build docs:build",
```
Shinigami92, line 59: I think we just need to move the `docs:api` to here and remove it from `build`

```suggestion
    "docs:build:ci": "run-s build docs:build docs:api",
```

package.json
Shinigami92:
```suggestion
    "build": "run-s build:clean build:code build:types",
```

package.json
Shinigami92: In addition to that we could document in e.g. the CONTRIBUTING.md that this `docs:api` needs to be runned once to prepare the `docs:dev`

Shinigami92: Or add two script like

```
"docs:dev": "run-s docs:dev:prepare docs:dev:run",
"docs:dev:prepare": "run-s build docs:api"
"docs:dev:run": "vitepress dev docs"
```

But 
1. it's a bit verbose
2. it would always run `build` and `docs:api` before starting the dev server, also if not need :thinking:

Shinigami92:
I will approve that now, but we should document
- which script should be run by humans in which order
- which scripts are required by CI/netlify-toml
- which are "just" sub-scripts
Could you open an issue for that, to track it?

JessicaSachs (approved):

PR #299, opened 2022-01-25
chore: add missing type on the system.commonFileName method

pkuczynski:


src/system.ts
```diff
@@ -58,7 +58,7 @@ export class System {
    * @method faker.system.commonFileName
    * @param ext
    */
-  commonFileName(ext): string {
+  commonFileName(ext?: string | number): string {
```
Shinigami92, line 61: **question:** where did you get these types from?
Currently it could be like any primitive like also boolean and bigint and such.
It's not clear to me where you did get this from

ST-DDT: Changing it to something string like it probably a good idea, but I it will be a breaking change.

Shinigami92: Maybe we can find something in the test suite how it wanted to be called.
I just currently say that it's not clear to me where the types are coming from

pkuczynski: I just _guessed_ the types, as they were the only ones which made sense as _extension_ really...

pkuczynski: `number` type represents any kind of number, so `int`, `float`, `bigint` right?

damienwebdev: @Shinigami92 you could check what was previously in definitely typed.

pkuczynski: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58218/files#diff-22605747c70de0fe07b686b06a1235f39c32faf85b32d66fc18c9623379d3f50L297

```js
commonFileName(ext?: string): string;
```

ST-DDT: I would go for the stricter string only version as used in DefinitelyTyped, because using unformatted numbers looks like an edge case for me.
E.g. you may use `.001` but I have never used just `.1` especially in random test data. But I would be okay with allowing numbers to, but then we shoulf make sure that it results in a decent extension even for non integer values.

Shinigami92: On top of that `0 || 'something' => 'something'`

ST-DDT: I assume this is the same for `''`. Lets got for string then and document that '' is considered to be not set.

Shinigami92: > I assume this is the same for `''`. Lets got for string then and document that '' is considered to be not set.

Oh... yeah, you are right :grimacing:

pkuczynski: Done!

ST-DDT (approved):

Shinigami92 (approved):

PR #300, opened 2022-01-25
fix: remove duplicated extension in system.filePath

pkuczynski:


Shinigami92:
For me it currently looks like you
1. change the runtime behavior => so 6.1
2. declare it as p4 but change the comment that explains that it have a double file extension?!

Please do not miss-use the priority labels, this is not a p4, max p3

ST-DDT:
> if this is even a bug at all 🤔

IMO it's a bug or at least unexpected behavior.

pkuczynski:
> 1. change the runtime behavior => so 6.1

Fixing a bug, so it is kind of "stability", but wasn't sure on your "terminology"

> 2. declare it as p4 but change the comment that explains that it have a double file extension?!

I am not sure what do you mean :/

> Please do not miss-use the priority labels, this is not a p4, max p3, if this is even a bug at all 🤔

I was not sure if my labels are correct, hence I reached out on discord asking for help with this. I strongly believe it is a bug though.

Shinigami92:
Tests are indicating that this was expected :thinking: 
Maybe we could just provide options and define how many extension will be added :thinking:

pkuczynski:
> Tests are indicating that this was expected 🤔
> Maybe we could just provide options and define how many extension will be added 🤔

These are the new tests. There were no old tests for this function AFAIK. And I can't really see why anyone would want to have dbl extensions in the `filePath` (by default)...?

pkuczynski:
> Maybe we could just provide options and define how many extension will be added 🤔

Added as a separate `feat` PR #395

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #307, opened 2022-01-26
fix: some return types

### LINKED ISSUE #305: Vet v6.0.0-alpha.5 types emitted types against v5.5.3
We need to ensure that the types emitted from our packages are API compatible with the previous DefinitelyTyped package.

See:

1. #166 
2. #198 
3. `faker.address.nearbyGPSCoordinate` (`coordinate` is a ReadOnlyArray in DefinitelyTyped)
4. `faker.address.companyName` has an arg of `format` which should be `string | number`
5. `faker.database.column` is implicitly return typed as `string` in 6.0.0-alpha.5 but should be set to return `string`
7. `faker.database.type()` is implicitly return typed as `string` in 6.0.0-alpha.5 but should be set to return `string`
8. `faker.database.collation()` is implicitly return typed as `string` in 6.0.0-alpha.5 but should be set to return `string`
9. `faker.database.engine()` is implicitly return typed as `string` in 6.0.0-alpha.5 but should be set to return `string`
10. `faker.finance.currencyCode()` is return typed `any`
11. `faker.finance.currencyName()` is return typed `any`
12. `faker.finance.currencySymbol()` is return typed `any`
13. `faker.hacker.abbreviation()` is implicitly return typed as `string`
14. `faker.hacker.adjective()` is implicitly return typed as `string`
15. `faker.hacker.noun()` is implicitly return typed as `string`
16. `faker.hacker.verb()` is implicitly return typed as `string`
17. `faker.hacker.ingverb()` is implicitly return typed as `string`
18. `faker.hacker.phrase()` is implicitly return typed as `string`

I'm sure there are a few others, but we need to check type by type to ensure spec compatibility.

ejcheng:
Partially fixes #305

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -38,7 +38,7 @@ export class Database {
    *
    * @method faker.database.column
    */
-  column() {
+  column(): string {
     return this.faker.random.arrayElement(
       this.faker.definitions.database.column
     );
@@ -49,7 +49,7 @@ export class Database {
    *
    * @method faker.database.type
    */
-  type() {
+  type(): string {
     return this.faker.random.arrayElement(this.faker.definitions.database.type);
   }
 
@@ -58,7 +58,7 @@ export class Database {
    *
    * @method faker.database.collation
    */
-  collation() {
+  collation(): string {
     return this.faker.random.arrayElement(
       this.faker.definitions.database.collation
     );
@@ -69,7 +69,7 @@ export class Database {
    *
    * @method faker.database.engine
    */
-  engine() {
+  engine(): string {
     return this.faker.random.arrayElement(
       this.faker.definitions.database.engine
     );
--- a/src/finance.ts
+++ b/src/finance.ts
@@ -153,7 +153,7 @@ export class Finance {
    *
    * @method faker.finance.currencyCode
    */
-  currencyCode() {
+  currencyCode(): string {
     // TODO @Shinigami92 2022-01-14: missing second parameter
     // @ts-expect-error
     return this.faker.random.objectElement(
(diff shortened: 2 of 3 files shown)


Shinigami92:
Currently one of our problems is that `arrayElement` doesn't guarantee type-safetynes (regrading to compile-time).
Our `definitions` currently are just _mostly_ arrays, but could be anything :shrug: 
I know that there are currently mostly `string[]` and we could just take over the types of `@types/faker`, but I would like to double-check that instead of assuming.

ST-DDT (dismissed):

ST-DDT:
IMO we should set these to the expected return types. Double checking with the existing ones would still be a good idea.

doughlass (dismissed):

import-brain:
> IMO we should set these to the expected return types. Double checking with the existing ones would still be a good idea.

All of the ones in this PR were checked against @types/faker latest version, they all seem reasonable and match up.

ST-DDT:
A probably wrong commit ended up here.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,7 +1,7 @@
 blank_issues_enabled: false
 contact_links:
   - name: Discord Chat
-    url: discord.gg/4qDjAmDj4P
+    url: chat.fakerjs.dev
```
ST-DDT, line 4: Please remove the last commit here.

ST-DDT (changes requested):

.github/ISSUE_TEMPLATE/config.yml
import-brain: Accidentally committed the chat.fakerjs.dev changes to this branch, just reset and force-pushed

ST-DDT (approved):

Shinigami92 (approved):

demipel8:
Wrong mention, sorry.

PR #309, opened 2022-01-26
fix: removed node v12 from the pipe

prisis:
Because vitest is not supporting this version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node_version: [12, 14, 16, 17]
+        node_version: [14, 16, 17]
         include:
           - os: macos-latest
             node_version: 16
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
   },
   "packageManager": "pnpm@6.26.1",
   "engines": {
-    "node": ">=12.0.0",
+    "node": ">=14.0.0",
     "npm": ">=7.0.0"
   }
 }


Shinigami92:
We should do this end of march

prisis:
Please check https://github.com/faker-js/faker/runs/4947011110?check_suite_focus=true as example the pipe never runs and vitest is returning 0 as error code

ST-DDT (dismissed):

ST-DDT:
The test pipeline does not work with node 12 and AFAICT the releases aren't build with node 12 either, so there is no point in keeping it.
If you wish to ensure that it can be build with node 12, then we should at least explicit turn of the test runs.

But I don't have a strong opinion on when to remove this.

Is there anything compatibility wise we have to consider when removing it?
(Aka does our build process ensure node 12 compatibility regardless of our build setup?)

doughlass (dismissed):

Shinigami92 (approved):
@damienwebdev @JessicaSachs Please point out in next release notes that we officially drop node 12 support

Shinigami92 (changes requested):
This line needs to be updated to v14

https://github.com/faker-js/faker/blob/3242dfc4b2f566931c3687173f6361a6384b6247/package.json#L99

ST-DDT:
Couldn't we just drop official support for node 12 but give the users at least the chance to use it?
AFAICT we don't use any node 14+ features.

Shinigami92:
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines
As far as we don't set `engine-strict`, only a warning will show up

Shinigami92 (approved):

ST-DDT (approved):

PR #310, opened 2022-01-26
chore(deps): configure and update lint-staged

Shinigami92:
Extracted from https://github.com/faker-js/faker/pull/151

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,8 +81,11 @@
     "*": [
       "prettier --write --ignore-unknown"
     ],
-    "/dist/**/*.{js,ts}": [
-      "eslint --ext .js,.ts"
+    "**/*.md": [
+      "eslint --ext .md"
+    ],
+    "{scripts,src,test}/**/*.ts": [
+      "eslint --ext .ts"
     ]
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


ST-DDT:
Is it intentional, that this doesn't setup the `pnpm run lint` command?

Shinigami92:
> Is it intentional, that this doesn't setup the `pnpm run lint` command?

Yes!
This will be done in separate PR, cause it will initially trigger many CI errors

ST-DDT (dismissed):

ST-DDT:
If I run the command locally I get the following error:

````txt
$ pnpm eslint .

Oops! Something went wrong! :(

ESLint: 6.8.0.

No files matching the pattern "." were found.
Please check for typing mistakes in the pattern.
````

Shinigami92:
The eslint setup is currently broken
It is _disabled_ in https://github.com/faker-js/faker/blob/3242dfc4b2f566931c3687173f6361a6384b6247/package.json#L63

#151 is trying to bring eslint support back to the repo
But because it is soo huge, I want to split up the PR to improve and speedup review process

ST-DDT:
I'm not familiar enough with eslint to review this PR without it working, so others have to step up for review.

Shinigami92:
I found out that I cannot split this up, cause we depend on the eslint version and the typescript plugin :slightly_frowning_face:

PR #318, opened 2022-01-27
test: use run for coverage

Shinigami92:


ST-DDT (approved):

griest024:
why is this needed?

ST-DDT:
Otherwise it will stay it the watch mode and not output the coverage.
(At least, that's what my local tests showed me as diff)

@Shinigami92 Maintainers should still write a description with a summary/explanation for their changes.

griest024 (approved):

Issue #322, opened 2022-01-27
Document npm scripts


ST-DDT:
We should document

- which script should be run by humans in which order
- which scripts are required by CI/netlify-toml
- which are "just" sub-scripts

Potentially a part of #308

_Originally posted by @Shinigami92 in https://github.com/faker-js/faker/issues/289#issuecomment-1023405156_

xDivisionByZerox:
This has been made a lot simpler #1846, because `preflight` is now the only script that needs to be run by a human.

In which form should the documentation be provided?

matthewmayer:
Could probably expand the line in CONTRIBUTING which says:

`Please make sure that you run pnpm run preflight before making a PR to ensure that everything is working from the start.`

To something like

```
Please make sure that you run `pnpm run preflight` before making a PR to ensure that everything is working from the start. This is a shorthand for running the following scripts in order:
- `pnpm install` - installs npm modules defined in package.json
- `pnpm run generate:locales` - generates locale files
- `pnpm run generate:api-doc` - generates API documentation
- `pnpm run format` - runs prettify to format code
- `pnpm run lint` - runs ESLint
- `pnpm run build` - builds the code, both CommonJS and ESM versions
- `pnpm run test:update-snapshots ` - runs all tests, and updates any snapshots if needed
- `pnpm run ts-check` - checks TypeScript
```

xDivisionByZerox:
That is a really good idea. 

Do you want to open a PR to make this change? Otherwise, I can do one and add you as a co-author.

matthewmayer:
you go ahead? you may be able to write nicer descriptions. i wasn't too clear what pnpm run ts-check actually does.

PR #325, opened 2022-01-27
build: add clean command

ST-DDT:


Shinigami92 (approved):

package.json
```diff
@@ -51,6 +51,7 @@
     "tsconfig.json"
   ],
   "scripts": {
+    "clean": "rimraf coverage dist docs/.vitepress/dist pnpm-lock.yaml node_modules",
```
griest024, line 54: `node_modules` and the lock file as well? Whats the rationale behind that?

ST-DDT: Fix issues with merge/rebase conflicts where pnpm picks up the old config regardless of it being updated.

griest024: okay I am not familiar with `pnpm`, I'll trust that this is the correct workflow

griest024 (approved):

package.json
Shinigami92: This is not pnpm related! It is the same for npm and yarn.
Deleting the lockfile locally and make a fresh install will generate a full new lock file instead of only the first level of dependencies.
pnpm also have a CLI flag I think like install -r for recursive, but this is kinda doing the same.

The clean also removes folders like dist and such so you can be clear that there is no caching involved somewhere at all. This especially can help when switching between branches that have e.g. differences in code or dependencies.

griest024: It seems to me that the lockfile is not useful for anything other than CI if it gets regenerated every time. Imo it should only change when deps are changed. With this workflow it seems as if it will change with features and big fixes.

Shinigami92: No no... you are indirectly right!
`pnpm run clean && pnpm install` should only be run by human!
It's a helper.
The lockfile should only be committed and updated if necessary.
And yes, it is especially useful for CI

griest024: Hm maybe this pr should have an edit to `CONTRIBUTING.md` explaining when and if this command should be used.

Shinigami92: 😅 please see #322

PR #335, opened 2022-01-28
fix: remove redundant precision from datatype.datetime options

pkuczynski:
The original DT types for this method looked like this:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b5bf4061b0811d2e16dc3c6b4c3c15881b73290f/types/faker/index.d.ts#L106-L107

```
datetime(max?: number): Date;
datetime(options?: { min?: number | undefined; max?: number | undefined }): Date;
```

When converting js to ts, it got accidentally typed taking all options of the `datatype.number`, but using `precision` here makes no sense, as expected output should be an `integer`. `new Date()` can handle floating numbers here

```js
> new Date(1020003333333.1345)
2002-04-28T14:15:33.333Z
> new Date(1020003333333)
2002-04-28T14:15:33.333Z
```

But I think we should build the interface in a way, which makes more sense.

Additionally I made some changes in the code to prevent `precision` leaking into `datatype.number`


pkuczynski:
Test failure in `test/unique.spec.ts` which is not relevant to changes from this PR.

ST-DDT (approved):

Shinigami92:
#312 needs to be merged first (I know you know it, but just to be clear for other maintainers so they don't accidentally merge this :slightly_smiling_face:)

ST-DDT:
@Shinigami92 Did you link the wrong PR?
`datetime()` is in `datatype` and not in `date`.

Shinigami92:
> @Shinigami92 Did you link the wrong PR? `datetime()` is in `datatype` and not in `date`.

Sorry, I think I loose the overview ...
So I think I need to takle next the datetype tests, will start with that soon

pkuczynski:
No worries! One step at the time :)

Shinigami92 (approved):

prisis (approved):

Issue #203, opened 2022-01-17
EPIC: Add JSDoc descriptions


Shinigami92:
### Clear and concise description of the problem

Most (if not all) of the functions already have JSDoc, but these are just simple e.g. `@param <fieldName>` without the descriptive text.
We found some documentation like https://github.com/faker-js/faker/pull/160, https://fakerjs.dev/api/datatype and e.g. these [schema definitions](https://github.com/faker-js/faker/blob/a973ee1ce64d2b593f64b633e927382280da8f27/src/internet.ts#L17-L180) that were just for generating documentation.

### Suggested solution

We would like to write these messages to the JSDoc comments, so that IDEs like VSCode can provide them via tooltips.

### Additional context

After this is done, we think that we can remove the schema definitions.

| Module      | Assignees   | #PR        | Done |
| ----------- | ----------- | ---------- | ---- |
| address.ts  | @mbokinala, @ST-DDT  | #230       |  ✔️ |
| animal.ts   | @mbokinala  | #250, #417 | ✔️   |
| commerce.ts | @ST-DDT     | #351       | ✔️ |
| company.ts  | @ST-DDT     | #423       | ✔️   |
| database.ts | @ST-DDT     | #424       | ✔️   |
| datatype.ts | @pkuczynski | #337       | ✔️   |
| date.ts     | @ST-DDT     | #427       | ✔️   |
| fake.ts     | @ST-DDT     | #406       | ✔️   |
| finance.ts  | @ST-DDT     | #434       | ✔️   |
| git.ts      | @ST-DDT     | #435       | ✔️   |
| hacker.ts   | @ST-DDT     | #436       | ✔️   |
| helpers.ts  | @ST-DDT     | #442 | ✔️ |
| image.ts    | @ST-DDT     | #437       | ✔️   |
| internet.ts | @ST-DDT | #447 | ✔️ |
| lorem.ts    | @ST-DDT | #458  | ✔️ |
| mersenne.ts | @ST-DDT     | #438       | ✔️   |
| music.ts    | @ST-DDT     | #390       | ✔️   |
| name.ts     | @ST-DDT     | #391       | ✔️   |
| phone.ts    | @pkuczynski | #397       | ✔️   |
| random.ts   | @pkuczynski @ST-DDT  | #408 #562  | ✔️ |
| system.ts   | @pkuczynski | #298       | ✔️   |
| time.ts     | @pkuczynski | #288       | ✔️   |
[...]

PR #337, opened 2022-01-28
chore: fix JSDoc comments in datatype.ts

pkuczynski:
Refs #203


src/datatype.ts
```diff
@@ -89,11 +109,17 @@ export class Datatype {
   }
 
   /**
-   * Returns a Date object using a random number of milliseconds since 1. Jan 1970 UTC
-   * Caveat: seeding is not working
+   * Returns a Date object using a random number of milliseconds since
+   * the [Unix Epoch](https://en.wikipedia.org/wiki/Unix_time) (1 January 1970 UTC).
```
Shinigami92, line 113: Nice idea :raised_hands:

src/datatype.ts
```diff
@@ -19,10 +19,20 @@ export class Datatype {
   }
 
   /**
-   * Returns a single random number based on a max number or range.
+   * Returns a single random number for the given max value or range and precision.
    *
-   * @method faker.datatype.number
-   * @param options
+   * @param [options] Maximum value or options object.
```
Shinigami92, line 24: I think these brackets are not needed :thinking: 
Could you check how it looks in VSCode?

As far as I know, these are needed in JavaScript, but not in TypeScript

Shinigami92: an example:

image

seems that it works, but the default values will get eaten :thinking:

pkuczynski: Brackets means it's an optional param. See https://jsdoc.app/tags-param.html#optional-parameters-and-default-values. I can't check how it looks in VSCode, as I am using JetBrains software... Here it looks like this:

image

Shinigami92:
@ST-DDT could you test out if your api docs generation script can handle this param syntax?

ST-DDT:
@Shinigami92 No it doesn't work.

The typedoc.json does not contain them either.

grafik

src/datatype.ts
```diff
@@ -19,10 +19,20 @@ export class Datatype {
   }
 
   /**
-   * Returns a single random number based on a max number or range.
+   * Returns a single random number for the given max value or range and precision.
```
ST-DDT, line 22:
```suggestion
   * Returns a single random number between zero and the given max value or the given range with the specified precision.
   * The bounds are inclusive.
```

pkuczynski:
> @Shinigami92 No it doesn't work.

There is an issue reported on typedoc and a PR with a fix: https://github.com/TypeStrong/typedoc/issues/567

ST-DDT:
@pkuczynski AFAICT That issue/PR isn't going to be fixed/merged soon and even if it gets merged, it does not solve the problem for us. See https://github.com/TypeStrong/typedoc/pull/1810#pullrequestreview-866938016

So maybe we should remove the `[]` for now?

Shinigami92 (changes requested):
Hey @pkuczynski, I also found out that the usage of square-brackets are not a good idea, cause it could be that JetBrains support it, but Microsoft doesn't support them in VSCode.
So we should try to find a solution to write docs for both worlds.

I also think you are mixing JSDoc and (what I call) TSDoc. It's kinda like JavaDoc vs KotlinDoc, they are similar but not the same.

So what I would suggest is something like
```
/**
 * @param options.min Lower bound for generated number. Defaults to `0`.
 */
```

pkuczynski:
I guess there is no other option than remove those defaults for now, and maybe add them at some point in the future when vscode and typedoc will support them. It's really a shame though, and it displays in the docs much nicer when the `default value` column is provided for a param vs when the information is embedded in the desc.

pkuczynski:
Done. Could you please re-review...?

Shinigami92 (dismissed):

src/datatype.ts
```diff
@@ -64,10 +75,20 @@ export class Datatype {
   }
 
   /**
-   * Returns a single random floating-point number based on a max number or range.
+   * Returns a single random floating-point number for the given precision or range and precision.
+   *
+   * @param options] Precision or options object.
```
ST-DDT, line 80:
```suggestion
   * @param options Precision or options object.
```

src/datatype.ts
```diff
@@ -141,9 +170,10 @@ export class Datatype {
   }
 
   /**
-   * uuid
+   * Returns UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
```
ST-DDT, line 173:
```suggestion
   * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
```

src/datatype.ts
```diff
@@ -156,23 +186,28 @@ export class Datatype {
   }
 
   /**
-   * boolean
+   * Returns boolean value true or false.
```
ST-DDT, line 189:
```suggestion
   * Returns the boolean value true or false.
```

src/datatype.ts
```diff
@@ -156,23 +186,28 @@ export class Datatype {
...
+   * @example
+   * faker.datatype.boolean() // false
    */
   boolean(): boolean {
     return !!this.faker.datatype.number(1);
   }
 
   /**
-   * hexaDecimal
+   * Returns [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) number.
```
ST-DDT, line 199:
```suggestion
   * Returns a [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) number.
```

src/datatype.ts
```diff
@@ -203,14 +238,15 @@ export class Datatype {
   }
 
   /**
-   * Returns json object with 7 pre-defined properties
+   * Returns string representing JSON object with 7 pre-defined properties.
```
ST-DDT, line 241:
```suggestion
   * Returns a string representing JSON object with 7 pre-defined properties.
```

src/datatype.ts
```diff
@@ -221,13 +257,15 @@ export class Datatype {
   }
 
   /**
-   * Returns an array with values generated by faker.datatype.number and faker.datatype.string
+   * Returns array with random strings and numbers.
```
ST-DDT, line 260:
```suggestion
   * Returns an array with random strings and numbers.
```

ST-DDT (changes requested):

src/datatype.ts
pkuczynski: omg! thanks!

ST-DDT (approved):

Shinigami92 (approved):

ST-DDT (approved):

PR #338, opened 2022-01-28
chore: welcome only new issue creators

pkuczynski:
Following [discussion](https://discord.com/channels/929487054990110771/936658360802115584/936661295854403625) this is an attempt to avoid creating the welcome message all the time. See https://actions-cool.github.io/issues-helper/en-US/base#welcome for more details. I also raised https://github.com/actions-cool/issues-helper/issues/102 to see if there are other alternatives...

Alternatively, we could use different action for this purpose, for example https://github.com/marketplace/actions/welcome-new-contributors


import-brain (approved):

ST-DDT (approved):

prisis (approved):

Shinigami92:
We can test it out, but I'm not sure if this is really the thing we wanted.
I copied the workflow over from Vite, and there we have this "problem" also, and decided to add the `... or another community member` to the message.
The `help wanted` label also indicates more an intent to be visible by sites like http://github-help-wanted.com/
A solution if this PR does not work as intended would be to just ignore as _internal_ contributor.

pkuczynski:
`Help wanted` should be still applied. If am not mistaken the only diff between `welcome` and `create comment` is that the first one does this only once per user...

Maybe I am wrong but it's wort to try. We can always revert if this does not work :)

Shinigami92:
we will just test it, if hell breaks lose we can revert it :P

PR #340, opened 2022-01-28
chore: rename phone_number to phone

pkuczynski:
Every other file is named like the class, except this one:

```
class: Phone
filename: phone_number
```

I aligned this with other files and simply renamed this one to match the class name.


pkuczynski:
The same thing should possibly happen to locales. Let me know if I should do it (in this or separate PR)

Shinigami92:
Where were not sure right now if the `fake` function somehow dynamically requests data with a pattern that uses `phone_number`
I would like to first update the tests for that.
Also I'm not sure if this counts to the project stability :thinking: 
I will move it to the parallel milestone

pkuczynski:
`fake` is referring to classes, not file names:

https://github.com/faker-js/faker/blob/main/test/fake.spec.ts#L7

and the class itself is using definitions in this way:

https://github.com/faker-js/faker/blob/main/src/phone_number.ts#L36

So I think its pretty safe change and its always good to verify via tests...

ST-DDT (approved):

prisis (approved):

import-brain (approved):

Issue #327, opened 2022-01-28
The last returns and example section on each api docs page vanishes after reload


ST-DDT:
### Describe the bug

This behaves exactly like https://github.com/faker-js/faker/issues/324 but only for the returns and example section of the last element.

````md
**Returns:** string | number

```ts
faker.address.zipCodeByState(state?: string): string | number
```
````

During reload:

![grafik](https://user-images.githubusercontent.com/1579362/151519551-131a4196-1e1b-422e-a612-45634cb55d04.png)

After a short period:

![grafik](https://user-images.githubusercontent.com/1579362/151519631-cf3077b7-9568-4ced-9b5a-f62d031ffded.png)

### Reproduction

- Enter https://fakerjs.dev/api/address.html#zip-code-by-state into your address bar and confirm

### Additional Info

We checked it and confirmed that it affects FF and Chrome, but only when serving.
You can avoid this by turning JS off or using dev mode.

ST-DDT:
The error is not fixed yet (and is wasn't related to #330 in the first place

https://user-images.githubusercontent.com/1579362/151629455-9e7bd826-082b-4414-a06f-c536e4cb30df.mp4

)

JessicaSachs:
Grrrrrrrr

PR #341, opened 2022-01-29
fix: removing toc to fix the docs html

JessicaSachs:
This removes the TOC from the codegen. It's okay. We'll improve the Docs UX shortly.

Should solve #327


cypress/e2e/smoke.cy.ts
```diff
@@ -1,11 +1,25 @@
...
-  //   cy.visit('/playground/');
-  //   cy.contains('Playground under construction');
-  // });
+  it('renders this last code example in the code', () => {
+    // Click on any section in the sidebar
+    cy.get('@firstSectionLink')
+      .click()
+      // Make sure the number of code examples is the same between reloads
+      .get('.container pre code')
+      .then(($codeBlocks) => {
```
pkuczynski, line 16: Minor thing, but does cypress not support async/await?

JessicaSachs: No. It's a long story. These aren't promises, they're Cypress Chainables (Bluebird, essentially)

pkuczynski: Cypress has a lot of weird quirks to my taste ;) But anyway, thanks for explaining...

ST-DDT (approved):
Strange cause, but works.

cypress/e2e/smoke.cy.ts
```diff
@@ -1,11 +1,25 @@
 describe('Smoke Test', () => {
-  it('compiles the guide page', () => {
+  beforeEach(() => {
     cy.visit('/guide/');
+    cy.get('a[href="/api/animal.html"]').as('firstSectionLink');
+  });
```
Shinigami92, line 5: **formatting:** just a missing empty line here

```suggestion
  });

```

cypress/e2e/smoke.cy.ts
```diff
@@ -1,11 +1,25 @@
...
+    // Click on any section in the sidebar
+    cy.get('@firstSectionLink')
+      .click()
+      // Make sure the number of code examples is the same between reloads
+      .get('.container pre code')
+      .then(($codeBlocks) => {
+        // Trigger a reload
+        cy.reload()
+          // Give the runtime a chance to update/fail
+          .wait(500)
```
Shinigami92, line 20: **suggestion (perf):** Do we have any chance to not use static time values here?
Maybe a wait on animation or something like this?

Shinigami92 (commented):
I'm okay for now with this change/removal of toc, because in the sidebar you can still the what functions are there.

cypress/e2e/smoke.cy.ts
Shinigami92: We will extract this into another issue, so we can merge this PR now

Shinigami92 (approved):

ST-DDT (approved):

PR #343, opened 2022-01-29
fix: datatype.datetime should use static boundaries

### LINKED ISSUE #334: datatype.datetime seed might not be working
### Describe the bug

There was this comment in the jsdoc
https://github.com/faker-js/faker/blob/60b71fd257c2b95a6997f5aac516b88327a78f4b/src/datatype.ts#L92-L98

We are not sure if this is true or not, however this method relies underneath on `datatype.number` which might be also affected.

Discussed here: https://discord.com/channels/929487054990110771/929544761122111498/936559737540071434

### Reproduction

Needs investigation

### Additional Info

_No response_

ST-DDT:
Fixes #334 

`new Date().setFullYear(...)` does not return a consistent value, as the actual time is not fixed an thus depends on the execution time + timezone. 

This PR uses UTC based static values instead.

### DIFF
--- a/src/datatype.ts
+++ b/src/datatype.ts
@@ -130,11 +130,11 @@ export class Datatype {
     let max = typeof options === 'number' ? options : options?.max;
 
     if (typeof min === 'undefined' || min < minMax * -1) {
-      min = new Date().setFullYear(1990, 1, 1);
+      min = Date.UTC(1990, 0);
     }
 
     if (typeof max === 'undefined' || max > minMax) {
-      max = new Date().setFullYear(2100, 1, 1);
+      max = Date.UTC(2100, 0);
     }
 
     return new Date(this.faker.datatype.number({ min, max }));
--- a/test/datatype.spec.ts
+++ b/test/datatype.spec.ts
@@ -22,8 +22,11 @@ const seededRuns = [
         withMinAndMaxAndPrecision: -0.4261,
       },
       datetime: {
-        // TODO @Shinigami92 2022-01-29: We will fix the deterministic in #343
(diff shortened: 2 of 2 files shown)


pkuczynski (approved):

pkuczynski:
Nice catch and fix! We use `new Date()` in `date.te` too, and I think I have seen some test failing around the midnight where the local time zone was the next day vs utc. Maybe it would be worth to apply it over there too?

pkuczynski (approved):

Shinigami92:
Oh my god (╯°□°)╯︵ ┻━┻

ST-DDT:
> Nice catch and fix! We use `new Date()` in `date.te` too, and I think I have seen some test failing around the midnight where the local time zone was the next day vs utc. Maybe it would be worth to apply it over there too?

The date module uses either `new Date()` with refers to a relative date or uses the `dateRef` parameter.
So it should be stable. If you catch a failing test, please report it so that we can analyze it and check only the related methods.

ST-DDT:
This needs to be rebased after #344 is merged.

ST-DDT:
Rebased

Shinigami92 (dismissed):

pkuczynski (dismissed):
Thats what I mentioned the other day. Cool that you fixed it :)

Shinigami92 (approved):

Shinigami92 (dismissed):

Shinigami92 (dismissed):

Shinigami92 (approved):

prisis (approved):

PR #346, opened 2022-01-29
chore(docs): reduce console spam and trim examples from code

ST-DDT:
Just some cleanup for things I forgot to remove previously.


Shinigami92 (approved):

scripts/apidoc.ts
```diff
@@ -222,10 +222,12 @@ async function build(): Promise<void> {
 
           content += `faker.${lowerModuleName}.${methodName}()`;
           content += (example ? ` // => ${example}` : '') + '\n';
-        } catch {
-          console.log(
-            `Failed to call: faker.${lowerModuleName}${methodName}()`
-          );
+        } catch (error) {
+          // If there is an error, we simply don't show the example call.
```
damienwebdev, line 226: Did you intend to leave this commented out code here?

damienwebdev (dismissed):

scripts/apidoc.ts
ST-DDT: Yes, this is intended to be useful for easy debugging. But  I can remove it as well. Which would you prefer?

Shinigami92 (approved):

scripts/apidoc.ts
prisis: Is this error happening that often?

ST-DDT: Well, we don't care about the error at all (because usually it is caused by functions that require parameters, but the generator doesn't give them any.
I guess I just remove it then.

Shinigami92 (dismissed):

Shinigami92 (approved):

prisis (approved):

PR #347, opened 2022-01-29
fix: fake is unable to return empty strings

ST-DDT:
Fake should behave exactly like the methods being called directly.
That is if the called method returns an empty string, then fake should do the same (and not fail like it does right now).

### DIFF
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -116,6 +116,10 @@ export class Fake {
     // replace the found tag with the returned fake value
     res = str.replace('{{' + token + '}}', result);
 
+    if (res === '') {
+      return '';
+    }
+
     // return the response recursively until we are done finding all tags
     return this.fake(res);
   }
--- a/test/fake.spec.ts
+++ b/test/fake.spec.ts
@@ -46,5 +46,9 @@ describe('fake', () => {
         Error('Invalid method: address.foo')
       );
     });
+
+    it('should be able to return empty strings', () => {
+      expect(faker.fake('{{helpers.repeatString}}')).toBe('');
+    });
   });
 });


src/fake.ts
```diff
@@ -114,6 +114,10 @@ export class Fake {
     // replace the found tag with the returned fake value
     res = str.replace('{{' + token + '}}', result);
```
Shinigami92, line 115: I think we can bring down the declaration of `res` down to this line, as it is not used above anyways

src/fake.ts
```diff
@@ -114,6 +114,10 @@ export class Fake {
...
 
+    if (res === '') {
+      return '';
+    }
+
     // return the response recursively until we are done finding all tags
     return this.fake(res);
```
Shinigami92, line 122: Not sure if we want to return empty string from `fake` instead, when `res` is empty

ST-DDT: IMO fake should behave exactly like calling the referenced methods directly.
Including returning an empty string of the called method does.
If we really don't want this, then we should at least throw here instead of in the next invocation to preserve enough context for debugging purposes.

src/fake.ts
ST-DDT: I think there are plenty of ways to improve this. However, these improvements are out of the scope of this PR.

src/fake.ts
prisis: The question is, should the function not just return the not resolved value out

ST-DDT: From discord:

> IMO `fake("{{x}}")` should behave like `x()`, thus if `x()` does return `''`, so should `fake("{{x}}")`.

Shinigami92 (dismissed):

import-brain (dismissed):

Shinigami92 (dismissed):

Shinigami92 (dismissed):

Shinigami92 (dismissed):

Shinigami92 (approved):

prisis (approved):

PR #351, opened 2022-01-29
docs: fix JSDoc comments in commerce.ts

ST-DDT:
Refs #203


Shinigami92 (approved):

Shinigami92 (approved):

src/commerce.ts
```diff
@@ -49,13 +55,15 @@ export class Commerce {
...
   /**
-   * price
+   * Generates a price between min and max (inclusive).
    *
-   * @method faker.commerce.price
-   * @param min
-   * @param max
-   * @param dec
-   * @param symbol
+   * @example
```
pkuczynski, line 60: I believe `@param` should come before `@example`?  At least that's how we structured in other files. I also provided examples for all optional params (at least most of them), so users can quickly see different input-output. See https://github.com/faker-js/faker/pull/337/files#diff-5699f20396717381133051426f9769465b1d8d460c942def407bb4f03907d5c3R31 as an example...

src/commerce.ts
```diff
@@ -1,5 +1,8 @@
 import type { Faker } from '.';
 
+/**
```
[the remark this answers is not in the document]
ST-DDT: Yes. That way we can add some kind of introduction to the module in the docs.
(Currently all is missing there)

src/commerce.ts
ST-DDT: I'll add the additional examples.

I myself don't have a preference for the location of the examples, but I was told to add them to the front.
@Shinigami92: What should we use as the standard?

Shinigami92: I personally was used to that examples are before params :shrug: 
Is there a eslint-plugin-jsdoc that says something about an order? If so I would like to stick to their default, cause I would like to set it up later.

pkuczynski: I don't see such a rule (or at least on a first look). But we should stick to the same order in all files, right?

Let me add eslint rule as a PR and let's see what it will report, ok?

Shinigami92: Yeah, I don't see it either :thinking: (https://github.com/gajus/eslint-plugin-jsdoc)

@brettz9 do you know such a rule in your plugin?
We want to sort/keep order of `@param` and `@example` but don't know if it's better to make example first or last :thinking:

brettz9: No, @Shinigami92 , there's currently no such rule. 

It looks like https://github.com/homer0/packages/tree/main/packages/public/prettier-plugin-jsdoc has one, but it is prettier specific. (though it

JSDoc does not prescribe any order, at least as far as anything I can recall.

brettz9: `comment-parser`, which we are using, does have a parser and stringifier, so such a rule should be doable. It'd be a good rule to have, in my opinion, but I'm rather constrained with other work these days to take it on.

ST-DDT: I fixed the examples and move the section.

Shinigami92 (approved):

pkuczynski (approved):

prisis (approved):

Issue #336, opened 2022-01-28
datatype.hexaDecimal should be called datatype.hexadecimal


pkuczynski:
### Describe the bug

The correct term is `hexadecimal`, without the capital `D`. See https://en.wikipedia.org/wiki/Hexadecimal

### Reproduction

N/A

import-brain:
Should alphaNumeric also be renamed to alphanumeric?

pkuczynski:
> Should alphaNumeric also be renamed to alphanumeric?

I think so...

prisis:
Let's go for aliases for now and add a deprecated notice, in 7 or 8 we can remove it.

Shinigami92:
As @damienwebdev and @mastermatt responded here https://github.com/faker-js/faker/pull/354#issuecomment-1026950523, I will close this issue for now as unwanted.
If there is still really a need, please reopen with new argumentations why it's useful.

PR #354, opened 2022-01-29
fix: rename hexaDecimal to hexadecimal

kupuma-ru21:
https://github.com/faker-js/faker/issues/336

I have dealt with the above issues.


ST-DDT (changes requested):
Your last commit reverts a lot of test upgrades.

Please revert/remove it and try again.

kupuma-ru21:
> Your last commit reverts a lot of test upgrades.
> 
> Please revert/remove it and try again. If you need help with that we can help you once v6.0 is released.

Reverted the last commit.

Please confirm.

ST-DDT:
This didn't work as intended.

Please run

````sh
git fetch --all
git reset --hard 262208750b712f793e8dadae40b295248de1cc73
git rebase origin/main
# Fix any rebase conflicts
#git rebase --continue
# Fix any linting, formatting or test issues
#git commit ...
git push -f
````

File: package.json [hunk 1]
```diff
@@ -11,7 +11,7 @@
     "fake data",
     "fake-data",
     "fake-generator",
-    "fake-data-generator",
+    "fake-data-generatgit rebase origin/mainor",
```
ST-DDT, line 14:
```suggestion
    "fake-data-generator",
```

ST-DDT (changes requested):
That looks better to me. You have a minor typo though. 😉 
Did you rebase all the way? Because there are still some conflicts.

test/datatype.spec.ts
```diff
@@ -248,7 +248,6 @@ describe('datatype', () => {
           expect(actual).toEqual(expectations[functionName].noArgs);
         });
       }
-
```
ST-DDT, line 251: Please restore this whitespace between the two sections.

package.json
kupuma-ru21: Fixed.
https://github.com/faker-js/faker/pull/354/commits/acea4a5d189e86d248166769eb70b3531623a9a8

test/datatype.spec.ts
kupuma-ru21: Fixed.
https://github.com/faker-js/faker/pull/354/commits/38f4a68ca6b764473a7c0668bf95feb53a8a413b

Shinigami92 (changes requested):
The current impl would introduce a breaking API change.
We could introduce an alias that calls the other function.

kupuma-ru21:
> The current impl would introduce a breaking API change. We could introduce an alias that calls the other function.

kupuma-ru21:
> The current impl would introduce a breaking API change. We could introduce an alias that calls the other function.

I'm sorry.
I didn't understand your point specifically.
What correction do I need to make?

damienwebdev:
@kupuma-ru21 

We have the function `hexadecimal`. Introducing a new function provides no value to the codebase other than breaking a consuming developer's codebase when they have previously relied on `hexaDecimal`. 

Put a different way, imagine the cost/benefit analysis of this commit:

1. Cost: All-consuming developers are forced to rename usage of `hexaDecimal` to `hexadecimal`, just making upgrades that much harder.
2. Benefit: It looks nicer.

Clearly, this is not worth it. As such, I'm closing this PR.

mastermatt:
@damienwebdev could you comment/close the initialing issue please https://github.com/faker-js/faker/issues/336

pkuczynski:
@damienwebdev this is a bit weak argumentation, given the fact that we are planing other of this type changes, for example moving `random.word` to `word.any`. Fact is that `hexaDecimal` is incorrect term. And it's shame that faker would keep using it, instead of keeping things clean. Refactoring should not be such of a big deal for customers of this library...

Shinigami92:
issues and PRs with label `s: needs decision` will be discussed in upcoming maintainer-team meeting

Shinigami92:
In yesterdays maintainers-team-meeting we discussed that we want to add `hexadecimal` and copy over the implementation from `hexaDecimal`
Then `hexaDecimal` will call `hexadecimal` and warn the user with a deprecation message and for the alternative
Also the JSDoc tag `@deprectated` needs to be added to `hexaDecimal`

We are doing this already in other places the same way, e.g. 
https://github.com/faker-js/faker/blob/c1c60649ee554f81887ab618ae8b2e2bd5a27d55/src/random.ts#L50-L59

pkuczynski:
@kupuma-ru21 let me know if you wanna continue this way, or I should take over as the author of the original issue #336

Shinigami92:
> @kupuma-ru21 let me know if you wanna continue this way, or I should take over as the author of the original issue #336

One week passed without any response of the author, @pkuczynski please proceed

PR #358, opened 2022-01-29
chore: add more type annotations

import-brain:
All added type annotations were checked against @types/faker@5.5.9 (most recent version that contains a index.d.ts file)


import-brain:
Not sure which milestone this would go under, please add one for me :)

ST-DDT (approved):

ST-DDT:
These types match the expectation and the underlying types should match, so lets get that in now.

PR #360, opened 2022-01-29
chore: use recommended no-var-requires

Shinigami92:


ST-DDT (approved):
Is it possible to turn this rule off for tests only?

Shinigami92:
> Is it possible to turn this rule off for tests only?

Yes, should I?

Shinigami92:
> Is it possible to turn this rule off for tests only?

Yes, should I?

prisis (approved):

Shinigami92:
I discussed with @ST-DDT that we don't want to **implicitly** allow it for tests, we can **explicitly** use the comments

PR #363, opened 2022-01-30
feat(types): provide strong typing for locales

ST-DDT:
In this first stage I only compare what the methods that use the definitions (likely) return and compare that to the locale en.


src/locales/en/address/index.ts
```diff
@@ -20,28 +21,38 @@ import street_name from './street_name';
...
+  time_zone,
+
+  // Extra
   street_address,
   street_name,
-  street_suffix,
-  time_zone,
-};
+  default_country,
+  building_number,
```
ST-DDT, line 55: Can someone please check whether and how these can currently be accessed and properly used by faker?
(The default country is usable as is via the definitions.address.default_country, but IMO for the others methods are missing for now).

ST-DDT:
The changes in the locales/index.ts files are only temporary to support filtering unneeded entries.

ST-DDT:
Please mark all types that I should inline.

I will apply the types/remove all anys programmatically from the LocaleDefinition children once this and this and #252 is merged.
(Via script)

src/definitions/definitions.ts
```diff
@@ -0,0 +1,138 @@
...
+  readonly finance: typeof finance;
+  readonly hacker: typeof hacker;
+  readonly internet: typeof internet;
+  readonly lorem: typeof lorem;
+  readonly name: typeof name;
+  readonly music: typeof music;
+  readonly phone_number: typeof phone_number;
+  readonly system: typeof system;
+  readonly vehicle: typeof vehicle;
+  readonly word: typeof word;
```
ST-DDT, line 111: Should we simplify these to just `[module: string]: string[]`?

ST-DDT:
All type definitions have been checked to work for both the en locale and the code using them (unless I didn't see all anys).
(With the expection of `postcode_by_state` that has a bug in both the code and the locales.)

src/definitions/address.ts
```diff
@@ -0,0 +1,114 @@
+import type { Format, Formats, Texts } from './utils';
+import { allOf } from './utils';
+
+/**
+ * The possible definitions related to addresses.
+ */
+export interface AddressDefinitions {
+  /**
+   * Postscodes patterns by state
```
ST-DDT, line 9:
```suggestion
   * Postcodes patterns by state
```

src/definitions/address.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  // TODO ST-DDT 2022-01-31: address.zipCodeByState() expects only { [state: string]: Range }
+  postcode_by_state: Formats | { [state: string]: Range };
+  /**
+   * Postscodes patterns
```
ST-DDT, line 14:
```suggestion
   * Postcodes patterns
```

ST-DDT:
Trivia: This PR reduces the lint warnings by more than 200 (731 -> 520).
(A huge amount of them are still scattered in the other locales and unused modules)

Shinigami92:
I will review this today in VSCode when I have a time-slot

src/index.ts
```diff
@@ -5,7 +5,12 @@ import { Company } from './company';
 import { Database } from './database';
 import { Datatype } from './datatype';
 import { _Date } from './date';
-import type { Definitions } from './definitions';
+import type {
+  Definitions,
+  DefinitionTypes,
+  LocaleDefinition,
+} from './definitions';
+import { definitions } from './definitions';
```
Shinigami92, line 13: Maybe we should use CONSTANT_CASE so we do not confuse with the Faker definitions.

```suggestion
import { definitions as DEFINITIONS } from './definitions';
```

src/index.ts
```diff
@@ -30,317 +35,26 @@ import { Word } from './word';
...
-      'interjection',
-      'noun',
-      'preposition',
-      'verb',
-    ],
-    title: '',
-    separator: '',
-  };
+  // This will eventually be replaced by just the imported definitions
+  private readonly definitionTypes: DefinitionTypes = definitions;
```
Shinigami92, line 57:
```suggestion
  private readonly definitionTypes: DefinitionTypes = DEFINITIONS;
```

src/locales/en/address/index.ts
```diff
@@ -20,28 +21,38 @@ import street_name from './street_name';
...
+} as Partial<AddressDefinitions>;
```
Shinigami92, line 56: Is it intended not to use `const address: Partial<AddressDefinitions> = ..-`?

_applies to other locations too_

src/definitions/company.ts
```diff
@@ -0,0 +1,33 @@
...
+import { allOf } from './utils';
+
+/**
+ * The possible definitions related to companies.
+ */
+export interface CompanyDefinitions {
+  // Business/products related words
+  bs_adjective: Texts;
+  bs_noun;
+  bs_verb;
```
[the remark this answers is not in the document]
ST-DDT: I missed them I guess. 😇

src/definitions/company.ts
```diff
@@ -0,0 +1,33 @@
+import type { Texts } from '.';
+import { allOf } from './utils';
+
+/**
+ * The possible definitions related to companies.
+ */
+export interface CompanyDefinitions {
+  // Business/products related words
```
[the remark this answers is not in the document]
ST-DDT: Will fix that

src/definitions/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  Texts,
+  Format,
+  Formats,
+  AnimalDefinitions,
+  AddressDefinitions,
+  CommerceDefinitions,
+  CommerceProductNameDefinitions,
+  CompanyDefinitions as CompanyDefinition,
+  DatabaseDefinitions as DatabaseDefinition,
+  DateDefinitions as DateDefinition,
```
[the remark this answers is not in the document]
ST-DDT: I renamed the files. Looks like vscode did something strange there.

src/index.ts
ST-DDT: Only here or also the exported values?

src/locales/en/address/index.ts
ST-DDT: Yes. If I add them to the end, the definitions contain values that are not specified in our API and we dont guarantee their existance or format at all. I only add them for backwards compatibility.

ST-DDT: We have to check/handle/support them in 6.2 or 7.0.

src/definitions/index.ts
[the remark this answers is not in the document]
Shinigami92: VSCode is doing it like this, so the renaming of a class doesn't break anything and the refactoring affects less then possible.

src/index.ts
Shinigami92: I think it makes sense also for the export

src/locales/en/address/index.ts
Shinigami92: I thought when we use `Partial` it will not complain

ST-DDT: `const x: Partial<X> =` is only possible if the object is a subset of X, but this isn't the case here.
As `x` contains elements that aren't in `X`.
The `as Partial<X>` allows for conjunction cases. Where some elements in `x` are not present in `X`.

src/definitions/finance.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * The possible definitions related to currencies.
+ */
+export interface FinanceCurrencyDefinitions {
+  /**
+   * The information about currencies by their full name.
+   */
+  [currencyName: string]: FinanceCurrencyEntryDefinitions;
+}
```
ST-DDT, line 36: Should I inline this as well?

src/definitions/system.ts
```diff
@@ -0,0 +1,37 @@
...
+  directoryPaths: string[];
+  mimeTypes: SystemMimeTypeDefinitions;
+}
+
+/**
+ * The mime type definitions with some additional information.
+ */
+export interface SystemMimeTypeDefinitions {
+  [mimeType: string]: SystemMimeTypeEntryDefinitions;
+}
```
ST-DDT, line 19: Should I inline this as well?

ST-DDT (commented):
I removed most of the simple types.

src/definitions/utils.ts
```diff
@@ -0,0 +1,8 @@
+// https://stackoverflow.com/a/53395649/4573065
+type AllOf<T> = ['Needs to be all of', T];
```
Shinigami92, line 2: Please also export this, so if someone really needs it, they can use it
```suggestion
export type AllOf<T> = ['Needs to be all of', T];
```

src/index.ts
```diff
@@ -30,317 +31,23 @@ import { Word } from './word';
...
-      'conjunction',
-      'interjection',
-      'noun',
-      'preposition',
-      'verb',
-    ],
-    title: '',
-    separator: '',
-  };
+  readonly definitions: LocaleDefinition = { title: 'loading' };
```
Shinigami92, line 50: loading? Theoretically this is a runtime change :thinking: 
I would suggest to just use 

```suggestion
  // Will be lazy init
  readonly definitions: LocaleDefinition = {} as LocaleDefinition;
```

src/index.ts
```diff
@@ -381,18 +88,17 @@ export class Faker {
...
    * Load the definitions contained in the locales file for the given types
-   *
-   * @param types
    */
-  private loadDefinitions(types: DefinitionTypes): void {
+  private loadDefinitions(): void {
     // TODO @Shinigami92 2022-01-11: Find a way to load this even more dynamically
     // In a way so that we don't accidentally miss a definition
-    Object.keys(types).forEach((t: keyof DefinitionTypes) => {
+    const types = DEFINITIONS;
```
Shinigami92, line 100: Maybe we should directly use it instead of re-variable it ^^

Shinigami92 (approved):

prisis (approved):

PR #362, opened 2022-01-29
feat: add more arabic names dataset

wael-Fadlallah:

PR #365, opened 2022-01-30
feat(revert): add more arabic names dataset

JessicaSachs:
Reverts faker-js/faker#362


Shinigami92 (approved):

ejcheng (approved):

ST-DDT (approved):

prisis (approved):

ST-DDT:
Does reverting this reduce our work?
If not, then we could just leave it in here as an accident.

Shinigami92:
> Does reverting this reduce our work?
> 
> If not, then we could just leave it in here as an accident.

Due to the CI fails and no other PR can be worked on / rebased, we will revert it for now
I talked already to the contributor and said that we are sorry. Everything is okay right now, we will do this in v6.2 as planned 👌

PR #372, opened 2022-01-30
test: rewrite name tests

Shinigami92:


ST-DDT (approved):

test/name.spec.ts
```diff
@@ -1,333 +1,520 @@
...
-        assertInArray(
-          name,
-          faker.definitions.name[genderString + '_first_name']
+  }
+
+  // Create and log-back the seed for debug purposes
+  faker.seed(Math.ceil(Math.random() * 1_000_000_000));
+
+  describe(`random seeded tests for seed ${faker.seedValue}`, () => {
+    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
```
piotrekn, line 169: @Shinigami92 what's the reason behind repeating tests `NON_SEEDED_BASED_RUN` times?

Shinigami92: Because these tests are random, we want to ensure that they are stable over multiple runs.
For most modules it's not such important so we just make currently `5` runs, but e.g. the datatype module we run `25` times.
Also if we need to, we have the same configuration for all tests and so we can quickly jump into a test and change these behalf.

PR #383, opened 2022-01-31
test: rewrite music tests

piotrekn:


test/music.spec.ts
```diff
@@ -1,14 +1,63 @@
...
+          expect(actual).toEqual(expectations[functionName].noArgs);
+        });
+      }
+    });
+  }
+
+  // Create and log-back the seed for debug purposes
+  faker.seed(Math.ceil(Math.random() * 1_000_000_000));
+
   describe('genre()', () => {
```
ST-DDT, line 55: AFAICT your don't use that seed here at all.

Maybe use it like this: https://github.com/faker-js/faker/blob/8bb1e6c068518c6c1ebafacc93d38cdbdb2710c5/test/date.spec.ts#L423-L428

ST-DDT (changes requested):
Looks good to me, expect one small detail.

ST-DDT (dismissed):

import-brain (approved):

Shinigami92 (changes requested):
Please add a `NON_SEEDED_BASED_RUN = 5` like in the other new tests, so we can ensure stability

ST-DDT (approved):

test/music.spec.ts
```diff
@@ -1,14 +1,69 @@
...
+  faker.seed(Math.ceil(Math.random() * 1_000_000_000));
+
+  describe(`random seeded tests for seed ${faker.seedValue}`, () => {
+    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
+      describe('genre()', () => {
+        it('should return a genre', () => {
+          const genre = faker.music.genre();
+
+          expect(typeof genre).toBe('string');
+          expect(genre).toBeTruthy();
```
Shinigami92, line 64: **suggestion:** I think this would be a bit better:
```suggestion
          expect(genre.length).greaterThen(0);
          expect(faker.definitions.music.genre).toContain(genre);
```

piotrekn: Checking `faker.definitions` - good idea 👍🏻 
Could you support with an argument, when `expect(genre.length).greaterThen(0` would be better than `expect(genre).toBeTruthy()`?
I see it exactly opposite. In case of `genre = null/undefined` you'd get generic error `Cannot read property 'length' of undefined`.

Shinigami92: Oh... yeah, you are right with the length :astonished: 
I was **wrong** with my thought that `expect('').toBeTruthy()` would not fail

ST-DDT (approved):

Shinigami92 (approved):

PR #384, opened 2022-02-01
chore: improve typesafety for locales

fpoppinga:
- DefinitionTypes now can only contain keys which are actually defined
  by the respective field in LocaleDefinition
- remove lots of  types from de locale
- fix (some) missing keys in DefinitionTypes


ST-DDT:
See also https://github.com/faker-js/faker/pull/363

Shinigami92:
Closing in favor of #363

PR #390, opened 2022-02-01
docs: update music jsdocs

ST-DDT:


Shinigami92 (approved):

import-brain (approved):

pkuczynski (approved):

src/music.ts
```diff
@@ -9,19 +12,13 @@ export class Music {
       }
       this[name] = this[name].bind(this);
     }
-
-    // TODO @Shinigami92 2022-01-12: We should find a better strategy as assigning this property to a function
-    // @ts-expect-error
-    this.genre.schema = {
-      description: 'Generates a genre.',
-      sampleResults: ['Rock', 'Metal', 'Pop'],
-    };
```
griest024, line 18: Why is this diff here? Seems like it doesn't belong

ST-DDT: We remove the schema in favor of plain jsdocs.

Shinigami92: The schema things were really old code that we want to get rid of
It was just for generating the old docs and examples
But because we now generate the docs out of the jsdocs we don't need these anymore

griest024: oh the schema is just for doc purposes? that is dumb haha

griest024 (dismissed):

griest024 (changes requested):
Please change your commit prefix to `docs:`

griest024 (approved):

Shinigami92 (approved):

PR #391, opened 2022-02-01
docs: update name jsdocs

ST-DDT:


Shinigami92 (approved):

src/name.ts
```diff
@@ -187,11 +214,13 @@ export class Name {
   }
 
   /**
-   * gender
+   * Return a random gender name.
```
griest024, line 217: It can just be `Return a random gender.`

src/name.ts
```diff
@@ -187,11 +214,13 @@ export class Name {
...
+   *
+   * @example
+   * faker.name.jobTitle() // 'Trans*Man'
+   * faker.name.jobTitle(true) // 'Female'
```
griest024, line 221: Example methods are wrong

griest024 (changes requested):

src/name.ts
ST-DDT: Fixed

src/name.ts
ST-DDT: Fixed

Shinigami92 (approved):

griest024 (approved):

PR #396, opened 2022-02-01
test: rewrite phone tests

piotrekn:


ST-DDT (approved):

test/phone.spec.ts
```diff
@@ -1,47 +1,125 @@
...
+          expect(phoneNumber).match(/\d/);
+          expect(fakeHelpers_replaceSymbolWithNumberSpy).toHaveBeenCalled();
+        });
+      });
+
+      describe('phoneNumberFormat()', () => {
+        it('should return phone number with proper US format (Array index)', () => {
+          faker.locale = 'en';
+          const phoneNumber = faker.phone.phoneNumberFormat(1);
+          expect(phoneNumber).match(/\([2-9]\d\d\) [2-9]\d\d-\d\d\d\d/);
```
Shinigami92, line 101: Please try to find something in `validator`

Shinigami92 (changes requested):

test/phone.spec.ts
ST-DDT: There is nothing in validators.
Phone numbers are just numbers with some "styling" to improve readability.

Shinigami92: :sob:

test/phone.spec.ts
```diff
@@ -1,47 +1,125 @@
...
+  }
+
+  // Create and log-back the seed for debug purposes
+  faker.seed(Math.ceil(Math.random() * 1_000_000_000));
+
+  describe(`random seeded tests for seed ${faker.seedValue}`, () => {
+    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
+      describe('phoneNumber()', () => {
+        it('should return a random phoneNumber with a random format', () => {
+          const fakeHelpers_replaceSymbolWithNumberSpy = vi.spyOn(
```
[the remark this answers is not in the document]
ST-DDT: Mhh, you use the spy to check the specific implementation of the method vs its api contract.
What if we refactor the method? The test might fail then, but the api contract is still fulfilled.
E.g. the method could return a "mock phone number prefix" like 555 and then an arbitrary number.

For me this isn't a blocker, but a cannot decide on this alone.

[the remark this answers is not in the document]
piotrekn: This single test kinda really more on implementation here than the API contract check, tbh. It is random number and random format, so API contract check is very limited.

I don't know the specifics why that was important to spy on in the first place. Maybe it was a way to be sure that creating a phone number of a format was done with this particular method. If it's an essential part of the implementation, it might be good to ensure its usage.

Not sure what to make of it. If you want it gone, let me know 👍🏻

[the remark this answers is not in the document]
Shinigami92: We currently try to remove all spies and mocks, so that we only test the "API" instead of implementation details.
So, yea, I would like to get this removed :slightly_smiling_face:

Shinigami92 (approved):

test/phone.spec.ts
```diff
@@ -1,47 +1,128 @@
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
Shinigami92, line 1: This should not be required anymore :thinking: 

```suggestion
import { beforeEach, describe, expect, it } from 'vitest';
```

Shinigami92 (changes requested):

Shinigami92 (approved):

ST-DDT (approved):

PR #397, opened 2022-02-01
chore: fix JSDoc comments in phone.ts

pkuczynski:
Refs #203


src/phone.ts
```diff
@@ -12,36 +12,46 @@ export class Phone {
   }
 
   /**
-   * phoneNumber
+   * Returns phone number.
```
ST-DDT, line 15:
```suggestion
   * Generates a random phone number.
```

src/phone.ts
```diff
@@ -12,36 +12,46 @@ export class Phone {
...
-    return this.faker.helpers.replaceSymbolWithNumber(format);
+    return this.faker.helpers.replaceSymbolWithNumber(
+      format || this.phoneFormats()
+    );
   }
 
-  // FIXME: this is strange passing in an array index.
   /**
-   * phoneNumberFormat
+   * Returns phone number in a format of the given index from `faker.definitions.phone_number.formats`.
```
ST-DDT, line 32:
```suggestion
   * Returns a phone number format of the given index from `faker.definitions.phone_number.formats`.
```

src/phone.ts
```diff
@@ -12,36 +12,46 @@ export class Phone {
...
+  // TODO @pkuczynski 2022-02-01: discuss removing this method as it tightly couples with localisation
+  phoneNumberFormat(phoneFormatsArrayIndex = 0): string {
     return this.faker.helpers.replaceSymbolWithNumber(
       this.faker.definitions.phone_number.formats[phoneFormatsArrayIndex]
     );
   }
 
   /**
-   * phoneFormats
+   * Returns phone number format.
```
ST-DDT, line 49:
```suggestion
   * Returns a random phone number format.
```

ST-DDT (changes requested):

src/phone.ts
pkuczynski: It does not return the format, but the phone number in this format...

ST-DDT: `Returns phone number in a format` sounds like it returns a phone number to me.
That's why I changed it to `Returns a phone number format `.

pkuczynski: Because it does return phone number :) Lookup the examples below @ST-DDT

ST-DDT: Oh you are right. I totally got confused by the method name.

ST-DDT (approved):

src/phone.ts
pkuczynski: Yeah, the method looks very odd, that's why I left a comment to consider dropping it in the future...

Shinigami92 (approved):

import-brain (approved):

Shinigami92 (approved):

PR #398, opened 2022-02-01
feat(types): provide strong typing for locales 2

ST-DDT:
Update `generateLocales.ts` to generate proper type definitions in `locales/.../index.ts` files.

The entirety of this PR was created by the included script with the exception of 5 files where I had to manually add `as X` (weekday, month, title).

@faker-js/contributors 

Currently the script has the following drawbacks/limitations:
- It is unable to detect that only optional properties are missing and thus it applies `Partial<X>` to often.
  - I could fix that by adding a `DEFINITIONS` variant, that only contains required properties. However, I haven't found a way to check that array for correctness at compile time
  - Use `LocaleDefinition['module']` as it is already wrapped in `Partial<X>`
  - Alternatively, we could mark all properties to be optional
  - Or ignore that for now

**Lint-Warnings:** 502 -> 205


src/locales/ar/vehicle/index.ts
```diff
@@ -1,12 +1,17 @@
...
-import module from './model';
+import model from './model';
 import type_ from './type';
 
-const vehicle: any = {
+const vehicle: Partial<VehicleDefinitions> = {
   fuel,
   manufacturer,
-  module,
+  model,
```
Shinigami92, line 14: Is this expected?

scripts/generateLocales.ts
```diff
@@ -16,36 +19,58 @@ const pathDocsApiLocalization = resolve(
...
+  const index = files.indexOf('index.ts');
+  if (index !== -1) {
+    files.splice(index, 1);
+  }
+  return files;
+}
 
-  // src/locale/<locale>.ts
+function removeTsSuffix(files: string[]): string[] {
+  return files.map((f) => f.replace('.ts', ''));
```
Shinigami92, line 43: Personally I don't like abbreviations, we should use `file`
_Applies to some other lines in this files, too_

src/locales/ar/vehicle/index.ts
ST-DDT: Seems to be a typo in the index, that got now regenerated.

Definitions:
https://github.com/faker-js/faker/blob/773bcec209da625cd1931d8bbb81e4d5159d9a89/src/definitions/vehicle.ts#L19-L22
Old:
https://github.com/faker-js/faker/blob/773bcec209da625cd1931d8bbb81e4d5159d9a89/src/locales/ar/vehicle/index.ts#L1-L13

scripts/generateLocales.ts
ST-DDT: Should be fixed. Please check whether I missed a part.

src/locales/ne/address/postcode.ts
```diff
@@ -1 +0,0 @@
-export default [0];
```
Shinigami92, line 1: :eyes:

Shinigami92 (dismissed):

src/locales/ne/address/postcode.ts
ST-DDT: This sole file content was invalid. I removed it to allow the fallback logic to take over and increase the visibility of a missing configuration by removing it from the file list.

Shinigami92 (approved):

Shinigami92 (approved):

prisis (approved):

PR #403, opened 2022-02-02
chore: enable eslint cache

pkuczynski:
This significantly speeds up subsequent eslint runs


ST-DDT (dismissed):

ST-DDT:
This works really nice. Where does it store the cache if I want to delete it?

pkuczynski:
> This works really nice. Where does it store the cache if I want to delete it?

Root of the project. Its already gitignored.

prisis (dismissed):

clarkerican (approved):

ST-DDT (approved):

Shinigami92:
Tested it locally and seems to work
Just asking myself why this is not the default behavior of eslint :thinking: 
Should we also clean the cache-file on `pnpm run clean`?

import-brain (approved):

package.json
```diff
@@ -66,7 +66,7 @@
     "docs:dev:run": "vitepress dev docs",
     "docs:serve": "vitepress serve docs",
     "format": "prettier --write .",
-    "lint": "eslint .",
+    "lint": "eslint --cache --cache-strategy content .",
```
Shinigami92, line 69: Let's do it this way, so users can decide to use it or not, also the CI and lint-stage will not get affected by it and always use a clean lint run

```suggestion
    "lint": "eslint .",
    "lint:cache": "eslint --cache --cache-strategy content .",
```

Shinigami92 (changes requested):

package.json
Shinigami92: https://github.com/eslint/eslint/issues/10712#issuecomment-409748549

pkuczynski: First of all the issue is 4y old and we are not using any of those plugins. Saying that, I am using `node` and `import` together with enabled cache and none of them ever caused any issues. I am not sure if anything changed since the linked issue was created and that's why I am not experiencing problems. But again we do not use those plugins and I would rather like to see a proof of malfunction than just make it very slow for everyone linting locally...

Shinigami92 (approved):
If we encounter any problems in the future, we can revert it

Shinigami92:
@pkuczynski uhm... sry, but
How will this behave when having a dependency-update? Or something similar.
So e.g. the eslint rule itself changed in behavior but the file did not change?

So maybe we should really add it to the `clean` statement :grimacing:

ST-DDT:
Is it based on the content of the file itself?
What if you change the api in your method and that causes a warning in another (untouched) file?
( Is that even possible? 🤔 )

Shinigami92:
> In the worse case scenario CI would caught the issue as it does not have the cache enabled

Okay :+1:

Shinigami92:
We await approval from @prisis (or a third maintainer)
So give it some minutes

prisis (approved):

PR #406, opened 2022-02-02
docs: update fake jsdocs

ST-DDT:
Ref:  #203

I noticed that the fake docs-page won't be updated by my script. I will fix that in another PR.


prisis (dismissed):

src/fake.ts
```diff
@@ -15,21 +15,30 @@ export class Fake {
...
+   * @param str The format string that will get interpolated.
    *
-   * @method faker.fake
-   * @param str
+   * @example
+   * faker.fake('{{name.lastName}}') // 'Barrows'
+   * faker.fake('{{name.lastName}}, {{name.firstName}} {{name.suffix}}') // 'Durgan, Noe MD'
+   * faker.fake('This is static test.') // 'This is static test.'
+   * faker.fake('Good Morning {{name.firstName}}!') // 'Good Morning Estelle!'
+   * faker.fake('You can call me at {{phone.phoneNumber(+!# !## #### #####!)}}.') // 'You can call me at +47 991 5679 973722.'
```
Shinigami92, line 40: Maybe we should use a 555 number here, to ensure it's a fake number: https://en.wikipedia.org/wiki/Fictitious_telephone_number#North_American_Numbering_Plan

ST-DDT: Only the example value or also the pattern?

Shinigami92: Only the example in the JSDocs

ST-DDT: Fixed

Shinigami92 (approved):

clarkerican (approved):

PR #407, opened 2022-02-03
chore: improve typings on random.objectElement

pkuczynski:
Refs #203

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -116,7 +116,6 @@ export class Random {
     return arrayCopy.slice(min);
   }
 
-  // TODO @Shinigami92 2022-01-28: This function needs types
   /**
    * Takes an object and returns a random key or value.
    *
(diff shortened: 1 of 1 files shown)


Shinigami92:
@pkuczynski could you try:

```ts
  objectElement<T extends Record<string, unknown>, K extends keyof T>(
    object: T,
    field: 'key'
  ): K;
  objectElement<T extends Record<string, unknown>, K extends keyof T>(
    object: T,
    field?: unknown
  ): T[K];
  objectElement<T extends Record<string, unknown>, K extends keyof T>(
    object = { foo: 'bar', too: 'car' } as unknown as T,
    field = 'value'
  ): K | T[K] {
    const key = this.faker.random.arrayElement(array);

    return field === 'key' ? (key as K) : (object[key] as T[K]);
  }
```

src/finance.ts
```diff
@@ -1,6 +1,7 @@
 import type { Faker } from '.';
 import type { Helpers } from './helpers';
 import ibanLib from './iban';
+import type { FinanceCurrencyEntryDefinitions } from './definitions';
```
Shinigami92, line 4: Is this a missing cleanup?

pkuczynski: yes! fixed

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox:
Wouldn't it be much easier from an architectural and types perspective, by splitting this method into 2 separate ones? One for object value and one for object key.

PR #408, opened 2022-02-03
chore: fix JSDoc comments in random.ts

pkuczynski:
Refs #203

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -14,6 +14,9 @@ function arrayRemove<T>(arr: T[], values: T[]): T[] {
   return arr;
 }
 
+/**
+ * Generates random values of different kinds. Some methods are deprecated and have been moved to dedicated modules.
+ */
 export class Random {
   constructor(private readonly faker: Faker, seed?: any[] | any) {
     // Use a user provided seed if it is an array or number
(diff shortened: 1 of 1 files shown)


src/random.ts
```diff
@@ -118,7 +153,7 @@ export class Random {
 
   // TODO @Shinigami92 2022-01-28: This function needs types
   /**
-   * Takes an object and returns a random key or value.
+   * Returns a random random key or value from given object.
```
Shinigami92, line 156: doubled `random`

prisis, line 156:
```suggestion
   * Returns a random key or value from given object.
```

ST-DDT (approved):

pkuczynski:
Guys, I haven't finished this one yet. Can only work on this next week as I am on holiday atm...

src/random.ts
```diff
@@ -216,25 +256,37 @@ export class Random {
...
       words.push(this.faker.random.word());
     }
+
     return words.join(' ');
   }
 
+  /**
+   * @see words()
+   */
   readonly randomWords: Random['words'] = this.words.bind(this);
```
Shinigami92, line 290: I think we can remove this, if it was never in `5.5.3`

Shinigami92 (changes requested):

ST-DDT:
If you (@pkuczynski ) are fine with it, we would like to merge these improvements in.
That way, we have at least some. Including the deprecation hints.
You can still do the remaining jsdocs when you are back from your holidays.

src/random.ts
Shinigami92: In worst case by accident / by me :laughing: 
The were stuff like

```ts
const words = function randomWords() { ...
```

And I tried to be smart and rebind it
But it could be that the `randomWords` was never exposed anyways and could only be available in a debugger.
We already found some other places like this and removed these also there

pkuczynski: We meant to deprecate this and remove in on of the future versions right? If it was in some old faker, we don't know if users would not be anyway upgrading from it straight to 6.x

Shinigami92: It is nowhere exposed in the old types: https://github.com/Ivan-Kuzmichev/DefinitelyTyped/blob/d04089192418ea5954df5abb0093f9e5c7e115a5/types/faker/index.d.ts
CTRL+F `randomWords` = no hits

Were as https://github.com/Ivan-Kuzmichev/DefinitelyTyped/blob/d04089192418ea5954df5abb0093f9e5c7e115a5/types/faker/index.d.ts#L229 is exposed

Shinigami92, line 290:
```suggestion
```

Shinigami92 (changes requested):

ST-DDT (approved):

Shinigami92 (approved):

PR #420, opened 2022-02-04
docs(fix): use named import

togami2864:
If we import `faker` by default import as in the sample, we will get the following error.
```js
import faker from '@faker-js/faker';

const randomName = faker.name.findName(); // Rowan Nikolaus
// when using default import , we need to write like `faker.faker.name.findName()`
```

```shell
// result
const randomName = faker.name.findName();
                              ^
TypeError: Cannot read property 'findName' of undefined
```
This can be avoided by using named import.

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -42,7 +42,7 @@ pnpm add @faker-js/faker --save-dev
 ### Node.js
 
 ```js
-import faker from '@faker-js/faker';
+import { faker } from '@faker-js/faker';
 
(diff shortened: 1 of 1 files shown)


Shinigami92:
Duplicate of https://github.com/faker-js/faker/pull/410

Shinigami92 (approved):

Shinigami92:
But we can merge this now, cause now I'm able to give my approval :wink:

ST-DDT (approved):

PR #431, opened 2022-02-05
fix: optional args on faker.finance.iban()

dantman:
`faker.finance.iban()` is a valid use of Faker. However in TypeScript you get a type error because the optional `formatted` and `countryCode` args are not marked as optional like in other functions.

Tests are written against the cjs bundle without types so did not catch the error.

### DIFF
--- a/src/finance.ts
+++ b/src/finance.ts
@@ -293,7 +293,7 @@ export class Finance {
    *
    * @method faker.finance.iban
    */
-  iban(formatted: boolean = false, countryCode: string): string {
+  iban(formatted: boolean = false, countryCode?: string): string {
     let ibanFormat: {
       bban: Array<{ type: string; count: number }>;
       country: string;


ST-DDT:
FFR: In definitely typed it was [supported](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58218/files#diff-22605747c70de0fe07b686b06a1235f39c32faf85b32d66fc18c9623379d3f50L140) and the code (both impl and test) looks like it is optional as well. So we should restore the original behavior.

If you are familiar with ts, could you please also remove the ts-expect-error from the `iban()` calls in the finance_iban.spec.ts test file?

Thanks for bringing this to our attention.

ST-DDT (approved):

ST-DDT:
I have to check why the code cov is reported as zero for this patch. The tests should be covering this.

dantman:
> I have to check why the code cov is reported as zero for this patch. The tests should be covering this.

Maybe because that's a diff, I only made a one character type change, and none of the tests cover types.

Shinigami92 (approved):

PR #432, opened 2022-02-05
test: cover source instead of bundled code

Shinigami92:


Shinigami92:
Please note that I explicitly left `test/locale-imports.spec.ts` as is, so we currently still need to **build** the code before running the tests

Shinigami92:
Currently trying to find a way to prevent the replacement of `||=` just to support Node 14.
The bundled code should already cover that.

Shinigami92:
For now we rewrote it, we can find a way later in another PR if needed

ST-DDT (approved):

ST-DDT (approved):

clarkerican (approved):

PR #434, opened 2022-02-05
docs: improve finance jsdocs

ST-DDT:
Ref: #203

### DIFF
--- a/src/definitions/finance.ts
+++ b/src/definitions/finance.ts
@@ -9,7 +9,7 @@ export interface FinanceDefinitions {
    */
   account_type: string[];
   /**
-   * The pattern by provider used to generate credit card codes.
+   * The pattern by (lowercase) provider name used to generate credit card codes.
    * `L` will be replaced by the check bit.
    *
    * @see Helpers.replaceCreditCardSymbols()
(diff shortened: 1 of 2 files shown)


Shinigami92 (approved):

Shinigami92:
https://github.com/bcoe/c8/issues/182

@bcoe @j03m
/cc @antfu 

Did we misc-configured something in our `vite.config.ts`

https://github.com/faker-js/faker/blob/832bf8a2ecaeb83458581212c1c9467575aa3488/vite.config.ts#L1-L14

Or is it really a bug by `c8` that JSDoc multi-line comments are count as _needs to be covered by tests_? :eyes: 
If so, this could be a huge bummer for `vitest` :grimacing:

bcoe:
@Shinigami92 I would take a look at the `.js` generated from the `.ts` file, sometimes there's an unexpected code path that gets generated, by virtue of the structure of the generated code.

Shinigami92:
@bcoe So we already want to switch to just cover the `src` folder instead of the `dist/cjs` (https://github.com/faker-js/faker/pull/432)
Do you think this could already fix it?

bcoe:
> Do you think this could already fix it?

There's a chance. The problem is generally a disconnect between the `.ts` and the `.js`, if the typescript compiler introduces an extra logical expression that's not exercised, or the source map remapping has a hiccup, there might be an uncovered element.

clarkerican (approved):

Issue #352, opened 2022-01-29
Use enum for male/female references


griest024:
### Clear and concise description of the problem

References to male and female are done with hardcoded strings and numbers. https://github.com/faker-js/faker/blob/44628ecbbf8967c3a1313c6abfc62ad653cafe29/src/name.ts#L29-L35.

### Suggested solution

These should be replaced with enum values:

```ts
enum Sex {
  MALE = 'male',
  FEMALE = 'female',
  INTERSEX = 'intersex'
}
```

### Additional context

As an aside, `gender` is an inaccurate name for male and female. The correct term is `sex`.

Shinigami92:
I would suggest to use type literals like `type Sex = 'male' | 'female'`

Beside that, I would highly like to have something like `type Sex = 'male' | 'female' | 'binary'` in the future

griest024:
> I would suggest to use type literals like `type Sex = 'male' | 'female'`
> 
> Beside that, I would highly like to have something like `type Sex = 'male' | 'female' | 'binary'` in the future

I much prefer `enum`s; simple types can't be used as runtime values and we can't use it to replace hardcoded values.

Shinigami92:
I like to provide enums, but not to force the usage
So we could to 

```ts
export enum Sex {
  MALE = 'male',
  FEMALE = 'female',
  INTERSEX = 'intersex'
}

export type SexType = 'male' | 'female' | 'intersex' | number

firstName(gender?: SexType): string { /*...*/ }
```

griest024:
Yes as to the external usage we should accept more than just the enum. But for the internal usage I want to see `case Sex.MALE:` instead of `case 0:`.

ST-DDT:
Please no numbers. Or what is your intention?

Shinigami92:
[TS Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBMAdgVwLZwMqjgbwLABQccAsgIIAyAonALxwDkqAhgDbAMA0hxAYleWp1GAM2At2XHnACSAOQAqVAEoYqADWEMAlohjAoAZ1ANCAX0KFQkWHBgBPMMEygFj5-QAGAEhxYQZp5wAD5wKKgARgaWBADGEIiG8CAAXC4gbk5aEhwA3DEiyIixMNoJcCLaRjByzKjAABQA5kgAJgYA-Gn+mcAAlGlJULpNuHBQwDDIUIiMDHBmQA)

[...]

PR #440, opened 2022-02-05
chore(locale): added curated names for default(en) locale

budaG:
@Shinigami92 review..


test/name.spec.ts
```diff
@@ -245,6 +245,17 @@ describe('name', () => {
...
+          faker.locale = 'en';
+
+          let name = faker.name.middleName();
+          expect(faker.definitions.name.middle_name).toContain(name);
+          name = faker.name.middleName(0);
+          expect(faker.definitions.name.male_middle_name).toContain(name);
+          name = faker.name.middleName(1);
+          expect(faker.definitions.name.female_middle_name).toContain(name);
+        });
+
```
ST-DDT, line 258: A few lines below this is a test that you can enable by removing the todo from it.

ST-DDT (commented):
Could you please add the source of these lists to this PR?

src/name.ts
```diff
@@ -126,8 +126,23 @@ export class Name {
...
       if (typeof gender !== 'number') {
-        gender = this.faker.datatype.number(1);
+        if (typeof this.faker.definitions.name.middle_name === 'undefined') {
+          gender = this.faker.datatype.number(1);
+        } else {
+          // Fall back to unisex middle names if they exist and gender wasn't specified
+          return this.faker.random.arrayElement(
+            this.faker.definitions.name.middle_name
+          );
+        }
```
Shinigami92, line 145: I think this code occurs multiple times in the name module, we should move it into a private helper method

Shinigami92 (changes requested):

test/name.spec.ts
budaG: @ST-DDT good catch, enabled the test case for 'uk' locale. 
https://docs.google.com/spreadsheets/d/1I2GBSy7Y63hNYYtYr_hYbhk0bCI3phW5WDnEBlHCsD8/edit?usp=sharing
here is my worksheet with source URL, I used to scrape those names.

src/name.ts
budaG: @Shinigami92 I agree. Isolated the duplicated code into a private method. Let me know if it looks good.

src/name.ts
```diff
@@ -332,4 +332,26 @@ export class Name {
...
+   * Determines gender in 0 or 1
+   *
+   * @param gender The optional gender to use.
+   *
+   * @returns 0 or 1 given a string param of 'male' or 'female' else returns default value
+   *
+   * @example Name.determineGenderByStringParam('male') // 0
+   *
+   */
+  private static determineGenderByStringParam(gender: string | number) {
```
ST-DDT, line 347: IMO this method should do most of the gender related stuff.
String handling, unknown number handling, undefined handling (choose random).
This method should also only ever return `0 | 1´.

ST-DDT (changes requested):

test/name.spec.ts
budaG: Just realized about the permission. Changed the permission so that everyone can see. Let me know if you see it otherwise.
Also that list has been sanitized for duplicates, trimmed and ordered before it made it to the code base.

src/name.ts
budaG: I see what you are saying and I agree we should have methods to handle all that separately rather than handling those scenarios in every method but that touches every method. I think it should be a separate story/enhancement/chore to do that. This was to fix the default locale middle name bug.

ST-DDT: There are three ways:
- Just fix the middle name method without adding a new helper method.
- Fix the middle name method by creating a separate method, but don't use it in the other methods yet.
- Fix all name methods by creating a separate method and using it everywhere.

I suggest going for 1 or 2 and then creating a separate PR with option 3 later once this is merged.

Shinigami92: In the longtime goal, I would really like to **deprecate** numeric values for **gender**.
Why is `male = 0` and `female = 1`? What's with binary genders?
The string values would not have a numeric order.

@griest024 also wanted to introduce the possibility to additionally use enum. #352

---

I would currently like to see just a switch case with default fallback.
But I'm not sure if this becomes to much for this PR and we may want to split it to another PR.
On the other side, as we already targeting 6.1 with this PR, maybe it's okay...

---

I would write it like so that we can call it this way:
```ts
export type Gender = ...;

firstName(gender?: Gender): string
  return this.<functionName>(gender, this.faker.definitions.name.<arrayToUse>);
}
```

budaG: I like the idea of Enum, but if we are choosing between lesser of the two evils, switch vs if, for now, I would stick with if, more concise and clear. Let me know your thoughts.

JessicaSachs: Non-technical nit: can we, like, literally remove the "gender binary" from our codebase (at some point)? I think we can still have the API, but I'd like to make an explicit effort to revamp everything surrounding gender in the lib. I feel pretty strongly about this.

JessicaSachs: > Why is male = 0 and female = 1? What's with binary genders?

lol I love you Shini.

Shinigami92:
@budaG This PR needs a rebase

Shinigami92 (approved):

PR #442, opened 2022-02-06
docs: improve helpers jsdocs

ST-DDT:
Ref: #203


src/helpers.ts
```diff
@@ -413,13 +477,18 @@ export class Helpers {
     return str;
   }
 
-  // TODO @Shinigami92 2022-01-11: We might have a bug with the `phone` definition
-  // I may be `phone_number` instead
```
ST-DDT, line 417: @Shinigami92 What does this comment refer to?

Shinigami92: I think it was just my confusion about that the module was named `phone` and the definition was named `phone_number` or something like that.
And I thought it would be dynamically resolved, because I assumes the code would be smart...
So I think you can ignore this comment

src/helpers.ts
```diff
@@ -110,22 +122,33 @@ export class Helpers {
   }
 
   /**
-   * backward-compatibility
+   * Backward-compatibility. Use `faker.random.arrayElement()` instead.
```
Shinigami92, line 125: We should deprecated this in an extra PR

prisis (approved):
nice :), great job

PR #446, opened 2022-02-07
docs: use vue components for api-docs

ST-DDT:
Refactor the docs generation to be easier to understand by using vue components and splitting the code into smaller methods.


docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -0,0 +1,32 @@
...
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Type</th>
+          <th>Default</th>
+          <th>Description</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="parameter of props.parameters" v-bind:key="parameter">
```
Shinigami92, line 21:
```suggestion
        <tr v-for="parameter of props.parameters" :key="parameter">
```

Shinigami92: I assume `parameter` is not of type `string`. We should take a unique value as `:key`
I think we can use `:key="parameter.name"` here, cause `name` should always be a unique key

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -0,0 +1,32 @@
...
+          <td>{{ parameter.name }}</td>
+          <td>{{ parameter.type }}</td>
+          <td>
+            <code v-if="parameter.default">{{ parameter.default }}</code>
+          </td>
+          <td v-html="parameter.description"></td>
```
Shinigami92, line 27: @JessicaSachs I would like to hear your points about the usage of `v-html`
As far as I know it is mostly a security issue, is it?

scripts/apidoc.ts
```diff
@@ -96,120 +89,77 @@ async function build(): Promise<void> {
...
+      for (
+        let index = 0;
+        signature.parameters && index < signature.parameters.length;
+        index++
+      ) {
+        const parameter = signature.parameters[index];
+
+        const parameterDefault = parameter.defaultValue;
+        const parameterRequired = typeof parameterDefault === 'undefined';
+        if (index == 0) {
```
Shinigami92, line 130:
```suggestion
        if (index === 0) {
```

Shinigami92 (changes requested):

docs/.vitepress/components/api-docs/method-parameters.vue
ST-DDT: If requested I can also apply HTML sanitize before setting it.

Shinigami92 (dismissed):
If this is actually working and serve benefits, I will not block it just because of the usage of `v-html`
We should just keep it in mind, and if we know better we could fix it or workaround it in the future :shrug:

docs/.vitepress/components/api-docs/method-parameters.vue
ST-DDT: I now sanitize the api docs before writing the files. That should be safe enough.

ST-DDT:
The changes to `random.ts` are required, because it does get treated as html attributes.

````txt
Rejected unsafe md: Defaults to { bannedChars: [] }
Rejected unsafe html: <p bannedChars:="" []="">Defaults to</p>

Expected safe html: <p>Defaults to</p>

````

**Comment by ST-DDT** [comment]
@Shinigami92 VSCode reports `string.replaceAll` to not exist in `apidoc.ts`. It runs fine though. Any idea why?

**Comment by Shinigami92** [comment]
> @Shinigami92 VSCode reports `string.replaceAll` to not exist in `apidoc.ts`. It runs fine though. Any idea why?

replaceAll is a really new js feature and was added last year, maybe not all node version support it yet
Use replace(/.../g, ...) instead

File: scripts/apidoc.ts [hunk 4]
```diff
@@ -99,120 +118,77 @@ async function build(): Promise<void> {
...
-        ::: v-pre
-
-        ${toBlock(signature.comment)}
-
-        :::
-
-        `.replace(/\n +/g, '\n');
-
-      // Generate parameter section
+      // typeParameters
```
Shinigami92, line 136: If this is a section comment, maybe use not camelCase here but Title Case or just a sentence

scripts/apidoc.ts
```diff
@@ -99,120 +118,77 @@ async function build(): Promise<void> {
...
-        // parameters
-        parameters.forEach((parameter, index) => {
-          const parameterDefault = parameter.defaultValue;
-          const parameterRequired = typeof parameterDefault === 'undefined';
-          if (index == 0) {
-            requiresArgs = parameterRequired;
-          }
-          const parameterName = parameter.name + (parameterRequired ? '?' : '');
-          const parameterType = parameter.type.toString();
+      // parameters
```
Shinigami92, line 147: Same here

scripts/apidoc.ts
```diff
@@ -221,34 +197,72 @@ async function build(): Promise<void> {
...
       }
-      const examples =
+      const exampleTags =
         signature?.comment?.tags
           .filter((tag) => tag.tagName === 'example')
           .map((tag) => tag.text.trimEnd()) || [];
 
-      if (examples.length !== 0) {
-        content += examples.join('\n').trim() + '\n';
+      if (exampleTags.length !== 0) {
```
Shinigami92, line 211: Should we use `>` for better readability?

scripts/apidoc.ts
```diff
@@ -1,9 +1,17 @@
...
 import { format } from 'prettier';
-import options from '../.prettierrc.cjs';
+import * as TypeDoc from 'typedoc';
+import { createMarkdownRenderer } from 'vitepress';
+import prettierConfig from '../.prettierrc.cjs';
+import type {
+  Method,
+  MethodParameter,
+} from '../docs/.vitepress/components/api-docs/method';
+// import vitepressConfig from '../docs/.vitepress/config';
```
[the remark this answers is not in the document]
ST-DDT: At some point I would like to actually import that config. Sadly, I cannot do so at this point in time.
Should I add a TODO here?

Shinigami92 (dismissed):

prisis (approved):

prisis (approved):

Shinigami92 (approved):

PR #455, opened 2022-02-08
fix: usable esm exports

### LINKED ISSUE #448: Feat: Instantiate standalone Faker instance, import locales (or extend for custom subclass)
Goal:

Create a subclass that extends the `Faker` class so proprietary fake-data generators can be used side by side with faker's built-in ones.

Problem:

There seems to be a problem with the bound methods getting bound to `undefined` instead of to the instance. This is likely due to the empty `locales` option:

- I didn't see any way to import all locales like Faker's own code does, or even an individual locale (see code example for my attempts)
- The `Faker` constructor initializes `locale` to `en`, but seems to use an empty object as the default for `locales`. 

Proposed solution(s):

- Set the default `locales` in `Faker` constructor to at least include the `en` locale so the class can be used via `new Faker()`
- Throw an error if `new Faker(...)` is instantiated with invalid options
- Fix whatever is preventing the locales from being imported, so users can instantiate their own `Faker` instances the "long" way.

Demo and other commentary here:

https://stackblitz.com/edit/faker-js-demo-q1dwn6?devtoolsheight=33&file=index.ts

The code is also below, in case that demo gets corrupted:

```typescript
import { Faker } from '@faker-js/faker';
// I would expect at least one these work:
// import allLocales from '@faker-js/faker/locales';
// import { locales } from '@faker-js/faker/locales';
// import en from '@faker-js/faker/locales/en';
// import { en } from '@faker-js/faker/locales/en';

// This code shows how the bind is failing, resulting in
// "Error: Cannot read properties of undefined (reading 'internet')"
const baseFaker = new Faker(/* {locales: {en}} */);
console.log('baseFaker.internet.email: ', baseFaker.internet.email);
console.log(baseFaker.internet.email());

// Below here is the code I'm actually trying to write.
// I don't think it's relevant to the bug, but I've left
// it here as an example of the end goal and a possible
// solution.

export class MyFake extends Faker {
  readonly customFake = new CustomFake(this);
}

export class CustomFake {
  private readonly fake: MyFake;

  constructor(fake: MyFake) {
    this.fake = fake;
  }

  /* Side note: this syntax seems a lot easier than the
   * bind() methods faker currently uses. */
  public someFakeThing = () => {
    return 'fake data';
  };
}

const myfaker = new MyFake(/* {locales: {en}} */);

// confirm this works
console.log(myfaker.customFake.someFakeThing());

// some debug logs to show values "exist"
// console.log("myfaker: ", myfaker);
// console.log("myfaker.internet: ", myfaker.internet);
console.log('myfaker.internet.email: ', myfaker.internet.email);

// but this fails, presumably because email() is not bound
console.log(myfaker.internet.email());
```

Shinigami92:
closes #448

This is a critical fix and should be done in v6.0 cause users want to explicitly import e.g. locales


Shinigami92:
I tried several things but nothing works right now and I do not fully understand the `exports` field in `package.json`
I will break up now and close this PR as failed

If someone else wants to tackle this, please feel free

PR #457, opened 2022-02-09
fix: fix unique method types

### LINKED ISSUE #456: `faker.unique` TS error
### Describe the bug

when using `faker.unique` with some other faker functions that take arguments, Typescript will throw an error

tried with:
* `faker.address.countryCode`
* `faker.datatype.number`
* `faker.random.arrayElement`
* `faker.internet.email`

### Reproduction

```javascript

const code =  faker.unique(faker.address.countryCode, []);
const code2 =  faker.unique(faker.address.countryCode);
// const code3 =  faker.unique(faker.address.countryCode, 'US');
const code4 =  faker.unique(faker.address.countryCode, ['US']);
```
all the above throw TS errors but result seem to be ok except for `code3`.

### Additional Info

_No response_

ST-DDT:
Fixes #456 

**Note:** There is a slight imprecision I couldn't remove.
It is possible to omit the parameters, although the referenced method requires them.
If I fix that, it is no longer possible to omit the empty parameter array.

### DIFF
--- a/src/unique.ts
+++ b/src/unique.ts
@@ -1,4 +1,5 @@
 import * as uniqueExec from './vendor/unique';
+import type { RecordKey } from './vendor/unique';
 
 /**
  * Module to generate unique entries.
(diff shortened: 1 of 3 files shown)


ST-DDT:
No, i didn't see it (or forgot about it).
This variant also covers non string returning methods.
(But I'm fine, with closing this one, if the other ones get updated and supports non string methods.)

prisis (approved):

PR #469, opened 2022-02-11
test: rewrite vehicle

Shinigami92:


ST-DDT (changes requested):

src/locales/en/vehicle/index.ts
```diff
@@ -6,7 +6,7 @@ import model from './model';
 import type_ from './type';
 
 const vehicle = {
-  bicycle,
+  bicycle_type: bicycle,
```
ST-DDT, line 9: Do you need this change for your tests?

Either revert this change or also rename the actual files, because this file will be generated soon.

ST-DDT (changes requested):

prisis (approved):

PR #483, opened 2022-02-13
types(date): allow `Date` obj as argument

fratzinger:
All methods in `faker.date` only allowed the type `string` but tests were also run with `Date` arguments. So passing a `Date` was possible but resulted in a TS error. This PR allows `Date` as argument for all relevant `faker.date` methods.


fratzinger:
saw #200 too late. Closing this in favor of #200.

PR #488, opened 2022-02-14
fix: remove invalid aliases

Shinigami92:
While rewriting to TS I added accidentally some aliases.
These are not needed


ST-DDT:
> These are not needed

= Not in 5.5.3

import-brain (approved):

prisis (approved):

PR #490, opened 2022-02-14
test: rewrite random

Shinigami92:


ST-DDT (approved):
Where are the seeded tests? This test class looks a lot different from the other ones.
Should we really structure this class differently?

Shinigami92:
> Where are the seeded tests? This test class looks a lot different from the other ones. Should we really structure this class differently?

Maybe if I have time and will, I can add seeded tests in next few days.
Or in another PR, so at least we can quick-merge this, so we benefit already from the improved variant.

prisis (approved):

prisis:
Im fine if we add the seed later :)

Issue #492, opened 2022-02-15
Add separate methods for random key and value


ST-DDT:
Wouldn't it be much easier from an architectural and types perspective, by splitting this method into 2 separate ones? One for object value and one for object key.

_Originally posted by @xDivisionByZerox in https://github.com/faker-js/faker/issues/407#issuecomment-1040411965_

xDivisionByZerox:
Since I, more or less, requested this feature, should I provide help by implementing this change and providing a pull request?

ST-DDT:
Contributions are always welcome. We might not be be able to merge it soon though.

import-brain:
@xDivisionByZerox Assigning to you :)

xDivisionByZerox:
I just opened #503

PR #503, opened 2022-02-17
feat: separate methods for object key value

xDivisionByZerox:
Created in relation to #492.

Thinks I have done:
- Create and implement 2 new methods `faker.random.objectValue` and `faker.random.objectKey`
- Wrote documentation for both methods
- Wrote test for both methods
- Added a deprecation flag and log for `faker.random.objectElement`

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -168,7 +168,7 @@ export class Random {
    * @param object The object to get the keys from.
    * @param field If this is set to `'key'`, this method will a return a random key of the given instance.
    *
-   * @see faker.random.objectKey()
+   * @see faker.helpers.objectKey()
    *
    * @example
    * const object = { keyA: 'valueA', keyB: 42 };
@@ -188,7 +188,7 @@ export class Random {
    * @param object The object to get the values from.
    * @param field If this is set to `'value'`, this method will a return a random value of the given instance.
    *
-   * @see faker.random.objectValue()
+   * @see faker.helpers.objectValue()
    *
    * @example
    * const object = { keyA: 'valueA', keyB: 42 };
@@ -211,8 +211,8 @@ export class Random {
    * If this is set to `'value'`, this method will a return a random value of the given instance.
    * Defaults to `'value'`.
    *
-   * @see faker.random.objectKey()
-   * @see faker.random.objectValue()
+   * @see faker.helpers.objectKey()
+   * @see faker.helpers.objectValue()
    *
    * @example
    * const object = { keyA: 'valueA', keyB: 42 };
@@ -236,8 +236,8 @@ export class Random {
    * If this is set to `'value'`, this method will a return a random value of the given instance.
    * Defaults to `'value'`.
    *
-   * @see faker.random.objectKey()
-   * @see faker.random.objectValue()
+   * @see faker.helpers.objectKey()
+   * @see faker.helpers.objectValue()
    *
    * @example
    * const object = { keyA: 'valueA', keyB: 42 };
--- a/src/random.ts
+++ b/src/random.ts
@@ -161,23 +161,18 @@ export class Random {
   }
 
   /**
-   * Returns a random key or value from given object.
+   * Returns a random key from given object.
    *
    * @template T The type of `Record` to pick from.
    * @template K The keys of `T`.
-   * @param object The object to get the keys or values from.
+   * @param object The object to get the keys from.
    * @param field If this is set to `'key'`, this method will a return a random key of the given instance.
-   * If this is set to `'value'`, this method will a return a random value of the given instance.
-   * Defaults to `'value'`.
    *
    * @see faker.random.objectKey()
-   * @see faker.random.objectValue()
    *
    * @example
    * const object = { keyA: 'valueA', keyB: 42 };
-   * faker.random.objectElement(object) // 42
    * faker.random.objectElement(object, 'key') // 'keyB'
-   * faker.random.objectElement(object, 'value') // 'valueA'
    *
    * @deprecated
    */
@@ -185,6 +180,23 @@ export class Random {
     object: T,
     field: 'key'
   ): K;
+  /**
+   * Returns a random value from given object.
+   *
+   * @template T The type of `Record` to pick from.
+   * @template K The keys of `T`.
+   * @param object The object to get the values from.
+   * @param field If this is set to `'value'`, this method will a return a random value of the given instance.
+   *
+   * @see faker.random.objectValue()
+   *
+   * @example
+   * const object = { keyA: 'valueA', keyB: 42 };
+   * faker.random.objectElement(object) // 42
+   * faker.random.objectElement(object, 'value') // 'valueA'
+   *
+   * @deprecated
+   */
   objectElement<T extends Record<string, unknown>, K extends keyof T>(
     object: T,
     field?: unknown
@@ -212,12 +224,8 @@ export class Random {
    * @deprecated
    */
   objectElement<T extends Record<string, unknown>, K extends keyof T>(
-    object: T,
-    field?: 'key' | 'value'
-  ): K | T[K];
-  objectElement<T extends Record<string, unknown>, K extends keyof T>(
-    object = { foo: 'bar', too: 'car' } as unknown as T,
-    field = 'value'
+    object: T = { foo: 'bar', too: 'car' } as unknown as T,
+    field: 'key' | 'value' = 'value'
   ): K | T[K] {
     deprecated({
       deprecated: 'faker.random.objectElement()',
--- a/src/random.ts
+++ b/src/random.ts
@@ -227,9 +227,10 @@ export class Random {
     object: T = { foo: 'bar', too: 'car' } as unknown as T,
     field: 'key' | 'value' = 'value'
   ): K | T[K] {
+    const useKey = field === 'key';
     deprecated({
-      deprecated: 'faker.random.objectElement()',
-      proposed: 'faker.random.objectKey() or faker.random.objectValue()',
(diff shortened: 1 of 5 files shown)


ST-DDT (approved):

src/random.ts
```diff
@@ -170,6 +170,8 @@ export class Random {
    * faker.random.objectElement(object) // 42
    * faker.random.objectElement(object, 'key') // 'keyB'
    * faker.random.objectElement(object, 'value') // 'valueA'
+   *
+   * @deprecated
```
ST-DDT, line 174: Please add `@see` references to the new methods.

Shinigami92: Do we need to add this annotation and the see to both JSDocs? I think so.

xDivisionByZerox: I'm more concerned why there are 2 JSDocs in the first place. Is this some kind of merge problem?

Shinigami92: Was added here: https://github.com/faker-js/faker/pull/596
Sadly I think I discussed with @ST-DDT in voice chat without writing a comment about the why
It was something like the one is shown in VSCode and the other is picked up from TypeDoc

xDivisionByZerox: Added the `@see` decorator to both JSDocs in 30711140ee4a7380853e7b4c91a0170aacfa694e

src/random.ts
```diff
@@ -189,6 +194,9 @@ export class Random {
    * If this is set to `'value'`, this method will a return a random value of the given instance.
    * Defaults to `'value'`.
    *
+   * @see faker.random.objectKey()
+   * @see faker.random.objectValue()
+   *
```
ST-DDT, line 199: This one is missing the deprecated flag.

ST-DDT: Please also add a `// For typedoc/api docs` comment above this jsdocs.
That way it's easy to tell why it is here.

xDivisionByZerox: Should be done now

ST-DDT (dismissed):

Shinigami92 (dismissed):

src/random.ts
```diff
@@ -203,10 +214,48 @@ export class Random {
     object = { foo: 'bar', too: 'car' } as unknown as T,
     field = 'value'
   ): K | T[K] {
+    console.warn(
+      'Deprecation Warning: faker.random.objectElement is now split into 2 seperate methods faker.datatype.objectKey and faker.datatype.objectValue'
```
import-brain, line 218:
```suggestion
      'Deprecation Warning: faker.random.objectElement() is now split into 2 separate methods: faker.datatype.objectKey() and faker.datatype.objectValue().'
```

import-brain: Very small change, but I think this makes the warning slightly easier to read.

xDivisionByZerox:
The failing CI doesn't seem to be related to the changes I made, rather than with the `fake` module.
```
2022-03-26T09:57:08.4524974Z  > Fake.fake src/fake.ts:85:12
2022-03-26T09:57:08.4525598Z
2022-03-26T09:57:08.4526162Z     if (this.faker[parts[0]] == null) {
2022-03-26T09:57:08.4527269Z           throw new Error('Invalid module: ' + parts[0]);
2022-03-26T09:57:08.4527942Z                             ^
2022-03-26T09:57:08.4528536Z     }

Error: invalid module: {datatype
```

I guess a rebase could fix this. (?)

Shinigami92 (changes requested):
1. We need to add test for the deprecation warning
2. We need to swap out all internal usages like in `finance.currencyCode` (please use search to find more occurrences)

Shinigami92:
> The failing CI doesn't seem to be related to the changes I made, rather than with the `fake` module.
> 
> ```
> 2022-03-26T09:57:08.4524974Z  > Fake.fake src/fake.ts:85:12
> 2022-03-26T09:57:08.4525598Z
> 2022-03-26T09:57:08.4526162Z     if (this.faker[parts[0]] == null) {
> 2022-03-26T09:57:08.4527269Z           throw new Error('Invalid module: ' + parts[0]);
> 2022-03-26T09:57:08.4527942Z                             ^
> 2022-03-26T09:57:08.4528536Z     }
> 
> Error: invalid module: {datatype
> ```
> 
> I guess a rebase could fix this. (?)

I'm not totally sure what's going on here, I hope @ST-DDT has some more in-depth inside into the fake function and what could be the problem here related to this PR changes
`main` branch is green and I did not see this previously somewhere

Here is a local failed test result

```js
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/all_functional.spec.ts > faker.fake functional tests > uk > datatype > array()
Error: Invalid module: {datatype
 ❯ Fake.fake src/fake.ts:85:12
     83| 
     84|     if (this.faker[parts[0]] == null) {
     85|       throw new Error('Invalid module: ' + parts[0]);
       |            ^
     86|     }
     87| 
 ❯ Fake.fake src/fake.ts:124:16
 ❯ test/all_functional.spec.ts:96:35

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```

ST-DDT:
I have identified the issue.
The generated string contains a `$` which causes the string replace method to insert the token again. In this case the `$` was directly after a `{` which then caused the issue.
I will provide a fix for the faker method.

ST-DDT:
Fix and test provided: https://github.com/faker-js/faker/pull/688/files#diff-d764addde9b944dba41e8666051bf020b5f44349c6043d6d7770bd58073eca25R83

src/random.ts
```diff
@@ -203,10 +214,48 @@ export class Random {
     object = { foo: 'bar', too: 'car' } as unknown as T,
     field = 'value'
   ): K | T[K] {
+    console.warn(
+      'Deprecation Warning: faker.random.objectElement() is now split into 2 separate methods: faker.datatype.objectKey() and faker.datatype.objectValue().'
+    );
```
ST-DDT, line 219: This should use the new deprecation helper method.

ST-DDT (changes requested):

src/random.ts
ST-DDT: @xDivisionByZerox This is almost ready to merge. Do you have time to address this soonish or should we change it for you?

Shinigami92:
I just found out that we need JSDoc with `@deprecated` on ALL signatures!



In the one line with not using 'key', it is not strikedthrough...

src/random.ts
```diff
@@ -194,24 +212,67 @@ export class Random {
...
+   * faker.random.objectKey({ myProperty: 'myValue' }) // 'myProperty'
+   */
+  objectKey<T extends Record<string, unknown>>(
+    object: T = { foo: 'bar', too: 'car' } as unknown as T
+  ): keyof T {
     const array: Array<keyof T> = Object.keys(object);
-    const key = this.arrayElement(array);
+    const key = this.faker.random.arrayElement(array);
+
+    return key;
```
ST-DDT, line 281: IMO you can directly return here instead of storing it in const.

src/random.ts
```diff
@@ -194,24 +212,67 @@ export class Random {
...
+   * faker.random.objectValue({ myProperty: 'myValue' }) // 'myValue'
+   */
+  objectValue<T extends Record<string, unknown>>(
+    object: T = { foo: 'bar', too: 'car' } as unknown as T
+  ): T[keyof T] {
+    const key = this.faker.random.objectKey(object);
+    const value = object[key];
 
-    return field === 'key' ? (key as K) : (object[key] as T[K]);
+    return value;
```
ST-DDT, line 299: Same here.

src/random.ts
```diff
@@ -194,24 +211,92 @@ export class Random {
...
+   * Returns a random key from given object.
+   *
+   * @param object The optional object to be used. If not specified, it defaults to `{ foo: 'bar', too: 'car' }`.
+   *
+   * @example
+   * faker.random.objectKey() // 'foo'
+   * faker.random.objectKey({ myProperty: 'myValue' }) // 'myProperty'
+   */
+  objectKey<T extends Record<string, unknown>>(
+    object: T = { foo: 'bar', too: 'car' } as unknown as T
```
ST-DDT, line 276: IMO this method should not have a default value.

src/random.ts
```diff
@@ -194,24 +211,92 @@ export class Random {
...
+   * Returns a random value from given object.
+   *
+   * @param object The optional object to be used. If not specified, it defaults to `{ foo: 'bar', too: 'car' }`.
+   *
+   * @example
+   * faker.random.objectValue() // 'bar'
+   * faker.random.objectValue({ myProperty: 'myValue' }) // 'myValue'
+   */
+  objectValue<T extends Record<string, unknown>>(
+    object: T = { foo: 'bar', too: 'car' } as unknown as T
```
ST-DDT, line 294: IMO this method should not have a default value.

ST-DDT (changes requested):

ST-DDT (approved):

Shinigami92 (approved):

PR #504, opened 2022-02-17
feat: add internet.emoji

### LINKED ISSUE #497: Add a faker.internet.emoji function

I think it would be useful to generate random emoji. My use case is that I have documents that can have an emoji icon (a bit like notion), so I would like to generate random emoji for this.

### Suggested solution

I suggest creating a `faker.internet.emoji()` function that return a random emoji

### Alternative

_No response_

### Additional context

_No response_

hugoattal:
fix: https://github.com/faker-js/faker/issues/497

So I added a function `faker.internet.emoji()` that generate emojis.
You can also filter only some groups of emoji with `faker.internet.emoji(["food"])` for example.

I took all the fully-qualified emoji of Unicode : https://unicode.org/Public/emoji/14.0/emoji-test.txt

I didn't separate each subgroup, as I don't think people need that much granularity.
I also disabled prettier on the `emoji.ts` file to have a more readable file so that you can spot the group names, but I can keep the prettier format if you prefer (I will have a huge lines number contribution 😛 !).

For the test, I only check if the string has a length of at least 1, as the regex to check emoji are quite hacky for now.
I found using a package (such as this one: https://github.com/mathiasbynens/emoji-regex) only for this test to be unnecessary.


src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,121 @@
...
+    '👌','👌🏻','👌🏼','👌🏽','👌🏾','👌🏿','🤌','🤌🏻','🤌🏼','🤌🏽','🤌🏾','🤌🏿','🤏','🤏🏻','🤏🏼','🤏🏽','🤏🏾','🤏🏿','✌️','✌🏻','✌🏼','✌🏽','✌🏾','✌🏿','🤞','🤞🏻','🤞🏼','🤞🏽','🤞🏾','🤞🏿','🫰','🫰🏻','🫰🏼','🫰🏽','🫰🏾','🫰🏿','🤟','🤟🏻','🤟🏼','🤟🏽','🤟🏾','🤟🏿','🤘','🤘🏻','🤘🏼','🤘🏽','🤘🏾','🤘🏿','🤙','🤙🏻','🤙🏼','🤙🏽','🤙🏾','🤙🏿',
+    '👈','👈🏻','👈🏼','👈🏽','👈🏾','👈🏿','👉','👉🏻','👉🏼','👉🏽','👉🏾','👉🏿','👆','👆🏻','👆🏼','👆🏽','👆🏾','👆🏿','🖕','🖕🏻','🖕🏼','🖕🏽','🖕🏾','🖕🏿','👇','👇🏻','👇🏼','👇🏽','👇🏾','👇🏿','☝️','☝🏻','☝🏼','☝🏽','☝🏾','☝🏿','🫵','🫵🏻','🫵🏼','🫵🏽','🫵🏾','🫵🏿',
+    '👍','👍🏻','👍🏼','👍🏽','👍🏾','👍🏿','👎','👎🏻','👎🏼','👎🏽','👎🏾','👎🏿','✊','✊🏻','✊🏼','✊🏽','✊🏾','✊🏿','👊','👊🏻','👊🏼','👊🏽','👊🏾','👊🏿','🤛','🤛🏻','🤛🏼','🤛🏽','🤛🏾','🤛🏿','🤜','🤜🏻','🤜🏼','🤜🏽','🤜🏾','🤜🏿',
+    '👏','👏🏻','👏🏼','👏🏽','👏🏾','👏🏿','🙌','🙌🏻','🙌🏼','🙌🏽','🙌🏾','🙌🏿','🫶','🫶🏻','🫶🏼','🫶🏽','🫶🏾','🫶🏿','👐','👐🏻','👐🏼','👐🏽','👐🏾','👐🏿','🤲','🤲🏻','🤲🏼','🤲🏽','🤲🏾','🤲🏿','🤝','🤝🏻','🤝🏼','🤝🏽','🤝🏾','🤝🏿','🫱🏻‍🫲🏼','🫱🏻‍🫲🏽','🫱🏻‍🫲🏾','🫱🏻‍🫲🏿','🫱🏼‍🫲🏻','🫱🏼‍🫲🏽','🫱🏼‍🫲🏾','🫱🏼‍🫲🏿','🫱🏽‍🫲🏻','🫱🏽‍🫲🏼','🫱🏽‍🫲🏾','🫱🏽‍🫲🏿','🫱🏾‍🫲🏻','🫱🏾‍🫲🏼','🫱🏾‍🫲🏽','🫱🏾‍🫲🏿','🫱🏿‍🫲🏻','🫱🏿‍🫲🏼','🫱🏿‍🫲🏽','🫱🏿‍🫲🏾','🙏','🙏🏻','🙏🏼','🙏🏽','🙏🏾','🙏🏿',
```
ST-DDT, line 26: Some of don't seem to work in my browser or IDE:

> '🫶','🫶🏻'

AFAICT these are added in `E14.0` . Should we exclude those then?
Or do you have an ETA when browsers/IDEs are going to add them?

hugoattal: Or I can simply use the E12.0 : https://unicode.org/Public/emoji/12.0/emoji-test.txt

src/definitions/internet.ts
```diff
@@ -16,6 +16,22 @@ export interface InternetDefinitions {
    * Some free-mail domains used in that country (e.g. `gmail.de`).
    */
   free_email: string[];
+  /**
+   * List of all fully-qualified emoji ordered by groups.
+   */
+  emoji: {
+    [key in
```
Shinigami92, line 23: This doesn't look correctly formatted with prettier

Maybe you also want to do something like `emoji: Record<EmojiGroup, string[]>`

src/internet.ts
```diff
@@ -405,4 +405,25 @@ export class Internet {
...
+  /**
+   * Generates a random emoji.
+   *
+   * @param filters A list of the emoji groups that should be used.
+   *
+   * @example
+   * faker.internet.emoji() // '🥰'
+   * faker.internet.emoji(['food', 'nature']) // '🥐'
+   */
+  emoji(filters?: Array<keyof InternetDefinitions['emoji']>): string {
```
Shinigami92, line 418: Maybe provide an option obj like

```ts
emoji(options: {
  filters?: Array<...>
  versions?: Array<EmojiVersion>
} = {})
```

Shinigami92 (changes requested):

src/internet.ts
ST-DDT: I object having an emoji version parameter.

Shinigami92: > I object having an emoji version parameter.

Any reason why?

ST-DDT: IMO this library is not about testing charsets or rendering pipelines.
If you need those, you have to use the full set of emojis/unicode anyway.
Thus we should not differenciate between versions "words". We also dont add support for switching between new or old language dictionaries.

Shinigami92: @ST-DDT Makes sense and I accept that :+1:

src/locales/en/internet/emoji.ts
hugoattal: > Support for Emoji 14.0 is now available via Google's [Android 12L](https://emojipedia.org/google/android-12l/) update and the first beta update of Apple's [iOS 15.4](https://blog.emojipedia.org/first-look-new-emojis-in-ios-15-4/).

Source: https://emojipedia.org/emoji-14.0/

---

I'm sorry, but I really don't understand why the fact that not all tools support it means we shouldn't include it 😕. If a user can write an emoji on their Android and people on Windows can't see it, well it's "too bad", but it's a real word situation.

I believe this library is to generate/test real-world data, not narrowing down to easy situations.

Shinigami92:

hugoattal:
Yep sorry, kinda forgot about it... I'll do the rebase and the updates tomorrow!

src/definitions/internet.ts
hugoattal: I did that here: https://github.com/faker-js/faker/blob/8731d5c3ff4d557c3a2eae54f46ee0042e7cc966/src/definitions/internet.ts#L34

This is correctly formatted by ESLint, I find this easier to read than having everything on a single line, but I can put it on one line if you prefer!

hugoattal:
Sorry for the little mess in the commits, first time I rebased a forked repo 😅...
Everything should be good now 🙂

ST-DDT (dismissed):

ejcheng (dismissed):

src/locales/en/internet/emoji.ts
pkuczynski: I am with @ST-DDT on this and would stick to version 13. Faker does not need to generate every possible piece of data, but rather data which looks like valid and coming from a user. If those new emojis are not supported yet on peoples devices, it's unlikely someone would use them, right? So there is no pressure to include them on our side either.

src/definitions/internet.ts
```diff
@@ -1,5 +1,17 @@
 import { allOf } from './utils';
 
+export type EmojiGroups =
```
pkuczynski, line 3:
```suggestion
export type EmojiGroup =
```

Personally I would call it a `type` and not a `group`, but that's only a side note...

src/definitions/internet.ts
```diff
@@ -1,5 +1,17 @@
...
+  | 'smileys'
```
pkuczynski, line 4:
```suggestion
  | 'smiley'
```

Can you keep singular form for consistency?

src/internet.ts
```diff
@@ -426,4 +426,25 @@ export class Internet {
...
```
pkuczynski, line 439:
```suggestion
  emoji(group?: Array<EmojiGroup>): string {
```

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,121 @@
+// prettier-ignore
+
+export default {
```
pkuczynski, line 3: Could you add a link to the source of this? And if we decide to limit on the version maybe a short note about it...

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,121 @@
+// prettier-ignore
```
pkuczynski, line 1: Emoji are not connected to i18n and they the same for every language, right? So is it a good place to put them here under `en` @ST-DDT @Shinigami92? Feels to me they should find a better location...

pkuczynski (changes requested):

src/definitions/internet.ts
pkuczynski: After a bit of thinking it really feels more like a `type`: `food`, `smiley`, `body`... and in the definitions, that just grouped by type. But what you want from the user in the method definitions is to _limit emoji to a certain type_...

src/definitions/internet.ts
hugoattal: Fixed here: https://github.com/faker-js/faker/pull/504/commits/6badfaaa55667c694d80753bbb63698be463229a

src/internet.ts
hugoattal: Done here: https://github.com/faker-js/faker/pull/504/commits/82cad3d5a251ec168f8266202449feca429106cf

src/definitions/internet.ts
hugoattal: From the unicode source, this is call a `group`. Can you confirm that you want me to rename it to `type`?
cf https://unicode.org/Public/emoji/14.0/emoji-test.txt

image

src/locales/en/internet/emoji.ts
hugoattal: Added here: https://github.com/faker-js/faker/blob/82cad3d5a251ec168f8266202449feca429106cf/src/locales/en/internet/emoji.ts#L3

> And if we decide to limit on the version maybe a short note about it...

I'm not sure to understand what you mean by that... Could you elaborate?

src/locales/en/internet/emoji.ts
hugoattal: I have no problem putting it somewhere else, but I couldn't find a function that were not inside the locals folder. (and the default one being `en`).

Could you point me to an example? Or should I update the architecture to have to generic functions?

src/definitions/internet.ts
pkuczynski: In the source file its called a group, which indeed collects emojis by type :) Its kind of like you would group cars by size: sedan, suv, etc. Collection of all SUV would be a group, but the attribute they are grouped by would be `size`. Does that make sense?

src/locales/en/internet/emoji.ts
pkuczynski: @ST-DDT suggested we stick with version 13.1 which seems to have wider adoption (and I agree with this suggestion). Than we should add a note about it in case someone else stumble upon this next time. Does this make sense?

I would also link https://www.unicode.org/versions/enumeratedversions.html as primary source as the `txt` is only a partial of the whole standard...

src/locales/en/internet/emoji.ts
pkuczynski: I see your pain as there is no standard for it. I would wait for what other guys think. but maybe same folder but `internet.emoji.ts`?

src/definitions/internet.ts
pkuczynski: And here they call them _category_ https://home.unicode.org/emoji/emoji-frequency/

Yeah, I am pretty sure it should be called differently then `group` as its too general word. You can group emojis in many different ways...

src/locales/en/internet/emoji.ts
Shinigami92: We are drafting that out in https://github.com/faker-js/faker/issues/823
We had unwritten plans for this for a month or two, but never wrote it down into an issue, so finally we have this issue now :slightly_smiling_face:

src/locales/en/internet/emoji.ts
ST-DDT: **Maintainer decision**: Please switch to emoji v13.0 for now.
For us, it is more important to generate valid/working data, then generating all possible data.
Most users use this framework in combination with existing tools/browsers and thus there is hardly anything they can do in their application to "fix" these unsupported emojis.

Tests that are intended to test compatibility with a certain new feature are best implemented using exactly that in their tests. Which also need special visual tests to very are they displayed as icon or as missing char text.

Tests that are intended to test the behavior with explicitly unknown emojis/characters are the same.

We may or may not add an option for a specific emoji versions later, but for now I consider this out of the scope of our library.

> I believe this library is to generate/test real-world data, not narrowing down to easy situations.

We do generate/use real world data, not arbitary data.
Our strings are limited to printable characters (emojis), not arbitary binary data displayed as string.
Tests for specific edge cases, should always be implemented using fixed values, not random ones, otherwise you have a test that fails 1 in 1000 or less times and you don't know why.

ST-DDT (changes requested):
Please switch to emoji v13.0.
Other then that, it looks good to me and ready to be merged in v6.2.

src/locales/en/internet/emoji.ts
ST-DDT: No need to move it yet. We will move it, once we have decided, where to put it.

src/locales/en/internet/emoji.ts
ST-DDT: Yes, please add the link below the other link:

````ts
// Source: https://unicode.org/Public/emoji/14.0/emoji-test.txt 
// Other versions: https://www.unicode.org/versions/enumeratedversions.html
````

File: src/definitions/internet.ts (hunk 4 above)
Comment (ST-DDT) [reply src/definitions/internet.ts]: @pkuczynski IMO since we use that source we should also stick to their names.
group and category are just synonmys. You could group/categorize them however you want, but we use the "official" groups.

If you have a better name for it, then we can decide on that then.

File: src/internet.ts [hunk 9]
```diff
@@ -426,4 +426,26 @@ export class Internet {
...
+   * @example
+   * faker.internet.emoji() // '🥰'
+   * faker.internet.emoji(['food', 'nature']) // '🥐'
+   */
+  emoji(groups?: Array<EmojiGroup>): string {
+    groups =
+      groups ||
+      (Object.keys(this.faker.definitions.internet.emoji) as Array<
+        keyof InternetDefinitions['emoji']
+      >);
```
ST-DDT, line 445:
```suggestion
      (Object.keys(this.faker.definitions.internet.emoji) as Array<EmojiGroup>);
```

is the cast needed at all?

hugoattal: Yes, this is needed:
image

Though, I updated the type in https://github.com/faker-js/faker/pull/504/commits/faba4db059a401d45c7521ca76931e94aa6eeaef

src/internet.ts
```diff
@@ -426,4 +426,24 @@ export class Internet {
     };
     return _password(len, memorable, pattern, prefix);
   }
+
+  /**
+   * Generates a random emoji.
+   *
+   * @param filters A list of the emoji types that should be used.
+   *
+   * @param types
```
ST-DDT, line 435:
```suggestion
   * @param types A list of the emoji types that should be used.
   *
```

src/locales/en/internet/emoji.ts
hugoattal: I updated to v13 and added the link: https://github.com/faker-js/faker/pull/504/commits/28f13af0a592f9417c11db4adf392873b2c176f0

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,125 @@
...
+
+// Source: https://unicode.org/Public/emoji/13.0/emoji-test.txt
+// Others: https://unicode.org/Public/emoji/
+// Versions: https://www.unicode.org/versions/enumeratedversions.html
```
ST-DDT, line 5:
```suggestion
// Versions: https://unicode.org/versions/enumeratedversions.html
```

src/definitions/internet.ts
hugoattal: Aaaah, I renamed it to type >< ... https://github.com/faker-js/faker/pull/504/commits/24727edfbe7c4f7863e77f91ccda147a9eba627f

Should I change it back?

src/locales/en/internet/emoji.ts
pkuczynski: I totally support what has been written here...

ST-DDT (changes requested):

src/definitions/internet.ts
pkuczynski: My better name is `Type` @ST-DDT :)

src/locales/en/internet/emoji.ts
hugoattal: Their base domain is with the `www` and all their link use the domain with the `www`, so I let it like this.
You really want to remove the `www` ?

image

src/definitions/internet.ts
pkuczynski: > Should I change it back?

Nah, I think you can keep it, unless @ST-DDT disagrees?

src/definitions/internet.ts
```diff
@@ -16,6 +28,10 @@ export interface InternetDefinitions {
    * Some free-mail domains used in that country (e.g. `gmail.de`).
    */
   free_email: string[];
+  /**
+   * List of all fully-qualified emoji ordered by groups.
```
pkuczynski, line 32:
```suggestion
   * List of all fully-qualified emojis.
```

pkuczynski (dismissed):

src/locales/en/internet/emoji.ts
ST-DDT: If they use www everywhere then please do the same in the comment. I just want all three links to be consistent.

hugoattal: Oops, yep, I missed this 😅... fixed: https://github.com/faker-js/faker/pull/504/commits/213ad4497f336c34f01ca7d75fa8f9ac84513102

ST-DDT (dismissed):

src/definitions/internet.ts
```diff
@@ -1,5 +1,17 @@
 import { allOf } from './utils';
 
+export type EmojiType =
```
Shinigami92, line 3: IMO we should move this type out of definitions into the module file
~~Also export it same as e.g. GenderType in name module~~ nevermind, already doing that :+1: just the file need to change

src/internet.ts
```diff
@@ -426,4 +426,23 @@ export class Internet {
...
+  /**
+   * Generates a random emoji.
+   *
+   * @param types A list of the emoji types that should be used.
+   *
+   * @example
+   * faker.internet.emoji() // '🥰'
+   * faker.internet.emoji(['food', 'nature']) // '🥐'
+   */
+  emoji(types?: Array<EmojiType>): string {
```
Shinigami92, line 439: Just to be a bit more future safe and we already try to do this in other places, could we use options here?

```suggestion
  emoji(options: { types?: Array<EmojiType> } = {}): string {
```

These can be extended on demand in the future

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,125 @@
+// prettier-ignore
```
Shinigami92, line 1: Do we really need to not format this file? It could raise potentially more git conflicts when working on this file later on.

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,125 @@
...
+
+// Source: https://www.unicode.org/Public/emoji/13.0/emoji-test.txt
+// Others: https://www.unicode.org/Public/emoji/
+// Versions: https://www.unicode.org/versions/enumeratedversions.html
+
+export default {
+  smiley: [
+    '😀','😃','😄','😁','😆','😅','🤣','😂','🙂','🙃','😉','😊','😇',
```
Shinigami92, line 9: Is it possible to generate these emojis with code-points like `U+1F600` and then just with a range?

test/internet.spec.ts
```diff
@@ -579,6 +583,16 @@ describe('internet', () => {
           expect(password).satisfy(validator.isStrongPassword);
         });
       });
+
+      describe('emoji', () => {
+        it('should return a random emoji', () => {
+          const emoji = faker.internet.emoji();
+
+          expect(emoji).toBeTruthy();
+          expect(typeof emoji).toBe('string');
```
Shinigami92, line 592:
```suggestion
          expect(emoji).toBeTypeOf('string');
```

test/internet.spec.ts
```diff
@@ -579,6 +583,16 @@ describe('internet', () => {
...
+          expect(emoji.length).toBeGreaterThanOrEqual(1);
```
Shinigami92, line 593: If testing against a lower bound, should we also test against a upper bound?

Shinigami92 (changes requested):

src/locales/en/internet/emoji.ts
hugoattal: I feel like it would be a bit overengineered, and very difficult to read... Also, emojis does not always have consecutive code-points. Here are the ones you selected:

```
1F600  ; fully-qualified     # 😀 E1.0 grinning face
1F603  ; fully-qualified     # 😃 E0.6 grinning face with big eyes
1F604  ; fully-qualified     # 😄 E0.6 grinning face with smiling eyes
1F601  ; fully-qualified     # 😁 E0.6 beaming face with smiling eyes
1F606  ; fully-qualified     # 😆 E0.6 grinning squinting face
1F605  ; fully-qualified     # 😅 E0.6 grinning face with sweat
1F923  ; fully-qualified     # 🤣 E3.0 rolling on the floor laughing
1F602  ; fully-qualified     # 😂 E0.6 face with tears of joy
1F642  ; fully-qualified     # 🙂 E1.0 slightly smiling face
1F643  ; fully-qualified     # 🙃 E1.0 upside-down face
1FAE0  ; fully-qualified     # 🫠 E14.0 melting face
1F609  ; fully-qualified     # 😉 E0.6 winking face
1F60A  ; fully-qualified     # 😊 E0.6 smiling face with smiling eyes
1F607  ; fully-qualified     # 😇 E1.0 smiling face with halo
```

test/internet.spec.ts
hugoattal: I don't think it's future-proof as some emojis are made of several emojis next to each others

image

src/locales/en/internet/emoji.ts
hugoattal: I believe it's way easier to work with this format, but I can change this if you confirm that you prefer a 5k lines file 😅.

src/definitions/internet.ts
hugoattal: Done here: https://github.com/faker-js/faker/pull/504/commits/0052ba02708cbd16cc3b538d34e074b3cd7e4ed7

src/internet.ts
hugoattal: Done here: https://github.com/faker-js/faker/pull/504/commits/af8f73bf96f5ea5bb58924b75da2829caeb02c56 !

src/internet.ts
```diff
@@ -426,4 +438,23 @@ export class Internet {
...
+   * faker.internet.emoji() // '🥰'
+   * faker.internet.emoji({types: ['food', 'nature']}) // '🥐'
+   */
+  emoji(options: { types?: Array<EmojiType> } = {}): string {
+    options.types =
+      options.types ||
+      (Object.keys(this.faker.definitions.internet.emoji) as Array<EmojiType>);
+    const group = this.faker.random.arrayElement(options.types);
+    return this.faker.random.arrayElement(
+      this.faker.definitions.internet.emoji[group]
```
Shinigami92, line 457:
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
   */
  emoji(options: { types?: ReadonlyArray<EmojiType> } = {}): string {
    const {
      types = Object.keys(this.faker.definitions.internet.emoji) as Array<EmojiType>
    } = options;
    const emojiType = this.faker.random.arrayElement(types);
    return this.faker.random.arrayElement(
      this.faker.definitions.internet.emoji[group]
```

hugoattal: Done: https://github.com/faker-js/faker/pull/504/commits/194604e32f9be56d21795649ca04fae19a084122

src/locales/en/internet/emoji.ts
Shinigami92: We also already have a 8k lines file here: https://github.com/faker-js/faker/blob/main/src/locales/en/system/mimeTypes.ts
So I would aim just for consistency here and format it. It's easier to work with them in the future when comparing diffs.

hugoattal: Done here: https://github.com/faker-js/faker/pull/504/commits/cdb59221298a4b6637602c11d20163040f5d0ad3 !

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🚺',
+    '🚻',
+    '🚼',
+    '🚾',
+    '🛂',
+    '🛃',
+    '🛄',
+    '🛅',
+    '⚠️',
+    '⚠',
```
Shinigami92, line 4042: This emoji is not correctly displayed on my system :thinking: 

image

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🚫',
+    '🚳',
+    '🚭',
+    '🚯',
+    '🚱',
+    '🚷',
+    '📵',
+    '🔞',
+    '☢️',
+    '☢',
```
Shinigami92, line 4054: There are some non-emoji codes here
image

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '♏',
+    '♐',
+    '♑',
+    '♒',
+    '♓',
+    '⛎',
+    '🔀',
+    '🔁',
+    '🔂',
+    '▶️',
```
Shinigami92, line 4129: And here 
image

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '😉',
+    '😊',
+    '😇',
+    '🥰',
+    '😍',
+    '🤩',
+    '😘',
+    '😗',
+    '☺️',
+    '☺',
```
Shinigami92, line 26: These are not displayed as emoji

image

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🥳',
+    '🥸',
+    '😎',
+    '🤓',
+    '🧐',
+    '😕',
+    '😟',
+    '🙁',
+    '☹️',
+    '☹',
```
Shinigami92, line 80: And here 
image

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🥱',
+    '😤',
+    '😡',
+    '😠',
+    '🤬',
+    '😈',
+    '👿',
+    '💀',
+    '☠️',
+    '☠',
```
Shinigami92, line 109: And here
image

Shinigami92 (commented):
I found some non-emoji emojis :thinking: Not sure what to do with them or it is my system and a non-installed font
Did not saw them before, cause it was just a large chunk of icons. Now were everything is in it's own line, it's much better readable :slightly_smiling_face:

src/locales/en/internet/emoji.ts
hugoattal: I think it depends on the system, it's very difficult to find information about emoji implementation state.
([this link](https://developer.android.com/guide/topics/ui/look-and-feel/emoji2) suggests that Android support 13.1, but it's not even clear)

What do you suggest?
- Remove the one that does not display on my system
- Try the [12.0](https://unicode.org/Public/emoji/12.0/emoji-test.txt) (or below)
- Keep it like this

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🤌🏾',
+    '🤌🏿',
+    '🤏',
+    '🤏🏻',
+    '🤏🏼',
+    '🤏🏽',
+    '🤏🏾',
+    '🤏🏿',
+    '✌️',
+    '✌',
```
import-brain, line 229: Not displayed as emoji

import-brain, line 229: Not displayed as emoji

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🖕🏾',
+    '🖕🏿',
+    '👇',
+    '👇🏻',
+    '👇🏼',
+    '👇🏽',
+    '👇🏾',
+    '👇🏿',
+    '☝️',
+    '☝',
```
import-brain, line 290: here

src/locales/en/internet/emoji.ts
```diff
@@ -0,0 +1,4604 @@
...
+    '🤲🏿',
+    '🤝',
+    '🙏',
+    '🙏🏻',
+    '🙏🏼',
+    '🙏🏽',
+    '🙏🏾',
+    '🙏🏿',
+    '✍️',
+    '✍',
```
import-brain, line 364: here

ejcheng (changes requested):

src/locales/en/internet/emoji.ts
hugoattal: Those are valid emojis and are displayed as such on some devices (android for example) 😅

But I still don't know what to do about this, cf https://github.com/faker-js/faker/pull/504#discussion_r848828082 ...

import-brain: > Those are valid emojis and are displayed as such on some devices (android for example) 😅
> 
> But I still don't know what to do about this, cf [#504 (comment)](https://github.com/faker-js/faker/pull/504#discussion_r848828082) ...

Hmm, this is definitely a big problem. It's pretty much impossible to ensure cross-compatibility with all systems (for example, the emojis I marked as not showing don't show on Mac).

src/locales/en/internet/emoji.ts
pkuczynski: I think we should go max 13.1

src/locales/en/internet/emoji.ts
hugoattal: Hum, I don't see this as a big problem, it's still displayed as a char...

Here is my point: https://github.com/faker-js/faker/pull/504#discussion_r809905180. However, the maintainers decided to use an older version of emoji (13.0): https://github.com/faker-js/faker/pull/504#discussion_r846792591

Despite the fact that I don't support this decision, I have no problem following it or doing anything to make you (the maintainers) happy with this PR.

I just need a directive concerning this issue, I listed 3 solutions here https://github.com/faker-js/faker/pull/504#discussion_r848828082, I just need someone to pick one 😅 !

src/locales/en/internet/emoji.ts
hugoattal: This is already the case, but there's still some emoji that display as char in some system: https://unicode.org/Public/emoji/13.1/emoji-test.txt

So, should I keep it? (I'm in favor of keeping it)

src/locales/en/internet/emoji.ts
Shinigami92: image
That's really interesting :eyes: For me (arch linux) they display as exact same emoji, so they are duplicates :thinking: 
I think we don't need duplicates.

import-brain: Very odd

hugoattal: Oooooh, I understand now, it's my fault, I forgot to remove the unqualified version of emoji when I updated to 3.0!

image


I'm sorry, I'm fixing it now...

hugoattal: I fixed it here: https://github.com/faker-js/faker/pull/504/commits/51a29d3762eb2f6fef3e4f4b7b42390a108b65d1

Shinigami92:
45 and 46 are duplicates but now removed both :thinking: 
51 and 68 was okay but removed now?
80 was correctly removed

I'm not sure if the _unqualified_ is the correct "filter"

image

hugoattal:
@Shinigami92 You're right some were missing, I fixed it here: https://github.com/faker-js/faker/pull/504/commits/4e239d760ff868b35a0690b447bd94fb07b49b7b

src/locales/en/internet/emoji.ts
Shinigami92: > I fixed it here: [51a29d3](https://github.com/faker-js/faker/commit/51a29d3762eb2f6fef3e4f4b7b42390a108b65d1) (Sorry again...)

Hey, no need to be sorry :slightly_smiling_face: 
Actually I really like that we have so much different systems and check from multiple perspectives so that we can bring the best to our users :raised_hands:

src/internet.ts
```diff
@@ -426,4 +438,25 @@ export class Internet {
...
   }
+
+  /**
+   * Generates a random emoji.
+   *
+   * @param options Options object.
+   * @param options.types A list of the emoji types that should be used.
+   * @example
+   * faker.internet.emoji() // '🥰'
+   * faker.internet.emoji({types: ['food', 'nature']}) // '🥐'
```
Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

Shinigami92, line 449: @ST-DDT will api-docs ignore such unformatted example docs? Or should we care about that this should be formatted?
```suggestion
   * faker.internet.emoji({ types: ['food', 'nature'] }) // '🥐'
```

ST-DDT: It will print them as is.
https://github.com/faker-js/faker/blob/360bec7c7f3269049eb748523f2d271b6ed36d75/scripts/apidoc/signature.ts#L128-L135
The examples are highlighted for ts though.
https://github.com/faker-js/faker/blob/360bec7c7f3269049eb748523f2d271b6ed36d75/scripts/apidoc/signature.ts#L149
We probably could run them past prettier, but then maybe as a check whether they are formatted correctly instead of just formatting them for the website.

Shinigami92: Oh I just recognized we don't have them only for the docs, but actually also inline docs. So we should format these in code! ^^

Shinigami92 (approved):

ST-DDT:
I'm very busy today, so I will have a look at this tomorrow.

ejcheng (approved):

ST-DDT (approved):

prisis (approved):

Issue #466, opened 2022-02-10
Deprecate `helpers.randomize`


Shinigami92:
### Clear and concise description of the problem

https://github.com/faker-js/faker/blob/999556bf83018daf010b05350d58934ad5bd0761/src/helpers.ts#L118-L122
This just forwards the call

### Suggested solution

Add a `@deprecated` + a `console.warn` like in other places

xDivisionByZerox:
Can I take this? Seems easy enough.

Shinigami92:
Let's put this into v6.1 so folks can just switch to 6.0.0 as a simple 5.5.3 replacement without doing to much stuff at once and instantly need to also change these methods.
I know it's just a warning, but I want to make a smooth transition as possible for the start.

PR #506, opened 2022-02-18
docs: deprecate randomize

xDivisionByZerox:
Created in relation to #466.

What has changed:
The jsdoc of `helpers.randomize` now includes the `@deprecated` decorator. 
Furthermore, the method will now print a `console.warn` to inform the end-user about the deprecation.


ST-DDT (changes requested):
Please move the `@deprecated` tag down to comply with #512

ST-DDT (approved):

Shinigami92 (commented):
Could you please add a check in the test that the warning is called?
We already do this for methods in random here: https://github.com/faker-js/faker/blob/2d68da1140a4f6d83a94e296506cfa5b126dd067/test/random.spec.ts#L239

Shinigami92 (approved):

ST-DDT (approved):

ST-DDT (approved):

Issue #375, opened 2022-01-30
Unreachable code in finance.creditCardNumber()


ST-DDT:
### Describe the bug

https://github.com/faker-js/faker/blob/d4125e2a7d69042c37ccecea89f1525c4424a45b/src/finance.ts#L231-L261

Due to 
https://github.com/faker-js/faker/blob/d4125e2a7d69042c37ccecea89f1525c4424a45b/src/finance.ts#L235
`localeFormat` cannot be a `string` thus
https://github.com/faker-js/faker/blob/d4125e2a7d69042c37ccecea89f1525c4424a45b/src/finance.ts#L247-L248
is unreachable

### Reproduction

Set the `faker.definitions.finance.credit_card` to a string and run the method.

### Additional Info

Discovered in #363

xDivisionByZerox:
I would provide a PR for this, Can someone assign me to this?

PR #508, opened 2022-02-18
fix: remove unreachable code finance

xDivisionByZerox:
Created in relation to #375

The original issue only stated code for the following `if` statement:
```typescript
if (provider in localeFormat) {
    formats = localeFormat[provider]; // there could be multiple formats
    if (typeof formats === 'string') {
        format = formats;
    } else {
        format = this.faker.random.arrayElement(formats);
    }
}
```  

I also found, that the last `else` statement:
```typescript
else {
    // Choose a random provider
    // TODO ST-DDT 2022-01-30: #375 This is impossible to access
    if (typeof localeFormat === 'string') {
        format = localeFormat;
    } else if (typeof localeFormat === 'object') {
        // Credit cards are in a object structure
        formats = this.faker.random.objectElement(localeFormat, 'value'); // There could be multiple formats
        if (typeof formats === 'string') {
          format = formats;
        } else {
          format = this.faker.random.arrayElement(formats);
        }
    }
}
```
has the same problem. So I removed the unreachable parts as well.

### DIFF
--- a/src/finance.ts
+++ b/src/finance.ts
@@ -266,7 +266,7 @@ export class Finance {
       format = provider;
     } else {
       // Choose a random provider
-      // Credit cards are in a object structure
+      // Credit cards are in an object structure
       const formats = this.faker.random.objectElement(localeFormat, 'value'); // There could be multiple formats
       format = this.faker.random.arrayElement(formats);
     }
--- a/src/finance.ts
+++ b/src/finance.ts
@@ -258,32 +258,17 @@ export class Finance {
    */
   creditCardNumber(provider = ''): string {
     let format: string;
-    let formats: string | string[];
     const localeFormat = this.faker.definitions.finance.credit_card;
     if (provider in localeFormat) {
-      formats = localeFormat[provider]; // there could be multiple formats
-      if (typeof formats === 'string') {
-        format = formats;
-      } else {
-        format = this.faker.random.arrayElement(formats);
-      }
+      format = this.faker.random.arrayElement(localeFormat[provider]);
     } else if (provider.match(/#/)) {
       // The user chose an optional scheme
       format = provider;
     } else {
       // Choose a random provider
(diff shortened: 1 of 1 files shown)


src/finance.ts
```diff
@@ -261,32 +261,17 @@ export class Finance {
...
-      } else if (typeof localeFormat === 'object') {
-        // Credit cards are in a object structure
-        formats = this.faker.random.objectElement(localeFormat, 'value'); // There could be multiple formats
-        if (typeof formats === 'string') {
-          format = formats;
-        } else {
-          format = this.faker.random.arrayElement(formats);
-        }
-      }
+      // Credit cards are in a object structure
```
Shinigami92, line 272:
```suggestion
      // Credit cards are in an object structure
```

Shinigami92 (commented):
Somehow the node 14 test fails, could you have a look?

ST-DDT (dismissed):

ST-DDT:
FFR:
grafik

ST-DDT:
@Shinigami92 I assume that test is wrong.

https://github.com/faker-js/faker/blob/431c10887ae43c5ace1f653db91b5540c511db7e/test/random.spec.ts#L180-L187

https://github.com/faker-js/faker/blob/431c10887ae43c5ace1f653db91b5540c511db7e/src/random.ts#L383

It just has a very low chance of failing (~0.2%).

The test should be fixed like this:

````diff
- expect(alphaText).match(/[b-oq-z]/);
+ expect(alphaText).match(/^[0-9b-oq-z]{5}$/);
````

The same is true for the following test.
I also recommend to ban more chars, as the current test only has a chance of about ~25%(?) to actually the test the expected result.

ST-DDT (approved):

Shinigami92 (dismissed):

Shinigami92 (approved):

PR #515, opened 2022-02-19
test: rewrite finance iban

Shinigami92:


ST-DDT (approved):

ST-DDT:
No seeded runs?

Shinigami92:
> No seeded runs?

I'm not sure if needed, this is a sub-test instead of like the others works
We have seeded already here: https://github.com/faker-js/faker/blob/431c10887ae43c5ace1f653db91b5540c511db7e/test/finance.spec.ts#L21

prisis (approved):

Issue #513, opened 2022-02-19
Fix broken random.alphaNumeric() test


ST-DDT:
https://github.com/faker-js/faker/blob/431c10887ae43c5ace1f653db91b5540c511db7e/test/random.spec.ts#L180-L187

https://github.com/faker-js/faker/blob/431c10887ae43c5ace1f653db91b5540c511db7e/src/random.ts#L383

It just has a very low chance of failing (~0.2%).

The test should be fixed like this:

````diff
- expect(alphaText).match(/[b-oq-z]/);
+ expect(alphaText).match(/^[0-9b-oq-z]{5}$/);
````

The same is true for the following test.
I also recommend to ban more chars, as the current test only has a chance of about ~25%(?) to actually the test the expected result.

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/508#issuecomment-1045997000_

**Comment by ST-DDT** [comment]
@Shinigami92 Do you want to fix it, or is it up for grabs?

**Comment by xDivisionByZerox** [comment]
I will do this. No problem.

**Comment by xDivisionByZerox** [comment]
Wouldn't it be easier to iterate over each banned character and expect the alpha text to not include it? Like this
```typescript
it('should be able to ban some characters', () => {
  const bannedChars = ['a', 'p'];
  const alphaText = faker.random.alphaNumeric(5, {
    bannedChars,
  });

  expect(alphaText).toHaveLength(5);
  for(const bannedChar of bannedChars) {
    expect(alphaText.includes(bannedChar)).toBe(false);
  }
});
```

This would allow to change the bannedChars Array without rewriting the RegEx.

ST-DDT:
IMO we should ban all characters `a` - `p` to have a higher probability to "trigger" a banned character.

As for "foreach vs regex": I think it's fine either way.

@Shinigami92 Any preference?

xDivisionByZerox:
We could also force test this by splitting this into 2 tests. One with all alphabetic characters banned and one with all numbers characters baned.

ST-DDT:
I really like that idea.

`foreach` vs `regex`:
If it isn't too much work, start with both variants in after each other (in the same method). 
That makes comparing it easier.
Then we can decide on one of them easily afterwards.

xDivisionByZerox:
Should the current test with some characters baned (instead of all alpha/number) still be reworked or replaced by the 2 tests I suggested?

[...]

PR #517, opened 2022-02-19
fix: test random.alphaNumeric

xDivisionByZerox:
Created in relation to #513.

- split the failing test case into 2 explicit once
- fixed typo in surrounding test case
- added todo test case

### DIFF
--- a/test/random.spec.ts
+++ b/test/random.spec.ts
@@ -186,7 +186,7 @@ describe('random', () => {
       expect(alphaText).toHaveLength(5);
       expect(alphaText).match(/^[0-9]{5}$/);
       for (const bannedChar of bannedChars) {
-        expect(alphaText.includes(bannedChar)).toBe(false);
+        expect(alphaText).not.includes(bannedChar);
       }
     });
 
@@ -199,7 +199,7 @@ describe('random', () => {
       expect(alphaText).toHaveLength(5);
       expect(alphaText).match(/^[a-z]{5}$/);
       for (const bannedChar of bannedChars) {
-        expect(alphaText.includes(bannedChar)).toBe(false);
+        expect(alphaText).not.includes(bannedChar);
       }
     });
 
--- a/test/random.spec.ts
+++ b/test/random.spec.ts
@@ -177,13 +177,30 @@ describe('random', () => {
       expect(actual).toHaveLength(5);
     });
 
-    it('should be able to ban some characters', () => {
+    it('should be able to ban all alphabetic characters', () => {
+      const bannedChars = 'abcdefghijklmnopqrstuvwxyz'.split('');
       const alphaText = faker.random.alphaNumeric(5, {
-        bannedChars: ['a', 'p'],
+        bannedChars,
       });
 
       expect(alphaText).toHaveLength(5);
-      expect(alphaText).match(/[b-oq-z]/);
+      expect(alphaText).match(/^[0-9]{5}$/);
+      for (const bannedChar of bannedChars) {
+        expect(alphaText.includes(bannedChar)).toBe(false);
+      }
+    });
+
+    it('should be able to ban all numeric characters', () => {
+      const bannedChars = '0123456789'.split('');
+      const alphaText = faker.random.alphaNumeric(5, {
+        bannedChars,
+      });
+
+      expect(alphaText).toHaveLength(5);
+      expect(alphaText).match(/^[a-z]{5}$/);
+      for (const bannedChar of bannedChars) {
+        expect(alphaText.includes(bannedChar)).toBe(false);
+      }
     });
 
     it('should be able handle mistake in banned characters array', () => {
--- a/test/random.spec.ts
+++ b/test/random.spec.ts
@@ -212,7 +212,14 @@ describe('random', () => {
       expect(alphaText).match(/[b-oq-z]/);
     });
 
-    it.todo('should be able to handle all possible characters being banned');
+    it.todo('should throw if all possible characters being banned', () => {
+      const bannedChars = 'abcdefghijklmnopqrstuvwxyz0123456789'.split('');
+      expect(() =>
+        faker.random.alphaNumeric(5, {
+          bannedChars,
+        })
+      ).toThrowError();
+    });
   });
 
   describe('deprecation warnings', () => {
--- a/test/random.spec.ts
+++ b/test/random.spec.ts
@@ -211,6 +211,8 @@ describe('random', () => {
       expect(alphaText).toHaveLength(5);
       expect(alphaText).match(/[b-oq-z]/);
     });
+
+    it.todo('should be able to handle all possible characters being banned');
   });
 
   describe('deprecation warnings', () => {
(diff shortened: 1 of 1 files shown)


test/random.spec.ts
```diff
@@ -177,23 +177,42 @@ describe('random', () => {
...
+    it('should be able to handle mistake in banned characters array', () => {
       const alphaText = faker.random.alphaNumeric(5, {
         bannedChars: ['a', 'p', 'a'],
       });
 
       expect(alphaText).toHaveLength(5);
       expect(alphaText).match(/[b-oq-z]/);
     });
+
+    it.todo('should be able to handle all possible characters being banned');
```
ST-DDT, line 215: Could you add the implementation here as well?
I would expect the method to throw if all chars are banned.

xDivisionByZerox: But this shouldn't be the case. Since the method 
1. removes all banned characters which would result in an empty `charsArray`
2. Chooses random elements from `charsArray` with the use of `faker.random.arrayElement`
3. `faker.random.arrayElement` input is `[]`
4. `faker.random.arrayElement` uses `faker.datatype.number` to generate a with input `-1`
5. `faker.datatype.number` returns with `-1`
6. `faker.random.arrayElement` returns `charArray[-1]` = undefined
7. Resolving in a string of undefiend times the `count` input of `faker.random.alphaNumeric`

This means that the implementation needs to be rewritten as well.

xDivisionByZerox: You also write [here](https://github.com/faker-js/faker/issues/513#issuecomment-1046019951) that this test should be implemented with a todo. Testcases with a todo should not have an actual implementation IMO since it negates the sense of it being marked as todo.

ST-DDT: > But this shouldn't be the case. Since the method
> 
>     1. removes all banned characters which would result `charsArray = []`
> 
>     2. Chooses random elements from `charsArray` with the use of `faker.random.arrayElement`
> 
>     3. `faker.random.arrayElement` uses `faker.datatype.number` to generate a with input `charsArray.length - 1` = -1
> 
>     4. `faker.datatype.number` returns with `-1`
> 
>     5. `faker.random.arrayElement` returns `charArray[-1]` = undefined
> 
>     6. Resolving in a string of undefiend times the `count` input of `faker.random.alphaNumeric`
> 
> 
> This means that the implementation needs to be rewritten as well.

Yes, it does return `'undefinedundefinedundefinedundefinedundefined'`, but that is not what I expect it to return in that case.
This is what I was referring to with "bug".

> You also write [here](https://github.com/faker-js/faker/issues/513#issuecomment-1046019951) that this test should be implemented with a todo. Testcases with a todo should not have an actual implementation IMO since it negates the sense of it being marked as todo.

The todo is not related to "write the test" but to enable it (which we cannot do at this point, because of the "bug").
But having or not having the test impl is not a blocker for me.

xDivisionByZerox: Ok, then I misunderstood you.
I will provide an implementation for the test case.

ST-DDT: You can add a comment with this issue: https://github.com/faker-js/faker/issues/518

xDivisionByZerox: f81edec68f496f3b57c0a5c177ad6f6ecf473584 implements the test case

test/random.spec.ts
```diff
@@ -177,23 +177,49 @@ describe('random', () => {
       expect(actual).toHaveLength(5);
     });
 
-    it('should be able to ban some characters', () => {
+    it('should be able to ban all alphabetic characters', () => {
+      const bannedChars = 'abcdefghijklmnopqrstuvwxyz'.split('');
       const alphaText = faker.random.alphaNumeric(5, {
```
Shinigami92, line 182: Maybe we should rename it to `numericText`? Or just `actual`
`alphaText` may be a bit confusing here, cause it has zero "alphas"

test/random.spec.ts
```diff
@@ -177,23 +177,49 @@ describe('random', () => {
...
-        bannedChars: ['a', 'p'],
+        bannedChars,
       });
 
       expect(alphaText).toHaveLength(5);
-      expect(alphaText).match(/[b-oq-z]/);
+      expect(alphaText).match(/^[0-9]{5}$/);
+      for (const bannedChar of bannedChars) {
+        expect(alphaText.includes(bannedChar)).toBe(false);
```
Shinigami92, line 189: I think this can be rewritten to

```ts
expect(alphaText).not.includes(bannedChar);
```

_referees to other lines too_

test/random.spec.ts
```diff
@@ -177,23 +177,49 @@ describe('random', () => {
...
+    it('should be able to ban all alphabetic characters', () => {
+      const bannedChars = 'abcdefghijklmnopqrstuvwxyz'.split('');
       const alphaText = faker.random.alphaNumeric(5, {
-        bannedChars: ['a', 'p'],
+        bannedChars,
       });
 
       expect(alphaText).toHaveLength(5);
-      expect(alphaText).match(/[b-oq-z]/);
+      expect(alphaText).match(/^[0-9]{5}$/);
```
Shinigami92, line 187: Maybe there is a validation for that :thinking: 
Something like (?):
```ts
expect(alphaText).satisfy(validator.isNumeric);
```

---

_We do not need to check length again, cause we already did in line 186_

Shinigami92 (changes requested):

test/random.spec.ts
xDivisionByZerox: I simply copied that from the preview approach without thinking about it. 
Nevertheless, by calling `alphaNumeric` you want to achieve the generation of an alphanumeric string regardless of what characters are actually contained in the string.

test/random.spec.ts
xDivisionByZerox: This is true but does the same thing in the end. If you prefer your style I can rewrite it.

test/random.spec.ts
xDivisionByZerox: Are you aiming for increased readability? Because the regex works perfectly fine.

test/random.spec.ts
Shinigami92: I prefer it, because then we get the specific underlying message of chai/vitest if the runner fails with that expectation

test/random.spec.ts
Shinigami92: Yes, somewhere else we decided to use as less regex as possible exactly for readability

test/random.spec.ts
xDivisionByZerox: Understood. 2a026b5ce9a7d78dac1354771c68cfc57ee0e211 fixes this.

test/random.spec.ts
xDivisionByZerox: In that case, should the regex statements be deleted completely?

Shinigami92: Yeah okay, let's remove it :slightly_smiling_face: and use just the loop

xDivisionByZerox: Removed the regex in 80793aa285df69d7121f89c1b35c38aa3783d281

ST-DDT (approved):

Shinigami92 (approved):

PR #519, opened 2022-02-19
test: cleanup finance test

Shinigami92:
Thanks to a random god (don't know how made that) the finance module was already rewritten :eyes: 
This is "just" a cleanup to bring it in structure like all other tests

Also it now runs all seeded tests inside the for loop

### DIFF
--- a/test/finance.spec.ts
+++ b/test/finance.spec.ts
@@ -9,22 +9,22 @@ const seedRuns = [
     expectations: {
       account: '37917755',
       accountName: 'Money Market Account',
+      routingNumber: '379177554',
+      mask: '(...3791)',
       amount: '374.54',
-      bic: 'UYEOSCP1514',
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):

test/finance.spec.ts
```diff
@@ -9,90 +9,90 @@ const seedRuns = [
...
+      bic: 'LXUEBTZ1',
       transactionDescription:
         'deposit transaction at Trantow - Sanford using card ending with ***(...8076) for PYG 928.52 in account ***62743167',
-      transactionType: 'invoice',
     },
   },
 ];
 
 const functionNames = [
   'account',
```
prisis, line 79: should this be in alphabetical order?

prisis (approved):

test/finance.spec.ts
ST-DDT: The other tests aren't either. I assume they are in declaration order.

PR #520, opened 2022-02-19
build: add redirect aliases for cjs

Shinigami92:
Our `package.json`s `exports` field seems not to work for any case :thinking: 
This creates a `locale/<locale>.js` file for every locale when bundling for cjs so node/npm versions that does not support the exports field can use these redirects.

At least this should work for now and we can release a 6.0.0 soon

### DIFF
--- a/scripts/bundle.ts
+++ b/scripts/bundle.ts
@@ -1,8 +1,24 @@
 import { buildSync } from 'esbuild';
 import { sync as globSync } from 'glob';
+import { existsSync, mkdirSync, rmSync, writeFileSync } from 'node:fs';
 import locales from '../src/locales';
 
 console.log('Building dist for node (cjs)...');
+
+// Generate entry-points for cjs compatibility
+const localeDir = 'locale';
+if (existsSync(localeDir)) {
+  rmSync(localeDir, { recursive: true, force: true });
+}
+mkdirSync(localeDir);
+for (const locale of Object.keys(locales)) {
+  writeFileSync(
+    `${localeDir}/${locale}.js`,
+    `module.exports = require('../dist/cjs/locale/${locale}');\n`,
+    { encoding: 'utf8' }
+  );
+}
+
 buildSync({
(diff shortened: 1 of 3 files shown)


Shinigami92:
Tested successfully with my locale playground against `esm` and `cjs` + with `json-schema-faker`

ST-DDT:
Should we do the same for `locales`?

Shinigami92:
> Should we do the same for `locales`?

locales is not exported from the `package.json`

This is only for usage of 
1. `require(@faker-js/faker)` (package.json -> https://github.com/faker-js/faker/blob/9138ea0fb7ca7d0247588eec7a7c8a7a3cfbbb37/package.json#L24)
2. `require(@faker-js/faker/locale/<locale>)`

everything else is not supported (_currently_)

scripts/bundle.ts
```diff
@@ -1,8 +1,24 @@
...
+// Generate entry-points for cjs compatibility
+const localeDir = 'locale';
+if (existsSync(localeDir)) {
+  rmSync(localeDir, { recursive: true, force: true });
+}
+mkdirSync(localeDir);
+for (const locale of Object.keys(locales)) {
+  writeFileSync(
+    `${localeDir}/${locale}.js`,
+    `module.exports = require('../dist/cjs/locale/${locale}');\n`,
```
ST-DDT, line 17: Please generate a comment, why this file is here (potentially linking to this issue too).

ST-DDT (changes requested):

scripts/bundle.ts
Shinigami92: Do we actually need this for `.gitignore`d files?
We also don't have this for files in `dist`

ST-DDT: Well, the other files are minified, so can skip it.

ST-DDT (approved):

prisis (approved):

PR #534, opened 2022-02-21
feat(time): add default case in time.recent()

bl0cknumber:
Add missing default case in time recent

Close #533

PR #536, opened 2022-02-21
fix : added the default date case to time.ts

kb-0311:
Hi , I just added the default switch case to time.js . where the date would then be initialized to `date.getUTCDate();` 

Let me know the changes changes to make and whether was there something specific you were looking for the default case.


bl0cknumber:
Hi brother

default format should be unix

duplicate of #534

Issue #345, opened 2022-01-29
datatype.float does not return a float with given precision


Shinigami92:
### Describe the bug

The code indicates that float can take a number that will be used as precision value.
But it returns always a integer number when using `typoef options === 'number'`

### Reproduction

```
faker.seed(42)
faker.datatype.float(6) === 37452
```

### Additional Info

I found this bug while rewriting the datatype tests in #344

ST-DDT:
This behavior might be intentional:

`faker.datatype.number(1000) // 64000`

piotrekn:
After further investigation, I think you're right @ST-DDT - it looks like it's a feature, not a bug. The `precision` option is set, but the name `precision` might be a little unfortunate here. It does not refer only to significant decimal digits, as [this unit test](https://github.com/faker-js/faker/blob/main/test/datatype.spec.ts#L423-L436) exposes.


> faker.datatype.number(1000) // 64000
faker.datatype.float(1000) // 64000

Are you 100% sure about the `faker.datatype.number(1000)` output? I think this should be random number form `[0;1000]`?

```javascript
faker.seed(42)
faker.datatype.number(1000) // 374

faker.seed(42)
faker.datatype.float(1000) // 37000
```

ST-DDT:
> Are you 100% sure about the `faker.datatype.number(1000)` output?

I meant `faker.datatype.number({ precision: 1000 })); // 37000`.
IMO precision is a good word for it (or at least I cannot think of a better one right now).

piotrekn:
>IMO precision is a good word for it (or at least I cannot think of a better one right now).

It resembles precision if represented with `10^n`, but in reality it can take any `Number`. **What is this number in those cases, then? How would you explain what's going on?** I find it hard to put it in simple words and the term "precision" is misused imho :) I also find it very weird how the precision influence the outcome:
```
faker.seed(42);
faker.datatype.float({ precision: 1 }   // 37454

faker.seed(42);
faker.datatype.float({ precision: 9 }   // 37449

faker.seed(42);
faker.datatype.float({ precision: 10 }  // 37450

faker.seed(42);
faker.datatype.float({ precision: 11 }  // 37455
```

Correct me if I'm wrong, but this has little in common to precision as typically the mathematical precision is represented as integer, and refers to the number of significant decimal digits.
[...]

PR #544, opened 2022-02-22
fix: datatype.float does not return a float with given precision

piotrekn:
Linked issue: #345 

Solution:
`float(N)` translates to `float(10^N)`. That apply to `number({precision: N})` as well, as `float` is mostly an alias.

### DIFF
--- a/src/datatype.ts
+++ b/src/datatype.ts
@@ -36,6 +36,7 @@ export class Datatype {
    * faker.datatype.number({ min: 1000000 }) // 431433
    * faker.datatype.number({ max: 100 }) // 42
    * faker.datatype.number({ precision: 0.01 }) // 64246.18
+   * faker.datatype.number({ precision: 2 }) // 64246.18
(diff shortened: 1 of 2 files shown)


piotrekn:
Here is the proposition, tho I'm thinking that maybe `float(N)` should behave as `number(N)` - setting `max`:
https://github.com/faker-js/faker/blob/main/src/datatype.ts#L44-L46

src/datatype.ts
```diff
@@ -59,6 +60,12 @@ export class Datatype {
       options.precision = 1;
     }
 
+    if (options.precision > 1) {
+      options.precision = Number(
+        (0.1 ** options.precision).toFixed(options.precision)
+      );
+    }
```
ST-DDT, line 67: This causes a break in behavior between <=1 and >1

faker.datatype.number({ precision: 1 }) // 64246
faker.datatype.number({ precision: 2 }) // 64246.18

A precision above 1 with the old logic may be useful too:

faker.datatype.number({ precision: 1000 }) // 64000

PR #548, opened 2022-02-23
chore: move faker into own file

Shinigami92:
https://github.com/faker-js/faker/issues/405#issuecomment-1048613530

PR #549, opened 2022-02-23
docs: rewrite tree-shaking part

Shinigami92:


JessicaSachs:
Content is good. Needs some rephrasing. Will check it out in a bit.

JessicaSachs:
I'm following the instructions above. Do you expect

```ts
import { faker } from '@faker-js/faker
```

to be tree-shaken? I wrote this in my other comment, but nothing is tree-shakeable right now.

### Steps to reproduce:

```sh
# clone faker
npm init vite
# (say yes to everything)
cd vite-project
pnpm install
cd ..
pnpm link --dir ./vite-project
```

Modify the JS to import faker.
```js
// within main.js
console.log(faker.locales) // shows all locales
```

Build and preview the dist for vite
```sh
pnpm vite build # See that the gzipped version is 678kb, should be ~200ish kb
```

If you modify the import to `const fakerEn = require('@faker-js/faker/locale/en')`, it's still not tree shaken and the bundle is ~600kb, even though `faker.locales` only shows that `en` is loaded.

We need to rewrite the docs to say that tree-shaking isn't supported right now, or we need to change the way we export the Faker class.

Shinigami92:
@JessicaSachs #548 will solve the tree-shaking problem for locale imports

Shinigami92:
@JessicaSachs could you have a new look into this? Now that #548 is merged?

Shinigami92:
Not sure how to proceed here and no response yet from @JessicaSachs :thinking: 
I think it's better then the current live documentation. So if we want to change something, we can open a commit in the future anyways :shrug:

JessicaSachs:
This is good! I'm sorry I hadn't checked back in.

JessicaSachs (approved):

ST-DDT (approved):

Issue #518, opened 2022-02-19
Unexpected behavior of alphaNumeric with all chars banned


ST-DDT:
### Describe the bug

When all characters are banned, then the method returns a string containing 'undefined' n times.
This violates the expectations of banning all characters.

````ts
faker.random.alphaNumeric(5, {
    bannedChars: '0123456789abcdefghijklmnopqrstuvwxyz'.split('')
}); // 'undefinedundefinedundefinedundefinedundefined'
````

I expect this method to throw an error in that case.

### Reproduction

````ts
it('should be able to handle all possible characters being banned', () => {
  const bannedChars = '0123456789abcdefghijklmnopqrstuvwxyz'.split('');
  const alphaText = faker.random.alphaNumeric(5, {
    bannedChars,
  });

  expect(() =>
    faker.random.alphaNumeric(5, {
      bannedChars,
    })
  ).toThrow();
});
````

### Additional Info

Discovered in https://github.com/faker-js/faker/issues/513#issuecomment-1046017475 / #517

ST-DDT:
On hold till #517 is merged.

ST-DDT:
#517 is merged

xDivisionByZerox:
I will work on this. Please assign me.

xDivisionByZerox:
Could someone please link PR #550 to this?

ST-DDT:
Done

PR #550, opened 2022-02-23
fix: alphaNumeric all chars banned

xDivisionByZerox:
Created in relation to #518.

- remove additional `if` check against `options` argument
- throw if no characters to pick from are available
- activate test case


ST-DDT (approved):

src/random.ts
```diff
@@ -479,11 +479,15 @@ export class Random {
...
-    if (options) {
-      if (options.bannedChars) {
-        charsArray = arrayRemove(charsArray, options.bannedChars);
-      }
+
+    if (options.bannedChars) {
+      charsArray = arrayRemove(charsArray, options.bannedChars);
     }
+
+    if (charsArray.length <= 0) {
```
Shinigami92, line 487: I'm personally not a fan of `length <= 0` because it should never be below 0 🤔 
Are there any reasons for not using `===` here?

src/random.ts
```diff
@@ -479,11 +479,15 @@ export class Random {
...
+      throw new Error('Not able to ban all charcters.');
```
Shinigami92, line 488: Maybe we could add a more specific reason **why** it's not able to do it? So we could potentially increase developer experience.

ST-DDT: Do you have an alternative suggestion for the text?

Maybe: `All characters are banned. Method is unable to generate a string.` ?

src/random.ts
xDivisionByZerox: No there is no this is just my default way to go, since I have some colleges at work that require this syntax, because "you never know what could happen under the hood"...
And since it doesn't change the logic behind this (catch every empty array) I never bothered and simply adapted to this syntax. I can change it to `===` if you require it.

Shinigami92: Yeah, lets change it, I could also try to find an eslint rule, but I think there is no such rule

xDivisionByZerox: There is a [eslint plugin](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-collection-size-mischeck.md) for that. Just so you know and can consider if that would be helpful for consistency.
Since IMO you, as reviewer, should not care abut this kind of thing^^

xDivisionByZerox: Updated in 7a9673adff99000770e2153b8fbaaf0e86d3f5a3

src/random.ts
xDivisionByZerox: Updated in  9bda61edb4a19b74d05b1f4470b90cef87b6e41b

src/random.ts
Shinigami92: > [eslint plugin](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-collection-size-mischeck.md?rgh-link-date=2022-03-12T11%3A11%3A31Z)

Sounds good, but I had some problems in the past with sonar and is was mostly to strict and opinionated :thinking:
Also it had issues with TypeScript. But this was more than 1 year ago and things could have changed.

ST-DDT (approved):

Shinigami92 (approved):

PR #551, opened 2022-02-23
ci: fix coverage patch threshold

Shinigami92:
The threshold should be `100% - target` to not confuse the `coverage/patch` ci job


ST-DDT:
FFR: https://docs.codecov.com/docs/commit-status#threshold

> number
> Allow the coverage to drop by X%, and posting a success status.

PR #552, opened 2022-02-23
chore(deps): add prettier-plugin-organize-imports

Shinigami92:
Add dev depenedency `prettier-plugin-organize-imports`

This will organize all imports with prettier

I will format all files when some other currently higher prio PRs are merged to reduce merge conflicts

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -4,9 +4,7 @@
  * @type {import('prettier').Options}
  */
 module.exports = {
-  plugins: [
-    /* require.resolve("prettier-plugin-organize-imports") */
-  ],
+  plugins: [require.resolve('prettier-plugin-organize-imports')],
   singleQuote: true,
   overrides: [
     {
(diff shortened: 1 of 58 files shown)


Shinigami92:
image

lel, ... I think `codecov/patch` is totally useless :rofl: 

Maybe we should just disable it
@ST-DDT any ideas?

ST-DDT:
The issue here is that the imports are considered code, but are never taken into consideration regarding usage.
IMO this is a bug in the code coverage engine that needs fixing. As for us, we could remove this until it is fixed upstream.
We just have to take this into consideration when looking at PRs that add features.

Shinigami92:
@bcoe Do you know anything about this? Is this a bug of c8?

ST-DDT (approved):

bcoe:
> @bcoe Do you know anything about this? Is this a bug of c8?

My hunch is that it might have to do with how imports are transpiled to JavaScript, vs., an issue with the coverage engine itself. What does the transpiled code look like?

prisis (approved):

PR #555, opened 2022-02-24
fix: decimal javascript calculation

### LINKED ISSUE #331: datatype.number does not always generate numbers with expected precision
### Describe the bug

Due to the limitations of float arithmetics the `precision` param is not always well reflected in the generated output.

### Reproduction

```
> faker.datatype.number({ precision: 0.01 })
64246.18
> faker.datatype.number({ precision: 0.001 })
75598.72
> faker.datatype.number({ precision: 0.0001 })
15719.824
> faker.datatype.number({ precision: 0.00001 })
57268.07910000001
```

### Additional Info

Suggested solution would be to use a better math library offering arbitrary-precision decimal arithmetic, eg. https://www.npmjs.com/package/big.js

franznoel:
## Description

* Fix decimal Javascript calculation using decimal.js library
* Added unit test for precision values 0.00001 and 0.000000001

closes #331

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "c8": "~7.11.0",
     "conventional-changelog-cli": "~2.2.2",
     "cypress": "~9.5.0",
+    "decimal.js": "^10.3.1",
     "esbuild": "~0.14.23",
     "eslint": "~8.9.0",
     "eslint-config-prettier": "~8.4.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,6 +12,7 @@ specifiers:
   c8: ~7.11.0
   conventional-changelog-cli: ~2.2.2
   cypress: ~9.5.0
+  decimal.js: ^10.3.1
   esbuild: ~0.14.23
   eslint: ~8.9.0
   eslint-config-prettier: ~8.4.0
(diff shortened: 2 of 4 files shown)


Shinigami92:
I would like to discuss with the maintainer team internally if dependencies should be really needed for faker or if we want to try to mostly not rely on dependencies to keep the size of faker as low as possible.

Shinigami92:
Yesterday we discussed in a maintainers team meeting that we want to go with an optional peer dependency and a log message that can be somehow toggled on and off
So when someone wants precise correct decimal values, they have the option to additionally install `decimal.js` and then benefit from the "better" variant
And if they don't need it, they will just get the value / variant that is currently available

Shinigami92:
One week passed without response from the author, it may be that this feature will fall out of the v6.1 milestone and will be just handled when it's done

prisis:
Should someone take if over?

Shinigami92:
> Should someone take if over?

I could take it over, but later
I'm a bit interested into how we handle optional peer dependencies, but it doesn't feel like it needs to be tackled immediately, so I focus my time for more important stuff right now

xDivisionByZerox:
I will close this due to inactivity and #779.

PR #559, opened 2022-02-24
docs: update maintainers

ST-DDT:


prisis (approved):

Shinigami92 (commented):
Please have a look into line 7 of this file, there is a date.
So I'm not sure if we want to preserve this document as a history/changelog or if we should update the timestamp and maybe link to the git history or somewhat

@JessicaSachs What are your thoughts about this?

import-brain (approved):

JessicaSachs (approved):
It's fine with me to add your name here because you were there during this entire statement period. For future members I don't think that placing their names here would make a lot of sense and we should get a teams page.

Shinigami92 (approved):

PR #563, opened 2022-02-25
refactor: set seed directly

ST-DDT:
Instead of abusing two constructors to call a method, we could just call the method directly.

### DIFF
--- a/src/datatype.ts
+++ b/src/datatype.ts
@@ -4,14 +4,7 @@ import type { Faker } from '.';
  * Module to generate various primitive values and data types.
  */
 export class Datatype {
-  constructor(private readonly faker: Faker, seed?: number | number[]) {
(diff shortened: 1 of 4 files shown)


ST-DDT:
Should this be in v6.0 or v6.1?

Shinigami92:
Because this is a breaking change, we cannot put it into 6.0

Shinigami92 (approved):

Shinigami92:
Maybe we need another label than `p1-chore` because that has the description that it does not change code behaviour. But this PR does change how the class constructors are called.

Shinigami92 (dismissed):

Shinigami92 (approved):

Shinigami92 (approved):

prisis (approved):

PR #564, opened 2022-02-25
docs: generate docs for fake() and unique()

ST-DDT:


ST-DDT:
Side question: Should I start splitting that script into multiple files? Where should I place them? `scripts/apiDocs/*`?

Shinigami92:
> Side question: Should I start splitting that script into multiple files? Where should I place them? `scripts/apiDocs/*`?

Yes, we don't have a strict current way of structure for scripts folder. Feel free to define one :slightly_smiling_face: 
If helpful, do it directly in this PR

scripts/apidoc/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { resolve } from 'path';
```
Shinigami92, line 1:
```suggestion
import { resolve } from 'node:path';
```

Shinigami92 (approved):

prisis (approved):

PR #570, opened 2022-02-26
docs: document options params in api docs

ST-DDT:
# Example

grafik

# Note

Only merge after #564 
Currently this does not work for union parameters e.g. `number | { ...}` https://github.com/TypeStrong/typedoc/issues/1876


Shinigami92 (dismissed):

prisis (dismissed):

ST-DDT:
Updated

prisis (approved):

scripts/apidoc/signature.ts
```diff
@@ -1,5 +1,5 @@
 import sanitizeHtml from 'sanitize-html';
-import type * as TypeDoc from 'typedoc';
+import * as TypeDoc from 'typedoc';
```
prisis, line 2: Just a question, why you removed the type?

ST-DDT: https://github.com/faker-js/faker/blob/a741d1f2a9b2a658512d4b99074d9695905c4b69/scripts/apidoc/signature.ts#L193-L195

Because I now have to access an actual value.

Shinigami92: Is it possible to split the imports into types and actual use?

ST-DDT: Like this?
````ts
import type {
  Comment,
  ParameterReflection,
  Reflection,
  SignatureReflection,
  SomeType,
} from 'typedoc';
import { ReflectionFlag, ReflectionKind } from 'typedoc';
````

Shinigami92: Yeah, I think so, the benefit of this, is that it uses tree-shaking
Not sure if this is helpful in a script, but maybe it could speed up loading by nanoseconds :P

prisis (approved):

griest024 (approved):

PR #574, opened 2022-02-26
docs: add test for api docs generation

ST-DDT:
This PR adds a test that ensures, that the api docs generation does not break (at least on the data side).

**Note:**

To be merged after #564


ST-DDT:
Updated

Shinigami92 (approved):

test/scripts/apidoc/signature.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  describe('analyzeSignature()', () => {
+    const actuals = {};
+
+    it('expected and actual methods are equal', () => {
+      expect(Object.keys(methods).sort()).toEqual(Object.keys(expected).sort());
+    });
+
+    it.each(Object.keys(expected))('%s', (name) => {
+      const method = methods[name];
+      const actual = analyzeSignature(method.signatures[0], null, method.name);
```
griest024, line 39: Methods can have multiple signatures? I guess this is to support overloading? Are we just going to ignore that because we don't use it?

ST-DDT: Well, that is what we do now. As we don't really use multiple signatures.
We might want to change it to the last signature as it might be more useful.

Here is the only method that does use multiple signatures at this point in time:
https://github.com/faker-js/faker/blob/60d3cc519cd2cab061a23fbe09164d8b59e177ff/src/random.ts#L174-L185
And it does kind of show up wrong:
https://fakerjs.dev/api/random.html#objectElement
This PR is here to prevent me accidentally breaking the docs generation, not fixing issues that it currently has.
I will create a new issue for that.

Shinigami92 (approved):

prisis (approved):

PR #576, opened 2022-02-26
refactor: date accept number input

### LINKED ISSUE #394: Accept numbers as params for Date methods

https://github.com/faker-js/faker/pull/200#issuecomment-1024802219

### Suggested solution

Accept timestamps

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
Created in relation to #394.

I changed the way dates are created from the `Date.parse` method to the `Date.constructor` since it uses that under the hood ([MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters)). This also allows adding `Date` arguments in the future (or even this PR).
The default date is assigned if the provided input creates an invalid date instance (if the value of the date is `NaN`). This was not checked till now whatsoever.


src/date.ts
```diff
@@ -26,11 +26,12 @@ export class _Date {
...
    */
-  past(years?: number, refDate?: string): Date {
-    let date = new Date();
-    if (typeof refDate !== 'undefined') {
-      date = new Date(Date.parse(refDate));
+  past(years?: number, refDate?: string | number): Date {
+    let date = new Date(refDate);
+    if (isNaN(date.valueOf())) {
+      date = new Date();
     }
```
ST-DDT, line 35: This logic seems to be repeated multiple times. Maybe extract it into its own helper method.

ST-DDT:
grafik

I assume the results are fluctuating by around ~1s.
For some strange reason, the new Date(string) method does discard the ms from the parsed date.
That's why [#200](https://github.com/faker-js/faker/pull/200/files#diff-cdc74ab5a2d5c1414321a9e339b60e2e79e1018a99e545d87653e3e0d972fed6R189) explicitly uses a separate method.

ST-DDT (changes requested):
Please add tests for `number` inputs.

xDivisionByZerox:
> For some strange reason, the new Date(string) method does discard the ms from the parsed date (sometimes).
That's why [#200](https://github.com/faker-js/faker/pull/200/files#diff-cdc74ab5a2d5c1414321a9e339b60e2e79e1018a99e545d87653e3e0d972fed6R189) explicitly uses new Date(Date.parse(string)).

That doesn't make sense. The date constructor under the hood uses the Date.parse method for string inputs. You can even verify that both methods get the same result by:
```javascript
const date1 = new Date('2020-01-01T00:00:00.000Z');
const date2 = Date.parse('2020-01-01T00:00:00.000Z');

console.log(date1); // 2020-01-01T00:00:00.000Z
console.log(date2); // 1577836800000
console.log(date1.valueOf() === date2); // true
```

src/date.ts
xDivisionByZerox: Fixed after the merge with #200

xDivisionByZerox:
Can you add a label or something that this should not get merged before #200?

src/date.ts
```diff
@@ -111,22 +108,18 @@ export class _Date {
    * faker.date.betweens('2020-01-01T00:00:00.000Z', '2030-01-01T00:00:00.000Z', 2)
    * // [ 2023-05-02T16:00:00.000Z, 2026-09-01T08:00:00.000Z ]
    */
-  betweens(from: string, to: string, num?: number): Date[] {
+  betweens(from: string | Date, to: string | Date, num?: number): Date[] {
```
ST-DDT, line 111: What about numbers here?

xDivisionByZerox: I'm sorry. Missed that during the merge.

xDivisionByZerox: fixed in 4e1d519

src/date.ts
```diff
@@ -1,5 +1,7 @@
 import type { Faker } from '.';
 
+type DateConstructor = number | string | Date;
```
Shinigami92, line 3: Maybe it's better to use native TypeScript typing system to get the params of DateConstructor

```ts
type DateConstructorParameters = ConstructorParameters<DateConstructor>[0]; // string | number | Date
```

Also using `DateConstructor` as type name is confusing with the real native type `DateConstructor`

Shinigami92 (changes requested):

Shinigami92:
> That doesn't make sense. The date constructor under the hood uses the Date.parse method for string inputs. You can even verify that both methods get the same result by:
> 
> ```js
> const date1 = new Date('2020-01-01T00:00:00.000Z');
> const date2 = Date.parse('2020-01-01T00:00:00.000Z');
> 
> console.log(date1); // 2020-01-01T00:00:00.000Z
> console.log(date2); // 1577836800000
> console.log(date1.valueOf() === date2); // true
> ```

You are checking that the milliseconds are removed but using `000` for milliseconds part?

You should compare e.g.

```ts
const date1 = new Date('2020-01-01T00:00:00.123Z');
const date2 = Date.parse('2020-01-01T00:00:00.123Z');

console.log(date1); // 2020-01-01T00:00:00.123Z
console.log(date2); // 1577836800123
console.log(date1.valueOf() === date2); // false
```

src/date.ts
xDivisionByZerox: Cool feature. Didn't know about the `ConstructorParameters` utility type.
Fixed in 90a52271c725a4083472ace430b05e50ee5614ac

xDivisionByZerox:
> Please add tests for `number` inputs.

Added in a709e9d

ST-DDT (commented):
For now, this looks good to me. I will re-review it once #200 is merged to main.

src/date.ts
```diff
@@ -256,4 +250,13 @@ export class _Date {
 
     return this.faker.random.arrayElement(source);
   }
+
+  private toDate(date?: DateConstructorParameters): Date {
```
Shinigami92, line 254: Due to how we handle our "module" classes, it's not possible to define `private` methods in them
So you need to extract `toDate` out of the class and define it above as a raw function (that is not exported!)
An example for that can be found here: https://github.com/faker-js/faker/pull/644/files#diff-f2b24a1df2ffea4071c80b9948ecf7ed59397468799db7ee1ec376b3c9740b06R17

Shinigami92 (changes requested):

src/date.ts
xDivisionByZerox: Done in a14f072

Shinigami92 (dismissed):

Shinigami92:
@xDivisionByZerox we merged #200, please rebase this PR and re-evaluate if and what can still be merged from this PR

xDivisionByZerox:
I'm going to close this, make another branch based on the current main, and will open another PR since I'm somewhat of a wizard and always completely destroy the git history.

xDivisionByZerox:
I opened #670 to deprecate this PR.

PR #580, opened 2022-02-28
fix: force passed locales into faker constructor

vith:
I saw that there's a `Faker` class being exported now, so I tried to use it, but it doesn't seem to work.

For example, when trying to do `faker = new Faker(); faker.name.firstName()` I get: 
```
TypeError: Cannot read properties of undefined (reading 'name')
```

I didn't see any examples of using this constructor in the docs, so it's possible I'm using it wrong. Anyway, I wrote these tests to check if it was a problem in my own project, but they fail as well in the same way.


Shinigami92:
I think it fails, because you don't provide any locales.

Maybe we need to change the type on your side to force to provide a `locales` :thinking:

Shinigami92:
Uhm... tests are green? :eyes:

ST-DDT:
I assume since it probably returns  undefined` in the middle the fallback of  `['a', 'b', 'c']` applies and thus passes this trivial test.
Maybe change it to length > 1 (as there shouldn't be any one char names in our data).

vith:
The tests are green just because I defined them with `it.fails`. If you want I can remove that so they actually cause the suite to fail now.

I see what you mean now about not providing any locales, i.e. I need replicate something like this:

```js
import { Faker } from '../faker';
import sv from '../locales/sv';

const faker = new Faker({
  locale: 'sv',
  localeFallback: 'en',
  locales: {
    sv,
    en,
  },
});

export = faker;
```

But I haven't found any way to actually import the locale data externally. I think the `exports` field of `package.json` only lets me import the pre-instantiated per-locale singletons like the one I pasted above from `src/locale/sv.ts`?

ST-DDT:
> The tests are green just because I defined them with `it.fails`. If you want I can remove that so they actually cause the suite to fail now.

Yes, that would make it more obvious for anyone looking at the tests.

> But I haven't found any way to actually import the locale data externally. I think the `exports` field of `package.json` only lets me import the pre-instantiated per-locale singletons like the one I pasted above from `src/locale/sv.ts`?

Yes, currently there isn't because it wasn't in 5.x. This is planned for a release after 6.0 when we actually start adding fixes and features.

vith:
> Yes, that would make it more obvious for anyone looking at the tests.

Done.

> Yes, currently there isn't because it wasn't in 5.x. This is planned for a release after 6.0 when we actually start adding fixes and features.

Ah, okay. I wonder if the Faker class shouldn't be exported from the package in 6.0 then, if the locale data it needs won't be?

I tried to use the class because I got it as a code completion via the TypeScript Language Server in VSCode.

Not sure if there's anything useful to make out of the tests in this PR then. I guess making the locales field required in the constructor would help like @Shinigami92 said, and/or a better error message when locale data is missing.

Shinigami92:
> Could you adjust the constructor to not have a default and require the locales?

@vith Would you like to open a new PR to do this or maybe just use this PR?

ST-DDT:
> > Could you adjust the constructor to not have a default and require the locales?
> 
> @vith Would you like to open a new PR to do this or maybe just use this PR?

I recommend using this one, as both parts fix+test belong together anyway.

Shinigami92:
I will jump into this PR now and fix the constructor

src/faker.ts
```diff
@@ -83,8 +83,18 @@ export class Faker {
   readonly vehicle: Vehicle = new Vehicle(this);
   readonly word: Word = new Word(this);
 
-  constructor(opts: FakerOptions = {}) {
-    this.locales = this.locales || opts.locales || {};
+  constructor(opts: FakerOptions) {
+    if (!opts) {
+      throw new Error('Options with at least one locale must be provided');
```
ST-DDT, line 88: Since there is also a `locale` parameter, this should refer to "one entry in locales".

ST-DDT (changes requested):

ST-DDT (dismissed):

ST-DDT:
This needs to be rebased

ST-DDT (approved):

src/faker.ts
```diff
@@ -81,8 +81,20 @@ export class Faker {
...
   readonly word: Word = new Word(this);
 
-  constructor(opts: FakerOptions = {}) {
-    this.locales = this.locales || opts.locales || {};
+  constructor(opts: FakerOptions) {
+    if (!opts) {
+      throw new Error(
+        'Options with at least one entry in locales must be provided'
+      );
+    }
```
xDivisionByZerox, line 89: For me, this seems like unreachable code, since the constructor type was changed.

xDivisionByZerox (approved):

PR #589, opened 2022-03-01
refactor(random.arrayElement): remove default parameter

### LINKED ISSUE #581: Remove default value from `helpers.arrayElement()`
https://github.com/faker-js/faker/blob/4eef328a90657ce64a5210862ad372ecd97d7e7a/src/random.ts#L106-L108

This causes issues with missing data, that then gets magically replaced with these defaults, which results in unexpected behavior.
This change could be considered breaking, but IMO calling that method without an array to choose from is a bug itself.
Thus why I consider this a bug, that needs fixing more than retaining backwards compatibility.

We observed this method misbehaving in our tests before.

**Alternatives considered**

Create a new method and deprecate the old one.

**Additional Information**

Related to #219

xDivisionByZerox:
Fixes #581.

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -99,20 +99,18 @@ export class Random {
    * Returns random element from the given array.
    *
    * @template T The type of the entries to pick from.
-   * @param array Array to pick the value from. Defaults to `['a', 'b', 'c']`.
+   * @param array Array to pick the value from.
    *
    * @example
-   * faker.random.arrayElement() // 'b'
    * faker.random.arrayElement(['cat', 'dog', 'mouse']) // 'dog'
+   * faker.random.arrayElement([0, 1, 2, 3]) // 3
    */
-  arrayElement<T = string>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
-    const index =
-      array.length > 1
-        ? this.faker.datatype.number({ max: array.length - 1 })
-        : 0;
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array == null || array.length === 0) {
+      throw new Error('Cannot get random element from empty array.');
+    }
 
+    const index = this.faker.datatype.number({ max: array.length - 1 });
     return array[index];
   }
 
@@ -120,20 +118,20 @@ export class Random {
    * Returns a subset with random elements of the given array in random order.
    *
    * @template T The type of the entries to pick from.
-   * @param array Array to pick the value from. Defaults to `['a', 'b', 'c']`.
+   * @param array Array to pick the value from.
    * @param count Number of elements to pick.
    *    When not provided, random number of elements will be picked.
    *    When value exceeds array boundaries, it will be limited to stay inside.
    *
    * @example
-   * faker.random.arrayElements() // ['b', 'c']
    * faker.random.arrayElements(['cat', 'dog', 'mouse']) // ['mouse', 'cat']
    * faker.random.arrayElements([1, 2, 3, 4, 5], 2) // [4, 2]
    */
-  arrayElements<T>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>,
-    count?: number
-  ): T[] {
+  arrayElements<T>(array: ReadonlyArray<T>, count?: number): T[] {
+    if (array == null || array.length === 0) {
+      throw new Error('Cannot get random element from empty array.');
+    }
+
     if (typeof count !== 'number') {
       count = this.faker.datatype.number({ min: 1, max: array.length });
     } else if (count > array.length) {
--- a/src/random.ts
+++ b/src/random.ts
@@ -129,7 +129,7 @@ export class Random {
    */
   arrayElements<T>(array: ReadonlyArray<T>, count?: number): T[] {
     if (array == null || array.length === 0) {
-      throw new Error('Cannot get random element from empty array.');
+      throw new FakerError('Cannot get random element from empty array.');
     }
 
     if (typeof count !== 'number') {
--- a/src/random.ts
+++ b/src/random.ts
@@ -107,7 +107,7 @@ export class Random {
    */
   arrayElement<T>(array: ReadonlyArray<T>): T {
     if (array == null || array.length === 0) {
-      throw new Error('Cannot get random element from empty array.');
+      throw new FakerError('Cannot get random element from empty array.');
     }
 
     const index = this.faker.datatype.number({ max: array.length - 1 });
--- a/src/word.ts
+++ b/src/word.ts
@@ -1,5 +1,17 @@
 import type { Faker } from '.';
 
+function selectWords(length?: number, words: string[] = []): string[] {
+  let wordList = words;
+  if (length) {
+    wordList = words.filter((word) => word.length === length);
+    if (wordList.length === 0) {
+      wordList = words;
+    }
+  }
+
+  return wordList;
+}
+
 /**
  * Module to return various types of words.
  */
@@ -25,7 +37,9 @@ export class Word {
    * faker.word.adjective(100) // 'complete'
    */
   adjective(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.adjective);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.adjective)
+    );
   }
 
   /**
@@ -39,7 +53,9 @@ export class Word {
    * faker.word.adverb(100) // 'sadly'
    */
   adverb(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.adverb);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.adverb)
+    );
   }
 
   /**
@@ -53,7 +69,9 @@ export class Word {
    * faker.word.conjunction(100) // 'as long as'
    */
   conjunction(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.conjunction);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.conjunction)
+    );
   }
 
   /**
@@ -67,7 +85,9 @@ export class Word {
    * faker.word.interjection(100) // 'yowza'
    */
   interjection(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.interjection);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.interjection)
+    );
   }
 
   /**
@@ -81,7 +101,9 @@ export class Word {
    * faker.word.noun(100) // 'care'
    */
   noun(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.noun);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.noun)
+    );
   }
 
   /**
@@ -95,7 +117,9 @@ export class Word {
    * faker.word.preposition(100) // 'an'
    */
   preposition(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.preposition);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.preposition)
+    );
   }
 
   /**
@@ -109,18 +133,8 @@ export class Word {
    * faker.word.verb(100) // 'mess'
    */
   verb(length?: number): string {
-    return this.selectOneWord(length, this.faker.definitions.word.verb);
-  }
-
-  private selectOneWord(length?: number, words: string[] = []): string {
-    let wordList = words;
-    if (length) {
-      wordList = words.filter((word) => word.length === length);
-      if (wordList.length <= 0) {
-        wordList = words;
-      }
-    }
-
-    return this.faker.random.arrayElement(wordList);
+    return this.faker.random.arrayElement(
+      selectWords(length, this.faker.definitions.word.verb)
+    );
   }
 }
--- a/src/word.ts
+++ b/src/word.ts
@@ -25,19 +25,7 @@ export class Word {
    * faker.word.adjective(100) // 'complete'
    */
   adjective(length?: number): string {
-    let wordList = this.faker.definitions.word.adjective;
-    if (length) {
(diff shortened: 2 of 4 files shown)


xDivisionByZerox:
Should this change also apply to `arrayElements`?

xDivisionByZerox:
Ok cool of course I forgot to look at the test cases on such a method.
Will provide updates for test files in near future.

xDivisionByZerox:
`address.streetPrefix` implementation:
https://github.com/faker-js/faker/blob/60d3cc519cd2cab061a23fbe09164d8b59e177ff/src/address.ts#L234-L238

Meaning that if a local doesn't export a `street_prefix` it will throw an error (example 'de'). 
To fix this we would need to make all definition maps required (since its not only `address.streetPrefix` that breaks from this).

xDivisionByZerox:
How to handle this?
@ST-DDT @Shinigami92

ST-DDT:
> Meaning that if a local doesn't export a `street_prefix` it will throw an error (example 'de').

It will only fail, if both the selected and the fallback locale don't have that value.
IMO throwing an error is still better than returning abc randomly.
That way you at least notice, that it's not supported.

src/random.ts
```diff
@@ -98,15 +98,17 @@ export class Random {
...
    * @example
-   * faker.random.arrayElement() // 'b'
    * faker.random.arrayElement(['cat', 'dog', 'mouse']) // 'dog'
+   * faker.random.arrayElement([0, 1, 2, 3]) // 3
    */
-  arrayElement<T = string>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array.length <= 0) {
```
ST-DDT, line 108: This needs to handle undefined arrays as well. When is the array length below zero?

Maybe like this?

```suggestion
    if (typeof array === 'undefined' || !array.length) {
```

Shinigami92: I would even go with `array == null || array.length === 0`

xDivisionByZerox: Handled both `null` and `undefined` in f9599b6aa35cac02f2dd9e859ae930218bc1dbd1

src/random.ts
```diff
@@ -98,15 +98,17 @@ export class Random {
...
    * @example
-   * faker.random.arrayElement() // 'b'
    * faker.random.arrayElement(['cat', 'dog', 'mouse']) // 'dog'
+   * faker.random.arrayElement([0, 1, 2, 3]) // 3
    */
-  arrayElement<T = string>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (typeof array === 'undefined' || array === null || array.length <= 0) {
```
Shinigami92, line 108: This can be shorten to
```suggestion
    if (array == null || array.length === 0) {
```

This does exactly the same but is much easier to read and shorter

xDivisionByZerox: I would argue about being easier to read because you use a loosely type check `==` for bullish values and a strong type check for the array length. Furthermore, I'm a friend of abandoning the loose type check operator since it brings consistency into a project (and somewhat safety over only using `==` as a consistent check).

When on that point? Why do you use `null` as compare value instead of the more often used `undefined`? Would also do the same.

xDivisionByZerox: That said, I'm not saying you are wrong. Just want to give you my opinion on that toppic.

Shinigami92: Using ` == null ` is a really common pattern in js.
I think there is even a eslint rule for that 👀 
Yeah https://eslint.org/docs/rules/eqeqeq
We can configure it like my suggested way

It's a special case when you want to compare something is `null` or `undefined`, then use `== null`

xDivisionByZerox: The `eqeqeq` eslint rule is exaclty what im referring to as well. Since it disallows the loose type check operation `==` and requires you to use the strong one `===`.
But I guess you referring to the `'smart'` configuration option instead of `'always'`.

ST-DDT: Having an eslint rule for that is probably good for consistency.

xDivisionByZerox: I might be biased but I rarely see `== null`. `== undefined` is what I come across all the time.

Shinigami92 (dismissed):

ST-DDT:
Please also update `random.arrayElements()` accordingly + fix the tests.

E.g. by changing:

````ts
return (
      this.faker.random.arrayElement(wordList) ||
      this.faker.random.arrayElement(this.faker.definitions.word.adjective)
    );
````

to

````ts
return (
      wordList.length > 0 ? this.faker.random.arrayElement(wordList) : this.faker.random.arrayElement(this.faker.definitions.word.adjective)
    );
````

Maybe add a check that checks the array before calling arrayElement and throws an error that contains a the method/definition that caused it.

----------

For `word.ts` it might be useful to extract the entire logic into a separate private method:

````ts
adjective(length?: number): string {
    return selectOneWord(length, this.faker.definitions.word.adjective);
}

private selectOneWord(length?: number, words: string[]): string {
  let wordList = words;
    if (length) {
      wordList = words.filter((word) => word.length == length);
      if (!wordList.length) {
        wordList = words;
      }
    }

    return this.faker.random.arrayElement(words);
}
````

**Comment by xDivisionByZerox** [comment]
> Maybe add a check that checks the array before calling arrayElement and throws an error that contains a the method/definition that caused it.

Can I expect intern-provided arrays (like `this.faker.definitions.word.adjective`) to be provided correctly or check them as well?

**Comment by ST-DDT** [comment]
Theoretically we don't have an `intern-provided` array, `en` is just a fallback, but you can specify a different one (and might be sensible to do so, if you don't a language without latin characters).
Maybe we should use special syntax for the definitions to make this implicitly?

- `this.faker.definitions.word.adjective` throw if the entry is missing
- `this.faker.extensions.word.adjective` return undefined

(both would use the same "data source")

**Comment by Shinigami92** [comment]
@xDivisionByZerox In preparation of starting to review some v6.1 PRs, could you rebase this one and fix conflicts?

**Comment by xDivisionByZerox** [comment]
Rebase is done and conflicts resolved. Haven't had time for the test and probably won't have this week either.

File: src/word.ts [hunk 3]
```diff
@@ -153,18 +81,18 @@ export class Word {
...
+    return this.selectOneWord(length, this.faker.definitions.word.verb);
+  }
+
+  private selectOneWord(length?: number, words: string[] = []): string {
+    let wordList = words;
     if (length) {
-      wordList = this.faker.definitions.word.verb.filter(
-        (word: string) => word.length === length
-      );
+      wordList = words.filter((word) => word.length == length);
```
Shinigami92, line 90:
```suggestion
      wordList = words.filter((word) => word.length === length);
```

Shinigami92 (changes requested):
Due to how we handle our "module" classes, it's not possible to define private methods in them
So you need to extract `selectOneWord` out of the class and define it above as a raw function (that is not exported!)
An example for that can be found here: https://github.com/faker-js/faker/pull/644/files#diff-f2b24a1df2ffea4071c80b9948ecf7ed59397468799db7ee1ec376b3c9740b06R17

Also this PR needs a rebase 🙂

xDivisionByZerox:
@Shinigami92 why do my rebases always put the changes from the branch I rebase to the end? Isn't it supposed to treat the rebased branch as a new source and not as a change?

Shinigami92:
> @Shinigami92 why do my rebases always put the changes from the branch I rebase to the end? Isn't it supposed to treat the rebased branch as a new source and not as a change?

Seems you definitely do something wrong, but a rebase puts the changes to the end, that's correct
But this PR is now messed up :(
Maybe I could show you next few days how you can rebase correctly

Shinigami92:
@xDivisionByZerox I hopefully restored this branch correctly
Please have a look and check if everything is normal again

For your local machine you now need to switch into this branch and then just run `git pull --rebase`

Additionally I already moved the private function out of the module and due to the this binding, I needed to alter the behavior a bit

Shinigami92:
Oh damn, I hope I did not messed up myself now :slightly_frowning_face: 
There is 1 lint error and 366 failing tests :eyes: 
After I did the rebase there were no test file changes

Shinigami92:
@xDivisionByZerox This PR is one of the last PRs that I would forcefully like to have in v6.1 milestone. Could you bring this to the finish goal or should I take it over myself?

src/random.ts
```diff
@@ -99,41 +99,39 @@ export class Random {
...
    * faker.random.arrayElements(['cat', 'dog', 'mouse']) // ['mouse', 'cat']
    * faker.random.arrayElements([1, 2, 3, 4, 5], 2) // [4, 2]
    */
-  arrayElements<T>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>,
-    count?: number
-  ): T[] {
+  arrayElements<T>(array: ReadonlyArray<T>, count?: number): T[] {
+    if (array == null || array.length === 0) {
+      throw new Error('Cannot get random element from empty array.');
```
Shinigami92, line 132: `FakerError` is already merged 😉

src/random.ts
```diff
@@ -99,41 +99,39 @@ export class Random {
...
-  arrayElement<T = string>(
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
-    const index =
-      array.length > 1
-        ? this.faker.datatype.number({ max: array.length - 1 })
-        : 0;
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array == null || array.length === 0) {
+      throw new Error('Cannot get random element from empty array.');
```
Shinigami92, line 110: `FakerError` is already merged 😉

src/random.ts
xDivisionByZerox: done in c0a28a2f43e5bb26277c9fdc4d151b8bafa29bc7

xDivisionByZerox:
As I said [here](https://github.com/faker-js/faker/pull/589#issuecomment-1057412956)
address.streetPrefix breaks the tests, since there are no locales in `en` for street_prefix

Shinigami92:
@xDivisionByZerox So now we just need to fix all the failing tests (?)

xDivisionByZerox:
Not really since the problem is the same I stated 1 month ago.
Or should I change the `address.streetPrefix` to expect an error?

Shinigami92:
> `address.streetPrefix` breaks our tests, since there are no locales in `en` for street_prefix

So yeah, but as I understand this PR: This is THE bug we want to fix!
So do we need to throw errors in this case?
@ST-DDT I would like you to decide it how to proceed

xDivisionByZerox:
I'll change the test to expect an Error and add a comment to expect a real value when locales were included.

xDivisionByZerox:
@ST-DDT Can you give insight into how the `BROKEN_LOCALE_METHODS` constant in `all_functional.spec.ts` works? Do I need to add the modules that don't work with `en` and provide an alternative local or does this object contain all functions not working in a specific local?

damienwebdev:
My opinion is that there's a no-trivial chance that this is breaking for someone. If we release this as `v6.1.2` or `v6.2` we would be breaking the semver contract.

ST-DDT:
> My opinion is that there's a non-trivial chance that this is breaking for someone. If we release this as `v6.1.2` or `v6.2` we would be breaking the semver contract.

I agree. We should move this to v7.

ST-DDT:
> @ST-DDT Can you give insight into how the `BROKEN_LOCALE_METHODS` constant in `all_functional.spec.ts` works? Do I need to add the modules that don't work with `en` and provide an alternative local or does this object contain all functions not working in a specific local?

Yes, if effectively contains all tests that are disabled for a certain locale.

xDivisionByZerox:
> > @ST-DDT Can you give insight into how the `BROKEN_LOCALE_METHODS` constant in `all_functional.spec.ts` works? Do I need to add the modules that don't work with `en` and provide an alternative local or does this object contain all functions not working in a specific local?
> 
> Yes, if effectively contains all tests that are disabled for a certain locale.

Well then this doesn't work I guess. If you have a look at the log files from node-16, ubuntu-latest line 22344:
```
2022-04-01T15:28:45.2251825Z  [31m×[39m test/all_functional.spec.ts[2m > [22mfaker.fake functional tests[2m > [22mru[2m > [22mname[2m > [22mprefix()
2022-04-01T15:28:45.2252117Z [31m   → Cannot get random element from empty array.[39m
```

Even tho we have in all_functions:
```
const BROKEN_LOCALE_METHODS = {
    ...,
    name: {
        prefix: [..., 'ru'],
        ...,
    },
    ...,
}
```

xDivisionByZerox:
Or can I remove this now since the `it.fails` doesn't apply anymore since the function will now throw an error? But for my understanding `it.fails` should be the right why to go when a function throws an error :/

ST-DDT:
The faker.fake method currently does not have an exclude handling, as it does not fail.

ST-DDT:
@xDivisionByZerox Can you split the word improvements to its own PR?
AFAICT these changes are independent and could be merged earlier.

xDivisionByZerox:
> @xDivisionByZerox Can you split the word improvements to its own PR? AFAICT these changes are independent and could be merged earlier.

Well [you yourself suggested this change](https://github.com/faker-js/faker/pull/589#issuecomment-1059073693) in this PR. This logic is not required to change if this PR is not in place. So a separate PR should IMO be dismissed because it doesn't provide any meaningful architectural refactoring/performance boosting.

xDivisionByZerox:
But yeah, I can provide a separate PR if you wish to merge this in 6.3.0.

ST-DDT:
> > @xDivisionByZerox Can you split the word improvements to its own PR? AFAICT these changes are independent and could be merged earlier.
> 
> Well [you yourself suggested this change](https://github.com/faker-js/faker/pull/589#issuecomment-1059073693) in this PR. This logic is not required to change if this PR is not in place. So a separate PR should IMO be dismissed because it doesn't provide any meaningful architectural refactoring/performance boosting.

It simplifies the code by removing duplicate code. IMO that is worth enough.
(I plan to extend the length to a range at some point, so any simplifications are welcome.)

xDivisionByZerox:
I will close this PR due to MASSIV changes in the file structure. I will provide a new PR with changes discussed in this PR.

PR #613, opened 2022-03-11
chore: fix some lint warnings

ST-DDT:


prisis (dismissed):

src/helpers.ts
```diff
@@ -342,7 +342,7 @@ export class Helpers {
     const RANGE_REP_REG = /(.)\{(\d+)\,(\d+)\}/;
     const REP_REG = /(.)\{(\d+)\}/;
     const RANGE_REG = /\[(\d+)\-(\d+)\]/;
-    let min, max, tmp, repetitions;
+    let min, max, tmp, repetitions: number;
```
Shinigami92, line 345: We should split this to separate lines

test/fake.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('fake', () => {
 
     it('does not allow undefined parameters', () => {
       expect(() =>
-        // @ts-expect-error
+        // @ts-expect-error The parameter is required
```
Shinigami92, line 33:
```suggestion
        // @ts-expect-error: The parameter is required
```

src/helpers.ts
ST-DDT: Do you have time to fix these?

src/helpers.ts
```diff
@@ -434,9 +437,9 @@ export class Helpers {
    * faker.helpers.uniqueArray(faker.definitions.name.first_name, 6)
    * faker.helpers.uniqueArray(["Hello", "World", "Goodbye"], 2)
    */
-  uniqueArray<T>(source: T[] | (() => T), length: number): T[] {
+  uniqueArray<T>(source: readonly T[] | (() => T), length: number): T[] {
     if (Array.isArray(source)) {
-      const set = new Set(source);
+      const set = new Set<T>(source);
       const array = Array.from(set);
       return this.faker.helpers.shuffle(array).splice(0, length);
```
Shinigami92, line 444: Just a note for a later PR, IMO we should use `slice` instead of `splice` here :thinking:

prisis (approved):

PR #616, opened 2022-03-12
feat: Add MongoDB ObjectId generation

nhammond101:
Adds a function to generate MongoDB [ObjectIds](https://docs.mongodb.com/manual/reference/bson-types/#std-label-objectid)

The function aims to be deliberately simple so any combination of `a-z` and `0-9` is generated as a valid ObjectId, instead of implementing ObjectIds according to the official spec

Futhermore, I was unsure where to add the function given it's a datatype for a database, so happy to refactor/move if required.


Shinigami92:
Would this fit better into `/src/database.ts`? I think we should move it to that module.

Shinigami92:
Not related to this PR, but IMO we should try to prepare that the datatype module only defines functions for `typeof` "return" values.
So only `string`, `number`, `boolean` and so on. But this would be something for `v7+`

nhammond101:
@Shinigami92 agree on the moving to database, and have done so now.

Is there anything I can add to the current PR to facilitate the future `typeof` improvements?

src/database.ts
```diff
@@ -59,4 +59,53 @@ export class Database {
...
+  /**
+   * Returns an [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string
+   *
+   * @param value When provided, and is a valid objectId, this method simply returns it as is.
+   *
+   * @example
+   * faker.database.objectId() // e175cac316a79afdd0ad3afb
+   * faker.database.objectId('6228fb6ec55ed495fc1e5f80') // 6228fb6ec55ed495fc1e5f80
+   */
+  objectId(value?: string): string {
```
Shinigami92, line 72: What's the benefit of having this param? I think currently it doesn't provide any value to faker and should just be checked/handled outside of the usage of faker.

Shinigami92 (changes requested):

Shinigami92:
> Is there anything I can add to the current PR to facilitate the future `typeof` improvements?

Mh, currently I don't think so. You can always join our discord if you haven't already, so we can communicate better there.
But for now, we would tackle all the v6.1 stuff first anyways.
We expect to release 6.0.0 on monday evening.

src/database.ts
nhammond101: not a great deal upon reflection.  i've removed it now.

src/database.ts
```diff
@@ -59,4 +59,46 @@ export class Database {
...
     );
   }
+
+  /**
+   * Returns an [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string
+   *
+   * @example
+   * faker.database.objectId() // e175cac316a79afdd0ad3afb
+   */
+  objectId(): string {
```
ST-DDT, line 69: Maybe prefix it with mongo(DB), because object id might be used in other contexts as well.

src/database.ts
```diff
@@ -59,4 +59,46 @@ export class Database {
...
+    const length = 24;
+
+    // a-f0-9
+    const charCodeOptions = [
```
ST-DDT, line 73: Why don't you use [faker.random.hexaDecimal](https://fakerjs.dev/api/random.html#hexaDecimal)

src/database.ts
nhammond101: done

src/database.ts
nhammond101: good point, i had avoided it thus far due to the `0x` prefix.  
i've refactored to use hexadecimal and it certainly shortens the function code!

ST-DDT (approved):

src/database.ts
Shinigami92: Oh yeah, good point, we already thought about somewhere else that we want to remove the `0x` part from that function.
But this will be in v7, cause it would introduce a "big" breaking change.

ST-DDT: Maybe we can start with a prefix parameter in v6.2?

Shinigami92: Oh yeah that could work :thinking: Good idea :+1: 
So we could add a `hexaDecimal(length = 1, prefix = '0x')`

damienwebdev:
This feels more like a plugin than a method of the main library. @Shinigami92 @ST-DDT have you seen a way to add custom user-defined methods to the Faker API?

ST-DDT:
> have you seen a way to add custom user-defined methods to the Faker API?

This is possible, but it requires casting or subclassing the faker instance. While I agree this kind of custom/specific, moving this to a plugin seems overengeneered to me. We can consider this again, if we have more of these.
We could nest these in their own submodule if you want.
`faker.database.mongodb.objectId()`

Shinigami92:
I'm with @ST-DDT, plugins would be over engineered.
We also already had the discussion about sub module, but IMO we can move it into that if really needed.

src/database.ts
```diff
@@ -59,4 +59,15 @@ export class Database {
       this.faker.definitions.database.engine
     );
   }
+
+  /**
+   * Returns an [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string
+   *
+   * @example
+   * faker.database.mongodbObjectId() // e175cac316a79afdd0ad3afb
```
ST-DDT, line 67:
```suggestion
   * faker.database.mongodbObjectId() // 'e175cac316a79afdd0ad3afb'
```

src/database.ts
```diff
@@ -59,4 +59,15 @@ export class Database {
       this.faker.definitions.database.engine
     );
   }
+
+  /**
+   * Returns an [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string
```
ST-DDT, line 64:
```suggestion
   * Returns a MongoDB [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string.
```

ST-DDT (approved):

pkuczynski (approved):

import-brain (approved):

src/database.ts
```diff
@@ -59,4 +59,15 @@ export class Database {
...
+
+  /**
+   * Returns a MongoDB [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) string.
+   *
+   * @example
+   * faker.database.mongodbObjectId() // 'e175cac316a79afdd0ad3afb'
+   */
+  mongodbObjectId(): string {
+    // strip the "0x" from the hexadecimal output
+    return this.faker.datatype.hexaDecimal(24).replace('0x', '').toLowerCase();
```
Shinigami92, line 71:
```suggestion
    return this.faker.datatype.hexadecimal(24).replace('0x', '').toLowerCase();
```

Shinigami92 (approved):

src/database.ts
ST-DDT: Good catch

ST-DDT (approved):

PR #640, opened 2022-03-21
feat: add os. as street prefix for pl

krychaxp:
Add street_prefix


ST-DDT (approved):
Format wise this looks good. Cannot check content wise.

pkuczynski (approved):
`os.` is shortcut form `osiedle` == `estate`

PR #642, opened 2022-03-21
feat: export locales

mshima:
Locales should be exported otherwise the only way to have localeFallback is to use the default exported Faker instance.

Example:
```js
const en = require('@fakerjs/faker/locale/en');
const pt_BR = require('@fakerjs/faker/locale/pt_BR');
const { Faker } = require('@fakerjs/faker');

const faker = new Faker({ locales: {en, pt_BR}, locale: 'pt_BR', fallbackLocale: 'en' });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,40 @@ See our documentation for a list of [provided languages](https://fakerjs.dev/api
 
 ### Individual Localization Packages
 
-Faker supports incremental loading of locales.
+Faker supports loading of individual locales.
 
 ```js
 // loads only de locale
 const { faker } = require('@faker-js/faker/locale/de');
+```
+
+Faker supports adding locale fallback.
+cjs:
+
+```cjs
+const { Faker } = require('@faker-js/faker');
+const { default: de } = require("@faker-js/faker/locale/de");
+const { default: fr } = require("@faker-js/faker/locale/fr");
+
+consta faker =   new Faker({
+  locales: { de, fr },
+  locale: "de",
+  localeFallback: "fr",
+});
+```
+
+esm:
+
+```mjs
+import { Faker } from '@faker-js/faker';
+import de from "@faker-js/faker/locale/de";
+import fr from "@faker-js/faker/locale/fr";
+
+consta faker =   new Faker({
+  locales: { de, fr },
+  locale: "de",
+  localeFallback: "fr",
+});
 ```
 
 ## Setting a randomness seed
--- a/src/locale/af_ZA.ts
+++ b/src/locale/af_ZA.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import af_ZA from '../locales/af_ZA';
 import en from '../locales/en';
 
+export { default } from '../locales/af_ZA';
+
 export const faker = new Faker({
   locale: 'af_ZA',
   localeFallback: 'en',
--- a/src/locale/ar.ts
+++ b/src/locale/ar.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import ar from '../locales/ar';
 import en from '../locales/en';
 
+export { default } from '../locales/ar';
+
 export const faker = new Faker({
   locale: 'ar',
   localeFallback: 'en',
--- a/src/locale/az.ts
+++ b/src/locale/az.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import az from '../locales/az';
 import en from '../locales/en';
 
+export { default } from '../locales/az';
+
 export const faker = new Faker({
   locale: 'az',
   localeFallback: 'en',
--- a/src/locale/cz.ts
+++ b/src/locale/cz.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import cz from '../locales/cz';
 import en from '../locales/en';
 
+export { default } from '../locales/cz';
+
 export const faker = new Faker({
   locale: 'cz',
   localeFallback: 'en',
--- a/src/locale/de.ts
+++ b/src/locale/de.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import de from '../locales/de';
 import en from '../locales/en';
 
+export { default } from '../locales/de';
+
 export const faker = new Faker({
   locale: 'de',
   localeFallback: 'en',
--- a/src/locale/de_AT.ts
+++ b/src/locale/de_AT.ts
@@ -7,6 +7,8 @@ import { Faker } from '../faker';
 import de_AT from '../locales/de_AT';
 import en from '../locales/en';
 
+export { default } from '../locales/de_AT';
+
 export const faker = new Faker({
   locale: 'de_AT',
   localeFallback: 'en',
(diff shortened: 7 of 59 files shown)


mshima:
Thanks @Shinigami92 for the quick feedback.
Marking as ready for review, since it the requested changes have been addressed.

Shinigami92:
It's working? omg that would be really nice, thank you @mshima for this fix, I assume you will make many happy with that

Shinigami92 (changes requested):
Currently we need to do something like this:

```ts
import { Faker } from "@faker-js/faker";
import es from "@faker-js/faker/locales/es";
import fr from "@faker-js/faker/locales/fr";

console.log(
  "Manually loaded locale:",
  new Faker({
    locales: { es: es.default, fr: fr.default },
    locale: "es",
    localeFallback: "fr",
  }).locales.es.title
); // Manually loaded locale: Spanish
```

This is working, but it would be much better if we can somehow make it possible to omit the need of using `<locale>.default`
Or do you know a better way of importing them?

But also there are no types defined for these exports :thinking: 
image
So at least this needs to be fixed before we can merge this PR!

**Edit 1:**
I found out that this `<locale>.default` is only an issue using `esm`, but not `cjs`
For `cjs` it's just:

```ts
import { Faker } from "@faker-js/faker";
import es from "@faker-js/faker/locales/es";
import fr from "@faker-js/faker/locales/fr";

console.log(
  "Manually loaded locale:",
  new Faker({
    locales: { es, fr },
    locale: "es",
    localeFallback: "fr",
  }).locales.es.title
); // Manually loaded locale: Spanish
```

**Edit 2:**
But only when you use TS with cjs :laughing: 
When using cjs with raw js you need `.default` again :shrug: 

```js
const { Faker } = require("@faker-js/faker");
const es = require("@faker-js/faker/locales/es");
const fr = require("@faker-js/faker/locales/fr");

console.log(
  "Manually loaded locale:",
  new Faker({
    locales: { es: es.default, fr: fr.default },
    locale: "es",
    localeFallback: "fr",
  }).locales.es.title
);
```

Shinigami92:
> I think this is because of:
> 
> https://github.com/faker-js/faker/blob/56424709b7ccd7aa7fe6130c78a770a0a901d7b5/package.json#L36
> 
> This makes node esm to import cjs version. Should be `require` instead of `node`.

I was not fully aware of what each of the things are doing when I crafted all together. Feel free to fix it until it's correct and I approved that everything is working.
I have a locale faker playground where I can now test esm TS/JS and cjs TS/JS.
I use pnpm's `overrides` to link to my local `@faker-js/faker`

mshima:
I will checkout the project to see the generated code. Until now it could be done using CI 😄.

mshima:
> Currently we need to do something like this:
> 
> ```ts
> import { Faker } from "@faker-js/faker";
> import es from "@faker-js/faker/locales/es";
> import fr from "@faker-js/faker/locales/fr";
> 
> console.log(
>   "Manually loaded locale:",
>   new Faker({
>     locales: { es: es.default, fr: fr.default },
>     locale: "es",
>     localeFallback: "fr",
>   }).locales.es.title
> ); // Manually loaded locale: Spanish
> ```
> 
> This is working, but it would be much better if we can somehow make it possible to omit the need of using `<locale>.default`

Looks like this is a pattern esbuild uses.
See:
```js
node -e 'console.log(require("@faker-js/faker"));'
{ Faker: [Getter], default: [Getter], faker: [Getter] }
```

Need to change the approach

mshima:
Should work like:
```
const { Faker } = require("@faker-js/faker");
const { es, fr } = require("@faker-js/faker/locales");

console.log(
  "Manually loaded locale:",
  new Faker({
    locales: { es, fr },
    locale: "es",
    localeFallback: "fr",
  }).locales.es.title
);
```

Shinigami92:
I will test this again this evening, due to I do not have the local playground on my workstation laptop but on my home pc

Shinigami92:
@mshima I created a playground here: https://github.com/faker-js/playground
Feel free to ask if something is unclear

Shinigami92:
@mshima are you still interested in this PR? Or should someone take it over?

mshima:
If someone else manages to finishes it, go ahead.
I want to finish it, I couldn’t find the time to do it yet.

The cjs part will probably need to expose a handcrafted api. I don’t think current approach will work.

mshima:
@Shinigami92 this need approach seems to do the trick.

Shinigami92:
@mshima I see this PR and I like it, but currently our maintainer resources are a bit limited and we have a bunch of other things on our agenda. So it still could take a bit until we can proceed with this wanted feature.
We want to internally discuss if e.g. we want imports like
- `import { locale } from '@faker-js/faker/locale/de'`
- `import { de } from '@faker-js/faker/locale/de'`
- `import { de } from '@faker-js/faker/locales'`
- `import { de } from '@faker-js/faker'`
And for the long-term (not this PR) we even think about using a mono-repo structure and make possible to use
```ts
import { de } from '@faker-js/de'
import { Faker } from '@faker-js/faker'

new Faker({ /* ... use de somehow */ })
```

mshima:
@Shinigami92 no problem, I cannot use v7 for now since we need node 12 support.

mshima:
> @mshima I see this PR and I like it, but currently our maintainer resources are a bit limited and we have a bunch of other things on our agenda. So it still could take a bit until we can proceed with this wanted feature. We want to internally discuss if e.g. we want imports like
> 
> * `import { locale } from '@faker-js/faker/locale/de'`
> * `import { de } from '@faker-js/faker/locale/de'`
> * `import { de } from '@faker-js/faker/locales'`
> * `import { de } from '@faker-js/faker'`
>   And for the long-term (not this PR) we even think about using a mono-repo structure and make possible to use
> 
> ```ts
> import { de } from '@faker-js/de'
> import { Faker } from '@faker-js/faker'
> 
> new Faker({ /* ... use de somehow */ })
> ```

@Shinigami92 is there a decision about the api? We are about to start a major version, so I may work on this soon.

ST-DDT:
We will discuss this in the next team meeting again.

ST-DDT:
@Shinigami92 will tackle this in a new PR for v8.
We would like to keep this open for now, to make the discussion that happened here more visible.

mshima:
As workaround:
```
  let nativeFakerInstance;
  // Faker >=6 doesn't exports locale by itself, it only exports a faker instance with the locale.
  // We need a Faker instance for each entity, to build additional fake instances, use the locale from the exported localized faker instance.
  // See https://github.com/faker-js/faker/pull/642
  try {
    // eslint-disable-next-line import/no-dynamic-require
    nativeFakerInstance = (await import(`@faker-js/faker/locale/${nativeLanguage}`)).faker;
  } catch (error) {
    // Faker not implemented for the native language, fallback to en.
    // eslint-disable-next-line import/no-unresolved, import/no-dynamic-require
    nativeFakerInstance = (await import('@faker-js/faker/locale/en')).faker;
  }

  const faker = new Faker({
    locales: nativeFakerInstance.locales,
    locale: nativeFakerInstance.locale,
    localeFallback: nativeFakerInstance.localeFallback,
  });
```

ST-DDT:
The locales can now be imported via `import { de } from '@faker-js/faker'` and `import { fakerDE as faker } from '@faker-js/faker'`.

Superseded by #1735

PR #644, opened 2022-03-21
fix: name module gender

Shinigami92:


src/name.ts
```diff
@@ -1,5 +1,83 @@
...
+      values = male;
+      break;
+    default:
+      values = generic;
+      break;
+  }
+
+  if (values == null) {
+    if (female != null && male != null) {
+      values = faker.random.arrayElement([female, male]);
```
ST-DDT, line 72: Can you add a test for this? Just use some locale without the default fallback.

src/name.ts
```diff
@@ -1,5 +1,83 @@
 import type { Faker } from '.';
 
+export enum Gender {
+  female = 'female',
+  male = 'male',
+}
+
+// TODO christopher 21-03-2022: Remove 0 and 1 in v7
```
prisis, line 8: Wrong todo :)

src/name.ts
```diff
@@ -1,5 +1,83 @@
...
+ * Normalize gender.
+ *
+ * @param gender Gender.
+ * @returns Normalized gender.
+ */
+function normalizeGender(
+  gender?: GenderType
+): Exclude<GenderType, number> | undefined {
+  if (gender == null || typeof gender === 'string') {
+    // TODO christopher 21-03-2022: Cast can be removed when we set `strict: true`
```
prisis, line 21: same

src/name.ts
Shinigami92, line 8:
```suggestion
// TODO @Shinigami92 21-03-2022: Remove 0 and 1 in v7
```

src/name.ts
Shinigami92, line 21:
```suggestion
    // TODO @Shinigami92 21-03-2022: Cast can be removed when we set `strict: true`
```

JessicaSachs:
Should we go ahead and add support for the gender non-binary in this PR? You're setting up the types and infrastructure, which is good, but maybe we should add `NonBinary` to the enum and support it?

Shinigami92:
> Should we go ahead and add support for the gender non-binary in this PR? You're setting up the types and infrastructure, which is good, but maybe we should add `NonBinary` to the enum and support it?

I would do that in v6.2
This is just a bug fix for now, features can be added later

ST-DDT (approved):

JessicaSachs (approved):
Great first step. We'll tackle Non Binary support in 6.2 then 👍🏻

PR #645, opened 2022-03-21
Deprecation decorator

pkuczynski:
What do you think about using decorator for marking methods as deprecated?

### DIFF
--- a/src/decorators/deprecated.ts
+++ b/src/decorators/deprecated.ts
@@ -1,0 +1,23 @@
+/**
+ *
+ * @param method
+ */
+export function Deprecated(method: string): any {
+  return function (
+    target: unknown,
+    propertyKey: string,
+    descriptor: PropertyDescriptor
+  ): any {
+    const originalMethod = descriptor.value;
+
+    descriptor.value = function (...args: unknown[]): any {
+      console.warn(
+        `Deprecation Warning: faker.random.number is now located in ${method}'`
+      );
+
+      return originalMethod.apply(this, args);
+    };
+
+    return descriptor;
+  };
+}
--- a/src/random.ts
+++ b/src/random.ts
@@ -1,4 +1,5 @@
 import type { Faker } from '.';
+import { Deprecated } from './decorators/deprecated';
 
 /**
  * Method to reduce array of characters.
@@ -56,12 +57,10 @@ export class Random {
    *
    * @deprecated
    */
+  @Deprecated('faker.datatype.number')
   number(
     options?: number | { min?: number; max?: number; precision?: number }
   ): number {
-    console.warn(
-      'Deprecation Warning: faker.random.number is now located in faker.datatype.number'
-    );
     return this.faker.datatype.number(options);
   }
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,8 @@
     "outDir": "dist",
     "declaration": true,
     "esModuleInterop": true,
-    "allowJs": true
+    "allowJs": true,
+    "experimentalDecorators": true
   },
   "include": ["src/**/*"],
   "exclude": ["node_modules"]


pkuczynski:
> What about the methods that are deprecated for removal?

This is just a concept applied to an example method. We can extend the decorator with more params if we want to, so the message can be shaped differently, but also the behaviour. For example `type: 'warn', 'error'` could in the first case just warn, but still execute the method, while `error` could just fail with suggestion. Or whatever else we can think of...

I didn't focused on tests, lint, etc. Just wanted to have your initial feedback on this proposal...

ST-DDT:
> We can extend the decorator with more params if we want to

IMO:
- No param => No alternative
- With param => respective alternative

Currently this project doesn't use annotations, not sure if we should start with them for this usecase.
I hope to get rid of most deprecated methods "soon".

> For example `type: 'warn', 'error'` could in the first case just warn, but still execute the method, while `error` could just fail with suggestion.

IMO we should not define a method that just errors. We deprecate them, to notify others of its state. When they are gone, they should be gone for good.

pkuczynski:
> I do not want `experimentalDecorators` to be activated in that project.
> If a second maintainer also reject this like me, please lets close this PR.
> IMO this is just overhead.

They Stage 2 and should rather sooner then later become part of the standard. Why don't you like them @Shinigami92 ?

pkuczynski:
> > We can extend the decorator with more params if we want to
> 
> IMO:
> 
> * No param => No alternative
> * With param => respective alternative

Can be done, no problem.

> Currently this project doesn't use annotations, not sure if we should start with them for this usecase. I hope to get rid of most deprecated methods "soon".

There will be always some deprecations or changes. They are in the nature of every project...

> > For example `type: 'warn', 'error'` could in the first case just warn, but still execute the method, while `error` could just fail with suggestion.
> 
> IMO we should not define a method that just errors. We deprecate them, to notify others of its state. When they are gone, they should be gone for good.

Sure.

Shinigami92:
> They Stage 2 and should rather sooner then later become part of the standard. Why don't you like them @Shinigami92 ?

Decorators were already there when I started 5 years ago with Angular v4+
I only like them when used in frameworks like NestJS or Angular, but it feels wrong when used in a library
As I said, IMO they just introduce not only runtime overhead but also mental complexity
And to introduce them only for one usage I think that is overkill

Would we could do instead would be to define a function to manage the deprecation messages
But even then, we have functions that warns conditionally based on input parameters, so it's really just overkill and cannot be applied constantly to every function

pkuczynski:
> > They Stage 2 and should rather sooner then later become part of the standard. Why don't you like them @Shinigami92 ?
> 
> Decorators were already there when I started 5 years ago with Angular v4+

Yes, they are around for quite some time, but given how many libraries relay on them and the fact they are built into TS, I can't see them not being introduced to the language.

> I only like them when used in frameworks like NestJS or Angular, but it feels wrong when used in a library

I know many popular libraries using them, like TypeORM for example. They very handy.

> As I said, IMO they just introduce not only runtime overhead but also mental complexity And to introduce them only for one usage I think that is overkill

What runtime overhead and mental complexity? They are very simple and very popular in many languages... The whole purpose of them is actually simplifying code, not making it more complex.
 
> Would we could do instead would be to define a function to manage the deprecation messages But even then, we have functions that warns conditionally based on input parameters, so it's really just overkill and cannot be applied constantly to every function

Thats why decorator comes super handy here. Can be applied very constantly to every method...

Shinigami92 (changes requested):
Beside the fakt that I'm against using decorators at all in our repo, I would highly suggest not to write them on our own but use something like this: https://github.com/vilic/deprecated-decorator

pkuczynski:
> Beside the fakt that I'm against using decorators at all in our repo, I would highly suggest not to write them on our own but use something like this: https://github.com/vilic/deprecated-decorator

Decorators are simple to write. The one I did is literally few lines of code and does the job nicely. Why do you think bringing just another dependency would help for our case? It's a bit like getting a gun to kill mosquito...

Shinigami92:
> Decorators are simple to write. The one I did is literally few lines of code and does the job nicely. Why do you think bringing just another dependency would help for our case? It's a bit like getting a gun to kill mosquito...

Because:
1. I as a maintainer of this repo do not want to maintain them in the long run
2. Would at least follow a standard instead of having this implementation in this package and another one in another package and so on

ST-DDT:
@Shinigami92 Lets discuss this at the next meeting with the entire team.

pkuczynski:
> Because:
> 
> 1. I as a maintainer of this repo do not want to maintain them in the long run

The package has been published last time 6y ago so either it's not maintained or there is no maintenance needed.

> 2. Would at least follow a standard instead of having this implementation in this package and another one in another package and so on

What I have implemented covers the most basic flow tailored to faker, which is aligned to what the package mentioned by you does as well. So they are inline.

pkuczynski:
> @Shinigami92 Lets discuss this at the next meeting with the other maintainers.

Let me know what's your call and I can either close or adopt to your wishes :)

Shinigami92:
In yesterdays maintainers-team-meeting we discussed that we don't want to use decorators for this approach, but we like your proposal to standardize deprecation warnings, so we would like to have a function that we can call.
It should have parameters for `version`, `alternative` and the possibility to override the default message.
It should always start with `Deprecation Warning:`

Further implementation details can be discussed while implementing this PR 🙂

pkuczynski:
OK :) I will close this PR and work on a new one...

ST-DDT:
Superseded by #689

PR #646, opened 2022-03-21
fix: corrected the Costa Rican IBAN format

### LINKED ISSUE #643: Wrong IBAN formatting for Costa Rica
### Describe the bug

@halvorv and I discovered an error in the generation of Costa Rican IBAN codes. According to multiple sources we have found online ([here](https://www.bccr.fi.cr/en/payments-system/general-information/iban) and [here](https://en.wikipedia.org/wiki/International_Bank_Account_Number#Structure) for example), a Costa Rican IBAN code should consist of in total 22 characters, following the structure like this:
`CR99 0000 0000 0000 8888 88`
- first two characters are `CR`
- two digit checksum
- one reserved character
- three digit bank code
- fourteen digit account number

This totals up to 22 characters in total. However, in Faker, it is currently defined like this:

```js
{
      country: 'CR',
      total: 21,
      bban: [
        {
          type: 'n',
          count: 3,
        },
        {
          type: 'n',
          count: 14,
        },
      ],
      format: 'CRkk bbbc cccc cccc cccc c',
    },
```
According to [this](https://bank-code.net/iban/structure/costa-rica-international-bank-account-number) source, we can include that one reserved character within the bank code.

Hence, the first object in the `bban` array should be defined with a count of `4` instead of `3`, the total should become `22` and the formatting should become `CRkk bbbb cccc cccc cccc cc`

### Reproduction

Generate an IBAN number for Costa Rica and it will generate an invalid one:
```js
faker.finance.iban(false, 'CR')
```

### Additional Info

_No response_

MrGussio:
Fixes #643 by correcting the Costa Rican IBAN formatting

### DIFF
--- a/test/finance_iban.spec.ts
+++ b/test/finance_iban.spec.ts
@@ -283,9 +283,9 @@ describe('finance_iban', () => {
           expect(
             iban.substring(0, 2),
             iban.substring(0, 2) +
-              ' must contains only characters in CR IBAN ' +
+              "must start with 'CR' in CR IBAN " +
               ibanFormated
-          ).match(/^[A-Z]{2}$/);
+          ).to.eq('CR');
 
           expect(
             iban.substring(2, 22),
--- a/test/finance_iban.spec.ts
+++ b/test/finance_iban.spec.ts
@@ -255,6 +255,50 @@ describe('finance_iban', () => {
             'the result should be equal to 1'
           ).toBe(1);
         });
+
+        it('IBAN for Costa Rica is correct', () => {
+          // Costa Rica
+          // https://wise.com/us/iban/costa-rica
+          // Length 22
+          // BBAN 1n,3n,14n
+          // GEkk xbbb cccc cccc cccc cccc cccc
+          // x = reserve digit
+          // b = National bank code (digits)
+          // c = Account number (digits)
+
+          // example IBAN CR05 0152 0200 1026 2840 66
+
+          const iban = faker.finance.iban(false, 'CR');
+
+          expect(iban).satisfy(validator.isIBAN);
+
+          const ibanFormated = iban.match(/.{1,4}/g).join(' ');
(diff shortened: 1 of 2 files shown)


ST-DDT (approved):
Checked and seems to be working according to: https://www.iban.de/iban-pruefen.html

Shinigami92 (changes requested):
Don't we have a test case for that? Nothing failed...
But I think we should add a test case for that because we also have tests for other ibans

See: https://github.com/faker-js/faker/blob/main/test/finance_iban.spec.ts

MrGussio:
> Don't we have a test case for that? Nothing failed... But I think we should add a test case for that because we also have tests for other ibans
> 
> See: https://github.com/faker-js/faker/blob/main/test/finance_iban.spec.ts

I added the test for Costa Rica to the test file and made issue #660 for the other tests of missing IBAN country codes.

In the latest commit I also splitted the reserved character from the banking digits of the Costa Rican IBAN formatting, to really cohere to the definitions I have found online.

test/finance_iban.spec.ts
```diff
@@ -255,6 +255,50 @@ describe('finance_iban', () => {
...
+            22,
+            `CR IBAN would be 22 chars length, given is ${iban.length}`
+          ).toBe(iban.length);
+
+          expect(
+            iban.substring(0, 2),
+            iban.substring(0, 2) +
+              ' must contains only characters in CR IBAN ' +
+              ibanFormated
+          ).match(/^[A-Z]{2}$/);
```
ST-DDT, line 288: IMO this should be `'CR'` not some kind of regex. Or am I missing something here?

MrGussio: You're right, that probably does make more sense. The other tests also used this regex matching for the first two characters, but I agree that matching it with the country code is more robust.

ST-DDT (approved):

Shinigami92 (approved):

PR #647, opened 2022-03-21
ci: configure codecov patch threshold

Shinigami92:
The threshold is not useful as it only fails PRs that are successfully but the threshold is not meet.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -14,4 +14,4 @@ coverage:
     patch:
       default:
         target: 95%
-        threshold: 2%
+        threshold: 1%
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -14,4 +14,4 @@ coverage:
     patch:
       default:
         target: 95%
-        threshold: 100%
+        threshold: 2%
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -10,4 +10,8 @@ coverage:
     project:
       default:
         target: 95%
+        threshold: 2%
+    patch:
+      default:
+        target: 95%
         threshold: 100%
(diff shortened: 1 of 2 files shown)


ST-DDT (dismissed):

griest024 (changes requested):
2% threshold seems large. Why not 0% or at worst, 1%?

Shinigami92:
> 2% threshold seems large. Why not 0% or at worst, 1%?

@griest024 Okay I may have misunderstood `threshold`, maybe you could try to explain me what is going on.

The initial issue here is that the CI failed several times with the `codecov/patch` action.
I found out that it doesn't take the config we set for codecov/project default, so I needed to explicitly configured it for patch.

I assume setting the threshold to 0% will not help again, cause the default may was be 0% by default previously. (I could be wrong)
But everytime we did a helpful refactoring that moved some parts, the patch action failed and marked the whole PR as ❌, were every other action was ✅ and the PR is good to go. So it just may confused people and especially maintainers that there was something broken or wrong, but it was not.

So I would like to first try and drive for a while with my proposed settings in this PR.
If something is still bad, we can change it in a future PR anyway.

griest024:
@Shinigami92 from the [docs](https://docs.codecov.com/docs/commit-status#threshold)

> Allow the coverage to drop by X%, and posting a success status.

It's possible that there is some broken logic for `0%` that will fail PRs that should have passed. I would think that `1%` would avoid this.

.github/.codecov.yml
```diff
@@ -11,3 +11,7 @@ coverage:
       default:
         target: 95%
         threshold: 2%
+    patch:
+      default:
+        target: 95%
+        threshold: 2%
```
Shinigami92, line 17:
```suggestion
        threshold: 1%
```

griest024 (approved):

ST-DDT (approved):

PR #650, opened 2022-03-21
refactor: use smart eqeqeq null checks

Shinigami92:
Instead of explicitly checking on undefined, I refactored the code to relax-checking on null with a non-strict `==`
These are checking for "null or undefined" and called smart null checks.
The benefit of them is, that folks that are using plain JS without type-checking can also pass `null` as an `undefined`/optional parameter and get the same behavior as not passing a value.


ST-DDT:
Is there a lint rule we can enable to enforce these checks?

ST-DDT (approved):

ST-DDT:
It feels like the usage of `v == null`, `!v`, `v != null`, ... is inconsistent.

Shinigami92:
> It feels like the usage of `v == null`, `!v`, `v != null`, ... is inconsistent.

It is not, cause `!0` is `false`. So there are some places where we explicitly count e.g. a length that is `0` as a `false` case and then we fall back to e.g. a defined value like `4`

ST-DDT:
This needs a rebase

ST-DDT (dismissed):

src/date.ts
```diff
@@ -123,7 +123,7 @@ export class _Date {
     to: string | Date | number,
     num?: number
   ): Date[] {
-    if (typeof num === 'undefined') {
+    if (num == null) {
       num = 3;
     }
```
ST-DDT, line 128: Just set it as default.

pkuczynski (approved):

PR #654, opened 2022-03-22
fix: only return word with desirable alpha characters

### LINKED ISSUE #113: `random.word()` can return undesirable non-alpha characters

Shinigami92:
fix #113

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -284,12 +284,37 @@ export class Random {
       'name.jobType',
     ];
 
-    // randomly pick from the many faker methods that can generate words
+    const bannedChars = [
+      ' ',
+      '!',
+      '#',
+      '%',
+      '&',
+      '*',
+      ')',
+      '(',
+      '+',
+      '=',
+      '.',
+      '<',
+      '>',
+      '{',
+      '}',
+      '[',
+      ']',
+      ':',
+      ';',
+      "'",
+      '"',
+      '_',
+      '-',
+    ];
     let result: string;
     do {
+      // randomly pick from the many faker methods that can generate words
       const randomWordMethod = this.faker.random.arrayElement(wordMethods);
       result = this.faker.fake('{{' + randomWordMethod + '}}');
-    } while (!/^[\w\s]+$/.test(result));
+    } while (bannedChars.some((char) => result.includes(char)));
     return this.faker.random.arrayElement(result.split(' '));
   }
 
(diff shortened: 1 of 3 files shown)


src/random.ts
```diff
@@ -286,8 +285,11 @@ export class Random {
     ];
 
     // randomly pick from the many faker methods that can generate words
-    const randomWordMethod = this.faker.random.arrayElement(wordMethods);
-    const result = this.faker.fake('{{' + randomWordMethod + '}}');
+    let result: string;
+    do {
+      const randomWordMethod = this.faker.random.arrayElement(wordMethods);
+      result = this.faker.fake('{{' + randomWordMethod + '}}');
+    } while (!/^[\w\s]+$/.test(result));
```
ST-DDT, line 292: Does this method work for non latin letter languages?

Shinigami92: Good question: I will research this...

Shinigami92: Please have a new look. I know we cannot handle every special case this way, but currently I do not see a better solution due to we do not have `\p` in js regex 🙁 
So I think for now this is enough to check and if there is any issue in the future, we can patch it

src/random.ts
```diff
@@ -285,9 +284,37 @@ export class Random {
...
+      '_',
+      '-',
+    ];
+    let result: string;
+    do {
+      // randomly pick from the many faker methods that can generate words
+      const randomWordMethod = this.faker.random.arrayElement(wordMethods);
+      result = this.faker.fake('{{' + randomWordMethod + '}}');
+    } while (bannedChars.some((char) => result.includes(char)));
     return this.faker.random.arrayElement(result.split(' '));
```
ST-DDT, line 318: Since space is banned there is no longer a need to select here.

Shinigami92: Oh damn, I implemented it wrong! There are some strings that contains spaces so I need to allow it back.

prisis (approved):

ST-DDT (approved):

PR #661, opened 2022-03-23
refactor: deprecate time.recent

Shinigami92:


ST-DDT (dismissed):

src/time.ts
```diff
@@ -2,6 +2,8 @@ import type { LiteralUnion } from './faker';
 
 /**
  * Module to generate time of dates in various formats.
+ *
+ * @deprecated Just use the native Date object.
```
prisis, line 6: Can we make this message a bit nicer  
  
Something like   
  
Since faker-js 6.0.0 The faker.time.recent is deprecated. You should stop using it, as it will be removed in the future. Start to use \`(new Date()).recent(abbr | wide | unix).

Shinigami92: Is that better?

ST-DDT (approved):

PR #664, opened 2022-03-23
fix: datatype.number when min = max + precision, throw when max > min

### LINKED ISSUE #329: datatype.number does not produce random number when min = max + precision
### Describe the bug

`faker.datatype.number` returns always the same value `=max+precision`, when `min = max + precision`

### Reproduction

```
faker.datatype.number({ min: 100000 })
// 100000 - on every run
```

### Additional Info

A check ensuring `max` is reasonably larger then `min` is missing...

pkuczynski:
Fixes #329


src/datatype.ts
```diff
@@ -41,39 +41,27 @@ export class Datatype {
...
-    if (typeof options.min === 'number') {
-      min = options.min;
-    }
-
-    if (typeof options.max === 'number') {
-      max = options.max;
-    }
+    const precision = typeof opts.precision === 'number' ? opts.precision : 1;
+    const min = typeof opts.min === 'number' ? opts.min : 0;
+    let max = typeof opts.max === 'number' ? opts.max : min + 99999;
```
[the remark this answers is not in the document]
pkuczynski: That is what was suggested in #329. For me, it makes sense that when max is not provided, to give it a nice, wide span. What would be your suggestion to do it differently?

src/datatype.ts
```diff
@@ -41,39 +41,27 @@ export class Datatype {
...
 
-    if (typeof options.precision === 'number') {
-      precision = options.precision;
+    if (max <= min) {
+      throw new Error(`Max should be larger then min: ${max} > ${min}`);
     }
```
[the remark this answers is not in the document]
pkuczynski: I am not sure if this is the right thing to do in this case. Imagine:

```faker.datatype.number({ min: 100, max: 10})```

I would be really surprised that values got swapped. Especially if they are not given directly as in the example above, but coming from some other parts of the code. It would just hide an error instead of surfacing it...

File: src/datatype.ts [hunk 3]
```diff
@@ -34,39 +34,27 @@ export class Datatype {
...
-    if (typeof options.min === 'number') {
-      min = options.min;
-    }
-
-    if (typeof options.max === 'number') {
-      max = options.max;
-    }
+    const precision = typeof opts.precision === 'number' ? opts.precision : 1;
+    const min = typeof opts.min === 'number' ? opts.min : 0;
+    let max = typeof opts.max === 'number' ? opts.max : min + 99999;
```
Shinigami92, line 41: Could you swap lines a bit to reduce diff-changes?

```suggestion
    const min = typeof opts.min === 'number' ? opts.min : 0;
    let max = typeof opts.max === 'number' ? opts.max : min + 99999;
    const precision = typeof opts.precision === 'number' ? opts.precision : 1;
```

min and max need to be swapped because now max uses min

src/datatype.ts
```diff
@@ -34,39 +34,27 @@ export class Datatype {
...
 
-    if (typeof options.precision === 'number') {
-      precision = options.precision;
+    if (max <= min) {
+      throw new Error(`Max should be larger then min: ${max} > ${min}`);
```
Shinigami92, line 44: We need a test case to test this error

Shinigami92 (changes requested):

src/datatype.ts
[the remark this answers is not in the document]
ST-DDT: ok, please update the docs accordingly

src/datatype.ts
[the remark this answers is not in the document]
ST-DDT: ok, i agree except for min == max then it should return just that.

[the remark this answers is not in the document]
pkuczynski: Done

src/datatype.ts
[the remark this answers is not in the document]
pkuczynski: Done

src/datatype.ts
pkuczynski: Done

ST-DDT (approved):

ST-DDT:
Please fix the failing tests.

pkuczynski:

Yeah, I was working on them :)

pkuczynski:

All fixed now...

src/address.ts
```diff
@@ -407,8 +407,8 @@ export class Address {
   latitude(max: number = 90, min: number = -90, precision: number = 4): string {
     return this.faker.datatype
       .number({
-        max: max,
-        min: min,
+        max,
+        min,
```
ST-DDT, line 411:
```suggestion
        min,
        max,
```

That way it feels more natural to me. But not really important.

test/address.spec.ts
```diff
@@ -411,7 +411,7 @@ describe('address', () => {
 
         it('returns latitude with min and max and default precision', () => {
           for (let i = 0; i < 100; i++) {
-            const latitude = faker.address.latitude(-5, 5);
+            const latitude = faker.address.latitude(5, -5);
```
ST-DDT, line 414: 😿 🤔

src/locales/ro/address/index.ts
```diff
@@ -25,10 +25,10 @@ const address = {
   secondary_address,
   state,
   state_abbr,
-  streets,
   street_address,
   street_name,
   street_suffix,
+  streets,
```
ST-DDT, line 31: Please revert this change.

test/address.spec.ts
pkuczynski: I know :/ The signature of this function was not really good. It should take `opts` instead of params `max, min`. Or maybe I should do the detection inside `latitude` and swap if needed? This sounds like a good idea... wdyt?

src/address.ts
pkuczynski: For me too. I kept the order but simplified the signature. I will apply our suggestion.

src/locales/ro/address/index.ts
pkuczynski: It must have been my editor doing that automatically, sorry...

test/address.spec.ts
ST-DDT: Well, I'm still a fan of swapping in `datatype.number()` as well. 😉

pkuczynski: The params are called `max` and `min` so user should provide valid values :)

Shinigami92 (approved):

ST-DDT:
Since my expectation of `generateMeARandomNumberBetween(-5, 5)` is to work, I will not approve this.
But I will accept it, if the others want it this PR's way.

My sole requirement: Add a hint to the changelog

Shinigami92:
> Since my expectation of `generateMeARandomNumberBetween(-5, 5)` is to work, I will not approve this. [...]

But I do understand it correctly that this PR did not introduce a change in that? And the test was just calling it wrongly in history?

ST-DDT:
> But I do understand it correctly that this PR did not introduce a change in that? And the test was just calling it wrongly in history?

It does: [src/datatype.ts:44](https://github.com/faker-js/faker/pull/664/files#diff-5699f20396717381133051426f9769465b1d8d460c942def407bb4f03907d5c3R43-R45)

src/random.ts
```diff
@@ -107,8 +107,12 @@ export class Random {
   arrayElement<T = string>(
     array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
   ): T {
-    const r = this.faker.datatype.number({ max: array.length - 1 });
-    return array[r];
+    const index =
+      array.length > 1
+        ? this.faker.datatype.number({ max: array.length - 1 })
+        : 0;
```
ST-DDT, line 113: I think this change is no longer needed

pkuczynski:
> > Since my expectation of `generateMeARandomNumberBetween(-5, 5)` is to work, I will not approve this. [...]
> 
> But I do understand it correctly that this PR did not introduce a change in that? And the test was just calling it wrongly in history?

This PR introduced an additional check if `max > min` and throws an error when they are not. We don't have a function that takes params like `(-5, 5)`. It takes options `({ min: -5, max: 5 })`. To see the error being thrown user would have to call it with `({ min: 5, max: -5 })` which is mathematically incorrect, right? So if anything we are only giving users better support now...

pkuczynski:
> My requirements:
> 
> * Add a hint to the next changelog/release

Not sure how shall I do that?

> * Please also rename the PR to reflect the changes to behavior of the method (e.g. `be more strict with datatype.number()'s behavior`).

Done. Additionaly added some jsdoc.

src/random.ts
pkuczynski: This change is a bit of perf optimization. Does not make sense to call `number` if the array length is 1. In such a case we should just return this element.

Shinigami92:
@pkuczynski Could you please shorten the title of this PR to something that is max 72 chars long?
The title is taken as commit title on squash-merge

pkuczynski:
> @pkuczynski Could you please shorten the title of this PR to something that is max 72 chars long?
> The title is taken as commit title on squash-merge

Done

ST-DDT (approved):
I changed my mind. number isn't the offender, latitude is.
So all is good.

PR #670, opened 2022-03-24
refactor: date accept number input

xDivisionByZerox:
Reopen of #576 in favor of cleaner git history.

Created in relation to https://github.com/faker-js/faker/issues/394.

I changed the way dates are created from the `Date.parse` method to the `Date.constructor` since it uses that under the hood ([MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters)). 
Invalid dates were not handled before.

### DIFF
--- a/src/date.ts
+++ b/src/date.ts
@@ -17,19 +17,6 @@ function toDate(date?: string | Date | number): Date {
 }
 
 /**
- * Converts date passed as a string or Date to milliseconds. If nothing passed, takes current date.
- *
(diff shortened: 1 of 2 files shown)


xDivisionByZerox:
This history graph looks so much better :D

Shinigami92 (approved):

ST-DDT (approved):

PR #671, opened 2022-03-25
chore: update codeowners file

Shinigami92:
- Formatted the file a bit better
- Added maintainers also to docs, so one of them is just required (codeowners file works with `or`-condition in a line)


Shinigami92:
We need to give the maintainers-docs group write access

ST-DDT (approved):
^

Shinigami92:
Okay, let's wait on @damienwebdev and the write access for `maintainers-docs`

Shinigami92:
Beside the fact that the docs owners doesn't work as expected, it will not break when I merge this...
So I will do that now

PR #680, opened 2022-03-26
refactor(address.secondaryAddress): use local definition

### LINKED ISSUE #676: `address.secondaryAddress` doesn't use `faker.definitions.address.secondary_address`

The current implementation of `faker.address.secondaryAddress` uses a hardcoded string array to choose from instead of the definition one. This means locales are ignored for this method.

https://github.com/faker-js/faker/blob/e62fd15cc7b83cc314a22ad9b13cd68f222d4076/src/address.ts#L311-L316

### Suggested solution

Use `faker.definitions.address.secondary_address` instead of the hardcoded string array

```typescript
secondaryAddress(): string {
  return this.Helpers.replaceSymbolWithNumber(
    this.faker.random.arrayElement(this.faker.definitions.address.secondary_address)
  );
}
```

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
Fixes #676.


Shinigami92:
Oh so there were already definitions in en :eyes: 
So we can move this to v6.1

Shinigami92 (approved):

pkuczynski:
There is no need for changes in tests?

import-brain (approved):

PR #683, opened 2022-03-26
fix(address.city): use string argument

### LINKED ISSUE #682: `address.city` string parameter ignored
### Describe the bug

When using the `address.city` function you are allowed to pass a format as a string (for custom format) or as a number for index-based selection. When providing a custom string format it will be ignored.

### Reproduction

faker.address.city('{{address.cityPrefix}} {{datatype.string}}')

### Additional Info

Is marked as todo in #541

xDivisionByZerox:
Fixes #682


xDivisionByZerox:
Should the JSDOCs have a `@see faker.fake` line in it, since it's used to parse the formats?

xDivisionByZerox:
Should I add test for `address.city` since there are non so far.

src/address.ts
```diff
@@ -173,11 +168,17 @@ export class Address {
...
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
+      if (typeof format === 'number') {
+        return formats[Math.max(0, Math.min(format, formats.length - 1))];
+      } else if (typeof format === 'string') {
+        return format;
+      } else {
+        return formats[this.faker.datatype.number(formats.length - 1)];
```
ST-DDT, line 177: Use random.arrayElement here.

ST-DDT (commented):
What is the added benefit of using `address.prefix('format')` to just using `faker.fake('format')`?

src/address.ts
xDivisionByZerox: fixed in d6a48c128476162b6ced1fdb72ee2082ab57c206

ST-DDT:
> Should I add tests for `address.city` since there are non so far. Or should this be done in a separate PR?

Tests would be useful.

Shinigami92 (approved):

src/address.ts
```diff
@@ -80,7 +77,6 @@ export class Address {
   readonly Helpers: Helpers;
 
   constructor(private readonly faker: Faker) {
-    f = this.faker.fake;
     this.Helpers = this.faker.helpers;
```
ST-DDT, line 80: If you remove `f`, then also remove the `Helpers`.

ST-DDT (commented):
Did I miss the test for `city(number)` and city(string)`?

src/address.ts
ST-DDT: This is what I meant.

xDivisionByZerox: done in 91c011196bd2bb0df9f16bf9fb70b2c7dcb8f4ed

src/address.ts
```diff
@@ -173,11 +163,17 @@ export class Address {
       formats.push('{{address.cityName}}');
     }
 
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
+      if (typeof format === 'number') {
+        return formats[Math.max(0, Math.min(format, formats.length - 1))];
```
ST-DDT, line 168: Do not box the values. Either throw or ignore invalid values.

Please emit a deprecation warning for both string and number parameters.
Instead of hiding the actual pattern (that may change between faker version), they should hardcode the pattern in combination with faker.fake.

ST-DDT (changes requested):
Could you please extract the tests for the other address methods into a separate PR?

src/address.ts
```diff
@@ -173,11 +163,17 @@ export class Address {
       formats.push('{{address.cityName}}');
     }
 
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
```
Shinigami92, line 166: Currently it feels like this PR does two things that can be separated

1. Just replace `fake` and `Helpers` with their `this.faker.` calls
2. This rework

IMO we should split them

Shinigami92 (changes requested):

src/address.ts
xDivisionByZerox: When #748 is merged I can rebase and continue?

src/address.ts
xDivisionByZerox: > they should hardcode the pattern in combination with faker.fake.

Sorry but I don't understand what you mean by that.

src/address.ts
Shinigami92: exactly

xDivisionByZerox:
So why do the `helpers` tests fail now?

Shinigami92:
> So why do the `helpers` tests fail now?

Seems like some seeded values need to be updated

src/address.ts
```diff
@@ -160,15 +159,24 @@ export class Address {
       '{{name.lastName}}{{address.citySuffix}}',
     ];
 
-    if (!format && this.faker.definitions.address.city_name) {
+    if (
+      typeof format === 'number' &&
+      this.faker.definitions.address.city_name
+    ) {
```
ST-DDT, line 165: I think the condition for the format flipped here.
Not sure whether `format` should have an impact here.

src/address.ts
```diff
@@ -160,15 +159,24 @@ export class Address {
...
       formats.push('{{address.cityName}}');
     }
 
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
+      if (typeof format === 'number') {
+        return formats[format];
```
ST-DDT, line 171: Log a deprecation warning here because using a fixed format this way may break during updates and thus should be hardcoded using `faker.fake(format)` instead.

src/address.ts
```diff
@@ -160,15 +159,24 @@ export class Address {
...
+      } else if (typeof format === 'string') {
+        return format;
```
ST-DDT, line 173: Same here

ST-DDT (changes requested):

src/address.ts
xDivisionByZerox: So if I understand correctly the `format` parameter, in general, should be depricated?

src/address.ts
xDivisionByZerox: The old condition was just `!format`. So if I provide a string format as a parameter `this.faker.definitions.address.city_name` was getting pushed, but we never access the formats array in that case. So we only case we need to push the definitions is when a number parameter is provided.

xDivisionByZerox: Or was this by design? That you could never choose the 5th format (`{{address.cityName}}`) when providing an index.

ST-DDT: I think they might have had the same intention as I have with deprecating the parameters. If you specify the index 4 then it would just be a call to cityName and they should do that directly instead.

I think it is safe to remove this limitation since we deprecate the name and index parameters anyway.

src/address.ts
ST-DDT: Yes. If they want a fixed format they should specify exactly that.

xDivisionByZerox:
Yo, could it be that the dates generated by us break the test if they run on machines in different timezones? All tests pass on my end but fail here. The logs show expected date values `2012-02-01T23:00:00.000Z` instead of `2012-02-02T00:00:00.000Z`. 
My timezone is UTC+1 without daylight saving times and UTC+2 with daylight saving times included.

xDivisionByZerox:
Nvm we expect any date in `helpers.spec`. But why can they fail in this case?

src/address.ts
xDivisionByZerox: So not the parameter itself but the number parameter should be deprecated?

ST-DDT: No, if they have a fixed format (either string or index) the should call `faker.fake` and not this method that just forwards the parameter there.

xDivisionByZerox: Ok I understand.

xDivisionByZerox: done

src/address.ts
ST-DDT: If you change this to

````ts
if (this.faker.definitions.address.city_name != null) {
````

then I think this PR is ready.

Comment (xDivisionByZerox) [reply src/address.ts]: done :)

**Comment by xDivisionByZerox** [comment]
I don't get this test

**Comment by ST-DDT** [comment]
@xDivisionByZerox I reran the pipeline and now its fixed. Looks like a temporarily codecov server issue. All green now.

**Review by ST-DDT — APPROVED** [review]

File: src/address.ts [hunk 8]
```diff
@@ -159,15 +159,32 @@ export class Address {
...
     }
 
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
+      if (typeof format === 'number') {
+        deprecated({
+          deprecated: 'address.city(number)',
+          since: 'v6.2.0',
```
Shinigami92, line 170: We are already targeting v6.3.0 as next

```suggestion
          since: 'v6.3.0',
```

Is there also a test that spies for this warning?

src/address.ts
```diff
@@ -159,15 +159,32 @@ export class Address {
...
+          until: 'v7.0.0',
+        });
+        return formats[format];
+      } else if (typeof format === 'string') {
+        deprecated({
+          deprecated: 'address.city(string)',
+          since: 'v6.2.0',
```
Shinigami92, line 177:
```suggestion
          since: 'v6.3.0',
```

src/address.ts
```diff
@@ -159,15 +159,32 @@ export class Address {
...
       formats.push('{{address.cityName}}');
     }
 
-    if (typeof format !== 'number') {
-      format = this.faker.datatype.number(formats.length - 1);
-    }
+    const useFormat = (() => {
+      if (typeof format === 'number') {
+        deprecated({
+          deprecated: 'address.city(number)',
```
Shinigami92, line 169: On all places where we use this function, we use `deprecated: 'faker.address.city(number)'`

```suggestion
          deprecated: 'faker.address.city(number)',
```

Also, we may provide a hint what the user should use instead.

src/address.ts
```diff
@@ -159,15 +159,32 @@ export class Address {
...
+          since: 'v6.2.0',
+          until: 'v7.0.0',
+        });
+        return formats[format];
+      } else if (typeof format === 'string') {
+        deprecated({
+          deprecated: 'address.city(string)',
+          since: 'v6.2.0',
+          until: 'v7.0.0',
+          proposed: 'faker.fake(string)',
```
Shinigami92, line 179:
```suggestion
          deprecated: 'faker.address.city(string)',
          proposed: 'faker.fake(string)',
          since: 'v6.3.0',
          until: 'v7.0.0',
```

Shinigami92 (changes requested):

src/address.ts
xDivisionByZerox: FYI:
https://github.com/faker-js/faker/blob/712b1de126ea6580660a320e065c35ac775f09b3/src/name.ts#L30-L35
https://github.com/faker-js/faker/blob/712b1de126ea6580660a320e065c35ac775f09b3/src/helpers.ts#L510-L515
https://github.com/faker-js/faker/blob/712b1de126ea6580660a320e065c35ac775f09b3/src/helpers.ts#L584-L588
https://github.com/faker-js/faker/blob/712b1de126ea6580660a320e065c35ac775f09b3/src/helpers.ts#L636-L641

xDivisionByZerox: done in 0747cb4aa2a2d0b6e49650019a018767ab8b29eb

src/address.ts
xDivisionByZerox: done in 0747cb4aa2a2d0b6e49650019a018767ab8b29eb

src/address.ts
xDivisionByZerox: version update in 0747cb4aa2a2d0b6e49650019a018767ab8b29eb
tests in 43c3d31c5a924826fe05a4a05297d2fbf736eebc

xDivisionByZerox:
@Shinigami92 can you have a look at the error in `address.nearbyGPSCoordinate` (since you wrote the test for it)? I don't get why changing a depreciation message resolves to this test failing.

Shinigami92:
> @Shinigami92 can you have a look at the error in `address.nearbyGPSCoordinate` (since you wrote the test for it)? I don't get how changing a depreciation message leads to this test failing.

Yeah I see, the seed for the execution of the `nearbyGPSCoordinate` check changed due to that the execution order of the tests changed because you added the `city` test above it.
It is not your fault! The altered seed just now shows that we have an expectation issue for the `nearbyGPSCoordinate` check.

@ST-DDT and I working on it now and try to solve it.

ST-DDT:
Superseded by #948

PR #688, opened 2022-03-26
fix: fake behavior with special replacement patterns

ST-DDT:
Fixes the issue described here: https://github.com/faker-js/faker/pull/503#issuecomment-1079744428

### DIFF
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -50,10 +50,10 @@ export class Fake {
 
     // find first matching {{ and }}
     const start = str.search(/{{[a-z]/);
-    const end = str.substring(start).search(/}}/) + start;
+    const end = str.indexOf('}}', start);
 
     // if no {{ and }} is found, we are done
-    if (start === -1 || end < start) {
+    if (start === -1 || end === -1) {
       return str;
     }
 
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -43,16 +43,13 @@ export class Fake {
    * faker.fake('I flipped the coin an got: {{random.arrayElement(["heads", "tails"])}}') // 'I flipped the coin an got: tails'
    */
   fake(str: string): string {
-    // setup default response as empty string
-    let res = '';
-
     // if incoming str parameter is not provided, return error message
     if (typeof str !== 'string' || str.length === 0) {
       throw new Error('string parameter is required!');
     }
 
     // find first matching {{ and }}
-    const start = str.search('{{');
+    const start = str.search('{{[a-z]');
     const end = str.search('}}');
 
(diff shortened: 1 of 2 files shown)


src/fake.ts
```diff
@@ -43,32 +43,25 @@ export class Fake {
...
-    let res = '';
-
     // if incoming str parameter is not provided, return error message
     if (typeof str !== 'string' || str.length === 0) {
       throw new Error('string parameter is required!');
     }
 
     // find first matching {{ and }}
-    const start = str.search('{{');
+    const start = str.search('{{[a-z]');
```
Shinigami92, line 52: This looks like a regex :thinking: 
Can `search` can take a real regex? Then we should use `str.search(/\{\{[a-z]/)`

ST-DDT: Fixed

test/fake.spec.ts
```diff
@@ -50,5 +50,39 @@ describe('fake', () => {
...
+      expect(faker.fake('}}hello{{random.alpha}}')).toMatch(/^}}hello[a-z]$/);
+    });
+
+    it('should be able to handle connected brackets', () => {
+      expect(faker.fake('{{{random.alpha}}}')).toMatch(/^{[a-z]}$/);
+    });
+
+    it('should be able to handle empty brackets', () => {
+      expect(faker.fake('{{}}')).toBe('{{}}');
+    });
```
ST-DDT, line 78: Please check whether these tests match your expectation regarding the `fake()` method.

I'm especially unsure about `{{}}`.

src/fake.ts
```diff
@@ -43,32 +43,25 @@ export class Fake {
...
     // if incoming str parameter is not provided, return error message
     if (typeof str !== 'string' || str.length === 0) {
       throw new Error('string parameter is required!');
     }
 
     // find first matching {{ and }}
-    const start = str.search('{{');
-    const end = str.search('}}');
+    const start = str.search(/{{[a-z]/);
+    const end = str.substring(start).search(/}}/) + start;
```
xDivisionByZerox, line 53: `end` can be determined by using `indexOf` instead of search since a RegEx is not required.

xDivisionByZerox: This would also drop the required `substring` call upfront since `indexOf` accepts a start position.

ST-DDT: Fixed

Shinigami92 (approved):

test/fake.spec.ts
xDivisionByZerox: I don't know to be honest. But these tests and the expected results seem reasonable to me.

import-brain (approved):

PR #689, opened 2022-03-26
chore: use a standardized deprecated function

Shinigami92:
closes #645


src/deprecated.ts
```diff
@@ -0,0 +1,21 @@
+export interface FunctionReplacement {
+  old: string;
+  new: string;
+}
+
+export type DeprecatedOptions = string | FunctionReplacement;
```
pkuczynski, line 6: Looking at your comment after team discussion: https://github.com/faker-js/faker/pull/645#issuecomment-1078783669

> It should have parameters for `version`, `alternative` and the possibility to override the default message.

src/deprecated.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Logs a warning message to the console prefixed with `Deprecation Warning: `.
+ *
+ * @param deprecatedOptions The options to log.
```
pkuczynski, line 11: In other JSDoc we present all options with examples...

src/deprecated.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function deprecated(deprecatedOptions: DeprecatedOptions): void {
+  if (typeof deprecatedOptions === 'string') {
+    console.warn(`Deprecation Warning: ${deprecatedOptions}`);
+  } else {
+    console.warn(
+      `Deprecation Warning: ${deprecatedOptions.old} is now located in ${deprecatedOptions.new}`
```
pkuczynski, line 18: A personal opinion: I think it would sound better written as: 

```js
`${deprecatedOptions.old} is deprecated. Please use ${deprecatedOptions.new} instead.`
```

Without the `Deprecation Warning:` in front.

src/helpers.ts
```diff
@@ -511,8 +513,8 @@ export class Helpers {
    * @deprecated If you need some specific object you should create your own method.
    */
   createCard(): Card {
-    console.warn(
-      'Deprecation Warning: If you need some specific object you should create your own method.'
+    deprecated(
```
pkuczynski, line 516: I think this message might be hard to narrow down in the larger code. We should express more explicitly what's deprecated - in this case `helpers.createCard`

src/name.ts
```diff
@@ -24,8 +25,8 @@ function normalizeGender(
 
   const normalizedGender = gender === 0 ? 'male' : 'female';
 
-  console.warn(
-    `Deprecation Warning: Please use '${normalizedGender}' for gender instead of ${gender}`
+  deprecated(
+    `Please use '${normalizedGender}' for gender instead of ${gender}`
```
pkuczynski, line 29: Similarly here. I can imagine someone would upgrade to faker 6.1 in some bigger project and run it. I think he would struggle big time to understand that those deprecation messages come from `faker`...

pkuczynski (changes requested):

src/deprecated.ts
Shinigami92: Do we really need this also for internal APIs? This function will be never exposed publicly 🤔

But I could add `@internal`

src/deprecated.ts
Shinigami92: Yeah, I can add that tomorrow 
What would you think how a message could be structured?

src/helpers.ts
Shinigami92: Great suggestion 👍
I will change that tomorrow

src/deprecated.ts
pkuczynski: If you put it like this, what's the point of writing JSDoc for it at all? :)

src/deprecated.ts
pkuczynski: Maybe something like this:

```js
Faker.js: ${opts.deprecated} is deprecated.
Faker.js: ${opts.deprecated} is deprecated and will be removed in ${opts.until}
Faker.js: ${opts.deprecated} is deprecated. Please use ${opts.proposed} instead.
Faker.js: ${opts.deprecated} is deprecated since ${opts.since}. Please use ${opts.proposed} instead.
Faker.js: ${opts.deprecated} is deprecated and will be removed in ${opts.until}. Please use ${opts.proposed} instead.
Faker.js: ${opts.deprecated} is deprecated since ${opts.since} and will be removed in ${opts.until}. Please use ${opts.proposed} instead.
```

src/deprecated.ts
Shinigami92: Make the linter happy 😏

pkuczynski: We can adjust the linter to expect jsdoc only on exported functions :)

Shinigami92: Yes, lets do this

Shinigami92: Oh, but how do we do this for this function? It is exported, but not made public via the bundling process :confused:

Shinigami92:
Before I update the tests, @pkuczynski could you have a new look?

src/deprecated.ts
pkuczynski: Put all internal functions into a subfolder (eg. `lib`) and then use `overrides` in eslint config to tweak settings for this particular folder

src/deprecated.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Logs a warning message to the console.
+ *
+ * @param opts The options to log.
+ *
+ * @internal
+ */
+export function deprecated(opts: DeprecatedOptions): void {
+  if (opts.since && opts.until && opts.proposed) {
```
pkuczynski, line 16: Functional wise it looks good, but code-wise there is a lot of repetition. You could do it way simpler I think. Something like:

```js
let message = `[@faker-js/faker]: ${opts.deprecated} is deprecated`

if (opts.since) {
  message += ` since ${opts.since}`
}

if (opts.until) {
  message += ` and will be removed in ${opts.until}`
}

if (opts.propsed) {
  message += `. Please use ${opts.proposed} instead`
}

return message + '.'
```

pkuczynski (dismissed):

pkuczynski:
> Before I update the tests, @pkuczynski could you have a new look?

Except for the single comment I left, it all looks good to me! :) Still think annotation would work better ;)

src/deprecated.ts
Shinigami92: Seems I cannot use
```
  overrides: [
    {
      files: ['src/**/*.ts', '!src/internal/**/*.ts'],
```

Results in infinit long waiting when running `pnpm run lint`

So I will ignore rules in file

pkuczynski: As we only need jsdoc on the _root_ files, change the line to `files: ['src/*.ts']` so it does not go deep

tsconfig.json
```diff
@@ -12,7 +12,8 @@
     // "strictPropertyInitialization": true,
     // "noImplicitAny": true,
     // "noImplicitThis": true,
-    "useUnknownInCatchVariables": true
+    "useUnknownInCatchVariables": true,
+    "stripInternal": true
```
pkuczynski, line 16: > This is an internal compiler option; use at your own risk, because the compiler does not check that the result is valid. If you are searching for a tool to handle additional levels of visibility within your d.ts files, look at [api-extractor](https://api-extractor.com/).

Do we really want this? Hmm, I would not do that tbh.

src/deprecated.ts
Shinigami92: This will change later, cause we may plan to move more stuff into folder later (v7+)

tsconfig.json
Shinigami92: > Do not emit declarations for code that has an @internal annotation in its JSDoc comment.

It's only for type declaration files, so I think it's safe :thinking:

pkuczynski: I am only citing official docs and it does not seem like recommended...

ST-DDT:
IMO we should always include the `()` in the message to make it easier to see that it refers to a method call

xDivisionByZerox (dismissed):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #695, opened 2022-03-27
feat: faker.finance.pin()

### LINKED ISSUE #472: PIN, account number

I need to generate 4-letter PIN numbers. Since `faker.datatype.number` cannot generate leading zeroes, I was looking at `faker.finance.account` which does exactly this thing. But it doesn't look right to me to use bank account generator for PIN, the bank account format might be localized in the future, since different countries use different formats.
Examples:
- [Belgium](https://www.ecbs.org/iban/belgium-bank-account-number.html): 539-0075470-34
- [Czechia](https://www.ecbs.org/iban/czech-replublic-bank-account-number.html), [Slovakia](https://www.ecbs.org/iban/slovak-republic-bank-account-number.html): 19-2000145399/0800
- [Italy](https://www.ecbs.org/iban/italy-bank-account-number.html): X 05428 11101 000000123456
- [Malta](https://www.ecbs.org/iban/malta-bank-account-number.html): 12345MTLCAST001S
- Some other countries is Europe use BBAN (part of IBAN) as their national format, many more have spaces or dashes in their account numbers.

### Suggested solution

I suggest we create a new generator `faker.finance.pin` or `faker.internet.pin` which generates string of numbers of provided length.
As for the bank account formats, we might start localizing them.

### Alternative

_No response_

### Additional context

_No response_

import-brain:
fixes #472


src/finance.ts
```diff
@@ -290,6 +290,22 @@ export class Finance {
...
+   * Generates a random PIN number.
+   *
+   * @param digits The number of digits to generate. Defaults to 4.
+   *
+   * @example
+   * faker.finance.pin() // '506'
+   */
+  pin(digits: number = 4): string {
+    let pin = '';
+    for (let i = 0; i < digits; i++) {
```
pkuczynski, line 303: Why not `'#'.repeat(digits)`?

pkuczynski (changes requested):

src/finance.ts
```diff
@@ -290,6 +290,22 @@ export class Finance {
     return cvv;
   }
 
+  /**
+   * Generates a random PIN number.
+   *
+   * @param digits The number of digits to generate. Defaults to 4.
+   *
+   * @example
+   * faker.finance.pin() // '506'
```
xDivisionByZerox, line 299: The example should be a 4 character string since that's the default length.

import-brain: Whoops, forgot to change the jsdoc😅

src/finance.ts
xDivisionByZerox: I'm not even sure why we need that extra step instead of choosing a random number right here and appending it.

src/finance.ts
```diff
@@ -290,6 +290,22 @@ export class Finance {
...
+   */
+  pin(digits: number = 4): string {
+    let pin = '';
+    for (let i = 0; i < digits; i++) {
+      pin += '#';
+    }
+    return this.faker.helpers.replaceSymbols(pin);
```
pkuczynski, line 306: What this function does is `this.faker.datatype.number(9)`. So maybe instead all of this extra processing and conversion, you should simply do:

```
Array.from({ length }, () => this.faker.datatype.number(9)).join('')
```

(after renaming `digits` to `length`




```

File: src/finance.ts (hunk 1 above)
Comment (pkuczynski) [reply src/finance.ts:303]: I agree! See my comment below...

File: src/finance.ts (hunk 3 above)
Comment (import-brain) [reply src/finance.ts]: Wow, I didn't realize you could do this, thank you so much!

**Comment by xDivisionByZerox** [comment]
What would be the expected behavior when a negative length is requested?

**Comment by import-brain** [comment]
> What would be the expected behavior when a negative length is requested?

I didn't consider that scenario. Do you think it should throw an error?

**Review by Shinigami92 — DISMISSED** [review]

**Comment by pkuczynski** [comment]
> > What would be the expected behavior when a negative length is requested?
> 
> I didn't consider that scenario. Do you think it should throw an error?

I personally think it should throw when `< 1`...

File: src/finance.ts [hunk 4]
```diff
@@ -290,6 +290,18 @@ export class Finance {
     return cvv;
   }
 
+  /**
+   * Generates a random PIN number.
+   *
+   * @param length The length of the PIN to generate. Defaults to 4.
+   *
+   * @example
+   * faker.finance.pin() // '5067'
```
pkuczynski, line 299: You should provide an example when `length` is given...

test/finance.spec.ts
```diff
@@ -452,6 +456,33 @@ describe('finance', () => {
         });
       });
 
+      describe('pin()', () => {
+        it('should return a string', () => {
+          const pin = faker.finance.pin();
+          expect(pin).toBeTypeOf('string');
+        });
+
+        it('should have all characters be a digit', () => {
```
pkuczynski, line 465:
```suggestion
        it('should contain only digits', () => {
```

pkuczynski (changes requested):

xDivisionByZerox (dismissed):

src/finance.ts
```diff
@@ -290,6 +290,23 @@ export class Finance {
...
+   * @param length The length of the PIN to generate. Defaults to 4.
+   * @throws Will throw an error if length is less than 1.
+   *
+   * @example
+   * faker.finance.pin() // '5067'
+   * faker.finance.pin(6) // '213789'
+   */
+  pin(length: number = 4): string {
+    if (length < 1) {
+      throw new Error('PIN length must be 1 or greater');
```
pkuczynski, line 305:
```suggestion
      throw new Error('mnimum length is 1');
```

pkuczynski (dismissed):

src/finance.ts
```diff
@@ -290,6 +290,23 @@ export class Finance {
     return cvv;
   }
 
+  /**
+   * Generates a random PIN number.
+   *
+   * @param length The length of the PIN to generate. Defaults to 4.
```
ST-DDT, line 296:
```suggestion
   * @param length The length of the PIN to generate. Defaults to `4`.
```

src/finance.ts
```diff
@@ -290,6 +290,23 @@ export class Finance {
...
+   * @throws Will throw an error if length is less than 1.
```
ST-DDT, line 297: IIRC we usually dont document the errors.

import-brain: ```finance.iban()``` documents the errors, so I went with that

ST-DDT (approved):

pkuczynski (dismissed):

src/finance.ts
```diff
@@ -291,6 +291,23 @@ export class Finance {
...
+   * @param length The length of the PIN to generate. Defaults to `4`.
+   * @throws Will throw an error if length is less than 1.
+   *
+   * @example
+   * faker.finance.pin() // '5067'
+   * faker.finance.pin(6) // '213789'
+   */
+  pin(length: number = 4): string {
+    if (length < 1) {
+      throw new Error('minimum length is 1');
```
Shinigami92, line 306: This needs to be a `FakerError`

Shinigami92 (changes requested):

pkuczynski (approved):

Shinigami92 (approved):

ST-DDT (approved):

PR #697, opened 2022-03-27
fix: use require/import export map in package.json

### LINKED ISSUE #696: Node w/ ESM uses CJS files
### Describe the bug

I'm trying to compile my NodeJS app with ESM. When I build with esbuild, it selects the CJS build of Faker instead of the ESM version.

image

```
    Checking for a package in the directory "/Users/cyber/dev/platform/node_modules/@faker-js/faker/locale/en_US"
    Read 8 entries for directory "/Users/cyber/dev/platform/node_modules/@faker-js/faker"
    Looking for "./locale/en_US" in "exports" map in "/Users/cyber/dev/platform/node_modules/@faker-js/faker/package.json"
      Checking object path map for "./locale/en_US"
      The key "./locale/*" matched with "en_US" left over
      Checking condition map for one of ["default", "import", "node"]
        The key "node" applies
        Checking path "en_US" against target "./dist/cjs/locale/*.js"
          Substituted "en_US" for "*" in "./dist/cjs/locale/*.js" to get "./dist/cjs/locale/en_US.js"
      The resolved path "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale/en_US.js" is exact
      Read 56 entries for directory "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale"
      Resolved to "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale/en_US.js"
  Read 56 entries for directory "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale"
  Ignoring "/Users/cyber/dev/platform/node_modules/@faker-js/faker/tsconfig.json" because "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale/en_US.js" is inside "node_modules"
  Primary path is "/Users/cyber/dev/platform/node_modules/@faker-js/faker/dist/cjs/locale/en_US.js" in namespace "file"
```

### Reproduction

I'm using esbuild with `--bundle --target node14 --mainFields module,main --format esm`

### Additional Info

I believe the problem is in faker's package.json:
```json
{
"exports": {
    ".": {
      "node": "./dist/cjs/index.js",
      "es2015": "./dist/esm/index.js",
      "default": "./dist/esm/index.js"
    },
    "./locale/*": {
      "node": "./dist/cjs/locale/*.js",
      "es2015": "./dist/esm/locale/*.js",
      "default": "./dist/esm/locale/*.js"
    },
  },
}
```

It appears to assume if I'm using node I want commonjs. I'm using node and I want ESM. 
I think it should be using `require` vs `import` - see the nodejs docs here https://nodejs.org/api/packages.html#conditional-exports

Then it will select the proper version depending on if the consumer is using `require` or `import`. No assumptions needed.

revmischa:
Closes #696 - please see there for more details

This is to make the export map compliant with NodeJS documentation and recommendations: https://nodejs.org/api/packages.html#conditional-exports

I believe the description and example there is pretty clear. It also mentions that using `node` as a mapping is usually wrong:
> "node" - matches for any Node.js environment. Can be a CommonJS or ES module file. In most cases explicitly calling out the Node.js platform is not necessary.


Shinigami92 (changes requested):
This does not work right now :thinking: 

I get 
```node
file:///home/shinigami/OpenSource/faker.js-tests/esm/index.js:2
import { faker } from "@faker-js/faker";
         ^^^^^
SyntaxError: Named export 'faker' not found. The requested module '@faker-js/faker' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@faker-js/faker';
const { faker } = pkg;
```

Please checkout the https://github.com/faker-js/playground (and comment out the testings for #642)

Shinigami92 (approved):
Tested locally with the playground, it's working :+1:

pkuczynski (approved):

Shinigami92:
Please unlock your repo / this PR so maintainers can make changes / rebase the main branch into it.
We wanted to release this in v6.1.0 but now we are stuck and need to wait on you :slightly_frowning_face:

PR #701, opened 2022-03-28
docs: mention template literals in fake docs

ST-DDT:
Created to clarify the usage recommendations of `faker.fake()`:
https://discord.com/channels/929487054990110771/957954416286261288

### DIFF
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -17,7 +17,10 @@ export class Fake {
   /**
    * Generator method for combining faker methods based on string input.
    *
-   * This will check the given string for placeholders and replace them by calling the specified faker method.
+   * Note: If you just want to create a string on the fly, we recommend using string template literals instead.
+   * This method is useful if you wish to choose a random format from a list or persistent storage (json etc.).
+   *
+   * It checks the given string for placeholders and replace them by calling the specified faker method.
    * E.g. the input `Hi, my name is {{name.firstName}}!`,
    * will use the `faker.name.firstName()` method to resolve the placeholder.
    * It is also possible to combine static text with placeholders,
--- a/src/fake.ts
+++ b/src/fake.ts
@@ -18,7 +18,7 @@ export class Fake {
    * Generator method for combining faker methods based on string input.
    *
    * Note: If you just want to create a string on the fly, we recommend using string template literals instead.
-   * This method is useful if you wish to choose a random format from a list or persistent storage (json etc.).
+   * This method is useful if you wish to choose a random format from a non-executable source or persistent storage (json etc.).
    *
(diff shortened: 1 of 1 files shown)


src/fake.ts
```diff
@@ -17,7 +17,10 @@ export class Fake {
   /**
    * Generator method for combining faker methods based on string input.
    *
-   * This will check the given string for placeholders and replace them by calling the specified faker method.
+   * Note: If you just want to create a string on the fly, we recommend using string template literals instead.
+   * This method is useful if you wish to choose a random format from a list or persistent storage (json etc.).
```
xDivisionByZerox, line 21: For choosing a random format from a list you could do:
```typescript
const list = [
  faker.name.firstName,
  faker.name.lastname,
  faker.name.middleName,
];
const randomFunction = faker.random.arrayElement(list);
const result = randomFunction();
console.log(result); // Hauck
```

xDivisionByZerox: > Then maybe from a non-executable source or persistent storage?

I really like that wording.

ST-DDT:
```suggestion
   * This method is useful if you wish to choose a random format from a non-executable source or persistent storage (json etc.).
```

import-brain (approved):

src/fake.ts
```diff
@@ -17,7 +17,10 @@ export class Fake {
   /**
    * Generator method for combining faker methods based on string input.
```
pkuczynski, line 18:
```suggestion
   * Generator method for combining faker methods based on static string input.
```

src/fake.ts
```diff
@@ -17,7 +17,10 @@ export class Fake {
...
    *
-   * This will check the given string for placeholders and replace them by calling the specified faker method.
+   * Note: If you just want to create a string on the fly, we recommend using string template literals instead.
```
pkuczynski, line 20:
```suggestion
   * Note: If you want to create a string combining a few different faker elements, we recommend using string template literals instead of `fake`, eg. `const address = `${faker.address.zipCode} ${faker.address.city}``
```

Shinigami92 (approved):

src/fake.ts
```diff
@@ -17,7 +17,10 @@ export class Fake {
...
+   * This method is useful if you wish to choose a random format from a non-executable source or persistent storage (json etc.).
```
pkuczynski, line 21:
```suggestion
   * This method is useful though if you wish to build a random string from a static, non-executable source (eg string coming from a user, stored in a database or a file).
```

pkuczynski (commented):
Some more suggestions...

Issue #694, opened 2022-03-27
deprecate `phone.phoneFormat`


xDivisionByZerox:
### Clear and concise description of the problem

Phone formats should not be exposed directly.

### Suggested solution

Deprecate `phone.phoneFormat` in favor of `phone.phoneNumber`. 

### Additional context

Marked as TODO in #541

ST-DDT:
IMO we should not expose the format directly.

xDivisionByZerox:
I agree with you on that (if you really need to access them you could simply `faker.definitions.phone_number.formats`). Should I rephrase this issue to remove the util functions that expose phone formats and mark for discussion?

PR #710, opened 2022-03-28
Update phone.ts

kum9748ar:
renamed `phoneFormats()` to `formats`
Ref : https://github.com/faker-js/faker/issues/694

### DIFF
--- a/src/phone.ts
+++ b/src/phone.ts
@@ -55,7 +55,7 @@ export class Phone {
    * faker.phone.phoneFormats() // '!##.!##.####'
    */
   // TODO @pkuczynski 2022-02-01: simplify name to `format()`
-  phoneFormats(): string {
+    formats(): string {
     return this.faker.random.arrayElement(
       this.faker.definitions.phone_number.formats
     );


src/phone.ts
```diff
@@ -55,7 +55,7 @@ export class Phone {
    * faker.phone.phoneFormats() // '!##.!##.####'
    */
   // TODO @pkuczynski 2022-02-01: simplify name to `format()`
```
ST-DDT, line 57: Please remove this comment.

src/phone.ts
```diff
@@ -55,7 +55,7 @@ export class Phone {
...
-  phoneFormats(): string {
+    formats(): string {
```
ST-DDT, line 58: Please format this correctly by running `pnpm run format`.
Please create an alias with a deprecation warning that points to this method to ease migration.

ST-DDT (changes requested):

ST-DDT:
We haven't decided whether we actually want to expose the method in general. So putting this on hold.

src/phone.ts
```diff
@@ -55,7 +55,7 @@ export class Phone {
    * faker.phone.phoneFormats() // '!##.!##.####'
```
Shinigami92, line 55: This would also need to be updated

kum9748ar: @Shinigami92 I am having some issues writing the commit message can you help me 
image
out

Shinigami92: Seems your commit message is not valid
You should stick to our message format and use something like `chore: your message`

ST-DDT:
Superseded by #724

PR #713, opened 2022-03-28
Changed deprecation warning

kum9748ar:
phoneFormats() to formats()
Changed deprecation warning
ref : https://github.com/faker-js/faker/issues/694


ST-DDT:
Duplicate of #710

PR #714, opened 2022-03-28
feat(word): add sample method

### LINKED ISSUE #339: Move random.word to word.any

Most of the methods of random already found their new home, like `random.number` > `datatype.number`

### Suggested solution

Move `random.word` to `word.any` and deprecate `random.word`.

### Alternative

_No response_

### Additional context

https://github.com/faker-js/faker/pull/276#issuecomment-1024251316

pkuczynski:
Fixes #276 

Fixes #339


src/word.ts
```diff
@@ -58,6 +58,40 @@ export class Word {
     );
   }
 
+  /**
+   * Returns a word of random or optionally specified length from
+   * list of adjectives, adverbs, nouns, prepositions or verbs.
+   * If specified length is unresolvable, returns word of a random length.
+   *
+   * @param length The optional length of word to return.
+   */
```
ST-DDT, line 67: `@example`s are missing

src/word.ts
```diff
@@ -58,6 +58,40 @@ export class Word {
...
+
+    let result = '';
+
+    do {
+      const index = this.faker.datatype.number({
+        min: 0,
+        max: generators.length - 1,
+      });
+
+      result = generators[index](length);
```
ST-DDT, line 85: Use `arrayElement()` instead

ST-DDT (changes requested):

src/word.ts
pkuczynski: I am not using `arrayElement`, because if the generator can't satisfy the given length, I remove it from the list and try with another one. So it needs to stay as it is... Makes sense?

src/word.ts
pkuczynski: omg! Of course!

pkuczynski: I actually copied it from the other method without noticing that all of them come without examples! I will add them in a separate PR...

src/word.ts
ST-DDT: The source/array selection shouldn't be impacted by this at all. 

AFAICT `arrayElement(generators)(length)` would still behave the same.

ST-DDT (approved):

src/word.ts
pkuczynski: Ok, let me explain then. If I call `arrayElement(generators)(length)` it will pick up a random generator, which might not be able to satisfy the given length. In such a case, I would like to eliminate this generator and try with a random one from the remaining list. How would I do that using purely `arrayElement`?

pkuczynski: Ah, didnt noticed you added `nvm` later... Was following what come with email notification, sorry. Case closed.

src/word.ts
```diff
@@ -58,6 +58,44 @@ export class Word {
...
+
+      if (length && generators.length > 0) {
+        generators.splice(index, 1);
+      }
+    } while (length && result.length < length && generators.length > 0);
```
Shinigami92, line 94: We should extract the `length` check out of the do-while, so it doesn't run once even if length is 0

test/word.spec.ts
```diff
@@ -204,6 +220,49 @@ describe('word', () => {
...
+          const length = 15;
+          const actual = faker.word.any(length);
+
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+          expect(actual.length).toBe(length);
+          expect(words).toContain(actual);
+        });
+
+        it('should return word with random length when given length can not be satisfied', () => {
```
Shinigami92, line 254: Not for this PR, but IMO it feels a bit weird if a word with given length is not found then use random length 🤔 
IMO we should change behavior for all such cases to a min/max length and search between that. If nothing found return `undefined` or throw an error?
I will open a proposal issue for that... we need to think about how we can handle that

Shinigami92 (changes requested):

src/random.ts
```diff
@@ -254,8 +254,17 @@ export class Random {
    *
    * @example
    * faker.random.word() // 'Seamless'
+   *
+   * @deprecated
    */
   word(): string {
+    deprecated({
+      deprecated: 'faker.random.word()',
+      proposed: 'faker.word.any()',
```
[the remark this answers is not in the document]
pkuczynski: Added this following your request on https://github.com/faker-js/faker/pull/276#issuecomment-1078777865

Or do you reffer to `proposed` only? Can you suggest some alternative?

src/word.ts
pkuczynski: It should run once even if length is `0`, as it needs to pick up first result regardless of length, right?

test/word.spec.ts
pkuczynski: Yes, this is a bit weird I agree, but I just followed the other methods from this class. Felt that consistency here would be important? Maybe we should call this param `maxLength`?

src/random.ts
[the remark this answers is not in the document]
Shinigami92: Yes I only mean the `proposed` param

test/word.spec.ts
Shinigami92: We should change all at once, so for now I think it's okay to follow all the other functions

src/word.ts
Shinigami92: Ah I see

ST-DDT, line 94: All other methods in this module use `=== length`. This method should follow suite.

ST-DDT (changes requested):

src/word.ts
ST-DDT: I'm fine with extending all methods later with a `length?: number | { min: number, max: number}` in a later PR.

pkuczynski: It is using `===` for length :) So basically if the length condition is not met, it will do another loop picking up the next random generator, unless there are no more generators left and then it keeps the last result.

src/random.ts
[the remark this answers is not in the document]
pkuczynski: I replaces with a new `deprecated` prop called `similar`. What do you think?

src/word.ts
ST-DDT:
```suggestion
    } while (length && result.length !== length && generators.length > 0);
```

ST-DDT (changes requested):

ST-DDT (dismissed):

pkuczynski:
@Shinigami92 anything else missing to get this merged?

Shinigami92:
Yes, all the v6.1 targeted things to be merged I assume 🤔

pkuczynski:
> Yes, all the v6.1 targeted things to be merged I assume 🤔

Not sure if we should be so stick to the _6.1 is only bug fixes_ as keeping PRs open for longer time means potential conflicts and maintenance, so I would be in favor of merging things as soon as they ready...

ST-DDT:
> > Yes, all the v6.1 targeted things to be merged I assume 🤔
> 
> Not sure if we should be so stick to the _6.1 is only bug fixes_ as keeping PRs open for longer time means potential conflicts and maintenance

That is true. Is is a small feature, requiring low effort for maintaining the PR from you.

> so I would be in favor of merging things as soon as they ready...

If we merge this one, then we have to explain why we didn't merge other features.
Sure we could start merging them all, but then everyone would start creating new PRs with new super fancy features
and we have to somehow interweave the required bug fix merges with the nice to have feature requests.
Also people tend to prefer adding features over fixing bugs.

We use the milestones as a means to communicate the order we discussed on when to tackle things/fix/add...
This feature being in v6.2 has been discussed/communicated a few months ago, we don't want to discourage anyone in contributing to the project, so we don't ban PRs for future milestones, but I hope you can understand that this may cause you additional work to maintain them for the time being.

We hope to finish all fixes/PRs for 6.1 by this monday evening, so you don't have to maintain it forever either.

pkuczynski:
Fair enough :)

ST-DDT:
**To consider:** Handle "empty" locale
This method should at least not behave worse than the other word methods.
Ref: https://github.com/faker-js/faker/issues/770#issuecomment-1087249521

src/random.ts
```diff
@@ -259,8 +259,17 @@ export class Random {
...
    * @example
    * faker.random.word() // 'Seamless'
+   *
+   * @deprecated
    */
   word(): string {
+    deprecated({
+      deprecated: 'faker.random.word()',
+      similar: 'faker.word.any()',
+      since: 'v6.1.0',
```
Shinigami92, line 269: The upcoming release is v6.2.0
```suggestion
      since: 'v6.2.0',
```

src/word.ts
```diff
@@ -66,6 +66,44 @@ export class Word {
...
+      this.faker.word.verb,
+    ];
+
+    let result = '';
+
+    do {
+      const index = this.faker.datatype.number({
+        min: 0,
+        max: generators.length - 1,
+      });
```
Shinigami92, line 95: I might be wrong, but I assume this can be simplified to

```suggestion
      const index = this.faker.datatype.number(generators.length - 1);
```

Shinigami92 (changes requested):

xDivisionByZerox:
This PR is really outdated. I guess its easier to close this and then recreate it on the current main branch.
@pkuczynski Are you ok with this? Do you want to do that? Otherwise, I can offer you my help.

src/random.ts
Shinigami92: That's really old :laughing:

src/modules/word/index.ts
```diff
@@ -171,4 +171,87 @@ export class Word {
...
+  /**
+   * Returns a string containing a number of space separated random words.
+   *
+   * @param count Number of words. Defaults to a random value between `1` and `3`.
+   *
+   * @example
+   * faker.word.words() // 'almost'
+   * faker.word.words(5) // 'before hourly patiently dribble equal'
+   */
+  words(count?: number): string {
```
ST-DDT, line 246: Not sure about this name.

src/modules/word/index.ts
```diff
@@ -171,4 +171,87 @@ export class Word {
...
+  /**
+   * Returns random word.
+   *
+   * @param length Expected word length. If specified length is unresolvable, returns verb of a random length.
+   *
+   * @example
+   * faker.word.random() // 'incidentally'
+   * faker.word.random(5) // 'fruit'
+   */
+  random(length?: number): string {
```
ST-DDT, line 184: Not sure about this name.

I eventually want a "random" method for some more modules.
(string.random, animal.random)

I prefer random over any, because it feels better when used with options.

````ts
faker.string.random({ allowedChars: [], bannedChars: []);
faker.string.any({ allowedChars: [], bannedChars: []);

faker.animal.random();
faker.animal.any();
````

**Comment by ST-DDT** [comment]
I updated this branch to retain the original contribution references to @pkuczynski .

**Comment by ST-DDT** [comment]
Test failures -> https://github.com/faker-js/faker/issues/1317 + https://github.com/faker-js/faker/issues/1318

File: src/modules/word/index.ts (hunk 8 above)
Comment (xDivisionByZerox) [reply src/modules/word/index.ts:246]: How about `many()`? Because it sounds nice when calling the entire thing: `faker.word.many()`.

Comment (pkuczynski) [reply src/modules/word/index.ts:246]: or `few()`

File: src/modules/word/index.ts (hunk 9 above)
Comment (pkuczynski) [reply src/modules/word/index.ts:184]: for me `faker.string.any({ allowedChars: [], bannedChars: []})` looks totally fine. Better than `random` as everything is `random` in faker, so it would be misleading that others methods are not. It also reads kind of naturally: _hey, give me any string. but without chars_

Comment (xDivisionByZerox) [reply src/modules/word/index.ts:184]: Well, the same goes for `random`. *hey, give me a random string. but without chars.*
TBH, I don't care as long as it's consistent across (official) modules. But as a developer, I guess I would search for `random` first (intellisense and documentation).

Comment (pkuczynski) [reply src/modules/word/index.ts:184]: Wouldn't you then expect `faker.animal.bear()` not to be random? :)

File: src/modules/word/index.ts (hunk 8 above)
Comment (import-brain) [reply src/modules/word/index.ts:246]: What about `multiple()`?

File: src/modules/word/index.ts (hunk 9 above)
Comment (xDivisionByZerox) [reply src/modules/word/index.ts:184]: I still would. TBH I don't see your problem there. I don't understand your argument:

>[...] as everything is random in faker, so it would be misleading that others methods are not

I understand, but I don't see it this way. But that is nothing we can change if you feel this way but I don't.

Comment (ST-DDT) [reply src/modules/word/index.ts:184]: Added to the meeting notes for voting.

Comment (damienwebdev) [reply src/modules/word/index.ts:184]: I've been thinking about this a lot, and I feel like `random` does not fit within given that everything in Faker is pseudorandom. When you call `word` it's somewhat implied that what you get back is random. 

In essence, we're by calling this "random", we end up with a meaning like "random random" which to me feels meaningless. 

In addition, `any` also feels wrong because it has a namespace conflict with Typescript `any`.

I spent some time thinking about this and I feel like `sample` is a decent option.

The phrase "random sample" is a term in STEM, specifically statistics. "sample", in the context of a module, indicates that you're taking a "sample" of the module. https://en.m.wikipedia.org/wiki/Sampling_(statistics)


Ultimately, I think this word is also generic and fits well with all of our modules, however I'm unclear on how exactly to type it:

```ts
export interface FakerModule {
  sample(): unknown 
}
```

```ts
export interface FakerModule {
  sample(): UnionOfReturnTypeOfAllPublicMethodsOfImplementationclass??;
}
```

ST-DDT: I think sample is a good term.

As for having a sample method for all modules, I'm a little bit conflicted.
I would not add a sample method to modules that return different kind of data such as the address and number module. 
The address module returns mostly strings but these have significantly different semantics e.g. building number, street name, coutry name, gps coordinates. Which would make the output random instead of being a sample.
The "number" module returns different datatypes e.g. number, BigInt, hex strings... which renders the output mostly unusable.
For e.g. the word or animal module the sample method seems suitable.

Shinigami92:
To be fair we would need to split this PR in two parts if @xDivisionByZerox's recommendation to have simple responsibilities by a PR counts for for this PR as well

So one PR introducing the new method `sample`
And one for deprecating the old method(s)

@xDivisionByZerox's point was also about that the commit history is more clean and if e.g. something need to be reverted, the commit in the default branch history can be better linked

ST-DDT:
@Shinigami92 I think you are mistaken here.
If we replace one thing with another, then we do it in one step (and always have been).
(Same as we did with renaming the name->person module, address->location, fake->helpers, phoneNumber->number, ...)

Shinigami92:
> @Shinigami92 I think you are mistaken here. If we replace one thing with another, then we do it in one step (and always have been). (Same as we did with renaming the name->person module, address->location, fake->helpers, phoneNumber->number, ...)

But is this really a one to one replacement? I thought the code differ much from what the original code did.
The [`wordMethods`](https://github.com/faker-js/faker/pull/714/files#diff-01ebf55dc10e16012b9f6e0b4342dd25453163010129e4ec5ab96f7b9efb3df0L119-L181) from the old method had MUCH more possible values.
You can also see that the `random.word` snapshot results did change drastically (not only one offset but totally different results)

ST-DDT:
So you are referring to not replacing the old implementation itself?

Shinigami92:
> referring

yes

like @xDivisionByZerox would say: this PR is doing two things
1. moving the old method to a new home
2. changing the implementation of that new method

ST-DDT:
I strongly disagree with moving the method implementation. 🙅 
The old implementation is a list of random methods that have very little in common with "words".
The word module should not gain dependencies on any external module (expect for the current helpers one).

AFAICT we have two choices here:
- Deprecate the old method and replace its implementation with a delegate to the new words.sample method.
- Deprecate the old method and retain it as it is till removal.

 (Potentially in a separate PR from the one introducing the word.sample method).

src/modules/word/index.ts
```diff
@@ -315,4 +315,110 @@ export class WordModule {
...
+      // randomly pick from the many faker methods that can generate words
+      const randomWordMethod = this.faker.helpers.arrayElement(wordMethods);
+
+      try {
+        result = randomWordMethod(options);
+      } catch {
+        // catch missing locale data potentially required by randomWordMethod
+        continue;
+      }
+    } while (!result || bannedChars.some((char) => result.includes(char)));
```
Shinigami92: Should we make this more `while(true)` safe in another PR?

src/modules/random/index.ts
```diff
@@ -25,140 +25,53 @@ export class RandomModule {
   /**
    * Returns random word.
    *
+   * @see faker.word.random()
```
Shinigami92:
```suggestion
   * @see faker.word.sample()
```

ST-DDT:
**Team decision**

- This PR will only add the sample method.
- The random.word will be deprecated in a separate PR but will retain the original implementation

ST-DDT (dismissed):

src/modules/word/index.ts
ST-DDT: Yes, since I would like to automatically only take existing word categories into account.

````ts
    const opts = typeof options === 'number' ? { length: options } : options;
    return this.faker.helpers.arrayElement(
      filterWordListByLength({
        ...opts,
        wordList: Object.values(this.faker.definitions.word).flatMap((v: string) => v).flatMap((v) => v.split(/ -/)),
      })
    );
````

But `Object.values(this.faker.definitions.word)` is currently not supported.

Comment (ST-DDT) [reply src/modules/word/index.ts]: Nevermind. I found a solution to avoid this problem.
We might want to add an additional filter option later though.

**Comment by ST-DDT** [comment]
Done. Ready for Review.

**Review by ST-DDT — DISMISSED** [review]

**Review by ejcheng — DISMISSED** [review]

File: src/modules/word/index.ts [hunk 12]
```diff
@@ -315,4 +316,83 @@ export class WordModule {
...
+    const wordMethods = [
+      this.adjective,
+      this.adverb,
+      this.conjunction,
+      this.interjection,
+      this.noun,
+      this.preposition,
+      this.verb,
+    ];
+    this.faker.helpers.shuffle(wordMethods);
```
Shinigami92: 🙅 Oh my good, we need to rewrite `helpers.shuffle` to not mutate the input array directly, but return a copy of the array.

I will open a PR for that later on

src/modules/word/index.ts
```diff
@@ -315,4 +316,83 @@ export class WordModule {
...
+  words(count?: number): string {
+    if (count == null) {
+      count = this.faker.datatype.number({ min: 1, max: 3 });
+    }
+
+    const words: string[] = [];
+    for (let i = 0; i < count; i++) {
+      words.push(this.sample());
+    }
+    return words.join(' ');
```
Shinigami92: Can we rewrite this to this?

```suggestion
    return Array.from({ lenght: count }, () => this.sample()).join(' ');
```

Shinigami92 (approved):

ST-DDT (approved):

ejcheng (approved):

PR #718, opened 2022-03-28
feat: FakerError

### LINKED ISSUE #707: Common `FakerJSError`

When faker-js throws an error it's hard to track for end-users where this error comes from

### Suggested solution

Provide a custom FakerJsError class that prefixes all error messages with the libraries name:

```typescript
class FakerJsErrorr extends Error {
    constructor(message: string) {
        super('[faker-js/faker]: ', + message);
    }
}
```

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
closes #707


src/internal/faker-js.error.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @internal
+ */
+export class FakerJsError extends Error {
```
pkuczynski, line 4: Why not `FakerError`? I find it quite odd to see the `Js` in the middle of it.

src/internal/faker-js.error.ts
```diff
@@ -0,0 +1,8 @@
...
+  constructor(message: string) {
+    super('[@faker-js/faker]: ' + message);
```
pkuczynski, line 6: I have doubts if we should prepend the error message with the library type. I have not seen this before. If a developer wants to know when the issue happens, he will see this in the stack trace. And such errors should not leak to end users...

test/unique.spec.ts
```diff
@@ -108,7 +109,7 @@ describe('unique', () => {
               maxRetries: 9999,
               exclude: ['https', 'http'],
             });
-          }).toThrowError(/^Exceeded maxTime:/);
+          }).toThrowError(/Exceeded maxTime:/);
```
pkuczynski, line 112: In some tests we check `.toThrowError(new FakerJsError(...))` in some others we use regex. This should be aligned. If we introduce custom error, we should use the first syntax.

pkuczynski (changes requested):

src/internal/faker-js.error.ts
xDivisionByZerox: Was just going with the package name as default. I have no problems renaming it to `FakerError`.

src/internal/faker-js.error.ts
xDivisionByZerox: Well, that's the entire point of this error class. It basically mimics the newly implemented [`deprecated()`](https://github.com/faker-js/faker/blob/main/src/internal/deprecated.ts#L14) log just as an error implementation. 

Well yes and no. By end-users you mean clients? Or developers? I agree on it being visible in the stack trace. But what if you use a bundler and your stack trace resolves to gibberish?

test/unique.spec.ts
xDivisionByZerox: The problem with providing an instance is that the error message has to match exactly (for my understanding) and not partialy.

src/internal/faker-js.error.ts
pkuczynski: `deprecated` is a bit different as there is no stack trace... I would assume that anybody willing to watch such an error would use source maps, otherwise debugging is a real pain. Looking at the other libraries, I haven't really seen such prepending on error messages, but maybe you can point us to some examples?

src/internal/faker-js.error.ts
ST-DDT: I prefer `FakerError` as well.

Shinigami92: I also would like to name it `FakerError`, we also do **not** export the normal `faker` instance as `import { fakerJs } from '@faker-js/faker'` 🤷

src/internal/faker-js.error.ts
Shinigami92: IMO @pkuczynski has a point here, but it's still good to have this `Error` because the `name` of it will reflect the needed info.

So a stacktrace would contain `FakerError: ...message...` and then the stacktrace

Could you test somehow once if this is like I think?

test/unique.spec.ts
Shinigami92: When using this own error, we should test for `.toThrowError(FakerError)` and maybe we just need to test the message in an additional step.

src/internal/faker-js.error.ts
pkuczynski: Ah, I just noticed it's not exported in the main module. I think it should be exported. @Shinigami92 would you agree?

xDivisionByZerox: Just asking so I'm sure I did understand correctly:

> So a stacktrace would contain FakerError: ...message... and then the stacktrace

The custom error class is helpful because tt provides a unique constructor instance that:
 - can be checked against and handled appropriately
 - will be displayed in the error stack and thus can be traced back to faker

> Ah, I just noticed it's not exported in the main module. I think it should be exported

The `@interal` decorator should be removed (?)

pkuczynski: Yes, I think we should remove `@internal`, move it out of `internal` folder, add jsdoc and export in `index.ts`...

xDivisionByZerox: I think it's a bit misleading to have the error file lying between all the module files. Should I create an `src/error/` directory and store the file there (and export from index.ts)

test/unique.spec.ts
xDivisionByZerox: This was already done in the original push in the next test case.

src/internal/faker-js.error.ts
xDivisionByZerox: `error` or `errors`? I wrote `src/error/` but its a directory/collections, so....

Shinigami92: We have definitions, image_providers and locales, so I think errors is the way to go

test/unique.spec.ts
xDivisionByZerox: Removed `@internal` decorator in 2e9d4a956bbfe50498076c6b244242a559b81c11
Moved in ccf11cdfebb3f84daf29d8eb3ad4d7218b17cdb3

src/internal/faker-js.error.ts
xDivisionByZerox: Removed `@internal` decorator in 2e9d4a956bbfe50498076c6b244242a559b81c11
Moved file to error dir in ccf11cdfebb3f84daf29d8eb3ad4d7218b17cdb3

test/unique.spec.ts
ST-DDT: IMO you can restore the original regex here and below, since we no longer have the custom prefix.

ST-DDT (changes requested):

test/unique.spec.ts
xDivisionByZerox: Yeah, my bad. Didn't even noticed that change in the RegEx.
Fixed in 69a84d25bdafe7ca53f2908eed0ab44291f87a77

Shinigami92 (changes requested):
Please re-export `FakerError` in `src/index.ts`

ST-DDT (approved):

src/errors/faker.error.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * An error instance that will be thrown by faker.
+ */
+export class FakerError extends Error {
```
pkuczynski, line 4: If I am not mistaken, `export class FakerError extends Error {}` should be enugh?

src/errors/faker.error.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
pkuczynski, line 1: We don't really have that pattern to name files with `.` in the middle. So I would rather expect `faker-error`. It's not a major thing though...

src/fake.ts
```diff
@@ -88,11 +89,11 @@ export class Fake {
     const parts = method.split('.');
 
     if (this.faker[parts[0]] == null) {
-      throw new Error('Invalid module: ' + parts[0]);
+      throw new FakerError('Invalid module: ' + parts[0]);
```
pkuczynski, line 92: Why not use string templates like we do in other parts of the code??

```suggestion
      throw new FakerError(`Invalid module: ${parts[0]}`);
```

pkuczynski: This applies in many places...

pkuczynski (changes requested):

src/fake.ts
xDivisionByZerox: We don't have the [prefer-template linter rule](https://eslint.org/docs/rules/prefer-template) so I don't see why this would be a required change. We do "normal" string concatenation all over the project.

src/errors/faker.error.ts
xDivisionByZerox: Change the file name in c7787c51e46b97e607384d03d04cb3fa7db2d233

src/errors/faker.error.ts
xDivisionByZerox: Changed in a26bffd96aebca686c839350a7e7520ed6017497

src/fake.ts
pkuczynski: Indeed. I will propose this change in a separate PR as we currently use it in some places already...

pkuczynski (approved):

src/fake.ts
xDivisionByZerox: Or should we rework this when you propose that change?

pkuczynski: FYI https://github.com/faker-js/faker/pull/732 - easily fixable with `eslint --fix .` so no worries, we can move forward with this PR.

xDivisionByZerox:
> Please re-export `FakerError` in `src/index.ts`

How are things exported in the root `index.ts`? Is there any import/export order?
Should there be an index file in the errors directory? IMO this would make sense. If yes, would I make implicit or explicit exports in the root index file?

Shinigami92:
@xDivisionByZerox You just want to do in `src/index.ts`

```ts
import { FakerError } from './errors/faker-error';
// ...
export { FakerError };
```

Shinigami92:
> But you know that this can be simplified, do you?
> 
> ```ts
> export{ FakerError } from './errors/faker-error';
> ```
> 
> That's why I was asking because I have seen these redundant import => export statements...

Yeah I know, but I'm not yet used to it

pkuczynski:
The export is still missing @xDivisionByZerox right?

Shinigami92 (changes requested):
index export still missing

xDivisionByZerox:
Yes the export was still missing but is implemented now.

Shinigami92 (approved):

src/errors/faker-error.ts
```diff
@@ -0,0 +1,4 @@
+/**
```
pkuczynski, line 1: In [this discussion](https://github.com/faker-js/faker/pull/733#issuecomment-1083158285) there was a suggestion to move it to `utils`

pkuczynski (approved):

src/errors/faker-error.ts
xDivisionByZerox: I didn't follow that discussion. But tbh I don't think errors should be located in utils. They should be located in their own directory, but that is my opinion. 
I would like to get an opinion  or even a specification of any @faker-js/maintainers

ST-DDT (approved):

PR #719, opened 2022-03-28
docs: add ts import example in readme

pkuczynski:
I felt that this example is missing somehow if we have examples for other envs...

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ pnpm install @faker-js/faker -D
 
 ```js
 const { faker } = require('@faker-js/faker');
+
 const randomName = faker.name.findName(); // Rowan Nikolaus
 const randomEmail = faker.internet.email(); // Kassandra.Haley@erich.biz
 const randomPhoneNumber = faker.phone.phoneNumber(); // (279) 329-8663 x30233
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,6 +96,12 @@ In order to have faker working properly, you need to check if these `compilerOpt
 }
 ```
 
+And then simply import it like everything else:
+
+```ts
+import { faker } from '@faker-js/faker';
+```
```
ST-DDT, line 103: Maybe add add a `import { fakerEn  from '@faker-js/faker/locale/en'; // Smaller` as well

pkuczynski: I could, but we don't really do that in other cases...?

import-brain (approved):

README.md
Shinigami92: @ST-DDT It's not possible to import a locale like that, you need to use `import fakerEn from '@faker-js/faker/locale/en';`

Shinigami92 (approved):
Don't have strong feelings about this suggestion
IMO it doesn't add currently no new value to the README, but if it helps others 🤷

Issue #693, opened 2022-03-27
rename `phone.phoneNumber` to `phone.number`


xDivisionByZerox:
### Clear and concise description of the problem

Function name is repeating the module name.

### Suggested solution

Create a new function `number` in the `phone` module that is a copy of the current `phoneNumber` function. Call this new function in the current function and add a deprecation warning.

### Additional context

Marked as todo in #541

import-brain:
How would we go about this? Would we simply rename all instances of ```phone.phoneNumber()``` to ```phone.number()```? Or should we keep ```phone.phoneNumber()``` and add a deprecation warning saying to use the new ```phone.number()``` function which will be a copy paste of ```phone.phoneNumber()```.

Shinigami92:
wait on #689, then yes, add deprecation warning + alias

xDivisionByZerox:
And remove completely in v7?

xDivisionByZerox:
This was resolved in #1063.

PR #722, opened 2022-03-29
chore: change some phone methods

kum9748ar:
Renamed `phoneFormats ()` to` formats()` and `PhoneNumber()` to   `number()`
- Removed todos 
- changed warnings

Ref : https://github.com/faker-js/faker/issues/694
https://github.com/faker-js/faker/issues/693

### DIFF
--- a/src/phone.ts
+++ b/src/phone.ts
@@ -20,14 +20,14 @@ export class Phone {
    * @param format Format of the phone number. Defaults to `faker.phone.phoneFormats()`.
    *
    * @example
-   * faker.phone.phoneNumber() // '961-770-7727'
-   * faker.phone.phoneNumber('501-###-###') // '501-039-841'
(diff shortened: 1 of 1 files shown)


src/phone.ts
```diff
@@ -52,10 +52,9 @@ export class Phone {
    * Returns a random phone number format.
    *
    * @example
-   * faker.phone.phoneFormats() // '!##.!##.####'
+   * faker.phone.formats() // '!##.!##.####'
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `format()`
-  phoneFormats(): string {
```
Shinigami92, line 58: We should add aliases for these and call the deprecation function, so migrations are more smooth

Shinigami92 (changes requested):

ST-DDT:
Superseded by #724

PR #726, opened 2022-03-29
feat: ```greekLetter()```

### LINKED ISSUE #558: Random greek alphabet letter

As a user I want to generate a random greek letter so that I can prepend this to a name of an object.

I would like to have the option to get a short or a long version of a letter (`α` or `'alpha'`).
Maybe `Α` and `άλφα` could also somehow get returned.

### Suggested solution

```ts
const cameraName = 'Camera ' + faker.<?>.greekLetter({ format: 'long' });

const cameraName2 = faker.helpers.mustache('Camera {{greekLetter({format:"long"})}}');
```

I'm open to other ideas

### Alternative

_No response_

### Additional context

https://en.wikipedia.org/wiki/Greek_alphabet#Letters

import-brain:
fixes #558

How would I write tests for this? I was planning on using Vitest's ```toContain``` method, but according to the Vitest docs, toContain only accepts a string parameter, and I was planning on feeding an array of Greek letters to it. My current method is a somewhat workaround, but it's super janky. Any ideas?

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -630,5 +630,6 @@ export class Random {
     } else if (options.format === 'long') {
       return this.faker.random.arrayElement(longGreekLetters);
     }
+    return this.faker.random.arrayElement(shortGreekLetters);
   }
 }
--- a/src/random.ts
+++ b/src/random.ts
@@ -557,4 +557,78 @@ export class Random {
     });
     return this.faker.datatype.hexaDecimal(count);
   }
+
+  /**
+   * Returns a Greek letter.
+   *
+   * @param options The options to use. Defaults to `{ format: 'short' }`.
+   * @param options.format The format of the letter. Defaults to `'short'`.
+   *
+   * @example
+   * faker.random.greekLetter() // 'θ'
+   * faker.random.greekLetter({ format: 'short' }) // 'ε'
+   * faker.random.greekLetter({ format: 'long' }) // 'lambda'
+   *
+   */
+  greekLetter(options?: { format: 'long' | 'short' }): string {
+    options.format = options.format || 'short';
+    const shortGreekLetters = [
+      'α',
+      'β',
+      'γ',
+      'δ',
+      'ε',
+      'ζ',
+      'η',
+      'θ',
+      'ι',
+      'κ',
+      'λ',
+      'μ',
+      'ν',
+      'ξ',
+      'ο',
+      'π',
+      'ρ',
+      'σ',
+      'τ',
+      'υ',
+      'φ',
+      'χ',
+      'ψ',
+      'ω',
+    ];
+    const longGreekLetters = [
+      'alpha',
+      'beta',
+      'gamma',
+      'delta',
+      'epsilon',
+      'zeta',
+      'eta',
+      'theta',
+      'iota',
+      'kappa',
+      'lambda',
+      'mu',
+      'nu',
+      'xi',
+      'omicron',
+      'pi',
+      'rho',
+      'sigma',
+      'tau',
+      'upsilon',
+      'phi',
+      'chi',
+      'psi',
+      'omega',
+    ];
+
+    if (options.format === 'short') {
+      return this.faker.random.arrayElement(shortGreekLetters);
+    } else if (options.format === 'long') {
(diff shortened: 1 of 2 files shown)


src/random.ts
```diff
@@ -557,4 +557,78 @@ export class Random {
...
+      'chi',
+      'psi',
+      'omega',
+    ];
+
+    if (options.format === 'short') {
+      return this.faker.random.arrayElement(shortGreekLetters);
+    } else if (options.format === 'long') {
+      return this.faker.random.arrayElement(longGreekLetters);
+    }
```
xDivisionByZerox, line 632: Missing default case. Currently, your function could return undefined.

xDivisionByZerox:
> How would I write tests for this? I was planning on using Vitest's toContain method, but according to the Vitest docs, toContain only accepts a string parameter, and I was planning on feeding an array of Greek letters to it. My current method is a somewhat workaround, but it's super janky. Any ideas?

How about getting the char code of the letter and checking against a range?

src/random.ts
```diff
@@ -557,4 +557,79 @@ export class Random {
...
+   *
+   * @example
+   * faker.random.greekLetter() // 'θ'
+   * faker.random.greekLetter({ format: 'short' }) // 'ε'
+   * faker.random.greekLetter({ format: 'long' }) // 'lambda'
+   *
+   */
+  greekLetter(options?: { format: 'long' | 'short' }): string {
+    options.format = options.format || 'short';
+    const shortGreekLetters = [
```
ST-DDT, line 575: Maybe move this to a file level const or separate locale file.

src/random.ts
```diff
@@ -557,4 +557,79 @@ export class Random {
...
+   * @param options.format The format of the letter. Defaults to `'short'`.
+   *
+   * @example
+   * faker.random.greekLetter() // 'θ'
+   * faker.random.greekLetter({ format: 'short' }) // 'ε'
+   * faker.random.greekLetter({ format: 'long' }) // 'lambda'
+   *
+   */
+  greekLetter(options?: { format: 'long' | 'short' }): string {
+    options.format = options.format || 'short';
```
ST-DDT, line 574:
```suggestion
    const format = options.format || 'short';
```

src/random.ts
```diff
@@ -557,4 +557,79 @@ export class Random {
...
+      'tau',
+      'upsilon',
+      'phi',
+      'chi',
+      'psi',
+      'omega',
+    ];
+
+    if (options.format === 'short') {
+      return this.faker.random.arrayElement(shortGreekLetters);
```
ST-DDT, line 629: Since this is the default case you can omit this.

test/random.spec.ts
```diff
@@ -278,6 +278,17 @@ describe('random', () => {
...
   });
 
+  describe('greekLetter', () => {
+    it('should return a single letter', () => {
+      const greekLetter = faker.random.greekLetter();
+      // if greekLetter does not start with an English alphabet letter,
+      // check if greekLetter has only one character (Greek letters are only one character)
+      if (!/[a-zA-Z]/.test(greekLetter)) {
+        expect(greekLetter).toHaveLength(1);
+      }
```
ST-DDT, line 288: Write separate tests for long and short versions.

src/random.ts
import-brain: I don't think we would need to create any other methods dealing with Greek letters in the future, though. Wouldn't it be easier just to keep it as-is?

Shinigami92:
I did not expect that someone suddenly start to implement this, so sorry for not inform you
We may want to have a more generic locale based function, like `letter(opts)` that returns greek letters when using `faker.locale = 'el'`
We should discuss that in discord and plan it more future-proof

import-brain:
> I did not expect that someone suddenly start to implement this, so sorry for not inform you
> 
> We may want to have a more generic locale based function, like `letter(opts)` that returns greek letters when using `faker.locale = 'el'`
> 
> We should discuss that in discord and plan it more future-proof

That's my bad, don't worry about it. I probably should have informed you earlier.

ST-DDT:
This PR has gone stale. The requirements have changed as well, so closing.

PR #728, opened 2022-03-29
chore: define const of a type instead of type casting in generateLocales script

pkuczynski:
This will start catching issues with our locales definitions...
Check #729 listing the issues...


pkuczynski:
How come this is a breaking change?

ST-DDT:
> How come this is a breaking change?

If we are strict about it, we loose like 20% of our locale data categories (Guestimated).
App, buisness, cell phone, team, ...

ST-DDT:
And I would like to check each of them, whether they can be useful for us before deleting them.

ST-DDT:
@pkuczynski I think we should change our definitions to be extensible like our base definitions:

https://github.com/faker-js/faker/blob/cc46a0c19af2752b6210c24b715fcce20197b6d9/src/definitions/definitions.ts#L54-L58

That way we would get rid of the `as X` casts, while still allowing custom entries.
Do you want to create your own PR for it or do you want me to create it?

ST-DDT:
Will be superseded by #915.

pkuczynski:
> Do you want to create your own PR for it or do you want me to create it? When that is merged, we can merge this one.

Go for it. I have super limited time at the moment :(

ST-DDT:
Fixed in #915 .

PR #729, opened 2022-03-29
chore: re-generate locales using the new script without type casting

pkuczynski:
includes #728


ST-DDT:
This will be in v7 at the earliest.

ST-DDT:
Will be superseded by #915.

ST-DDT:
Fixed in https://github.com/faker-js/faker/pull/915 .

PR #732, opened 2022-03-30
chore: prefer string templates over string concatenation

pkuczynski:
String templates are more readable, compact, and a bit faster than string concatenation. Edding eslint rule to force them.

See test results here:
https://jsben.ch/dWhr2

Or benchmark yourself:

```js
const Benchmark = require('benchmark')
const suite = new Benchmark.Suite()
const text = 'some text'
const number = 12

suite.add('concatenation', () => { const output = 'prefix' + text + 'middle' + number + 'postfix' })
suite.add('template', () => {const output = 'prefix' + text + 'middle' + number + 'postfix' })
suite.on('cycle', (event) => { console.log(String(event.target)) })
suite.on('complete', function() { console.log('Fastest is ' + this.filter('fastest').map('name')) })
suite.run({ async: true })
```

My results in node:

```
concatenation x 690,777,686 ops/sec ±1.55% (83 runs sampled)
template x 694,885,002 ops/sec ±1.58% (84 runs sampled)
Fastest is template,concatenation
```


pkuczynski:
Needs #733 no to fail...

Shinigami92 (changes requested):
Please also update _vendor_ files, we could rename the folder later if you like to

pkuczynski:
> Please also update _vendor_ files, we could rename the folder later if you like to

Check #733 which excludes vendor from eslint... This should be sufficient, right?

pkuczynski:
> Please also update _vendor_ files, we could rename the folder later if you like to

Done...

ST-DDT (approved):
The performance gain is minimal and it has some implications on code readability.
Note sure, about it.

pkuczynski:
> The performance gain is minimal and it has some implications on code readability.
> Note sure, about it.

The performance gain is minimal, but the readability increases big time. At least to my taste.

src/finance.ts
```diff
@@ -384,20 +384,20 @@ export class Finance {
   bic(): string {
     const vowels = ['A', 'E', 'I', 'O', 'U'];
     const prob = this.faker.datatype.number(100);
-    return (
+    return `${
       this.Helpers.replaceSymbols('???') +
       this.faker.random.arrayElement(vowels) +
```
Shinigami92, line 389: If we already force to use it, then also use it everywhere

src/commerce.ts
```diff
@@ -45,13 +45,7 @@ export class Commerce {
...
    */
   productName(): string {
-    return (
-      this.faker.commerce.productAdjective() +
-      ' ' +
-      this.faker.commerce.productMaterial() +
-      ' ' +
-      this.faker.commerce.product()
-    );
+    return `${this.faker.commerce.productAdjective()} ${this.faker.commerce.productMaterial()} ${this.faker.commerce.product()}`;
```
Shinigami92, line 48: I question myself why prettier allows this `printWidth`? It's far over 80 columns 🤔

src/address.ts
```diff
@@ -255,26 +255,23 @@ export class Address {
     let address = '';
     switch (this.faker.datatype.number(2)) {
       case 0:
-        address =
-          this.Helpers.replaceSymbolWithNumber('#####') +
-          ' ' +
-          this.faker.address.streetName();
+        address = `${this.Helpers.replaceSymbolWithNumber(
+          '#####'
```
Shinigami92, line 259: This is not really more readable than before 😕

scripts/verifyCommit.ts
```diff
@@ -21,13 +21,14 @@ if (!isMergeCommit && !releaseRE.test(msg) && !commitRE.test(msg)) {
...
-      `    ${colors.green(`fix: handle events on blur (close #28)`)}\n\n` +
-      colors.red(`  See .github/commit-convention.md for more details.\n`)
+    )}\n\n${colors.red(
+      `  Proper commit message format is required for automated changelog generation. Examples:\n\n`
+    )}    ${colors.green(`feat: add 'comments' option`)}\n` +
+      `    ${colors.green(
+        `fix: handle events on blur (close #28)`
+      )}\n\n${colors.red(
+        `  See .github/commit-convention.md for more details.\n`
+      )}`
```
Shinigami92, line 31: This is horrible 😱 
Also, not sure if this helps at least a little bit, but template literals allows real newlines. So try to replace the `\n` with just newlines.
To be safe in terms of indentation, this will then stick far to the left.

scripts/generateLocales.ts
```diff
@@ -240,8 +240,9 @@ for (const locale of locales) {
     locale,
     'LocaleDefinition',
     1,
-    `title: '${localeTitle}',` +
-      (localeSeparator ? `\nseparator: '${localeSeparator}',` : '')
+    `title: '${localeTitle}',${
+      localeSeparator ? `\nseparator: '${localeSeparator}',` : ''
+    }`
```
Shinigami92, line 245: Should we introduce a new variable for that?

Shinigami92 (dismissed):

scripts/verifyCommit.ts
pkuczynski: It was quite terrible before too ;) Prettier forcing formatting also does not help :/ @Shinigami92 have a look if it is better now?

src/address.ts
pkuczynski: This is changing in #754 so I will leave it as it is for now, ok?

src/commerce.ts
pkuczynski: 80 columns are anyway quite narrow limit to my taste...

pkuczynski:
@ST-DDT made some improvements. Can you have a look again and comment on places that you don't like?

pkuczynski:
@Shinigami92 @ST-DDT all changes applied, ready for your re-review

scripts/verifyCommit.ts
```diff
@@ -18,16 +18,22 @@ const isMergeCommit = msg.startsWith('Merge remote-tracking-branch');
 
 if (!isMergeCommit && !releaseRE.test(msg) && !commitRE.test(msg)) {
   console.log();
+
   console.error(
-    `  ${colors.bgRed(colors.white(' ERROR '))} ${colors.red(
+    `  ${colors.bgRed(colors.white(' ERROR '))}${colors.red(
```
ST-DDT, line 23: Is this change intentional?

grafik

ST-DDT (changes requested):

scripts/verifyCommit.ts
pkuczynski: Not at all, not sure how did it happend. Thank you for a good catch!

src/internet.ts
```diff
@@ -314,15 +304,9 @@ export class Internet {
...
-      (redStr.length === 1 ? '0' : '') +
-      redStr +
-      (greenStr.length === 1 ? '0' : '') +
-      greenStr +
-      (blueStr.length === 1 ? '0' : '') +
-      blueStr
-    );
+    return `#${redStr.length === 1 ? '0' : ''}${redStr}${
+      greenStr.length === 1 ? '0' : ''
+    }${greenStr}${blueStr.length === 1 ? '0' : ''}${blueStr}`;
```
ST-DDT, line 309: Please use one line per color here to improve readability.

ST-DDT (commented):
One last quibble, otherwise it looks good to me.

src/internet.ts
pkuczynski: I fixed it like this: https://github.com/faker-js/faker/pull/732/commits/b1f44be9e6ffb76acda10978271668d922a0a861

Does this work? I think @Shinigami92 mentioned that we should not use `private` methods, not sure why? But then I needed `this.faker.datatype.number` in this function, so I could not take it out from the class...

And recommendation welcome! All in all I think it looks way cleaner now :)

Shinigami92 (changes requested):

src/internet.ts
```diff
@@ -278,35 +278,44 @@ export class Internet {
...
-   * @param baseRed255 The optional base red. Used for aesthetically pleasing color palettes. Supports values between `0` and `255`. Defaults to `0`.
-   * @param baseGreen255 The optional base green. Used for aesthetically pleasing color palettes. Supports values between `0` and `255`. Defaults to `0`.
-   * @param baseBlue255 The optional base blue. Used for aesthetically pleasing color palettes. Supports values between `0` and `255`. Defaults to `0`.
+   * Based on:
+   * http://stackoverflow.com/questions/43044/algorithm-to-randomly-generate-an-aesthetically-pleasing-color-palette
+   *
+   * @param base Base color value in range form 0 to 255
+   * @private
+   */
+  private colorFromBase(base: number): string {
```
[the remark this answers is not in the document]
Shinigami92: It has something to do with the
https://github.com/faker-js/faker/blob/7017e07ca4215f00f0bc23f8a2c33b08242a4d84/src/internet.ts#L7-L16

Even if the tests are green, these private functions are not real-js private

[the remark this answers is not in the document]
pkuczynski: Yeah, I was wondering why do we do this bind? So we can use `faker.bla.bla` without instantiating the class with `new`?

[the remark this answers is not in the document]
pkuczynski: Then the only option I have is to define this function inside `color` but that sucks a bit, don't you think so?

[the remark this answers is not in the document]
pkuczynski: Than why don't we export faker like:

```
const faker = new Faker();

export { faker }
export { Faker }
```

This would solve those issues and would not require this funky, counterintuitive constructor. Right?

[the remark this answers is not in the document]
Shinigami92: Exact, this was a historical possibility from pre 5.5.3 times. I personally don't like it really much, but folks are doing stuff like:
```ts
const genColor = faker.internet.color; // This would not work without our class-constructor-binding-this-hack
const color = genColor();
```
😣
And we did not want to break their source code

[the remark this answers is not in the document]
pkuczynski: Yeah, but if we would export `const faker = new Faker()` the example above wold work nicely, right?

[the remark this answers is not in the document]
ST-DDT: IMO, the method is quite small, if you use a lambda, so its shouldn't be that much of an issue.

````ts
const colorFromBase = (base: number) => Math.floor((this.faker.datatype.number(256) + base) / 2)
      .toString(16)
      .padStart(2, '0');
````

It's at least more readable, then repeating it 3 times.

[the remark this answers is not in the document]
pkuczynski: Yeah, I could use the lambda and size wise its fine, but I would expect it to be slower? Run the benchmark and to my surprise it was not, not sue if this is correct results: https://jsben.ch/xyjlE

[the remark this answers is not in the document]
ST-DDT: For me lambda is ~1% faster.
But as long as the impact isn't huge I'm not that concerned about performance.
We are a lib for testing, not for production.
Please note that a console.log is 20xtimes slower than our function (not sure whether the benchmark requires a black hole) in order to be accurate.

Lambda/Static only = ~5 000 000
With Console log = ~220 000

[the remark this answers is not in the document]
pkuczynski: I updated the benchmark to remove the console.log and all depends on the env https://jsben.ch/BqH74 Safari seems to be ~5-10% slower. In Chrome ~5%.

We are testing library which means we could be also used for performance testing to generate hundreds of millions of records, where performance would matter...

But I can do lambda if you think its better...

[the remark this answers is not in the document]
pkuczynski: Done

ST-DDT (approved):

pkuczynski:
@ST-DDT @Shinigami92 all changes applied... Can you re-review, please?

ST-DDT (approved):

prisis (approved):

Shinigami92 (approved):

PR #733, opened 2022-03-30
chore: do not eslint src/vendor/*

pkuczynski:
There is no point in eslinting external libraries, right? 

Takes down number of eslint errors from 71 to 60

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = defineConfig({
   ignorePatterns: [
     ...readGitignoreFiles(),
     '.eslintrc.js', // Skip self linting
+    'src/vendor/**/*'
   ],
   root: true,
   env: {


import-brain (dismissed):

ST-DDT:
How about utils? We could also put the `errors` in there.

pkuczynski:
> IMO this is not totally true, we already converted these files to newer syntax and the unique and user-agend were already heavily altered by us. It's just in vendor for historical reasons, we may want to move them out into a different folder 🤷

I think we should annotate them on the top saying that they have been modified. Don't you think?

pkuczynski:
> How about `utils`? We could also put the `errors` in there.

Sounds good, but I think I would vote for `lib`.

I can create a PR with both changes if we settle on what should be done?

ST-DDT:
> > How about `utils`? We could also put the `errors` in there.
> 
> Sounds good, but I think I would vote for `lib`.

Lib is fine as well.

Shinigami92:
> Lib is fine as well.

I would prefer `utils`, `lib` is a somewhat bad word because it's like `dist`
In JS ecosystem dist and lib are mostly the folders that are served in the emitted bundle that is published on npm, so I'm used to gitignore these folder and only generate these on build

Issue #541, opened 2022-02-22
Source code todo list


xDivisionByZerox:
This issue serves as a container for all todos listed in the source code from 03c3d1156312153428f2db5649b24dd5fc046d43.

This covers only `TODO` and `FIXME` and no `it.todo` statements for now.

Column `method` is a static link to the original comment in 03c3d1156312153428f2db5649b24dd5fc046d43. The other columns will be updated with time so it might be that the description you see in the table will not match with the description in the source file, because it was updated.

| description                                                                                                                      | link                                                                                                                                   | issue/pr  | fixed |
| -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- | --------- | ----- |
| Fix types                                                                                                                        | [address.zipCodeByState](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L69)            | #759      | ✔     |
| The string parameter doesn't work as expected.                                                                                   | [address.city](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L94)                      | #682, #948    | ✔    |
| change all these methods that accept a boolean to instead accept an options hash.                                                | [address](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L175)                          | #649      | ✔     |
| this.faker.definitions.address.secondary_address                                                                                 | [address.secondaryAddress](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L248)         | #676      | ✔     |
| Limit the parameter to the two values.                                                                                           | [address.countryCode](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L289)              | #587 #588 | ✔    |
| useAbbr not in use                                                                                                               | [address.state](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L316)                    | #677      | ✔     |
| Allow coordinate parameter to be [string, string].                                                                               | [address.nearbyGPSCoordinate](https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d4/src/address.ts#L458)      | #780      | ❌    |
[...]

PR #753, opened 2022-04-01
chore: fix mustache type warning and add some tests

ST-DDT:
Ref: #541


src/helpers.ts
```diff
@@ -482,20 +482,20 @@ export class Helpers {
...
-        // TODO @Shinigami92 2022-01-14: Try to improve the type or maybe use `if`
-        // @ts-expect-error
-        data[p]
-      );
+      const value = data[p];
+      if (typeof value === 'string') {
+        str = str.replace(re, value);
+      } else {
+        str = str.replace(re, value);
+      }
```
ST-DDT, line 498: I'm not sure why I have to do this, but otherwise I get the ts error.

test/helpers.spec.ts
```diff
@@ -730,6 +730,44 @@ describe('helpers', () => {
             faker.helpers.mustache()
           ).toBe('');
```
ST-DDT, line 731: This test feels wrong to me.

src/helpers.ts
```diff
@@ -482,20 +482,20 @@ export class Helpers {
     str: string | undefined,
     data: Record<
       string,
-      string | ((substring: string, ...args: any[]) => string)
+      string | ((substring: string, ...args: unknown[]) => string)
```
ST-DDT, line 485: Does this change have any negative influence on the users?
As an alternative I could use `string | Parameters<String.prototype.replace>[1]` not sure whether that is easier to read.
TypeScript doesn't seem to detect that both are possible here either.

src/helpers.ts
Shinigami92: Looks weird, but it's a solution for now
Feels somewhat like a TS bug :shrug:

test/helpers.spec.ts
Shinigami92: Maybe it is a leftover from historical times

ST-DDT: Should I remove it?

Shinigami92: yes

Shinigami92 (approved):

PR #757, opened 2022-04-02
ci: auto npm publish script

### LINKED ISSUE #268: Automatic npm publish on commit

As a developer using faker I want to get the latest merged changes on npm so that I can get access to the latest changes without local npm linking.

Right now, it seems that tag creation is done manually (maybe?) and there are some changes that are not available on the published npm version.

### Suggested solution

We could setup some action to execute on master commits to automatically bump versions and publish to npm.

I've used https://www.npmjs.com/package/semantic-release in the past, config is a bit daunting but the result is pretty great.

Additional branches could be created to generate release candidate/alpha/beta releases as well (but personally, I don't really think this is useful).

### Alternative

An alternative would be to keep pushing tags manually but maybe do it at a higher frequency ?

### Additional context

_No response_

ejcheng:
fixes #268


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
import-brain, line 18: This needs to be configured for this workflow to work.

Shinigami92:
This looks ridiculously easy :thinking:

ST-DDT (dismissed):
I cannot test this, but I think it looks good.
We can merge and check it Monday with our v6.1.2 release.

Shinigami92 (dismissed):
But we need @prisis or @damienwebdev to add the token

import-brain:
> This looks ridiculously easy :thinking: _sus_?

May or may not have been copy-pasted from GitHub documentation😂

xDivisionByZerox:
Isn't the point of this pipeline to ensure that things like v6.1.0 don't happen again? Well, the pipeline doesn't build the newest state of the library or did I miss something?

Shinigami92:
> Does the `npm publish` automatically include a build?

Yes it does, since 6.1.1

https://github.com/faker-js/faker/commit/62dcfc9658e6d3ee23b9e674d10b9a229b026d0e#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R76

xDivisionByZerox:
Ok, I didn't notice that. But tbh I'm not a fan of relying on pre/post scripts for pipelines since:
1. I want to see what a pipeline does by looking at its config file
2. A change in this script can slip in too easy  without noticing it directly

Just some points you might want to consider.

ST-DDT:
@xDivisionByZerox So you would like us to explicitly call `build` or what?

xDivisionByZerox:
> @xDivisionByZerox So you would like us to explicitly call build or what?
I would do so for my pipelines, yes. But I don't think I'm in the position to demand anything. 

I guess it would be helpful enough, for anyone that might have a look at the config in the future, to add a comment over the `npm publish`, that there is a pre-publish script in place.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,18 @@
...
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
```
pkuczynski, line 14: Looks like it's default so does not need to be specified?

.github/workflows/npm-publish.yml
pkuczynski: Why NODE_TOKEN_AUTH and not NPM_TOKEN?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,18 @@
+name: Publish to NPM
+on:
+  # When a release is created on GitHub, publish to NPM
+  release:
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
pkuczynski, line 8: Does it have to be Ubuntu? Alpine are usually smaller = faster execution

pkuczynski (changes requested):

pkuczynski:
I am with @xDivisionByZerox on this...

ST-DDT:
> I am with @xDivisionByZerox on this...

Would something like this suffice for you?

````yaml
# Relies on the prepublishOnly hook to build the project
- run: npm publish
````

Unfortunately it isn't possible to add comments to the package.json to add a hint there too.
IMO the info in the pipeline is not really helpful, because it only state what it depends on (and that it's nothing extraordinary for an Node dev), but the other side side contains no reference to dependent steps (and AFAICT these vary greatly between projects).

**Comment by xDivisionByZerox** [comment]
> Would something like this suffice for you?
> 
> ```yaml
> # Relies on the prepublishOnly hook to build the project
> - run: npm publish
> ```
> 
> Unfortunately it isn't possible to add comments to the package.json to add a hint there too. IMO the info in the pipeline is not really helpful, because it only state what it depends on (and that it's nothing extraordinary for an Node dev), but the other side contains no reference to dependent steps (and AFAICT these vary greatly between projects).

I mean if you state it that way, why not explicitly call the necessary build steps? Yes, it's a bit redundant and takes some more time to complete, on the other side this is run like once a week, so it shouldn't matter.
Even going a step further: Why not remove the `prepublish` script from the package? When a release pipeline is in place NOONE should make a manual release EVER.

**Comment by prisis** [comment]
I would go more in this direction [https://www.npmjs.com/package/@changesets/cli,](https://www.npmjs.com/package/@changesets/cli,) check [https://github.com/Thinkmill/manypkg](https://github.com/Thinkmill/manypkg) as example how it works, it creates a .changeset folder with all the changes than can be than released

**Comment by ST-DDT** [comment]
Looks like we won't have a pipeline by Monday.

**Comment by prisis** [comment]
Nope, sorry... , we can talk about it tomorrow in the call :)

File: .github/workflows/npm-publish.yml [hunk 4]
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
```
Shinigami92, line 13: We might want to use node 18

```suggestion
          node-version: 18
```

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,18 @@
...
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish
```
Shinigami92, line 16: We might want to use pnpm instead of npm for the automated publish process
That's because pnpm can later be used to release mono-repo structures
and so this is a preparation for when we extract locales out to own packages

.github/workflows/npm-publish.yml
Shinigami92: Yeah that looks wrong :thinking: 

https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#set-the-token-as-an-environment-variable-on-the-cicd-server

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,17 @@
...
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: pnpm ci
```
ST-DDT, line 14: Does this command exist only in ci? I cannot run it locally.

ST-DDT (changes requested):

.github/workflows/npm-publish.yml
Shinigami92: I think this is the native **npm** not pnpm command
The theoretically equivalent would be something like using `--frozen-lockfile`
But this might be auto-enables in `env CI=true`
https://pnpm.io/cli/install#--frozen-lockfile

ejcheng:
Superseded by #1096

ST-DDT:
@import-brain Do you still need the branch? If not, please delete it.

ejcheng:
> @import-brain Do you still need the branch? If not, please delete it.

Done

PR #764, opened 2022-04-03
refactor: rename hexaDecimal to hexadecimal

### LINKED ISSUE #336: datatype.hexaDecimal should be called datatype.hexadecimal

pkuczynski:
Fixes #354
Fixes #336


test/datatype.spec.ts
```diff
@@ -629,14 +629,38 @@ describe('datatype', () => {
       });
 
       describe('hexaDecimal', () => {
+        it('should display deprecated message', () => {
+          const spy = vi.spyOn(console, 'warn');
```
xDivisionByZerox, line 633: Should this not spy on the `deprecate` function?

ST-DDT: I think so too, that would be more refactor safe.

pkuczynski: We could, but we dont in the other tests, so I followed what we do already. I suggest we merge it as it is and I will create separate PR which fix all the tests to spy on `deprecated` instead of `console.warn`. Make sense?

ST-DDT: Sounds good to me.

ST-DDT (approved):

test/datatype.spec.ts
Shinigami92: I think we should test the assumption, not the implementation.
We assume that a warning is logged to the console. How the code does it, is implementation detail.
We also did not export by design the deprecation function and marked it explicitly as internal.
So IMO spying on the `console.warn` here is indeed correct

src/datatype.ts
```diff
@@ -185,11 +186,35 @@ export class Datatype {
...
    * faker.datatype.hexaDecimal() // '0xb'
    * faker.datatype.hexaDecimal(10) // '0xaE13F044fb'
+   *
+   * @deprecated
    */
   hexaDecimal(length = 1): string {
+    deprecated({
+      deprecated: 'faker.datatype.hexaDecimal()',
+      proposed: 'faker.datatype.hexadecimal()',
+      since: 'v6.1.0',
```
Shinigami92, line 201: Actually this will be 6.1.2

```suggestion
      since: 'v6.1.2',
```

Shinigami92 (approved):

Shinigami92 (approved):

Shinigami92:
Oh crap ^^ I forgot to also update the test, sry

test/datatype.spec.ts
```diff
@@ -629,14 +629,38 @@ describe('datatype', () => {
...
+
+          faker.datatype.hexaDecimal();
+
+          expect(spy).toHaveBeenCalledWith(
+            '[@faker-js/faker]: faker.datatype.hexaDecimal() is deprecated since v6.1.0 and will be removed in v7.0.0. Please use faker.datatype.hexadecimal() instead.'
```
Shinigami92, line 638:
```suggestion
            '[@faker-js/faker]: faker.datatype.hexaDecimal() is deprecated since v6.1.2 and will be removed in v7.0.0. Please use faker.datatype.hexadecimal() instead.'
```

Shinigami92 (approved):

ST-DDT (approved):

PR #765, opened 2022-04-03
test: check that system.filePath() generates file extensions

ST-DDT:
Ref: #541


test/system.spec.ts
```diff
@@ -2,7 +2,6 @@ import validator from 'validator';
 import { afterEach, describe, expect, it } from 'vitest';
 import { faker } from '../src';
 
-// TODO @prisis 2022-01-31: Add multiple seed based expectations.
```
ST-DDT, line 5: We already have 3, so this should suffice for now.

test/system.spec.ts
```diff
@@ -235,12 +234,16 @@ describe('system', () => {
...
           const filePath = faker.system.filePath();
+          const parts = filePath.split('/');
 
           expect(
             filePath.startsWith('/'),
             'generated filePath should start with /'
           ).toBeTruthy();
-          // TODO @prisis 2022-01-26: Add test to validate if the path has ext on the end.
+          expect(
+            parts[parts.length - 1],
```
Shinigami92, line 244: Would it be better to match against a regex?
Like `/\w+\.\w+/`, so we test that the period is somewhere in the middle

Shinigami92 (approved):

import-brain (approved):

PR #767, opened 2022-04-03
refactor: remove inconsistent defaults from internet.password()

ST-DDT:
The defaults in the password method are inconsistent, I consolidated them and removed useless lines.

### DIFF
--- a/src/internet.ts
+++ b/src/internet.ts
@@ -358,7 +358,8 @@ export class Internet {
    *
    * @param len The length of the password to generate. Defaults to `15`.
    * @param memorable Whether the generated password should be memorable. Defaults to `false`.
-   * @param pattern The pattern that all chars should match should match. Defaults to `/\w/`.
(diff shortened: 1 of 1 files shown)


src/internet.ts
```diff
@@ -369,15 +370,11 @@ export class Internet {
...
-    len?: number,
-    memorable?: boolean,
-    pattern?: RegExp,
-    prefix?: string
+    len: number = 15,
+    memorable: boolean = false,
+    pattern: RegExp = /\w/,
+    prefix: string = ''
   ): string {
-    len = len || 15;
```
Shinigami92, line 377: Just to notice, previously an input of `0` was fallbacked to `15`
Now `0` will be tacken as `0`
So at least we should switch from chore to refactor

import-brain (approved):

Shinigami92 (approved):

prisis (approved):

PR #771, opened 2022-04-04
fix: random word fails on undefined

ST-DDT:


src/random.ts
```diff
@@ -326,7 +326,7 @@ export class Random {
       const randomWordMethod = this.faker.random.arrayElement(wordMethods);
 
       result = randomWordMethod();
-    } while (bannedChars.some((char) => result.includes(char)));
+    } while (!result || bannedChars.some((char) => result.includes(char)));
```
Shinigami92, line 329: Can this result in an infinit loop?

Shinigami92 (approved):

src/random.ts
ST-DDT: Theoretically yes. But since we are deprecating this method in #714 anyway I don't want to put too much effort in it.
This is only a problem, if your locale is mostly empty and without fallback.

HonzaMac: Shouldn't then `fallbackLocale` take over and fill this entropy gap?

ST-DDT: > Shouldn't then `fallbackLocale` take over and fill this entropy gap?

Yes, but only have one. Theoretically, you could use an empty fallback.
If you make your life hard, then you shouldn't wonder why things aren't as easy.
This issue is originally caused by this exact problem. State isn't provided by both 'sk' and 'en', thus returns undefined.
Unfortunately, we cannot do much about it without breaking the API.

There is a loop safe loophole we could go for:

````diff
- result = randomWordMethod();
+ result = randomWordMethod() || randomWordMethod.name;
````

HonzaMac: Ok. 
`en` seems to [have state filled](https://github.com/faker-js/faker/blob/main/src/locales/en/address/state.ts). It's probably from other data set.

ra100: I guess the issue is, that `sk`, `cz` and others do not have states, so adding some states to those languages would break behaviour, if suddenly state would appear, that would be unexpected :)

ra100: Would it be fine, to modify `wordMethods` array and remove those methods, which can return empty/undefined in any language?

ST-DDT: > Ok. `en` seems to [have state filled](https://github.com/faker-js/faker/blob/main/src/locales/en/address/state.ts?rgh-link-date=2022-04-04T09%3A35%3A10Z). Is it from other dataset?

I have checked, it is the state method, the fallback code returns an empty array which is really weird.
@Shinigami92 and I will look into it some more later today.

ST-DDT: Found it the `sk` locale defines states as empty array.

ST-DDT: > Would it be fine, to modify `wordMethods` array and remove those methods, which can return empty/undefined in any language?

Most method can return empty/undefined results at this time. Especially with user provided locales.
In v7 we plan to create separate accessors for definitions we require for a method or just want to check for existence.
That way the methods could throw `FakerMissingDefinitionsError` (or similar).

PR #781, opened 2022-04-05
chore: fix finance.iban warnings

ST-DDT:
Ref: #541

### DIFF
--- a/src/iban.ts
+++ b/src/iban.ts
@@ -30,21 +30,13 @@ export = {
   pattern10: ['01', '02', '03', '04', '05', '06', '07', '08', '09'],
   pattern100: ['001', '002', '003', '004', '005', '006', '007', '008', '009'],
   toDigitString: (str: string): string =>
-    str.replace(
-      /[A-Z]/gi,
-      (match) =>
-        // TODO @Shinigami92 2022-01-13: This needs to be converted to string
-        // @ts-expect-error
-        match.toUpperCase().charCodeAt(0) - 55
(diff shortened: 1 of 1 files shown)


Shinigami92 (approved):

src/iban.ts
```diff
@@ -30,21 +30,13 @@ export = {
...
     ),
   mod97: (digitStr: string): number => {
     let m = 0;
     for (let i = 0; i < digitStr.length; i++) {
-      m =
-        (m * 10 +
-          // @ts-expect-error: We need to convert this properly
-          (digitStr[i] | 0)) %
-        97;
+      m = (m * 10 + +digitStr[i]) % 97;
```
pkuczynski, line 39: I am not sure if we can be 100% sure what's coming in `digitStr`? If its numbers, than it works:

```js
2 * 10 + +'923'[1] // 22
```

But if that's not numbers:

```js
2 * 10 + ('abc'[1] | 0) // 20 << before
2 * 10 + +'abc'[1] // NaN << now
```

pkuczynski (changes requested):
Change in `mod97` needs to be clarified...

src/iban.ts
ST-DDT: Yes, it's only numbers. See the above method that converts the alpha chars to digits.

https://en.wikipedia.org/wiki/International_Bank_Account_Number#Validating_the_IBAN

pkuczynski (approved):

PR #791, opened 2022-04-06
feat: reimplement datatype.bigInt

Shinigami92:
Depends on #797


src/datatype.ts
```diff
@@ -286,21 +286,30 @@ export class Datatype {
...
+   * faker.datatype.bigInt({ min: 10n, max: 100n }) // 36n
    */
-  bigInt(value?: string | number | bigint | boolean): bigint {
-    if (value === undefined) {
-      value =
-        Math.floor(this.faker.datatype.number() * 99999999999) + 10000000000;
+  bigInt(options?: bigint | { min?: bigint; max?: bigint }): bigint {
+    const opts = typeof options === 'bigint' ? { max: options } : options ?? {};
+
+    const min = typeof opts.min === 'bigint' ? opts.min : BigInt(0);
```
pkuczynski, line 305: So if I pass `min: 100` faker would ignore this and use 0? I would be really surprised when that happens...

src/datatype.ts
```diff
@@ -286,21 +286,30 @@ export class Datatype {
   /**
    * Returns a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#bigint_type) number.
    *
-   * @param value When provided, this method simply converts it to `BigInt` type.
+   * @param options Maximum value or options object.
```
pkuczynski, line 289: This is a breaking change. How do we warn users about it?

src/datatype.ts
```diff
@@ -286,21 +286,30 @@ export class Datatype {
...
+   * faker.datatype.bigInt(100n) // 52n
+   * faker.datatype.bigInt({ min: 1000000n }) // 431433n
+   * faker.datatype.bigInt({ max: 100n }) // 42n
+   * faker.datatype.bigInt({ min: 10n, max: 100n }) // 36n
    */
-  bigInt(value?: string | number | bigint | boolean): bigint {
-    if (value === undefined) {
-      value =
-        Math.floor(this.faker.datatype.number() * 99999999999) + 10000000000;
+  bigInt(options?: bigint | { min?: bigint; max?: bigint }): bigint {
```
pkuczynski, line 302: I think we should accept numbers here too, not only bigint...

pkuczynski (changes requested):

src/datatype.ts
Shinigami92: Currently it is not (TS-)possible to pass a `number`, so what you are asking for is to make it possible to pass `number`?

src/datatype.ts
Shinigami92: We write in changelog that we fully reimplemented this function
I do not expect that anyone currently use the function, as it is currently totally useless :shrug:

src/datatype.ts
Shinigami92: So yeah, that is what I was asking with the BigIntConstructor params
Also it's the answer to https://github.com/faker-js/faker/pull/791#discussion_r844266361 :slightly_smiling_face: 

BitIntConstructor params are `number | bigint | string | boolean`

test/datatype.spec.ts
```diff
@@ -692,6 +703,65 @@ describe('datatype', () => {
           const generateBigInt = faker.datatype.bigInt();
           expect(generateBigInt).toBeTypeOf('bigint');
         });
+
+        it('should generate a big bigInt value with low delta', () => {
+          const generateBigInt = faker.datatype.bigInt({
+            min: 999999999n,
+            max: 1000000000n,
+          });
+          expect(generateBigInt).toBeTypeOf('bigint');
```
ST-DDT, line 712: This should check the bounds.

test/datatype.spec.ts
```diff
@@ -692,6 +703,65 @@ describe('datatype', () => {
...
+          const initialMin = 1n;
+          const initialOtherProperty = 'hello darkness my old friend';
+          const input: {
+            min?: bigint;
+            max?: bigint;
+            otherProperty: string;
+          } = {
+            min: initialMin,
+            otherProperty: initialOtherProperty,
+          };
```
ST-DDT, line 757: Wouldn't it be easier to freeze the input and check for exceptions due to attempted mutations?

src/datatype.ts
```diff
@@ -286,21 +286,61 @@ export class Datatype {
...
+      BigInt(
+        Array.from(
+          { length },
+          () => '0123456789'.split('')[this.number({ min: 0, max: 9 })]
+        ).join('')
+      );
+
+    const delta = max - min;
+
+    const offset = generateRandomBigInt(delta.toString(10).length) % delta;
```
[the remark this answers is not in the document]
ST-DDT: Loops have a worst case success rate of 1/9. You might run a few loops before you get a valid value.
Not sure what is more important "no bias" or "better performance".

ST-DDT (changes requested):

ST-DDT:
Please also add a test, that uses min:0, max: 1000000000000 (it is important that these are 1 and then only zeros, as these are harder to "find" for some implementations (e.g. loop based ones).

00-99 => 100% (100/100) success rate
000-100 => 10% (100/999) success rate

src/datatype.ts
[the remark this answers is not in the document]
ST-DDT: We could do 10 tries via loop before we fall back to the modulo operation.

src/datatype.ts
```diff
@@ -286,21 +286,61 @@ export class Datatype {
...
     }
 
-    return BigInt(value);
+    const generateRandomBigInt = (length: number) =>
+      BigInt(
+        Array.from(
+          { length },
+          () => '0123456789'.split('')[this.number({ min: 0, max: 9 })]
+        ).join('')
+      );
```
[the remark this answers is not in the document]
Shinigami92: I will implement this in another PR to track a specific commit message in git history

src/datatype.ts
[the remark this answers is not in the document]
Shinigami92: I would like not to call `faker.datatype.number` over and over again, because that way we waste seed sequence 🤔

src/datatype.ts
```diff
@@ -286,21 +286,62 @@ export class Datatype {
...
+      min = BigInt(0);
+      max = BigInt(options ?? 999999999999999);
+    }
+
+    if (max === min) {
+      return min;
+    }
+
+    if (max < min) {
+      throw new Error(`Max ${max} should be larger then min ${min}`);
```
Shinigami92, line 329: Needs to be FakerError

src/datatype.ts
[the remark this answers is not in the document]
ST-DDT: Are they limited?

Shinigami92:
Awaiting other PRs like #797 to be done

src/datatype.ts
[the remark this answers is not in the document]
Shinigami92: Let's start with the current implementation and if users raising issue for that, we can improve it later

ST-DDT (approved):

pkuczynski (dismissed):

ST-DDT (dismissed):

ST-DDT (approved):

prisis (approved):

src/modules/datatype/index.ts
```diff
@@ -261,20 +261,59 @@ export class Datatype {
...
+  ): bigint {
+    let min: bigint;
+    let max: bigint;
+
+    if (typeof options === 'object') {
+      min = BigInt(options.min ?? 0);
+      max = BigInt(options.max ?? min + BigInt(999999999999999));
+    } else {
+      min = BigInt(0);
+      max = BigInt(options ?? 999999999999999);
```
xDivisionByZerox, line 296: Why can you pass `options` directly here? [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt#syntax) documentation states that the value can "May be a string or an integer". So why the `boolean` (in general) or the object?

I've never worked with BigInt so sorry if this makes no sense.

src/modules/datatype/index.ts
```diff
@@ -261,20 +261,59 @@ export class Datatype {
...
+   * faker.datatype.bigInt({ max: 100n }) // 42n
+   * faker.datatype.bigInt({ min: 10n, max: 100n }) // 36n
    */
-  bigInt(value?: string | number | bigint | boolean): bigint {
-    if (value === undefined) {
-      value = Math.floor(this.number() * 99999999999) + 10000000000;
+  bigInt(
+    options?:
+      | bigint
+      | boolean
```
xDivisionByZerox, line 280: What would `faker.datatype.bigInt(false)` return? I don't understand what advantage we have from accepting a boolean parameter.

Shinigami92: image

It's just the official constructor argument

PR #797, opened 2022-04-07
feat: random numeric

Shinigami92:
This introduce a new function beside `alpha` and `alphaNumeric` in the random module.
This function will later be used as a base for other faker functions like `finance.pin` and `datatype.bigInt` (#791).
It can also be used for potential new faker functions like `internet.authCode`.


ST-DDT (changes requested):

src/random.ts
```diff
@@ -540,6 +540,41 @@ export class Random {
...
   }
 
+  /**
+   * Generating a string consisting of digits based on length.
+   *
+   * The minimum value will be `'1'`.
+   *
+   * @param length The number of digits to generate. Defaults to `1`.
+   *
+   * @throws When `length` is less than or equals to `0`.
```
import-brain, line 550:
```suggestion
   * @throws When `length` is less than or equal to `0`.
```

import-brain:
Doesn't this do the same thing as ```finance.pin()```?

src/random.ts
```diff
@@ -540,6 +540,41 @@ export class Random {
     return wholeString;
   }
 
+  /**
+   * Generating a string consisting of digits based on length.
```
pkuczynski, line 544: Numeric in many languages and SQL databases is a number with certain precision: `NUMERIC(precision, scale)`. So having a method with this name would be misleading. If user would like to have a number with certain number of digits he could just call `faker.datatype.number({ min: 100, max: 999 })`, correct? And if he would like to make a string out of it, he could typecast it `String(faker.datatype.number(...))`.

So in principle I am a bit against this feature as it is unclear and duplicates what's already achievable by other method.

pkuczynski: As a side node, I think we could consider splitting `faker.datatype` into `faker.number`, `faker.string` etc...

Shinigami92:
> Doesn't this do the same thing as `finance.pin()`?

@import-brain: a little bit yes, but it is more like `finance.pin` can use this function. For the UX I would like to have both functions, but with a limited option set that is specialised for generating pin numbers. Maybe we also could add something like `internet.authCode`?
`datatype.bigInt` will also use this function.

@pkuczynski: @ST-DDT and I have short communication paths via personal friendship and discord, so we sometimes implement stuff together in our private space. I understand your suggestions, but sometimes we are a little bit ahead in thinking processes (e.g. in terms of how we want to re-structure the modules in "vFuture"). Not everything of that is somewhere to read because we are just thinking by communicating. So it might be that we want to go into another direction as you suggested.
But re-structuring modules will be anyways something for that far future.

src/random.ts
```diff
@@ -540,6 +540,41 @@ export class Random {
...
+    if (length < 1) {
+      throw new FakerError(
+        'Minimum length for numeric string should be greater than or equals to 1.'
+      );
+    }
+
+    let result = Array.from(
+      { length },
+      () =>
+        '0123456789'.split('')[this.faker.datatype.number({ min: 0, max: 9 })]
```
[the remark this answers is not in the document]
Shinigami92: I think I will make a screenshot of that, post it on twitter and say that sometimes a break could be healthy 🤣

pkuczynski:
I noticed that you have short paths of communication, which is fine of course, but keeping the community away from it, is kind of demotivating... :/

ST-DDT:
> I noticed that you have short paths of communication, which is fine of course, but keeping the community away from it, is kind of demotivating... :/

Thanks for letting us know. We try to improve on this and establish regular team-meetings.
You can always ping us, if you have a specific topic you would like to discuss.
Things like this tend to come up during our reviews and it's hard to always communicate them.
(I usually tend to create issues for those things, but since this is only at the stage of we are thinking about it, I haven't done so yet)
I created a discussion for this idea of ours now: https://github.com/faker-js/faker/discussions/805

If you notice again, that there are some things that haven't been communicated (well), then let us know.

src/random.ts
ST-DDT: IMO this is a very basic feature, that is useful in many places. I mostly consider this to be the base for the other methods.

You have a point with this method's name being a bit ambigous for now.
Should we name it `numericString` until we restructure the modules in v7?

src/random.ts
```diff
@@ -540,6 +540,73 @@ export class Random {
     return wholeString;
   }
 
+  /**
+   * Generating a string consisting of digits based on length.
```
ST-DDT, line 544:
```suggestion
   * Generates a string consisting of digits based on length.
```

test/random.spec.ts
```diff
@@ -278,6 +280,86 @@ describe('random', () => {
     });
   });
 
+  describe('numeric', () => {
+    it('should return single digit when no length provided', () => {
+      const actual = faker.random.numeric();
+
+      expect(actual).toHaveLength(1);
+    });
```
ST-DDT, line 288: This should check that a digit has been returned.

test/random.spec.ts
```diff
@@ -278,6 +280,86 @@ describe('random', () => {
...
+
+    it.each(times(100))(
+      'should generate random value with a length of %s',
+      (length) => {
+        const actual = faker.random.numeric(length);
+
+        expect(actual).toHaveLength(length);
+      }
+    );
```
ST-DDT, line 297: Same here

test/random.spec.ts
```diff
@@ -278,6 +280,86 @@ describe('random', () => {
...
+
+      expect(actual).toHaveLength(0);
+    });
+
+    it('should return a valid numeric string', () => {
+      const actual = faker.random.numeric();
+
+      expect(actual).toBeTypeOf('string');
+      expect(actual).toMatch(/^[1-9]$/);
+    });
```
ST-DDT, line 316: This is kind of redundant. Please merge it with the above tests.

test/random.spec.ts
```diff
@@ -278,6 +280,86 @@ describe('random', () => {
...
+    it('should allow leading zeros via option and all other digits banned', () => {
+      const actual = faker.random.numeric(4, {
+        allowLeadingZeros: true,
+        bannedDigits: '123456789'.split(''),
+      });
+
+      expect(actual).toBe('0000');
+    });
+
+    it('should allow leading zeros via option and all other digits banned', () => {
```
ST-DDT, line 340:
```suggestion
    it('should fail on leading zeros via option and all other digits banned', () => {
```

test/random.spec.ts
```diff
@@ -278,6 +280,86 @@ describe('random', () => {
...
+          bannedDigits: '123456789'.split(''),
+        })
+      ).toThrowError(
+        new FakerError(
+          'Unable to generate numeric string, because all possible digits are banned.'
+        )
+      );
+    });
+
+    it('should ban all digits pass via bannedDigits', () => {
```
ST-DDT, line 353:
```suggestion
    it('should ban all digits passed via bannedDigits', () => {
```

ST-DDT (changes requested):

test/random.spec.ts
```diff
@@ -1,7 +1,9 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { faker } from '../src';
+import { faker, FakerError } from '../src';
 import { times } from './support/times';
 
+// TODO @Shinigami92 2022-04-08: Add seeded runs
```
Shinigami92, line 5: I extracted an issue #809

ST-DDT (approved):

src/random.ts
pkuczynski: I think we should go straight to `faker.string.numbers`. Please note that I used plural form here, as what we generate is not a number but string consisting of numbers...

src/random.ts
```diff
@@ -540,6 +540,73 @@ export class Random {
     return wholeString;
   }
 
+  /**
+   * Generates a string consisting of digits based on length.
```
pkuczynski, line 544:
```suggestion
   * Generates a given length string of digits.
```

src/random.ts
```diff
@@ -540,6 +540,73 @@ export class Random {
...
+   * faker.random.numeric(42, { allowLeadingZeros: true }) // '00564846278453876543517840713421451546115'
+   * faker.random.numeric(6, { bannedDigits: ['0'] }) // '943228'
+   */
+  numeric(
+    length: number = 1,
+    options: {
+      allowLeadingZeros?: boolean;
+      // TODO @Shinigami92 2022-04-08: Make `bannedDigits` LiteralUnion<Digit>[] in a separate PR
+      // Make something like this also for the other functions (alpha, alphaNumeric)
+      bannedDigits?: ReadonlyArray<string>;
```
pkuczynski, line 564: We could make the type more strict here:

```
      bannedDigits?: ReadonlyArray<'0' | '1' | '2'...>;
```

src/random.ts
```diff
@@ -540,6 +540,73 @@ export class Random {
...
+    if (length <= 0) {
+      return '';
+    }
+
+    const { allowLeadingZeros = false, bannedDigits = [] } = options;
+
+    const joinedBannedDigits =
+      bannedDigits.length > 0
+        ? Array.from(new Set([...bannedDigits]))
+            .filter((char) => /^[0-9]$/.test(char))
```
pkuczynski, line 576: I believe in the other (or maybe the same?) PR you said that we should not validate user input, as we do have contract defined and users are responsible to follow it. What changed your mind?

src/random.ts
pkuczynski: We could also let users provide a string instead of array, so the can say `bannedDigits: '236'`

src/random.ts
```diff
@@ -540,6 +540,73 @@ export class Random {
...
+  ): string {
+    if (length <= 0) {
+      return '';
+    }
+
+    const { allowLeadingZeros = false, bannedDigits = [] } = options;
+
+    const joinedBannedDigits =
+      bannedDigits.length > 0
+        ? Array.from(new Set([...bannedDigits]))
```
pkuczynski, line 575: Isn't this premature optimization? It does not make any functionally any difference if provided array is uniq or not, right?

pkuczynski (changes requested):

src/random.ts
Shinigami92: > We could make the type more strict here:
> 
> ```
>       bannedDigits?: ReadonlyArray<'0' | '1' | '2'...>;
> ```

That is literally the TODO above that line 😉 

> We could also let users provide a string instead of array, so the can say `bannedDigits: '236'`

I currently mimic the behavior of `bannedChars` from `alpha` and `alphaNumeric`. So if we allow a string, let's change that for them consistently in a separate PR.

src/random.ts
Shinigami92: This is not for validating user input, but just filtering out only the necessary chars and the user can pass just any chars they want (for better UX, not the need to especially cast a provided value forcefully with TypeScript types).

ST-DDT: IIRC there was some code that relied on `0` being the first value in there, but I think its gone now.
IMO we can remove the `joinedBannedDigits` entirely now, as we could just use `bannedDigits.includes` in line 582.

src/random.ts
Shinigami92: Yeah, I assume it is just pre-optimization 🤔

src/random.ts
Shinigami92: Done via https://github.com/faker-js/faker/pull/797/commits/9f6cedab0aae7d6bb5f80c1028d955fbfa0492c2

src/random.ts
Shinigami92: I started to draft #819

ST-DDT (approved):
The implementation looks good to me. But we still have to decide where to put it.

src/random.ts
pkuczynski: Why not doing the proper type straight away? :)

pkuczynski (dismissed):
As suggested before I would put it in faker.string.numbers

src/random.ts
ST-DDT: - Consistency of the code (optimally across the entire code base :flushed:), we still have to work on this one
- Traceability in the git history
- But most importantly because it is easier to spot mistakes, if you change these things in multiple places at once.
  (Both implementation and review wise)

import-brain (approved):
I think this is fine where it is, don't think there's really a need to move it

Shinigami92:
> I think this is fine where it is, don't think there's really a need to move it

Thx, I also would like to move the module discussion into a later major and not bind it to these feature implementations

pkuczynski:
> Thx, I also would like to move the module discussion into a later major and not bind it to these feature implementations

Its a new function so it's not about moving it but rather putting in a future destination place... otherwise we will have to deprecate it and people will need to make an update on the next major release. Double the work.

Shinigami92:
Deprecations should not be removed directly, but marked as for removal or moved to new location

Shinigami92:
This PR adds a function in v6.2 to a module that already exists in v6
Adding a whole new module is a separate PR and targeted for v7 or later

test/random.spec.ts
```diff
@@ -304,6 +304,82 @@ describe('random', () => {
...
+    });
+
+    it.each(times(100))(
+      'should generate random value with a length of %s',
+      (length) => {
+        const actual = faker.random.numeric(length);
+
+        expect(actual).toHaveLength(length);
+        expect(actual).toMatch(/^[1-9]/);
+        expect(actual).toMatch(/[0-9]+$/);
```
ST-DDT, line 322: You should merge these checks:

```suggestion
        expect(actual).toMatch(/^[1-9][0-9]+$/);
```

test/random.spec.ts
```diff
@@ -304,6 +304,82 @@ describe('random', () => {
...
+
+      expect(actual).toHaveLength(0);
+    });
+
+    it('should return a valid numeric string with provided length', () => {
+      const actual = faker.random.numeric(1000);
+
+      expect(actual).toBeTypeOf('string');
+      expect(actual).toMatch(/^[1-9][0-9]+$/);
+    });
```
ST-DDT, line 343: Please add a length check.

ST-DDT (changes requested):
Just some minor things left.

Also what about:
https://github.com/faker-js/faker/pull/797#discussion_r845942681

test/random.spec.ts
Shinigami92: How does this work with a given `length = 1`?

ST-DDT: Use * instead of +

ST-DDT (approved):

import-brain (approved):

PR #798, opened 2022-04-07
refactor: datatype number

Shinigami92:


ST-DDT (approved):

src/datatype.ts
```diff
@@ -37,14 +38,26 @@ export class Datatype {
...
-    const opts = typeof options === 'number' ? { max: options } : options ?? {};
+    let min: number;
+    let max: number;
+    let precision: number;
 
-    const min = typeof opts.min === 'number' ? opts.min : 0;
-    let max = typeof opts.max === 'number' ? opts.max : min + 99999;
-    const precision = typeof opts.precision === 'number' ? opts.precision : 1;
+    if (typeof options === 'object') {
+      min = options.min ?? 0;
```
pkuczynski, line 46: The issue with this approach is that it does not check proper type of this property. I know we are using types, but they would only be checked if someone is using TS. When he is not, he could easily call this function as `faker.datatype.number({min: 'a'}). How should it work in such case? Throw an error? Assume option is invalid and take default? Right now it will just accept it and (I think) return `NaN`... Maybe that's fine, don't know. Before it checking for the type and ignoring invalid options, so in a way its a breaking change...

Shinigami92: > what's the value of this refactor?

To align it to the #791 impl

> easily call this function as `faker.datatype.number({min: 'a'})`

If someone would do this, they just disrespect the API contract and we would need to check every single input parameter in every single function.
We don't do that. We do not want to do that.
If someone disrespect the API contract, they need to life with unsupported behavior.
This PR does not change the API contact / function signature.

pkuczynski: Just my $.2 I preferred the old implementation. Felt simpler and shorter. #791 is new so it should rather align to what we had then vice versa :)

ST-DDT: After having another look at this. I think I like the options remapping L40 (but maybe directly overwrite the options).

````ts
if (typeof options === number) {
 options = { ... }
}
````

The default should be set in the signature as well. That would remove the need for the fallback from that check too.
It would also explicitly document the default in our api docs.

After that we don't need the big if anymore either as well as allowing us to use const for the values again.
So that should hopefully solve most of @pkuczynski's issues with these changes.

However, I agree with @Shinigami92 that we don't have to check for most of out API contract.
If we do check, then we should explicitly throw an error instead of silently ignoring the wrong value.

So please revert this partially/change this to be closer to pkuczynski's implementation.

**Review by ST-DDT — CHANGES_REQUESTED** [review]
See my comment above.

**Review by ST-DDT — COMMENTED** [review]
Not sure if you want to fix this here, but there is a bug in this method for negative numbers.

````ts
faker.datatype.number({ min: -5, max: -4 })
````

never generates `-4` due to the if being there (should always apply)

https://github.com/faker-js/faker/blob/0b8cac057a471d1b392a44f1327df05b35c9f9b5/src/datatype.ts#L57

As an alternative you could just remove the if and its body and add +1 to the max  side of the `mersenne.rand()` call. That way `max` could be a const as well.

File: src/datatype.ts [hunk 2]
```diff
@@ -36,22 +36,15 @@ export class Datatype {
...
-    } else {
-      min = 0;
-      max = options ?? 99999;
-      precision = 1;
+    if (typeof options === 'number') {
+      options = { max: options };
     }
 
+    const { min = 0, precision = 1 } = options;
+    let max = options.max ?? min + 99999;
```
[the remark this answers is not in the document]
pkuczynski: I like the destructuring with defaults

ST-DDT (approved):

pkuczynski (approved):

PR #799, opened 2022-04-07
refactor!: remove faker default export

### LINKED ISSUE #777: Remove faker default export from src/index
Remove this line https://github.com/faker-js/faker/blob/b34ef465fb23829e4f07f3ffbba4203737f74b08/src/index.ts#L40
So users are forced to use `import { faker } from '@faker-js/faker';`

This will also help for auto-imports to directly use the requested import-syntax

Shinigami92:


ST-DDT:
Why exactly are you removing the switch imports docs?

Shinigami92:
> Why exactly are you removing the switch imports docs?

Because this is targeted for v7 and we will not need to further document such stuff from then on.

ST-DDT (approved):

prisis (approved):

PR #800, opened 2022-04-07
feat: resettable unique store

Shinigami92:
#810 should be merged first


src/utils/unique.ts
```diff
@@ -68,6 +73,7 @@ export function exec<Method extends (...parameters) => RecordKey>(
   opts.maxRetries = opts.maxRetries || 50;
   opts.exclude = opts.exclude || exclude;
   opts.compare = opts.compare || defaultCompare;
+  opts.store = opts.store || GLOBAL_UNIQUE_STORE;
```
ST-DDT, line 76: What about immutable options object? (not necessarily the values in it)

test/unique.spec.ts
```diff
@@ -161,4 +161,22 @@ describe('unique', () => {
       })
     ).toThrow();
   });
+
+  it('should be possible to pass a user-specific store', () => {
+    const store = {};
+
+    const method = () => 'with conflict: 0';
+
+    expect(store).toEqual({});
```
ST-DDT, line 170: No need to check that here.

ST-DDT:
> I'm not yet sure how to allow user to reset props inside `GLOBAL_UNIQUE_STORE`.

Don't think too much.
Either move it inside the unique class (effectively removing it global state) or leave it there until we can remove it in v7.

test/unique.spec.ts
```diff
@@ -161,4 +161,22 @@ describe('unique', () => {
...
+
+    expect(faker.unique(method, [], { store })).toBe('with conflict: 0');
```
Shinigami92, line 172:
```suggestion
    expect(faker.unique(method, [], { store })).toBe('with conflict: 0');
```

ST-DDT (approved):

pkuczynski (approved):

PR #801, opened 2022-04-07
feat: color module

### LINKED ISSUE #286: Add more random color generators

As a developer using faker I want to generate random hexadecimal colors so that I can use it to generate random colors for things in my applications.

One use case example is the ability to generate a random color for a team/user in the application.

**I'd love to submit a PR for this if it's a welcomed change! 😄**

### Suggested solution

I'm not positive which namespace this would fit under, but an implementation like this could work to generate the color.

```js
const hexColorOptions = ['0','1','2','3','4','5','6','7','8','9','A','B','C','D','E','F'];
const color = ['#'];

for (let i = 0; i < 6; i++) {
  const index = faker.datatype.number({
    min: 0,
    max: hexColorOptions.length - 1,
  });

  color.push(hexColorOptions[index]);
}

return color.join('');
```

### Alternative

We could also generate RGBA, CMYK, HSL colors as well.

### Additional context

_No response_

harsohailB:
## Resolves:

#286 

## Key Changes:

- [x]  Migrate color API from `commerce` to its own `color` module
- [x] Deprecate color method in `commerce` module (since v7 until v8)
- [x] Add random `RGB` generation and tests (Red Green Blue)
- [x] Add random `RGBA` generation and tests (Red Green Blue Alpha)
- [x] Add random `CMYK` generation and tests (Cyan, Magenta, Yellow, and Key (black))
- [x] Add random `HSL` generation and tests (Hue, Saturation, Lightness)
- [x] Add random `HSLA` generation and tests (Hue, Saturation, Lightness, Alpha)
- [x] Add random `HWB` generation and tests (Hue, Whiteness, Blackness)
- [x] Add random `LAB` generation and tests (Lightness, followed by the A-axis (green to red) and B-axis (blue to yellow))
- [x] Add random `LCH` generation and tests (Lightness, Chroma, and Hue)
- [x] Migrate existing tests from `commerce.spec.ts` to `color.spec.ts`

**Add support for color gamuts:** (see those supported by CSS media queries [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut))

- [x] sRGB
- [x] display-p3
- [x] rec2020 
- [x] Add random `space` generation and tests (color space list [here](https://en.wikipedia.org/wiki/List_of_color_spaces_and_their_uses))

**Add CSS support:** (perhaps in `options` have `{ format: 'css' }`)

- [x] RGB - `rgb(255, 0, 0)` (covers sRGB gamut for CSS)
- [x] RGBA - `rgba(255, 0, 0, 0.5)`
- [x] HSL - `hsl(0deg 100% 80%)`
- [x] HSLA - `hsl(0deg 100% 50% / 0.5)`
- [x] HWB - `hwb(194 0% 0%)`
- [x] CMYK - `cmyk(100%, 0%, 0%, 0%)`
- [x] LAB - `lab(29.2345% 39.3825 20.0664)`
- [x] LCH - `lch(52.2345% 72.2 56.2)`
- [x] display-p3 - `color(display-p3 0 1 0)`

**Design Decisions: (WIP)**

- [x] Combine `hex` and `numeric` methods and provide the option in args
- [x] Include `binary` as a return format

## CSS Colors Reference

[Modern CSS Colors](https://www.smashingmagazine.com/2021/11/guide-modern-css-colors/)


prisis:
Thank you very much for your Pull Request :) we a discussing in what version of faker it would make sense to put this feature into, but it looks like it will go into faker v7. We will update you on the process :)

src/color.ts
```diff
@@ -0,0 +1,114 @@
...
+    return this.faker.random.arrayElement(this.faker.definitions.color.human);
+  }
+
+  /**
+   * Returns a RGB color in hex format
+   *
+   * @example
+   * faker.color.rgb() // '0xffffff'
+   */
+  rgb(): string {
```
Shinigami92, line 30: For hex colors I would assume `rgbHex`

---

Additionally, we could provide optional options for 

- allow only upper or lower or mixed case
- prefix or not prefix with '0x'

src/color.ts
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @example
+   * faker.color.human() // 'red'
+   */
+  human(): string {
+    return this.faker.random.arrayElement(this.faker.definitions.color.human);
+  }
+
+  /**
+   * Returns a RGB color in hex format
```
Shinigami92, line 25: Please end JSDoc sentence with period `.`

src/color.ts
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @example
+   * faker.color.rgb() // '0xffffff'
+   */
+  rgb(): string {
+    return this.faker.datatype.hexadecimal(6);
+  }
+
+  /**
+   * Returns a RGB color in decimal format
```
Shinigami92, line 35: I'm not sure about `decimal` or another wording here, because I see a difference between `[255, 255, 255]` and `[1.0, 1.0, 1.0]`.
Both could be used e.g. for something like `three.js`

src/color.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  rgba(): string {
+    return this.faker.datatype.hexadecimal(8);
+  }
+
+  /**
+   * Returns a RGBA color in decimal format
+   *
+   * @example
+   * faker.color.rgbaNumeric() // '[255, 255, 255, 0.54]'
```
Shinigami92, line 60: Looks weird for me to suddenly treat alpha channel differently 🤔 
What were your source for relying on a base/definition?

src/color.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Return a percentage value in decimal format betwene 0 and 1
+   * with percision of two decimal place
+   *
+   * @example
+   * this.getPercentage() // 0.36
+   */
+  private getPercentage(): number {
```
Shinigami92, line 111: We do not use `private` functions, instead inline the code or use a non-exported function above the class.

It may change on the road to v7+ 🤔 But currently we iterate in test over all member functions of a module and test them.
JS doesn't have the ability to detect that a function was marked with private due to it is just a TypeScript concept and not real-js.

src/locales/ar/commerce/index.ts
```diff
@@ -3,14 +3,12 @@
...
 import department from './department';
 import product_name from './product_name';
 
-const commerce: Partial<CommerceDefinitions> = {
-  color,
+const commerce = {
   department,
   product_name,
-};
+} as Partial<CommerceDefinitions>;
```
ST-DDT, line 12: Please run `pnpm run generate:locales` again.

ST-DDT (commented):
I'll review it, when we start with v7.

Shinigami92 (changes requested):
We should think about if we want to go more in a direction of supporting CSS stuff or more like generating values for e.g. `three.js`.

For CSS I would also like to see something like get a [color gamut](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut) `display-p3`

But for three.js I would like to have e.g. possibility to generate colors for a specific channel or stuff like that: https://threejs.org/docs/#api/en/math/Color

And maybe there is also something else non-css/non-three.js

src/commerce.ts
```diff
@@ -14,18 +14,6 @@ export class Commerce {
...
 
-  /**
-   * Returns a human readable color name.
-   *
-   * @example
-   * faker.commerce.color() // 'red'
-   */
-  color(): string {
-    return this.faker.random.arrayElement(
-      this.faker.definitions.commerce.color
```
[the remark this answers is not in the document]
Shinigami92: Yeah, we should reuse the new color function and write a deprecation message to remove it in v8

src/locales/ar/commerce/index.ts
harsohailB: Generated!

src/color.ts
harsohailB: At the bottom of this source, a floating point between 0 and 1 is used to show the alpha percentage:
https://www.w3schools.com/css/css_colors_rgb.asp

harsohailB:
> We should think about if we want to go more in a direction of supporting CSS stuff or more like generating values for e.g. `three.js`.
> 
> For CSS I would also like to see something like get a [color gamut](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut) `display-p3`
> 
> But for three.js I would like to have e.g. possibility to generate colors for a specific channel or stuff like that: https://threejs.org/docs/#api/en/math/Color
> 
> And maybe there is also something else non-css/non-three.js

@Shinigami92 I will look into this further and update the PR description with new tasks.

src/color.ts
harsohailB: Done!

src/commerce.ts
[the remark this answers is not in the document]
harsohailB: Completed in commit c7f255e. If someone can confirm that the change is correct, that would be awesome!

src/color.ts
harsohailB: Done!

src/color.ts
harsohailB: Should we allow `options` here to specify the numeric type to be `int` or `float` for clarity (will default to `int`)? Docs can be changed to "numeric format" instead of "decimal format"?

src/commerce.ts
```diff
@@ -19,11 +20,17 @@ export class Commerce {
...
-    );
+    deprecated({
+      deprecated: 'faker.commerce.color()',
+      proposed: 'faker.color.human()',
+      since: 'v7.0.0',
+      until: 'v8.0.0',
+    });
+    return this.faker.random.arrayElement(this.faker.definitions.color.human);
```
Shinigami92, line 33: You should call 
```suggestion
    return this.faker.color.human();
```

Shinigami92 (changes requested):

src/color.ts
ST-DDT: How about merging rgbHex and rgbNumeric into just `rgb(options: { format: 'hex', 'number255', 'percentage' } )`?
(Not sure about naming it `'number255'`, maybe just `'number'` to work better with the other color formats?)

We would need multiple signatures to make most use of TS features though.

Shinigami92: What about naming it `8bit`?

ST-DDT: The bit/byte representation is not exactly standardized. It can be binary, decimal or hex.

harsohailB: Should we split decimal into `int` and `float` though to support something like three.js as @Shinigami92 mentioned in the first comment? So format would be: `binary`, `int`, `float`, or `hex`

src/color.ts
harsohailB: Created a percentage function in commerce module as I think it fits there. Used that in this color module.

src/commerce.ts
```diff
@@ -38,6 +45,24 @@ export class Commerce {
...
+   * Return a percentage value in decimal format between 0 and 1
+   * with percision of two decimal place.
+   *
+   * @param percision Precision of the generated number. Defaults to `0.01`.
+   *
+   * @example
+   * this.getPercentage(0.1) // 0.4
+   * this.getPercentage(0.01) // 0.36
+   */
+  percentage(percision: number): number {
```
ST-DDT, line 58: This method should probably in our future `number` module.

src/color.ts
```diff
@@ -0,0 +1,334 @@
...
+      css = `lab(${percentages[0]}% ${values[1]} ${values[2]})`;
+      break;
+    case 'lch':
+      css = `lch(${percentages[0]}% ${values[1]} ${values[2]})`;
+      break;
+    case 'display-p3':
+      css = `color(display-p3 ${values[0]} ${values[1]} ${values[2]})`;
+      break;
+  }
+  return css;
```
ST-DDT, line 99: Maybe return directly from with the case statements
and add a default case that throws for unsupported cases or uses rgb as default.

src/color.ts
```diff
@@ -0,0 +1,334 @@
...
+   * @param options.format Format of generated RGB color. Defaults to `hex`.
+   * @param options.includeAlpha Adds an alpha value to the color (RGBA). Defaults to `false`.
+   *
+   * @example
+   * faker.color.rgb() // '0xffffFF'
+   * faker.color.rgb({ prefix: '#' }) // '#ffffFF'
+   * faker.color.rgb({ case: 'upper' }) // '0xFFFFFF'
+   * faker.color.rgb({ case: 'lower' }) // '0xffffff'
+   * faker.color.rgb({ prefix: '#', case: 'lower' }) // '#ffffff'
+   * faker.color.rgb({ format: 'decimal' }) // '[255, 255, 255]'
```
ST-DDT, line 168: Does it return a string in this case? If not, remove the confusing single quotes.

src/color.ts
```diff
@@ -0,0 +1,334 @@
...
+ * @param values Array of color values to be converted.
+ * @param format Format of generated RGB color.
+ * @param colorSpace Color space to format CSS string for. Defaults to `rgb`.
+ */
+function toColorFormat(
+  values: number[],
+  format: 'decimal' | 'css' | 'binary',
+  colorSpace?: ColorSpace
+): string | number[] {
+  colorSpace = colorSpace || 'rgb';
```
ST-DDT, line 114: Maybe move the fallback to the signature?

src/commerce.ts
harsohailB: Is there a ticket for this where I can mention it?

src/color.ts
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
harsohailB: I'm not quite understand what you mean here. Can you please elaborate?

src/commerce.ts
ST-DDT: https://github.com/faker-js/faker/discussions/805

src/color.ts
ST-DDT:
````ts
function toColorFormat(
  values: number[],
  format: 'decimal' | 'css' | 'binary',
  colorSpace: ColorSpace = 'rgb'
): string | number[] {
````

Comment (harsohailB) [reply src/color.ts]: Oh yea, that would be much cleaner. Thanks for pointing that out! Done!

File: src/commerce.ts (hunk 9 above)
Comment (pkuczynski) [reply src/commerce.ts:58]: It really does not make a lot of sense to create a function which has almost exactly the same signature like already existing one. I can easily imagine that in not that long time someone will create a PR extending `percentage` with `min` and `max` options and then it will be identical as `float`.

If I would see `percentage` without looking at the documentation I would expect it generates something like `'25%‘`. And not a float value...

So I am against this new method. Let's keep things [stupid simple](https://en.wikipedia.org/wiki/KISS_principle)

File: src/color.ts [hunk 13]
```diff
@@ -0,0 +1,337 @@
+import type { Faker } from '.';
+
+type ColorSpace =
```
Shinigami92, line 3: In this example, the color spaces are define much differently :thinking: 
https://www.w3.org/TR/css-color-4/#terminology

harsohailB: If I understand it correctly now, is the `colorGamut()` function supported to be a `colorSpace()` generator and the CSS string generator should take in a `ColorGamut` type?

Shinigami92: > If I understand it correctly now, is the `colorGamut()` function supposed to be a `colorSpace()` generator and the CSS string generator should take in a `ColorGamut` type?

I think yes :thinking: 
But also it seems we need both, `colorGamut` (or maybe just `gamut()`, because we are already in the color module) and `colorSpace` (or maybe just `space`?)
`colorSpace` can return values that can be applies as the CSS value, so e.g. `display-p3` or `xyz`

A helpful video I can recommend it: https://www.youtube.com/watch?v=cGyLHxn16pE

harsohailB: I think `gamut()` should return the values that can be applied as the CSS value as gamut is a [CSS media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut).

Shinigami92: @harsohailB Yes!

src/color.ts
```diff
@@ -0,0 +1,338 @@
+import type { Faker } from '.';
+import type gamut from './locales/en/color/gamut';
+
+type Gamut = typeof gamut[number];
```
harsohailB, line 4: I am not sure if this is the best way to do this. If someone can advise, that would be great!

src/color.ts
harsohailB: Done in [0272c06](https://github.com/faker-js/faker/pull/801/commits/0272c06cb17fc7ef2a60387f7e554d7a94b28dd7)!

harsohailB: Done in [0272c06](https://github.com/faker-js/faker/pull/801/commits/0272c06cb17fc7ef2a60387f7e554d7a94b28dd7)!

src/locales/en/color/gamut.ts
```diff
@@ -0,0 +1,11 @@
+export default [
+  'rgb',
+  'rgba',
+  'hsl',
+  'hsla',
+  'hwb',
+  'cmyk',
+  'lab',
+  'lch',
+  'display-p3',
```
Shinigami92, line 10: The [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut.#syntax) says there are only these 3 values: 'srgb', 'p3' and 'rec2020'
From where are the others from?

src/locales/en/color/space.ts
```diff
@@ -0,0 +1,43 @@
...
+  'Munsell Color System',
+  'Natural Color System (NSC)',
+  'Pantone Matching System (PMS)',
+  'RAL',
+  'Federal Standard 595C',
+  'British Standard Colour (BS)',
+  'HKS',
+  'LMS',
+  'RG',
+  'RGK',
```
Shinigami92, line 42: I would assume that these are the values `'sRGB', 'display-p3', 'a98-rgb', 'prophoto-rgb', 'rec2020', 'xyz'` as the values described here https://drafts.csswg.org/css-color/#terminology (First param of `color`)

The current values of this file could be still shipped but under a different name :thinking: How about "space human"? like readable human colors are defined also?
Maybe `spaceTerms`?

Shinigami92 (changes requested):

src/locales/en/color/gamut.ts
Shinigami92: If you need the current values, they may need to be stored on a different extra place

src/locales/en/color/space.ts
harsohailB: I am confused because https://drafts.csswg.org/css-color/#terminology states the first param of `color` as color space but https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut states it as gamut...

src/locales/en/color/gamut.ts
harsohailB: Check https://www.smashingmagazine.com/2021/11/guide-modern-css-colors/ for the additional gamuts supported by CSS.

src/commerce.ts
harsohailB: @pkuczynski You make some really good points here! @ST-DDT and @Shinigami92 what do you think about this? I am happy to revert to using `float` inline to keep it simpler.

src/locales/en/color/space.ts
Shinigami92: > I am confused because [drafts.csswg.org/css-color/#terminology](https://drafts.csswg.org/css-color/#terminology) states the first param of `color` as color space but [developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut) states it as gamut...

Yes, I'm also confused by that :thinking:
But there is definitely a difference
The one is pointing out `p3` (I assume gamut) and the other `display-p3` (I assume space)

src/locales/en/color/gamut.ts
Shinigami92: Ah so I assume your actual defined values are the color "function"
But `cmyk` and `display-p3` are wrong for them

harsohailB: Oh I see. I should probably not return `color(` with `display-p3`. How is CMYK wrong?

src/locales/en/color/space.ts
harsohailB: [Here](https://en.wikipedia.org/wiki/DCI-P3) it states the opposite :laughing: 

> **P3** is an [RGB **color space**](https://en.wikipedia.org/wiki/RGB_color_space). DCI-P3 (Digital Cinema Initiative) is used with digital theatrical motion picture distribution[[1]](https://en.wikipedia.org/wiki/DCI-P3#cite_note-Technicolor-1) ([DCDM](https://www.loc.gov/preservation/digital/formats/fdd/fdd000177.shtml)). **Display-P3** is a variant developed by [Apple Inc.](https://en.wikipedia.org/wiki/Apple_Inc.) for **wide-gamut displays**.

src/locales/en/color/gamut.ts
Shinigami92: Oh, seems like `cmyk` exists for PhotoShop :thinking: (https://www.mediaevent.de/css/color-cmyk.html)
But not sure if we should put this into here.

Also instead of `display-p3` you may want `color`
I just had a look into your description of this PR
image
And `display-p3` is the first parameter, not the function-name itself, that is `color` in this case.

Shinigami92: ... https://www.w3schools.com/colors/colors_cmyk.asp :eyes: 
Why don't they have just consistent documentation at one place 😬

harsohailB: The W3 is the source I used for CMYK. And I see what you mean by  `color` is needed for `display-p3`. So no changes needed?

Shinigami92: > The W3 is the source I used for CMYK. And I see what you mean by color is needed for display-p3. So no changes needed?

I assume we need to just replace `display-p3` with `color` as `display-p3` is the possible parameter and `color` is just a possible function like `rgba` or `hsla`
Then rename it file to `function` and add a `gamut` file with the 3 possible values

src/locales/en/color/space.ts
Shinigami92: Oof... But I think we should stay close to CSS docs and not Wikipedia :thinking: Even if it is theoretically wrong :thinking:

Shinigami92: image
These are color `space`s and the (so I call them) first parameter for the color function

---

image
This is the color `gamut` (for media queries)

Shinigami92: @harsohailB Please do not misunderstand me and accidentally throw away the current defined values, these should just be relocated into another file

src/commerce.ts
Shinigami92: For now I think yes, using it just inline is okay for now

src/locales/en/color/space.ts
harsohailB: No worries! I really appreciate the clarifications you are making. I will try to take this in and hopefully get us closer in the next commit. Will re-request review from you once I make the new files for CSS functions vs spaces vs gamuts.

Shinigami92:
@harsohailB From tomorrow on some of us in the team (also me) are in vacation, so I expect it will get very silent for the next days.
I'm back on next saturday.

src/locales/en/color/space.ts
harsohailB: Refactored the locales and some functions. Let me know what you think!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * Converts an array of numbers into binary string format.
+ *
+ * @param values Array of values to be converted.
+ */
+function toBinary(values: number[]): string {
+  const binary: string[] = values.map((value: number) => {
+    const isFloat: boolean = Number(value) === value && value % 1 !== 0;
+    if (isFloat) {
+      const buffer: ArrayBuffer = new ArrayBuffer(4);
```
Shinigami92, line 45: Type can be inferred
```suggestion
      const buffer = new ArrayBuffer(4);
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+  });
+  return binary.join(' ');
+}
+
+/**
+ * Converts an array of numbers into CSS accepted format.
+ *
+ * @param values Array of values to be converted.
+ * @param cssFunction CSS function to be generated for the color. Defaults to `rgb`.
+ * @param space Color space to format CSS color function with. Defaults to `srgb`.
```
Shinigami92, line 60:
```suggestion
 * @param space Color space to format CSS color function with. Defaults to `sRGB`.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+  }
+}
+
+/**
+ * Converts an array of color values to the specified color format.
+ *
+ * @param values Array of color values to be converted.
+ * @param format Format of generated RGB color.
+ * @param cssFunction CSS function to be generated for the color. Defaults to `rgb`.
+ * @param space Color space to format CSS color function with. Defaults to `srgb`.
```
Shinigami92, line 98:
```suggestion
 * @param space Color space to format CSS color function with. Defaults to `sRGB`.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+  cssSupportedSpace(): string {
+    return this.faker.random.arrayElement(
+      this.faker.definitions.color.cssSpaces
+    );
+  }
+
+  /**
+   * Returns a RGB color.
+   *
+   * @param options options object.
```
Shinigami92, line 182:
```suggestion
   * @param options Options object.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+      color.push(this.faker.commerce.percentage(0.01));
+      cssFunction = 'rgba';
+    }
+    return toColorFormat(color, options.format, cssFunction);
+  }
+
+  /**
+   * Returns a CMYK color.
+   *
+   * @param options options object.
```
Shinigami92, line 227:
```suggestion
   * @param options Options object.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+   * @example
+   * faker.color.cmyk() // [0.31, 0.52, 0.32, 0.43]
+   * faker.color.cmyk({ format: 'decimal' }) // [0.31, 0.52, 0.32, 0.43]
+   * faker.color.cmyk({ format: 'css' }) // cmyk(100%, 0%, 0%, 0%)
+   * faker.color.cmyk({ format: 'binary' }) // (8-32 bits) x 4
+   */
+  cmyk(options?: { format?: 'decimal' | 'css' | 'binary' }): string | number[] {
+    const color: string | number[] = [0, 0, 0, 0].map(() =>
+      this.faker.commerce.percentage(0.01)
+    );
```
Shinigami92, line 239:
```suggestion
    const color = Array.from({ length: 4 }, (() =>
      this.faker.commerce.percentage(0.01)
    ));
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+
+    return toColorFormat(color, options?.format || 'decimal', 'cmyk');
+  }
+
+  /**
+   * Returns a HSL color.
+   *
+   * @param options options object.
```
Shinigami92, line 247:
```suggestion
   * @param options Options object.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+    }
+    return toColorFormat(
+      hsl,
+      options?.format || 'decimal',
+      options?.includeAlpha ? 'hsla' : 'hsl'
+    );
+  }
+
+  /**
+   * Returns a HWB color.
```
Shinigami92, line 276:
```suggestion
   * Returns an HWB color.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+    for (let i = 0; i < 2; i++) {
+      hsl.push(this.faker.commerce.percentage(0.01));
+    }
+    return toColorFormat(hsl, options?.format || 'decimal', 'hwb');
+  }
+
+  /**
+   * Returns a LAB (CIELAB) color.
+   *
+   * @param options options object.
```
Shinigami92, line 298:
```suggestion
   * @param options Options object.
```

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+    return toColorFormat(lab, options?.format || 'decimal', 'lab');
+  }
+
+  /**
+   * Returns a LCH color. Even though upper bound of
+   * chroma in LCH color space is theoretically unbounded,
+   * it is bounded to 230 as anything above will not
+   * make a noticable difference in the browser.
+   *
+   * @param options options object.
```
Shinigami92, line 323:
```suggestion
   * @param options Options object.
```

src/commerce.ts
```diff
@@ -38,6 +45,24 @@ export class Commerce {
     );
   }
 
+  /**
+   * Return a percentage value in decimal format between 0 and 1
+   * with percision of two decimal place.
+   *
+   * @param percision Precision of the generated number. Defaults to `0.01`.
```
Shinigami92, line 52:
```suggestion
   * @param precision Precision of the generated number. Defaults to `0.01`.
```

src/commerce.ts
```diff
@@ -38,6 +45,24 @@ export class Commerce {
     );
   }
 
+  /**
+   * Return a percentage value in decimal format between 0 and 1
+   * with percision of two decimal place.
```
Shinigami92, line 50:
```suggestion
   * with precision of two decimal place.
```

src/commerce.ts
Shinigami92, line 58:
```suggestion
  percentage(precision: number = 0.01): number {
```

src/commerce.ts
```diff
@@ -38,6 +45,24 @@ export class Commerce {
...
+   *
+   * @example
+   * this.getPercentage(0.1) // 0.4
+   * this.getPercentage(0.01) // 0.36
+   */
+  percentage(percision: number): number {
+    return this.faker.datatype.float({
+      min: 0,
+      max: 1,
+      precision: percision || 0.01,
```
Shinigami92, line 62:
```suggestion
      precision,
```

Shinigami92 (changes requested):

src/commerce.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * Formats the hex format of a generated color string according
+ * to options specified by user.
+ *
+ * @param hexColor Hex color string to be formated.
+ * @param options options object.
+ * @param options.prefix Prefix of the generated hex color. Defaults to `0x`.
+ * @param options.case Letter case of the generated hex color. Defaults to `mixed`.
+ */
+function applyHexFormat(
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * Converts an array of numbers into binary string format.
+ *
+ * @param values Array of values to be converted.
+ */
+function toBinary(values: number[]): string {
+  const binary: string[] = values.map((value: number) => {
+    const isFloat: boolean = Number(value) === value && value % 1 !== 0;
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+ * @param values Array of values to be converted.
+ * @param cssFunction CSS function to be generated for the color. Defaults to `rgb`.
+ * @param space Color space to format CSS color function with. Defaults to `srgb`.
+ */
+function toCSS(
+  values: number[],
+  cssFunction: CSSFunction = 'rgb',
+  space: CSSSpace = 'sRGB'
+): string {
+  const percentages = values.map((value: number) => Math.round(value * 100));
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+  cmyk(options?: { format?: 'decimal' | 'css' | 'binary' }): string | number[] {
+    const color: string | number[] = [0, 0, 0, 0].map(() =>
+      this.faker.commerce.percentage(0.01)
+    );
+
+    return toColorFormat(color, options?.format || 'decimal', 'cmyk');
+  }
+
+  /**
+   * Returns a HSL color.
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+      hsl,
+      options?.format || 'decimal',
+      options?.includeAlpha ? 'hsla' : 'hsl'
+    );
+  }
+
+  /**
+   * Returns a HWB color.
+   *
+   * @param options options object.
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+    for (let i = 0; i < 2; i++) {
+      lab.push(
+        this.faker.datatype.float({ min: -100, max: 100, precision: 0.0001 })
+      );
+    }
+    return toColorFormat(lab, options?.format || 'decimal', 'lab');
+  }
+
+  /**
+   * Returns a LCH color. Even though upper bound of
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+        this.faker.datatype.number({ min: 0, max: 230, precision: 0.1 })
+      );
+    }
+    return toColorFormat(lch, options?.format || 'decimal', 'lch');
+  }
+
+  /**
+   * Returns a random color based on CSS color space specified.
+   *
+   * @param options options object.
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    color = [0, 0, 0].map(() =>
+      this.faker.datatype.number({ min: 0, max: 255 })
+    );
+    if (options?.includeAlpha) {
+      color.push(this.faker.commerce.percentage(0.01));
+      cssFunction = 'rgba';
+    }
+    return toColorFormat(color, options.format, cssFunction);
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,370 @@
...
+   * faker.color.colorByCSSColorSpace({ format: 'binary' }) // (8-32 bits x 3)
+   */
+  colorByCSSColorSpace(options?: {
+    format?: 'decimal' | 'css' | 'binary';
+    space?: CSSSpace;
+  }): string | number[] {
+    if (options?.format === 'css' && !options?.space) {
+      options = { ...options, space: 'sRGB' };
+    }
+    const color = [0, 0, 0].map(() => this.faker.commerce.percentage(0.0001));
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
harsohailB: Done!

src/commerce.ts
harsohailB: Change no longer required as percentage generator is commerce is now deleted. Refer to [this](https://github.com/faker-js/faker/pull/801#discussion_r846716417).

src/commerce.ts
harsohailB: Change no longer required as percentage generator is commerce is now deleted. Refer to [this](https://github.com/faker-js/faker/pull/801#discussion_r846716417).

src/commerce.ts
harsohailB: Change no longer required as percentage generator is commerce is now deleted. Refer to [this](https://github.com/faker-js/faker/pull/801#discussion_r846716417).

src/commerce.ts
harsohailB: Changed to inline and removed from commerce.

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+    case 'hwb':
+      return `hwb(${values[0]} ${percentage(values[1])}% ${percentage(
+        values[2]
+      )}%)`;
+    case 'lab':
+      return `lab(${percentage(values[0])}% ${values[1]} ${values[2]})`;
+    case 'lch':
+      return `lch(${percentage(values[0])}% ${values[1]} ${values[2]})`;
+    default:
+      return toCSS(values, 'rgb');
```
Shinigami92, line 99: On the first look that looks like an infinit loop (it's not)
But I would rewrite it to e.g.

```ts
    case 'rgb':
    default:
      return `rgb(${values[0]}, ${values[1]}, ${values[2]})`;
```
And remove the first case.

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+  let result: string | number[];
+  switch (format) {
+    case 'css':
+      result = toCSS(values, cssFunction, space);
+      break;
+    case 'binary':
+      result = toBinary(values);
+      break;
+  }
+  return result;
```
Shinigami92, line 128:
```suggestion
  switch (format) {
    case 'css':
      return toCSS(values, cssFunction, space);
    case 'binary':
      return toBinary(values);
    default:
      return values;
  }
```

Shinigami92 (changes requested):

src/color.ts
```diff
@@ -0,0 +1,388 @@
+import type { Faker } from '.';
+import type cssFunctions from './locales/en/color/css_functions';
+import type cssSpaces from './locales/en/color/css_spaces';
+
+type CSSFunction = typeof cssFunctions[number];
+type CSSSpace = typeof cssSpaces[number];
```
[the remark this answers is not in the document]
pkuczynski: Agree 100x!

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+   * @param options.case Letter case of the generated hex color. Defaults to `mixed`. Only applied when 'hex' format is used.
+   * @param options.format Format of generated RGB color. Defaults to `hex`.
+   * @param options.includeAlpha Adds an alpha value to the color (RGBA). Defaults to `false`.
+   *
+   * @example
+   * faker.color.rgb() // '0xffffFF'
+   * faker.color.rgb({ prefix: '#' }) // '#ffffFF'
+   * faker.color.rgb({ case: 'upper' }) // '0xFFFFFF'
+   * faker.color.rgb({ case: 'lower' }) // '0xffffff'
+   * faker.color.rgb({ prefix: '#', case: 'lower' }) // '#ffffff'
```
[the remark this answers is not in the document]
pkuczynski: Agree...

src/color.ts
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/**
+ * Converts an array of numbers into binary string format.
+ *
+ * @param values Array of values to be converted.
+ */
+function toBinary(values: number[]): string {
+  const binary: string[] = values.map((value) => {
+    const isFloat = value === value && value % 1 !== 0;
```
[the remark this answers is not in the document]
harsohailB: First value should be `Number(value)`. Changed!

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+   * @example
+   * faker.color.rgb() // '0xffffFF'
+   * faker.color.rgb({ prefix: '#' }) // '#ffffFF'
+   * faker.color.rgb({ case: 'upper' }) // '0xFFFFFF'
+   * faker.color.rgb({ case: 'lower' }) // '0xffffff'
+   * faker.color.rgb({ prefix: '#', case: 'lower' }) // '#ffffff'
+   * faker.color.rgb({ format: 'decimal' }) // [255, 255, 255]
+   * faker.color.rgb({ format: 'css' }) // 'rgb(255, 0, 0)'
+   * faker.color.rgb({ format: 'binary' }) // '10000000 00000000 11111111'
+   * faker.color.rgb({ format: 'decimal', includeAlpha: true }) // [255, 255, 255, 0.4]
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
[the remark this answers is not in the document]
harsohailB: I agree. I will make this change.

src/color.ts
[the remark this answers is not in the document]
harsohailB: Done!

src/definitions/color.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  human: string[];
+  /**
+   * Color space names.
+   */
+  space: string[];
+  /**
+   * Color gamut names supported by CSS.
+   */
+  cssSpaces: readonly string[];
```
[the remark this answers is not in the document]
harsohailB: This should not be an issue anymore as moved the css functions and spaces into `color.ts`

src/color.ts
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
```diff
@@ -0,0 +1,388 @@
...
+   */
+  rgb(options?: {
+    prefix?: string;
+    case?: 'upper' | 'lower' | 'mixed';
+    format?: 'hex' | ColorFormat;
+    includeAlpha?: boolean;
+  }): string | number[] {
```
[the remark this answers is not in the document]
harsohailB: Added in [11fdcc8](https://github.com/faker-js/faker/pull/801/commits/11fdcc87847eb404e3a0c6cbc464cc8bd8f0a822). Let me know if this is correct!

Shinigami92 (changes requested):
I'm not sure about the new naming `typeCase` :thinking: 
I think we already have the option `case` for something else, but I didn't checked that yet

harsohailB:
> I'm not sure about the new naming `typeCase` thinking I think we already have the option `case` for something else, but I didn't checked that yet

Option case was giving parsing issues such as when I was trying to `const { case = 'lower' } = options` it wanted to treat case as a case in switch statement. Therefore, I renamed to `typeCase`

src/color.ts
```diff
@@ -0,0 +1,453 @@
...
+}
+
+/**
+ * Converts an array of numbers into binary string format.
+ *
+ * @param values Array of values to be converted.
+ */
+function toBinary(values: number[]): string {
+  const binary: string[] = values.map((value) => {
+    const isFloat = Number(value) === value && value % 1 !== 0;
```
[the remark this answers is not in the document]
harsohailB: Yea I guess if we are passing in a `number[]` we do not need that first condition as it'll always be true. Will take that out.

src/color.ts
```diff
@@ -0,0 +1,453 @@
+import type { Faker } from '.';
+
+/**
+ * Color space names supported by CSS.
+ */
+export const cssSpaces = [
```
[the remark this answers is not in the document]
harsohailB: Done!

src/color.ts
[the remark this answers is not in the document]
harsohailB: Done!

Shinigami92:
> Option case was giving parsing issues such as when I was trying to `const { case = 'lower' } = options` it wanted to treat case as a case in switch statement. Therefore, I renamed to `typeCase`

Yes you are right! I checked the other places and they use `upcase: boolean` (we don't want that here)
We could use `casing`

Shinigami92:
> Option case was giving parsing issues such as when I was trying to `const { case = 'lower' } = options` it wanted to treat case as a case in switch statement. Therefore, I renamed to `typeCase`

Yes you are right! I checked the other places and they use `upcase: boolean` (we don't want that here)
We could use `casing`

harsohailB:
> > Option case was giving parsing issues such as when I was trying to `const { case = 'lower' } = options` it wanted to treat case as a case in switch statement. Therefore, I renamed to `typeCase`
> 
> Yes you are right! I checked the other places and they use `upcase: boolean` (we don't want that here) We could use `casing`

Done!

Shinigami92 (dismissed):

Shinigami92:
Hi @harsohailB 👋
Could you update this PR? We moved modules into `src/modules/<modulename>/index.ts`

harsohailB:
> Hi @harsohailB 👋 Could you update this PR? We moved modules into `src/modules/<modulename>/index.ts`

Will do!

harsohailB:
```
test/scripts/apidoc/examplesAndDeprecations.spec.ts > examples and deprecations > Color > cmyk
     → Color.cmyk to have examples: expected '' not to be '' // Object.is equality
```

Receiving these failures from the tests. Not sure why this is happening as there are examples for each method in the color module. If someone could point me in the right direction with this, that would be much appreciated!

Shinigami92:
> ```
> test/scripts/apidoc/examplesAndDeprecations.spec.ts > examples and deprecations > Color > cmyk
>      → Color.cmyk to have examples: expected '' not to be '' // Object.is equality
> ```
> 
> Receiving these failures from the tests. Not sure why this is happening as there are examples for each method in the color module. If someone could point me in the right direction with this, that would be much appreciated!

Ah! @ST-DDT's test is working here, but the message is maybe a bit cryptic
It says that the JSDoc for cmyk function is missing `@example`
We are now forcefully requiring these as we want to use these code example to auto-call them in tests later on. @ST-DDT can tell you more about that if needed.
All you need to know is: add JSDoc with `@example` and orientate yourself on other already existing code 🙂

src/modules/color/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    return toColorFormat(color, format, cssFunction);
+  }
+
+  /**
+   * Returns a CMYK color.
+   *
+   * @param options Options object.
+   * @param options.format Format of generated CMYK color. Defaults to `decimal`.
+   *
+   * @example
```
harsohailB, line 281: @Shinigami92 But all the color module function have the @example JSDoc in them already? 🤔

Shinigami92: @ST-DDT 👀

src/locales/ar/commerce/index.ts
```diff
@@ -3,12 +3,10 @@
  * Run 'pnpm run generate:locales' to update.
  */
 import type { CommerceDefinitions } from '../../..';
-import color from './color';
 import department from './department';
 import product_name from './product_name';
 
-const commerce: CommerceDefinitions = {
-  color,
+const commerce: Partial<CommerceDefinitions> = {
```
ST-DDT, line 9: AFAICT you have to edit https://github.com/faker-js/faker/blob/main/scripts/generateLocales.ts#L50 and run `pnpm run generate:locales` again.

src/modules/color/index.ts
ST-DDT: We use the docs from the last api signature to generate the docs.
https://github.com/faker-js/faker/blob/6219c9414953021be50cb794b6658f23f2f65f0a/scripts/apidoc/moduleMethods.ts#L54
(You might have to duplicate the impl signature, because typedoc only reads api signatures.)
This is required because otherwise we would show `cmyk()` as the signature in the docs and typedoc removes tags without target. I haven't found a better way for that yet.

You can see this very easily when running `pnpm run docs:dev`:
grafik

Example:
https://github.com/faker-js/faker/blob/c115056e04d1e42f97c8d77daed3d9056c375953/src/random.ts#L174-L205

ST-DDT (changes requested):

src/modules/color/index.ts
```diff
@@ -0,0 +1,453 @@
...
+      cssFunction = 'rgba';
+    }
+    return toColorFormat(color, format, cssFunction);
+  }
+
+  /**
+   * Returns a CMYK color.
+   *
+   * @param options Options object.
+   * @param options.format Format of generated CMYK color. Defaults to `decimal`.
```
ST-DDT, line 279:
```suggestion
   * @param options.format Format of generated CMYK color. Defaults to `'decimal'`.
```

harsohailB: Done!

src/locales/ar/commerce/index.ts
harsohailB: Done!

src/modules/color/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    casing?: Casing;
+    format?: 'hex' | StringColorFormat;
+    includeAlpha?: boolean;
+  }): string;
+  rgb(options?: {
+    prefix?: string;
+    casing?: Casing;
+    format?: NumberColorFormat;
+    includeAlpha?: boolean;
+  }): number[];
```
Shinigami92, line 221: I know it's odd, but every single signature needs a JSDoc
the last one will be used for TSDoc but the others are used by e.g. VSCode
You can check that by manually trying it out

Shinigami92:
@harsohailB I will jump into this PR and fix everything until it's green

harsohailB:
> @harsohailB I will jump into this PR and fix everything until it's green

Thank you! Sorry been a little busy with school and work. I will try to contribute where ever I can still.

src/modules/color/index.ts
harsohailB: The docs issue is now resolved!

Shinigami92 (dismissed):

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * faker.color.rgb({ format: 'css' }) // 'rgb(255, 0, 0)'
+   * faker.color.rgb({ format: 'binary' }) // '10000000 00000000 11111111'
+   * faker.color.rgb({ format: 'decimal', includeAlpha: true }) // [255, 255, 255, 0.4]
+   */
+  rgb(options?: {
+    prefix?: string;
+    casing?: Casing;
+    format?: 'hex' | ColorFormat;
+    includeAlpha?: boolean;
+  }): string | number[] {
```
ST-DDT, line 291: This needs to be duplicated as api signature to be picked up by typedoc.

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * faker.color.cmyk({ format: 'decimal' }) // [0.31, 0.52, 0.32, 0.43]
+   * faker.color.cmyk({ format: 'css' }) // cmyk(100%, 0%, 0%, 0%)
+   * faker.color.cmyk({ format: 'binary' }) // (8-32 bits) x 4
+   */
+  cmyk(options?: { format?: ColorFormat }): string | number[] {
+    const color: string | number[] = Array.from({ length: 4 }).map(() =>
+      this.faker.datatype.float({ min: 0, max: 1, precision: 0.01 })
+    );
+    return toColorFormat(color, options?.format || 'decimal', 'cmyk');
+  }
```
ST-DDT, line 365: Same here

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * faker.color.hsl({ format: 'decimal', includeAlpha: true }) // [300, 0.21, 0.52, 0.28]
+   * faker.color.hsl({ format: 'css' }) // hsl(0deg, 100%, 80%)
+   * faker.color.hsl({ format: 'css', includeAlpha: true }) // hsl(0deg 100% 50% / 0.5)
+   * faker.color.hsl({ format: 'binary' }) // (8-32 bits) x 3
+   * faker.color.hsl({ format: 'binary', includeAlpha: true }) // (8-32 bits) x 4
+   */
+  hsl(options?: {
+    format?: ColorFormat;
+    includeAlpha?: boolean;
+  }): string | number[] {
```
ST-DDT, line 424: And here

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * @param options Options object.
+   * @param options.format Format of generated RGB color. Defaults to `decimal`.
+   *
+   * @example
+   * faker.color.hwb() // [201, 0.21, 0.31]
+   * faker.color.hwb({ format: 'decimal' }) // [201, 0.21, 0.31]
+   * faker.color.hwb({ format: 'css' }) // hwb(194 0% 0%)
+   * faker.color.hwb({ format: 'binary' }) // (8-32 bits x 3)
+   */
+  hwb(options?: { format?: ColorFormat }): string | number[] {
```
ST-DDT, line 478: And here

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * @param options Options object.
+   * @param options.format Format of generated RGB color. Defaults to `decimal`.
+   *
+   * @example
+   * faker.color.lab() // [0.832133, -80.3245, 100.1234]
+   * faker.color.lab({ format: 'decimal' }) // [0.856773, -80.2345, 100.2341]
+   * faker.color.lab({ format: 'css' }) // lab(29.2345% 39.3825 20.0664)
+   * faker.color.lab({ format: 'binary' }) // (8-32 bits x 3)
+   */
+  lab(options?: { format?: ColorFormat }): string | number[] {
```
ST-DDT, line 528: And here

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * @param options Options object.
+   * @param options.format Format of generated RGB color. Defaults to `decimal`.
+   *
+   * @example
+   * faker.color.lch() // [0.522345, 72.2, 56.2]
+   * faker.color.lch({ format: 'decimal' }) // [0.522345, 72.2, 56.2]
+   * faker.color.lch({ format: 'css' }) // lch(52.2345% 72.2 56.2)
+   * faker.color.lch({ format: 'binary' }) // (8-32 bits x 3)
+   */
+  lch(options?: { format?: ColorFormat }): string | number[] {
```
ST-DDT, line 594: And here

src/modules/color/index.ts
```diff
@@ -0,0 +1,674 @@
...
+   * @example
+   * faker.color.colorByCSSColorSpace() // [0.93, 1, 0.82]
+   * faker.color.colorByCSSColorSpace({ format: 'decimal' }) // [0.12, 0.21, 0.31]
+   * faker.color.colorByCSSColorSpace({ format: 'css', space: 'display-p3' }) // color(display-p3 0.12 1 0.23)
+   * faker.color.colorByCSSColorSpace({ format: 'binary' }) // (8-32 bits x 3)
+   */
+  colorByCSSColorSpace(options?: {
+    format?: ColorFormat;
+    space?: CSSSpace;
+  }): string | number[] {
```
ST-DDT, line 660: And here

ST-DDT (changes requested):
Please copy the impl signature so it will be picked up as the last api signature.
AFAICT you don't have to add the jsdocs to the impl signature.

Shinigami92 (dismissed):

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+    casing?: Casing;
+    format?: 'hex' | ColorFormat;
+    includeAlpha?: boolean;
+  }): string | number[];
+  /**
+   * Returns an RGB color.
+   *
+   * @param options Options object.
+   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `0x`.
+   * @param options.casing Letter type case of the generated hex color. Only applied when 'hex' format is used. Defaults to `mixed`.
```
ST-DDT, line 297:
```suggestion
   * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'mixed'`.
```

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+   * @param options.format Format of generated RGB color. Defaults to `hex`.
```
ST-DDT, line 298:
```suggestion
   * @param options.format Format of generated RGB color. Defaults to `'hex'`.
```

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+export type ColorFormat = StringColorFormat | NumberColorFormat;
+export type Casing = 'lower' | 'upper' | 'mixed';
+
+/**
+ * Formats the hex format of a generated color string according
+ * to options specified by user.
+ *
+ * @param hexColor Hex color string to be formatted.
+ * @param options Options object.
+ * @param options.prefix Prefix of the generated hex color. Defaults to `0x`.
```
ST-DDT, line 43:
```suggestion
 * @param options.prefix Prefix of the generated hex color. Defaults to `'0x'`.
```

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+ * @param options.casing Letter type case of the generated hex color. Defaults to `mixed`.
```
ST-DDT, line 44:
```suggestion
 * @param options.casing Letter type case of the generated hex color. Defaults to `'mixed'`.
```

ST-DDT (changes requested):

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+    casing?: Casing;
+    format?: 'hex' | StringColorFormat;
+    includeAlpha?: boolean;
+  }): string;
+  /**
+   * Returns an RGB color.
+   *
+   * @param options Options object.
+   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `'0x'`.
+   * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'mixed'`.
```
ST-DDT, line 246: These will never be used in this case, so should they still be included in the signature/jsdocs here?

src/modules/color/index.ts
```diff
@@ -0,0 +1,793 @@
...
+   * faker.color.rgb({ casing: 'upper' }) // '0xFFFFFF'
+   * faker.color.rgb({ casing: 'lower' }) // '0xffffff'
+   * faker.color.rgb({ prefix: '#', casing: 'lower' }) // '#ffffff'
+   * faker.color.rgb({ format: 'hex', casing: 'lower' }) // '#ffffff'
+   * faker.color.rgb({ format: 'decimal' }) // [255, 255, 255]
+   * faker.color.rgb({ format: 'css' }) // 'rgb(255, 0, 0)'
+   * faker.color.rgb({ format: 'binary' }) // '10000000 00000000 11111111'
+   * faker.color.rgb({ format: 'decimal', includeAlpha: true }) // [255, 255, 255, 0.4]
+   */
+  rgb(options?: {
```
ST-DDT, line 313: AFAIK implementation signatures don't need jsdocs.

src/modules/color/index.ts
Shinigami92: @ST-DDT please check if my update was what you had thought

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #808, opened 2022-04-08
feat: Updated mimeType extensions

prisis:
feat: Added mime-db as dev dep  
feat: Added script to sync mime-db  
  
Im proposing a different approach to update the mime-types


package.json
```diff
@@ -122,7 +123,8 @@
     "validator": "~13.7.0",
     "vite": "~2.9.1",
     "vitepress": "~0.22.3",
-    "vitest": "~0.8.4"
+    "vitest": "~0.8.4",
+    "mime-db": "^1.52.0"
```
Shinigami92, line 127: Please use 
```suggestion
    "mime-db": "~1.52.0"
```

After that you need to clean and re-install to generate a new lock file

Also you did not use `pnpm add -D mime-db` because it's not on the correct sorted position

pnpm-lock.yaml
```diff
@@ -21,6 +21,7 @@ specifiers:
   eslint-plugin-prettier: ~4.0.0
   esno: ~0.14.1
   lint-staged: ~12.3.7
+  mime-db: ^1.52.0
```
Shinigami92, line 24:
```suggestion
  mime-db: ~1.52.0
```

Shinigami92 (changes requested):

package.json
prisis: Nope i didnt call it, i like to open a new pr where i like to call, npx sort-package-json

prisis: was faster for me, its fixed :)

Shinigami92 (dismissed):

src/locales/en/system/mimeTypes.ts
```diff
@@ -1,30 +1,30 @@
-/*
-
-The MIT License (MIT)
+// This file is generated by scripts/copyMimeTypes.ts
+// Do not edit this file directly. Instead, update mime-db and run pnpm run copy:mime-types
```
ST-DDT, line 2:
```suggestion
// Do not edit this file directly. Instead, update mime-db and run `pnpm run copy:mime-types`
```

scripts/copyMimeTypes.ts
```diff
@@ -0,0 +1,39 @@
...
+  'src/locales/en/system/mimeTypes.ts'
+);
+const prettierTsOptions: Options = { ...options, parser: 'typescript' };
+fs.readFile(mimeDbPath, 'utf8', (err, data) => {
+  if (err) {
+    throw err;
+  }
+
+  const licence = fs.readFileSync(mimeDbLicencePath, { encoding: 'utf8' });
+  const mimeTypeFileContent = `// This file is generated by scripts/copyMimeTypes.ts\n// Do not edit this file directly. Instead, update mime-db and run pnpm run copy:mime-types\n\n/*\n${
```
ST-DDT, line 24:
```suggestion
  const mimeTypeFileContent = `// This file is generated by scripts/copyMimeTypes.ts\n// Do not edit this file directly. Instead, update mime-db and run \`pnpm run copy:mime-types\`\n\n/*\n${
```

ST-DDT (changes requested):

Shinigami92 (approved):

ST-DDT (approved):

HonzaMac:
Thank you. Great idea.

pkuczynski:
@prisis just out of the interest: why do we copy it to our source instead of just keeping it as dependency and importing like any other module?

prisis:
It's the first step, i would like to discuss how we will handle dependencies in the future for faker.

For now I just want to provide a way without adding something new :)

PR #807, opened 2022-04-08
chore: fix todo comments

Shinigami92:

PR #810, opened 2022-04-08
refactor: improve unique

Shinigami92:
bumped `maxTime` from 3 to 50 ms because there was a flaky test in https://github.com/faker-js/faker/pull/807


src/unique.ts
```diff
@@ -31,38 +36,41 @@ export class Unique {
...
-    if (typeof opts.maxTime !== 'number') {
-      opts.maxTime = this.maxTime;
-    }
-    if (typeof opts.maxRetries !== 'number') {
-      opts.maxRetries = this.maxRetries;
-    }
-    opts.currentIterations = 0;
-    return uniqueExec.exec(method, args, opts);
+    const {
+      startTime = new Date().getTime(),
```
ST-DDT, line 63: AFAICT: This value will no longer be overwritten. Remove the line here and add it below

src/unique.ts
```diff
@@ -31,38 +36,41 @@ export class Unique {
...
    * @param args The arguments used to call the method.
-   * @param opts The optional options used to configure this method.
-   * @param opts.startTime The time this execution stared. This will be ignored/overwritten.
-   * @param opts.maxTime The time this method may take before throwing an error.
-   * @param opts.maxRetries The total number of attempts to try before throwing an error.
-   * @param opts.currentIterations The current attempt. This will be ignored/overwritten.
-   * @param opts.exclude The value or values that should be excluded/skipped.
-   * @param opts.compare The function used to determine whether a value was already returned.
+   * @param options The optional options used to configure this method.
+   * @param options.startTime The time this execution stared. This will be ignored/overwritten. Defaults to `new Date().getTime()`.
```
ST-DDT, line 40: Since the value will be overwritten, the default does not matter. In fact you could change it to "this parameter does nothing".

src/utils/unique.ts
```diff
@@ -2,21 +2,31 @@ import { FakerError } from '../errors/faker-error';
...
-const found: Record<RecordKey, RecordKey> = {};
-
-// global exclude list of results
-// defaults to nothing excluded
-const exclude: RecordKey[] = [];
-
-// current iteration or retries of unique.exec ( current loop depth )
+/**
+ * Global store of unique values.
+ * Uniqueness for entire faker instance.
```
ST-DDT, line 7: This is wrong. It is global, and not bound to a single faker instance.

src/utils/unique.ts
```diff
@@ -2,21 +2,31 @@ import { FakerError } from '../errors/faker-error';
...
+ * This means that faker should *never* return duplicate values across all API methods when using `Faker.unique` without passing `options.store`.
```
ST-DDT, line 8: There is no `options.store` yet.

ST-DDT (changes requested):

ST-DDT (approved):

import-brain (approved):

Issue #778, opened 2022-04-05
Cannot directly import _Date


JontyMC:
### Describe the bug

After importing the _Date class like this:

`import { _Date } from '@faker-js/faker/date';`

When running, I get the following error:

`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './date' is not defined by "exports" in ******\node_modules\@faker-js\faker\package.json`

### Reproduction

`import { _Date } from '@faker-js/faker/date';`

### Additional Info

I want to import _Date because I want to use it in a delegate for configuring a test data builder:

`date?: (date: _Date) => Date;`

Shinigami92:
Yeah, we could export it, but we should then name it like `FakerDate` or something like this
Currently it is an internal not exported type

JontyMC:
Thanks for workaround.

The type is in the public API, so it is really external. What would be the downside in exporting it?

ST-DDT:
> I want to import _Date because I want to use it in a delegate for configuring a test data builder:
> 
> `date?: (date: _Date) => Date;`

You should probably use the `Faker` instance there instead. Since you cannot create the `_Date` type independently.
Is there a specific reason, why you want to use the `_Date` instance instead of the `Faker` instance?

JontyMC:
Using the _Date instance creates a nicer API, as you would ony ever want to use that module in this case.

ST-DDT:
I like both the `FakerDate` and `DateModule` names. Any preference?

But I'm unsure whether these names are really easier to use then `Faker['date']` and how often they will be used.
I assume most of the time Faker will be used like this `x: (faker: Faker) => X`

JontyMC:
I don't see a downside to exporting the types (probably prefer FakerDate), it might stop someone raising the same issue. Workarounds are fine now I know them.

JontyMC:
The faker.date API is externally facing and has a contract in terms of the functions and properties it provides. I assume these won't change without deprecation and version bump, so in that sense the "type" is officially supported. Given this contract has a type already, it may as well be exported (perhaps it would be better as a typescript type or interface).

Also, the typescript _Date class is exported in the dist folder and visual studio auto imports @faker-js/faker/date when I used it.

ST-DDT:
Note: The import will probably stop working in v6.2 because we intend to move the modules from `src/<module>.ts` to `src/modules/<module>.ts`

Issue #751, opened 2022-04-01
Reorganize src folder structure


Shinigami92:
Move all modules into a `src/modules` folder

As this will mostly destroy work from active open PRs, we should do this when we have a low amount of open PRs
As I also don't know which side effects this could potentially cause, I would like to earliest target this for v7, also so we can currently focus on more important stuff right now

Shinigami92:
I highly recommend to follow a file structure like 

```
- src/
  - modules/
    - address/
      - index.ts
      - module.ts
      - definition.ts
      - <...>.ts
    - /animal
      - ...
    - index.ts
  - index.ts
```

That way we do not need to open multiple path and search for loosely-coupled files and as a VSCode user I could benefit from file-nesting

ST-DDT:
What about the actual locales data?
The current structure is based on types.
If you structure them based on usage you have to do it consistently.

Shinigami92:
> What about the actual locales data? The current structure is based on types (the definition files describe the folder structure for locales). If you structure them based on usage you have to do it consistently.

Yeah, I would like to move the content inside locales completely out of the `src` code and define them via static files that are accessible like assets and can be accessed from `dist/esm` and `dist/cjs` without duplicating them. So maybe bundled minified in `dist/assets` or `dist/locales`.
But I see this change after we are done with other stuff.
Also I think we should not move them before that into my suggested `src/modules/*` because then I assume stuff inside https://github.com/faker-js/faker/blob/main/scripts/generateLocales.ts would break or at least will be harder to do.

PR #816, opened 2022-04-08
chore: add Module suffix to module classes

Shinigami92:
Referencing: #778 and #751

- Re-structure / moving of modules will be done later
- Exporting modules as type can be done in a separate PR to preserve commit message history / changelog


ST-DDT (changes requested):
This breaks our docs generation.

**New**

grafik

All class modules are missing.

ST-DDT (approved):

pkuczynski (changes requested):
Hmm, what's the point of this change? I see no value in calling every class a module and it is confusing with es6 modules...

Shinigami92:
> Hmm, what's the point of this change? I see no value in calling every class a module and it is confusing with es6 modules...

It hasn't anything to do with es6
We already named them modules everywhere, in docs, in discussions, when we are talking about them, in unit tests, etc...

We have an issue requesting for exporting the `_Date` module as type #778 and we discussed there to first rename all the modules with that prefix to be
1. consistent
2. prevent conflicts with e.g. js `Date`
3. We decided against to name them e.g. `FakerDate`, because this could come into confusion with that it e.g. wrappes the `Date` which is not the case.

ST-DDT:
I have to agree with @Shinigami92 here.
In my mind, these are our modules.
Unless you have a better suggestion on how to name these, I'm in favor of merging this.

ST-DDT (approved):

ST-DDT:
IMO there is no need to explicitly call the modules `XyzModule`.
How about naming the modules `XyzFaker`. That way its obvious, that it's a faker module and not a datatype.

e.g.
- DateFaker
- NumberFaker
- StringFaker
- FinanceFaker
- MusicFaker
- VehicleFaker
- ....

Later on, if we actually manage to tree-shake them, then its probably easier to use a `DateFaker` than a `DateModule`.

pkuczynski:
> IMO there is no need to explicitly call the modules `XyzModule`. How about naming the modules `XyzFaker`. That way its obvious, that it's a faker module and not a datatype.

This makes much more sense to me and less chance to conflict with some other package also exporting `DateModule`. The only difference to @ST-DDT proposal, I would rather call them `FakerDate`, then `DateFaker`.

Saying all of that I am stil wondering if use case presented in #778 is a valid one. We use classes construct for namespacing only. They don't create real objects in a sense, that there is no `new FakerDate()`. @JontyMC should be rather passing whole faker instance to his _test data builder_, which would be more syntactically correct, but also more readable:

```
date?: (faker: Faker) => Date;
```

Shinigami92:
Since I totally disagree with the suggestions, I'm closing this PR from my side as I don't want to be associated with the commit history. You are free to make the changes yourself.  I've given several reasons why it's a bad idea to name it something else.

Shinigami92:
So just to summarise up why I only see that it can be named `*Module`:

- All of [#816 pr-description](https://github.com/faker-js/faker/pull/816#issue-1197468027)
- All of [#816 comment(1094238003)](https://github.com/faker-js/faker/pull/816#issuecomment-1094238003)
- The fact that we are calling them this way when talking to each other everywhere
- The fact that we have currently 50x the naming `module` in JSDoc comments and even variable names already in our codebase (`src/`)
- The fact that a user can be confused about that e.g. `FakerDate` is a wrapper around `Date`.
- `DateFaker` is not really well, because I would assume that if I have e.g. `(faker: Faker) => ...` I can call it with `(faker.date)`, which is NOT the case! It is not an abstraction or interface but just a module inside Faker.
- I even want to move code out of `FakeModule` (_fake not faker!_), because I would like to prevent "naming collision" while using autocomplete, as same as with `database` vs `datatype`. For more see: https://github.com/faker-js/faker/discussions/805

And most hilariously, the issue author could just use `(dateModule: typeof faker.date) => ...` or `(dateModule: Faker['date'])` and archive exactly what they want, today.

JontyMC:
I changed the design of our test data builders after the previous conversation because I realised it was over complicating things, so we no longer need to explicitly import the Date module.

However, I still think the API of faker would benefit from explicit typings. I outlined this before, but the Date module is part of the external facing API. There is no way to hide this. Eg, if you right click > goto definition on faker.date in vs code, it will take you to that type (currently _Date). It will also happily import the type for you and blow up at runtime, which is really why I raised the bug in the first place to help others who might do the same thing (probably unlikely...).

I didn't know you could do either of typeof faker.date or Faker['date'], so I appreciate the learning. However, all this is actually doing here is aliasing the underlying _Date class.

IMO, given it is impossible to hide the type of the modules, it would be nicer for documentation purposes if it were a type or interface and named consistently. Whether this is a class or whatever behinds the scenes is then actually an implementation detail. And for what its worth, I think xxxFaker is the least bad naming convention. You can still call them modules, just the xxxFaker module.

PR #819, opened 2022-04-08
feat: allow banned as string

Shinigami92:
Following PRs need to be merged before this PR:

- #790
- #797 

---

I would love to even type the non-array version, but sadly we still have not RegexTypes (https://github.com/microsoft/TypeScript/issues/41160) in TS and we are limited (https://stackoverflow.com/questions/66294091/how-to-create-a-type-that-describes-string-with-only-digits)

### DIFF
--- a/src/modules/random/index.ts
+++ b/src/modules/random/index.ts
@@ -1,12 +1,84 @@
 import type { Faker } from '../..';
 import { FakerError } from '../../errors/faker-error';
 import { deprecated } from '../../internal/deprecated';
+import type { LiteralUnion } from '../../utils/types';
 
 export type Casing = 'upper' | 'lower' | 'mixed';
 
 const UPPER_CHARS: readonly string[] = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
 const LOWER_CHARS: readonly string[] = 'abcdefghijklmnopqrstuvwxyz'.split('');
 const DIGIT_CHARS: readonly string[] = '0123456789'.split('');
+
+export type LowerAlphaChar =
+  | 'a'
+  | 'b'
+  | 'c'
+  | 'd'
+  | 'e'
+  | 'f'
+  | 'g'
+  | 'h'
+  | 'i'
+  | 'j'
+  | 'k'
+  | 'l'
+  | 'm'
+  | 'n'
+  | 'o'
+  | 'p'
+  | 'q'
+  | 'r'
+  | 's'
+  | 't'
+  | 'u'
+  | 'v'
+  | 'w'
+  | 'x'
+  | 'y'
+  | 'z';
+
+export type UpperAlphaChar =
+  | 'A'
+  | 'B'
+  | 'C'
+  | 'D'
+  | 'E'
+  | 'F'
+  | 'G'
+  | 'H'
+  | 'I'
+  | 'J'
+  | 'K'
+  | 'L'
+  | 'M'
+  | 'N'
+  | 'O'
+  | 'P'
+  | 'Q'
+  | 'R'
+  | 'S'
+  | 'T'
+  | 'U'
+  | 'V'
+  | 'W'
+  | 'X'
+  | 'Y'
+  | 'Z';
+
+export type NumericChar =
+  | '0'
+  | '1'
+  | '2'
+  | '3'
+  | '4'
+  | '5'
+  | '6'
+  | '7'
+  | '8'
+  | '9';
+
+export type AlphaChar = LowerAlphaChar | UpperAlphaChar;
+export type AlphaNumericChar = AlphaChar | NumericChar;
 
 /**
  * Method to reduce array of characters.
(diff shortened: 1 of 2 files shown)


src/random.ts
```diff
@@ -443,8 +513,13 @@ export class Random {
       'y',
       'z',
     ];
-
-    charsArray = arrayRemove(charsArray, bannedChars);
+    // TODO @Shinigami92 2022-01-11: A default empty array gets assigned above, we should check the length against 0 or not here
```
ST-DDT, line 516: There is no need for that IMO.

import-brain (approved):

src/random.ts
```diff
@@ -511,7 +588,12 @@ export class Random {
       'z',
     ];
 
-    charsArray = arrayRemove(charsArray, bannedChars);
+    if (bannedChars) {
+      if (typeof bannedChars === 'string') {
+        bannedChars = bannedChars.split('');
+      }
```
ST-DDT, line 594: Please make all 3? instances consistent where and how you do this check.

ST-DDT (changes requested):

ST-DDT (dismissed):

xDivisionByZerox:
What would be the use case of providing a string instead of an array? This would allow me to provide an option like `bannedChars: 'aaaaaaaaaaaaaaaaaaaaaaaaaaA'` where I would expect only this combination of charters to be baned.

Shinigami92:
> What would be the use case of providing a string instead of an array? This would allow me to provide an option like `bannedChars: 'aaaaaaaaaaaaaaaaaaaaaaaaaaA'` where I would expect only this combination of charters to be baned.

Yes exactly, it's the same as `bannedChars: 'aaaaaaaaaaaaaaaaaaaaaaaaaaA'.split('')`, just a better UX

Shinigami92:
@xDivisionByZerox Do I need to update the docs? Or maybe ask @pkuczynski what the initial thought was by his feature request.

ST-DDT:
IMO having a string is a simplification over the array (you have to write less). Having an object or indizies would require more writing/reduce readability.

I would adjust the jsdocs though (param and example).

Shinigami92:
First I would love to let @pkuczynski come to word :slightly_smiling_face: 
Then I can work further on this PR

pkuczynski:
@Shinigami92 can you remind me when did I request this? :) I tried to click around for a while and dust off my memory, but don't recon this request? :(

pkuczynski:
And btw. with `bannedChars` option I would expect this chars will not appear in the result string. Chars, not whole string. So in the example from @xDivisionByZerox 

```
// this is fine
const result = faker.datatype.string({
    bannedChars: 'abcd',
    count: 4,
});
console.log(result); // abcz
```

I would expect that result would not contain neither `a`, `b`, `c` or `d`.

Shinigami92:
> @Shinigami92 can you remind me when did I request this? :) I tried to click around for a while and dust off my memory, but don't recon this request? :(

https://github.com/faker-js/faker/pull/797#discussion_r846391378

ST-DDT (changes requested):
This fails the following tests:

````ts
it('should be able to ban some characters via string', () => {
  const actual = faker.random.alpha({
    count: 5,
    bannedChars: [
      'B',
      'C',
      'D',
      'E',
      'F',
      'G',
      'H',
      'I',
      'J',
      'K',
      'L',
      'M',
      'N',
      'O',
      'P',
      'Q',
      'R',
      'S',
      'T',
      'U',
      'V',
      'W',
      'X',
      'Y',
      'Z',
    ],
    upcase: true,
  });

  expect(actual).toHaveLength(5);
  expect(actual).toMatch(/^A{5}$/);
});
````

I know this isn't directly caused by this PR, but this PR explicitly lists uppercase characters as allowed (to be banned).

Shinigami92:
@ST-DDT You mean we need to add them to line 248 to 275?

ST-DDT:
I would like have a `case: upper|lower|mixed` parameter (mixed being the default).

Shinigami92:
> I would like have a `case: upper|lower|mixed` parameter (mixed being the default).

okay, then this will be definitely another PR/issue

ST-DDT:
> > I would like have a `case: upper|lower|mixed` parameter (mixed being the default).
> 
> okay, then this will be definitely another PR/issue

I consider, this "fixed", if you remove the uppercase characters from the input (for now).

ST-DDT:
(and edit the jsdocs accordingly)

ST-DDT:
Could you add a hint to the jsdocs regarding that behavior?

Shinigami92:
Await merge of #955

pkuczynski:
> > @Shinigami92 can you remind me when did I request this? :) I tried to click around for a while and dust off my memory, but don't recon this request? :(
> 
> [#797 (comment)](https://github.com/faker-js/faker/pull/797#discussion_r846391378)

Ah yes, sorry! :) Indeed I requested this :) Then I think I had only this comment https://github.com/faker-js/faker/pull/819#issuecomment-1117956389 which seems to be addressed?

pkuczynski (approved):

ST-DDT (approved):

pkuczynski (approved):

PR #822, opened 2022-04-09
feat: dynamic definitions tree

ST-DDT:
Instead of making only our definitions/locale data accessible via `faker.definitions`, this PR allows devs to access any data, that are added to the locale itself.

Or in other words:
Previously `faker.definitions.business` returned `undefined` even though `faker.locales.en.buisness` is defined.


ST-DDT:
Partially addresses #823 .

src/faker.ts
```diff
@@ -98,43 +98,56 @@ export class Faker {
...
-          get: (): unknown =>
-            this.locales[this.locale][moduleName]?.[entryName] ??
-            this.locales[this.localeFallback][moduleName]?.[entryName],
-        });
-      }
-    }
+    };
+
+    // Returns the first LocaleDefinition[module][entry] in any locale
+    const resolveModuleData = (
```
Shinigami92, line 130: I would prefer to swap these `resolveModuleData` lines with `moduleLoader` lines cause in line 120 `resolveModuleData` is used but `moduleLoader` is not used in `resolveModuleData`

src/faker.ts
```diff
@@ -48,7 +48,7 @@ export class Faker {
   localeFallback: UsableLocale;
 
   // Will be lazy init
-  readonly definitions: LocaleDefinition = {} as LocaleDefinition;
+  readonly definitions: LocaleDefinition = this.initDefinitions();
```
Shinigami92, line 51: Looks like it's not lazy init anymore, so I would just remove the comment

Shinigami92 (commented):
To really understand what's going on in `initDefinitions` I would like to have a screenshare after my vacation with you, where I go through the code and you explain what I don't understand :slightly_smiling_face:

src/faker.ts
ST-DDT: Fixed

src/faker.ts
ST-DDT: Fixed

Shinigami92 (approved):
Tested it out locally with the new tests but without the changes in `src`
To me this is not a feature but a bugfix!

prisis (approved):

PR #828, opened 2022-04-10
refactor: log messages for unique fields

Shinigami92:
This is a addendum to #810

### DIFF
--- a/docs/.vitepress/tsconfig.json
+++ b/docs/.vitepress/tsconfig.json
@@ -1,0 +1,5 @@
+{
+  "compilerOptions": {
+    "target": "ES2020"
+  }
+}
--- a/src/unique.ts
+++ b/src/unique.ts
@@ -1,3 +1,4 @@
+import { deprecated } from './internal/deprecated';
 import type { RecordKey } from './utils/unique';
 import * as uniqueExec from './utils/unique';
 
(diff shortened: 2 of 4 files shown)


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "target": "ES5",
```
Shinigami92, line 3: I needed to set this explicit, so accessors are possible

docs/.vitepress/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "compilerOptions": {
+    "target": "ES2020"
```
Shinigami92, line 3: Somehow vitepress use target as build target and cannot handle const without that

ST-DDT (approved):
Does this need tests?

pkuczynski (approved):

src/unique.ts
```diff
@@ -10,14 +11,74 @@ export class Unique {
...
+
+  /**
+   * Maximum time `unique.exec` will attempt to run before aborting.
+   *
+   * @deprecated Use options instead.
+   */
+  get maxTime(): number {
+    deprecated({
+      deprecated: 'faker.unique.maxTime',
+      proposed: 'Options',
```
pkuczynski, line 24: I am not sure if this is clear where those Options should be applied?

pkuczynski (approved):

src/unique.ts
ST-DDT: Yeah, you are right. @Shinigami92 Please improve this.

Shinigami92: IMO it's okay, because there is only exactly one function that is related in the context with that.

PR #829, opened 2022-04-10
feat: phone IMEI

### LINKED ISSUE #826: Random IMEI number

I want to be able to generate a random IMEI number.

### Suggested solution

Would this be under ```phone``` or ```internet```? I think it could go under either of them, honestly.

### Alternative

_No response_

### Additional context

IMEI numbers follow the same Luhn algorithm that is used for credit cards, the ```luhnCheck.ts``` file can be used to check for IMEI validity. Also, I'm pretty sure ```helpers.replaceCreditCardSymbols()``` could be adapted for this usage.

import-brain:
closes #826

### DIFF
--- a/src/phone.ts
+++ b/src/phone.ts
@@ -1,7 +1,7 @@
 import type { Faker } from '.';
 
 /**
- * Module to generate phone-related numbers.
+ * Module to generate phone-related data.
  */
 export class Phone {
   constructor(private readonly faker: Faker) {
@@ -62,7 +62,7 @@ export class Phone {
   }
 
   /**
-   * Returns a random IMEI number.
+   * Generates IMEI number.
    *
    * @example
    * faker.phone.imei() // '13-850175-913761-7'
--- a/src/phone.ts
+++ b/src/phone.ts
@@ -1,7 +1,7 @@
 import type { Faker } from '.';
 
 /**
- * Module to generate phone numbers.
+ * Module to generate phone-related numbers.
  */
 export class Phone {
   constructor(private readonly faker: Faker) {
@@ -58,6 +58,19 @@ export class Phone {
   phoneFormats(): string {
     return this.faker.random.arrayElement(
       this.faker.definitions.phone_number.formats
+    );
+  }
+
+  /**
+   * Returns a random IMEI number.
+   *
+   * @example
+   * faker.phone.imei() // '13-850175-913761-7'
+   */
+  imei(): string {
+    return this.faker.helpers.replaceCreditCardSymbols(
+      '##-######-######-L',
+      '#'
     );
   }
 }
--- a/test/phone.spec.ts
+++ b/test/phone.spec.ts
@@ -153,7 +153,8 @@ describe('phone', () => {
         });
 
         it('should be Luhn-valid', () => {
-          expect(luhnCheck(faker.phone.imei())).toBeTruthy();
+          const imei = faker.phone.imei();
+          expect(imei).satisfy(luhnCheck);
         });
       });
     }
--- a/test/phone.spec.ts
+++ b/test/phone.spec.ts
@@ -147,6 +147,11 @@ describe('phone', () => {
           expect(imei).toBeTypeOf('string');
         });
 
+        it('should have a length of 18', () => {
+          const imei = faker.phone.imei();
+          expect(imei).toHaveLength(18);
+        });
+
         it('should be Luhn-valid', () => {
           expect(luhnCheck(faker.phone.imei())).toBeTruthy();
         });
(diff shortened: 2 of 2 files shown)


src/phone.ts
```diff
@@ -60,4 +60,17 @@ export class Phone {
...
+   * Returns a random IMEI number.
+   *
+   * @example
+   * faker.phone.imei() // '13-850175-913761-7'
+   */
+  imei(): string {
+    return this.faker.helpers.replaceCreditCardSymbols(
+      '##-######-######-L',
+      '#'
+    );
```
import-brain, line 74: Should we create a new function and use that instead of ```replaceCreditCardSymbols```? As of right now, this works and is Luhn-valid.

src/phone.ts
```diff
@@ -60,4 +60,17 @@ export class Phone {
       this.faker.definitions.phone_number.formats
     );
   }
+
+  /**
+   * Returns a random IMEI number.
```
pkuczynski, line 65:
```suggestion
   * Generates IMEI number.
```

In faker everything is random :)

src/phone.ts
```diff
@@ -1,7 +1,7 @@
 import type { Faker } from '.';
 
 /**
- * Module to generate phone numbers.
+ * Module to generate phone-related numbers.
```
pkuczynski, line 4:
```suggestion
 * Module to generate phone-related data.
```

?

src/phone.ts
pkuczynski: Or rename the helper to give it more broad meaning?

pkuczynski (dismissed):

test/phone.spec.ts
```diff
@@ -125,6 +140,16 @@ describe('phone', () => {
           expect(faker.definitions.phone_number.formats).contain(phoneFormat);
         });
       });
+
+      describe('imei()', () => {
+        it('should return a string', () => {
+          const imei = faker.phone.imei();
+          expect(imei).toBeTypeOf('string');
+        });
+        it('should be Luhn-valid', () => {
```
Shinigami92, line 149:
```suggestion
        });

        it('should be Luhn-valid', () => {
```

src/phone.ts
import-brain: No-coffee documentation 🙃

src/phone.ts
ST-DDT: Since I don't have a better name right now, I think it's ok to start like this.

Shinigami92 (approved):

test/phone.spec.ts
```diff
@@ -125,6 +140,17 @@ describe('phone', () => {
           expect(faker.definitions.phone_number.formats).contain(phoneFormat);
         });
       });
+
+      describe('imei()', () => {
+        it('should return a string', () => {
+          const imei = faker.phone.imei();
+          expect(imei).toBeTypeOf('string');
```
ST-DDT, line 147: Please also test the length.

src/phone.ts
import-brain: I could make another PR renaming ```replaceCreditCardSymbols()```, change every place where it is called, and put this one on hold until that one is merged, if you want.

test/phone.spec.ts
```diff
@@ -125,6 +140,17 @@ describe('phone', () => {
...
+        });
+
+        it('should be Luhn-valid', () => {
+          expect(luhnCheck(faker.phone.imei())).toBeTruthy();
```
ST-DDT, line 151: IMO you can switch this to

````ts
const actual = ...

expect(actual).toSatisfy(luhnCheck);
````

ST-DDT (changes requested):

test/phone.spec.ts
import-brain: This throws a "Property 'toSatisfy' does not exist on type 'Assertion<string>'" error.

src/phone.ts
Shinigami92: Or the other way around, merge this as it is kinda ready and rename the internally used function in this just also.
You are currently just calling something that already works and it's not see-able from outside by a user.

test/phone.spec.ts
Shinigami92: I think it's just called `satisfy` or `satisfies`, pls look via autocomplete, we also use it in other places

import-brain: > I think it's just called `satisfy` or `satisfies`, pls look via autocomplete, we also use it in other places

Ohhh I see, it didn't show up in Vitest docs, so I was confused.

Shinigami92 (approved):

ST-DDT (approved):

PR #838, opened 2022-04-14
test: introduce seededTests factory

ST-DDT:
This PR introduces a seededTest factory that simplifies
There are probably a lot of ways in which this can be improved,
but for now I would just like to get feedback on the concept itself.
Should we do something like this or should we scrap this idea?

Current benefits of this approach:

- Simplified tests (~1 line instead of ~5 per test)
- Ensure that all methods in a module have a test
- Ensure that all methods with arguments have explicit tests (that hopefully covers all the argument combinations)

### DIFF
--- a/test/address.spec.ts
+++ b/test/address.spec.ts
@@ -71,7 +71,8 @@ describe('address', () => {
 
     t.describe('latitude', (t) => {
       t.it('noArgs');
-    }).describe('longitude', (t) => {
+    });
+    t.describe('longitude', (t) => {
       t.it('noArgs');
     });
 
--- a/test/address.spec.ts
+++ b/test/address.spec.ts
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, it } from 'vitest';
 import { faker } from '../src';
-import { seededRuns } from './support/seededRuns';
+import { seededTests } from './support/seededRuns';
 import { times } from './support/times';
 
 function degreesToRadians(degrees: number) {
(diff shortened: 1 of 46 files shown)


Shinigami92:
Could we try snapshots first?

pkuczynski:
I must say, I am not the biggest fan of those tests, as I dont feel they really test anything. Whenever they fail, I just simply replace the value with the expected one and that's it. So what's the point? I much more prefer functional tests...

ST-DDT:
> I must say, I am not the biggest fan of those tests, as I dont feel they really test anything. Whenever they fail, I just simply replace the value with the expected one and that's it. So what's the point? I much more prefer functional tests...

The most of the functional tests are pretty vague.
Our library generates random data so there is hardly anything we can test precisely.
Most tests I have written so far anywhere consist of mostly equality checks, the seeded test are our only way to do that.
The seeded test exist to assist us, and notice changes/side effects.
E.g. if arrayElement would always return the first element, the random tests would still pass, but the seeded would fail. They won't tell you why, but at least they tell you to have a look.
Asside from notifying us, they also give us the chance to look at the returned/expected values.
Does it look like an address or phone number? Are there any unresolved parts in there?

ST-DDT:
I updated this PR to use test snapshots instead of implementing that myself.
It would be nice, if you could give me some input regarding the test setup and whether it is worth continuing (using the seededTests factory).

I'm not really happy with the test snapshot file format, but I guess it's better then using our own proprietary format.

The PR's changes to the tests are probably reviewed best on a side by side view.
The changes to the snapshot file are only caused by the changed names (to help with readability of the tests).

ST-DDT:
@Shinigami92 Why is the type check failing?

ST-DDT:
Since the new factory explicitly forces tests to be created if a method has parameters, there are some new tests.

ST-DDT:
All test converted. ~400/3 new ones added.

test/support/seededRuns.ts
```diff
@@ -1 +1,315 @@
...
+   * This method will automatically be called by the default methods
+   * and should be called at the beginning of custom vite `it` blocks.
+   */
+  setup(): void {
+    this.faker.seed(this.seed);
+    this.faker.locale = 'en';
+  }
+
+  /**
+   * Runs the instructions for an vite `it` block.
```
Shinigami92, line 137:
```suggestion
   * Runs the instructions for a vitest `it` block.
```

test/support/seededRuns.ts
```diff
@@ -1 +1,315 @@
...
+  }[MethodOf<ObjectType, () => unknown>];
+
+/**
+ * Method that prepares seeded tests.
+ *
+ * It ensures that all methods in that module have exactly one test case or block associated to them.
+ * Duplicate calls to `t.it(methodName)` or `t.describe(methodName)` will directly throw an error.
+ * Before the method returns it will check that there are tests for all methods of the module.
+ *
+ * You may add custom vite `it` and `describe` blocks both on a module and method level,
```
Shinigami92, line 50:
```suggestion
 * You may add custom vitest `it` and `describe` blocks both on a module and method level,
```

test/address.spec.ts
```diff
@@ -37,133 +37,70 @@ function haversine(
...
+    });
 
-          expect(cardinalDirection).toMatchSnapshot();
-        });
-      });
+    t.describe('latitude', (t) => {
+      t.it('noArgs');
+    }).describe('longitude', (t) => {
+      t.it('noArgs');
+    });
```
Shinigami92, line 76: *nit-pick*: Is this prettier formatting? Or can we add a newline before `.desc...`?
Otherwise, could we use `t.desc...` instead to force a linebreak?

test/datatype.spec.ts
```diff
@@ -1,220 +1,89 @@
...
+          min: -42,
+          max: 69,
+        })
+        .it('with min, max and precision', {
+          min: -42,
+          max: 69,
+          precision: 0.01,
+        });
+
+      // TODO: Migrate?
```
Shinigami92, line 24: OPEN TODO (with no responsible), need to be solved before merge

Shinigami92 (changes requested):

xDivisionByZerox:
I really have to say that I'm with @pkuczynski here. Even after the [explanation](https://github.com/faker-js/faker/pull/838#issuecomment-1100046325) I don't see how this test (the current one as well) will help us in any way.

> Most of the functional (random) tests are pretty vague and have flakey test coverage at best.

So why spend time implementing other vague tests that don really test anything. Creating solid tests that check for logical flaws is much better than 100% coverage.

> Our library generates random data so there is hardly anything we can test precisely.

Most of our functions just return a random member of the locales definitions. I don't see how we can't test this. And if we do something other than that, there you have the thing you need to test.

> Most tests I have written so far anywhere consist of mostly contains or regex checks, the seeded test are our only way to do equalitiy tests.

How is a test that consists of "Oh the current value ain't correct, lemme override it. Nice now it works." better than a RegExp check? I truly don't understand this, If a case is too complex for logical checks a RegExp should be fine. That said, it should be the last option to choose as we have written in our intern coding guidelines.

> The seeded test exist to assist us, and notice changes/side effects.

I understand this and am partially on your side for this. But this seems more like an e2e test case for me. A unit test should not be affected by any side effects since the dependency in use should be mocked in that case (what we don't do at all).

 > E.g. if arrayElement would always return the first element, the random tests would still pass, but the seeded would fail. They won't tell you why, but at least they tell you to have a look.

Why would a seeded test catch that? The example implies that the current implementation of a function is changed, otherwise, it would just generate a snapshot with the first element from the array as expectation. In that case (implementation is changed) the seeded test would fail anyway since it's very likely that the return value differs from the original implementation. I think that's what you wanted to say with "They won't tell you why, [...]", but I think that is a very weak point since any of these cases can be implemented by a concrete, random test (that checks for logic errors) instead of a seeded, snapshot test.

> Asside from notifying us, they also give us the chance to look at the returned/expected values.
Does it look like an address or phone number? Are there any unresolved parts in there?

I mean, yes, obviously. But is this the point of the unit test here? If I take the claim from before that most values are checked against RegExp, I can be very sure what my final result will look like as well.

> They are not intended to be just replaced without a thought.

If they are, I think this is very poorly executed. No one will have a deep look into a completely changed snapshot file, since I expect this things to pass against the current test suites.

Shinigami92 (dismissed):

ST-DDT:
The codecov/patch think is still failing (because I only add types to src), but this is ready to review.

Shinigami92 (approved):

Shinigami92 (approved):

prisis (approved):

PR #839, opened 2022-04-15
docs: fix mersenne rand min max confusion

### LINKED ISSUE #775: Small issue with mersenne jsdoc
### Describe the bug

https://fakerjs.dev/api/mersenne.html#rand

In the examples, it says ```faker.mersenne.rand(500, 1000)```. However, the ```max``` parameter is listed first. I saw the code on how we got around listing ```max``` first (L33 to L37 of mersenne.ts), but is there any possible way we could improve the mersenne jsdoc examples? Could be a bit confusing, considering ```max``` is shown first in the function signature.

### Reproduction

faker.mersenne.rand(500, 1000)

### Additional Info

_No response_

gregoriomario:
although mersenne automatically swap the value if min is bigger than max, but it's a good practice to clear out the confusion.

Fix #775

### DIFF
--- a/src/modules/mersenne/index.ts
+++ b/src/modules/mersenne/index.ts
@@ -20,14 +20,14 @@ export class Mersenne {
   }
 
   /**
-   * Generates a random number between `[min, max)`.
+   * Generates a random number between `[max, min)`.
    *
    * @param max The maximum number. Defaults to `0`.
    * @param min The minimum number. Defaults to `32768`.
    *
    * @example
    * faker.mersenne.rand() // 15515
-   * faker.mersenne.rand(500, 1000) // 578
+   * faker.mersenne.rand(1000, 500) // 578
    */
   rand(max = 32768, min = 0): number {
     if (min > max) {
--- a/src/modules/mersenne/index.ts
+++ b/src/modules/mersenne/index.ts
@@ -20,10 +20,10 @@ export class Mersenne {
   }
 
   /**
-   * Generates a random number between `[max, min)`.
+   * Generates a random number between `[min, max)`.
    *
-   * @param max The maximum number. Defaults to `0`.
-   * @param min The minimum number. Defaults to `32768`.
+   * @param max The maximum number. Defaults to `32768`.
+   * @param min The minimum number. Defaults to `0`.
    *
    * @example
    * faker.mersenne.rand() // 15515


src/mersenne.ts
```diff
@@ -20,14 +20,14 @@ export class Mersenne {
   }
 
   /**
-   * Generates a random number between `[min, max)`.
+   * Generates a random number between `[max, min)`.
```
import-brain, line 23:
```suggestion
   * Generates a random number between `[min, max)`.
```
I think this should be changed back to what it was before (as seen in the suggestion block). The original description said that the function generates a number with an inclusive lower bound of ```min``` and an exclusive upper bound of ```max```, so this change will change the meaning of the JSDoc. Also, we should probably discuss putting a "max is first param" text in the JSDoc, but I'm not sure what format that would take.

import-brain (changes requested):

src/mersenne.ts
gregoriomario: I've been thinking about this too, which one should we change to clear out the confusion. Is it the docs that's inconsistent with the implemenetation of the param, or is it the param which might be better if we put min first instead of max. 

Maybe it's better to fix the param and keep the docs as it is before, but I'm not sure about this. Wdyt? 🤔

import-brain: I think fixing the param itself would be a breaking API change, so we should just change the doc. @Shinigami92 @ST-DDT @pkuczynski What do you think?

pkuczynski: Yes, this would be a breaking change so quite problematic...

ST-DDT: For now, please only the jsdocs please.

src/mersenne.ts
```diff
@@ -20,14 +20,14 @@ export class Mersenne {
...
    *
    * @param max The maximum number. Defaults to `0`.
    * @param min The minimum number. Defaults to `32768`.
    *
    * @example
    * faker.mersenne.rand() // 15515
-   * faker.mersenne.rand(500, 1000) // 578
+   * faker.mersenne.rand(1000, 500) // 578
```
Shinigami92, line 30: Seems the "example" not wrong, should we just add another line?

src/mersenne.ts
Shinigami92, line 23: Please re-check if the math symbols are correct (in-/ex-clusive)

Shinigami92 (commented):
I assume it was implemented with max as first param, so that a user can use it with `faker.mersenne.rand(6)` so without a min argument and it just has a default min = 0

src/mersenne.ts
Shinigami92: I take somewhat back my statement

max is in first place so it can potentially used without providing manually a min argument
I personally don't like that kind of parameter usage and would like to call it with options instead of two params, but that is not related to this PR 🤷

Shinigami92 (approved):

src/mersenne.ts
```diff
@@ -20,14 +20,14 @@ export class Mersenne {
   }
 
   /**
-   * Generates a random number between `[min, max)`.
+   * Generates a random number between `[max, min)`.
    *
    * @param max The maximum number. Defaults to `0`.
    * @param min The minimum number. Defaults to `32768`.
```
Shinigami92, line 26: The default values are also wrong here 🤔 are they?

Shinigami92 (changes requested):

Shinigami92 (approved):

ST-DDT (approved):

import-brain (approved):

xDivisionByZerox (approved):

PR #847, opened 2022-04-20
build: add node v18 support

### LINKED ISSUE #42: FEAT: Support Node18 when it drops
https://nodejs.org/en/about/releases/

Node 18 is slated to release on `2022-04-19`, we need to update the ci pipeline to this on that day.

Shinigami92:
- Use node v18 as new default
- Still support v14 and v16, v16 specially as latest LTS
- ~~Remove support for node v12 completely (maybe we can do this in a separate PR if we need to long for other v6.x stuff)~~ see #850

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
   },
   "devDependencies": {
     "@types/markdown-it": "~12.2.3",
-    "@types/node": "~16.11.33",
+    "@types/node": "~17.0.31",
     "@types/prettier": "~2.6.0",
     "@types/sanitize-html": "~2.6.2",
     "@types/validator": "~13.7.2",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -90,7 +90,7 @@
   },
   "devDependencies": {
     "@types/markdown-it": "~12.2.3",
-    "@types/node": "~16.11.27",
+    "@types/node": "~17.0.25",
```
Shinigami92, line 93: We need to wait for v18 types

Shinigami92:
According to this table https://kangax.github.io/compat-table/es2016plus/#node14_0 I think we can also switch to target 2020 🥳 
_maybe in a separate PR_

package.json
Shinigami92: Work on `@types/node` has started:
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60047
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60042

ST-DDT (approved):

ST-DDT:
This probably need admin powers to be merged.

pkuczynski (approved):

PR #848, opened 2022-04-20
refactor!: target es2020

Shinigami92:
Together with #847 according to this table https://kangax.github.io/compat-table/es2016plus/#node14_0 I think we can also switch to target 2020 🥳


tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES5",
```
pkuczynski, line 3: Hmm, should we drop support for older browsers? Do you know which would be affected?

Shinigami92: Accroding to this table: https://kangax.github.io/compat-table/es2016plus/
Yes, we are really safe

ST-DDT (approved):

tsconfig.json
pkuczynski: Not that safe for older iOS or other server side VMs, but I guess thats fine for faker..

pkuczynski (approved):

PR #849, opened 2022-04-20
refactor!: get rid of export =

Shinigami92:
This PR is an requirement for #848

### DIFF
--- a/scripts/bundle.ts
+++ b/scripts/bundle.ts
@@ -33,7 +33,7 @@ buildSync({
   // splitting: true, // Doesn't work with cjs
   format: 'cjs',
   platform: 'node',
-  target: 'node12',
+  target: 'node14',
 });
 
 console.log('Building dist for node type=module (esm)...');
@@ -48,6 +48,6 @@ buildSync({
   minify: true,
   splitting: true,
   format: 'esm',
-  target: 'node12.20',
+  target: 'node14',
   outExtension: { '.js': '.mjs' },
 });
--- a/src/locale/af_ZA.ts
+++ b/src/locale/af_ZA.ts
@@ -7,7 +7,7 @@ import { Faker } from '../faker';
 import af_ZA from '../locales/af_ZA';
 import en from '../locales/en';
 
-const faker = new Faker({
+export const faker = new Faker({
   locale: 'af_ZA',
   localeFallback: 'en',
   locales: {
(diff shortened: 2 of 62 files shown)


Shinigami92:
We need to decide if we want to introduce a breaking change and users need to use `require('@faker-js/faker/locale/en').default` or if we want back somehow the current behavior without `.default`.
It looks like we are trapped in a legacy design issue with that.

ST-DDT:
I like that proposal.

Shinigami92:
> I like that alternative proposal. I assume most of the time you only need a specific language anyway.

@prisis I would like to hear your vote for that, if you also accept, I will refactor it to that

src/locale/af_ZA.ts
```diff
@@ -7,13 +7,11 @@ import { Faker } from '../faker';
 import af_ZA from '../locales/af_ZA';
 import en from '../locales/en';
 
-const faker = new Faker({
+export const faker = new Faker({
```
ST-DDT, line 10: Can you add a test for these/fix the coverage?

Shinigami92: That's an issue with `c8` cannot do anything about it from faker side
Feel free to open an issue, could even be that I still have one open there
But I know that it's covered and you can also see in the changed files that I updated the tests that coveres this part, if I would not touched the test, it would fail :shrug:

ST-DDT (dismissed):

scripts/bundle.ts
```diff
@@ -34,7 +34,7 @@ buildSync({
   // splitting: true, // Doesn't work with cjs
   format: 'cjs',
   platform: 'node',
-  target: 'node12',
+  target: 'node14',
```
pkuczynski, line 37: So we are dropping node 12 support? This should be mentioned in release notes and in README

pkuczynski (dismissed):

scripts/bundle.ts
Shinigami92: That will already be done with #850

ST-DDT (approved):

ST-DDT:
Doesn't this contain the changes from?
https://github.com/faker-js/faker/pull/850

pkuczynski (approved):

Shinigami92:
> Doesn't this contain the changes from? #850

Not anymore 😉

PR #852, opened 2022-04-20
test: use consistent assertion functions

Shinigami92:
Waiting on https://github.com/vitest-dev/vitest/pull/1172

Theoretically we would also need that `greaterThanOrEquals` behave the same as `toBeGreaterThanOrEquals`
The one use `number | bigint` and the other uses `number | Date`

### DIFF
--- a/test/commerce.spec.ts
+++ b/test/commerce.spec.ts
@@ -101,7 +101,7 @@ describe('commerce', () => {
       describe(`productName()`, () => {
         it('should return random values from product arrays', () => {
           const name = faker.commerce.productName();
-          expect(name.split(' ').length).greaterThanOrEqual(3);
+          expect(name.split(' ').length).toBeGreaterThanOrEqual(3);
 
           const parts = name.split(' ');
           expect(faker.definitions.commerce.product_name.adjective).toContain(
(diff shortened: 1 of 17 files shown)


test/commerce.spec.ts
```diff
@@ -146,7 +146,7 @@ describe('commerce', () => {
           expect(
             amount,
             'The expected match should not include a currency symbol'
-          ).match(/[0-9.]/);
+          ).toMatch(/[0-9.]/);
```
ST-DDT, line 149: Not sure whether we should fix it here or in a separate PR, but this should probably look like this:

```suggestion
          ).toMatch(/^[0-9\.]+$/);
```

test/finance.spec.ts
```diff
@@ -230,18 +235,19 @@ describe('finance', () => {
           expect(
             amount,
             'The expected match should not include a currency symbol'
-          ).match(/[0-9.]/);
+          ).toMatch(/[0-9.]/);
```
ST-DDT, line 238: Same here.

ST-DDT (dismissed):

test/commerce.spec.ts
ST-DDT: Fixed in #865

test/finance.spec.ts
ST-DDT: Fixed in #865

pkuczynski (approved):

ST-DDT (approved):

import-brain (approved):

PR #857, opened 2022-04-20
fix: dont log deprecations on startup

Shinigami92:


pkuczynski (approved):

test/faker.spec.ts
```diff
@@ -31,6 +31,15 @@ describe('faker', () => {
...
   });
 
+  it('should not log any warnings on startup', () => {
+    const spy = vi.spyOn(console, 'warn');
+
+    new Faker({ locales: { en: { title: '' } } });
+
+    expect(spy).not.toHaveBeenCalled();
+    spy.mockRestore();
+  });
```
ST-DDT, line 41: Not sure why you tested a custom locale. You should also test the default instance.
(Maybe use `require(../src).faker` to check all locale loading).

Please also check that there is no other console log calls.

ST-DDT:
AFAICT We haven't released this bug yet, so there is no need for a hotfix release.

Shinigami92:
> AFAICT We haven't released this bug yet, so there is no need for a hotfix release.

I set that to p3 cause it's blocking us from releasing a v6.2.0, but I'm okay with any prio anyways 🤷

import-brain (approved):

Issue #830, opened 2022-04-10
Load locales from static files and only bundle once


Shinigami92:
### Clear and concise description of the problem

Currently locales are bundled from the `src` code to `dist/esm` and `dist/cjs`. This doubles the needed bundle-size by 2x as much as already needed.

### Suggested solution

We should move all locales outside the `src` code/folder into its own `locales` folder and bundle them once into `dist/locales` (minified).
Also we should use another format that is static like `json` or `yaml`.
The source locale files can be in a human readable format and they just need to be converted into a best minifyable format possible to minimize the bundle-size of the package.
Then code running from `esm` or `cjs` need to access somehow the content of `dist/locales`. Maybe possible via `node:fs`?
It might also be that we want to define schema for the locales, so we can validate that no invalid data slip into the locales, checked by CI.

### Additional context

Other issues like #751 and #823 should be done before starting with this issue.
I just opened this issue so it's transparent where I would like to go in long-time-goal.

ST-DDT:
This may conflict with possible changes to lazyly loaded locale data.

Shinigami92:
> This may conflict with possible changes to lazyly loaded locale data.

Either I didn't fully understand what you are meaning by this, or I think no, due to the content of the locales would be loaded eagerly as they are loaded now.
Only how they are loaded from static files instead of import statements would change.

ST-DDT:
> > This may conflict with possible changes to lazyly loaded locale data.
> 
> Either I didn't fully understand what you are meaning by this, or I think no, due to the content of the locales would be loaded eagerly as they are loaded now. Only how they are loaded from static files instead of import statements would change.

For now, this doesn't have any immediate negative impact. But we have to keep in mind that this prevents us for loading the locale data lazily/dynamically in the future.

Shinigami92:
Theoretically this issue could be extended to load locales from anywhere via an abstraction, so it could be loaded from a file, a DB, an web-endpoint or anything else. Only the schema needs to be fulfilled or the data cannot be correctly accesses by faker.

But first I will aim for moving our existing files into static locales as planned, so the bundle size can be halved in size.

ST-DDT:
This should be at least a bit flexible to allow for constructs like these:

````ts
export first_name = [...import(female_first_name), ... import(male_first_name)].sort().unique();
````

I'm not sure but maybe it is possible to do something like this:

````ts
if (import(moduleEntry).isDynamic) {
    // generate dynamic entrypoint
} else {
[...]

Issue #823, opened 2022-04-09
Improve locale loading and global locale data


ST-DDT:
### Clear and concise description of the problem

The current system of loading locale data has some drawbacks, that needs to be addresses:

1) It heavily depends on the `en` locale, even though some of the data (emojis, iban, country codes) aren't related to that
2) Is is not possible to base region/country specific locales, with more than a single fallback (which is `en` by default)
3) Using custom locale entries is currently not supported (via `faker.definitions`)

### Suggested solution

1) Introduce a `global` locale with all the locale data that aren't bound to a specific locale
2) Replace locale and fallback locale with an locale[], that are searched in order #858 
3) Provide all locale data dynamically via (via `faker.definitions`) #822 

### Alternative

1) Remove these data entirely from the locales, making it impossible to change/filter them.
2) Require the users to manually build their locales, but this is the exact opposite of our existing `faker.definitions`
3) Let them access the locale data via the locale, which once again is the exact opposite of our existing `faker.definitions`

### Additional context

1) could be introduced in v6.x, while keeping the data in en till v7 (No 2 is implement).
2) Is definitely a breaking change and can only be in v7+
3) Can be added now, as it is a non-breaking improvement

ST-DDT:
I already have an outdated branch with a potential solution/implementation for 2: https://github.com/faker-js/faker/compare/main...ST-DDT:feature/locales/multi-fallback

pkuczynski:
We should definitely have a space where we could put all locale-independent definitions. Not sure we should call it `global` locale though as it's not connected with anything i18n. I would rather see two folders here:
* definitions (or data)
* i18n (or locales)

Being able to create a path of multilevel fallbacks sounds like a good idea too! I can at least see 3 levels: dialect (eg `de-AT`), language (`de`), default (always `en`). But we should let user construct it they way he wants, including dynamic set created on his project side:

```
faker.locale = ['de-AT', {
  // custom user locale as Partial<FakerLocale>
}
```

We should also define default fallbacks so `en-GB` should always fallback into `en`. This way it could have a small subset of data specific to British English. What I mean here by _fallback_ is that when user specifies `en-GB` as his language it should be actually a deep merge of definitions from `en-GB` with `en`. This way GB can be a super small file with only very specific things on British. Does this make sense?

ST-DDT:
> We should definitely have a space where we could put all locale-independent definitions.

I specifically used a locale for these data to allow for easy and consistent overwriting of the data.
[...]

Issue #547, opened 2022-02-23
Incorrect behavior of `name.firstName()` function


JoCat:
### Describe the bug

When using the faker.name.firstName() function with a non-standard locale and without passing a gender value to the function, the name is still displayed in English.

### Reproduction

For example:
```js
import { faker } from '@faker-js/faker';

faker.locale = 'ru'; // Set another locale

const incorrectFirstName = faker.name.firstName(); // Jack (incorrect)
const correctFirstName = faker.name.firstName(0); // Евгений (correct)
const correctFirstName2 = faker.name.firstName(1); // Жанна (correct)
```

demo: https://stackblitz.com/edit/faker-js-demo-6fb7x2?file=index.ts

JoCat:
thanks, maybe there is a problem in the code here:
https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d43/src/name.ts#L38-L65
at least it differs from what is in the lastName and middleName functions, which work correctly

Shinigami92:
I mean... that is quite the whole function except of 2-3 lines :wink: 
Maybe also have a look into https://github.com/faker-js/faker/pull/440

Shinigami92:
Needs re-evaluation after #644 was merged

JoCat:
Now displays Deprecation Warning
![image](https://user-images.githubusercontent.com/13483132/159767273-be931528-a490-4f0d-8ce7-ac3ed5be8c75.png)

Shinigami92:
Yes, that is freshly implemented by us 🙂 
Please do not use `0` or `1` as parameter, but `'male'` or `'female'`

JoCat:
> Yes, that is freshly implemented by us 🙂
> Please do not use 0 or 1 as parameter, but 'male' or 'female'

yes, didn't notice
It turns out the bug is still there

![image](https://user-images.githubusercontent.com/13483132/159795516-d1507568-31f7-4bc2-a13d-503843015d85.png)

ST-DDT:
This issue is more complex then it looks like.

The basics:

- `female` is defined in `ru`
- `male` is defined in `ru`
- `generic` is defined in `en` (fallback language)
- If `generic` is not defined, it will use `female` or `male` randomly.

The possible solutions:
- Create/Configure a faker instance without the `en` fallback (Suggested workaround for now)
- Check whether `generic` is defined in the same language as `female` and `male`, if not treat it as not defined
[...]

PR #858, opened 2022-04-20
feat!: multi locale fallback

ST-DDT:
This PR introduces multiple fallback locales using an array of formats.

**Old**

````ts
faker.locale = 'fr_CH';
faker.fallbackLocale = 'en';
````

**New**

````ts
faker.localeOrder = ['fr_CH', 'fr', 'en'];
````

I'm not sure about removing support for these to setters:

````ts
faker.locale = 'fr_CH';
faker.fallbackLocale = 'en';
````

and the following constructor parameters

````ts
const faker = new Faker({
    locale: 'fr_CH',
    fallbackLocale: 'en',
    locales: { ... }
});
````

The coverage gets fixed in #849.

### CODE REVIEW

File: src/faker.ts [hunk 1]
```diff
@@ -38,14 +38,12 @@ export type UsedLocales = Partial<Record<UsableLocale, LocaleDefinition>>;
...
-  locale?: UsableLocale;
-  localeFallback?: UsableLocale;
+  localeOrder?: UsableLocale[];
 }
 
 export class Faker {
   locales: UsedLocales;
-  locale: UsableLocale;
-  localeFallback: UsableLocale;
+  localeOrder: UsableLocale[];
```
Shinigami92, line 46: I assume we need to make these readonly, so we can control when setting it

src/faker.ts
```diff
@@ -159,11 +167,25 @@ export class Faker {
...
    * @param locale The locale to set (e.g. `en` or `en_AU`, `en_AU_ocker`).
+   * @param fallbacks The fixed fallbacks to use. Defaults to `['en']`.
    */
-  setLocale(locale: UsableLocale): void {
-    this.locale = locale;
+  setLocale(locale: UsableLocale, fallbacks: UsableLocale[] = ['en']): void {
+    const localeOrder = [];
+    const parts = locale.split('_');
+    for (let i = parts.length; i > 0; i--) {
+      const subLocale = parts.slice(0, i).join('_');
```
Shinigami92, line 179: **question:** could you explain me whats going on here? if you would like via voice chat :slightly_smiling_face:

ST-DDT: `'en_AU_ocker'` -> `['en', 'AU', 'ocker']` // L177

-> `['en', 'AU', 'ocker']` -> `'en_AU_ocker'` // L179
-> `['en', 'AU']` -> `'en_AU'` // L179
-> `['en']` -> `'en'` // L179

Shinigami92: Oh... I see, and it will not cause issues, if a in-between locale could not be found cause it's not defined?
Also maybe it should be documented that this exists / handle like this

src/locale/ar.ts
```diff
@@ -8,8 +8,7 @@ import ar from '../locales/ar';
 import en from '../locales/en';
 
 const faker = new Faker({
-  locale: 'ar',
-  localeFallback: 'en',
+  localeOrder: ['ar', 'en'],
```
pkuczynski, line 11: As `en` is presented in every locale definition, I think we should skip it and add it as default on usage. Would make the config much simpler.

src/locale/ar.ts
```diff
@@ -8,8 +8,7 @@ import ar from '../locales/ar';
...
   locales: {
```
pkuczynski, line 12: If this would be an array and not object, you would not need the extra prop `localeOrder` right?

src/faker.ts
pkuczynski: Side note: 

ES6 way:

```js
'en_AU_ocker'.split('_').reduce((result, _, i, parts) => {
    return result.concat(parts.slice(0, parts.length - i).join('_'))
}, [])
```

No split:

```js
'en_AU_ocker'.split('_').reduce((result, _, i, parts) => {
    parts.pop()
    return result.concat(parts.join('_'))
}, ['en_AU_ocker'])
```

src/locale/ar.ts
Shinigami92: https://github.com/faker-js/faker/pull/642

src/faker.ts
```diff
@@ -38,14 +38,12 @@ export type UsedLocales = Partial<Record<UsableLocale, LocaleDefinition>>;
 
 export interface FakerOptions {
   locales: UsedLocales;
-  locale?: UsableLocale;
-  localeFallback?: UsableLocale;
```
Shinigami92, line 42: We should somehow deprecate these with since v7 until v8 to make a smooth transition possible for users

src/locale/ar.ts
ST-DDT: So instead of configuring it explicitly you want to move it to a default / implicit value?
I explicitly didn't choose that approach to allow users to omit the `en` locale to avoid certain fallback mechanism issues.

Shinigami92 (changes requested):

src/faker.ts
ST-DDT: There needs to be a way to set them, it may also be useful to get them.
Especially so, if you consider it in combination with seed.

src/locale/ar.ts
pkuczynski: I don't think this should be possible tbh. `en` is the only _full_ locale and mostly maintained. If you would like to create a new feature which needs locale, we can't block it until its implemented on every locale, right?

The only interface we should offer for locales should be:

```js
const faker = new Faker({ locales: ['de_AT', 'fr'] })
```

which would mean following fallback line: `de_AT > de > fr > en`

I can't imagine someone would prefer not working feature over getting a string in English when its not supported in a locale selected by him.

src/locale/ar.ts
pkuczynski: See my comment above...

src/faker.ts
ST-DDT: > Oh... I see, and it will not cause issues, if a in-between locale could not be found cause it's not defined?

https://github.com/faker-js/faker/blob/75ffa2ef271238fb277175c4ed6d4058432e5a12/src/faker.ts#L180

> Also maybe it should be documented that this exists / handle like this

I will try to do that.

> Side note:

IMO my "old school" approach is way easier to read for me.

src/locale/ar.ts
Shinigami92: `const faker = new Faker({ locales: ['de_AT', 'fr'] })`

1. `locales` is already in use and reserved for the locales itself that can be passed
2. That would resolve to `de_AT > de > fr` not `de_AT > de > fr > en`, maybe you don't want `en`, if you want it, pass it

src/locale/af_ZA.ts
```diff
@@ -8,8 +8,7 @@ import af_ZA from '../locales/af_ZA';
 import en from '../locales/en';
 
 const faker = new Faker({
-  locale: 'af_ZA',
-  localeFallback: 'en',
+  localeOrder: ['af_ZA', 'en'],
```
ST-DDT, line 11: Why isn't this shown as covered?

https://github.com/faker-js/faker/blob/75ffa2ef271238fb277175c4ed6d4058432e5a12/test/locale-imports.spec.ts#L15-L23

src/locale/ar.ts
pkuczynski: > 1 `locales` is already in use and reserved for the locales itself that can be passed

I am not sure if I understand? This is exactly what we want, initialise faker with locales, right?

> 2.

I think it should always add `en` no matter what. Imagine user calls `faker.animal.dog()` and there is no data for it in neither of the locales he specified. What should faker do then:

1. return empty string?
2. throw an error?
3. or return something from `en`

I would vote for 3. Better to return English name, than nothing. But maybe we should throw?

src/locale/ar.ts
pkuczynski: How do you switch between locales?

src/faker.ts
pkuczynski: > IMO my "old school" approach is way easier to read for me.

Both fine with me. I just posted this for fun :) Possibly the _no split_ version could be understood better then `slice`, not sure...

src/locale/ar.ts
ST-DDT: > > 1 `locales` is already in use and reserved for the locales itself that can be passed
> 
> I am not sure if I understand? This is exactly what we want, initialise faker with locales, right?

`locales` -> available data; `localeOrder` -> active data (+ their order)


> > 2.
> 
> I think it should always add `en` no matter what. Imagine user calls `faker.animal.dog()` and there is no data for it in neither of the locales he specified. What should faker do then:

We add it by default, to exactly ensure this behavior.
However, to avoid an issue like #547, we have to avoid hardcoded fallbacks.
Sometimes having an undefined/error is easier to detect, then a random english word.

See also #823

src/locale/ar.ts
ST-DDT: faker.localeOrder = ['fr'];
faker.setLocale('fr_BE');

ST-DDT:
The coverage gets fixed in #849.

src/locale/ar.ts
ST-DDT: > Can you come up with sensible example when switching locales on single version of faker is useful?

Because its easier, otherwise we wouldn't do that in our tests.

> > locales -> available data; localeOrder -> active data (+ their order)
> 
> This is not intuitive at all.

For me it is, also I don't want to change the base logic too much with this PR.
(I intend to make this change backwards compatible, switching to ordered locale data instead of keys might make this impossible)

Please note, that this might be due to some thoughts I have regarding the locale loading.
I consider always adding all locale data to the faker instance, but load the actual contents only lazily.
See also #830 
Since that's somewhat in the future, I would like to take one step at a time and fix some of our limitations, regarding the use of multiple locales.
In this PR, I would like to mostly stick to existing API. We can replace/improve the existing API in a separate PR to make reviews easier and focused on a single change. 

I'm not really happy of how big of a change this fix actually is and think about ways to simplify it (e.g. don't replace fallbackLocale with localeOrder and just make locale a single key or array).

ST-DDT:
@pkuczynski: @Shinigami92 also wants to merge `localeOrder` and `locales` into one.
However, he doesn't want to just drop `locale` and `localeFallback` without deprecating it first.
I'll check whether I can incorporate the backwards compatibility in this PR, otherwise this would be moved to v8 (as a hard break), so we don't have to many breaking changes in a single version.

ST-DDT:
We will use a different approach.

PR #864, opened 2022-04-22
docs: make clear that latitude is first on nearbyGPSCoordinate

efstathiosntonas:


ST-DDT (dismissed):
Just curious, isn't that the standard?

efstathiosntonas:
@ST-DDT no, [mongodb](https://www.mongodb.com/docs/manual/geospatial-queries/#geojson-objects) wants the longitude first. Google maps wants latitude first.

ST-DDT:
Could you please add it to L31 (`coordinateWithOffset()`) as well?

efstathiosntonas:
sure, missed that one, sorry.

ST-DDT (approved):

src/address.ts
```diff
@@ -28,7 +28,7 @@ function kilometersToMiles(miles: number): number {
 }
 
 /**
- * Calculates coordinates with offset.
+ * Calculates coordinates with offset (latitude first).
```
Shinigami92, line 31: Instead of writing it into the JSDoc comment, I would suggest to bring it directly into the TS definition
So in line 39 we can write `coordinate: [latitude: number, longitude: number]`

If I'm correct this is supported by TS >= 4.0, but feel free to double-check

Shinigami92 (changes requested):

src/address.ts
efstathiosntonas: @Shinigami92 tested it on TS 3.6.3 and it [works](https://codesandbox.io/s/bold-glade-dzwj9j?file=/src/index.ts), will change it now.

src/address.ts
```diff
@@ -478,7 +478,7 @@ export class Address {
   }
 
   /**
-   * Generates a random GPS coordinate within the specified radius from the given coordinate.
+   * Generates a random GPS coordinate within the specified radius from the given coordinate (latitude first).
```
Shinigami92, line 481: same here

Shinigami92 (changes requested):

src/address.ts
efstathiosntonas: done!

src/address.ts
```diff
@@ -492,7 +492,7 @@ export class Address {
    */
   // TODO ST-DDT 2022-02-10: Allow coordinate parameter to be [string, string].
   nearbyGPSCoordinate(
-    coordinate?: [number, number],
+    coordinate?: [latitude: number, longitude: number],
     radius?: number,
     isMetric?: boolean
   ): [string, string] {
```
Shinigami92, line 498: If possible, I think we could also use this here for the return value

Shinigami92 (changes requested):
Also line 43 could be adjusted 🤔

src/address.ts
efstathiosntonas: About L43, you mean something like: `{latitude: number, longitude: number}[]` instead of `number[]`?

ST-DDT (approved):

ST-DDT:
Seems to work as expected:
grafik

Shinigami92 (approved):

efstathiosntonas:
Thank you @ST-DDT, keep it up! I think we saved some time for the fellow devs in this one, had to go through the source code to see witch came first.

PR #869, opened 2022-04-24
test: refactor random tests to use validator.js methods

import-brain:


test/random.spec.ts
```diff
@@ -257,9 +257,7 @@ describe('random', () => {
       });
 
       expect(alphaText).toHaveLength(5);
-      for (const bannedChar of bannedChars) {
-        expect(alphaText).not.includes(bannedChar);
-      }
+      expect(alphaText).toMatch(/^[0-9]*$/);
```
import-brain, line 260: This matches numbers only.

test/random.spec.ts
```diff
@@ -269,9 +267,7 @@ describe('random', () => {
       });
 
       expect(alphaText).toHaveLength(5);
-      for (const bannedChar of bannedChars) {
-        expect(alphaText).not.includes(bannedChar);
-      }
+      expect(alphaText).toMatch(/^([^0-9]*)$/);
```
import-brain, line 270: This matches everything except numbers.

test/random.spec.ts
Shinigami92, line 260: While reading this test, I find it confusing that the variable is called `alphaText`. I assume this is a copy-paste issue from history. Should we name it maybe `actual` or `alphaNumericText`?
I would prefer `actual` for simplicity.

Applies for the part below as well.

test/random.spec.ts
Shinigami92: Should we use `\d` for this?

test/random.spec.ts
Shinigami92: Should we use `\d` for this?

xDivisionByZerox:
I don't know about this PR. If I remember correctly we have an unwritten rule to use as less RegExp as possible.
It can get really fast out of hand and is for most people not that readable. Furthermore (in this specific case) using the provided matcher (`not.includes`) you receive a much nicer error message to debug.

Shinigami92:
> I don't know about this PR. If I remember correctly we have an unwritten rule to use as less RegExp as possible. It can get really fast out of hand and is for most people not that readable. Furthermore (in this specific case) using the provided matcher (`not.includes`) you receive a much nicer error message to debug.

Would you like to add this statement to our docs? And yes I also se it like you and I already thought about to suggest using `toSatisfy` with validators. But if you even say that the output message is better readable, then we may even not do anything here.

test/random.spec.ts
ST-DDT:
```suggestion
      expect(alphaText).toMatch(/^[a-z]*$/);
```

test/random.spec.ts
ST-DDT: I think this is good enough.
If we rename the variables, then we should do so consistently in another PR.

ST-DDT:
Maybe instead of regex use the related validator.js methods in compination with satisfy?

test/random.spec.ts
Shinigami92: Are upper case letters also allowed? Or is that another case not covered in this test-path?

import-brain:
> Maybe instead of regex use the related validator.js methods in compination with satisfy?

I didn't use satisfy directly, but I think my method also works. There might be a better way to write this, though.

ST-DDT (approved):

ST-DDT (changes requested):
Oh no. Now the failure message no longer contains the bad value.

Shinigami92 (approved):

ST-DDT (approved):

PR #874, opened 2022-04-26
feat: faker.helpers.maybe

### LINKED ISSUE #870: feat: faker optional with probability

I need to create mock data and would like to generate some optional fields for entities.
I would also like to pass a probability so I can control a bit how often a value will be set.
The function needs to be deterministic.

### Suggested solution

My current own written function looks like this: 

```ts
function optional<T>(
  callback: () => T,
  { probability = 0.5 }: { probability: number } = { probability: 0.5 },
): T | undefined {
  if (faker.datatype.float({ min: 0, max: 1 }) < probability) {
    return callback();
  }
  return undefined;
}
```

### Alternative

_No response_

### Additional context

_No response_

Shinigami92:
closes #870 

GitHub Copilot generated some JSDocs and provided the naming `maybe` instead of `optional`, and I like that more, so I took that as name

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -691,4 +691,27 @@ export class Helpers {
       account: this.faker.finance.account(),
     };
   }
+
+  /**
+   * Returns the result of the callback if the probability check was successful, otherwise `undefined`.
+   *
+   * @param callback The callback to that will be invoked if the probability check was successful.
+   * @param options The options to use. Defaults to `{}`.
+   * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
+   *
+   * @example
+   * faker.random.maybe(() => 'Hello World!') // 'Hello World!'
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.1 }) // undefined
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.9 }) // 'Hello World!'
+   */
+  maybe<T>(
+    callback: () => T,
+    options: { probability?: number } = {}
+  ): T | undefined {
+    const { probability = 0.5 } = options;
+    if (this.faker.datatype.float({ min: 0, max: 1 }) < probability) {
+      return callback();
+    }
+    return undefined;
+  }
 }
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -695,6 +695,7 @@ export class Helpers {
   /**
    * Returns the result of the callback if the probability check was successful, otherwise `undefined`.
    *
+   * @template T The type of result of the given callback.
    * @param callback The callback to that will be invoked if the probability check was successful.
    * @param options The options to use. Defaults to `{}`.
    * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
--- a/src/random.ts
+++ b/src/random.ts
@@ -614,7 +614,7 @@ export class Random {
    *
    * @param callback The callback to that will be invoked if the probability check was successful.
    * @param options The options to use. Defaults to `{}`.
-   * @param options.probability The probability of the callback being invoked. Defaults to `0.5`.
+   * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
    *
    * @example
    * faker.random.maybe(() => 'Hello World!') // 'Hello World!'
--- a/src/random.ts
+++ b/src/random.ts
@@ -608,27 +608,4 @@ export class Random {
 
     return this.faker.datatype.hexadecimal(count);
   }
-
-  /**
-   * Returns the result of the callback if the probability check was successful, otherwise `undefined`.
-   *
-   * @param callback The callback to that will be invoked if the probability check was successful.
-   * @param options The options to use. Defaults to `{}`.
-   * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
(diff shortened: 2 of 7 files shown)


src/random.ts
```diff
@@ -608,4 +608,27 @@ export class Random {
...
+   * @param options The options to use. Defaults to `{}`.
+   * @param options.probability The probability of the callback being invoked. Defaults to `0.5`.
+   *
+   * @example
+   * faker.random.maybe(() => 'Hello World!') // 'Hello World!'
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.1 }) // undefined
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.9 }) // 'Hello World!'
+   */
+  maybe<T>(
+    callback: () => T,
```
Shinigami92, line 625: @ST-DDT When I try to use `(() => T) | T` here, I get into some issues on line 630 as `T` can also be a `Function` 🤔 
Is it worth trying to find a way to support this? Or would you be okay with letting users just use the callback for at least the initial state of this PR 🤔

ST-DDT: Lets keep it simple for now and don't support plain `T`s for now.
If there is a feature request for it, then we can add it later.

Shinigami92: Wait... no, plain `T` cannot be! It is an explicit feature of this PR to not invoke the callback if not necessary!

src/random.ts
```diff
@@ -608,4 +608,27 @@ export class Random {
 
     return this.faker.datatype.hexadecimal(count);
   }
+
+  /**
+   * Returns the result of the callback if the probability check was successful, otherwise `undefined`.
+   *
+   * @param callback The callback to that will be invoked if the probability check was successful.
+   * @param options The options to use. Defaults to `{}`.
+   * @param options.probability The probability of the callback being invoked. Defaults to `0.5`.
```
ST-DDT, line 617:
```suggestion
   * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
```

Maybe write a hint that the precision is capped at 2 decimal places?

Shinigami92: It's because of the default precision of `float`?

ST-DDT: Yes, for the boolean probability version I thought about deriving the precision from the inputs.

src/random.ts
ST-DDT: Why would that prevent plain `T`? If callback is a function, we will call it (when necessary), if it is something else we return it (when necessary).

Shinigami92: image

Shinigami92: If I just use `callback: T` how do I access the return type when it is a function?

ST-DDT (dismissed):

xDivisionByZerox:
@ST-DDT proposed [here](https://github.com/faker-js/faker/issues/870#issuecomment-1108436787) to place this method in the helpers module. I think that would be better than exposing it in the random module.

Shinigami92:
> @ST-DDT proposed [here](https://github.com/faker-js/faker/issues/870#issuecomment-1108436787) to place this method in the helpers module. I think that would be better than exposing it in the random module.

Oh, I oversaw this. But my current understatement is that `random` should handle values that are based on deterministic and `helpers` contains only stuff that are like utilities :thinking: 
Like `slugify` or `repeatString`
I also wrote that down here: https://github.com/faker-js/faker/discussions/805#discussioncomment-2538958

But I can be wrong and if you are voting for handling it differently, let's define that now and write that down somewhere.

ST-DDT:
> But my current understatement is that `random` should handle values that are based on deterministic

For the return value that is. Here the value/source is passed by the user.

> and `helpers` contains only stuff that are like utilities 🤔 Like `slugify` or `repeatString` 

IMO this maybe is exactly a helper, as it simplifies certain tasks for the user, but the output is heavily dependent on the users input and not some locale.

> I also wrote that down here: [#805 (reply in thread)](https://github.com/faker-js/faker/discussions/805#discussioncomment-2538958)

I couldn't find the exact part that you were referring to, probably not:

> Random
> 
> * It could be fully removed? ...

Shinigami92:
I was talking about
> - Helpers
>   - Should only contain stuff that is not based on locale definitions!

For deterministic, I talked about https://github.com/faker-js/faker/blob/f3dab6777b017f7005c40e293f1cc75dd1c882ab/src/random.ts#L629

ST-DDT:
IMO:
- Helpers is for everything that transforms the input
- Random generates new stuff

xDivisionByZerox (approved):

src/helpers.ts
```diff
@@ -691,4 +691,27 @@ export class Helpers {
...
+   *
+   * @param callback The callback to that will be invoked if the probability check was successful.
+   * @param options The options to use. Defaults to `{}`.
+   * @param options.probability The probability (`[0.00, 1.00]`) of the callback being invoked. Defaults to `0.5`.
+   *
+   * @example
+   * faker.random.maybe(() => 'Hello World!') // 'Hello World!'
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.1 }) // undefined
+   * faker.random.maybe(() => 'Hello World!', { probability: 0.9 }) // 'Hello World!'
+   */
```
ST-DDT, line 706: grafik

ST-DDT (changes requested):

ST-DDT (approved):
grafik

xDivisionByZerox (approved):

PR #875, opened 2022-04-26
test: migrate datatype to test snapshots

ST-DDT:
We should migrate from our old inline seeded tests to using test snapshots, as these are way easier to update.
I plan to update the other modules as well, but this one is the first in case something needs to be changed.


xDivisionByZerox:
I'm sorry but in what world is this easier to maintain? By changing any label in the entire test chain, you have an invalid expectation key.  

Also since every snapshot value is a string, prettier can't apply the style guide correctly. Or am I missing something?

Can you please elaborate on how this helps maintainability wise?

Shinigami92:
Snapshots can be updated by just using [`--update`](https://vitest.dev/guide/cli.html#options)

But yeah, I was not aware that it just uses strings and therefore we loose the types 🤔

xDivisionByZerox:
Ok, I didn't know there was such an option. Thought the migration was done by hand.

Shinigami92:
> Ok, I didn't know there was such an option. Thought the migration was done by hand.

That's the whole point of using snapshots 😉

Shinigami92:
I would prefer to have all tests migrated in one PR so it is only one commit in main branch history. You could create commits test by test to make review easier.

import-brain (dismissed):

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):

import-brain (dismissed):

Shinigami92 (approved):

import-brain (approved):

PR #876, opened 2022-04-26
fix: address.nearbyGPSCoordinate

### LINKED ISSUE #321: Investigate why test address gps coords for non-metric sometimes fails
### Describe the bug

[`218ae3a` (#285)](https://github.com/faker-js/faker/pull/285/commits/218ae3ace1085244a83ac32f56cea93f123cf389)

### Reproduction

comment in the outcommented code and see that it sometimes fails

### Additional Info

Not sure if the test is written wrongly or if the function really is broken

Shinigami92:
closes #321

### DIFF
--- a/src/address.ts
+++ b/src/address.ts
@@ -41,22 +41,22 @@ function coordinateWithOffset(
   distance: number,
   isMetric: boolean
 ): [latitude: number, longitude: number] {
-  const R = 6378.137; // Radius of the Earth (http://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html)
+  const EARTH_RADIUS = 6378.137; // http://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html
   const d = isMetric ? distance : kilometersToMiles(distance); // Distance in km
 
-  const lat1 = degreesToRadians(coordinate[0]); //Current lat point converted to radians
-  const lon1 = degreesToRadians(coordinate[1]); //Current long point converted to radians
+  const lat1 = degreesToRadians(coordinate[0]); // Current lat point converted to radians
+  const lon1 = degreesToRadians(coordinate[1]); // Current long point converted to radians
 
   const lat2 = Math.asin(
-    Math.sin(lat1) * Math.cos(d / R) +
-      Math.cos(lat1) * Math.sin(d / R) * Math.cos(bearing)
+    Math.sin(lat1) * Math.cos(d / EARTH_RADIUS) +
+      Math.cos(lat1) * Math.sin(d / EARTH_RADIUS) * Math.cos(bearing)
   );
 
   let lon2 =
     lon1 +
     Math.atan2(
-      Math.sin(bearing) * Math.sin(d / R) * Math.cos(lat1),
-      Math.cos(d / R) - Math.sin(lat1) * Math.sin(lat2)
+      Math.sin(bearing) * Math.sin(d / EARTH_RADIUS) * Math.cos(lat1),
+      Math.cos(d / EARTH_RADIUS) - Math.sin(lat1) * Math.sin(lat2)
     );
 
   // Keep longitude in range [-180, 180]
--- a/src/address.ts
+++ b/src/address.ts
@@ -454,15 +454,9 @@ export class Address {
 
     const distanceInDegree = distanceInKm / kmPerDegree; // in °
 
-    const offsetLatitude = Math.sin(angleRadians) * distanceInDegree;
-    const newLatitude = coordinate[0] + offsetLatitude;
-
-    const offsetLongitude = Math.cos(angleRadians) * distanceInDegree;
-    const newLongitude = coordinate[1] + offsetLongitude;
-
     const newCoordinate: [latitude: number, longitude: number] = [
-      newLatitude,
-      newLongitude,
+      coordinate[0] + Math.sin(angleRadians) * distanceInDegree,
+      coordinate[1] + Math.cos(angleRadians) * distanceInDegree,
     ];
 
     // Box latitude [-90°, 90°]
--- a/src/address.ts
+++ b/src/address.ts
@@ -41,22 +41,24 @@ function coordinateWithOffset(
   distance: number,
   isMetric: boolean
 ): [latitude: number, longitude: number] {
-  const EARTH_RADIUS = 6378.137; // http://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html
+  const EQUATORIAL_EARTH_RADIUS = 6378.137; // http://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html
   const d = isMetric ? distance : kilometersToMiles(distance); // Distance in km
 
   const lat1 = degreesToRadians(coordinate[0]); // Current lat point converted to radians
   const lon1 = degreesToRadians(coordinate[1]); // Current long point converted to radians
 
   const lat2 = Math.asin(
-    Math.sin(lat1) * Math.cos(d / EARTH_RADIUS) +
-      Math.cos(lat1) * Math.sin(d / EARTH_RADIUS) * Math.cos(bearing)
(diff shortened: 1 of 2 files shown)


ST-DDT:
It looks like the float values caused massive issues especially since we stacked the calls on those values.
With the simplified implementation we have far less error and use a much larger part of the circle.

**Actual**
Min: 0.00042
Avg: 0.07065
Max: 0.14406

**Expected**
Min: 0.00000
Avg: 0.07500
Max: 0.15000

That's 96% coverage vs the old ~0.08%

@Shinigami92 Anything you want to change/fix here?

src/address.ts
```diff
@@ -1,73 +1,9 @@
...
-  // Keep longitude in range [-180, 180]
-  if (lon2 > degreesToRadians(180)) {
-    lon2 = lon2 - degreesToRadians(360);
-  } else if (lon2 < degreesToRadians(-180)) {
-    lon2 = lon2 + degreesToRadians(360);
-  }
-
-  return [radiansToDegrees(lat2), radiansToDegrees(lon2)];
-}
+const kmPerDegree = 40_000 / 360; // in km/°
```
Shinigami92, line 6: Script constants should be written in CONSTANT_CASE

But this should also be named with earth, cause maybe we later want to have some other circumferences for other custom planetoids

test/address.spec.ts
```diff
@@ -1,6 +1,36 @@
...
+}
+
+function haversine(
+  latitude1: number,
+  longitude1: number,
+  latitude2: number,
+  longitude2: number,
+  isMetric: boolean
+) {
+  const EQUATORIAL_EARTH_RADIUS = 6378.137;
```
Shinigami92, line 19: We should move this out of the function

test/address.spec.ts
```diff
@@ -89,12 +119,12 @@ const seededRuns = [
...
       cardinalDirectionAbbr: 'W',
       timeZone: 'Asia/Magadan',
-      nearbyGpsCoordinates: ['0.0503', '-0.0242'],
+      nearbyGpsCoordinates: ['-0.0287', '0.0596'],
     },
   },
 ];
 
-const NON_SEEDED_BASED_RUN = 5;
+const NON_SEEDED_BASED_RUN = 1;
```
Shinigami92, line 127:
```suggestion
const NON_SEEDED_BASED_RUN = 5;
```

test/address.spec.ts
```diff
@@ -556,39 +586,14 @@ describe('address', () => {
...
-          let latFloat1: number;
-          let lonFloat1: number;
-          let isMetric: boolean;
-
-          for (let i = 0; i < 10000; i++) {
-            latFloat1 = parseFloat(faker.address.latitude());
-            lonFloat1 = parseFloat(faker.address.longitude());
-            const radius = Math.random() * 99 + 1; // range of [1, 100)
-            isMetric = Math.round(Math.random()) === 1;
+          for (let i = 0; i < 100; i++) {
```
Shinigami92, line 589: At least thousends, maybe even back to 10_000?

```suggestion
          for (let i = 0; i < 1_000; i++) {
```

```suggestion
          for (let i = 0; i < 10_000; i++) {
```

test/address.spec.ts
```diff
@@ -1,6 +1,36 @@
...
+  latitude1: number,
+  longitude1: number,
+  latitude2: number,
+  longitude2: number,
+  isMetric: boolean
+) {
+  const EQUATORIAL_EARTH_RADIUS = 6378.137;
+  const distanceLatitude = degreesToRadians(latitude2 - latitude1);
+  const distanceLongitude = degreesToRadians(longitude2 - longitude1);
+  const a =
```
Shinigami92, line 22: Could we find a better name for `a`? I currently don't know what it is :thinking:

test/address.spec.ts
```diff
@@ -556,39 +586,14 @@ describe('address', () => {
...
-          for (let i = 0; i < 10000; i++) {
-            latFloat1 = parseFloat(faker.address.latitude());
-            lonFloat1 = parseFloat(faker.address.longitude());
-            const radius = Math.random() * 99 + 1; // range of [1, 100)
-            isMetric = Math.round(Math.random()) === 1;
+          for (let i = 0; i < 100; i++) {
+            const latitude1 = +faker.address.latitude();
+            const longitude1 = +faker.address.longitude();
+            const radius = faker.datatype.float({ min: 1, max: 100 });
+            const isMetric = faker.datatype.boolean();
```
Shinigami92, line 593: I would like to move this out to the `it` as `it.each` and then split the runs into half for each, so maybe 5_000/5_000?

ST-DDT (dismissed):

test/address.spec.ts
```diff
@@ -556,39 +586,14 @@ describe('address', () => {
...
-
-          for (let i = 0; i < 10000; i++) {
-            latFloat1 = parseFloat(faker.address.latitude());
-            lonFloat1 = parseFloat(faker.address.longitude());
-            const radius = Math.random() * 99 + 1; // range of [1, 100)
-            isMetric = Math.round(Math.random()) === 1;
+          for (let i = 0; i < 100; i++) {
+            const latitude1 = +faker.address.latitude();
+            const longitude1 = +faker.address.longitude();
+            const radius = faker.datatype.float({ min: 1, max: 100 });
```
Shinigami92, line 592: We could control this one also a bit more with it.each together with the `isMetric`

so what about e.g. `it.each(times(100).map(i=>[i+1,isMetric,radius]))` ...

Okay maybe a 3-stacked for-loop is better here?
```ts
for (const isMetric of [true, false])
  for (const radius of times(100))
    for (const run of times(1000))
      it(...)
```

test/address.spec.ts
ST-DDT: I wouldn't recommend that, as this just wastes time. We don't do the NON_SEEDED_BASE_RUN that often either.

test/address.spec.ts
ST-DDT: Me neither. Maybe check the internet?

test/address.spec.ts
```diff
@@ -555,119 +588,45 @@ describe('address', () => {
...
-                Math.sin(dLon / 2);
-            const distance = R * 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
-
-            return isMetric ? distance : kilometersToMiles(distance);
+        for (const isMetric of [true, false]) {
+          for (const radius of times(100)) {
+            it(`should return random gps coordinate within a distance of another one (${JSON.stringify(
+              { isMetric, radius }
+            )})`, () => {
+              for (let i = 0; i < 100; i++) {
```
ST-DDT, line 596: A single run is probably enough.
Otherwise we end up with too many checks:
- `runs * isMetric * radius * i`
- `5 * 2 * 100 * 100`
- `= 100000`

If you omit the i-for-loop, then we still have `1000` test runs.
Maybe leave the loop there, but just reduce it to `5`,
that way it's very easy to increment the runs if there is a need to debug it.

ST-DDT (changes requested):

src/address.ts
```diff
@@ -452,6 +447,12 @@ export class Address {
         precision: 0.001,
       }) * errorCorrection; // in km
 
+    /**
+     * The distance in km per degree for earth.
+     */
+    // TODO @Shinigami92 2022-04-26: Provide an option property to provide custom circumferences.
```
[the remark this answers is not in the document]
Shinigami92: Done https://github.com/faker-js/faker/issues/878

test/address.spec.ts
Shinigami92: I found this https://www.movable-type.co.uk/scripts/latlong.html
But even there it is called `a`. :shrug: 
Because I don't want to stop this PR to get merged just based on that, lets move on :shrug:

ST-DDT (approved):

import-brain (approved):

PR #877, opened 2022-04-26
refactor(word): extract wordlist filter in separate method

xDivisionByZerox:
Refactor shared logic in the word module in a separate function.

Designed in #589.

### DIFF
--- a/src/word.ts
+++ b/src/word.ts
@@ -1,6 +1,31 @@
 import type { Faker } from '.';
 
 /**
+ * Filters a string array for values with a specific length.
+ * If length is not provided or no values with this length there found a copy of the original array is returned.
+ *
+ * @param options The options to provide
+ * @param options.wordList A list of word to filter
+ * @param options.length The exact length words should have
+ */
+function filterWordListByLength(options: {
+  wordList: string[];
+  length?: number;
+}): string[] {
+  if (!length) {
+    return options.wordList;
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):

src/word.ts
```diff
@@ -1,5 +1,30 @@
...
+    return options.wordList;
+  }
+
+  const wordListWithLengthFilter = options.wordList.filter(
+    (word) => word.length === options.length
+  );
+
+  return wordListWithLengthFilter.length > 0
+    ? wordListWithLengthFilter
+    : [...options.wordList];
```
ST-DDT, line 25: Why do you create a copy here?

xDivisionByZerox: I guess just a default thing I do to ensure that you don't accidentally mutate an existing object. this for example:
```
const myArr= ['a', 'b', 'c'];
const myResult = filterWordListByLength({
    wordList: myArray,
    length: 0,
});
cons lastElement = myResult.pop(); // mutates myResult array by removing the last element

console.log(lastElement); // 'c'
console.log(myResult) // ['a', 'b'] - as expected
console.log(myArr) // ['a', 'b'] - not expected since we never touched myArr
```

ST-DDT (approved):

Shinigami92 (approved):

PR #884, opened 2022-04-28
feat: support locale definitions directly from faker.fake

ST-DDT:
**Previously**

- `fake('{{<pattern>}}')`
  - -> `fake.<pattern>()`

**Now**

- `fake('{{<pattern>}}')`
  - -> `faker.<pattern>()`
  - -> `faker.random.arrayElement(faker.definitions.<pattern>)`


src/fake.ts
```diff
@@ -88,16 +88,26 @@ export class Fake {
...
+    let currentDefinitions: unknown = this.faker.definitions;
+
+    for (const part of parts) {
+      currentModuleOrMethod = currentModuleOrMethod?.[part];
+      currentDefinitions = currentDefinitions?.[part];
     }
 
-    if (this.faker[parts[0]][parts[1]] == null) {
-      throw new FakerError(`Invalid method: ${parts[0]}.${parts[1]}`);
+    let fn: (args?: unknown) => string;
```
xDivisionByZerox, line 99: In theory, this could return anything not only strings. So typing the return type as unknown to catch unhandled cases should be more secure IMO.

```suggestion
    let fn: (args?: unknown) => unknown;
```

Shinigami92 (approved):
Looks like your enhancements doesn't break anything, so I approve

src/modules/fake/index.ts
```diff
@@ -88,16 +88,26 @@ export class Fake {
...
+      currentDefinitions = currentDefinitions?.[part];
     }
 
-    if (this.faker[parts[0]][parts[1]] == null) {
-      throw new FakerError(`Invalid method: ${parts[0]}.${parts[1]}`);
+    let fn: (args?: unknown) => unknown;
+    if (typeof currentModuleOrMethod === 'function') {
+      fn = currentModuleOrMethod as (args?: unknown) => string;
+    } else if (Array.isArray(currentDefinitions)) {
+      fn = () => this.faker.random.arrayElement(currentDefinitions as []);
```
Shinigami92, line 103:
```suggestion
      fn = () => this.faker.random.arrayElement(currentDefinitions as unknown[]);
```

src/modules/fake/index.ts
```diff
@@ -88,16 +88,26 @@ export class Fake {
...
+    } else {
+      throw new FakerError(`Invalid module method or definition: ${method}
+faker.${method} = ${typeof currentModuleOrMethod} (must be function)
+faker.definitions.${method} = ${typeof currentDefinitions} (must be an array)`);
```
Shinigami92, line 107: This may lead to potential confusion as `typeof [] === 'object'`

ST-DDT: What would you recommend then?

Shinigami92: Maybe we can just leave out the type and write human answers like
```
Invalid module method or definition: ${method}
- faker.${method} is not a function
- faker.definitions.${method} is not an array
```

Also not sure if a 3-line error is a good idea, but I may be biased about that.

I would like to hear a second voice of a member/maintainer :slightly_smiling_face:

import-brain: > Maybe we can just leave out the type and write human answers like
> 
> ```
> Invalid module method or definition: ${method}
> - faker.${method} is not a function
> - faker.definitions.${method} is not an array
> ```
> 
> Also not sure if a 3-line error is a good idea, but I may be biased about that.
> 
> I would like to hear a second voice of a member/maintainer 🙂

I would like these human answers, too. Seems a bit more readable

pkuczynski (approved):

PR #897, opened 2022-05-01
feat(system): add cron

nhammond101:
Adds new functionality to generate cron expressions.
---

**New Function:**
* faker.system.cron()

**Usage:**  

`faker.system.cron()`:

```
// generates a random cron expression
console.log(faker.system.cron()): // '45 23 * * 6'
```
supports an optional options object, which currently has a year component

```
// generates a random cron expression with a year
console.log(faker.system.cron()): // '45 23 * * 6 2067'
```

**Unit tests for the functions added.**


Shinigami92:
Targeting it for v6.3, but I don't think it will make it so fast
We plan to work on v7 in next "sprint"

src/system.ts
```diff
@@ -191,4 +191,46 @@ export class System {
...
+      '?',
+    ];
+    const years = [this.faker.datatype.number({ min: 1970, max: 2099 }), '*'];
+
+    const minute = this.faker.random.arrayElement(minutes);
+    const hour = this.faker.random.arrayElement(hours);
+    const day = this.faker.random.arrayElement(days);
+    const month = this.faker.random.arrayElement(months);
+    const dayOfWeek = this.faker.random.arrayElement(daysOfWeek);
+    const year = this.faker.random.arrayElement(years);
```
Shinigami92, line 226: I think these are perfect use-cases for using `this.faker.helpers.maybe` with `?? '*'` as fallbacks

Shinigami92 (changes requested):
Should we also try to generate some more advanced cron strings like [23 0-20/2 * * *](https://crontab.guru/#23_0-20/2_*_*_*)?

ST-DDT:
Should we add some range for the cron: E.g. `yearly`, `daily`, ...

src/system.ts
```diff
@@ -191,4 +191,46 @@ export class System {
       this.faker.datatype.number(9),
     ].join('.');
   }
+
+  /**
+   * Returns a cron expression
+   *
+   * @param options The optional options to use.
+   * @param options.includeYear Whether to include a year in the generated expression.  Default to `false`
```
ST-DDT, line 199:
```suggestion
   * @param options.includeYear Whether to include a year in the generated expression.  Default to `false`.
```

src/system.ts
```diff
@@ -191,4 +191,46 @@ export class System {
...
+   *
+   * @example
+   * faker.system.cron() // '45 23 * * 6'
+   * faker.system.cron({ includeYear: true }) // '45 23 * * 6 2067'
+   * faker.system.cron({ includeYear: false }) // '45 23 * * 6'
+   */
+  cron(options?: { includeYear: boolean }): string {
```
ST-DDT, line 206:
```suggestion
  cron(options: { includeYear: boolean } = {}): string {
```

src/system.ts
```diff
@@ -191,4 +191,46 @@ export class System {
...
+    // create the arrays to hold the available values for each component of the expression
+    const minutes = [this.faker.datatype.number({ min: 0, max: 59 }), '*'];
+    const hours = [this.faker.datatype.number({ min: 0, max: 23 }), '*'];
+    const days = [this.faker.datatype.number({ min: 1, max: 31 }), '*', '?'];
```
ST-DDT, line 210: What is `?` and why is it only for days?

src/system.ts
ST-DDT: Maybe, maybe not.

ST-DDT (changes requested):

src/system.ts
nhammond101: the `?` represents "no specific value" whereas `*` is any valid value. 

An example is `dayOfWeek` (ie `SAT`) specified, the `dayOfMonth` wouldn't be all values (ie 1-31), but rather a subset, and therefore the `?` is used instead.

I added it for value completeness, but logically the two properties (days, dayOfWeek) aren't linked

nhammond101:
> Should we also try to generate some more advanced cron strings like [23 0-20/2 * * *](https://crontab.guru/#23_0-20/2_*_*_*)?

i'd like to, I wanted to get the simplest expressions generated first and potentially gauge demand

src/modules/system/index.ts
```diff
@@ -191,4 +204,47 @@ export class System {
       this.faker.datatype.number(9),
     ].join('.');
   }
+
+  /**
+   * Returns a cron expression.
+   *
+   * @param options The optional options to use.
+   * @param options.includeYear Whether to include a year in the generated expression. Default to `false`.
```
ST-DDT, line 212:
```suggestion
   * @param options.includeYear Whether to include a year in the generated expression. Defaults to `false`.
```

ST-DDT (changes requested):

Shinigami92 (changes requested):
This PR needs to update their test but we found out that to many other stuff not related to this PR within the test file is messed up
So it will not be merged in v7.0.0

ST-DDT:
@nhammond101 Could you please update this PR?

src/modules/system/index.ts
```diff
@@ -22,6 +22,19 @@ const commonInterfaceSchemas = {
   pci: 'p',
 } as const;
 
+/**
+ * cron days are only in English
+ */
```
Shinigami92, line 27: Not sure if this comment is worth at all 🤷
Maybe just remove it

src/modules/system/index.ts
```diff
@@ -264,4 +277,45 @@ export class System {
 
     return `${prefix}${interfaceType}${commonInterfaceSchemas[interfaceSchema]}${suffix}`;
   }
+
+  /**
+   * Returns a cron expression
+   *
+   * @param options The optional options to use.
+   * @param options.includeYear Whether to include a year in the generated expression.  Default to `false`
```
Shinigami92, line 285:
```suggestion
   * @param options.includeYear Whether to include a year in the generated expression. Default to `false`.
```

test/system.spec.ts
```diff
@@ -385,6 +392,32 @@ describe('system', () => {
           ).toMatch(/^enx[a-f\d]{12}$/);
         });
       });
+
+      describe('cron()', () => {
+        const regex = new RegExp(
+          /^([1-9]|[1-5]\d|\*) ([0-9]|1\d|2[0-3]|\*) ([1-9]|[12]\d|3[01]|\*|\?) ([1-9]|1[0-2]|\*) ([0-6]|\*|\?|[A-Z]{3})/
+        );
```
Shinigami92, line 399: Is there a reason why this is wrapped inside the `RegExp` constructor instead of just using a regex literal?

src/modules/system/index.ts
ST-DDT, line 285:
```suggestion
   * @param options.includeYear Whether to include a year in the generated expression.  Default to `false`.
```

src/modules/system/index.ts
```diff
@@ -264,4 +277,45 @@ export class System {
...
+   *
+   * @example
+   * faker.system.cron() // '45 23 * * 6'
+   * faker.system.cron({ includeYear: true }) // '45 23 * * 6 2067'
+   * faker.system.cron({ includeYear: false }) // '45 23 * * 6'
+   */
+  cron(options?: { includeYear: boolean }): string {
```
ST-DDT, line 292:
```suggestion
  cron(options: { includeYear: boolean } = {}): string {
    const { includeYear = false } = options;
```

test/system.spec.ts
```diff
@@ -385,6 +392,32 @@ describe('system', () => {
...
+        it('should return cron expression with 5 elements', () => {
+          expect(
+            faker.system.cron(),
+            `generated cron, string should contain 5 space-separated values`
```
Shinigami92, line 403: Maybe we should just check with `split(' ')` `length === 5`?
Alternatively the message should be changed
With split we could also check each part individually and server better readable native vitest messages with that

src/modules/system/index.ts
```diff
@@ -264,4 +277,45 @@ export class System {
...
+    const minute = this.faker.helpers.arrayElement(minutes);
+    const hour = this.faker.helpers.arrayElement(hours);
+    const day = this.faker.helpers.arrayElement(days);
+    const month = this.faker.helpers.arrayElement(months);
+    const dayOfWeek = this.faker.helpers.arrayElement(daysOfWeek);
+    const year = this.faker.helpers.arrayElement(years);
+
+    // create and return the cron expression string
+    let expression = `${minute} ${hour} ${day} ${month} ${dayOfWeek}`;
+    if (options?.includeYear) {
```
ST-DDT, line 315:
```suggestion
    if (includeYear) {
```

src/modules/system/index.ts
```diff
@@ -22,6 +22,19 @@ const commonInterfaceSchemas = {
   pci: 'p',
 } as const;
 
+/**
+ * cron days are only in English
```
ST-DDT, line 26:
```suggestion
 * Cron days are only in English.
```

ST-DDT (changes requested):

src/modules/system/index.ts
```diff
@@ -264,4 +277,65 @@ export class System {
...
+        '@hourly',
+        '@monthly',
+        '@reboot',
+        '@weekly',
+        '@yearly',
+      ];
+      availableExpressions.push(...nonStandardExpressions);
+    }
+
+    return this.faker.helpers.arrayElement(availableExpressions);
```
ST-DDT, line 339: IMO this should use a `randomBoolean()? standardExpression : nonStandardExpression`

src/modules/system/index.ts
import-brain: Wouldn't it be "defaults to" instead of "default to"?

src/modules/system/index.ts
```diff
@@ -264,4 +277,65 @@ export class System {
...
+   * faker.system.cron({ includeYear: true }) // '45 23 * * 6 2067'
+   * faker.system.cron({ includeYear: false }) // '45 23 * * 6'
+   * faker.system.cron({ includeNonStandard: false }) // '45 23 * * 6'
+   * faker.system.cron({ includeNonStandard: true }) // '@yearly'
+   */
+  cron(options?: {
+    includeYear?: boolean;
+    includeNonStandard?: boolean;
+  }): string {
+    // create the arrays to hold the available values for each component of the expression
```
import-brain, line 299:
```suggestion
  cron(
  	options: {
      includeYear?: boolean;
      includeNonStandard?: boolean;
    } = {}
  ): string {
  	const { includeYear = false, includeNonStandard = false } = options;
    // create the arrays to hold the available values for each component of the expression
```

src/modules/system/index.ts
```diff
@@ -264,4 +277,65 @@ export class System {
...
+    const year = this.faker.helpers.arrayElement(years);
+
+    // create and return the cron expression string
+    let standardExpression = `${minute} ${hour} ${day} ${month} ${dayOfWeek}`;
+    if (options?.includeYear) {
+      standardExpression += ` ${year}`;
+    }
+
+    const availableExpressions = [standardExpression];
+    if (options?.includeNonStandard) {
```
import-brain, line 326:
```suggestion
    if (includeNonStandard) {
```

src/modules/system/index.ts
import-brain: If you accept this change, please also accept the other one in this review and the change in the review here: https://github.com/faker-js/faker/pull/897/files#r953432873

ejcheng (changes requested):

src/modules/system/index.ts
nhammond101: Removed

test/system.spec.ts
nhammond101: none that I can remember, i'll convert to literal

src/modules/system/index.ts
nhammond101: i've changed it now

src/modules/system/index.ts
nhammond101: done

src/modules/system/index.ts
nhammond101: Updated the lines

ST-DDT (dismissed):

src/modules/system/index.ts
```diff
@@ -264,4 +274,67 @@ export class System {
 
     return `${prefix}${interfaceType}${commonInterfaceSchemas[interfaceSchema]}${suffix}`;
   }
+
+  /**
+   * Returns a cron expression
+   *
+   * @param options The optional options to use.
+   * @param options.includeYear Whether to include a year in the generated expression. Defaults to `false`.
+   * @param options.includeNonStandard Whether to include a @yearly, @monthly, @daily, etc text labels in the generated expression.  Default to `false`
```
ST-DDT, line 283:
```suggestion
   * @param options.includeNonStandard Whether to include a @yearly, @monthly, @daily, etc text labels in the generated expression.  Default to `false`.
```

ST-DDT:
Please rebase your commit:

````sh
git fetch --all
git rebase origin/main
git push -f
````

Alternatively, we can do that for you, but then you have to reset your local checkout.

File: src/modules/system/index.ts [hunk 17]
```diff
@@ -264,4 +274,67 @@ export class System {
 
     return `${prefix}${interfaceType}${commonInterfaceSchemas[interfaceSchema]}${suffix}`;
   }
+
+  /**
+   * Returns a cron expression
+   *
+   * @param options The optional options to use.
+   * @param options.includeYear Whether to include a year in the generated expression. Defaults to `false`.
+   * @param options.includeNonStandard Whether to include a @yearly, @monthly, @daily, etc text labels in the generated expression.  Default to `false`.
```
Shinigami92, line 283:
```suggestion
   * @param options.includeNonStandard Whether to include a @yearly, @monthly, @daily, etc text labels in the generated expression. Defaults to `false`.
```

test/system.spec.ts
Shinigami92: I think my request here is still valid, @ST-DDT what's your opinion about that?

src/modules/system/index.ts
```diff
@@ -264,4 +274,67 @@ export class System {
 
     return `${prefix}${interfaceType}${commonInterfaceSchemas[interfaceSchema]}${suffix}`;
   }
+
+  /**
+   * Returns a cron expression
```
ST-DDT, line 279:
```suggestion
   * Returns a cron expression.
```

ST-DDT (changes requested):
I think these are the last things that needs to be changed.
Otherwise it looks good to me.

test/system.spec.ts
ST-DDT: @Shinigami92 Maybe write an example?

Shinigami92:
```ts
const actual = faker.system.cron()
const parts = actual.split(' ')
expect(parts).toHaveLength(5)

const [minute, hour, day, month, dayOfWeek] = parts
expect(minute).toMatch(/([1-9]|[1-5]\d|\*)/)
expect(hour).toMatch(/([0-9]|1\d|2[0-3]|\*)/)
// ...
```

This is longer but much more readable, understandable and produce much better vitest messages

Shinigami92 (approved):
I would have solved it differently and maybe more readably, but my case/request are tackled and solved :+1:

ST-DDT (approved):

nhammond101:
@Shinigami92 if you get a few minutes, would you post your approach?

Shinigami92:
> if you get a few minutes, would you post your approach?

See here: https://github.com/faker-js/faker/pull/897#discussion_r956063709

ejcheng (approved):

src/modules/system/index.ts
```diff
@@ -264,4 +274,67 @@ export class System {
...
+      '@hourly',
+      '@monthly',
+      '@reboot',
+      '@weekly',
+      '@yearly',
+    ];
+
+    return !includeNonStandard || this.faker.datatype.boolean()
+      ? standardExpression
+      : this.faker.helpers.arrayElement(nonStandardExpressions);
```
xDivisionByZerox, line 338: This can potentially return a `nonStandardExpressions` even with `nonStandardExpressions` explicitly set to `false`.

xDivisionByZerox (changes requested):

src/modules/system/index.ts
xDivisionByZerox: Or am I having trouple with my logic?

xDivisionByZerox: This should have been a normal comment not a change request, my bad.

nhammond101: depends on use case.  For Cron implementations that support the non-standard values, i'd expect to be able to generate both standard and non-standard values, given that both are supported by that implementation.

xDivisionByZerox (approved):

PR #903, opened 2022-05-02
fix: replace deprecated arrayElement calls

### LINKED ISSUE #902: Deprecation message before refactoring - arrayElement()
### Describe the bug

On release 6.3.0,  it was added a warning that `faker.random.arrayElement` is deprecated - but other functions like `faker.random.alphaNumeric` are still referencing the obsolete one.

### Reproduction

just call `fake.random.alphaNumeric()`: 

```
$ node                                                                                                                                                              ─╯
Welcome to Node.js v16.14.0.
Type ".help" for more information.
> const {faker} = require('@faker-js/faker');
undefined
> faker.random.alphaNumeric()
[@faker-js/faker]: faker.random.arrayElement() is deprecated since v6.3.0 and will be removed in v7.0.0. Please use faker.helpers.arrayElement() instead.
'l'
```

### Additional Info

should be considered holding adding deprecation notices until all internal functions are refactored to remove the dependency from the deprecated method.

Shinigami92:
close #902

### DIFF
--- a/src/random.ts
+++ b/src/random.ts
@@ -360,12 +360,12 @@ export class Random {
 
     do {
       // randomly pick from the many faker methods that can generate words
-      const randomWordMethod = this.arrayElement(wordMethods);
+      const randomWordMethod = this.faker.helpers.arrayElement(wordMethods);
 
       result = randomWordMethod();
     } while (!result || bannedChars.some((char) => result.includes(char)));
 
-    return this.arrayElement(result.split(' '));
+    return this.faker.helpers.arrayElement(result.split(' '));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


xDivisionByZerox (approved):

rota-rossi (approved):

Shinigami92:
@rota-rossi seems like all the other maintainers are currently unavailable.
So sorry if this will be fixed tomorrow, because could be that I'm already sleeping when they approve.
But as this is only a annoying warning message, it just can be ignored and it is not really urgent that it will be fixed immediately.
Maybe someone will also find other bugs that need a hotfix, so maybe it is good that we need to wait a bit :shrug:

ST-DDT (approved):

PR #904, opened 2022-05-02
fix(helpers): change missed references to arrayElements from random to helpers

rota-rossi:


Shinigami92:
Oh I'm sorry, I was a fraction of a second faster the you with #903 :laughing:

PR #908, opened 2022-05-02
test: ensure working examples and no console spam

ST-DDT:
May be merged after #907.
Prevents issues such as #902.

This PR adds tests ensuring, that every method
- has working examples
- and running these does not log anything, unless the method is deprecated

Can be tested by:
- removing or adding `@deprecated`
- replacing `faker.helpers.arrayElement` with `faker.random.arrayElement` (or any other deprecated method)
- removing or modifying examples


test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -0,0 +1,110 @@
+/**
```
Shinigami92, line 1: Using a JSDoc comment here may lead to documenting the `locales` var in line 15.
Maybe you wanted to use a block comment with `/*` instead of `/**`?

But I would prefer to bring this block comment below the imports, just personal preferences :slightly_smiling_face:

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+        (_, locale: string) => `faker.locale = '${locales[locale]}';\nfaker.`
+      );
+
+      expect(examples, `${moduleName}.${methodName} to have examples`).not.toBe(
+        ''
+      );
+
+      // Save examples to a file to run it
+      const dir = resolve(__dirname, 'temp', moduleName);
+      mkdirSync(dir, { recursive: true });
```
Shinigami92, line 89: Would we benefit in any way using the non sync functions with `await`?

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { mkdirSync, writeFileSync } from 'fs';
+import { resolve } from 'path';
```
Shinigami92, line 2:
```suggestion
import { mkdirSync, writeFileSync } from 'node:fs';
import { resolve } from 'node:path';
```

Shinigami92 (approved):
This is really nice 👌

import-brain (approved):

PR #909, opened 2022-05-03
refactor!: reorganize src folder

Shinigami92:


ST-DDT (approved):

ST-DDT (approved):

xDivisionByZerox:
While we at it: Should we provide a directory for each module?  Like address.ts would be located under `src/modules/address/address.ts` (or index.ts). That would allow us to scope additional requirements for this module in this directory, for example, image providers in the image directory.

xDivisionByZerox:
> While we at it: Should we provide a directory for each module? Like `address.ts` would be located in `src/modules/address/address.ts` (or index.ts). That would allow us to scope additional requirements for this module in this directory, for example `image-providers` in the `modules/image` directory.

This could be a bit overkill for most modules for now but should give more reliability in the future (perhaps, IDK TBH)

ST-DDT:
I don't think we should do it for all modules, but maybe for all that would benefit from it?
AFAICT there isn't much difference between import from `address.ts` and `address/index.ts` so it should be save to update/change later on.

xDivisionByZerox:
The difference would be that we would not lose file history again if we need to reorganize since all files are already in their own module directory.

Shinigami92:
To be fair, I had the same idea as @xDivisionByZerox 🤔 
Also maybe we would e.g. split every function into it's own file but let them in their "scopes" (in a far far away future)
It's just more future proof

Shinigami92:
I'm still thinking that a folder for each module could be a good idea due to now some modules are not "sorted" by alphabet because they are folders instead of files

image

pkuczynski (dismissed):

pkuczynski:
> I'm still thinking that a folder for each module could be a good idea due to now some modules are not "sorted" by alphabet because they are folders instead of files

It makes a lot of sense for me to put each of them into folders. Its more future proof and we could use folder to split some modules into smaller chunks, like for example moving into those folders some of the module-specific utils (for example for colors) or some of the const definitions (instead of keeping them in i18n).

What I also personally like and always do is to keep tests close to the code. So instead of having `test/address.spec.ts` we could have it at `modules/address/address.spec.ts`

Shinigami92:
> What I also personally like and always do is to keep tests close to the code. So instead of having `/test/address.spec.ts` we could have it at `/src/modules/address/address.spec.ts`

I use this pattern also at my work for a NestJS application. But if we want that, let's do it in a separate PR as it is not that important as the src files itself.

xDivisionByZerox:
Do we need to change the test directory at all? Otherwise the PR title could be confusing.

pkuczynski:
> I use this pattern also at my work for a NestJS application. But if we want that, let's do it in a separate PR as it is not that important as the src files itself.

Sure! It should be done in a separate PR. I was just giving a rationale why its worth to have subfolders.

pkuczynski (dismissed):

pkuczynski (approved):

ST-DDT (dismissed):

xDivisionByZerox (changes requested):
Do we need to change the test directory at all? Otherwise, the PR title could be confusing. So either revert the test changes from the current PR or change the title, please.

ST-DDT:
Yeah, moving the tests twice is kind of redundant.

Shinigami92:
I extracted the test restructure into its own issue: #911

ST-DDT (approved):

xDivisionByZerox (approved):

PR #910, opened 2022-05-03
chore: remove obsolete definition key constants

ST-DDT:
Since #822, we no longer need the internal definitions tree, so we should remove it.


pkuczynski (approved):

Shinigami92 (approved):

Shinigami92:
While working on # 909, I run `pnpm run generate:locales` and it throwed errors at me unrelated to # 909

PR #915, opened 2022-05-03
fix(generate:locale): make the definition types extendible

ST-DDT:
I messed up and removed code (#910), that was used by our scripts, so now I have to adjust the scripts to work without them.

We could revert that commit, but we want to get ride of it in the long run anyway.

I based this PR on #728 (#729) and adjusted the locale definitions to be `Partial` by default.
All usages were already partial and optional anyway:

https://github.com/faker-js/faker/blob/79884641bb89136158bd3ff75e78b89dde32a13c/src/definitions/definitions.ts#L45-L54

I recommend, to review only the first commit/changes to the script and definitions as the second commit is only the execution of the script.

~~I will try to create a separate PR,~~ that will check the scripts at least type wise (Issue #917).


Shinigami92 (approved):

scripts/generateLocales.ts
```diff
@@ -33,6 +32,37 @@ const pathDocsApiLocalization = resolve(
...
+  : Capitalize<S>;
+
+type DefinitionsType = {
+  [key in keyof Definitions]: PascalCase<`${key}Definitions`>;
+};
+
+/**
+ * The types of the definitions.
+ */
+const definitionsTypes: DefinitionsType = {
```
pkuczynski, line 47: This was not needed before. Having a need to define it manually defeats the purpose of having a generator script...

ST-DDT: This was required before as well but I (prematurely) removed it in #910.
(In fact it was vastly simplified in comparison to the previous version.)

pkuczynski: OK!

pkuczynski (approved):

PR #921, opened 2022-05-04
fix: empty array passed into helpers.arrayElements

c0per:


Shinigami92 (commented):
What happens if you pass an empty array but also pass a count greater than 0?
Could you add a second test for this case?

c0per:
> What happens if you pass an empty array but also pass a count greater than 0?
> Could you add a second test for this case?

Sure, added.

Shinigami92 (approved):

ST-DDT (approved):

PR #922, opened 2022-05-04
ci: check scripts

### LINKED ISSUE #917: ci: check whether scripts still compile

In #910 I removed code from the sources, that were used in the scripts.
Unfortunately, there isn't a build step that checks them with TS powers.

### Suggested solution

Check the types in the ci pipeline.

### Alternative

Run all scripts (#813), however that still wouldn't catch most type errors.

### Additional context

_No response_

Shinigami92:
closes #917


Shinigami92:
TS-Errors are reported successfully:

Shinigami92:
awaiting merge of #915

Shinigami92:
I don't know why but the results differs when I run
`pnpm run ts-check:scripts` vs `tsc --noEmit --esModuleInterop --skipLibCheck scripts/**/*.ts` manually 🤔

ST-DDT:
This looks like the perfect place for using [extends](https://www.typescriptlang.org/tsconfig#extends).

Shinigami92:
> This looks like the perfect place for using [extends](https://www.typescriptlang.org/tsconfig#extends).

I didn't use it for `tsconfig.json` as I'm to afraid to break something in build process.
Also I needed to reverse some settings explicit set for scripts and tests and therefore the file is nearly same length 🤷

ST-DDT:
#915 is merged.

package.json
```diff
@@ -110,6 +112,7 @@
     "eslint-plugin-jsdoc": "~39.2.9",
     "eslint-plugin-prettier": "~4.0.0",
     "esno": "~0.14.1",
+    "glob": "~8.0.1",
```
ST-DDT, line 115: Why did you add this one?

ST-DDT (approved):

package.json
Shinigami92: https://github.com/faker-js/faker/blob/b4cadfaae924218ae605c94f91e1206990d818d9/scripts/bundle.ts#L2

ST-DDT: That file was not edited by this PR.

Shinigami92: It was not explicitly in dependencies 
Now it checks the ts code and needs it as direct dependency 
Check the main branch and look into this file, you will see red squiggles but previously they could be ignored because one of our dependencies already depend on glob
Use `pnpm why glob` in main to find out which one

import-brain (approved):

PR #925, opened 2022-05-04
fix: typo in error message

azangru:
This PR fixes a typo in the preposition in an error message.

An example of what the error message reads currently:

```
Max 0 should be larger then min 2.
```

This erroneously uses the word [then](https://www.merriam-webster.com/dictionary/then) instead of the preposition [than](https://www.merriam-webster.com/dictionary/than) used for comparisons. The message should have read instead, `Max 0 should be larger than min 2.`

Might also want to consider changing _larger than_ to _greater than_.


ST-DDT:
> Might also want to consider changing _larger than_ to _greater than_.

👍

src/modules/datatype/index.ts
```diff
@@ -49,7 +49,7 @@ export class Datatype {
     }
 
     if (max < min) {
-      throw new FakerError(`Max ${max} should be larger then min ${min}.`);
+      throw new FakerError(`Max ${max} should be larger than min ${min}.`);
```
Shinigami92, line 52:
```suggestion
      throw new FakerError(`Max ${max} should be greater than min ${min}.`);
```

test/datatype.spec.ts
```diff
@@ -297,7 +297,7 @@ describe('datatype', () => {
 
           expect(() => {
             faker.datatype.number({ min, max });
-          }).toThrowError(`Max ${max} should be larger then min ${min}.`);
+          }).toThrowError(`Max ${max} should be larger than min ${min}.`);
```
Shinigami92, line 300:
```suggestion
          }).toThrowError(`Max ${max} should be greater than min ${min}.`);
```

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #932, opened 2022-05-07
feat(modules): export module interfaces

### LINKED ISSUE #778: Cannot directly import _Date

Shinigami92:
closes #778

### DIFF
--- a/src/faker.ts
+++ b/src/faker.ts
@@ -1,32 +1,32 @@
 import type { LocaleDefinition } from './definitions';
 import { FakerError } from './errors/faker-error';
 import type { KnownLocale } from './locales';
-import { Address } from './modules/address';
-import { Animal } from './modules/animal';
-import { Color } from './modules/color';
-import { Commerce } from './modules/commerce';
-import { Company } from './modules/company';
-import { Database } from './modules/database';
-import { Datatype } from './modules/datatype';
-import { _Date } from './modules/date';
-import { Fake } from './modules/fake';
-import { Finance } from './modules/finance';
-import { Git } from './modules/git';
-import { Hacker } from './modules/hacker';
-import { Helpers } from './modules/helpers';
-import { Image } from './modules/image';
-import { Internet } from './modules/internet';
-import { Lorem } from './modules/lorem';
-import { Mersenne } from './modules/mersenne';
-import { Music } from './modules/music';
-import { Name } from './modules/name';
-import { Phone } from './modules/phone';
-import { Random } from './modules/random';
-import { Science } from './modules/science';
-import { System } from './modules/system';
-import { Unique } from './modules/unique';
-import { Vehicle } from './modules/vehicle';
-import { Word } from './modules/word';
+import { AddressModule } from './modules/address';
+import { AnimalModule } from './modules/animal';
+import { ColorModule } from './modules/color';
+import { CommerceModule } from './modules/commerce';
+import { CompanyModule } from './modules/company';
+import { DatabaseModule } from './modules/database';
+import { DatatypeModule } from './modules/datatype';
+import { DateModule } from './modules/date';
+import { FakeModule } from './modules/fake';
+import { FinanceModule } from './modules/finance';
+import { GitModule } from './modules/git';
+import { HackerModule } from './modules/hacker';
+import { HelpersModule } from './modules/helpers';
+import { ImageModule } from './modules/image';
+import { InternetModule } from './modules/internet';
+import { LoremModule } from './modules/lorem';
+import { MersenneModule } from './modules/mersenne';
+import { MusicModule } from './modules/music';
+import { NameModule } from './modules/name';
+import { PhoneModule } from './modules/phone';
+import { RandomModule } from './modules/random';
+import { ScienceModule } from './modules/science';
+import { SystemModule } from './modules/system';
+import { UniqueModule } from './modules/unique';
+import { VehicleModule } from './modules/vehicle';
+import { WordModule } from './modules/word';
 import type { LiteralUnion } from './utils/types';
 
 export type UsableLocale = LiteralUnion<KnownLocale>;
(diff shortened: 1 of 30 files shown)


ST-DDT:
https://github.com/faker-js/faker/discussions/805#discussioncomment-2599628

Shinigami92:
@pkuczynski Could you go through this PR/code and reevaluate if you REALLY want to have the modules renamed not to suffixed with `Module`? I still don't think your initial thought that it could conflict with node's modules is valid in any case.
Beside that, I personally find it confusing to name modules everywhere modules but then the module itself is not named module but faker were it actually is not e.g. a sub-instance of Faker.
It can be even more confusing if we have later a `MinimalFaker`, `BaseFaker`, `Faker`. (See https://github.com/faker-js/faker/discussions/805#discussioncomment-2690233)

And then we would even have a `FakeFaker` 👀 confusing like hell 🤷

ST-DDT:
> And then we would even have a `FakeFaker` 👀 confusing like hell 🤷

No, because fake -> helpers.fake

src/faker.ts
```diff
@@ -1,32 +1,32 @@
...
-import { Music } from './modules/music';
-import { Name } from './modules/name';
-import { Phone } from './modules/phone';
-import { Random } from './modules/random';
-import { Science } from './modules/science';
-import { System } from './modules/system';
-import { Unique } from './modules/unique';
-import { Vehicle } from './modules/vehicle';
-import { Word } from './modules/word';
+import { AddressModule } from './modules/address';
```
pkuczynski, line 4: I still think they should be called `XyzFaker`, as when they imported it would be clear what they are and would never conflict with any other package. For me `AdressModule` is too general. So either `FakerAddress`, `FakeAddress` or `AddressFaker`. All equally good... Probably `FakeAddress` being my favorite. `FakeAdress.streetName` reads very nicely.

I am also under impression @ST-DDT thinks the same?

pkuczynski (changes requested):

src/faker.ts
Shinigami92: FakeAddress/FakerAddress is new to me 🤔 
This might solve the problem of not getting confused with something like LocalizedFaker, BaseFaker or CoreFaker which is part of the darfted roadmap you have skipped in the last meeting(s)

Also I saw what @xDivisionByZerox did in the number and string modules, so I hoped that the discussion just came to an end and we silently just accepted that XyzModule as a name for a module in faker is just okay

pkuczynski: Sorry for skipping the last 2 meetings. Not always easy to balance work, personal duties and open source.

`FakeXyz` is a new thing which came to my mind just today. But it feels like it saves the purpose very well.  Let's see what others think... :)

ST-DDT: I like the Fake(r) prefix/suffixes way more than the generic Module.
I like XyzFaker the most because it puts the emphasis on it faking stuff.
Whereas Fake(r)Xyz sounds like a Faker/fake variant of Xyz. This might be a German implicit thing. It would be nice if a English native speaker could (in-)validate my impression on that.

For the same reason and to avoid naming conflicts I would name the Faker classes FakerCore/FakerBase...

There is one argument for FakerXyz and that is IDE autocompletion type Faker and the you get all variants (if you are lucky).

Please keep in mind that the later (not v8) treeshaking refactor will remove the classes in favor of factory functions. So I would like to get a quick decision on this. Ibwill open a poll later when Ibhave some spare time. EOV Thursday during the team meeting.

xDivisionByZerox: I like XyzFaker, for the same reasons @ST-DDT just stated. I just used XyzModule as a working title in the string and number module. 

Just leaving this here, so you have my opinion.

xDivisionByZerox:
Would it be possible to integrate this change as a non-breaking change? Maybe with some proxy magic and deprecation of the current classes (in the constructor)?

src/faker.ts
Shinigami92: > Please keep in mind that the later (not v8) treeshaking refactor will remove the classes in favor of factory functions.

Please keep in mind that this will **NOT** solve the problem! We will not rename the module interfaces back to e.g. just `Date` because also interfaces would conflict with native JS constructors.
You cannot write code like:
```ts
function myFun(fakerDateInstance: Date) {
//                                ^^^^ this is from Faker
	const recent: Date = fakerDateInstance.recent();
//                ^^^^ This is native JS/TS
}
```

So we need a permanent final solution for the entire future of FakerJS with no turning back!

ST-DDT:
> Would it be possible to integrate this change as a non-breaking change? Maybe with some proxy magic and deprecation of the current classes (in the constructor)?

The classes have never been part of the API, so not sure if that is worth it.

src/faker.ts
ST-DDT: https://github.com/faker-js/faker/discussions/1291

import-brain: > I like the Fake(r) prefix/suffixes way more than the generic Module. I like XyzFaker the most because it puts the emphasis on it faking stuff. Whereas Fake(r)Xyz sounds like a Faker/fake variant of Xyz. This might be a German implicit thing. It would be nice if a English native speaker could (in-)validate my impression on that.
> 
> For the same reason and to avoid naming conflicts I would name the Faker classes FakerCore/FakerBase...
> 
> There is one argument for FakerXyz and that is IDE autocompletion type Faker and the you get all variants (if you are lucky).
> 
> Please keep in mind that the later (not v8) treeshaking refactor will remove the classes in favor of factory functions. So I would like to get a quick decision on this. I will open a poll later when I have some spare time. EOV Thursday during the team meeting.

I would also prefer XyzFaker.

Shinigami92: Current voting says 5 of 6 voters prefer suffixing with `Module`
Also Damien got some independent arguments for using this and why the other ideas are not that good
So for now I will mark this as resolved

ST-DDT (approved):

ST-DDT:
**Team decision**

We will use `XyzModule`.

pkuczynski (approved):

xDivisionByZerox (approved):

import-brain (approved):

PR #933, opened 2022-05-07
refactor: rename helpers to helper

Shinigami92:
todos:
- [ ] needs test for deprecation message


ST-DDT:
I like `helpers` more than `helper`.
If you want to use singular names everywhere, then maybe use `util`? (It is as generic as `helpers`)

Shinigami92:
> I like `helpers` more than `helper`. If you want to use singular names everywhere, then maybe use `util`? (It is as generic as `helpers`)

I had the exact same thought! 🤔 
Let's ask other members or put this on meeting notes.

pkuczynski:
I don't like both names :) We should keep the whole thing private and used only internally, move things like `replaceCreditCardSymbols` closer to the usage and maybe expose some (`arrayElement`) in `faker.random`... Faker is meant to generate data and not provide utilities to replaces characters in strings - so those utility functions do not belong to it I think...

pkuczynski:
> Do you have an alternative? Is util at least better or worse than helper(s)?

Don't really see much different between those 2 (3) to be honest.

> If you have a specific suggestion, consider opening an issue/PR or document your idea here:
https://github.com/faker-js/faker/discussions/805

Sure! Done here: https://github.com/faker-js/faker/discussions/805#discussioncomment-2717280 - once it will get our consensus, I can work on PRs :) It does not make sense to do that if we won't agree on it...

> However random is once again really generic. I think util is the better name for those though.
> (Assuming the original random will be split and purged eventually.)

I think `random` is a good name and can be described as picking up a random value from collection (see my proposal of `random.element` or `random.order` - that reads very well).

Stuff currently in `random` should go away to more dedicated modules, as those methods _generate_ data, while `random` could be only _picking_ data from given collection.

> then we should retain and consider keeping them public

I listed in #805 what I think make sense to keep public and what not. Lets discuss this over there...

As of now renaming `helpers` to `helper` or `util` brings no major value from my pov...

Shinigami92:
As in v6 we currently marked things like random.arrayElement to be moved to helpers.arrayElement, I would suggest we will make the renaming / moval of helpers in v7->v8
Additionally we should write a complete and final solution down what we want to move where and make final decisions so we do not move stuff again and again

Shinigami92:
This got stale

PR #948, opened 2022-05-12
feat(address): use localized fake pattern in city

ST-DDT:
This PR changes the `address.city` method to use the `address.city` locale data with fake to generate city names.

For that I checked, that all city files are named correctly (city -> patterns; city_name -> actual city names).
I did not check whether the specified patterns make sense or the city names are real.

I generated the missing city files with the original pattern from the method to ensure it does not break between versions.
I did not add `'{{address.city_name}}'` to the city patterns if both the city and city_name files exist (except for the en locale because it affected a test, see my second commit).
I did not remove it from the patterns if the city_name file is missing in that locale.

Everything in the first commit was generated by using this code:

<details>
<summary>generateLocales.ts</summary>

````ts

const processed = new Set<string>();

/**
 * Use this hook method to selectively update locale data files (not for index.ts files).
 * This method is intended to be temporarily overwritten for one-time updates.
 *
 * @param filePath The full file path to the file.
 * @param locale The locale for that file.
 * @param localePath The locale path parts (after the locale).
 */
function updateLocaleFileHook(
  filePath: string,
  locale: string,
  localePath: string[]
): void {
  if (processed.has(locale)) {
    return;
  }
  if (localePath[0] !== 'address') {
    return;
  }
  processed.add(locale);

  let cityPattern: string[];
  let cityNames: string[];

  const cityFiles = ['city', 'city_name', 'city_root'];
  for (const name of cityFiles) {
    if (existsSync(`src/locales/${locale}/address/${name}.ts`)) {
      let content: string[] =
        require(`../src/locales/${locale}/address/${name}`).default;
      unlinkSync(resolve(`src/locales/${locale}/address/${name}.ts`));
      if (content[0].includes('{{')) {
        for (const replaced of cityFiles) {
          content = content.map((e) =>
            e.replace(
              new RegExp(`{{address.${replaced}}}`, 'g'),
              '{{address.city_name}}'
            )
          );
        }

        cityPattern = content;
      } else {
        cityNames = content;
      }
    }
  }

  if (cityPattern == null) {
    if (cityNames == null) {
      cityPattern = [
        '{{address.cityPrefix}} {{name.firstName}}{{address.citySuffix}}',
        '{{address.cityPrefix}} {{name.firstName}}',
        '{{name.firstName}}{{address.citySuffix}}',
        '{{name.lastName}}{{address.citySuffix}}',
      ];
    } else {
      cityPattern = [
        '{{address.cityPrefix}} {{name.firstName}}{{address.citySuffix}}',
        '{{address.cityPrefix}} {{name.firstName}}',
        '{{name.firstName}}{{address.citySuffix}}',
        '{{name.lastName}}{{address.citySuffix}}',
        '{{address.city_name}}',
      ];
    }
  }

  let cityPatternContent = `export default ${JSON.stringify(
    cityPattern,
    undefined,
    '  '
  )};\n`;
  cityPatternContent = formatTypescript(cityPatternContent);
  writeFileSync(
    resolve(`src/locales/${locale}/address/city.ts`),
    cityPatternContent
  );

  if (cityNames != null) {
    let cityNamesContent = `export default ${JSON.stringify(
      cityNames,
      undefined,
      '  '
    )};\n`;
    cityNamesContent = formatTypescript(cityNamesContent);

    writeFileSync(
      resolve(`src/locales/${locale}/address/city_name.ts`),
      cityNamesContent
    );
  }
}
````

</details>

And executed it using this helper:

````sh
untilSuccess() {
        until "$@"; do :; done
}

untilSuccess pnpm run generate:locales
pnpm run generate:locales
````


pkuczynski:
What's the value of having both methods `city` and `cityName`? As a first time user I would feel a bit confused...

ST-DDT:
For now, mainly legacy reasons, but for some usecases it might be required to have actual city names instead of generated ones.
Maybe we could (later) merge them together using a `generated: boolean = true` flag.

ST-DDT:
(I do plan to deprecate cityPrefix and citySuffix at some point though)

pkuczynski:
Following the comment in the other PR we both agreed that it would be better to have data close to reality (zodiac I think) so I think `city` should always return real city names if we have localization data and only generated ones when we don't...

ST-DDT:
> Following the comment in the other PR we both agreed that it would be better to have data close to reality (zodiac I think) so I think `city` should always return real city names if we have localization data and only generated ones when we don't...

That sounds like a reasonable suggestion. I would like to do that in a separate PR though, as these mass locale changes get confusing/unreviewable really fast (and I would have to purge the method references from all locale files).

pkuczynski:
> That sounds like a reasonable suggestion. I would like to do that in a separate PR though, as these mass locale changes get confusing/unreviewable really fast.

Make sense :)

> Should I also remove the city (pattern) files, if there are actual city name files?

Yes. And deprecate `cityName`...

pkuczynski (approved):

Shinigami92 (approved):

PR #724, opened 2022-03-29
fix: Updated src/phone.ts

kum9748ar:
## renamed functions and changed warnings 
- updated `phone.phoneFormat` to `format`, `phone.phoneNumber` to `number`   
- removed to-dos 
- formatted code using `npm format`
- fixed minor issues 
Ref: https://github.com/faker-js/faker/issues/694 , https://github.com/faker-js/faker/issues/693

PR #958, opened 2022-05-15
chore: deprecate phone.phoneFormats

ST-DDT:
Supersedes #694 and partially #724

The `phoneFormats` gives the user access to an underlying phone format, but there is no real usecase where the user might want to use that. Potentially encouraging the user to use our formats in their example patterns without actually copying them over, potentially causing issues if we change the phone patterns.

<details>
<summary>Docs preview</summary>

grafik

</details>


Shinigami92 (approved):

src/modules/phone/index.ts
```diff
@@ -51,11 +55,21 @@ export class Phone {
...
    * @example
    * faker.phone.phoneFormats() // '!##.!##.####'
+   *
+   * @deprecated
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `format()`
   phoneFormats(): string {
+    deprecated({
+      deprecated: 'faker.phone.phoneFormats',
+      proposed: 'faker.phone.phoneNumber',
```
Shinigami92, line 69: Previously we used `()` like:
```suggestion
      deprecated: 'faker.phone.phoneFormats()',
      proposed: 'faker.phone.phoneNumber()',
```
I'm fine with both :shrug: just pops up in my mind

src/modules/phone/index.ts
```diff
@@ -51,11 +55,21 @@ export class Phone {
...
+   * @deprecated
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `format()`
   phoneFormats(): string {
+    deprecated({
+      deprecated: 'faker.phone.phoneFormats()',
+      proposed: 'faker.phone.phoneNumber()',
+      since: 'v7.0',
+      until: 'v8.0',
+    });
```
Shinigami92, line 72: Should we add tests for these, or is the _exampleAndDeprecations_ enough for all future deprecations?

ST-DDT: IMO its enough.

Shinigami92 (approved):

xDivisionByZerox (approved):

import-brain (approved):

PR #959, opened 2022-05-15
docs: add hint that the generated data might be real ones

ST-DDT:
We should warn our users that the generated data may be working ones and not the well known `555` dummy ones.


Shinigami92:
Random thought: why not for:
- email
- iban
- vrm
- vim
- bitcoin address
- ipv4/6
- ...

I'm not sure if this hint is helpful :thinking:

ST-DDT:
We might want to add a general hint at some point (or maybe actually generate fake numbers).

But I created this one to address the issue raised in Discord: 
https://discord.com/channels/929487054990110771/929544565348777984/974285076064002148

Shinigami92:
> We might want to add a general hint at some point (or maybe actually generate fake numbers).
> 
> But I created this one to address an issue raised in Discord: [discord.com/channels/929487054990110771/929544565348777984/974285076064002148](https://discord.com/channels/929487054990110771/929544565348777984/974285076064002148)

IMO we should just add the hint now globally. For me it doesn't make really sense to add it here specifically just because one person asked in discord if values are real or not.
Some day just the next person would come up and ask for another function and so on :shrug: 
I also would say it is not in scope for Faker at all to generate or prevent real data. Just data that "looks" real and if they are "accidentally" real, so they are.

prisis (dismissed):

pkuczynski:
I agree with @Shinigami92 that it should be done as a general hint and not for a specific function. The whole concept of faker is to generate data that looks real, so as he pointed out, they might be actually real!

ST-DDT:
Yeah, I will think of a way to put it somewhere else.

README.md
```diff
@@ -30,6 +30,10 @@
 - 🧍 Names - Generate virtual humans with a complete online and offline identity.
 - 🔢 Numbers - Of course, we can also generate random numbers and strings.
 
+> **Note**: Faker tries to generate life like data and not specifically test data.
+> So the generated names, addresses, emails, phone numbers or other data might be existing and coincidentally valid information.
```
ST-DDT, line 34: Not sure about this sentence.
I hope the intention is clear, but the wording is probably not the best.

import-brain: What about "Faker tries to generate realistic data and not obvious fake data. The generated names, addresses, emails, phone numbers, and/or other data might be coincidentally valid information."?

Shinigami92 (approved):

ejcheng (approved):

PR #960, opened 2022-05-15
chore(deps): update esno

Shinigami92:
This will auto-close some other current open deps PR


Shinigami92:
https://github.com/faker-js/faker/blob/f296cb9e38a585778c479624245ab9a812875585/scripts/apidoc/utils.ts#L5

image

@antfu is this a bug in `esno`? We explicitly import the `cjs` file instead of `cts` :thinking: 
Also `cts` is not yet supported by prettier as far as I know, so we cannot just update the file extension.

privatenumber:
Thanks for the report @Shinigami92 @antfu

I fixed the `.cjs` resolution error, but there's another bug that will come up:

```
/faker-js/faker/scripts/apidoc/utils.ts:15
const pathRoot = resolve(__dirname, '..', '..');
                         ^

ReferenceError: __dirname is not defined in ES module scope
    at <anonymous> (/faker-js/faker/scripts/apidoc/utils.ts:15:26)
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)
    at async loadESM (node:internal/process/esm_loader:85:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

I've actually been working on fixing it and it's almost done: https://github.com/esbuild-kit/tsx/pull/5

Hope to have it released by tomorrow or day after. 🤞

Shinigami92:
@ST-DDT I fixed usages of `__dirname`
Is it correct that I get logs like `Failed to load /home/shinigami/OpenSource/faker/src/locales/<locale>. Attempting manual parse instead...`?
I'm not sure if something broke in the `generate:locales` script

ST-DDT:
> Is it correct that I get logs like `Failed to load`

If you run the command again, then it should be gone. If not, we have to check why it fails.

ST-DDT:
Maybe log the __dirname then I can tell you if its the right value.

ST-DDT (dismissed):

privatenumber:
Sounds like ya'll figured out a work around but just want to update that [tsx@3.0.0](https://github.com/esbuild-kit/tsx/releases/tag/v3.0.0) fixes the issue too.

Shinigami92:
Let's wait on the update and see if we need to rename the files to `mts`

Shinigami92:
@privatenumber yeah, with the newest update I needed to revert all changes due to I would need to rewrite everything to `mts` files
This is more problematic as just revert the changes, so I did that instead
Seems now everything works as before, thx :+1:

PR #979, opened 2022-05-21
fix: type exports for module NodeNext, Node16

andrew-w-ross:
When using typescript 4.7rc with module resolution `NodeNext` or `Node16` typescript will report `Could not find a declaration file for module  @faker-js/faker`.  I assume this is because the fallback property `types` is pointed set to `index.d.ts` not  `dist/index.d.ts` and I guess typesVersions is skipped in those module modes.


andrew-w-ross:
Just to note, I didn't change the `types` to `dist/types/index.d.ts` as I wasn't sure if that was intentional.

Shinigami92:
> I assume this is because the fallback property `types` is pointed set to `index.d.ts` not `dist/index.d.ts`

You assume? Or you tested it? You can use the playground project to test it: https://github.com/faker-js/playground

andrew-w-ross:
@Shinigami92 Tested locally as it's an issue with my project currently and playground isn't all that helpful for checking resolutions.

Trace resolution without any changes:

```bash
tsc --traceResolution | grep faker
======== Module name '@faker-js/faker' was successfully resolved to '/Users/andrewross/Projects/faker/dist/esm/index.mjs' with Package ID '@faker-js/faker/dist/esm/index.mjs@6.3.1'. ========
src/index.ts(1,23): error TS7016: Could not find a declaration file for module '@faker-js/faker'
```

And with my change:

```bash
tsc --traceResolution | grep faker
======== Resolving module '@faker-js/faker' from '/Users/andrewross/Projects/foo/src/index.ts'. ========
Loading module '@faker-js/faker' from 'node_modules' folder, target file type 'TypeScript'.
Scoped package detected, looking in 'faker-js__faker'
Scoped package detected, looking in 'faker-js__faker'
Scoped package detected, looking in 'faker-js__faker'
Found 'package.json' at '/Users/andrewross/Projects/foo/node_modules/@faker-js/faker/package.json'.
File '/users/andrewross/projects/foo/node_modules/@faker-js/faker/dist/types/index.d.ts' exist - use it as a name resolution result.
Resolving real path for '/users/andrewross/projects/foo/node_modules/@faker-js/faker/dist/types/index.d.ts', result '/Users/andrewross/Projects/faker/dist/types/index.d.ts'.
======== Module name '@faker-js/faker' was successfully resolved to '/Users/andrewross/Projects/faker/dist/types/index.d.ts' with Package ID '@faker-js/faker/dist/types/index.d.ts@6.3.1'. ========
```

Trace resolution only with types set to  `./dist/types/index.d.ts` surprisingly seems to have no affect.

Shinigami92 (commented):
I currently question myself if https://github.com/andrew-w-ross/faker/blob/f26464e4fda548f90238a81f12491c196f6d7ae7/package.json#L26 should be `./dist/types/index.d.ts` :thinking: 
Do we even have a root `index.d.ts`?

andrew-w-ross:
@Shinigami92 You don't and by the looks of it but you're only going to run into it if you someone uses typescript version < 4 and is that something you really care to support?

Shinigami92 (approved):

ST-DDT (approved):

import-brain (approved):

PR #980, opened 2022-05-21
fix: Luhn generation algorithms and tests

### LINKED ISSUE #952: Luhn algorithm generation not working
### Describe the bug

The functions using the Luhn algorithm to generate numbers (`finance.creditCardNumber`, `phone.imei`) don't generate Luhn-valid numbers, however, they still pass the tests in `luhnCheck.ts`. On digging deeper, I found that both of those functions use `helpers.replaceCreditCardSymbols`, which has a function inside of it to help calculate the Luhn checksum. I suspect that function is faulty.

### Reproduction

Screen Shot 2022-05-14 at 12 51 43 PM
Screen Shot 2022-05-14 at 12 51 55 PM

Numbers were checked on multiple Luhn validation sites, none of them said it worked.

### Additional Info

_No response_

import-brain:
fixes #952


import-brain:
@prisis Seems like the new luhnCheck function I pasted in causes tests to fail...👀

https://github.com/faker-js/faker/runs/6537655385?check_suite_focus=true

import-brain:
Just tested the function: https://replit.com/@ericcheng2005/luhnTest#index.js

Made sure the testcases I used were right too:
Screen Shot 2022-05-21 at 12 11 42 PM

ST-DDT:
Probably because the luhn value we are generating is broken too.
Maybe we should change that.
I created https://github.com/faker-js/faker/pull/981 to improve the error messages for the failures.

import-brain:
> Probably because the luhn value we are generating is broken too. Maybe we should change that. I created #981 to improve the error messages for the failures.

Yeah, something's wrong with `helpers.replaceCreditCardSymbols`.

import-brain:
Note: all of the numbers I changed the seeded tests to expect are Luhn-valid numbers, verified using https://simplycalc.com/luhn-validate.php

ST-DDT (approved):

prisis (approved):
Good job :)

Shinigami92 (approved):

PR #986, opened 2022-05-23
feat: added sponsor and community members to the readme

prisis:


Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

BACKERS.md
```diff
@@ -0,0 +1,11 @@
+<h1 align="center">Sponsors &amp; Backers</h1>
+
+Faker is an MIT-licensed open source project with its ongoing development made possible entirely by the support of the awesome sponsors and backers listed in this file. If you'd like to join them, please consider [sponsor Faker development](https://opencollective.com/fakerjs).
+
+### Sponsors
```
ST-DDT, line 5: Is there a way to avoid the duplicate here?

README.md
```diff
@@ -195,8 +195,22 @@ const secondRandom = faker.datatype.number();
 console.log(firstRandom === secondRandom);
 ```
 
+## Sponsors
+
+Faker is an MIT-licensed open source project with its ongoing development made possible entirely by the support of these awesome backers
+
+### Sponsors
```
ST-DDT, line 202: Is there a way to avoid the duplicate here?

README.md
```diff
@@ -244,6 +258,20 @@ pnpm run docs:serve # Serve docs from /dist
 
 The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
 
+### Thank you
+
+Thank you to all the people who already contributed to Faker!
+
+<a href="https://github.com/faker-js/faker/graphs/contributors"><img src="https://opencollective.com/fakerjs/contributors.svg?width=890" /></a>
```
ST-DDT, line 265: Not sure whether the width is "correct".
If you open this as a file then, it will be shown with a wider display: https://github.com/faker-js/faker/blob/b8dfab0be850668e433e503c8b0a7561dc721114/README.md (890×262)
If you display it as a Readme, then it will be shown slightly smaller: https://github.com/faker-js/faker/tree/b8dfab0be850668e433e503c8b0a7561dc721114 (830×244.333)

Maybe a designer with a keen eye should check whether the images get blurred/messed up.
(I don't care)

ST-DDT: Maybe just use 800, then the images aren't as small.
Or is the 890 some recommended value?

README.md
prisis: Can write a script to add it to both files  
  
Sorry i dont have a better idea, and i think that this will not be touched so much at all

Shinigami92: > Can write a script to add it to both files
> 
> Sorry i dont have a better idea, and i think that this will not be touched so much at all

I'm with @prisis, I assume these will not be touched as the images are auto-updating

Shinigami92: That was really not clear in the first place 😅 
I also thought that you mean that the content in both files are kinda the same

prisis: What you think if we let it be like this, for now. And if we agree on the new tiers we adjust this list based on the new tiers?

Shinigami92 (approved):

README.md
ST-DDT: Sounds good!

ST-DDT (approved):

PR #1001, opened 2022-05-25
feat: adding a browser build

prisis:
This is a test to add a browser build with iife, still need to test it


Shinigami92:
This would add additional ~4-5 MB to the package size 👀 
We should highly ask ourself if that is worth it 😬 
I even would like to get rid of the cjs version one day...

Each of these bundles has its own locales and the locales are around 4 MB in size.
If we found a way where every bundle can reuse one bundled locales that would be much better and reduce the size dramatically.

prisis:
How about we move the locales out into packages and throw a error if you want to use, lets say de like "You trying to use the locale de, please install @faker-js/faker-de" with this the core package would be pretty small

Shinigami92:
Closing this as not planned
Please use `script type=module` and follow the docs
Old browsers not supporting `type=module` are not supported, use faker `5.5.3` for this

PR #1010, opened 2022-05-26
fix: patched logic that was incompatible with webpack

chutchi2:
- Updated `src/modules/name/index.ts` to use a ternary operation (`a ? a : b`) instead of the webpack incompatible (`a ?? b`).

### DIFF
--- a/src/modules/name/index.ts
+++ b/src/modules/name/index.ts
@@ -152,8 +152,9 @@ export class Name {
     let prefix = '';
     let suffix = '';
 
-    const normalizedGender: GenderType =
-      gender ?? this.faker.helpers.arrayElement(['female', 'male']);
+    const normalizedGender: GenderType = gender
+      ? gender
+      : this.faker.helpers.arrayElement(['female', 'male']);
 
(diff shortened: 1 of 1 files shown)


chutchi2:
I only did this because the 7.1.0 release notes indicated the webpack comp error, and i tried it for myself before bringing it into my project and sure enough there was the comp error. I suppose I may have fixed it in a way that just happened to fix the compilation error, but it seemed weird to me that what i fixed was actually a fix. Considering that this syntax is used all over the project. My own project does use es2020 as it's target so otherwise it is a bit strange this issue exists.

chutchi2:
the issue actually seems to be pretty prominent for webpack and all of the `modules` directory. I'll close this because I'm wondering why this has suddenly popped up.

ST-DDT:
IMO this is a webpack framework issue and not a faker issue.
All currently supported node versions support this too.
Please check, whether there is an updated version of webpack available, that supports this.
I'm not even sure whether our build process produces the given syntax to achieve certain optimizations.

PR #543, opened 2022-02-22
chore: deprecate helpers.*Card methods

sprohaszka:
Related to issue #528 

Deprecate the following methods:
* helpers.userCard
* helpers.contextualCard
* helpers.createCard

Replace example in README.

Issue #528, opened 2022-02-20
Deprecate helpers.userCard/contextualCard/createCard


ST-DDT:
These methods place a maintenance burden on this library while providing little benefit to the users.
Users that need a specific data object usually require custom data structures that can be easily defined in a self-managed method.
We should only provide methods for standardized data structures directly.

sprohaszka:
Hello @ST-DDT.
To be sure to understand, you only want to add `deprecated` JSDoc flag?

ST-DDT:
The deprecation consists of two parts:
`@deprecated` jsdocs and a console.log warning.

Example: https://github.com/faker-js/faker/blob/03c3d1156312153428f2db5649b24dd5fc046d43/src/random.ts#L57-L64

sprohaszka:
Tell me you're ok with the wording.

Shinigami92:
Was not linked to the PR, already solved

PR #1013, opened 2022-05-27
chore: remove helpers.userCard() from README.md

andrecastelo:
Related to https://github.com/faker-js/faker/issues/528 and https://github.com/faker-js/faker/pull/543. Since the `helpers.*Card` methods were deprecated we need to remove the example from `README.md`.


Shinigami92:
I also had this in mind, but we may want to replace it with e.g. the `arrayElement` function, so at least there is one example.

Alternatively we could remove the whole section and just refer to the docs :thinking:

andrecastelo:
> I also had this in mind, but we may want to replace it with e.g. the `arrayElement` function, so at least there is one example.
> 
> Alternatively we could remove the whole section and just refer to the docs 🤔

True, true. IMO it is better if we leave the example there - it helps with discoverability, and then the user can look into the docs for more helper methods. Also the docs link is not very prominent (I think would be nice if the link was moved to the header, next to the stackblitz link).

andrecastelo:
Added `faker.helpers.arrayElement()` example, and added a link to the [Documentation](https://fakerjs.dev/guide/) at the top of `README.md`

ejcheng (dismissed):

Shinigami92 (approved):

ST-DDT (approved):

PR #1014, opened 2022-05-27
feat: science module

### LINKED ISSUE #944: Science module

I think it would be a good idea to add a science module to Faker to generate random scientific data. See below for examples.

### Suggested solution

Create a new module called science

`faker.science.elementName() // "Lithium"`
`faker.science.elementSymbol() // "F"`
`faker.science.siUnit() // "kg"`
`faker.science.siUnit({ long: true, plural: false }) // "meter"`

### Alternative

_No response_

### Additional context

_No response_

ejcheng:
closes #944


src/definitions/science.ts
```diff
@@ -0,0 +1,33 @@
+import type { LocaleEntry } from './definitions';
+
+/**
+ * The possible definitions related to science.
+ */
+export type ScienceDefinitions = LocaleEntry<{
+  /**
+   * Some science units (short).
+   */
+  shortUnit: string[];
```
ST-DDT, line 10: Maybe unitAbbr, as we already use that term elsewhere.

src/definitions/science.ts
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Some science units (long).
+   */
+  longUnit: string[];
+  /**
+   * Some element informtion.
+   */
+  element: Element[];
```
ST-DDT, line 18: I think we should explicitly refer to chemical elements. As `element` is kind of generic.

src/modules/science/index.ts
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param options The options to use. Defaults to `{ long: true }`.
+   * @param options.long Whether the long or short version of the unit should be returned. Defaults to `true`.
+   *
+   * @example
+   * faker.science.unit() // 'second'
+   * faker.science.unit({ long: false }) // 'Pa'
+   * faker.science.unit({ long: true }) // 'joule'
+   */
+  unit(options?: { long?: boolean }): string {
```
ST-DDT, line 52:
```suggestion
  unit(options: { long?: boolean } = {}): string {
```

src/modules/science/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    const { long = true } = options || {};
+    options = { long };
```
ST-DDT, line 54:
```suggestion
```

src/modules/science/index.ts
```diff
@@ -0,0 +1,65 @@
...
+   * @param options The options to use. Defaults to `{ long: true }`.
+   * @param options.long Whether the long or short version of the unit should be returned. Defaults to `true`.
+   *
+   * @example
+   * faker.science.unit() // 'second'
+   * faker.science.unit({ long: false }) // 'Pa'
+   * faker.science.unit({ long: true }) // 'joule'
+   */
+  unit(options?: { long?: boolean }): string {
+    const { long = true } = options || {};
```
ST-DDT, line 53:
```suggestion
    const { long = true } = options;
```

ST-DDT (changes requested):

src/definitions/science.ts
Shinigami92: How about structuring it like `unit: { full, abbr }`? That way, the data can be coupled.
SO you could use 

```ts
const { abbr, full } = science.unit()
```

ST-DDT: That doesn't match up with our other methods like date.month and weekday.
And I dont see a usecase for it either.

Shinigami92: The use-case would be that you just want one unit and show it in a table or whatever
So you have e.g. `Take 200 g (gram) of salt`

Shinigami92: I may want to design the httpStatus call in a same way
But you are right as currently we don't have that yet (maybe currency?)
I wont gave a veto if it gets merged as is, but it would be a bit better UX in my eyes :shrug:

src/definitions/science.ts
```diff
@@ -13,9 +13,9 @@ export type ScienceDefinitions = LocaleEntry<{
    */
   longUnit: string[];
   /**
-   * Some element informtion.
+   * Some chemical element informtion.
    */
-  element: Element[];
+  chemicalElement: Element[];
```
pkuczynski, line 18: Is it worth changing the name? Do we have other `element` function we would like to develop? If we make the change, we should also change `Element` type to `ChemicalElement`, but I doubt the value in this change tbh.

src/definitions/science.ts
pkuczynski: I agree with @Shinigami92 it would be better. Gave the same feedback to `element` in the #944. Otherwise you could not get both short and long name for the unit. And if user wants only one, he could do: `faker.unit().short`

src/modules/science/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  /**
+   * Returns a random element name.
+   *
+   * @example
+   * faker.science.elementName() // 'Praseodymium'
+   */
+  elementName(): string {
```
pkuczynski, line 23: Haven't you agreed in #944 to implement it as `faker.element().name`?

src/modules/science/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    ).name;
+  }
+
+  /**
+   * Returns a random element symbol.
+   *
+   * @example
+   * faker.science.elementSymbol() // 'Rg'
+   */
+  elementSymbol(): string {
```
pkuczynski, line 35: `faker.element().symbol`

src/modules/science/index.ts
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * @param options The options to use. Defaults to `{ long: true }`.
+   * @param options.long Whether the long or short version of the unit should be returned. Defaults to `true`.
+   *
+   * @example
+   * faker.science.unit() // 'second'
+   * faker.science.unit({ long: false }) // 'Pa'
+   * faker.science.unit({ long: true }) // 'joule'
+   */
+  unit(options: { long?: boolean } = {}): string {
```
pkuczynski, line 52: `faker.unit().long`, `faker.unit().short`

pkuczynski (changes requested):

src/definitions/science.ts
Shinigami92: IMO yes, we should use `chemicalElement` and `ChemicalElement`

src/modules/science/index.ts
import-brain: *facepalm* Whoops, I forgot

src/definitions/science.ts
pkuczynski: In science? What other elements in science there are?

src/definitions/science.ts
pkuczynski: I am wondering if we should not use just `element` as what other elements there might be in the science module? Its even called `periodic table of elements` isn't it?

import-brain: > I am wondering if we should not use just `element` as what other elements there might be in the science module? Its even called `periodic table of elements` isn't it?

Yeah, I can't recall any other thing commonly referred to as an "element" that's widely-used in science.

pkuczynski: Me neither, that's why I would keep it as discussed initially: `element`. @ST-DDT @Shinigami92 wdyt?

src/modules/science/index.ts
```diff
@@ -0,0 +1,58 @@
...
+   * faker.science.element() // { name: 'Hydrogen', symbol: 'H' }
+   * faker.science.element() // { name: 'Xenon', symbol: 'Xe' }
+   * faker.science.element() // { name: 'Cerium', symbol: 'Ce' }
+   */
+  element(): Element {
+    const index = this.faker.datatype.number({
+      min: 0,
+      max: this.faker.definitions.science.chemicalElement.length - 1,
+    });
+    return this.faker.definitions.science.chemicalElement[index];
```
ST-DDT, line 40: Why don't you use `helpers.arrayElement`?
```suggestion
    return this.faker.helpers.arrayElement(this.faker.definitions.science.chemicalElement);
```

src/definitions/science.ts
ST-DDT: I agree with you, that's it's the most common one in science.
But it's not the only one: https://en.wikipedia.org/wiki/Electrical_element
Even wikipedia refers to it as https://en.wikipedia.org/wiki/Chemical_element

Similar to `statusCode`s in the internet referring mostly to http status codes, but there are also other status codes such as `grpc`'s one.

Should we discuss this in one of our next team meetings?

src/locales/en/science/unit.ts
```diff
@@ -0,0 +1,90 @@
...
+    long: 'coulomb',
+    short: 'C',
+  },
+  {
+    long: 'volt',
+    short: 'V',
+  },
+  {
+    long: 'ohm',
+    short: 'Ω',
```
Shinigami92, line 64: Should we also name `short` just `symbol`?

src/definitions/science.ts
pkuczynski: Wikipedia is a bit different than us because they do not have namespaces, so they sometimes need a bit longer article names. The elements are equally chemical and physical, so I don't think this is a nice prefix. Electrical elements would probably be included in a different module I can imagine as the more for designing devices then pure scientific data, isn't it. Even in Wikipedia it says _periodic table of the (chemical) elements_ https://en.wikipedia.org/wiki/Periodic_table, where _chemical_ is in brackets...

src/modules/science/index.ts
```diff
@@ -0,0 +1,58 @@
+import type { Faker } from '../..';
+
+export type Element = {
+  name: string;
+  symbol: string;
+};
```
Shinigami92, line 6:
```suggestion
export interface Element {
  name: string;
  symbol: string;
};
```

src/modules/science/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type Unit = {
+  long: string;
+  short: string;
+};
```
Shinigami92, line 11:
```suggestion
export interface Unit {
  long: string;
  short: string;
};
```

Shinigami92 (changes requested):

Shinigami92 (changes requested):
Remember to re-export the interfaces ChemicalElement and Unit in `src/index.ts`

src/locales/en/science/unit.ts
import-brain: If we rename `short` to `symbol`, should we rename `long` to `full` or `name`?

Shinigami92: yes (😉 => _not sure whether `full` or `name`_)

src/modules/science/index.ts
```diff
@@ -0,0 +1,54 @@
...
+   * faker.science.unit() // { long: 'meter', short: 'm' }
+   * faker.science.unit() // { long: 'second', short: 's' }
+   * faker.science.unit() // { long: 'mole', short: 'mol' }
+   */
+  unit(): Unit {
+    const index = this.faker.datatype.number({
+      min: 0,
+      max: this.faker.definitions.science.unit.length - 1,
+    });
+    return this.faker.definitions.science.unit[index];
```
Shinigami92, line 52: pls use `arrayElement` for such things

test/science.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+  // Create and log-back the seed for debug purposes
+  faker.seed(Math.ceil(Math.random() * 1_000_000_000));
+
+  describe(`random seeded tests for seed ${JSON.stringify(
+    faker.seed()
+  )}`, () => {
+    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
+      describe(`elementName()`, () => {
+        it('should return a string', () => {
+          const name = faker.science.elementName();
```
Shinigami92, line 62: tests need updates

src/modules/science/index.ts
```diff
@@ -0,0 +1,54 @@
+import type { Faker } from '../..';
+
+export interface Element {
+  name: string;
+  symbol: string;
+};
```
Shinigami92, line 6: run `pnpm run format`

Shinigami92 (changes requested):

src/modules/science/index.ts
Shinigami92: `pnpm run clean && pnpm install && pnpm run format`

src/locales/en/science/unit.ts
pkuczynski: I vote for `name`

ST-DDT: I also prefer `name`.

src/modules/science/index.ts
Shinigami92: is something broken with prettier?

Shinigami92: it's already solved :thinking:

src/locales/en/science/unit.ts
import-brain: Okay, changing to `name`

src/modules/science/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface ChemicalElement {
+  name: string;
+  symbol: string;
+}
+
+export interface Unit {
+  name: string;
+  symbol: string;
+}
```
ST-DDT, line 11: Should we reuse the types in the definitions (or vice versa)?

src/locales/en/science/chemicalElement.ts
```diff
@@ -0,0 +1,474 @@
+export default [
+  {
+    symbol: 'H',
```
Shinigami92, line 3: Should we add a third field `atomicNumber`?

ST-DDT (changes requested):

src/locales/en/science/chemicalElement.ts
import-brain: Sure. Should the number be a string or an number type?

ST-DDT (dismissed):

src/locales/en/science/chemicalElement.ts
```diff
@@ -0,0 +1,474 @@
+export default [
```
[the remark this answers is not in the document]
import-brain: By "freeze", do you mean use `Object.freeze()`?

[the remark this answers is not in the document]
ST-DDT: Yes, so that unit().name = 123 doesnt do anything unexpected.

src/locales/en/science/chemicalElement.ts
Shinigami92: Mh... I think a number would fit here, or are there special cases?

Shinigami92 (approved):

ST-DDT (approved):

pkuczynski (approved):

PR #993, opened 2022-05-24
docs: use vitepress v1

Shinigami92:

PR #1017, opened 2022-05-28
docs: rework readme

Shinigami92:
https://github.com/faker-js/faker/blob/rework-readme/README.md

https://github.com/faker-js/faker/blob/rework-readme/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ You can view a code coverage report generated in `coverage/index.html`.
 
 ## Adding new locale or updating existing one
 
-After you have updated the locale definitions, you need to run `pnpm run generate:locales` to generate/update the related files.
+After adding new or updating existing locale data, you need to run `pnpm run generate:locales` to generate/update the related files.
 
 ## Developing the docs
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,6 +57,7 @@ pnpm run docs:dev
 
 ## Building and serving the docs statically
 
+If you changed something heavily in the docs, like auto-generating content, you should check the docs statically, because it could differ from the dev version.
 Before running the docs, build the Faker dist, it's used inside of certain routes.
 
 ```shell
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ pnpm run build
 
 ## Testing
 
-Before you can run the tests, you need to install all dependencies and also need to build the project, because some tests depend on the bundled content.
+Before you can run the tests, you need to install all dependencies and build the project, because some tests depend on the bundled content.
 
 ```shell
 pnpm install
@@ -43,7 +43,7 @@ pnpm run test
 pnpm run coverage
 ```
 
-You can view a code coverage report generated in `coverage/index.html`.
+You can view a generated code coverage report at `coverage/index.html`.
 
 ## Adding new locale or updating existing one
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,8 @@ pnpm run build
 
 ## Testing
 
+Before you can run the tests, you need to install all dependencies and also need to build the project, because some tests depend on the bundled content.
+
 ```shell
 pnpm install
 pnpm run build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,9 +47,9 @@ You can view a code coverage report generated in `coverage/index.html`.
 
 ## Developing the docs
 
+Before running the docs, build the Faker dist, it's used inside of certain routes.
+
 ```shell
-# build the Faker dist
-# it's used inside of certain routes
 pnpm run build
 
 pnpm run docs:dev
@@ -57,9 +57,9 @@ pnpm run docs:dev
 
 ## Building and serving the docs statically
 
+Before running the docs, build the Faker dist, it's used inside of certain routes.
+
 ```shell
-# build the Faker dist
-# it's used inside of certain routes
 pnpm run build
 
 pnpm run docs:build # Output docs to /dist
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,4 +69,4 @@ pnpm run docs:serve # Serve docs from /dist
 
 ## Deploying documentation
 
-The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
+See the [netlify.toml](netlify.toml) for configuration.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,6 @@ pnpm run docs:build # Output docs to /dist
 pnpm run docs:serve # Serve docs from /dist
 ```
 
-## Deploying Documentation
+## Deploying documentation
 
 The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,10 @@ pnpm run coverage
 
 You can view a code coverage report generated in `coverage/index.html`.
 
+## Adding new locale or updating existing one
+
+After you have updated the locale definitions, you need to run `pnpm run generate:locales` to generate/update the related files.
+
 ## Developing the docs
 
 Before running the docs, build the Faker dist, it's used inside of certain routes.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
 The sources are located in the [src](src) directory.
 All fake data generators are divided into namespaces (each namespace being a separate module).
 Most of the generators use the _definitions_, which are just plain JavaScript objects/arrays/strings that are separate for each [locale](src/locales).
+
+## Building Faker
+
+The project is being built by [esbuild](https://esbuild.github.io) (see [bundle.ts](scripts/bundle.ts))
+
+```shell
+pnpm install
+pnpm run build
+```
+
+## Testing
+
+```shell
+pnpm install
+pnpm run build
+
+pnpm run test
+# or
+pnpm run coverage
+```
+
+You can view a code coverage report generated in `coverage/index.html`.
+
+## Developing the docs
+
+```shell
+# build the Faker dist
+# it's used inside of certain routes
+pnpm run build
+
+pnpm run docs:dev
+```
+
+## Building and serving the docs statically
+
+```shell
+# build the Faker dist
+# it's used inside of certain routes
+pnpm run build
+
+pnpm run docs:build # Output docs to /dist
+pnpm run docs:serve # Serve docs from /dist
+```
+
+## Deploying Documentation
+
+The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
--- a/README.md
+++ b/README.md
@@ -3,158 +3,102 @@
   <h1>Faker</h1>
   <p>Generate massive amounts of fake (but realistic) data for testing and development.</p>
   
-  [![Chat on Discord](https://img.shields.io/discord/929487054990110771)](https://chat.fakerjs.dev)
-  [![Continuous Integration](https://github.com/faker-js/faker/actions/workflows/ci.yml/badge.svg)](https://github.com/faker-js/faker/actions/workflows/ci.yml)
-  [![codecov](https://codecov.io/gh/faker-js/faker/branch/main/graph/badge.svg?token=N61U168G08)](https://codecov.io/gh/faker-js/faker)
   [![npm version](https://badgen.net/npm/v/@faker-js/faker)](https://www.npmjs.com/package/@faker-js/faker)
   [![npm downloads](https://badgen.net/npm/dm/@faker-js/faker)](https://www.npmjs.com/package/@faker-js/faker)
+  [![Continuous Integration](https://github.com/faker-js/faker/actions/workflows/ci.yml/badge.svg)](https://github.com/faker-js/faker/actions/workflows/ci.yml)
+  [![codecov](https://codecov.io/gh/faker-js/faker/branch/main/graph/badge.svg?token=N61U168G08)](https://codecov.io/gh/faker-js/faker)
+  [![Chat on Discord](https://img.shields.io/badge/chat-discord-blue?style=flat&logo=discord)](https://chat.fakerjs.dev)
   [![Open Collective](https://img.shields.io/opencollective/backers/fakerjs)](https://opencollective.com/fakerjs#section-contributors)
   [![sponsor](https://img.shields.io/opencollective/all/fakerjs?label=sponsors)](https://opencollective.com/fakerjs)
   
 </div>
 
-### Try it Online ⚡️
-
-[fakerjs.dev/new](https://fakerjs.dev/new)
+## ⚡️ Try it Online
 
 [![](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://fakerjs.dev/new)
 
 [API Documentation](https://fakerjs.dev/guide/)
 
-## Installation
+## 🚀 Features
 
-Please replace your `faker` dependency with `@faker-js/faker`. This is the official, stable fork of Faker.
+- 💌 Addresses - Generate valid looking Addresses, Zip Codes, Street Names, States, and Countries!
+- ⏰ Time-based Data - Past, present, future, recent, soon... whenever!
+- 🌏 Localization - Set a locale to generate realistic looking Names, Addresses, and Phone Numbers.
+- 💸 Finance - Create stubbed out Account Details, Transactions, and Crypto Addresses.
+- 👠 Products - Generate Prices, Product Names, Adjectives, and Descriptions.
+- 👾 Hacker Jargon - “Try to reboot the SQL bus, maybe it will bypass the virtual application!”
+- 🔢 Numbers - Of course, we can also generate random numbers and strings.
 
-```shell
-npm install @faker-js/faker --save-dev
+## 📦 Install
+
+```bash
+npm install --save-dev @faker-js/faker
 ```
 
-or yarn
-
-```shell
-yarn add @faker-js/faker -D
-```
-
-or pnpm
-
-```shell
-pnpm install @faker-js/faker -D
-```
-
-### Browser
-
-```html
-<script src="faker.js" type="text/javascript"></script>
-<script>
-  const randomName = faker.name.findName(); // Caitlyn Kerluke
-  const randomEmail = faker.internet.email(); // Rusty@arne.info
-  const randomPhoneNumber = faker.phone.phoneNumber(); // (746) 637-3344 x8083
-</script>
-```
-
-### Node.js
-
-```js
-const { faker } = require('@faker-js/faker');
-
-const randomName = faker.name.findName(); // Rowan Nikolaus
-const randomEmail = faker.internet.email(); // Kassandra.Haley@erich.biz
-const randomPhoneNumber = faker.phone.phoneNumber(); // (279) 329-8663 x30233
-```
-
-### CDN/Deno
-
-```js
-import { faker } from 'https://cdn.skypack.dev/@faker-js/faker';
-
-const randomName = faker.name.findName(); // Willie Bahringer
-const randomEmail = faker.internet.email(); // Tomasa_Ferry14@hotmail.com
-const randomPhoneNumber = faker.phone.phoneNumber(); // 938-672-1359 x418
-```
-
-#### Alternative CDN links
-
-**esm:**
-
-- https://esm.sh/@faker-js/faker
-- https://cdn.jsdelivr.net/npm/@faker-js/faker/+esm
-
-**cjs:**
-
-- https://cdn.jsdelivr.net/npm/@faker-js/faker
-
-### TypeScript Support
-
-Since version `v6+` there is native TypeScript support.
-
-In order to have faker working properly, you need to check if these `compilerOptions` are set correctly in your `tsconfig` file:
-
-```json
-{
-  "compilerOptions": {
-    "esModuleInterop": true,
-    "moduleResolution": "Node"
-  }
-}
-```
-
-And then simply import it like everything else:
+## 🪄 Usage
 
 ```ts
 import { faker } from '@faker-js/faker';
-```
+// import { faker } from '@faker-js/faker/locale/de';
 
-If you want for whatever reason the versions prior to `v6`,
-you can use `@types/faker` and rebind the declarations to the `@faker-js/faker` package with a `faker.d.ts` file in your e.g. src folder.
+export const USERS: User[] = [];
 
-```ts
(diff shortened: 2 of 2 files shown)


ST-DDT (dismissed):

ejcheng (dismissed):

CONTRIBUTING.md
```diff
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+The project is being built by [esbuild](https://esbuild.github.io) (see [bundle.ts](scripts/bundle.ts))
+
+```shell
+pnpm install
+pnpm run build
+```
+
+## Testing
+
+```shell
```
xDivisionByZerox, line 35: Maybe some words what framework we use for testing (like line 26) or why we think tests are important.

README.md
```diff
@@ -213,67 +148,22 @@ Faker is an MIT-licensed open source project with its ongoing development made p
...
-pnpm run docs:serve # Serve docs from /dist
-```
-
-### Deploying Documentation
-
-The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
+<a href="https://github.com/faker-js/faker/graphs/contributors"><img src="https://opencollective.com/fakerjs/contributors.svg?width=800" /></a>
 
-### Thank you
+The [fakerjs.dev](https://fakerjs.dev) website is kindly hosted by the Netlify Team.
```
prisis, line 158: Can we add algolia, they Sponsoring our search

CONTRIBUTING.md
pkuczynski: For me it would be more important to mention that you always have to run build when changing the code, as tests are run from built code, not from source, which was quite unusual for me...

I would not mention `install` as this seems to be quite obvious step.

CONTRIBUTING.md
```diff
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+# or
+pnpm run coverage
+```
+
+You can view a code coverage report generated in `coverage/index.html`.
+
+## Developing the docs
+
+```shell
+# build the Faker dist
```
pkuczynski, line 49: This should be mention as a text, and not inside the code example I think...

CONTRIBUTING.md
```diff
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+# it's used inside of certain routes
+pnpm run build
+
+pnpm run docs:dev
+```
+
+## Building and serving the docs statically
```
pkuczynski, line 56: It would be nice to explain why this is needed and how this differ from the docs:dev

CONTRIBUTING.md
```diff
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+```shell
+# build the Faker dist
+# it's used inside of certain routes
+pnpm run build
+
+pnpm run docs:build # Output docs to /dist
+pnpm run docs:serve # Serve docs from /dist
+```
+
+## Deploying Documentation
```
pkuczynski, line 67:
```suggestion
## Deploying documentation
```

README.md
```diff
@@ -3,159 +3,103 @@
...
-## Installation
-
-Please replace your `faker` dependency with `@faker-js/faker`. This is the official, stable fork of Faker.
-
-```shell
-npm install @faker-js/faker --save-dev
-```
-
-or yarn
+## 🚀 Features
```
pkuczynski, line 22: I don't think it's worth to mention specific features. They keep changing, we are constantly adding new stuff. This list will get outdated quite quickly. And even now it does not cover everything.

It would be better to mention philosophy behind faker, that we try to generate sensible, realistic data when possible, we offer localization in many languages and ability to pin the generated by setting seed value - I didn't know about this feature until I started to work on faker.

README.md
```diff
@@ -3,159 +3,103 @@
...
+- 💌 Addresses - Generate valid looking Addresses, Zip Codes, Street Names, States, and Countries!
+- ⏰ Time-based Data - Past, present, future, recent, soon... whenever!
+- 🌏 Localization - Set a locale to generate realistic looking Names, Addresses, and Phone Numbers.
+- 💸 Finance - Create stubbed out Account Details, Transactions, and Crypto Addresses.
+- 👠 Products - Generate Prices, Product Names, Adjectives, and Descriptions.
+- 👾 Hacker Jargon - “Try to reboot the SQL bus, maybe it will bypass the virtual application!”
+- 🔢 Numbers - Of course, we can also generate random numbers and strings.
 
-or pnpm
+## 📦 Install
```
pkuczynski, line 32: Those icons look childish, do not bring value and are hard to provide for every section. I wouldn't add them...

README.md
```diff
@@ -3,159 +3,103 @@
...
-```ts
-// faker.d.ts
-declare module '@faker-js/faker' {
-  import faker from 'faker';
-  export default faker;
+// import { faker } from '@faker-js/faker/locale/de';
+
+export const USERS: User[] = [];
+
+export function createRandomUser(): User {
```
pkuczynski, line 46: Why not using arrow function here too?

README.md
```diff
@@ -3,159 +3,103 @@
...
-| Phone    | `faker.phone.phoneNumber()`                   | +1 291-299-0192                                                                                                                                                                                                         |
-| Random   | `faker.random.locale()`                       | fr_CA                                                                                                                                                                                                                   |
-| System   | `faker.system.directoryPath()`                | C:\Documents\Newsletters\                                                                                                                                                                                               |
-| Vehicle  | `faker.vehicle.vehicle()`                     | 2011 Dodge Caravan                                                                                                                                                                                                      |
+Array.from({ length: 10 }).forEach(() => {
```
pkuczynski, line 58: `Array.from({ length: 10 }, createRandomUser)`

Would be shorter...

README.md
```diff
@@ -3,159 +3,103 @@
...
-
-If you want for whatever reason the versions prior to `v6`,
-you can use `@types/faker` and rebind the declarations to the `@faker-js/faker` package with a `faker.d.ts` file in your e.g. src folder.
-
-```ts
-// faker.d.ts
-declare module '@faker-js/faker' {
-  import faker from 'faker';
-  export default faker;
+// import { faker } from '@faker-js/faker/locale/de';
```
pkuczynski, line 42: We have a separate section about localization so I would not leave this comment here...

README.md
```diff
@@ -3,159 +3,103 @@
...
-| Random   | `faker.random.locale()`                       | fr_CA                                                                                                                                                                                                                   |
-| System   | `faker.system.directoryPath()`                | C:\Documents\Newsletters\                                                                                                                                                                                               |
-| Vehicle  | `faker.vehicle.vehicle()`                     | 2011 Dodge Caravan                                                                                                                                                                                                      |
+Array.from({ length: 10 }).forEach(() => {
+  USERS.push(createRandomUser());
+});
+```
+
+## Modules
```
pkuczynski, line 63: I would drop this whole section. In favor of something like "Check documentation for complete list of data you can generate"

README.md
```diff
@@ -3,159 +3,103 @@
...
+| Music    | `faker.music.genre()`                         | R&B                                                                                                |
+| Name     | `faker.name.firstName()`                      | Cameron                                                                                            |
+| Phone    | `faker.phone.phoneNumber()`                   | +1 291-299-0192                                                                                    |
+| Random   | `faker.random.locale()`                       | fr_CA                                                                                              |
+| System   | `faker.system.directoryPath()`                | /root                                                                                              |
+| Vehicle  | `faker.vehicle.vehicle()`                     | Lamborghini Camry                                                                                  |
+| Word     | `faker.word.adjective()`                      | adorable                                                                                           |
 
 ### faker.fake()
```
pkuczynski, line 93: Name it "Templates"?

pkuczynski (changes requested):

CONTRIBUTING.md
```diff
@@ -20,3 +20,50 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+
+The website is kindly hosted for free by the Netlify team under their Open Source plan. See the [netlify.toml](netlify.toml) for configuration.
```
pkuczynski, line 69: This is already mentioned on the main Readme so is this worth the repetition? It's not connected with contribution to the project, right?

**localisation**

I would add new section documenting how to add a new locale. People ask about it. See https://discord.com/channels/929487054990110771/929544823109742593/981252262515060817

pkuczynski (changes requested):

README.md
```diff
@@ -164,27 +108,18 @@ The default language locale is set to English.
 
 Setting a new locale is simple:
 
-```js
+```ts
```
pkuczynski, line 111: We should mention that user needs to load the locale first. Correct?

README.md
```diff
@@ -164,27 +108,18 @@ The default language locale is set to English.
...
 // sets locale to de
 faker.locale = 'de';
 ```
 
 See our documentation for a list of [provided languages](https://fakerjs.dev/api/localization.html#localization)
```
pkuczynski, line 116: Should we also mention that not all locales provide data for every method and then we fallback to English (do we?)? And also encourage users to provide PR with locale data pointing them to dedicated section in the contributing guide.

pkuczynski (changes requested):

README.md
ST-DDT: I think we should keep the feature list, as it easier to understand than our philosphy. Kind of a TLDR.

We should still explain our philosophy though as it provides deeper insides into what we might consider to be a part of faker.

README.md
```diff
@@ -3,159 +3,103 @@
...
-```shell
-yarn add @faker-js/faker -D
-```
+- 💌 Addresses - Generate valid looking Addresses, Zip Codes, Street Names, States, and Countries!
+- ⏰ Time-based Data - Past, present, future, recent, soon... whenever!
+- 🌏 Localization - Set a locale to generate realistic looking Names, Addresses, and Phone Numbers.
+- 💸 Finance - Create stubbed out Account Details, Transactions, and Crypto Addresses.
+- 👠 Products - Generate Prices, Product Names, Adjectives, and Descriptions.
+- 👾 Hacker Jargon - “Try to reboot the SQL bus, maybe it will bypass the virtual application!”
+- 🔢 Numbers - Of course, we can also generate random numbers and strings.
```
ST-DDT, line 30: I think we should also add 
`- 🧍 Names - Generate virtual humans with a complete online and offline identity`

README.md
ST-DDT: IMO they loosen the atmosphere and highlight the different sections very distictively.
In presentations, you also have images on your slides to retain the viewers attention.

We should do it consistently for all headlines though (All or none).

README.md
ST-DDT: IMO most users don't read that far:
- They will read whether it suits they needs.
- Then they will check how to use it
- Then they will explore the API from within their IDE or search for specific terms

So if we plant this knowledge in their minds early on they at least notice, that there are more ways to do things.
I would potentially include a `// faker.locale = 'en'` as well.

README.md
ST-DDT: > Should we also mention that not all locales provide data for every method and then we fallback to English (do we?)?

Yes, and yes for now.

> And also encourage users to provide PR with locale data pointing them to dedicated section in the contributing guide.

Yes, but I would like to also have a section like build your own locale dynamically/indepently.
Althoug we would have to rewrite it soon.

README.md
pkuczynski: If we would mention earlier that faker is localized i to X languages, if they need it, they will seek it. TOC at the top would also build some expectations.

Leaving commented code looks like someone forgot to remove it...

README.md
```diff
@@ -3,159 +3,103 @@
...
-  [![codecov](https://codecov.io/gh/faker-js/faker/branch/main/graph/badge.svg?token=N61U168G08)](https://codecov.io/gh/faker-js/faker)
   [![npm version](https://badgen.net/npm/v/@faker-js/faker)](https://www.npmjs.com/package/@faker-js/faker)
   [![npm downloads](https://badgen.net/npm/dm/@faker-js/faker)](https://www.npmjs.com/package/@faker-js/faker)
+  [![Continuous Integration](https://github.com/faker-js/faker/actions/workflows/ci.yml/badge.svg)](https://github.com/faker-js/faker/actions/workflows/ci.yml)
+  [![codecov](https://codecov.io/gh/faker-js/faker/branch/main/graph/badge.svg?token=N61U168G08)](https://codecov.io/gh/faker-js/faker)
+  [![Chat on Discord](https://img.shields.io/badge/chat-discord-blue?style=flat&logo=discord)](https://chat.fakerjs.dev)
   [![Open Collective](https://img.shields.io/opencollective/backers/fakerjs)](https://opencollective.com/fakerjs#section-contributors)
   [![sponsor](https://img.shields.io/opencollective/all/fakerjs?label=sponsors)](https://opencollective.com/fakerjs)
   
 </div>
```
pkuczynski, line 15: Add TOC here

pkuczynski (changes requested):

README.md
Shinigami92: I tried to find icons for now for all first lvl headlines
Please check them out and check if the fit

README.md
Shinigami92: This is my personal style I always use in projects :thinking: 
I never export arrow functions but only real functions.

Do you have any reason why I should change that?

README.md
Shinigami92: This would not push the users into the `USERS` array.
Also it could be that something inside the function accesses the `USERS` array, and that's why I don't create first all at once and then push them all at once into the array.

README.md
Shinigami92: If we mention more like `faker.locale = 'en'` we may also need to mention `seed` :thinking: 
But that way the usage would get longer and longer :confused:

README.md
Shinigami92: I would like to drop this section a bit later after:
1. we finished the whole migration to VitePress v1 final (currently it's still in alpha) (#993)
2. then create a site that is searchable like https://vuejs.org/api/

README.md
Shinigami92: Will add this as suggested for now, but I really want to switch that to `person` in the future (rename the module).

README.md
Shinigami92: Did you know that GitHub as an auto-toc?


IMO we don't need a toc

CONTRIBUTING.md
```diff
@@ -20,3 +20,57 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+pnpm run test
+# or
+pnpm run coverage
+```
+
+You can view a code coverage report generated in `coverage/index.html`.
+
+## Adding new locale or updating existing one
+
+After you have updated the locale definitions, you need to run `pnpm run generate:locales` to generate/update the related files.
```
Shinigami92, line 50: @ST-DDT please check this section and provide suggestion or even update this yourself as you have access to the branch :slightly_smiling_face:

README.md
Shinigami92: :eyes: You somewhat got a point here, but instead of documenting that, we should throw a warning or even an error that the requested locale cannot be switched to due to it was forgotten to get loaded :thinking: 
This should be it's own PR

CONTRIBUTING.md
ST-DDT, line 50:
```suggestion
After adding new or updating existing locale data, you need to run `pnpm run generate:locales` to generate/update the related files.
```

README.md
ST-DDT: Since we are about to remove the current way of setting locaes, there is no need to mention it now.

README.md
```diff
@@ -197,7 +135,7 @@ const secondRandom = faker.datatype.number();
 console.log(firstRandom === secondRandom);
 ```
 
-## Sponsors
+## 💵 Sponsors
```
ST-DDT, line 138: Maybe?

```suggestion
## 🤝 Sponsors
```

README.md
pkuczynski: I dont like those icons, so I will just remain silent...

README.md
pkuczynski: We should give example on faker usage, and not how to make a module. Those exports are not really needed as they are irrelevant to the faker. I prefer the minimalistic approach.

Btw. why do you choose to export function over arrow? In all my projects always use arrow and have not really seen people using old plain `function`...

README.md
pkuczynski: It would. Have a look at this example: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#using_arrow_functions_and_array.from - you provide options and mapper.

README.md
pkuczynski: That's why I think TOC and simple short examples targeting the section subject are better.

README.md
pkuczynski: But we already have the documentation at https://fakerjs.dev/api/company.html, which is already better than readme and more "auto-updated" with the code...

README.md
pkuczynski: I know, but it's not that visible to the user, I think.

README.md
pkuczynski: We should show the user a working example, and not something which throws a warning...

README.md
```diff
@@ -197,7 +135,7 @@ const secondRandom = faker.datatype.number();
 console.log(firstRandom === secondRandom);
 ```
 
-## Sponsors
+## 🤝 Sponsors
 
 Faker is an MIT-licensed open source project with its ongoing development made possible entirely by the support of these awesome backers
```
pkuczynski, line 140:
```suggestion
Faker is an open source project with its ongoing development made possible entirely by the support of these awesome backers
```

pkuczynski (changes requested):

CONTRIBUTING.md
```diff
@@ -20,3 +20,57 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+The project is being built by [esbuild](https://esbuild.github.io) (see [bundle.ts](scripts/bundle.ts))
+
+```shell
+pnpm install
+pnpm run build
+```
+
+## Testing
+
+Before you can run the tests, you need to install all dependencies and also need to build the project, because some tests depend on the bundled content.
```
import-brain, line 35:
```suggestion
Before you can run the tests, you need to install all dependencies and build the project, because some tests depend on the bundled content.
```

CONTRIBUTING.md
```diff
@@ -20,3 +20,57 @@ If you want to make `Faker` a better, please read the following contribution gui
...
+```shell
+pnpm install
+pnpm run build
+
+pnpm run test
+# or
+pnpm run coverage
+```
+
+You can view a code coverage report generated in `coverage/index.html`.
```
import-brain, line 46:
```suggestion
You can view a generated code coverage report at `coverage/index.html`.
```

README.md
Shinigami92: I'm missing the **why** here, but if another maintainer is okay with removing this, I will apply the suggestion

README.md
Shinigami92: You may have misinterpreted something here :thinking: 
This is not a module but a function that generates an object e.g. for a database tuple.

---

I normally write TypeScript code all the day and the code is shorter when using a function.
I'm also not fully sure, but I think when using `function` it will be correctly named in debug or source maps, where as an arrow-function is anonymous.
Other projects like VueUse also use the same kind of exported function.
I only use arrow functions for lambda syntax like in a Array.map function or when I really need to prevent a collision with `this` keyword. But in most cases when I externalize a function, it doesn't depend on the `this` keyword.

README.md
Shinigami92: Yes, I use this myself, but only when I need to return an Array.
In this case, I use a terminating `forEach` which doesn't return an Array but void.

README.md
Shinigami92: > Re-reviewing it now again and it really looks weird that we keep this commented line...

Please suggest something that explains that it's also possible to use our pre-made locale faker instances which results in much smaller bundle size.
This line is somewhat important for us to be mentioned such high in the readme as we answer many times that this is possible. In GitHub Discussions and in Discord we get asked again and again and in 99% this was the solution.

README.md
pkuczynski: Well, if it's a plain function as an example, no need to export. If you export, it becomes a module :) And I agree it should be just a plain function, that's why I am suggesting removing all those exports.

Not sure what do you mean by shorter as arrow functions are much shorter:

```js
export function createRandomUser(): User {
  return {
     userId: faker.datatype.uuid(),
     ...
   };
 }
 ```

vs

```js
export const createRandomUser = (): User => ({
     userId: faker.datatype.uuid(),
     ...
})
```

Not sure about `Vue` not using it. React (and many other libraries) use arrow functions everywhere. I use them everywhere, hardly every use old plan `function`.

README.md
pkuczynski: I know, but instead of writing this like you did:

```ts
export const USERS: User[] = [];

Array.from({ length: 10 }).forEach(() => {
  USERS.push(createRandomUser());
 });
```

this can be shortened to:

```ts
export const USERS: User[] = Array.from({ length: 10 }, createRandomUser)
```

Isn't that nicer? (Although I would drop `export`)

README.md
pkuczynski: I would remove the commented code and add new subsection to usage:

```ms
### Optimizing size

By default faker loads all locales. If you only use few of them or care about the size of your bundle, you should rather use single-locale instances:

```ts
import { faker } from '@faker-js/faker/locale/de';

faker.street.name() // ...
```

or for multiple locale usage:

```ts
import { faker as en } from '@faker-js/faker/locale/en';
import { faker as de } from '@faker-js/faker/locale/de';

const faker = { en, de }

faker.en.street.name() // ...
faker.de.street.name() // ...
```
```

(code examples from top of my head, but you get the drill)

README.md
pkuczynski: You think adding licence type creates this *why*?

README.md
Shinigami92: I think we can agree here that this is opinionated and doesn't serve anymore any valuable stuff to the context itself. I will mark this as resolved now, let's move on.

README.md
Shinigami92: https://github.com/faker-js/faker/pull/1017#discussion_r889681341

README.md
Shinigami92: I mean why removing this. Please provide a reason why removing this info is valuable for this sentence.
It tells the valuable info that it is licensed under MIT license within the sponsor/contribution section.

README.md
pkuczynski: The value in my proposal was to keep things simple to focus on the faker example. No exports, shortest code. You argue into "this is my personal style and I want to keep it". There are no arguments to why your personal style is better. Exposing personal style does not sound like a good open source cooperative atmosphere... You didn't even let other maintenances raise their voice...

pkuczynski (approved):

README.md
Shinigami92: _resolve to allow merge, we could improve in another PR_

README.md
Shinigami92: _resolve to allow merge, we could improve in another PR_

README.md
Shinigami92: _resolve to allow merge, we could improve in another PR_

prisis (approved):

PR #1026, opened 2022-05-31
feat: add en_JP locale

revmischa:
Added the amazing Japanese interpretations of American-sounding names from https://docs.google.com/spreadsheets/d/1KOP4mNAX5R0N_dODE6j2gHTK2oTXCIrhAQ0dEniaYFw/edit#gid=0

image


src/locales/en_JP/index.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+import type { LocaleDefinition } from '../..';
+import name_ from './name';
+
+const en_JP: LocaleDefinition = {
+  title: 'English Names - 1994 Fighting Baseball (SNES) Japan Full Rosters',
```
ST-DDT, line 9: I'm not sure about this name.
Maybe `English (Japanese)`?

src/locales/en_JP/name/first_name.ts
```diff
@@ -0,0 +1,703 @@
+export default [
+  'Kevin',
+  'Luis',
+  'Derrick',
```
ST-DDT, line 4: Are these "westernized" versions of Japanese names?

src/locales/en_JP/index.ts
revmischa: As you wish, but it's just from this one source..

src/locale/en_JP.ts
```diff
@@ -0,0 +1,19 @@
...
+const faker = new Faker({
+  locale: 'en_JP',
+  localeFallback: 'en',
+  locales: {
+    en_JP,
+    en,
+  },
+});
+
+export = faker;
```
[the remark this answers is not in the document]
revmischa: No I had an old fork, let me export it properly

src/locales/en_JP/name/first_name.ts
revmischa: https://www.reddit.com/r/baseball/comments/8klxnf/the_full_fighting_baseball_snes_rosters_more_than/

src/locales/en_JP/index.ts
Shinigami92: Yeah, but maybe... just maybe, someone in the future would like to add something else to the en_JP set and they definitely will not think about this `title` field :wink: 
All out other locales have this not-written-down convention :thinking: 
Something in me tend more to `English (Japanese)`

src/locales/en_JP/name/index.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Run 'pnpm run generate:locales' to update.
+ */
+import type { NameDefinitions } from '../../..';
+import first_name from './first_name';
+import last_name from './last_name';
+
+const name = {
+  first_name,
+  last_name,
+} as Partial<NameDefinitions>;
```
ST-DDT, line 12: Please run `pnpm run generate:locales` again.

src/locales/en_JP/index.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+import type { LocaleDefinition } from '../..';
+import name_ from './name';
+
+const en_JP: LocaleDefinition = {
+  title: 'English Names (Japan)',
```
ST-DDT, line 9:
```suggestion
  title: 'English (Japan)',
```

@Shinigami92 What do you think?

ST-DDT (changes requested):

src/locales/en_JP/name/index.ts
revmischa: ok done

ST-DDT:
Blocked by #1027 and I cannot push the fix directly to this branch.

revmischa:
> Blocked by #1027 as I cannot push the fix directly to this branch.

I rebased and regenerated

src/locales/en_JP/index.ts
Shinigami92:
```suggestion
  title: 'English (Japan)',
```

Shinigami92:
@revmischa Please use `English (Japan)` instead of `English Names (Japan)` because we use the same of e.g. `English (Ghana)` or `English (India)`
I cannot apply suggestions on your PR, you may have set your branch to not allow maintainers edit

ST-DDT (commented):
Let me ask a few more questions regarding how the westernized names work in Japanese.

Is the list you provided (a subset of) a defined list of allowed names that Japanese people can use or is this just a fun/famous list of names in a Japanese context?

revmischa:
> Let me ask a few more questions regarding how the westernized names work in Japanese.
> 
> Is the list you provided (a subset of) a defined list of allowed names that Japanese people can use or is this just a fun/famous list of names in a Japanese context?

The latter. I know of no "allowed list" that exists

Shinigami92:
> The latter. I know of no "allowed list" that exists

In that case, please explain deeply why this is worthfull to be included in Faker. Do you need it for your own app?
Is this needed by others not just your app? Otherwise you also could just use `faker.helpers.arrayElement(yourNameArray)` :thinking:

revmischa:
> > The latter. I know of no "allowed list" that exists
> 
> In that case, please explain deeply why this is worthfull to be included in Faker. Do you need it for your own app? Is this needed by others not just your app? Otherwise you also could just use `faker.helpers.arrayElement(yourNameArray)` 🤔

Because it produces amusing names. Others will enjoy and appreciate it. Same reason we have en_BORK.  https://github.com/faker-js/faker/blob/main/src/locales/en_BORK/lorem/words.ts

xDivisionByZerox:
The problem with intruducing locales is: They are hard to maintain (like really hard). If you introduce this new locale we have to support this best case all features we have. This does not only include names but any module (like interent, finance, etc). If you would currently load an instance of your locale most faker methods would fail (or fall back to english). 

Thats why we have to be very thoughtfull about introducing new locales. 

For existing locales its a bit harder (like en_BORK). They are already out in the wild, we need to keep supporting them unless we drop them. But that would require us to make major releases with not much impact.

Hope that clearyfication helps.

src/locales/en_JP/name/first_name.ts
```diff
@@ -0,0 +1,702 @@
...
+  'Jeff',
+  'Delino',
+  'Tim',
+  'Jarvis',
+  'Robby',
+  'Vince',
+  'Mickey',
+  'Jeff',
+  'Ave',
+  'Bobby',
```
kid-icarus, line 182: There are many duplicate entries here which is very wasteful and bandwidth intensive. Just search for 'Bob' or 'Bobby'.

ST-DDT:
Some statistics for future reference:
From the firstnames 190 out of 310 are also present in `en`, the remaining 120 names mostly only differ in the spelling of the name.
The lastname look like mutated original names.

These names might be funny when matching first and last names are returned because you can guess the original player name. However randomly shuffled together, they are just random names.
If you think the missing first names are important, then we can consider adding them directly to `en`
For the last names, I'm not sure as I'm not sure whether these are/feel like English last names (I'm probably not the best one to judge that).

src/locales/en_JP/name/first_name.ts
kid-icarus: Specifically, there are 118 duplicate first names, and 30 duplicate laste names.

kid-icarus:
> These names might be funny when matching first and last names are returned because you can guess the original player name. However randomly shuffled together, they are just random names. If you think the missing first names are important, then we can consider adding them directly to `en` For the last names, I'm not sure whether these are/feel like English last names (I'm probably not the best one to judge that).

As an english speaker I can confirm these are English-sounding last names.

revmischa:
I think random names are fine. I could remove the duplicates.

ST-DDT:
I'm closing this as the additional maintenance efforts outweighs the benefits/fun.

PR #1034, opened 2022-06-03
feat: always use https for loremflickr

bfanger:
When using https for you testpage you can only use the imageUrl function, all the other image  helper functions will generate http:// urls.

This results in the image being blocked by the browser:
```
Mixed Content: The page at 'https://domain.tld/storybook/' was loaded over HTTPS,
but requested an insecure image 'http://loremflickr.com/100/100/nightlife?w=200&fm=webp&q=85'. 
This request has been blocked; the content must be served over HTTPS.
```

https image urls also work when the the page is hosted on http, in fact the http image urls will be redirected to https by lorumflicker.

This PR changes the default to `https://` and because the image is served over https anyway i've removed the `https?: boolean` parameter instead of swapping the default value.

The only code change is in the `imageUrl` function( line 93), the other changes are inside comments.

### DIFF
--- a/src/modules/image/index.ts
+++ b/src/modules/image/index.ts
@@ -36,9 +36,9 @@ export class Image {
    * @param randomize Whether to randomize the image or not. Defaults to `false`.
    *
    * @example
-   * faker.image.image() // 'http://loremflickr.com/640/480/city'
-   * faker.image.image(1234, 2345) // 'http://loremflickr.com/1234/2345/sports'
-   * faker.image.image(1234, 2345, true) // 'http://loremflickr.com/1234/2345/nature?56789'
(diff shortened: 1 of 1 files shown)


Shinigami92 (approved):

ST-DDT (approved):
Should we not consider this a breaking change?

import-brain (approved):

Shinigami92:
> Should we not consider this a breaking change?

IMO it's okay, only very few customers would use this last parameter and even if they would use it, they only get a quick compile error in TS if they even have TS. In JS the last parameter would just be ignored.
I would count it as a minor _breaking_ change, we will highlight it in next release notes at top.

PR #1042, opened 2022-06-08
feat: improve norwegian translations

Johannes-Andersen:
Added a few missing Norwegian Bokmål translations that where missing

My first PR to the project, so let me know if I have done something stupid :)


ST-DDT (dismissed):
Checked the translations using google and looks good to me.

ST-DDT:
Can you check why Github shows your commit as unverified?
grafik

Johannes-Andersen:
> Can you check why Github shows your commit as unverified?
> grafik

Ah, I’ll get that fixed tomorrow. Didn’t have my hardware key with my GPG key plugged in so the commit did not get signed

Johannes-Andersen:
> Can you check why Github shows your commit as unverified? grafik

Should be sorted now. All commits on the PR should be signed 👌🏻

ST-DDT (approved):

ST-DDT:
> Should be sorted now. All commits on the PR should be signed 👌🏻

Works as expected. Thanks!

src/locales/nb_NO/color/human.ts
```diff
@@ -0,0 +1,19 @@
+export default [
+  'beige',
+  'grønn',
+  'grå',
+  'gul',
+  'hvit',
+  'indigo',
+  'karmin',
+  'blå',
```
Shinigami92, line 9: Did we already defined somewhere that we want sorted arrays? Or will this be in a separated PR done via automation?
In both cases, we may also want a CI check to check if the arrays are sorted :thinking: ... could get complicated...

Shinigami92 (approved):

src/locales/nb_NO/color/human.ts
xDivisionByZerox: We discussed that but wanted a way to automatically do this (like an eslint rule). So as long as this isn't in place it shouldn't be a requirement.

import-brain (approved):

PR #1043, opened 2022-06-08
feat: `hacker.hashAlgorithm`

### LINKED ISSUE #1038: Blockchain module

I wish to be able to use Faker to be able to generate random cryptocurrency and blockchain related entries.

### Suggested solution

`faker.blockchain.crypto() // { name: "Cardano", symbol: "ADA" }`
`faker.blockchain.crypto() // { name: "Ripple", symbol: "XRP" }`
As you can see, for `blockchain.crypto()`, the backend should work pretty similarly to the functions in `science`.

`faker.blockchain.hashAlgorithm() // "SHA-256"`
`faker.blockchain.hashAlgorithm() // "X11"`

In addition, deprecate the `finance.bitcoinAddress()`, `finance.ethereumAddress()`, and `finance.litecoinAddress()` functions and move them to the blockchain module.

### Alternative

_No response_

### Additional context

_No response_

ejcheng:
Partially completes #1038


src/locales/en/hacker/hashAlgorithm.ts
```diff
@@ -0,0 +1,97 @@
...
+  'SHA-224',
+  'Cunningham chains',
+  'Dcrypt',
+  'Equihash 210_9',
+  'Grøstl-512',
+  'Lyra2RE',
+  'DEDAL',
+  'Lyra2vc0ban',
+  'Prime Six',
+];
```
xDivisionByZerox, line 97: This is not language-dependent and should probably be scoped in a file in the hacker module subdirectory.

xDivisionByZerox: Yes, I would put it in the `src/modules/hacker` directory.  The `index.ts` file itself is the module, so don't name it that way. An example of this would be `src/modules/internet/user-agend.ts`. 
So I would name this file `hash-algorythm.ts`.

import-brain: Ah, I see. Thanks for the clarification!

pkuczynski: `hash-algorithm`

xDivisionByZerox: Yeah you can't rely on my English spelling 😁

ST-DDT:
@import-brain Do you have a need for this method/these data?

import-brain:
> @import-brain Do you have a need for this method/these data? If not, I would like to mark this as "waiting for user interest".

I don't have any urgent need for this right now, just thought it would fit well with the hacker module.

ST-DDT:
@import-brain Did you close this PR by accident?

import-brain:
> @import-brain Did you close this PR by accident?

@ST-DDT Since we do not need this right now, I was hoping to keep it closed for now and reopen it later when we need it.

Issue #1044, opened 2022-06-09
Source code todo list v2


ST-DDT:
This list contains all `TODO`s as of c5ac124ecafafc89bdd31a5870f325d671c7e9cf.

| Done | Issue/PR | File | Reporter | Description |
| - | - | - | - | --- |
| ⛔ | #1066 | [src/modules/address/index.ts:426](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/address/index.ts#L426) | @ST-DDT | Allow coordinate parameter to be [string, string]. |
| ❓ | - | [src/modules/address/index.ts:455](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/address/index.ts#L455) | @Shinigami92 | Provide an option property to provide custom circumferences. |
| ❓ | - | [src/modules/helpers/index.ts:225](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/helpers/index.ts#L225) | unknown | implement for letters e.g. [0-9a-zA-Z] etc. |
| ❓ | - | [src/modules/helpers/index.ts:398](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/helpers/index.ts#L398) | @Shinigami92 | We want to remove this default value, but currently it's not possible because some definitions could be empty |
| ❓ | - | [src/modules/helpers/index.ts:424](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/helpers/index.ts#L424) | @Shinigami92 | We want to remove this default value, but currently it's not possible because some definitions could be empty |
| ⛔ | - | [src/modules/image/providers/lorempicsum.ts:6](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/image/providers/lorempicsum.ts#L6) | @ST-DDT | Rename to picsum? |
| ❓ | - | [src/modules/image/providers/lorempicsum.ts:69](https://github.com/faker-js/faker/blob/c5ac124ecafafc89bdd31a5870f325d671c7e9cf/src/modules/image/providers/lorempicsum.ts#L69) | @ST-DDT | This method does the same as image url, maybe generate a seed, if it is missig? |
[...]

PR #1045, opened 2022-06-09
test: improve internet tests

Shinigami92:
related with #1044

### DIFF
--- a/test/internet.spec.ts
+++ b/test/internet.spec.ts
@@ -477,7 +477,8 @@ describe('internet', () => {
 
           expect(ua).toBeTruthy();
           expect(ua).toBeTypeOf('string');
-          // TODO @Shinigami92 2022-02-11: Make tests more strict
+          expect(ua.length).toBeGreaterThanOrEqual(1);
+          expect(ua).toMatch(/^(Mozilla|Opera)\/.*([Mm]ac|[Ww]in|[Ll]inux).*/);
         });
       });
 
--- a/test/internet.spec.ts
+++ b/test/internet.spec.ts
@@ -84,7 +84,7 @@ const seededRuns = [
   },
 ];
 
-const NON_SEEDED_BASED_RUN = 500;
+const NON_SEEDED_BASED_RUN = 5;
 
 const functionNames = [
   'avatar',
--- a/test/internet.spec.ts
+++ b/test/internet.spec.ts
@@ -478,7 +478,9 @@ describe('internet', () => {
           expect(ua).toBeTruthy();
           expect(ua).toBeTypeOf('string');
           expect(ua.length).toBeGreaterThanOrEqual(1);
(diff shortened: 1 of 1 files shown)


import-brain (dismissed):

test/internet.spec.ts
```diff
@@ -477,7 +477,8 @@ describe('internet', () => {
 
           expect(ua).toBeTruthy();
           expect(ua).toBeTypeOf('string');
-          // TODO @Shinigami92 2022-02-11: Make tests more strict
+          expect(ua.length).toBeGreaterThanOrEqual(1);
+          expect(ua).toMatch(/^(Mozilla|Opera)\/.*([Mm]ac|[Ww]in|[Ll]inux).*/);
```
xDivisionByZerox, line 481: Can we split this into 2 matches?
So if the first fails we know the string didn't start with the expected value and if the second fails we know a value is missing in the expected string.

Shinigami92: Meh... the string gets trimmed 
image

Okay I will split it

Shinigami92: Okay tried a bit around... this should be "fixed" in vitest, not here

xDivisionByZerox: Furthermore, I think the test itself isn't well decoupled from the implementation. We expect a valid user agent from this function. According to [RFC 2616 (HTTP protocol) - section 14.43 User-Agend](https://datatracker.ietf.org/doc/html/rfc2616#section-14.43) any string following the format `/^([^\d]+\/[\dA-Za-z\.]+)$|^([^\d]+\/[\dA-Za-z\.]+) ([^\d]+\/[\dA-Za-z\.]+)*$/` should be valid. But that very hard to read i guess.

Shinigami92: The match against this pattern fails :thinking:

xDivisionByZerox: Ok, the error was that I forgot to check for potential comments.
The new regex (with checks for comments) should work:
`/^(([^\d]+\/[\dA-Za-z\.]+(\s\(.*\)))|([^\d]+\/[\dA-Za-z\.]+(\s\(.*\)*))(\s[^\d]+\/[\dA-Za-z\.]+(\s\(.*\)*))*)$/` and works with my current user agent 
```
/^(([^\d]+\/[\dA-Za-z\.]+(\s\(.*\)))|([^\d]+\/[\dA-Za-z\.]+(\s\(.*\)*))(\s[^\d]+\/[\dA-Za-z\.]+(\s\(.*\)*))*)$/.test('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.63 Safari/537.36 Edg/102.0.1245.33');
// true

ST-DDT (approved):

import-brain (approved):

PR #1046, opened 2022-06-09
chore(unsplash): deprecate unused const

import-brain:
This is a todo on "source code todo list v2"


src/modules/image/providers/unsplash.ts
```diff
@@ -4,16 +4,6 @@ import type { Faker } from '../../..';
  * Module to generate links to random images on `https://source.unsplash.com/`.
  */
 export class Unsplash {
-  // TODO ST-DDT 2022-03-11: Remove unused(?) constant
-  categories = [
```
Shinigami92, line 8: This could be used like `faker.image.unsplash.categories`, but I assume no one in the world does this :thinking:

xDivisionByZerox: Even if, this isn't any kind of data you would relate with faker. So a removal should be fine. 

Still a breaking change.

ST-DDT: This could potentially be used for: https://github.com/faker-js/faker/blob/3f49ac1e38263064372fc1970d08f23e00cd580b/src/modules/image/providers/unsplash.ts#L40-L45

So we should follow the normal deprecation procedure.

import-brain: Do you want the const itself to be somehow deprecated, or do you want the parameter to be deprecated?

xDivisionByZerox: > This could potentially be used for:
> 
> https://github.com/faker-js/faker/blob/3f49ac1e38263064372fc1970d08f23e00cd580b/src/modules/image/providers/unsplash.ts#L40-L45
> 
> So we should follow the normal deprecation procedure.

In this case, the `imageUrl()` `category` parameter should be made more strict with a `TemplateLiteral<currentCategoryConst>`

ST-DDT: > In this case, the `imageUrl()` `category` parameter should be made more strict with a `TemplateLiteral<currentCategoryConst>`

The category is an external thing, so I don't know how feasible this is.
If there is a limited set of values that are supported, then we could use them.

> The parameter is fine IMO.

Yes

> Do you want the const itself to somehow be deprecated


Replace the field with a getter that is deprecated.

import-brain: https://github.com/faker-js/faker/pull/1046/commits/94e655dd9ef3abf4952a6e96b10a5e6acab7f44a#diff-cca52b6b29e5adca3e58806e1cfcf625710fc61b723a23ad8b86468880869f7fR18-R37

You mean like this?

src/modules/image/providers/unsplash.ts
```diff
@@ -1,21 +1,40 @@
...
-  ];
-
   constructor(private readonly faker: Faker) {}
 
+  /**
+   * Generates a valid unsplash category.
+   *
+   * @deprecated
+   */
+  category(): Category {
```
xDivisionByZerox, line 23: The `get` keyword is missing. See [typescript getter/setter].(https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters)

Otherwise, you would convert the field/property to a callable function and thus have a breaking change.

src/modules/image/providers/unsplash.ts
xDivisionByZerox: No, you just removed the current field and return ONE element instead of the entire array. This is a breaking behavior. We still need to support the current syntax (`faker..image.unsplash.categories`). For your (now function) you would need to call `faker.image.unsplash.category()`. You have provided a new feature, not a rework.

import-brain: Ohhhh, now I see what you mean, sorry for the confusion. I'll change it.

src/modules/image/providers/unsplash.ts
```diff
@@ -4,15 +4,14 @@ import type { Faker } from '../../..';
...
-    'objects',
-    'buildings',
-  ];
+  /**
+   * Gets the list of supported unsplash categories.
+   *
+   * @deprecated
+   */
+  get categories(): string[] {
+    return ['food', 'nature', 'people', 'technology', 'objects', 'buildings'];
```
xDivisionByZerox, line 13: Still needs the deprecation log by calling the internal `deprecated()` method.

import-brain:
@xDivisionByZerox Would you prefer I write a test case for the deprecation log, or is it fine in this case?

xDivisionByZerox:
Normally our procedure when deprecating something consists of 3 steps:
1. Use the internal `deprecated` function
2. Add a `@deprecated` JSDoc parameter
3. Add a test case that ensures that the deprecation log is being printed to the console

So I guess to be exact a test would be required, if you don't mind.

import-brain:
> Normally our procedure when deprecating something consists of 3 steps:
> 
> 1. Use the internal `deprecated` function
> 2. Add a `@deprecated` JSDoc parameter
> 3. Add a test case that ensures that the deprecation log is being printed to the console
> 
> So I guess to be exact a test would be required, if you don't mind.

Sure, I'll get on it :)

xDivisionByZerox (dismissed):

ST-DDT (dismissed):

src/modules/image/providers/unsplash.ts
```diff
@@ -1,18 +1,23 @@
...
-  ];
+  /**
+   * Gets the list of supported unsplash categories.
+   *
+   * @deprecated
+   */
+  get categories(): string[] {
+    deprecated({
+      deprecated: 'faker.image.unsplash.categories',
+      since: 'v7.2.0',
```
Shinigami92, line 16:
```suggestion
      since: 'v7.3.0',
```

Shinigami92 (changes requested):

ST-DDT (dismissed):

test/image.spec.ts
```diff
@@ -155,6 +155,20 @@ describe('image', () => {
   });
 
   describe('unsplash', () => {
+    describe('categories', () => {
```
Shinigami92, line 158: We discussed in several other PRs that we don't need these deprecation tests anymore
So we should revert this test file

Shinigami92 (dismissed):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #1048, opened 2022-06-09
feat: mime-db as optional peer dependency

Shinigami92:


src/modules/system/index.ts
```diff
@@ -45,6 +48,21 @@ export class System {
...
     }
+
+    if (!mimeTypes) {
+      try {
+        // @ts-expect-error: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'. ts(1323)
+        import('mime-db')
+          .then((mod) => {
+            mimeTypes = mod.default;
+          })
+          .catch((error) => {
```
xDivisionByZerox, line 59: If we catch here we shouldn't need a try-catch block around this.

src/modules/system/index.ts
```diff
@@ -80,7 +98,11 @@ export class System {
...
    */
   mimeType(): string {
-    const mimeTypeKeys = Object.keys(this.faker.definitions.system.mimeTypes);
+    if (!mimeTypes) {
+      throw new FakerError('mime-db is required');
+    }
+
+    const mimeTypeKeys = Object.keys(mimeTypes);
 
     return this.faker.helpers.arrayElement(mimeTypeKeys);
```
xDivisionByZerox, line 107:
```suggestion
    return this.faker.helpers.objectKey(mimeTypes);
```

src/modules/system/index.ts
```diff
@@ -135,12 +160,14 @@ export class System {
...
    */
   fileExt(mimeType?: string): string {
+    if (!mimeTypes) {
+      throw new FakerError('mime-db is required');
+    }
+
     if (typeof mimeType === 'string') {
-      const mimes = this.faker.definitions.system.mimeTypes;
-      return this.faker.helpers.arrayElement(mimes[mimeType].extensions);
+      return this.faker.helpers.arrayElement(mimeTypes[mimeType].extensions);
```
xDivisionByZerox, line 168: Not this PR, but this can potentially return `undefined`.

Shinigami92:
At least for the esm version, this doesn't seem to work at all :eyes: 

Screenshot_20220609_211734

The file gets just embedded into the dist :eyes: :shrug:

src/modules/system/index.ts
Shinigami92: I'm testing... this is a draft PR

src/modules/system/index.ts
```diff
@@ -45,6 +60,26 @@ export class System {
...
+      try {
+        import('mime-db')
+          .then((mod) => {
+            mimeTypes = mod.default;
+          })
+          .catch((error) => {
+            console.warn(error);
+          });
+      } catch (error) {
+        console.warn(error);
```
ST-DDT, line 80: Could we use like 10 common mime types as default? If the db is absent, so that at least the file extension work without the entire lib?

ST-DDT (changes requested):

src/modules/system/index.ts
xDivisionByZerox: That was exactly the proposal in the [poll](https://github.com/faker-js/faker/discussions/1009) we made.

ST-DDT:
> I have a new idea 🤔
> We could try to allow the user to pass mime-db object to faker.system from outside.

I will only allow this if we treat it as locale data. 
I don't want multiple ways to provide user-specific/custom data to faker (asside from method params).
The one and only way should be locale data.

Shinigami92:
I hoped for an easy solution that could also deliver a good DX.
I put to much work from my side to tackle this task and in the end, I don't use this functionality for my own projects.
Therefore if someone else need this, want to provide a better DX and reduce the bundle size by solving this, feel free to tackle this issue and earn credits for that.

Issue #976, opened 2022-05-20
Proposal: enhance JSDoc deprecate parameter description


xDivisionByZerox:
### Clear and concise description of the problem

As a developer using faker I want to see some intel about a deprecated method in intellisense. This could help me to figure out what I should use instead without waiting for a log that tells me.

### Suggested solution

Add meaningful documentation to JSDocs `deprecated` parameter.
 
Like:
```
/**
  * @deprecated
  * Use faker.name.firstName instead.
  * /
faker.name.firstname()
```
This would result in a more excellent information box resulting and in a better developer experience by not only telling me that a method is deprecated but telling me what I should do instead.
Like:
![image](https://user-images.githubusercontent.com/77127505/169600210-89bb69c2-224f-4659-815e-622dd57bf977.png)

### Alternative

Don't do anything

### Additional context

[rxjs is doing the same for their operators](https://github.com/ReactiveX/rxjs/blob/d25b3528a4168c9ba76f81ec84ee20f6c241903b/src/internal/operators/combineLatest.ts) (just ONE reference)

ST-DDT:
Should we include that also in our api docs page?

xDivisionByZerox:
> Should we include that also in our api docs page?

I think that would be a great idea. I don't use the API docs TBH so I didn't realize they could be in need of this feature as well.

import-brain:
> Should we include that also in our api docs page?

Yeah, I also think this would be a great idea.

xDivisionByZerox:
Currently, we provide this information via the `@see` parameter. Would we move this into the `@deprecated` parameter description and no longer use it (`@see`) for deprecations?

import-brain:
> Currently, we provide this information via the `@see` parameter. Would we move this into the `@deprecated` parameter description and no longer use it (`@see`) for deprecations?

Yeah, I think it would be more clear if we moved the information to the `@deprecated` parameter and wrote something like "Use this function instead".

ST-DDT:
If it generates a clickable link for the docs, sure go for it.

xDivisionByZerox:
Do you mean in the docs? Because currently there is no link in the JSDoc in my IDE :/

ST-DDT:
I'm referring to the api docs website:
Example: https://fakerjs.dev/api/address.html#cityPrefix

Shinigami92:
Here an image, as the link posted above by @ST-DDT will get stale some day

[...]

PR #1053, opened 2022-06-10
chore: deprecate address.streetPrefix and suffix

ST-DDT:
The street prefix and suffix methods only generate parts of the data and not all of the locales support these.
IMO the methods aren't useful by themselves and thus should be removed.


xDivisionByZerox:
IMO either keep them as util functions (for custom tooling) or remove them completely, like don't advertise the usage with `faker.fake()`. Doesn't make sense to me to tell the user "hey can still use this, but not as nice as it was for now" (write a "complex" non-typesafe string).

ST-DDT:
> IMO either keep them as util functions (for custom tooling) or remove them completely, like don't advertise the usage with `faker.fake()`.

I do want to remove them, but I want to give everyone a chance to migrate.
I could switch to `faker.helpers.arrayElement(faker.definitions.address.street_prefix)`, but that is way longer.
The current usages of these methods are mostly limited to `faker.fake()`, so thats why I have chosen that one.

xDivisionByZerox:
> I do want to remove them, but I want to give everyone a chance to migrate.
> I could switch to `faker.helpers.arrayElement(faker.definitions.address.street_prefix)`, but that is way longer.
> The current usages of these methods are mostly limited to `faker.fake()`, so thats why I have chosen that one.

Nah, I think you misunderstood me. If we deprecate these functions, why leave the definitions?

src/modules/address/index.ts
```diff
@@ -216,10 +216,21 @@ export class Address {
...
    * @example
    * faker.address.streetSuffix() // 'Streets'
+   *
+   * @deprecated
    */
   streetSuffix(): string {
+    deprecated({
+      deprecated: 'faker.address.streetSuffix()',
+      proposed:
+        "faker.address.street() or faker.fake('{{address.street_suffix}}')",
```
xDivisionByZerox, line 230:
```suggestion
        "faker.address.street()",
```

I wouldn't want to advertise the direct usage of definitions.

ST-DDT:
> If we deprecate these functions, why leave the definitions?

To give users a hint how they should structure their definitions to be compatible with our format/our existing locales.
If possible I still want all locales to have the same set of possible files.
But there might be other/better ways to achieve that (e.g. a file-name whitelist, so it's easier to tell when a file is added that doesn't match our existing files).

test/random.spec.ts
```diff
@@ -33,7 +33,7 @@ const seededRuns = [
       locale: 'ur',
       numeric: '9',
       word: 'Division',
-      words: 'Turnpike Frozen Handcrafted',
+      words: 'functionalities Frozen Baby',
```
Shinigami92, line 36: 🥶

Shinigami92 (approved):
IMO it is okay to remove / hide this from the end-user as I would not know what a street suffix/prefix would even be 🤷
We also didn't explain it anywhere but just provide it
In the end Faker should be a tool to generate meaningful data, and streetnames are totally okay, but what is a street prefix/suffix?

xDivisionByZerox (dismissed):

import-brain (approved):

xDivisionByZerox (approved):

pkuczynski (dismissed):

src/modules/address/index.ts
```diff
@@ -218,10 +218,20 @@ export class Address {
   /**
    * Returns a random localized street suffix.
    *
+   * @see faker.address.street
+   *
    * @example
    * faker.address.streetSuffix() // 'Streets'
+   *
+   * @deprecated
```
Shinigami92, line 226: I might be wrong, but did we discuss to start writing human readable messages on deprecation JSDoc tags?
https://github.com/faker-js/faker/issues/976

xDivisionByZerox: Yes, good catch. See our [contribution guidelines, section deprecation workflow]:

> Afterwards, add a @deprecated parameter to the end of the JSDoc with a human readable description message with a suitable replacement for the deprecated function.

ST-DDT: Done

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1059, opened 2022-06-11
feat(internet)!: ip now returns ipv4 and ipv6

xDivisionByZerox:
Is listed in #1044.

This is a breaking change and should not be merged before the next major release.


test/internet.spec.ts
```diff
@@ -194,7 +198,7 @@ describe('internet', () => {
           const [prefix, suffix] = email.split('@');
 
           expect(prefix).toMatch(/^思源_唐3/);
-          expect(faker.definitions.internet.free_email).toContain(suffix);
+          expect(faker.definitions.internet?.free_email).toContain(suffix);
         });
 
         it('should return an email with special characters', () => {
```
xDivisionByZerox, line 204: The correct signature for `faker.internet.email()` has `string | undefined` as 3rd parameter. Is the `null` here intentional or from a legacy signature?

test/internet.spec.ts
```diff
@@ -153,7 +157,7 @@ describe('internet', () => {
           expect(email).toSatisfy(validator.isEmail);
 
           const [, suffix] = email.split('@');
-          expect(faker.definitions.internet.free_email).toContain(suffix);
+          expect(faker.definitions.internet?.free_email).toContain(suffix);
```
ST-DDT, line 160: Is this warning new? I havent seen any lint errors related to this before.

Shinigami92:
> I also updated the access to any definition.internet with an optional chaining operator since my IDE was complaining that the property is potentially undefined.

@xDivisionByZerox I would prefer that you extract this into its separate PR
But I also never had an issue with that :thinking: Also TypeScript check didn't complain yet :thinking: 
So is it maybe just your IDE? Did you miss to switch e.g. to the installed TypeScript instead of the embedded IDE TS version? Sometimes that can lead to such issues.

xDivisionByZerox:
> > I also updated the access to any definition.internet with an optional chaining operator since my IDE was complaining that the property is potentially undefined.
> 
> @xDivisionByZerox I would prefer that you extract this into its separate PR But I also never had an issue with that 🤔 Also TypeScript check didn't complain yet 🤔 So is it maybe just your IDE? Did you miss to switch e.g. to the installed TypeScript instead of the embedded IDE TS version? Sometimes that can lead to such issues.

I have no problems providing an additional PR regarding this topic, so we can focus on the logic changes in this PR.

test/internet.spec.ts
xDivisionByZerox: I mean you can just have a look at the definitions
https://github.com/faker-js/faker/blob/74e93811e578ec610dfd916d03276a59475f9e7a/src/definitions/definitions.ts#L20-L23

The `Partial<T>` lets typescript know that the entire object has optional keys. Or am I missing something?
I can over any definition's entry in my IDE, and it always tells me that the type is `T | undefined` (where `T` is the module you are currently working with, example: InternetDefinitions)

xDivisionByZerox:
I did revert the optional chaining. @Shinigami92 @ST-DDT please rereview.

ST-DDT (dismissed):

ST-DDT:
> I have no problems providing an additional PR regarding this topic, so we can focus on the logic changes in this PR.

IMO there is no need to change it.
- There is no lint warning
- It doesnt make the code more readable or safe.
- Its only for tests

test/internet.spec.ts
```diff
@@ -354,9 +355,19 @@ describe('internet', () => {
       });
 
       describe('ip()', () => {
-        it('should return a random IPv4 address with four parts', () => {
+        it('should return a random IPv4 or IPv6 address', () => {
           const ip = faker.internet.ip();
 
+          expect(ip).toBeTruthy();
+          expect(ip).toBeTypeOf('string');
+          expect(ip).toSatisfy((value: string) => validator.isIP(value));
```
ST-DDT, line 363:
```suggestion
          expect(ip).toSatisfy(validator.isIP);
```

test/internet.spec.ts
```diff
@@ -237,7 +238,7 @@ describe('internet', () => {
           expect(username).toBeTypeOf('string');
           expect(username).includes('Aiden');
           expect(username).toMatch(
-            /^Aiden((\d{2})|([._]Harann\d{2})|([._](Harann)))/
+            /^Aiden((\d{1,2})|([._]Harann\d{1,2})|([._](Harann)))/
```
ST-DDT, line 241: Did we miss these in the last PR?

xDivisionByZerox: Yes, since the number generated for the user name is between 0 and 99 (inclusive) there is a 10% chance if the number is only one digit. No test was running into that case, so they did pass.

ST-DDT (dismissed):

ejcheng (dismissed):

ST-DDT (dismissed):

ejcheng (dismissed):

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #1061, opened 2022-06-11
feat(location)!: `nearbyGPSCoordinate` returns number tuple

xDivisionByZerox:
Discussion reference #1058.

Changes the return type of `address.nearbyGPSCoordinate()` from a string tuple to a number tuple.

This is a breaking change and should not be merged before the next major release.


test/address.spec.ts
```diff
@@ -487,42 +487,39 @@ describe('address', () => {
...
-                  latitude1,
-                  longitude1,
-                  latitude2,
-                  longitude2,
-                  isMetric
-                );
-                expect(actualDistance).toBeLessThanOrEqual(radius);
-              }
-            );
+            it(`should return random gps coordinate within a distance of another one (${JSON.stringify(
```
xDivisionByZerox, line 490: I removed the rerun of the test (`.each(times(5))`) because it doesn't bring any value other than extending the test time and bumping the number of test cases, IMO. 

If this was to catch some edge cases, the testcases designed should be changed instead of run the same test multiple times.

src/modules/address/index.ts
```diff
@@ -439,19 +439,19 @@ export class Address {
...
   nearbyGPSCoordinate(
     coordinate?: [latitude: number, longitude: number],
     radius: number = 10,
     isMetric: boolean = false
-  ): [latitude: string, longitude: string] {
+  ): [latitude: number, longitude: number] {
     // If there is no coordinate, the best we can do is return a random GPS coordinate.
     if (coordinate === undefined) {
-      return [this.latitude(), this.longitude()];
+      return [parseFloat(this.latitude()), parseFloat(this.longitude())];
```
Shinigami92, line 454: Do we also want to change return type of `latitude` and `longitude`?

xDivisionByZerox: I wanted to provide for them as well. But since they are separate methods from this one I wanted to separate them and make this one depend on the other PR (or something like this).

IMO returning strings doesn't make sense for this method, but I think we are on the same side for this.

ST-DDT: You can remove the Todo for string inputs too.

ST-DDT (commented):
The description text should include a hint that the user might want to use toFixed on the result and some reference how much another digit is in precision. E.g. 1 degree = 100 km. 1km = 0.01degree/ToFixed(2). We dont have to use precises values but some rough estimates. I guessed these values so you should probably verify them first.

src/modules/address/index.ts
xDivisionByZerox: I wanted to leave them in since this PR technically doesn't fix what the todo states. But if you say so I will gladly remove it.

xDivisionByZerox:
> The description text should include a hint that the user might want to use toFixed on the result and some reference how much another digit is in precision. E.g. 1 degree = 100 km. 1km = 0.01degree/ToFixed(2). We dont have to use precises values but some rough estimates. I guessed these values so you should probably verify them first.

Don't know about this. That wasn't the case before either. IMO, this is on the user side, since our responsibility is the (fake) data generation not being a wiki on how specific data should be handled. If someone requests GPS coordinates I would expect them to know how to work with them.

src/modules/address/index.ts
```diff
@@ -491,7 +490,7 @@ export class Address {
     // Box longitude [-180°, 180°]
     newCoordinate[1] = (((newCoordinate[1] % 360) + 540) % 360) - 180;
 
-    return [newCoordinate[0].toFixed(4), newCoordinate[1].toFixed(4)];
+    return [newCoordinate[0], newCoordinate[1]];
```
ST-DDT, line 493: In the longitude and latitude PR you return the values with a specific presion.
IMO we should do the same here.

xDivisionByZerox: So you mean adding a parameter? Because currently, the user has no possibility to provide precision. I'm just working with what already exists here. If that is what you mean, I would want to provide an additional PR since this would be a feature (not like currently refactor) and is more comprehensible on the autogenerated changelog.

Otherwise, please elaborate on what you mean.

ST-DDT: Its okay to open a separate PR for it.
IMO the values should have the same precision regardless of whether the are genrated randomly (4) or nearby an existing location (10?).

src/modules/address/index.ts
```diff
@@ -439,19 +439,18 @@ export class Address {
    * @param isMetric If `true` assume the radius to be in kilometers. If `false` for miles. Defaults to `false`.
    *
    * @example
-   * faker.address.nearbyGPSCoordinate() // [ '33.8475', '-170.5953' ]
-   * faker.address.nearbyGPSCoordinate([33, -170]) // [ '33.0165', '-170.0636' ]
-   * faker.address.nearbyGPSCoordinate([33, -170], 1000, true) // [ '37.9163', '-179.2408' ]
+   * faker.address.nearbyGPSCoordinate() // [ 33.8475, -170.5953 ]
+   * faker.address.nearbyGPSCoordinate([33, -170]) // [ 33.0165, -170.0636 ]
+   * faker.address.nearbyGPSCoordinate([33, -170], 1000, true) // [ 37.9163, -179.2408 ]
```
Shinigami92, line 444: Do we need to update these examples due to the output numbers could be now much more precise? (please see also https://github.com/faker-js/faker/pull/1064#discussion_r895224981)

xDivisionByZerox: I mean I did the bare minimum to achieve the signature change. But you are somewhat right. I'll provide some more diverse examples

ST-DDT (dismissed):

ST-DDT:
Considering the other PR, should we rather introduce a `format` parameter?

xDivisionByZerox:
> Considering the other PR, should we rather introduce a `format` parameter?

It doesn't make sense to me that coordinates should be strings. We simply allow the input of stringified numbers due to the nature of JS. At least that's what I like to think.

ST-DDT:
Cross referencing related PRs: 
- https://github.com/faker-js/faker/pull/1066
- https://github.com/faker-js/faker/pull/1064
- https://github.com/faker-js/faker/pull/1061

src/modules/address/index.ts
ST-DDT: Please also address this.

src/modules/address/index.ts
Shinigami92: @xDivisionByZerox Please extract it to a ticket then, so e don't miss it in v8.0

Shinigami92:
I eagerly put the PR title to scope `location` as we will solve 
- #1344

in v8.0 and this will then be grouped together

src/modules/address/index.ts
Shinigami92: Nevermind :slightly_smiling_face: 

- #1064

src/modules/address/index.ts
Shinigami92: The default precision of lat and long functions is `4`!, so actually this is correct :grimacing:

Shinigami92: Or ... maybe we just should add a `precision` option? 🙃
The lat and long function also do have this option. :thinking:

xDivisionByZerox: This is not in the scope of this PR...

Was already suggested by @ST-DDT [here](https://github.com/faker-js/faker/pull/1061#discussion_r895200965)

Shinigami92: > This is not in the scope of this PR...

Yeah yes yes I know :sweat_smile: 
I just speak out loud

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #1063, opened 2022-06-12
refactor(phone): rename phoneNumber to number

xDivisionByZerox:
Listed as a todo in #1044.

This PR does the following things:
- move the implementation from `phone.phoneNumber()` into a new function called `phone.number()`
- deprecate `phone.phoneNumber()` via the internal deprecated method
- add a `@deprecated` JSDoc parameter to the `phone.phoneNumber()` signature documentation
- duplicates the test checks from `phoneNumber()` to `number()`
- add a test case for `phoneNumber()` deprecation log

### DIFF
--- a/src/modules/phone/index.ts
+++ b/src/modules/phone/index.ts
@@ -20,6 +20,8 @@ export class Phone {
    *
    * @param format Format of the phone number. Defaults to a random phone number format.
    *
+   * @see faker.phone.number
+   *
    * @example
    * faker.phone.phoneNumber() // '961-770-7727'
    * faker.phone.phoneNumber('501-###-###') // '501-039-841'
@@ -43,9 +45,9 @@ export class Phone {
    * @param format Format of the phone number. Defaults to a random phone number format.
    *
    * @example
-   * faker.phone.phoneNumber() // '961-770-7727'
-   * faker.phone.phoneNumber('501-###-###') // '501-039-841'
-   * faker.phone.phoneNumber('+48 91 ### ## ##') // '+48 91 463 61 70'
+   * faker.phone.number() // '961-770-7727'
+   * faker.phone.number('501-###-###') // '501-039-841'
+   * faker.phone.number('+48 91 ### ## ##') // '+48 91 463 61 70'
    */
   number(format?: string): string {
     format =
--- a/src/modules/phone/index.ts
+++ b/src/modules/phone/index.ts
@@ -24,15 +24,17 @@ export class Phone {
    * faker.phone.phoneNumber() // '961-770-7727'
    * faker.phone.phoneNumber('501-###-###') // '501-039-841'
    * faker.phone.phoneNumber('+48 91 ### ## ##') // '+48 91 463 61 70'
+   *
+   * @deprecated
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `number()`
   phoneNumber(format?: string): string {
-    format =
-      format ??
-      this.faker.helpers.arrayElement(
-        this.faker.definitions.phone_number.formats
-      );
-    return this.faker.helpers.replaceSymbolWithNumber(format);
+    deprecated({
+      deprecated: 'faker.phone.phoneNumber()',
+      proposed: 'faker.phone.number()',
+      since: 'v7.2.0',
+      until: 'v8.0.0',
+    });
+    return this.faker.phone.number(format);
   }
 
   /**
--- a/src/modules/phone/index.ts
+++ b/src/modules/phone/index.ts
@@ -36,6 +36,25 @@ export class Phone {
   }
 
   /**
+   * Generates a random phone number.
+   *
+   * @param format Format of the phone number. Defaults to a random phone number format.
+   *
+   * @example
+   * faker.phone.phoneNumber() // '961-770-7727'
+   * faker.phone.phoneNumber('501-###-###') // '501-039-841'
+   * faker.phone.phoneNumber('+48 91 ### ## ##') // '+48 91 463 61 70'
+   */
+  number(format?: string): string {
+    format =
+      format ??
+      this.faker.helpers.arrayElement(
+        this.faker.definitions.phone_number.formats
+      );
+    return this.faker.helpers.replaceSymbolWithNumber(format);
+  }
+
+  /**
    * Returns phone number in a format of the given index from `faker.definitions.phone_number.formats`.
    *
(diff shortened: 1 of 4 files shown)


xDivisionByZerox:
The ci is failing with an error I couldn't figure out. 
```
FAIL  test/scripts/apidoc/examplesAndDeprecations.spec.ts > examples and deprecations > Phone > number
AssertionError: expected "warn" to not be called at allexpected "warn" to not be called at all
Received: 
    1st warn call:
    Array [
❯test/scripts/apidoc/examplesAndDeprecations.spec.ts:107:26
      "[@faker-js/faker]: faker.phone.phoneNumber() is deprecated since v7.2.0 and will be removed in v8.0.0. Please use faker.phone.number() instead.",
    ]
105|       } else {
106|         for (const spy of consoleSpies) {
107|           expect(spy).not.toHaveBeenCalled();
   | ^
108|         }
109|       }
    2nd warn call:
    Array [
      "[@faker-js/faker]: faker.phone.phoneNumber() is deprecated since v7.2.0 and will be removed in v8.0.0. Please use faker.phone.number() instead.",
Test Files 1 failed | 32 passed (33)
    ]
    3rd warn call:
     Tests 2 failed | 37156 passed | 9 todo (37167)
    Array [
      "[@faker-js/faker]: faker.phone.phoneNumber() is deprecated since v7.2.0 and will be removed in v8.0.0. Please use faker.phone.number() instead.",
    ]
Number of calls: 3
```

The error states that `phone.number()` is logging to the console, but if you have a look at the message (and the source code) you can clearly see that `phone.phoneNumber()` is the one doing the log. 

What am I missing?

src/modules/phone/index.ts
```diff
@@ -24,9 +24,30 @@ export class Phone {
    * faker.phone.phoneNumber() // '961-770-7727'
    * faker.phone.phoneNumber('501-###-###') // '501-039-841'
    * faker.phone.phoneNumber('+48 91 ### ## ##') // '+48 91 463 61 70'
+   *
+   * @deprecated
```
ST-DDT, line 28: Add `@see`

ST-DDT (changes requested):

src/modules/phone/index.ts
```diff
@@ -24,9 +24,30 @@ export class Phone {
...
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `number()`
   phoneNumber(format?: string): string {
+    deprecated({
+      deprecated: 'faker.phone.phoneNumber()',
+      proposed: 'faker.phone.number()',
+      since: 'v7.2.0',
```
ST-DDT, line 34: The next version is 7.3

ST-DDT (changes requested):

src/modules/phone/index.ts
```diff
@@ -24,9 +24,30 @@ export class Phone {
...
+    return this.faker.phone.number(format);
+  }
+
+  /**
+   * Generates a random phone number.
+   *
+   * @param format Format of the phone number. Defaults to a random phone number format.
+   *
+   * @example
+   * faker.phone.phoneNumber() // '961-770-7727'
```
ST-DDT, line 46: You have to update the examples.

ST-DDT (changes requested):

src/modules/phone/index.ts
xDivisionByZerox: done eaa71c0

ST-DDT (changes requested):
Faker.fake still uses this method in the example.

xDivisionByZerox:
> Faker.fake still uses this method in the example.

I understand, I usually search with "Find All References" when replacing something. Obviously, these references do not get found.

ST-DDT:
> > Faker.fake still uses this method in the example.
> 
> I understand, I usually search with "Find All References" when replacing something. Obviously, these references do not get found.

Thats why the test is there 😉

ST-DDT (dismissed):

xDivisionByZerox:
> Thats why the test is there 😉

Yeah, but I was totally lost when looking at the error message since I wasn't able to find any (direct) call reference. We could potentially extend the message for these cases (faker.fake)? Not this PR tho.

test/phone.spec.ts
```diff
@@ -109,6 +119,28 @@ describe('phone', () => {
 
           expect(phoneNumber).toMatch(/\d/);
         });
+
+        it('should log a deprecation message to the console', () => {
```
Shinigami92, line 123: @ST-DDT correct me if I'm wrong, but I thought you wrote a test suite that automatically checks for these deprecations 🤔 
So I assume we don't need this test (?)

xDivisionByZerox: I'm not aware of a dx feature like this. And last time I did a deprecation this kind of test was required. But yeah, wait for @ST-DDT to give some insights.

ST-DDT: Yes, the deprecations and examples test that failed before contains the deprecation checks.
I consider the extra deprecation tests optional and wouldnt add them myself.

xDivisionByZerox: Good to know :/

src/modules/phone/index.ts
```diff
@@ -20,13 +20,36 @@ export class Phone {
...
+   *
+   * @deprecated
    */
-  // TODO @pkuczynski 2022-02-01: simplify name to `number()`
   phoneNumber(format?: string): string {
+    deprecated({
+      deprecated: 'faker.phone.phoneNumber()',
+      proposed: 'faker.phone.number()',
+      since: 'v7.3.0',
+      until: 'v8.0.0',
```
Shinigami92, line 37: New schema `7.3` and `8.0`? Or was that a not yet merged PR 🤔

test/phone.spec.ts
```diff
@@ -111,6 +121,14 @@ describe('phone', () => {
         });
       });
 
+      describe('number()', () => {
+        it('should return a random phoneNumber with a random format', () => {
+          const phoneNumber = faker.phone.number();
+
+          expect(phoneNumber).toMatch(/\d/);
```
Shinigami92, line 128: IMO this is not a good check, it could just contain one digit and it would be valid

Shinigami92 (changes requested):

test/phone.spec.ts
xDivisionByZerox: Yes, I'm with you. But I don't see the responsibility to change this in this PR, since it only renames the `phone.number()` function and doesn't change any implementation logic.

Shinigami92 (approved):

ST-DDT (approved):

PR #1071, opened 2022-06-14
feat: throw error on unknown locale

Shinigami92:


xDivisionByZerox (approved):

xDivisionByZerox:
Should this be marked as `breaking change`? As far as I can tell this can lead to unhandled errors on the end-user side.

Shinigami92:
> Should this be marked as `breaking change`? As far as I can tell this can lead to unhandled errors on the end-user side.

I would say no, because otherwise it would should just fail later when accessing a function that depends on it as you would get a "null-pointer error"
This brings the error just a little bit close upfront and you know whats going on

PR #1080, opened 2022-06-16
refactor(name.findName): allow prefix and suffix

xDivisionByZerox:
Refactors the behavior of `faker.name.findName()` to possible return a full name with `prefix` AND `suffix`.
This also fixes a to-do list in #1044 about the unwanted fallthrough in the switch statement.


src/modules/name/index.ts
```diff
@@ -148,33 +148,33 @@ export class Name {
...
-        }
-      // TODO @Shinigami92 2022-01-21: Not sure if this fallthrough is wanted
-      // eslint-disable-next-line no-fallthrough
-      case 1:
-        suffix = this.suffix();
-        if (suffix) {
-          return `${firstName} ${lastName} ${suffix}`;
-        }
+    const nameParts: string[] = [];
+    const prefix = this.faker.helpers.maybe(() => this.prefix(gender), {
```
xDivisionByZerox, line 158: Just to make sure. The description of `faker.helpers.maybe` states that the callback is invoked with the provided possibility and otherwise returns undefined. But the return type of `maybe` is always the same as the one of the callback. Is that a design decision to work around possible undefined values or is this an implementation error?

ST-DDT (dismissed):
Sadly we don't hit any of these cases in our seeded tests.

src/modules/name/index.ts
Shinigami92: Looked into the implement PR, but didn't read anything about written down :thinking: 
I slightly remember that it could be a design decision somehow, but yeah it might be wrong
Would you try open a PR to _fix_ this?

Shinigami92 (approved):

src/modules/name/index.ts
xDivisionByZerox: I can do some tests and see if this would "break" our code base. But since the method is exposed to the public it should be typed correctly (I guess). I will do some investigations and open another PR later.

Shinigami92 (approved):

ST-DDT (approved):

PR #1052, opened 2022-06-10
build: bump lowest supported node version to node14.13

Shinigami92:
This bumps the target to the lowest supported target

https://esbuild.github.io/api/#target

It could be that this doesn't change anything, but we are more safe with that

PR #1085, opened 2022-06-17
fix: lower target to support Webpack 4

### LINKED ISSUE #1073: Error: Webpack Compilation Error ./node_modules/@faker-js/faker/dist/esm/chunk-4J2PVEV7.mjs 1:1430
### Describe the bug

Error: Webpack Compilation Error
./node_modules/@faker-js/faker/dist/esm/chunk-Y64C4BL4.mjs 1:1430
Module parse failed: Unexpected token (1:1430)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders

### Reproduction

1. add @faker-js/faker to package.json and npm install
2. import { faker } from '@faker-js/faker'

### Additional Info

version: 7.2.0

JessicaSachs:
This PR  would fix #1073 and target es2019 instead of dropping support for Webpack 4 (which I don't think we should do -- a lot of users are still stuck on Webpack 4, despite how long it has been out).

We're mainly having issues because we use nullish coalescing internally. This requires a target of es2019 instead of es2020. https://esbuild.github.io/content-types/#javascript

Also would fix https://github.com/cypress-io/cypress/issues/21874

### DIFF
--- a/scripts/bundle.ts
+++ b/scripts/bundle.ts
@@ -7,6 +7,8 @@ console.log('Building dist for node (cjs)...');
 
 // Generate entry-points for cjs compatibility
 const localeDir = 'locale';
+const target = ['ES2019', 'node14.6'];
+
 if (existsSync(localeDir)) {
   rmSync(localeDir, { recursive: true, force: true });
 }
@@ -33,7 +35,7 @@ buildSync({
   // splitting: true, // Doesn't work with cjs
   format: 'cjs',
   platform: 'node',
-  target: 'es2019',
+  target,
 });
 
 console.log('Building dist for node type=module (esm)...');
@@ -48,6 +50,6 @@ buildSync({
   minify: true,
   splitting: true,
   format: 'esm',
-  target: 'es2019',
+  target,
   outExtension: { '.js': '.mjs' },
 });
--- a/scripts/bundle.ts
+++ b/scripts/bundle.ts
@@ -33,7 +33,7 @@ buildSync({
   // splitting: true, // Doesn't work with cjs
   format: 'cjs',
   platform: 'node',
-  target: 'node14',
+  target: 'es2019',
 });
 
 console.log('Building dist for node type=module (esm)...');
@@ -48,6 +48,6 @@ buildSync({
   minify: true,
   splitting: true,
   format: 'esm',
-  target: 'node14',
+  target: 'es2019',
   outExtension: { '.js': '.mjs' },
 });
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2019",
+    "target": "ES2019",
     "moduleResolution": "Node",
     "rootDir": "src",
     "outDir": "dist",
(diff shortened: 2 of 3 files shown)


scripts/bundle.ts
```diff
@@ -33,7 +33,7 @@ buildSync({
   // splitting: true, // Doesn't work with cjs
   format: 'cjs',
   platform: 'node',
-  target: 'node14',
+  target: 'es2019',
```
Shinigami92, line 36: maybe we should revert this, as everyone reporting bugs wrote that they have problems with the esm version, but this is the cjs version and I got 0 issues about that

scripts/bundle.ts
```diff
@@ -48,6 +48,6 @@ buildSync({
   minify: true,
   splitting: true,
   format: 'esm',
-  target: 'node14',
+  target: 'es2019',
```
Shinigami92, line 51: It's possible to set multiple targets and the lowest one wins. So we should use e.g. this:
```suggestion
  target: ['es2019', 'node14.6'],
```
Sadly I already tried node14.6 and the output still was not compatible with our tests when running on node14.6, but it was green up from node14.13+ 🤔

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "es2019",
```
Shinigami92, line 3: I personally prefer capital ES, because this is the auto-suggested value when pressing ctrl+space in vscode 👀 
```suggestion
    "target": "ES2019",
```
But will this somehow affect our code? I mean, esbuild has its own target, so if I'm not wrong, this line doesn't affect anything 🤔

Shinigami92 (changes requested):
We should add a CI run for node14.6 or at least the lowest target we support
See also #1052 for my own tries

Shinigami92:
For tomorrows meeting: I personally find it totally weird that we need to support >2 years old out of date packages, and even if this lowers our target I lost trust in esbuild a bit, because of https://github.com/evanw/esbuild/issues/2315
They are saying that it was because we used esno/tsx, but even now we build it to use [native js](https://github.com/faker-js/faker/pull/1052/files#diff-2c253df4d2d7d76359657b571f3f3e82c573214e78bb77aad3eb86af9ea21da1) and it produced the same result in my other branch 🤷
I also don't like that we cannot see which tools and packages out there in the world are possibly not compatible to es20XX and which not. We have a potential table about node versions, but from where can we find out which tools support version xy?
That said, we said that we do not want to support node v12 because it got EOL and so node v14 is the latest "legacy" LTS and it fully supports es2020 🤷
I mean, it is just weird that folks depend on webpack v4 that is out of date, don't get updates (as far as I know) and is not compatible to modern syntax anymore
Is webpack v4 now the new IE11 or what?!

... oops, this turned into a little rant ... 😅

Shinigami92 (approved):

JessicaSachs:
Shini and I spoke. We'll be supporting es2019 + Webpack 4.

xDivisionByZerox (approved):
I totally do not understand how this works but if you ensure that it does I'm fine with this.

PR #1086, opened 2022-06-17
feat: add abbreviations to `hacker.abbreviation()` in the `en` locale

import-brain:
some logic gates, other acronyms and abbreviations


xDivisionByZerox:
Aren't the "logic gates" just plain words? I'm questioning the decision to put them in the abbreviations file.

src/locales/en/hacker/abbreviation.ts
```diff
@@ -28,4 +29,18 @@ export default [
...
   'JBOD',
+  'AND',
+  'OR',
+  'XOR',
+  'NOT',
+  'NAND',
+  'NOR',
+  'XNOR',
+  'DNS',
+  'ASCII',
```
xDivisionByZerox, line 40: What about other encodings? (UTF8, HEX, USC2, ...)

ST-DDT: Utf8 and hex yes.
IMO USC2 is probably not common enough.

How about TLS and API?

import-brain: Hmm, okay. I'll remove the logic gates and add TLS, API, and the two encodings.

ST-DDT (dismissed):

Shinigami92 (commented):
SSD!

xDivisionByZerox:
> SSD!

What? We already have that.

ST-DDT:
> > SSD!
> 
> What? We already have that.

Can you give me a line number?
I only see SDD (Same Day Delivery?).

xDivisionByZerox:
> Can you give me a line number? I only see SDD (Same Day Delivery?).

Damn, your write. I literally read that as SSD...

Well in that case we can add `'HDD'` as well.

ST-DDT:
Looks like your brain is melting just like mine 😉😅🥵

ST-DDT:
Please either sort them alphabetically or sort them by their category as before.

src/locales/en/hacker/abbreviation.ts
```diff
@@ -1,31 +1,43 @@
...
+  'RSS',
   'SAS',
-  'IB',
   'SCSI',
-  'JSON',
+  'SDD',
+  'SMS',
+  'SMTP',
+  'SQL',
+  'SSL',
```
Shinigami92, line 33:
```suggestion
  'SSD',
  'SSL',
```

ST-DDT (approved):

Shinigami92 (dismissed):

Shinigami92 (approved):

import-brain:
I don't know if I'm doing this correctly or not, but the results that the seeded tests return seem to change on every run of the tests now that there are snapshots. I've manually changed them multiple times, but the received value from the tested function seems to change everytime. Also, updating the snapshot by pressing "u" after the tests run doesn't seem to do anything, what's up with that?

Shinigami92:
You left the seededRuns object and didn't remove it

Shinigami92 (approved):

ST-DDT (approved):

import-brain:
> You left the seededRuns object and didn't remove it

Ahhh, I see. Thanks Shini :)

PR #1088, opened 2022-06-18
test: fix cypress doc tests

Shinigami92:


ST-DDT:
The e2e tests failed both in the terminal and in electron until I added a wait.
However, since I started to write this comment the failure is no longer reproducible both with and without the wait😕 .

I still get this error/message though:

````txt
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
````

**Comment by ST-DDT** [comment]
The e2e tests failed both in the terminal and in electron until I added a wait.
However, since I started to write this comment the failure is no longer reproducible both with and without the wait😕 .

I still get this error/message though:

````txt
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
````

**Comment by JessicaSachs** [comment]
Ignore the error message... known issue.

File: cypress.config.ts [hunk 1]
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'cypress';
+
+export default defineConfig({
+  video: false,
+  e2e: {
+    chromeWebSecurity: false,
+    baseUrl: 'http://localhost:5000',
+    supportFile: false,
+    fixturesFolder: false,
```
xDivisionByZerox, line 9: What is with the missing configuration properties (integrationFolder, testFiles, pluginsFile)? Have they been deprecated?
I'm sorry if these are stupid questions but the PR doesn't have any description.

Shinigami92: https://docs.cypress.io/guides/references/migration-guide#Configuration-File-Changes
Happy reading :slightly_smiling_face:

xDivisionByZerox (approved):

ST-DDT (approved):

PR #1090, opened 2022-06-18
build: just read filenames of locales

Shinigami92:
This is an optimization separated from #1052

Instead of loading all locales while bundling, we can just read their filenames and so it don't transpile all the typescript content in memory


ST-DDT (approved):

Shinigami92:
The way going to #1287 solves this much better, so this PR would get stale anyways

PR #395, opened 2022-02-01
feat: add ability to define number of generated extensions in system.filePath

pkuczynski:
Includes #300 which should be merged first

PR #1101, opened 2022-06-22
feat(system.fileName): file extension count

xDivisionByZerox:
Implementation is discussed in #395 ([direct comment ref](https://github.com/faker-js/faker/pull/395#discussion_r843866722)).

This is implemented in a non-breaking manner.


prisis:
> Since @pkuczynski somewhat provided the implementation in his PR, I want to make them coauthor in this PR. How would this be possible? Is assigning them to the PR enough?

It is possible if you merge the commit to add a coauthor

src/modules/system/index.ts
```diff
@@ -31,13 +31,38 @@ export class System {
...
+      options = {
+        extensionCount: 1,
+      };
+    }
+    const baseName = this.faker.random
+      .words()
+      .toLowerCase()
+      .replace(/\W/g, '_');
+
+    const extCount = options.extensionCount;
```
prisis, line 56: can you move this call after the if?

xDivisionByZerox: I first had the mapping in line 62 implemented as a `while` loop and outsourced the count in this constant to not mutate the potential input object. But since this is no longer the case, I think the const is obsolete. 

I will remove it instead and pass the array length directly from the options.

xDivisionByZerox: Nvm, I saw that I never default `extensionCount` if an empty object is provided. Guess I will implement a test case for that.

src/modules/system/index.ts
```diff
@@ -31,13 +31,40 @@ export class System {
...
+    /**
+     * Define how many extensions the file name should have. A negative number will be treated as `0`. Defaults to `1`.
+     */
+    extensionCount?: number;
+  }): string {
+    if (options == null) {
+      options = {
+        extensionCount: 1,
+      };
+    }
```
ST-DDT, line 50:
```suggestion
  fileName(options: {
    /**
     * Define how many extensions the file name should have. A negative number will be treated as `0`. Defaults to `1`.
     */
    extensionCount?: number;
  } = {}): string {
```

ST-DDT (changes requested):
Please add a seeded test, that uses the new option.

ST-DDT:
> So by adding new test cases, existing ones (not the new one) are randomly failing? This doesn't seem right to me. I also can reproduce the values in the test by setting the seed to the one in the test.

The random seeded tests are less random than one would like. I will try to have a look at the randomness issue soon.
The test previously passing was probably a false positive.

xDivisionByZerox:
> Please add a seeded test, that uses the new option.

Honestly, no. I don't see the point in the seeded tests. No other function has an extra seeded test case for their argument, why would this have? I already added some test cases that check for the argument options in the non-seeded test runs. They are much more valuable IMO.

test/system.spec.ts
```diff
@@ -180,6 +181,41 @@ describe('system', () => {
...
         });
+
+        it('should return filenames with 1 ext per default', () => {
+          const fileName = faker.system.fileName();
+          const parts = fileName.split('.');
+
+          expect(parts).length(2);
+        });
+
+        it('should return filenames without a extension when extensionCount is 0', () => {
```
Shinigami92, line 192:
```suggestion
        it('should return filenames without an extension when extensionCount is 0', () => {
```

test/system.spec.ts
```diff
@@ -180,6 +181,41 @@ describe('system', () => {
...
+          const fileName = faker.system.fileName({
+            extensionCount: 0,
+          });
+
+          expect(fileName).not.toContain('.');
+        });
+
+        it('should return filenames without a extension when extensionCount is negative', () => {
```
Shinigami92, line 200:
```suggestion
        it('should return filenames without an extension when extensionCount is negative', () => {
```

test/system.spec.ts
```diff
@@ -180,6 +181,41 @@ describe('system', () => {
...
+          const fileName = faker.system.fileName({
+            extensionCount: -1,
+          });
+
+          expect(fileName).not.toContain('.');
+        });
+
+        it.each(times(10))(
+          'should return filenames with a %s extensions',
```
Shinigami92, line 209:
```suggestion
          'should return filenames with %s extensions',
```

test/system.spec.ts
```diff
@@ -180,6 +181,41 @@ describe('system', () => {
...
+          (extCount) => {
+            const fileName = faker.system.fileName({
+              extensionCount: extCount,
```
Shinigami92, line 212:
```suggestion
          (extensionCount) => {
            const fileName = faker.system.fileName({
              extensionCount,
```

xDivisionByZerox: done in 0bd31da

test/system.spec.ts
xDivisionByZerox: fixed in d2ecae9

test/system.spec.ts
xDivisionByZerox: fixed in d2ecae9

test/system.spec.ts
xDivisionByZerox: fixed in d2ecae9

src/modules/system/index.ts
xDivisionByZerox: Done in aa351b26e7765d9e1a78cd65557fbf10e8d70cff

src/modules/system/index.ts
```diff
@@ -31,13 +31,40 @@ export class System {
   /**
    * Returns a random file name with extension.
    *
+   * @param options An options object.
+   * @param options.extensionCount Define how many extensions the file name should have. A negative number will be treated as `0`. Defaults to `1`.
    * @example
```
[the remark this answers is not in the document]
xDivisionByZerox: Is there a possibility to automate this spacing preference via a linter/formatter?

xDivisionByZerox:
The failing test is expecting a common file extension to be included in an array provided in the test. According to our implementation of `system.commonFileExt()` all of the following would be valid common file types
```ts
[
  'gif', 'htm', 'html', 'jpe',
  'jpeg', 'jpg', 'm1v', 'm2a',
  'm2v', 'm3a', 'mp2', 'mp2a',
  'mp3', 'mp4', 'mp4v', 'mpe',
  'mpeg', 'mpg', 'mpg4', 'mpga',
  'pdf', 'png', 'shtml', 'wav',
]
```

But our expected values are only:
[
  'gif', 'htm', 'html', 'jpeg',
  'm2a', 'm2v', 'mp2', 'mp3',
  'mp4', 'mp4v', 'mpeg', 'mpg',
  'pdf', 'png', 'shtml', 'wav',
]
```

ST-DDT:
Then update the expected values!?
I'm not sure whether I would consider all of them "common".

xDivisionByZerox:
> [...] I would consider all of them "common".

That's where I wasn't sure about and wanted to have some "backup" from you.

ST-DDT:
I have rarily/never seen m2a/v mp2 or shtml.
So I  would remove them. Lets discuss this tomorrow with everyone.
I miss json and txt.

xDivisionByZerox:
For now, I updated the test expectation. In the future we should rework the function to pick extensions from a custom-defined array or something like that.

ST-DDT (approved):

import-brain (approved):

xDivisionByZerox:
This is non-blocking. Can we merge?

Shinigami92 (approved):

PR #1102, opened 2022-06-22
feat(locales.en): extend vehicle

duvall:


prisis (approved):

prisis (changes requested):
Can you please change the commit message?

duvall:
> Can you please change the commit message?

You have a preference ?

xDivisionByZerox:
Please don't use exclamation marks ("!") in your commit messages if they are not introducing a breaking change (even if they had to be in another location). Please also note that smilies are generally considered unprofessionally for commit messages (you can still write them in GitHub comments, tho). 
Last but not least, your commit message should caption what you did in this commit. In your case "chore(locales.en): extend vehicle" would be a much better fit.

xDivisionByZerox:
The new commit message should be fine as well, IMO.

xDivisionByZerox:
Please update the test expectation snapshots. To do that run the test script in your project with the update flag `vitest -u` OR delete the according snapshot file (in your case `test\__snapshots__\vehicle.spec.ts.snap`) and run the basic test script `vitest` or `npm run test`. The snapshot file should be updated automatically.

ST-DDT (approved):

PR #1122, opened 2022-07-01
feat(number): move methods to new module

### LINKED ISSUE #1342: Move Number methods to own module

Currently the number methods are spread in multiple modules and it isn't easy to guess them correctly.

### Suggested solution

- All random number methods should be in their own module
- The old methods are preserved for a full major version (deprecated)

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
This move all number generators into their own (number) module.

Idea for this is from #805.


xDivisionByZerox:
Currently, I have removed all functions that were moved to the number module. I guess that wasn't the best move from me since we should deprecate them in v8 and remove them in v9. Correct?

src/modules/number/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    const { precision = 2 } = options;
+
+    const int = this.int(options);
+    if(precision <= 0) {
+      return int;
+    } else if(int === options.max) {
+      return int;
+    }
```
xDivisionByZerox, line 99: This is currently not working as expected when non-int ranges are provided. Example `float({ min: 0.1, max: 0.5 })`.

src/modules/number/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Returns a [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) number.
+   *
+   * @param length Length of the generated number. Defaults to `1`.
+   *
+   * @example
+   * faker.datatype.hexadecimal() // 'b'
+   * faker.datatype.hexadecimal(10) // 'ae13f044fb'
+   */
+  hexadecimal(length = 1): string {
```
ST-DDT, line 122: This method shouldnt be here as it produces strings.

xDivisionByZerox: By that logic `bigInt` shouldn't be located here as well. It does produce a hex number. I don't see anything wrong here.

xDivisionByZerox: If I as a user want to have a hexadecimal number I would check in the numbers module for it, not in the string module. Also, I think it's okay to put implementations in multiple modules if that makes sense. So `hexadecimal` could be a function in number AND string. But that's a different topic.

xDivisionByZerox:
Suggestion from a discord user:

> I would love to have a number.int, number.bigInt and something with [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754)

src/modules/number/index.ts
```diff
@@ -0,0 +1,212 @@
...
+   *
+   * @throws When options define `max < min`.
+   *
+   * @example
+   * faker.number.int() // 55422
+   * faker.number.int(100) // 52
+   * faker.number.int({ min: 1000000 }) // 1031433
+   * faker.number.int({ max: 100 }) // 42
+   */
+  int(options: number | Range = {}): number {
```
xDivisionByZerox, line 46: What would be the expected behavior when inputs are non-integers? Should they be rounded up/down or round the result from `mersenne.rand()` entirely?

src/modules/number/index.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    const { min = 0 } = options;
+    const max = options.max ?? min + 99999;
+
+    if (max === min) {
+      return min;
+    }
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be greater than min ${min}.`);
```
xDivisionByZerox, line 59: Testcase for this error is missing.

Shinigami92 (dismissed):
**BLOCKED**: we cannot name the module `Number` as this is the JavaScript global Number constructor and we would run into the same issues as with `_Date` module (were we actually also want to find a better name for, like e.g. `DateModule` or `Temporal`)

If we name it `NumberModule` please first merge #932

src/modules/helpers/index.ts
```diff
@@ -354,7 +354,7 @@ export class Helpers {
     options: { probability?: number } = {}
   ): T | undefined {
     const { probability = 0.5 } = options;
-    if (this.faker.datatype.float({ min: 0, max: 1 }) < probability) {
+    if (this.faker.number.int({ min: 0, max: 100 }) / 100 < probability) {
```
ST-DDT, line 357: Why did you change this?

src/modules/number/index.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const randomInt = this.faker.mersenne.rand(max + 1, min);
+
+    return randomInt;
```
Shinigami92, line 52: Just nit-picking, but why do you name this `randomInt` but down you name it `float`?
Stick to one convention and make it consistent 🙂

src/modules/number/index.ts
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * @throws When options define `max < min`.
+   *
+   * @example
+   * faker.number.int() // 55422
+   * faker.number.int(100) // 52
+   * faker.number.int({ min: 1000000 }) // 1031433
+   * faker.number.int({ max: 100 }) // 42
+   */
+  int(options: number | { min?: number; max?: number } = {}): number {
```
Shinigami92, line 34: Should we use overloads with different parameter name?

Like we could do:

```ts
int(): number;
int(max: number): number;
int(options: { min?: number; max?: number }): number;
int(options: number | { min?: number; max?: number } = {}): number {
  // ...
```

That way the DX is slightly increased as this `max` will be shown in editors like VSCode if you turn on the feature for it

src/modules/datatype/index.ts
```diff
@@ -302,11 +302,11 @@ export class Datatype {
...
-   * faker.datatype.bigInt() // 55422n
-   * faker.datatype.bigInt(100n) // 52n
-   * faker.datatype.bigInt({ min: 1000000n }) // 431433n
-   * faker.datatype.bigInt({ max: 100n }) // 42n
-   * faker.datatype.bigInt({ min: 10n, max: 100n }) // 36n
+   * faker.number.bigInt() // 55422n
+   * faker.number.bigInt(100n) // 52n
+   * faker.number.bigInt({ min: 1000000n }) // 431433n
+   * faker.number.bigInt({ max: 100n }) // 42n
+   * faker.number.bigInt({ min: 10n, max: 100n }) // 36n
```
xDivisionByZerox, line 309: This documentation is wrong. Prob introduced while rebasing.

src/modules/helpers/index.ts
xDivisionByZerox: Don't know :/ The last rebase was very big, guess I introduced some errors during that. I'll revert to float.

src/modules/helpers/index.ts
```diff
@@ -446,7 +446,7 @@ export class Helpers {
 
     while (i-- > min) {
       index = Math.floor(
-        (i + 1) * this.faker.datatype.float({ min: 0, max: 0.99 })
```
xDivisionByZerox, line 449: This should use `float`.

src/modules/mersenne/index.ts
```diff
@@ -22,7 +22,7 @@ export class Mersenne {
   /**
    * Generates a random number between `[min, max)`.
    *
-   * @param max The maximum number. Defaults to `32768`.
```
xDivisionByZerox, line 25: This should probably not be in this PR. I mean it's somewhat related but it should be discussed separatly if we want to add this doc snippet (since all other range properties in the codebase are inclusive).

src/modules/number/index.ts
xDivisionByZerox: I remember we didn't want to do this because it requires a shit ton of duplicated JSDocs for the documentation. Please correct me if I'm wrong.

src/modules/number/index.ts
xDivisionByZerox: So `value` all the way. Gotcha!^^

src/modules/mersenne/index.ts
ST-DDT: IMO the Mersenne class should be removed from the public api mid term.

xDivisionByZerox: Well, it's still nice to know that the max value is exclusive, for intern usage.

src/modules/number/index.ts
Shinigami92: If we don't need to explicitly check it, I would just say: leave it as it behaves on default like yet

src/modules/number/index.ts
Shinigami92: If this is also needed in main branch, I would prefer open a separate PR and rebase this one after it got merged

xDivisionByZerox: Nah, we have this test case for `datatype.number`, but I forgot to migrate it.

src/modules/helpers/index.ts
```diff
@@ -446,7 +448,7 @@ export class Helpers {
...
+        (i + 1) * this.faker.number.float({ min: 0, max: 1, precision: 2 })
```
ST-DDT, line 451: AFAICT this would be better off with `int(0, i+1)` or whatever it does. I assume the 0.99 is there to prevent the result from overshooting.

src/modules/number/index.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (typeof options === 'number') {
+      options = { max: options };
+    }
+
+    const { min = 0 } = options;
+    const max = options.max ?? min + 99999;
```
ST-DDT, line 40: Maybe use:

````suggestion
const { min = 0, max = min + 99999 } = options;
````

File: src/modules/number/index.ts [hunk 13]
```diff
@@ -0,0 +1,184 @@
...
+      return min;
+    }
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be greater than min ${min}.`);
+    }
+
+    const int = this.faker.mersenne.rand(max + 1, min);
+
+    return int;
```
ST-DDT, line 52: Maybe return directly?

src/modules/number/index.ts
```diff
@@ -0,0 +1,184 @@
...
+      | number
+      | string
+      | {
+          min?: bigint | boolean | number | string;
+          max?: bigint | boolean | number | string;
+        }
+  ): bigint {
+    let min: bigint;
+    let max: bigint;
+
```
ST-DDT, line 155: Maybe use the same kind of remapping the options as you do with float?

ST-DDT (changes requested):

src/modules/number/index.ts
xDivisionByZerox: Done as far as possible. I can't do `const { min = BigInt(0), max = BigInt(min + 999999999999999) } = options` since if someone inputs `{min: true, max: true }` properties would be of type `boolean` after destruction.

src/modules/number/index.ts
xDivisionByZerox: So can I resolve this and we tackle this issue separately?

src/modules/number/index.ts
xDivisionByZerox: I'm used to this style due to debugability coding guidelines from my job. But I don't care, so done.

test/__snapshots__/helpers.spec.ts.snap
```diff
@@ -112,33 +113,34 @@ exports[`helpers > 1211 > arrayElement > with array 1`] = `"!"`;
...
 ]
 `;
 
 exports[`helpers > 1211 > arrayElements > with array 2`] = `
 [
-  "r",
-  "o",
+  "l",
+  " ",
+  undefined,
```
xDivisionByZerox, line 143: This doesn't look correct.
@ST-DDT guess we can replace the index in `arrayElements` with `int({ min: 0, max: i + 1 })`. 
But `int({ min: 0, max: i })` should work (?)

src/modules/helpers/index.ts
xDivisionByZerox: Look [here](https://github.com/faker-js/faker/pull/1122/files#r954159610).

test/__snapshots__/helpers.spec.ts.snap
ST-DDT: Yeah, I probably got confused with the old `(i+1) * x`

ST-DDT: Yeah, I probably got confused with the old `(i+1) * x`

test/number.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+import { faker, FakerError } from '../src';
+import { seededTests } from './support/seededRuns';
+
+describe('number', () => {
+  afterEach(() => {
+    faker.locale = 'en';
+  });
+
+  seededTests(faker, 'number', (t) => {
+    t.describeEach('int', 'hex')((t) => t.it('noArgs'));
```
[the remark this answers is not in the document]
xDivisionByZerox: "with value" means a primitive and "with options" means with an options object. Just asking so I'm understanding you correctly.

test/__snapshots__/helpers.spec.ts.snap
xDivisionByZerox: Fixes the problem by applying `int({ min: 0, max: i + 1 })`

src/modules/number/index.ts
```diff
@@ -0,0 +1,182 @@
...
+   * @example
+   * faker.number.float() // 51696.36
+   * faker.number.float(1) // 52023.2
+   * faker.number.float({ min: 1000000 }) // 212859.76
+   * faker.number.float({ max: 100 }) // 28.11
+   * faker.number.float({ precision: 1 }) // 84055.3
+   * faker.number.float({ min: 10, max: 100, precision: 3 }) // 57.315
+   */
+  float(
+    options: number | { min?: number; max?: number; precision?: number } = {}
```
Shinigami92, line 69: I currently think about the `precision` value now changed...
This could come into confusion when using it with native JS as there it has another meaning: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision#parameters

> An integer specifying the number of **significant** digits.

It's different to what e.g. `toFixed` is doing: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#parameters

> The number of digits to appear **after the decimal point**

So we might want to rename it, or do alternative ways 🤔 
I'm also not sure if we e.g. accept negative exponents, but this might work but lead just to more _roundings_

src/modules/number/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    return this.faker.mersenne.rand(max + 1, min);
+  }
+
+  /**
+   * Returns a single random floating-point number for the given precision or range and precision.
+   *
+   * @param options Precision or options object. Defaults to `{}`.
+   * @param options.min Lower bound for generated number. Defaults to `0`.
+   * @param options.max Upper bound for generated number. Defaults to `99999`.
+   * @param options.precision Number of digits after the decimal point. Negative numbers will be treated as `0`. Defaults to `2`.
```
ST-DDT, line 58: > Negative numbers will be treated as `0`.

Currently I don't see anything related to that in the code.

xDivisionByZerox: So, that's definitly a missing test case

test/number.spec.ts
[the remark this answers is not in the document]
xDivisionByZerox: Done in dac2931d4c075921bf605b7a8d7336b3352c2721.

test/number.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    t.describe('bigInt', (t) => {
+      t.it('noArgs').it('with value', 42);
```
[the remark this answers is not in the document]
xDivisionByZerox: Is one value and one option test enough here or should there be a test for each possible input type (`string`, `number`, `boolean`, `bigint`, and everything as options)

src/modules/number/index.ts
xDivisionByZerox: Should we discuss this in a meeting? I'm not sure if I should already implement this, because of @Shinigami92's [comment](https://github.com/faker-js/faker/pull/1122#discussion_r954573961).

> I'm also not sure if we e.g. accept negative exponents, but this might work but lead just to more roundings

xDivisionByZerox:
Is this ready review? #932 was merged, so this is no longer blocked.

ST-DDT:
Review yes, merge no. We will release "one last" v7.x first.

test/number.spec.ts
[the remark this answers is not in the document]
ST-DDT: No, one for each option as well as all at once is fine.

src/modules/mersenne/index.ts
```diff
@@ -22,7 +22,7 @@ export class MersenneModule {
...
+   * @param max The maximum number. This value is exclusiv. Defaults to `32768`.
```
ST-DDT, line 25:
```suggestion
   * @param max The maximum number. This value is exclusive. Defaults to `32768`.
```

src/modules/number/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    const int = this.int({
+      max: max * factor,
+      min: min * factor,
+    });
+
+    return int / factor;
+  }
+
+  /**
+   * Returns a [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) number.
```
ST-DDT, line 101:
```suggestion
   * Returns a lowercase [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) number.
```

ST-DDT (changes requested):

ST-DDT:
Please also include the migration guide docs in this PR.
There are some changes (e.g. precision: 0.01 -> 2) that needs to be documented.

src/modules/number/index.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    const { precision = 2, min = 0, max = min + 99999 } = options;
+
+    if (max === min) {
+      return min;
+    }
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be greater than min ${min}.`);
+    }
```
xDivisionByZerox, line 89: Do we need these checks? We call `this.int` later, that does these checks anyway.

src/modules/number/index.ts
xDivisionByZerox: Done in 774e1f9f5018f559537e0dee23ae63f07bdfa7a6.

xDivisionByZerox:
> Please also include the migration guide docs part in this PR. There are some changes (e.g. precision: 0.01 -> 2) that needs to be documented.

I'm not sure. I don't see the responsibility in this PR. There should already be a migration from v7 to v8 where I only need to include these changes.

ST-DDT:
> I'm not sure. I don't see the responsibility in this PR. There should already be a migration from v7 to v8 where I only need to include these changes.

Documentation is a very important part of our work.
And yes, the migration page itself will already be there before this PR is merged, however we can still "mock" the page for now and just add the migration section related to this module, so it can be reviewed and potential usage caveeats be detected early.
IMO the author of a PR has the best insights into the changes and thus can describe/explain them well.
If we delay the documentation, then we might not write them at all, which would be bad.
What do you think?

xDivisionByZerox:
> And yes, the migration page itself will already be there before this PR is merged,

In that case, I'm okay with that. But I don't see the responsibility of this PR to create the entire "migration boilerplate".

src/modules/number/index.ts
```diff
@@ -0,0 +1,190 @@
...
+   * @param options.max Upper bound for generated number. Defaults to `min + 99999`.
+   *
+   * @throws When options define `max < min`.
+   *
+   * @example
+   * faker.number.int() // 55422
+   * faker.number.int(100) // 52
+   * faker.number.int({ min: 1000000 }) // 1031433
+   * faker.number.int({ max: 100 }) // 42
+   *
```
import-brain, line 33: Maybe we should add an example here with `min` and `max`?

test/number.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+          otherProperty: string;
+        } = Object.freeze({
+          min: initialMin,
+          otherProperty: initialOtherProperty,
+        });
+
+        expect(() => faker.number.bigInt(input)).not.toThrow();
+      });
+
+      it('should throw when min bigger max', () => {
```
import-brain, line 236:
```suggestion
      it('should throw when min > max', () => {
```

src/modules/number/index.ts
Shinigami92: I would say fail fast in this case

src/modules/number/index.ts
Shinigami92: I extracted this to its own issue
- #1591

src/modules/number/index.ts
Shinigami92: In this case I would say fail fast and so: yes

src/modules/datatype/index.ts
```diff
@@ -332,6 +351,13 @@ export class DatatypeModule {
           max?: bigint | boolean | number | string;
         }
   ): bigint {
+    deprecated({
+      deprecated: 'faker.datatype.bigInt()',
+      proposed: 'faker.number.bigInt()',
+      since: '8.0',
+      until: '9.0',
+    });
```
ST-DDT, line 325: Is there a reason this method isn't redirecting to the new method?

src/modules/finance/index.ts
```diff
@@ -231,7 +231,7 @@ export class FinanceModule {
    * @since 3.1.0
    */
   bitcoinAddress(): string {
-    const addressLength = this.faker.datatype.number({ min: 25, max: 39 });
+    const addressLength = this.faker.number.int({ min: 25, max: 39 });
```
ST-DDT, line 234: Instead of calculating that here you can pass it directly to `alphanumeric`.

src/modules/finance/index.ts
```diff
@@ -342,7 +338,7 @@ export class FinanceModule {
     if (length < 1) {
       throw new FakerError('minimum length is 1');
     }
-    return Array.from({ length }, () => this.faker.datatype.number(9)).join('');
+    return Array.from({ length }, () => this.faker.number.int(9)).join('');
```
ST-DDT: This should use string.numeric instead.

src/modules/datatype/index.ts
```diff
@@ -37,10 +37,19 @@ export class DatatypeModule {
    * faker.datatype.number({ min: 10, max: 100, precision: 0.01 }) // 36.94
    *
    * @since 5.5.0
+   *
+   * @deprecated Use `faker.number.int` instead.
```
xDivisionByZerox:
```suggestion
   * @deprecated Use `faker.number.int()` instead.
```

src/modules/datatype/index.ts
```diff
@@ -86,10 +95,18 @@ export class DatatypeModule {
    * faker.datatype.float({ min: 10, max: 100, precision: 0.001 }) // 57.315
    *
    * @since 5.5.0
+   *
+   * @deprecated Use `faker.number.float` instead.
```
xDivisionByZerox:
```suggestion
   * @deprecated Use `faker.number.float()` instead.
```

src/modules/commerce/index.ts
```diff
@@ -67,14 +67,10 @@ export class CommerceModule {
       return `${symbol}${0.0}`;
     }
 
-    const randValue = this.faker.datatype.number({ max: max, min: min });
+    const randValue = this.faker.number.int({ max, min });
```
ST-DDT: Please order `min, max` consistently across all modules/invocations.

src/modules/number/index.ts
ST-DDT: Still nothing related to this in the sources.

src/modules/vehicle/index.ts
```diff
@@ -100,7 +100,7 @@ export class VehicleModule {
       length: 1,
       casing: 'upper',
       exclude,
-    })}${this.faker.datatype.number({ min: 10000, max: 99999 })}` // return five digit #
+    })}${this.faker.number.int({ min: 10000, max: 99999 })}` // return five digit #
```
ST-DDT: This should string.numeric instead.

Shinigami92: This changes returned result, so not part of this PR 🙅

src/modules/finance/index.ts
Shinigami92: This changes returned result, so not part of this PR 🙅

test/number.spec.ts
```diff
@@ -0,0 +1,319 @@
...
+                min: 0,
+                max: 1.5,
+                // TODO @Shinigami92 2022-11-23: Which precision should be used to get only values with 0.0, 0.5, 1.0 and 1.5?
+                precision: 0.5,
+              })
+            )
+          )
+        ).sort();
+
+        expect(results).toEqual([0, 0.5, 1, 1.5]);
```
Shinigami92: I found this missing test-case and the previous implementation could handle it to assume values like non `0.1`, `0.01` or `0.001` and so on.
So how would this be supported by the now changed precision parameter?
My brain cannot do the math 😣

I see three options:
1. Revert to use precision with `0.x`
2. Explain me the math and add this to JSDoc
3. Let the user round the result by them self like https://stackoverflow.com/a/34591063/6897682

src/modules/number/index.ts
Shinigami92: I added test cases for that: https://github.com/faker-js/faker/pull/1122/commits/f2cf2ce4f743f920b8b5d9423c9fcec7818db964

Shinigami92:
@ST-DDT Okay I now added a test case for the `0.5` steps but how would I use steps `0.3` or `0.4`?
Like getting `0.0`, `0.4`, `0.8`, `1.2`, `1.6`, ...

I would like to provide JSDoc examples and/or a source for people like me that cannot do the math out of their brain

ST-DDT:
> @ST-DDT Okay I now added a test case for the `0.5` steps but how would I use steps `0.3` or `0.4`? Like getting `0.0`, `0.4`, `0.8`, `1.2`, `1.6`, ...
> 
> I would like to provide JSDoc examples and/or a source for people like me that cannot do the math out of their brain

IIRC Precision = log10(0.4) or ln(0.4) / ln(10)

Shinigami92:
> IIRC Precision = log10(0.4) or ln(0.4) / ln(10)

Both wrong and both return the same result :confused: 

```ts
faker.number.float({ min: 0, max: 1.9, precision: Math.log(0.4) / Math.LN10 })
faker.number.float({ min: 0, max: 1.9, precision: Math.log10(0.4) })
// => [0, 2.5] :eyes: 
```

test/number.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+
+      it('should return a random bigint given a maximum value as BigInt', () => {
+        expect(faker.number.bigInt(10n)).toBeGreaterThanOrEqual(0n);
+        expect(faker.number.bigInt(10n)).toBeLessThanOrEqual(10n);
+      });
+
+      it('should return a random bigint given a maximum value as Object', () => {
+        expect(faker.number.bigInt({ max: 10n })).toBeGreaterThanOrEqual(0n);
+        expect(faker.number.bigInt({ max: 10n })).toBeLessThanOrEqual(10n);
+      });
```
ejcheng: Should we declare a const variable here and use that to validate bounds of the generated bigint instead of calling `number.bigInt` twice? I think that would keep it consistent with the other tests

test/number.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+
+      it('should return a random bigint given a negative bigint minimum and maximum value of 0', () => {
+        expect(
+          faker.number.bigInt({ min: -100n, max: 0n })
+        ).toBeGreaterThanOrEqual(-100n);
+        expect(
+          faker.number.bigInt({ min: -100n, max: 0n })
+        ).toBeLessThanOrEqual(0n);
+      });
```
ejcheng: Same here

test/number.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+
+      it('should return a random bigint between a range', () => {
+        const randomBigInt = faker.number.bigInt({ min: 22, max: 33 });
+        expect(randomBigInt).toBeGreaterThanOrEqual(22);
+        expect(randomBigInt).toBeLessThanOrEqual(33);
+      });
+
+      it('should succeed with success-rate', () => {
```
ejcheng:
```suggestion
      it('should succeed with success-rate', () => {
```
What does 'succeed with success-rate' mean?

src/modules/number/index.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    const mersenne: Mersenne =
+      // @ts-expect-error: access private member field
+      this.faker._mersenne;
+
+    return mersenne.next({ min, max: max + 1 });
+  }
+
+  /**
+   * Returns a single random floating-point number for the given precision or range and precision.
```
```suggestion
   * Returns a single random floating-point number for a given precision or range and precision.
```

test/number.spec.ts
Shinigami92: I don't know :thinking: 
Somehow it feels like we can just remove this test, can we?

ST-DDT:
> > IIRC Precision = log10(0.4) or ln(0.4) / ln(10)
> 
> Both wrong and both return the same result 😕
> 
> ```ts
> faker.number.float({ min: 0, max: 1.9, precision: Math.log(0.4) / Math.LN10 })
> faker.number.float({ min: 0, max: 1.9, precision: Math.log10(0.4) })
> // => [0, 2.5] :eyes: 
> ```

Correction
Log10(2.5)

Shinigami92:
> Correction
> Log10(2.5)

... well, we have a problem there:

image

So beside that I absolutely do not understand this math magic you are doing with these `log10`s,
it results in values greater then the requested max :eyes:

ST-DDT:
> > Correction
> > Log10(2.5)
> 
> ... well, we have a problem there:
> 
> image
> 
> So beside that I absolutely do not understand this math magic you are doing with these `log10`s, it results in values greater then the requested max 👀

Math: `1 / x = 0.4` => `x = 2.5` and `10 ** log10(x) = x`

Range: The issue here is probably with number.int
It uses +1 to make max inclusive.

1.9 × 2.5 = 4.75
4.75 + 1 = 5.75 (5 as int)
5 / 2.5 = 2

So instead of adding one we might want to add a very small number and then round up.
Alternatively substract precision from max before passing it  to int. Not sure whether that creates other issues.

ST-DDT:
We might want to be more strict in number.int anyway and ceil min and floor max before checking and using it. Because then we would pass 5 to mersenne, but since it is exclusive we will only get 4 as max returned value => 4 / 2.5 = 1.6.
Not tested yet.

ST-DDT:
Also since that math confuses even our own developers how about having both options?
Precision as used previously and decimal digits as used here?

Shinigami92:
I ensured together with @ST-DDT that the tests / bugs are already there before this PR and so I extracted it to #1595 so we do not pollute this PR with fixing this bug.

Shinigami92:
Due to the confusion with the log10 stuff, @ST-DDT and I decided for now to revert the parameter precision change from this PR.
But we will create a new issue to re-think about the change in another PR. (rename `precision` to `fractionDigits`)

Shinigami92 (dismissed):

src/modules/commerce/index.ts
```diff
@@ -67,14 +67,10 @@ export class CommerceModule {
       return `${symbol}${0.0}`;
     }
 
-    const randValue = this.faker.datatype.number({ max: max, min: min });
+    const randValue = this.faker.number.int({ min, max });
```
ST-DDT, line 71: Maybe simplify this to:

```suggestion
    const randValue = this.faker.number.float({ min, max, precision: 10 ** -dec });
```

?

src/modules/datatype/index.ts
```diff
@@ -86,10 +95,18 @@ export class DatatypeModule {
...
    */
   float(
     options?: number | { min?: number; max?: number; precision?: number }
   ): number {
+    deprecated({
+      deprecated: 'faker.datatype.float()',
+      proposed: 'faker.number.float()',
+      since: '8.0',
+      until: '9.0',
+    });
```
ST-DDT, line 88: Why don't you forward this method to faker.number.float()?

src/modules/datatype/index.ts
```diff
@@ -37,10 +37,19 @@ export class DatatypeModule {
...
    */
   number(
     options: number | { min?: number; max?: number; precision?: number } = 99999
   ): number {
+    deprecated({
+      deprecated: 'faker.datatype.number()',
+      proposed: 'faker.number.int()',
+      since: '8.0',
+      until: '9.0',
+    });
```
ST-DDT, line 49: Why don't you forward to number.float() here?

src/modules/number/index.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /**
+   * Returns a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#bigint_type) number.
+   *
+   * @param options Maximum value or options object. Defaults to `{}`.
+   * @param options.min Lower bound for generated bigint. Defaults to `0n`.
+   * @param options.max Upper bound for generated bigint. Defaults to `min + 999999999999999n`.
+   *
+   * @throws When options define `max < min`.
```
ST-DDT, line 140: We don't have that for any of the other methods.

src/modules/datatype/index.ts
Shinigami92: Good catch, it was not forwarded because previously the implementation of precision was differently.
Now as we reverted that, we can forward it

src/modules/number/index.ts
Shinigami92: This was just copied over from the old JSDocs... so at least there we had it
Instead of removing it, we might more think about to add it to more places :thinking:

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #1127, opened 2022-07-04
refactor(name.findName): rename to fullName

xDivisionByZerox:
This PR renames the existing `faker.name.findName()` to `faker.name.fullName()`, while keeping the existing function but marking them as deprecated. 
I also changed the signature of `fullName()` to an object for maintainability purposes.

I had a discussion with @Shinigami92 last week about deprecations. We came to the conclusion that they should be categorized as `refactor` instead of `chore`, what I will start to implement with this PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -50,7 +50,7 @@ module.exports = defineConfig({
         allowBoolean: true,
       },
     ],
-    '@typescript-eslint/unbound-method': 'warn',
+    '@typescript-eslint/unbound-method': 'off',
   },
   overrides: [
     {


ST-DDT (approved):

import-brain (approved):

xDivisionByZerox:
This is a none blocking/breaking change. So, can we merge this?

pkuczynski (dismissed):

ST-DDT (dismissed):

xDivisionByZerox:
TBH, I don't see what's blocking the merge. Most likely just leaving this open in case anyone still wants a change. Often, many changes are merged in the release week. So I guess in one or two weeks.

pkuczynski:
Lets just merge it then! No point to keep it hanging. @ST-DDT @Shinigami92

Shinigami92 (approved):

ST-DDT (approved):

PR #1128, opened 2022-07-04
chore(linting): disable `unbound-method` warnings

xDivisionByZerox:
This PR fixes nearly all linter warnings for the [`unbound-method`](https://typescript-eslint.io/rules/unbound-method/) rule.


ST-DDT (approved):

import-brain (dismissed):

ST-DDT:
> I could call them in an anonym function as well but that would instantiate a new `Fake` or `Unique` every function call and I'm not happy with that.

I currently see two ways for this:
- convert them to standalone functions
- use a private method to initialize the module instance and return the anonymous function call.

Shinigami92:
> > I could call them in an anonym function as well but that would instantiate a new `Fake` or `Unique` every function call and I'm not happy with that.
> 
> I currently see two ways for this:
> 
> * convert them to standalone functions
> * use a private method to initialize the module instance and return the anonymous function to initialize the field with.

What is the general problem with how it currently works in main branch?
We already hacked it with the implementation in all the constructors using:

```ts
for (const name of Object.getOwnPropertyNames(<Module>.prototype)) {
    continue;
  }
}
```

So if this PR just aims to get rid of the warning, we might just want to turn the warning off and don't care
Please also note that folks in real world already uses code like
```ts

const uuid = faker.datatype.uuid

const id = uuid()
```
and until now, we support this
_even if I don't like our current hack, _ but the only way to really fix that, would be to rewrite EVERYTHING back to composable functions like

```ts
function datatype(faker = faker() /* or = new Faker() ?! */ ) {
    uuid() {
        return ... using something with faker. ....
    }
}
```

This would be a complete breaking change and currently I don't see that (at least not before we tackled more important parts like outsource locales / maybe using a monorepo)

xDivisionByZerox:
> So if this PR just aims to get rid of the warning, we might just want to turn the warning off and don't care

So I'm reverting all changes and changing the linter config instead?  Just give me a thumbs up or down and I will act accordingly. My only plan was to get rid of the warnings since they are triggering my OCD during development.

ST-DDT:
I'm fine either way, but the warnings bother me as well. Maybe put this on our agenda for tomorrow.

xDivisionByZerox:
Decision from todays meeting:
Remove the linter rule and revert the changes done in this PR. Explanation can be found in [Shinigami's comment](https://github.com/faker-js/faker/pull/1128#issuecomment-1174712194).

Shinigami92:
> I didn't close this PR :/ if you drop all changes you made, does it auto-close?

Never heard of this behavior, in the log history of GitHub it says you closed it.

Shinigami92:
> I know, but I promise I didn't click the close button. Since we decided to not "fix" these warnings I first wanted to revert all commits I did and then just thought. Fk it, I'll rebase and drop all commits (basically resetting). After force pushing "the changes" I received a notification that I closed the PR ¯\\_(ツ)_/¯

RIP

Shinigami92 (approved):

ST-DDT (approved):

PR #1133, opened 2022-07-07
feat(system.networkInterface): add networkInterface faker

iandrewt:
I built this to use in one of my own projects, based on the [SystemD predictable network interface device names schema](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/ch-consistent_network_device_naming) (although this is likely the antithesis of that 😛 ). Figured it might be worth sharing, but if it's too niche then no hard feelings 😄


src/modules/system/index.ts
```diff
@@ -171,4 +179,69 @@ export class System {
...
+   */
+  networkInterface(
+    options: {
+      interfaceType?: typeof commonInterfaceTypes[number];
+      interfaceSchema?: keyof typeof commonInterfaceSchemas;
+    } = {}
+  ): string {
+    const {
+      interfaceType = this.faker.helpers.arrayElement(commonInterfaceTypes),
+      interfaceSchema = this.faker.helpers.arrayElement(
```
ST-DDT, line 204: You could also use 
https://fakerjs.dev/api/helpers.html#objectKey

iandrewt: Oh that's nifty, update incoming

ST-DDT:
**Preview:** ✅ 

grafik

ST-DDT (dismissed):

Shinigami92 (dismissed):
I'm personally not really happy what prettier does for the formatting here but that is not the authors problem
Also the test regex are very heavy, but they work
The PR works, so I will approve

test/system.spec.ts
```diff
@@ -46,6 +47,29 @@ describe('system', () => {
...
+              interfaceSchema ? ` interfaceSchema ${interfaceSchema}` : ''
+            }`,
+            {
+              interfaceType,
+              interfaceSchema,
+            }
+          );
+        }
+      }
+    });
```
Shinigami92, line 72: @ST-DDT please check if I made that right (was my first time working with that new factory)
feedback: you cannot easily click anymore on the function that is tested to get to the definition of it :confused: 
luckily I can jump to it via click on one of the option params, but for functions without options it's not possible

ST-DDT: I would probably use `with ${options}` for simplicity sake, but this works for me as well.

ST-DDT (approved):

test/system.spec.ts
Shinigami92: Oh I thought using `${options}` would result in `'[object Object]'`

ST-DDT: Well, you could still jsonify it.

Shinigami92 (approved):

ST-DDT (approved):

PR #1137, opened 2022-07-08
test(locale): check for duplicated entries

xDivisionByZerox:
This PR will introduce a test that checks every `LocaleDefinition` for duplicated entries.

The most important file that needs review is `test/locales.spec.ts`. Everything else are just changes that are required by this test case.


xDivisionByZerox:
I expect the test pipeline to fail for demonstration purposes.

@ST-DDT would you have a look at the implementation and the error log in general? I trust in you and your abilities on such things.

ST-DDT:
The implementation looks good, maybe you could use this expectation:

````ts
expect(
  uniqueDuplication,
  `Found duplicated values [${uniqueDuplication
    .sort()
    .join(', ')}]`
).toBe([]);
````

The toBe check focuses more on the content of the array than the length.

grafik

What do you think?

(PS: We have to check if there are/should be exceptions to this rule)

File: test/locales.spec.ts [hunk 1]
```diff
@@ -12,4 +14,45 @@ describe('locale', () => {
...
+          continue;
+        }
+
+        describe(moduleName, () => {
+          for (const [definitionName, entries] of Object.entries(
+            definitionMap
+          )) {
+            describe(definitionName, () => {
+              if (Array.isArray(entries)) {
+                it('should have no duplicate entries', () => {
```
ST-DDT, line 34:
```suggestion
                it('should not have duplicate entries', () => {
```
Not sure about this one.

xDivisionByZerox:
So I need to change about 1,4k locale files since they contain duplicates. Is it reasonable to fix all of this in one (THIS) PR? My current workflow is:
1. sort all entries in a locale files
2. commit with message `'chore(locales): *localeName* *fileName* sort alphabetically'`
3. remove duplicated entries
4. commit with message `'chore(locales): *localeName* *fileName* remove duplicates'`

My question: Should I provide an additional PR where I sort ALL locale files alphabetically first. This could be much clearer for the git history.

I use the [`Tyriar.sort-lines`](https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines) VS Code extension to sort locale file entries.

Shinigami92:
@xDivisionByZerox yes, for me that sounds more safe
1. PR for sort
2. PR for dedup
That way we have a better commit history in main branch

test/locales.spec.ts
```diff
@@ -12,4 +14,79 @@ describe('locale', () => {
...
+                    it('cant be tested', () => {
+                      expect(true).toBe(true);
+                    });
+                  }
+                }
+              } else {
+                it('needs to be tested', () => {
+                  expect(false).toBe(true);
+                });
+              }
```
xDivisionByZerox, line 85: Please help me optimize this. My brain can stand it and came up with this monster xD

xDivisionByZerox, line 85: Please help me optimize this. My brain can stand it and come up with this monster xD

test/locales.spec.ts
```diff
@@ -12,4 +14,79 @@ describe('locale', () => {
...
+                it('should not have duplicate entries', () => {
+                  const uniques = new Set(entries);
+                  const duplications = entries.filter((entry) => {
+                    if (uniques.has(entry)) {
+                      uniques.delete(entry);
+                      return false;
+                    } else {
+                      return true;
+                    }
+                  });
```
ST-DDT, line 43:
```suggestion
                  const duplications = entries.filter((entry) => !uniques.delete(entry));
```

ST-DDT (commented):
The check itself looks good, but duplicated. You might want to want to move the "describe if array ..." block into a helper method allowing recursive calls to it.
I could provide a patch/diff, if my explanation isn't easy to understand.

ST-DDT:
It looks like this PR doesn't really reduce the line count: `+10,463` `-4,859`
(not that it is important)

Shinigami92:
> It looks like this PR doesn't really reduce the line count: `+10,463` `-4,859` (not that it is important)

Yeah, you are right
Somehow some files are generated with fully new content 🤔 
Example:
image

xDivisionByZerox:
> Somehow some files are generated with fully new content 🤔

Oh damn. Good catch. I will revert those files. I guess they slipped through as I played with the `generate-locales` script.

xDivisionByZerox:
Those "additions" should be reverted now. They happened due to those files being implemented in a dynamic/logical way, not with a static data set.

ST-DDT:
I found a way to simplify the test a bit, feel free to merge: [diff](https://github.com/faker-js/faker/compare/test/locales/duplicated-entries...test/locales/duplicated-entries2?diff=split)

ST-DDT (dismissed):

ST-DDT:
@xDivisionByZerox Did you see my comment here: https://github.com/faker-js/faker/pull/1137#issuecomment-1258454562

src/locales/en/name/title.ts
```diff
@@ -1,6 +1,4 @@
-import type { NameTitleDefinitions } from '../../../definitions';
-
-const title: NameTitleDefinitions = {
```
Shinigami92, line 3: Was this change intended?

src/locales/en_NG/address/city_prefix.ts
```diff
@@ -218,8 +218,6 @@ export default [
   'Eggan',
   'Egudu Bale',
   'Egun',
-  'name',
-  ' ',
```
Shinigami92, line 222: Do we want also to remove line 120?

ST-DDT: Yes.

src/locales/en/name/title.ts
ST-DDT: It is one of the few files that still had this.

xDivisionByZerox: It was intended in that I would not change all those files by hand and had to make a generalization. And that one was:
- get exported value
- filter out duplicates
- rewrite the file with the new value (if changed) with `export default *value*;`

This obviously leads to everything else being overwritten in that file, but I thought it would be okay.

src/locales/en_NG/address/city_prefix.ts
xDivisionByZerox: Sorry, I didn't saw this comment. It should be done now.

Shinigami92 (approved):

ST-DDT (approved):

import-brain (approved):

PR #1147, opened 2022-07-12
docs(address.streetPrefix): refactor JSDOC example

xDivisionByZerox:
Updated a JSDOC example that was not using the default `faker` module. Could also have been a typo, IDK.

### DIFF
--- a/src/modules/address/index.ts
+++ b/src/modules/address/index.ts
@@ -231,7 +231,7 @@ export class Address {
    * Returns a random localized street prefix.
    *
    * @example
-   * fakerGH.address.streetPrefix() // 'Boame'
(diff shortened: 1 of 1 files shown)


pkuczynski (approved):

ST-DDT:
AFAICT this is not an accidential typo, but explicitly chosen.
The default en locale does not have street_prefixes. The (en_)GH locale does.

This is one of the reasons I would like to deprecate/remove these methods.
Not sure what to do until then.
The test examples and deprecation code has a special handling for these.

xDivisionByZerox:
Wow, that's crazy. In that case, I'll close this since everything is "working as expected".

PR #1148, opened 2022-07-12
chore(locale.de): update word definition

xDivisionByZerox:
This PR updates the word definitions for the german locale.

 Updated:
- adjective

Added:
- noun
- adverb
- verb
- preposition


pkuczynski (approved):
I can't judge `de` but I trust that you know what you are doing :)

import-brain (approved):

ST-DDT (approved):
I really like it.
Have you checked whether the adjective list allows for rebundling?
The page sells a product (writing/author class) so they may have interest in keeping it to their own.

src/locales/de/word/adjective.ts
```diff
@@ -1,193 +1,1136 @@
+// source: https://buchkodex.de/blog/adjektive/
```
Shinigami92, line 1: @ST-DDT informed me about this PR and I'm of the opinion that we should not directly include this link in the source code but maybe just in the PR description (which is linked by the commit)
The adjectives here are not all from that list and in future we might add more to it and therefore just merge more and more together, so the source is not directly true anymore
So please remove it from this source file

Shinigami92: I'm ok with your suggestion

pkuczynski:
I would consider Wikipedia more solid source of information...

pkuczynski (approved):

ST-DDT (approved):

PR #1155, opened 2022-07-15
feat(string): move methods to new module

### LINKED ISSUE #1341: Move String methods into own module

Currently the string methods are spread in multiple modules and it isn't easy to guess them correctly.

### Suggested solution

- All random string methods should be in their own module
- The old methods are preserved for a full major version (deprecated)

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
This PR introduces a new module: `String`. This was heavily discussed in #805.

I did not create any new functions. I just moved a bunch of them into the new module and deprecated the old ones.

## Which functions made it to the `String` module?

| was | now |
| - | - |
| datatype.string() | string.random() |
| datatype.uuid() | string.uuid() |
| datatype.hexadecimal() | string.hexadecimal() |
| random.alpha() | string.alpha() |
| random.alphaNumeric() | string.alphaNumeric() |
| random.numeric() | string.numeric() |


pkuczynski:
I would avoid word `random`, as everything in faker is random, so it creates impression that this is more random than others... Instead I propose to do `string.any()`

src/modules/string/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  /**
+   * Returns a string containing UTF-16 chars between 33 and 125 (`!` to `}`).
+   *
+   * @param length Length of the generated string. Max length is `2^20`. Defaults to `10`.
+   *
+   * @example
+   * faker.string.random() // 'Zo!.:*e>wR'
+   * faker.string.random(5) // '6Bye8'
+   */
+  random(length = 10): string {
```
pkuczynski, line 105: I would avoid word `random`, as everything in faker is random, so it creates impression that this is more random than others... Instead I propose to do `string.any()`

src/modules/string/index.ts
```diff
@@ -0,0 +1,390 @@
...
+   * @param options.count The number of characters and digits to generate. Defaults to `1`.
+   * @param options.casing The casing of the characters. Defaults to `'mixed'`.
+   * @param options.bannedChars An array of characters and digits which should be banned in the generated string. Defaults to `[]`.
+   *
+   * @example
+   * faker.string.alphaNumeric() // '2'
+   * faker.string.alphaNumeric(5) // '3e5v7'
+   * faker.string.alphaNumeric(5, { bannedChars: ["a"] }) // 'xszlm'
+   */
+  alphaNumeric(
```
pkuczynski, line 207: It should be written as `alphanumeric`. See https://en.wikipedia.org/wiki/Alphanumericals

src/modules/string/index.ts
```diff
@@ -0,0 +1,390 @@
...
+    options:
+      | number
+      | {
+          count?: number;
+          casing?: Casing;
+          bannedChars?: readonly LiteralUnion<AlphaNumericChar>[] | string;
```
pkuczynski, line 213: This is not exactly related to this PR, but maybe good moment to propose it. Instead of `bannedChars` and `bannerDigits`, I propose to use common option for such cases: `except`. If I remember well, we also have similar way of excluding some data in other methods too? Alternatively it could be also named  `exclude`, but I prefer `except`. What others think?

pkuczynski (commented):
Code wise look good, I am not approving only to have a chance to discuss the comments I left...

src/modules/string/index.ts
xDivisionByZerox: I, myself, have no problem with that. But, I think there were members that disliked the `any` function name in the number module since it has collisions with the typescript `any` keyword. But I really like `any()` as a generic solution, to the problem of the module (generating strings), TBH.

pkuczynski: Well, I don't see a conflict here. I just dbl checked, and typescript also has no problem when variable or function is named `any`. And in practice, this function generates any kind for string, right?

pkuczynski: Those faker members, who didnt like it, please raise your voice or stay silent forever ;)

Shinigami92: Ok I raise my hand to not be silence, but I'm also not sure what could be better in alternative. It's a little bit same as the discussion about number. It just doesn't feel right.

ST-DDT: This only generates ascii characters and no utf8 characters.
Please give me some time to think about.
I have to connect to a free wifi in order to respond to this, and I would like to spend my holidays with non-faker thoughts as well. 😉

pkuczynski: Hahaha, I didn't meant you have to speak immediately :) No rush, enjoy your stay and we can talk about it when you are both back :)

import-brain: Hmm, this is a really difficult decision for me. I think `random` is more clear, but it implies this function generates "more random" data than other functions, as @pkuczynski said. However, I don't think `any` is a good choice either, it seems a bit ambiguous. Does anyone have any third ideas for this function name?

src/modules/string/index.ts
import-brain: I definitely prefer `except` for this param.

src/modules/string/index.ts
pkuczynski: @import-brain and what's wrong with `any`?

import-brain: > @import-brain and what's wrong with `any`? In which way it is ambiguous?

It just doesn't sit right with me. If I looked at a function named `string.any()`, I would ask "any what?", and I would have to hover over the function and look at the jsdoc before realizing what it does. In that regard, `random` is a lot better, you can tell instantly that the function generates a random string. However, `random()` has the problems you mentioned.

pkuczynski: For my `string.any()` gives me a string with no constraints. Same as `any` in typescript... `random` is more explicit of course, but does not play so nicely with the other functions: `string.(numeric|alphanumeric|alpha)` - or `any` of those three...

import-brain: > For my `string.any()` gives me a string with no constraints. Same as `any` in typescript... `random` is more explicit of course, but does not play so nicely with the other functions: `string.(numeric|alphanumeric|alpha)` - or `any` of those three...

Hmm, you're right, I didn't consider that people would think "no constraints". Honestly, I would be fine with either "random" or "any".

src/modules/string/index.ts
ST-DDT: Or maybe `without`?

src/modules/string/index.ts
ST-DDT: > but does not play so nicely with the other functions: `string.(numeric|alphanumeric|alpha)` - or `any` of those three...

Note its not any of those three its a mixture of them (and more).

src/modules/string/index.ts
Shinigami92: I'm tempted to exclude but didn't thought about it enough yet

src/modules/string/index.ts
xDivisionByZerox: This looks really weird, but done.

src/modules/random/index.ts
```diff
@@ -388,44 +231,16 @@ export class Random {
...
-    while (result.length < length) {
-      result += this.faker.helpers.arrayElement(allowedDigits);
-    }
-
-    return result;
+    deprecated({
+      deprecated: 'faker.random.numeric()',
+      proposed: 'faker.string.numeric()',
+      since: '8.0',
+      until: '8.0',
```
Shinigami92, line 238:
```suggestion
      until: '9.0',
```

src/modules/database/index.ts
```diff
@@ -70,6 +70,6 @@ export class Database {
    */
   mongodbObjectId(): string {
     // strip the "0x" from the hexadecimal output
-    return this.faker.datatype.hexadecimal(24).replace('0x', '').toLowerCase();
+    return this.faker.string.hexadecimal(24).replace('0x', '').toLowerCase();
```
ST-DDT, line 73: This can be simplified using options.

src/modules/finance/index.ts
```diff
@@ -321,8 +321,9 @@ export class Finance {
    * faker.finance.ethereumAddress() // '0xf03dfeecbafc5147241cc4c4ca20b3c9dfd04c4a'
    */
   ethereumAddress(): string {
-    const address = this.faker.datatype.hexadecimal(40).toLowerCase();
-    return address;
+    const addressPrefix = '0x';
+    const address = this.faker.string.hexadecimal(40).toLowerCase();
+    return addressPrefix + address;
```
ST-DDT, line 326: This can be simplified using options.

src/modules/string/index.ts
```diff
@@ -0,0 +1,390 @@
...
+   * @example
+   * faker.string.alpha() // 'b'
+   * faker.string.alpha(10) // 'qccrabobaf'
+   * faker.string.alpha({ count: 5, casing: 'upper', bannedChars: ['A'] }) // 'DTCIC'
+   */
+  alpha(
+    options:
+      | number
+      | {
+          count?: number;
```
ST-DDT, line 145: Maybe instead of `count` use `length`?

src/modules/string/index.ts
```diff
@@ -0,0 +1,390 @@
...
+   * @example
+   * faker.string.alphanumeric() // '2'
+   * faker.string.alphanumeric(5) // '3e5v7'
+   * faker.string.alphanumeric(5, { bannedChars: ["a"] }) // 'xszlm'
+   */
+  alphanumeric(
+    options:
+      | number
+      | {
+          count?: number;
```
ST-DDT, line 211: Maybe instead of `count` use `length`?

test/string.spec.ts
```diff
@@ -0,0 +1,466 @@
...
+  'numeric',
+];
+
+describe('string', () => {
+  afterEach(() => {
+    faker.locale = 'en';
+  });
+
+  for (const seed of seededRuns) {
+    describe(`seed: ${seed}`, () => {
```
ST-DDT, line 24: This should use the seededTestFactory.

ST-DDT (changes requested):

Shinigami92 (changes requested):
**BLOCKED**: we cannot name the module `String` as this is the JavaScript global String constructor and we would run into the same issues as with `_Date` module (were we actually also want to find a better name for, like e.g. `DateModule` or `Temporal`)

If we name it `StringModule` please first merge #932

src/modules/string/index.ts
```diff
@@ -0,0 +1,402 @@
...
+   *
+   * @param length Length of the generated string. Defaults to `1`.
+   *
+   * @example
+   * faker.string.hexadecimal() // 'b'
+   * faker.string.hexadecimal(10) // 'aE13F044fb'
+   *
+   * @since 8.0.0
+   */
+  hexadecimal(length = 1): string {
```
ST-DDT, line 370: Why no `case` option here?

src/modules/string/index.ts
```diff
@@ -0,0 +1,402 @@
...
+   * @example
+   * faker.string.uuid() // '4136cd0b-d90b-4af7-b485-5d1ded8db252'
+   *
+   * @since 8.0.0
+   */
+  uuid(): string {
+    const RFC4122_TEMPLATE = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
+    const replacePlaceholders = (placeholder: string) => {
+      const random = this.faker.datatype.number({ min: 0, max: 15 });
+      const value = placeholder === 'x' ? random : (random & 0x3) | 0x8;
```
ST-DDT, line 353:
```suggestion
      const value = placeholder === 'x'
          ? this.faker.datatype.number({ min: 0, max: 15 }) 
          : this.faker.datatype.number({ min: 8, max: 11 });
```

Or however this should be formatted.

xDivisionByZerox:
We should consider a `string.special` method which returns a string of a given length (or default) with only special characters (example: '!', '"', '$').

src/modules/string/index.ts
xDivisionByZerox: Why is this implemented like this anyway? Can we just do 2 `replace` cycles on the template (one for 'x' and one for 'y')?

ST-DDT: It might be a neglible performance thing.
For me, it isn't that important.
What do the others think is best?

ST-DDT:
Sounds like a useful method.
It should have a docs comment that it can be used to test value escaping, so that users can tell what the method is for more easily.

I would also suggest a method that returns higher unicode/UTF-8 characters (e.g. `ä¿`, some non latin letters,...) to check for encoding errors.

src/modules/string/index.ts
Shinigami92: I named it `sample`

src/modules/string/index.ts
Shinigami92: Uh this looks heavy... 😖
Is it okay if I don't touch this for now in this PR?
Should we add a TODO, create an issue or just open another PR after this got merged?

Shinigami92:
> We should consider a `string.special` method which returns a string of a given length (or default) with only special characters (example: '!', '"', '$').

> Sounds like a useful method.
It should have a docs comment that it can be used to test value escaping, so that users can tell what the method is for more easily.
> 
> I would also suggest a method that returns higher unicode/UTF-8 characters (e.g. `ä¿`, some non latin letters,...) to check for encoding errors.

Lets extract this to it's own issue (if not already happen)

Edit: Extracted to https://github.com/faker-js/faker/issues/1467

Shinigami92:
@ST-DDT Okay I think I have done everything except the rewrite of the tests to use the factory.
I run out of time for today, would you like to rewrite them today? Otherwise I will do that tomorrow. 🤗

src/modules/string/index.ts
Shinigami92: Extracted to https://github.com/faker-js/faker/issues/1465#issue-1412183179

src/modules/string/index.ts
Shinigami92: Extracted to https://github.com/faker-js/faker/issues/1466

ST-DDT:
I'm not sure whether casing should be case instead.
In JS the method is called to lower/upper case as well.

If I google for [character case](https://www.google.com/search?q=character+case) I get 6.390.000.000 results.
If I google for [character casing](https://www.google.com/search?q=character+casing) I only get 21.700.000 results.
That's more than two orders of magnitude in difference!

Could a native English speaker comment whether (char) casing is a real word and which one is the correct one context wise?

src/modules/string/index.ts
```diff
@@ -0,0 +1,425 @@
...
+          bannedChars?: readonly LiteralUnion<AlphaNumericChar>[] | string;
+        } = {}
+  ): string {
+    if (typeof options === 'number') {
+      options = {
+        length: options,
+      };
+    }
+
+    const { casing = 'mixed', length = 1 } = options;
```
ST-DDT: Might be nit picky:

```suggestion
    const { length = 1, casing = 'mixed' } = options;
```

For consistency with the signature + other methods.

Alternatively, sort the deconstructors alphabetically, if possible.

Shinigami92:
I decided against case because in some context it can be conflict with switch-case.
I explicitly want to move away from Parameter names that can collide with JS or TS syntax.

If you have another parameter name proposal, please tell us.

src/modules/random/index.ts
```diff
@@ -337,12 +222,16 @@ export class RandomModule {
    * @param options.casing The casing of the characters. Defaults to `'lower'`.
    * @param options.bannedChars An array of characters and digits which should be banned in the generated string. Defaults to `[]`.
    *
+   * @see faker.string.alphaNumeric()
```
ejcheng:
```suggestion
   * @see faker.string.alphanumeric()
```

test/random.spec.ts
```diff
@@ -194,6 +194,7 @@ describe('random', () => {
             faker.random.alpha({
               count: 5,
               bannedChars,
+              casing: 'lower',
```
ST-DDT: AFAICT this isn't needed because all characters are banned.

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+
+describe('string', () => {
+  seededTests(faker, 'string', (t) => {
+    t.describe('alpha', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
+        .it('with length', { length: 6 })
+        .it('with casing', { casing: 'lower' })
+        .it('with casing', { casing: 'upper' })
+        .it('with casing', { casing: 'mixed' })
```
ST-DDT: Please give them unique names, so it doesn't break if you move them.

e.g. `with casing = lower`

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        });
+    });
+
+    t.describe('alphanumeric', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
+        .it('with length', { length: 6 })
+        .it('with casing', { casing: 'lower' })
+        .it('with casing', { casing: 'upper' })
+        .it('with casing', { casing: 'mixed' })
```
ST-DDT: Same here.

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+          bannedChars: 'lmnopqrstu67890',
+        });
+    });
+
+    t.describe('hexadecimal', (t) => {
+      t.it('noArgs')
+        .it('with length', { length: 6 })
+        .it('with casing', { casing: 'lower' })
+        .it('with casing', { casing: 'upper' })
+        .it('with casing', { casing: 'mixed' })
```
ST-DDT: And here

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+          casing: 'lower',
+          prefix: '',
+        });
+    });
+
+    t.describe('numeric', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
+        .it('with length', { length: 6 })
+        .it('with casing', { allowLeadingZeros: true })
```
ST-DDT:
```suggestion
        .it('with allowLeadingZeros', { allowLeadingZeros: true })
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        .it('with bannedChars', { bannedDigits: '12345' })
+        .it('with length, casing and bannedChars', {
+          length: 7,
+          bannedDigits: '12345',
+          allowLeadingZeros: true,
+        });
+    });
+
+    t.describe('sample', (t) => {
+      t.it('noArgs').itRepeated('repeated', 5, 5);
```
ST-DDT:
```suggestion
      t.it('noArgs').itRepeated('with length parameter', 5, 5);
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+    });
+
+    t.describe('uuid', (t) => {
+      t.itRepeated('repeated', 5);
+    });
```
ST-DDT:
```suggestion
    t.itRepeated('uuid', 5);
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        .it('with length, casing and empty prefix', {
+          length: 7,
+          casing: 'lower',
+          prefix: '',
+        });
+    });
+
+    t.describe('numeric', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
```
ST-DDT:
```suggestion
        .itRepeated('with length parameter', 5, 5)
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        .it('with length, casing and bannedChars', {
+          length: 7,
+          casing: 'lower',
+          bannedChars: 'lmnopqrstu67890',
+        });
+    });
+
+    t.describe('hexadecimal', (t) => {
+      t.it('noArgs')
+        .it('with length', { length: 6 })
```
ST-DDT, line 42: Why no length parameter check here?

```suggestion
        .itRepeated('with length parameter', 5, 5)
        .it('with length', { length: 6 })
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        .it('with length, casing and bannedChars', {
+          length: 7,
+          casing: 'lower',
+          bannedChars: 'lmnopqrstu',
+        });
+    });
+
+    t.describe('alphanumeric', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
```
ST-DDT:
```suggestion
        .itRepeated('with length parameter', 5, 5)
```

test/string.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+import { seededTests } from './support/seededRuns';
+import { times } from './support/times';
+
+const NON_SEEDED_BASED_RUN = 5;
+
+describe('string', () => {
+  seededTests(faker, 'string', (t) => {
+    t.describe('alpha', (t) => {
+      t.it('noArgs')
+        .itRepeated('repeated', 5, 5)
```
ST-DDT:
```suggestion
        .itRepeated('with length parameter', 5, 5)
```

src/modules/vehicle/index.ts
```diff
@@ -123,16 +125,19 @@ export class VehicleModule {
...
       min: 0,
       max: 9,
     })}${this.faker.datatype.number({
       min: 0,
       max: 9,
-    })}${this.faker.random.alpha({ count: 3, casing: 'upper' })}`.toUpperCase();
+    })}${this.faker.string.alpha({
+      length: 3,
+      casing: 'upper',
+    })}`.toUpperCase();
```
ejcheng:
```suggestion
    })}${this.faker.string.alpha({
      length: 3,
      casing: 'upper',
    })}`
```
Isn't the `toUpperCase()` call here redundant?

src/modules/string/index.ts
```diff
@@ -0,0 +1,425 @@
...
+   *
+   * @param length Length of the generated string. Max length is `2^20`. Defaults to `10`.
+   *
+   * @example
+   * faker.string.sample() // 'Zo!.:*e>wR'
+   * faker.string.sample(5) // '6Bye8'
+   *
+   * @since 8.0.0
+   */
+  sample(length = 10): string {
```
ST-DDT, line 387: Why doesn't this have options?

ejcheng (changes requested):

ST-DDT (changes requested):

src/modules/string/index.ts
Shinigami92: We copied it over from `random.string`
Lets update signatures in #1349

test/string.spec.ts
Shinigami92: `hexadecimal` doesn't have a only number option => also moved to #1349

test/string.spec.ts
```diff
@@ -0,0 +1,483 @@
...
+      t.it('noArgs')
+        .itRepeated('with length parameter', 5, 5)
+        .it('with length', { length: 6 })
+        .it('with allowLeadingZeros', { allowLeadingZeros: true })
+        .it('with bannedChars', { bannedDigits: '12345' })
+        .it('with length, allowLeadingZeros and bannedChars', {
+          length: 7,
+          allowLeadingZeros: true,
+          bannedDigits: '12345',
+        });
```
ejcheng:
```suggestion
        .it('with bannedDigits', { bannedDigits: '12345' })
        .it('with length, allowLeadingZeros and bannedDigits', {
          length: 7,
          allowLeadingZeros: true,
          bannedDigits: '12345',
        });
```
Should this be `bannedDigits` because the parameter is called `bannedDigits`, or are we using `bannedChars` for everything?

Shinigami92: You are right, was a copy paste issue

ejcheng (dismissed):

Shinigami92 (approved):

ejcheng (approved):

ST-DDT (approved):
There are some inconsistencies in this PR that needs to be resolved in later PRs as part of https://github.com/faker-js/faker/issues/1349 and avoided in future PRs .

e.g. the alphaNumeric function changed its signature during the move, but similar signature adjustments in other methods were moved to new/separate issues.

I only approve this because this is a massive PR, fixing these would make the PR even harder to review than it is already. 
It also blocks further development and we have issues to track the remaining todos.

PR #1161, opened 2022-07-18
refactor(fake): move to `helpers`

xDivisionByZerox:
This PR moves the `fake()` method from the `Fake` module into the `Helpers` module. The reason is, that there is no reason for it being a standalone module when it is clearly a utility method. 

This was also discussed in #805 ([direct link](https://github.com/faker-js/faker/discussions/805#discussioncomment-2539551)).


pkuczynski (dismissed):

Shinigami92 (commented):
Also just add a deprecation JSDoc Tag to the module + its constructor

Did you also update references in docs and readme?

xDivisionByZerox:
> Also just add a deprecation jedoch Tag to the module + its constructor

This is not possible since this would log a warning on Faker instance creation, which violates one of our test suites. I mean I can still add a JSDoc with a `@deprecated` parameter, but no deprecated function call in constructor.

Shinigami92:
> > Also just add a deprecation jedoch Tag to the module + its constructor
> 
> 
> 
> This is not possible since this would log a warning on Faker instance creation, which violates one of our test suites. I mean I can still add a JSDoc with a `@deprecated` parameter, but no deprecated function call in constructor.

Good catch, that's true

Shinigami92 (dismissed):

ST-DDT (dismissed):

pkuczynski (dismissed):

ejcheng (dismissed):

xDivisionByZerox:
We had 4 individual approvals and this is non blocking. I think we can merge this.

ST-DDT:
The tests seem to fail.

xDivisionByZerox:
That's odd, but I'll have a look.

xDivisionByZerox:
> The tests seem to fail.

It was a missing test with the new test factory. @ST-DDT can you have a look if I've done this the right way?

test/helpers.spec.ts
```diff
@@ -86,6 +86,10 @@ describe('helpers', () => {
     t.describe('objectValue', (t) => {
       t.it('simple', { a: 1, b: 2, c: 3 });
     });
+
+    t.describe('fake', (t) => {
+      t.it('with plain string', 'my test string');
```
ST-DDT, line 91: Yes, this is correct, but please also try a with an actual template and a template with params.

ST-DDT (changes requested):

test/helpers.spec.ts
xDivisionByZerox: But then the test for no seeds is useless or not? In that case, I don't get why we need to do this. May you explain or link to a source where you already gave information for the why? I just want to understand.

xDivisionByZerox: So what would we need all of this for?
https://github.com/faker-js/faker/blob/dd6f5a29d978177864c80438f05231f53b903d89/test/helpers.spec.ts#L487-L609

ST-DDT: The fixed seeded tests are meant to check the values with an equality check.
The random seeded tests are meant to check the values with a regex or a similar vague check.

The tests you quoted would still pass, if the result would be the original input with the template input joined together.

`faker.helpers.fake('{{phone.number}}')` -> `'{{phone.number}}1234'`
The seeded tests uncover these issues, if you read the snap file changes carefully.
The random seeded don't (always) do the same.

ST-DDT (dismissed):

test/helpers.spec.ts
```diff
@@ -479,6 +486,130 @@ describe('helpers', () => {
...
+          expect(faker.helpers.fake('{{hello')).toBe('{{hello');
+          expect(faker.helpers.fake('hello{{')).toBe('hello{{');
+        });
+
+        it('should be able to handle only }} brackets', () => {
+          expect(faker.helpers.fake('hello}}')).toBe('hello}}');
+          expect(faker.helpers.fake('}}hello')).toBe('}}hello');
+        });
+
+        it('should be able to handle reverted brackets', () => {
```
import-brain, line 585:
```suggestion
        it('should be able to handle inverted brackets', () => {
```

pkuczynski:
I am wondering if we should keep this function public? Whatever this function does is available via string templates. As a downside it requires users to know shape of the definitions and prevent us from strong typings. So what's the point of having it exposed publicly?

xDivisionByZerox:
> I am wondering if we should keep this function public? Whatever this function does is available via string templates. As a downside it requires users to know shape of the definitions and prevent us from strong typings. So what's the point of having it exposed publicly?

So you propose to deprecate this feature (which would make moving it unnecessary). Do I understand you correctly? If so please open a discussion or issue where we can debate the pros and cons of moving and deprecating. Please also link to this PR and other discussions, issues, and comments where this was already proposed/argued.

This PR has the goal to move the `fake` function into a different module. I don't want to discuss anything groundbreaking different here. 

That said, I will put this PR `on-hold` so there is no work done until your proposal is discussed.

ST-DDT:
IMO we should move faker.fake short term.

If we are going to remove it, then we need a concept on how to deal with the templates that currently rely on the faker.fake method such as findName, fullName etc. There might be a way to do that, but this requires detailed planning and feedback from the community.
We can discuss the next steps in our next team meeting.

src/modules/helpers/index.ts
```diff
@@ -454,4 +455,134 @@ export class Helpers {
...
+
+  /**
+   * Generator for combining faker methods based on a static string input.
+   *
+   * Note: We recommend using string template literals instead of `fake()`,
+   * which are faster and strongly typed (if you are using TypeScript),
+   * e.g. ``const address = `${faker.address.zipCode()} ${faker.address.city()}`;``
+   *
+   * This method is useful if you have to build a random string from a static, non-executable source
+   * (e.g. string coming from a user, stored in a database or a file).
```
Shinigami92, line 467: We have another issue/pr somewhere that thinks about fully remove `fake`
But this is IMO the reason why we CANNOT remove it at all and why it has an existential purpose!

Shinigami92 (approved):
The implementation is approved, but I think this PR needs a rebase as it is older than the test overhaul, is it?

src/modules/helpers/index.ts
pkuczynski: We can remove it from the public interface as it looks to deeply into our localization shapes. It made sense when string templates were not the thing, but now they are, so it can be deprecated in favor of it. I listed all the reasons in our meeting notes from 2022-07-28, have a look...

ST-DDT:
> but I think this PR needs a rebase as it is older than the test overhaul, is it?

It has been updated since.

ST-DDT (approved):

1LISA2:
下载
下载 (1)
very nice

PR #1163, opened 2022-07-18
chore: rename Gender to Sex

### LINKED ISSUE #352: Use enum for male/female references

hankucz:
Fixes #352

Questions:
* I don't know how to deprecate `Gender` type in favor of `Sex`
* I haven't added `Intersex` nor `Nonbinary` to the enum, as I felt that the way it is used, it would be better not to specify the sex, which would lead to randomizing the value despite the sex. For example 'faker.name.firstName()` would provide either a female or male first name, which IMHO would provide inclusivity for non-binary data...
* I could also create a separate category of names, which are not sex-specific, like `Alex, Max, Sam, etc` and then I could add a new enum value for picking them up. Although some (or maybe even many) languages do not have such names (eg. in Polish, every female name must end with an `a`, and if a name does not, it is considered male - there are no exceptions to this rule).
* Some other fixes in this module will follow in a separate PR, for example #352


xDivisionByZerox:
> I don't know how to deprecate Gender type in favor of Sex

As far as I can tell there is no way to do that other than documentation since the type will be gone as soon as the TypeScript code is compiled to JavaScript.

> I haven't added Intersex nor Nonbinary to the enum, as I felt that the way it is used, it would be better not to specify the sex, which would lead to randomizing the value despite the sex. For example 'faker.name.firstName()` would provide either a female or male first name, which IMHO would provide inclusivity for non-binary data...

> I could also create a separate category of names, which are not sex-specific, like Alex, Max, Sam, etc and then I could add a new enum value for picking them up. Although some (or maybe even many) languages do not have such names (eg. in Polish, every female name must end with an a, and if a name does not, it is considered male - there are no exceptions to this rule).

We have already talked about this possibility in a team meeting. But, you also came to the conclusion that it's not as easy as just providing an extra set of names to pick from (your Polish example). Because of that, we decided to keep the options as they are, for now. When the library is considered stable and easier maintainable we will think about such features.

src/modules/name/index.ts
```diff
@@ -1,26 +1,26 @@
 import type { Faker } from '../..';
 
-export enum Gender {
-  female = 'female',
-  male = 'male',
+export enum Sex {
+  Female = 'female',
+  Male = 'male',
```
xDivisionByZerox, line 5: Seems like we do not have any linter rules for this. But, it's common sense/best practice to have enum values written in UPPER_SNAKE_CASE. Might also be that I'm just really biased about formatting. I'd like at least one maintainer to comment on this, before flagging it as required.

xDivisionByZerox:
The current change would be considered a breaking change since the Gender enum would be removed. But like I said in an earlier comment, I'm not sure if you can deprecate and enum. I'm just pinging @Shinigami92 and @ST-DDT so they will see this (but they are currently on vacation).

One other think is that I'd like to get the additional type (now `SexValue`) removed. But this doesn't belong in this PR. Just leaving this here for reference.

Otherwise this looks good to me.

src/index.ts
```diff
@@ -37,8 +37,7 @@ export type {
   NumberColorFormat,
   StringColorFormat,
 } from './modules/color';
-export { Gender } from './modules/name';
-export type { GenderType } from './modules/name';
```
Shinigami92, line 41: These should not be removed, but just marked as deprecated

So I expect this:

```ts
export { Gender, Sex } from './modules/name';
export type { GenderType, SexType } from './modules/name';
```

src/modules/name/index.ts
Shinigami92: That is more a Java thing I think
If you look into https://www.typescriptlang.org/docs/handbook/enums.html it is more driven by C# and they use PascalCase

src/modules/name/index.ts
```diff
@@ -1,26 +1,26 @@
...
 }
 
-export type GenderType = 'female' | 'male';
+type SexValue = `${Sex}`;
```
Shinigami92, line 8:
```suggestion
export type SexValue = `${Sex}`;
```

src/modules/name/index.ts
```diff
@@ -1,26 +1,26 @@
...
+ * @param param2.generic Non-sex definitions.
  * @param param2.female Female definitions.
  * @param param2.male Male definitions.
- * @returns Definition based on given gender.
+ * @returns Definition based on given sex.
  */
 function selectDefinition(
   faker: Faker,
-  gender: GenderType | undefined,
+  sex: Sex | SexValue | undefined,
```
Shinigami92, line 23: Theoretically this is duplicated 🤔 
The enum in TypeScript is not really a real thing but will be resolved as its string replacement
So now we have the final type: `'female' | 'male' | 'female' | 'male' | undefined`

I personally would prefer to only use the string value or the types everywhere but not the enum directly
The enum should only be something for an enduser for better UX to be used

I will let decide and discuss with @ST-DDT which strategy we should use (also for future occurrences)

ST-DDT: IMO we should only apply minimal neccessary changes in this PR e.g. deprecate Gender* and replace it with Sex*.

src/modules/name/index.ts
hankucz: @pkuczynski also favored using `FEMALE`, but I checked what TypeScript documentation says, and their examples were written with PascalCase. So I am not sure what to do now...

Shinigami92: PascalCase

hankucz:
> The current change would be considered a breaking change since the Gender enum would be removed. But like I said in an earlier comment, I'm not sure if you can deprecate and enum.

I am not sure if anybody would use the enum, as initially, I changed types to use it, and all tests were failing compilation when called with strings like `firstName('female')`, which is how they all were documented. I was thinking whether it makes sense to export the enum at all, and maybe only use the string type? Eventually, I decided to go with both but export only enum. What do you think?

> One other think is that I'd like to get the additional type (now SexValue) removed. But this doesn't belong in this PR. Just leaving this here for reference.

I can't do it, if we want to support param values as strings, eg `firstName('female')` - or I don't know how to do it...

src/index.ts
hankucz: How to make it depreacted? Both @pkuczynski and @xDivisionByZerox said it's not possible...

src/modules/name/index.ts
hankucz: What's the value of exporting it? I thought we would rather like to drive people to use enum?

src/modules/name/index.ts
hankucz: @Shinigami92 as I explained in the other comment, when I kept using only enum, it was getting type errors in my IDE:

image

Bear in mind, that this enum was not used internally either...

hankucz: @ST-DDT changing only the enum was not enough, as I also had to change param names and documentation. So I felt that using this enum would make even more sense, as now we have it done properly... Please advise what shall I do?

src/modules/name/index.ts
Shinigami92: No, we do not force people to use enums.
Enums in TypeScript are theoretically a design issue, as enums doesn't exist in JavaScript.
At the end of the day only numbers or strings are there after the compiled output.
So these are only there for TypeScript users who like using enums. It's only for UX

src/index.ts
Shinigami92:
```ts
/**
 * @deprecated
 */
export enum Gender {
  female = 'female',
  male = 'male',
}
```

src/modules/name/index.ts
Shinigami92: Wait what? 👀 I need to test that out, I was not aware that these wont work. Maybe we decided to not use const enums but with non const enums it is not working anymore or something like this.

Shinigami92: When I try it in main branch it works fine:

image

src/index.ts
hankucz: Done...

src/modules/name/index.ts
hankucz: Ok, I understand; however, if we export the enum, should we still export the basic type? Wouldn't this be redundant?

src/modules/name/index.ts
xDivisionByZerox: TBH, I don't really care, as long as it is consistent. We could use [`@typescript-eslint´/naming-convention`](https://typescript-eslint.io/rules/naming-convention/) to ensure this. But that's a topic for another PR.

src/modules/name/index.ts
xDivisionByZerox: It's not redundant since a user (that don't like enums) could potentially explicitly type their variables.
```
import { faker, SexValue } from '@faker-js/faker';

const sex: SexValue = 'female';
const name = faker.name.firstName(sex);
```

src/modules/name/index.ts
hankucz: Yes, this works if I keep definition as `firstName(sex?: Sex | SexValue): string` or  change it to `firstName(sex?: SexValue): string` (as it is currently).

It did not work when I had it as enum: `firstName(sex?: Sex): string`

So should I keep `sex?: Sex | SexValue` or just change it to plain `sex?: SexValue`?

src/modules/name/index.ts
pkuczynski: Do you really think someone would do that instead of:

```
import { faker, Sex } from '@faker-js/faker';

const sex: Sex = Sex.Female;
const name = faker.name.firstName(sex);
```

I really doubt it. If anybody is into typescript, he would use enums to abstract string values. If he is into plain js, he would use strings.

pkuczynski: I also think that `SexValue` sucks as a name and having both `Sex` and `SexValue` is odd...

src/modules/name/index.ts
Shinigami92: `firstName(sex?: SexType): string`
and rename `SexValue` "back" to `SexType` as previous `GenderType`

src/modules/name/index.ts
Shinigami92: @xDivisionByZerox would you like to open a PR for https://typescript-eslint.io/rules/naming-convention/?

src/index.ts
Shinigami92, line 41: The `GenderType` is still missing, as same as the `SexType`

src/modules/name/index.ts
```diff
@@ -70,19 +81,19 @@ export class Name {
...
+   * @param sex The optional sex to use.
    * Can be either `'female'` or `'male'`.
    *
    * @example
    * faker.name.firstName() // 'Antwan'
    * faker.name.firstName("female") // 'Victoria'
    * faker.name.firstName("male") // 'Tom'
    */
-  firstName(gender?: GenderType): string {
+  firstName(sex?: Sex | SexValue): string {
```
Shinigami92, line 92:
```suggestion
  firstName(sex?: SexType): string {
```

src/modules/name/index.ts
```diff
@@ -92,19 +103,19 @@ export class Name {
...
+   * @param sex The optional sex to use.
    * Can be either `'female'` or `'male'`.
    *
    * @example
    * faker.name.lastName() // 'Hauck'
    * faker.name.lastName("female") // 'Grady'
    * faker.name.lastName("male") // 'Barton'
    */
-  lastName(gender?: GenderType): string {
+  lastName(sex?: Sex | SexValue): string {
```
Shinigami92, line 114:
```suggestion
  lastName(sex?: SexType): string {
```

src/modules/name/index.ts
```diff
@@ -114,19 +125,19 @@ export class Name {
...
+   * @param sex The optional sex to use.
    * Can be either `'female'` or `'male'`.
    *
    * @example
    * faker.name.middleName() // 'Доброславівна'
    * faker.name.middleName("female") // 'Анастасівна'
    * faker.name.middleName("male") // 'Вікторович'
    */
-  middleName(gender?: GenderType): string {
+  middleName(sex?: Sex | SexValue): string {
```
Shinigami92, line 136:
```suggestion
  middleName(sex?: SexType): string {
```

src/modules/name/index.ts
```diff
@@ -152,14 +162,19 @@ export class Name {
    *
    * @deprecated Use faker.name.fullName() instead.
    */
-  findName(firstName?: string, lastName?: string, gender?: GenderType): string {
+  findName(
+    firstName?: string,
+    lastName?: string,
+    sex?: Sex | SexValue
```
Shinigami92, line 168: We need to bring back the gender option property as this is needed as a fallback and a deprecation warning should be logged when using it

Shinigami92, line 168:
```suggestion
    sex?: SexType
```

src/modules/name/index.ts
```diff
@@ -182,19 +196,20 @@ export class Name {
     options: {
       firstName?: string;
       lastName?: string;
-      gender?: GenderType;
+      sex?: Sex | SexValue;
```
Shinigami92, line 199: Same here with gender property

+

```suggestion
      sex?: SexType;
```

src/modules/name/index.ts
```diff
@@ -182,19 +196,20 @@ export class Name {
...
     } = {}
   ): string {
     const {
-      gender = this.faker.helpers.arrayElement(['female', 'male']),
-      firstName = this.firstName(gender),
-      lastName = this.lastName(gender),
+      sex = this.faker.helpers.arrayElement([Sex.Female, Sex.Male]),
```
Shinigami92, line 203:
```suggestion
      sex = this.faker.helpers.arrayElement<SexType>(['female', 'male']),
```

src/modules/name/index.ts
```diff
@@ -236,18 +250,17 @@ export class Name {
...
-   * Can be either `'female'` or `'male'`.
+   * @param sex The optional sex to use. Can be either `'female'` or `'male'`.
    *
    * @example
    * faker.name.prefix() // 'Miss'
    * faker.name.prefix('female') // 'Ms.'
    * faker.name.prefix('male') // 'Mr.'
    */
-  prefix(gender?: GenderType): string {
+  prefix(sex?: Sex | SexValue): string {
```
Shinigami92, line 260:
```suggestion
  prefix(sex?: SexType): string {
```

src/modules/name/index.ts
xDivisionByZerox: Yes, I really like having rules and documentation^^

src/modules/name/index.ts
hankucz: OK. Done

src/index.ts
```diff
@@ -37,8 +37,7 @@ export type {
...
+export type { Gender, GenderType, Sex, SexType } from './modules/name';
```
Shinigami92, line 40: `Gender` and `Sex` (the enums) needs to be a **real** export, not a `export type`

src/modules/name/index.ts
```diff
@@ -30,13 +43,16 @@ function selectDefinition(
...
   let values: string[] | undefined;
-  switch (gender) {
-    case 'female':
+
+  switch (sex) {
+    case Sex.Female:
       values = female;
       break;
-    case 'male':
+
```
Shinigami92, line 51: We can remove these unnecessary newline
```suggestion
```

src/modules/name/index.ts
```diff
@@ -30,13 +43,16 @@ function selectDefinition(
...
+    case Sex.Male:
       values = male;
       break;
+
```
Shinigami92, line 55: We can remove these unnecessary newline
```suggestion
```

src/modules/name/index.ts
```diff
@@ -152,14 +167,15 @@ export class Name {
...
-  findName(firstName?: string, lastName?: string, gender?: GenderType): string {
+  findName(firstName?: string, lastName?: string, sex?: SexType): string {
     deprecated({
       deprecated: 'faker.name.findName()',
       proposed: 'faker.name.fullName()',
       since: '7.4',
       until: '8.0',
     });
-    return this.fullName({ firstName, lastName, gender });
+
```
Shinigami92, line 177: We can remove these unnecessary newline
```suggestion
```

Shinigami92 (changes requested):

src/modules/name/index.ts
```diff
@@ -152,14 +167,15 @@ export class Name {
    *
    * @deprecated Use faker.name.fullName() instead.
    */
-  findName(firstName?: string, lastName?: string, gender?: GenderType): string {
+  findName(firstName?: string, lastName?: string, sex?: SexType): string {
```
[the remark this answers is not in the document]
pkuczynski: @Shinigami92 it does not make sense what you are saying. It's not a hash, these were just params. So there is no need nor way to deprecate :)

src/modules/name/index.ts
pkuczynski: Its a nice separation of the cases and its not against prettier or eslint settings. I like it and I would keep it. Keeps the code more readable.

src/modules/name/index.ts
pkuczynski: Its a nice separation of the cases and its not against prettier or eslint settings. I like it and I would keep it. Keeps the code more readable.

src/modules/name/index.ts
pkuczynski: Its a nice separation of the cases and its not against prettier or eslint settings. I like it and I would keep it. Keeps the code more readable.

src/modules/name/index.ts
pkuczynski: What's the point of having enum if we can't use it? I am against the suggested change.

src/modules/name/index.ts
```diff
@@ -30,13 +38,16 @@ function selectDefinition(
   }: { generic?: string[]; female?: string[]; male?: string[] } = {}
 ) {
   let values: string[] | undefined;
-  switch (gender) {
-    case 'female':
+
+  switch (sex) {
+    case Sex.Female:
```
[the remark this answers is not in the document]
pkuczynski: What's the point of having enum if we can't use it? I am against the suggested change.

hankucz:
@xDivisionByZerox @ST-DDT @Shinigami92 is this ready for merge? I left some unresolved comments from @Shinigami92 - can you help me with this, as I am not sure if I should tackle them or not as it seems there are some contradictory opinions...

pkuczynski (approved):

pkuczynski (approved):

src/modules/name/index.ts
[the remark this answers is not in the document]
ST-DDT: @pkuczynski is right, this isn't an options object where the name is important, in this case its only relevant that the index remains the same.

ST-DDT (dismissed):
Looks good to me. For me it works with both the enum values and inline string values.

import-brain (approved):

Shinigami92 (approved):

src/modules/name/index.ts
```diff
@@ -1,30 +1,43 @@
 import type { Faker } from '../..';
 import { deprecated } from '../../internal/deprecated';
 
-// disabled until renamed to Sex
+/**
+ * @deprecated
```
xDivisionByZerox, line 5: In our [deprecation workflow documentation](https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md#deprecation-workflow) we require a human-readable message for the `deprecated` JSDoc parameter.

```suggestion
 * @deprecated Use Sex enum instead.
```

src/modules/name/index.ts
```diff
@@ -1,30 +1,43 @@
...
+  // disabled until renamed to Sex
   // eslint-disable-next-line @typescript-eslint/naming-convention
   female = 'female',
   // eslint-disable-next-line @typescript-eslint/naming-convention
   male = 'male',
 }
 
-export type GenderType = 'female' | 'male';
+/**
+ * @deprecated
```
xDivisionByZerox, line 16: In our [deprecation workflow documentation](https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md#deprecation-workflow) we require a human-readable message for the `deprecated` JSDoc parameter.

```suggestion
 * @deprecated Use SexType instead.
```

src/modules/name/index.ts
xDivisionByZerox: Someone should definitely propose a linter setting that handles these cases. IMO, a reviewer should not have to bother with this kind of change. If it is allowed (pipeline doesn't fail), cool - get on.

hankucz:
Suggestions from @xDivisionByZerox are done. Is this ready for merge?

pkuczynski (dismissed):

src/modules/name/index.ts
```diff
@@ -171,33 +187,45 @@ export class Name {
...
+      gender,
+      sex = gender || this.faker.helpers.arrayElement([Sex.Female, Sex.Male]),
+      firstName = this.firstName(sex),
+      lastName = this.lastName(sex),
     } = options;
 
+    if (gender) {
+      deprecated({
+        deprecated: `faker.name.fullName({ gender: '...'})`,
+        proposed: `faker.name.fullName({ sex: '...'})`,
```
Shinigami92, line 218:
```suggestion
        deprecated: `faker.name.fullName({ gender: '...' })`,
        proposed: `faker.name.fullName({ sex: '...' })`,
```

Shinigami92 (approved):

ST-DDT (approved):

Issue #1142, opened 2022-07-11
Rename `company.companyName()` to `company.name()`


xDivisionByZerox:
### Clear and concise description of the problem

The function `companyName()` in the `company` module has the redundant word 'company' in its name.

### Suggested solution

Apply our [`deprecation workflow`](https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md#deprecation-workflow) to the function.

### Alternative

Don't.

PR #1166, opened 2022-07-19
refactor(company): rename companyName to name

Minozzzi:
Hi guys, I applied te deprecated workflow on function `company.companyName()`. And I have a suggestion to the new function created, the parameters could be an object instead array positions. What do you guys think?

Related issue: #1142


Shinigami92:
We are already really slowly migrate nearly all functions to using an options object as parameter, so yes 🙂

ST-DDT (commented):
Instead of just copying the method you should call the new method from the old one.

src/modules/company/index.ts
```diff
@@ -30,10 +31,43 @@ export class Company {
...
    * faker.company.companyName() // 'Zieme, Hauck and McClure'
+   *
+   * @deprecated Use `faker.company.name()` instead
    */
   companyName(format?: number): string {
+    deprecated({
+      deprecated: 'faker.company.companyName()',
+      proposed: 'faker.company.name()',
+    });
+
```
xDivisionByZerox, line 70: Please call `this.name(format);` here instead of "reimplementing" the function as @ST-DDT stated.

xDivisionByZerox (changes requested):

xDivisionByZerox (dismissed):

import-brain (approved):

import-brain (changes requested):
Hmm, deprecation tests are failing...👀

ST-DDT:
> FAIL  test/scripts/apidoc/examplesAndDeprecations.spec.ts > examples and deprecations > Finance > transactionDescription
> AssertionError: expected "warn" to not be called at allexpected "warn" to not be called at all
> Received: 
>     1st warn call:
>     Array [
>       "[@faker-js/faker]: faker.company.companyName() is deprecated. Please use faker.company.name() instead..",
>     ]

It looks like finance.transactionDescription still calls company.companyName.

https://github.com/faker-js/faker/blob/05712b0c4d6fd6b2433eaa59c090f3f2a9290a17/src/modules/finance/index.ts#L427

src/modules/company/index.ts
```diff
@@ -47,6 +50,27 @@ export class Company {
...
+   * @example
+   * faker.company.companyName() // 'Zieme, Hauck and McClure'
+   *
+   * @deprecated Use `faker.company.name()` instead
+   */
+  companyName(format?: number): string {
+    deprecated({
+      deprecated: 'faker.company.companyName()',
+      proposed: 'faker.company.name()',
+    });
```
xDivisionByZerox, line 69: Could you please add the `since` and `until` properties? `since` represents the version that the deprecation was merged in. So the next version would be `7.4`. `until` is the next major version, since we are allowed to make breaking changes there.

xDivisionByZerox: Correct

Minozzzi:
Corrected your comments 😁

xDivisionByZerox (approved):

import-brain (dismissed):

src/modules/company/index.ts
```diff
@@ -30,10 +31,12 @@ export class Company {
    *
    * @param format The optional format index used to select a format.
    *
+   *
    * @example
-   * faker.company.companyName() // 'Zieme, Hauck and McClure'
+   * faker.company.name() // 'Zieme, Hauck and McClure'
+   *
```
ST-DDT, line 37: Here are two new unnecessary empty lines.

Minozzzi: Good!

import-brain (approved):

ST-DDT (approved):

PR #1168, opened 2022-07-19
feat(name): extract sex generator from gender to sex

### LINKED ISSUE #353: Clarify usages of sex vs gender

The codebase names variables as `gender` but the values are used as male and female. "Male" and "female" are sexes not genders. Some locale name definitions are organized by sex but also have references to gender. https://github.com/faker-js/faker/blob/44628ecbbf8967c3a1313c6abfc62ad653cafe29/src/locales/en/name/binary_gender.ts#L1

### Suggested solution

We need to understand the distinction between sex and gender and decide what role they should play in the organization of name definitions. Variable names need to be updated to reflect exactly what kind of values they can hold and in what context they should be used.

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #776: Typing of gender conflicts with common usage patterns
### Describe the bug

Although in practice `faker.name.firstName` _et al._ can accept any value, their typing limits them to `GenderType | undefined`, which is currently `'female' | 'male' | 0 | 1 | undefined`. The specific problem with this is that you can't use the return value from `faker.name.gender()`, whether or not you set `binary` to `true`, as the argument to the other methods.

### Reproduction

This works fine in JavaScript:
```javascript
faker.name.firstName(faker.name.gender());
```
but the exact same code in TypeScript gives:
```
Argument of type 'string' is not assignable to parameter of type 'GenderType | undefined'.ts (2345)
```

### Additional Info

I found out about this via a Stack Overflow question: https://stackoverflow.com/q/71743871/3001761

hankucz:
Fixes #353
Fixes #776

* extract sex generator from name.gender to name.sex
* deprecate `name.gender(true)`, which generated _binary gender_, what we would consider now as `sex`
* rename locale definitions from `binary_gender` to `sex`
* lowercase english values for `sex`
* extend english locale with `non-binary`


src/definitions/name.ts
```diff
@@ -5,7 +5,7 @@ import type { LocaleEntry } from './definitions';
  */
 export type NameDefinitions = LocaleEntry<{
   gender: string[];
-  binary_gender: string[];
+  sex: string[];
```
xDivisionByZerox, line 8: I'm unsure about this since it would elevate this change to a breaking one, which would block an early merge of all other changes in this PR. I guess a comment with a todo (to rename it on the next major release) is better for now and leave it as `binary_gender` even if we interpret it as `sex`. 

Need conformation for this from at least one other maintainer/member.

xDivisionByZerox: Nvm, I just can't read. Leave it as is.

hankucz: Why would this be a breaking change? I updated all methods to use this new locale anyway...

src/modules/name/index.ts
```diff
@@ -217,22 +217,34 @@ export class Name {
...
   }
 
+  /**
+   * Return a random sex.
+   *
+   * @example
+   * faker.name.sex() // 'female'
+   */
+  sex(): string {
+    return this.faker.helpers.arrayElement(this.faker.definitions.name.sex);
```
ST-DDT, line 245: There should be a way to generate only binary genders.

src/modules/name/index.ts
```diff
@@ -217,22 +217,34 @@ export class Name {
   /**
    * Return a random gender.
    *
-   * @param binary Whether to return only binary gender names. Defaults to `false`.
```
ST-DDT, line 220: The param still exists.

ST-DDT (changes requested):

src/modules/name/index.ts
hankucz: What we consider `sex` is equal to binary genders.

ST-DDT: Then the sex file should only contain that, but it does not:
https://github.com/faker-js/faker/blob/6b6c9b5a8a1f1ef8874a200953bf45d60a94a73c/src/locales/en/name/sex.ts#L1
I'm not sure whether non-binary is the correct term here, but AFAIK there are 3 sexes.

ST-DDT: Maybe we should adjust our datastructure accordingly:
`[string, string, string]`?

ST-DDT: @xDivisionByZerox @JessicaSachs Any suggestions here?

Shinigami92: Just to be clear:
In the past we had an issue where customers tried to use gender as input for firstName, like:
```ts
faker.name.firstName(faker.name.gender())
```

This is obviously WRONG and should not be used like this!
faker.name.gender (or sex) is a localized string which can not be used as input for other functions as we are using switch cases based on that value.
So we should try to make the users absolutely clear that this is not usable as inputs.
`sex` now can and will be differ from what SexType defines, as the return value of `faker.name.sex()` could return `männlich` in `de`

Beside all of this, sex doesn't really have something todo with "name" as this is still a "name" module... I would like to introduce a new module like "person" and move some functions when not all to this. But that's another task for the future.
So maybe we should leave the sex function away for now and wait on the new person module to add it?!

Shinigami92 (commented):
This is a breaking change as `binary_gender` could be used in `faker.faker('{{binary_gender}}')` without that we know it or users know that it had changed
I mark this as blocked for now and would like to first discuss if this PR is needed anyway

src/modules/name/index.ts
pkuczynski: I agree with @Shinigami92 100%. I created discussion https://github.com/faker-js/faker/discussions/1194 to decide how to take it further. Until then, we can close this thread and keep `sex` and `gender` as two separate functions. What do you think guys?

pkuczynski:
> This is a breaking change as `binary_gender` could be used in `faker.faker('{{binary_gender}}')` without that we know it or users know that it had changed
> I mark this as blocked for now and would like to first discuss if this PR is needed anyway

If we are killing `gender` enum, it feels like this is a natural good step forward to make the differentiation between `gender` and `sex`.

In my opinion `faker.fake` should be deprecated, as the same result can be achieved with string templates, which additionally allow user to provide options to method calls, instead of just using locales.

So yes, technically it _**could be**_ a breaking change, but at the same time locale formats are not officially published, so I have doubts if anybody would hook up to the internals so deep to find it. And if he did, it's kind of not our fault...

@ST-DDT @xDivisionByZerox wdyt?

pkuczynski (dismissed):

src/modules/name/index.ts
xDivisionByZerox: > Maybe we should adjust our datastructure accordingly: `[string, string, string]` or `{female: string, male: string, other: string}`?

That's exactly what I proposed in a meeting 3 weeks ago (I would link to out meeting notes but there is sadly no note about my proposal).

xDivisionByZerox:
> In my opinion faker.fake should be deprecated, as the same result can be achieved with string templates, which additionally allow user to provide options to method calls, instead of just using locales.

TBH, I'm with you on this one. Never understood the use of `fake` at all. Gues its legacy from pre-v5.5.

> So yes, technically it could be a breaking change, but at the same time locale formats are not officially published, so I have doubts if anybody would hook up to the internals so deep to find it. And if he did, it's kind of not our fault...

Not "technically". In this case, it is since the feature to use locale definition with `fake` is documented AND pushed by using it in our own codebase. Have a look at these deprecation statements:

https://github.com/faker-js/faker/blob/ca7cde59ec3747a93d1d47355a83287696628eb5/src/modules/address/index.ts#L103-L109

https://github.com/faker-js/faker/blob/ca7cde59ec3747a93d1d47355a83287696628eb5/src/modules/address/index.ts#L126-L132

So for me, there is no question whether renaming `binary_gender` is a breaking change. It clearly is.

pkuczynski:
> TBH, I'm with you on this one. Never understood the use of fake at all. Gues its legacy from pre-v5.5.

See my (same) comment https://github.com/faker-js/faker/pull/1161#issuecomment-1195852602

> Not "technically". In this case, it is since the feature to use locale definition with `fake` is documented AND promoted by using it in our own codebase. Have a look at these deprecation statements:
> 
> https://github.com/faker-js/faker/blob/ca7cde59ec3747a93d1d47355a83287696628eb5/src/modules/address/index.ts#L103-L109
> 
> https://github.com/faker-js/faker/blob/ca7cde59ec3747a93d1d47355a83287696628eb5/src/modules/address/index.ts#L126-L132
> 
> So for me, there is no question whether renaming `binary_gender` is a breaking change. It clearly is.

These are two statements that do not mention `binary_gender`, so it's unlikely someone would be using it. Especially this data is provided in 5 locales only, so very few... So yes, this is technically breaking, but the impact is low...

src/modules/name/index.ts
pkuczynski: Which data structures? Locales?

xDivisionByZerox: Definitions, which are basically types for the locales structure. See current name definition:
https://github.com/faker-js/faker/blob/0198cb086e0121c4472187910c65ae4dbe5ffd1f/src/definitions/name.ts#L6-L34

The proposed change would look something like this:
```ts
type NameSet = {
  female: string;
  male: string;
  other: string;
}

type NameGroup = {
  female: string[];
  male: string[];
  other: string[];
}

export type NameDefinitions = LocaleEntry<{
  gender: string[];
  sex: NameSet;

  prefix?: NameGroup;
  first_name?: NameGroup;
  middle_name?: NameGroup;
  last_name?: NameGroup;

  // there are no gender-specific suffixes
  suffix: string[];

  /**
   * A list of patterns used to generate names (Fake-Pattern[]).
   */
  name: string[];

  title: NameTitleDefinitions;
}>;
```

BTW this is totally out of scope for this PR.

hankucz:
What are the next steps? Should I do anything about it?

src/locales/en/name/sex.ts
```diff
@@ -0,0 +1 @@
+export default ['female', 'male', 'non-binary'];
```
Shinigami92, line 1: IMO this is a wrong change :thinking: 
Originally this was binary_gender file, which contains translated strings which could also be used in `faker.helpers.fake` (which can be used for e.g. stored database strings)
Also now the values are in lower case as before they were in normal case starting with an upper case letter
Beside all of that, if we name this sex it could get into confusion with the sex enum / sex property of functions, but this would not work by design as this locale definitions are translated values and e.g. in locale fr it is called `Femme` and `Homme`, so they have a totally different technical meaning as the property/enum.

src/modules/name/index.ts
```diff
@@ -220,22 +220,36 @@ export class Name {
...
-   * faker.name.gender(true) // 'Female'
    */
   gender(binary?: boolean): string {
     if (binary) {
-      return this.faker.helpers.arrayElement(
-        this.faker.definitions.name.binary_gender
-      );
+      deprecated({
+        deprecated: 'faker.name.gender(true)',
+        proposed: 'faker.name.sex()',
```
Shinigami92, line 232: This change doesn't make sense at all
`sex != gender`!
Sex is the real biological / physical thing
Gender is the identity somebody feels to be

So we cannot just replace this function with sex

Shinigami92 (changes requested):

src/locales/en/name/sex.ts
pkuczynski: This continues to be stored in the dictionary, just `sex` instead of `binary_gender`. This can continue to be used as it was, just in a better way, as it no includes 3 items instead of 2, so it does express better sex selection on different forms during the registration process - more and more websites offer you to specify "non-binary" as an option. So if anything, we improved the data.

There is no reason why this data should be written in capital form, as this is just a pure noun. So why do you think it should be written in capital form? Other nouns we always write in small letters...

I believe we specify `sex` parameter as `SexType` so there should be no confusion when used via typescript. When used in plain JS, we should probably throw when an unsupported value is provided, which would solve the problem...

The naming of the function does not make any difference in this proposal vs the current situation, as even currently people were already trying to misuse it. So we should solve it on the code level and not by naming the funct this or the other way...

src/modules/name/index.ts
pkuczynski: It makes a lot of sense. So-called _binary genders_, are equal sex. If this is not true, why did we change the function's names in #1163?

JessicaSachs: Oh, we've finally gotten to _this_ discussion.

Gender identity and biological sex are different. [This is a really nice article that talks about it.](https://www.healthline.com/health/different-genders).

**Some examples of gender identities (abbreviated list):** male, female, transgender, gender neutral, non-binary, agender, pangender, genderqueer, two-spirit, and third gender. 

**Some examples of sexes:** male, female, or intersex.

With regards to the API design for things like names, I think we should support `femme`, `masculine` (or `masc`), and `unisex` as options.

With regards to supporting the binary or not... I believe the enum of "sexes" will more closely match the gender binary that some web forms assume exists.

Shinigami92: Oh my god I think I found the best source we could depend on 🙌 
[https://www.who.int/health-topics/gender](https://www.who.int/health-topics/gender#tab=tab_1:~:text=Gender%20interacts%20with%20but%20is%20different%20from%20sex%2C%20which%20refers%20to%20the%20different%20biological%20and%20physiological%20characteristics%20of%20females%2C%20males%20and%20intersex%20persons%2C%20such%20as%20chromosomes%2C%20hormones%20and%20reproductive%20organs.)

> Gender interacts with but is different from sex, which refers to the different biological and physiological characteristics of females, males and intersex persons, such as chromosomes, hormones and reproductive organs.

So I think we should use `enum Sex = female, male and intersex`

src/modules/name/index.ts
pkuczynski: Ah yes, it would be a good change, but beyond scope of this pr. And I am not sure how `faker.helpers.fake` would handle this. probably it would not?

src/modules/name/index.ts
pkuczynski: You are mixing up things. This is not about enum. Enum has been done in #1163. This is about changing `faker.name.gender(true)` to `faker.name.sex()`. As `faker.name.gender(true)` looks odd and not really logical. Is `female` and `male` the `true` gender?!?

Shinigami92: I didn't read correctly 
I thought this PR wants to replace the gender function entirely with a sex function 
But it just replaces the property passed to the gender function and introduce a new addition sex function 

So this misunderstanding was my bad 👍

Shinigami92:
Should we duplicate the binary_gender locale definition for now and remove it in v8?
That way we could merge this PR now and don't need to wait because of the breaking change for the fake function 🤔

pkuczynski:
> Should we duplicate the binary_gender locale definition for now and remove it in v8?
> That way we could merge this PR now and don't need to wait because of the breaking change for the fake function 🤔

We could if you think it's worth it. Hard to say if anybody was using it in `fake`, but I kind of doubt it...

src/modules/name/index.ts
Shinigami92: According to WHO `intersex` is the correct term
[who.int/health-topics/gender~intersex](https://www.who.int/health-topics/gender#tab=tab_1:~:text=females%2C%20males%20and-,intersex,-persons%2C%20such%20as)

Shinigami92 (changes requested):

src/modules/name/index.ts
```diff
@@ -220,22 +220,36 @@ export class Name {
...
+      });
+
+      return this.faker.name.sex();
     }
 
     return this.faker.helpers.arrayElement(this.faker.definitions.name.gender);
   }
 
+  /**
+   * Return a random sex.
```
ST-DDT, line 244: Not sure about the phrasing here:
```suggestion
   * Return a random sex (localized versions of `female`, `intersex`, or `male`).
```
So that it's clear that there are three of them and this method returns them, although I would kind of like to have an optional param to return only a binary sex.

Please also add a `@see` targeting the gender function and vice versa to mentally bind them together.

ST-DDT (changes requested):
Please merge the latest changes from master and don't forget to update/rename `src\locales\pl\name\binary_gender.ts`.

Also, if we are going to touch this method, I expect it to behave consistently (e.g. either to have three sexes or two, but not a mixture of both across different locales).

src/modules/name/index.ts
hankucz: I changed it slightly to express its localized. I can add a line that the output of this function should not be used as a param value whenever we ask for `sex`, but not sure if it's needed...

`@see` added

src/modules/name/index.ts
hankucz: `intersex` is a biological thing but does not function within laws and documents. If you are born `intersex` doctors/parents would decide for you and assign you one of the binary sexes. So I think it would be enough to stick to it, as we had it so far.

src/modules/name/index.ts
ST-DDT: > I can add a line that the output of this function should not be used as a param value whenever we ask for `sex`, but not sure if it's needed...

I think it doesnt hurt to have this. In the worst case we can link to it if someone missused that method.

src/locales/en/name/sex.ts
```diff
@@ -0,0 +1 @@
+export default ['female', 'male'];
```
Shinigami92, line 1:
```suggestion
export default ['female', 'intersex', 'male'];
```

Shinigami92 (changes requested):

Shinigami92 (commented):
Could you please add `src/locales/de/name/sex.ts` with `['divers', 'männlich', 'weiblich']`?

src/locales/en/name/sex.ts
pkuczynski: I agree with @hankucz that we should stick to _binary sex_ here. `intersex` is a biological term, which would not end up in your documents. And what this function could be used is to generate value for `sex` field on the `passport` for example. The `intersex` would not exists there, as when you are born doctors would decide on your sex and if need be run an operation... Maybe this would change in the future world, but not now. Even in US there are only few states which would do that: https://www.usbirthcertificates.com/articles/gender-neutral-birth-certificates-states

src/modules/name/index.ts
hankucz: Done

pkuczynski (dismissed):

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
...
+   *
+   * Output of this method is localised, so it should not be used to fill the parameter `sex`
+   * available in some other modules for example faker.name.firstName().
```
ST-DDT, line 276:
```suggestion
   * available in some other modules for example `faker.name.firstName()`.
```

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
...
+      });
+
+      return this.faker.name.sex();
     }
 
     return this.faker.helpers.arrayElement(this.faker.definitions.name.gender);
   }
 
+  /**
+   * Return a random sex.
```
ST-DDT, line 273:
```suggestion
   * Returns a random sex.
```

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
   /**
    * Return a random gender.
```
ST-DDT, line 248:
```suggestion
   * Returns a random gender.
```

ST-DDT (changes requested):

ST-DDT:
<details>
<summary><strong>Docs-Preview (Click to expand)</strong></summary>


grafik

grafik


</details>

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
...
    *
-   * @param binary Whether to return only binary gender names. Defaults to `false`.
+   * @param binary (deprecated) Whether to return only binary gender names. Defaults to `false`.
+   *
+   * @see faker.name.sex() if you would like to generate binary-gender value
```
import-brain, line 252:
```suggestion
   * @see faker.name.sex() - if you would like to generate binary genders
```

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
...
     return this.faker.helpers.arrayElement(this.faker.definitions.name.gender);
   }
 
+  /**
+   * Return a random sex.
+   *
+   * Output of this method is localised, so it should not be used to fill the parameter `sex`
+   * available in some other modules for example faker.name.firstName().
+   *
+   * @see faker.name.gender() if you would like to generate gender related values.
```
import-brain, line 278:
```suggestion
   * @see faker.name.gender() - if you would like to generate gender related values.
```

src/modules/name/index.ts
```diff
@@ -247,22 +247,43 @@ export class Name {
...
    */
   gender(binary?: boolean): string {
     if (binary) {
-      return this.faker.helpers.arrayElement(
-        this.faker.definitions.name.binary_gender
-      );
+      deprecated({
+        deprecated: 'faker.name.gender(true)',
+        proposed: 'faker.name.sex()',
+        since: '7.4',
```
Shinigami92, line 262:
```suggestion
        since: '7.5',
```

If there are other places in this PR, please change them as well

src/modules/name/index.ts
hankucz: I could not find an example where `-` is used, so I will not make this change, ok?

src/modules/name/index.ts
hankucz: Same thing with `-`

hankucz:
I believe I applied all of the feedback. Is there anything else I should do about it?

pkuczynski (approved):

pkuczynski (approved):

Shinigami92 (approved):

Shinigami92:
Ping @ST-DDT, you need to approve to stale your _requested changes_

ST-DDT (approved):

PR #1170, opened 2022-07-19
chore(linting): add naming convention rule

xDivisionByZerox:
Add a basic linting rule to apply naming conventions.
For now, I only added the rule that classes, interfaces, type declarations, and enum members have to be written in `PascalCase`.

I'm open to more strict definitions in this PR.


pkuczynski (dismissed):

import-brain (approved):

ST-DDT (approved):

Shinigami92 (approved):
I'm ok with that for now
But is there maybe a no-config default-preset?

xDivisionByZerox:
> I'm ok with that for now 
But is there maybe a no-config default-preset?

Not that I'm aware of. All configs we extend from do not set this rule. 

TBH, I'm a big fan of explicitly settings configs and not relying on defaults. That way you ensure your config always stays the same even if the preset you extend from decides to change their settings. But that's my opinion.

pkuczynski:
> TBH, I'm a big fan of explicitly settings configs and not relying on defaults. That way you ensure your config always stays the same even if the preset you extend from decides to change their settings. But that's my opinion.

I second that.

PR #1171, opened 2022-07-20
fix(finance.bic): remove hardcoded elements and simplify function

### LINKED ISSUE #1159: BIC generation is not realistic
While looking at the code mentioned in #1158, I noticed that `finance.bic()` generates BIC codes in a weird way that might not be representative of real SWIFT/BIC codes. There are two issues:

1. According to the real SWIFT codes listed on [this page](https://bank.codes/guides/swift-codes-guide/) (scroll down a bit), it doesn't seem like the 4th character in the code has to be a vowel. https://github.com/faker-js/faker/blob/df9c650432de2d119b4f17d61a0c554b744debd0/src/modules/finance/index.ts#L403-L404 However, the function uses `helpers.replaceSymbols()` to pick three random uppercase letters and then it appends a random vowel to that. To fix this, we could delete line 404 and replace line 403 with `this.faker.helpers.replaceSymbols('????')`.

2. Also based on the real SWIFT codes mentioned above, it doesn't seem like the 8th char of a SWIFT code is required to be a `1`. However, `1` is always the 8th char, because it is hardcoded. https://github.com/faker-js/faker/blob/df9c650432de2d119b4f17d61a0c554b744debd0/src/modules/finance/index.ts#L407 This char should be randomly generated.

Also, is there any possible way we could make this function easier to read? Personally, the ternaries spread over multiple lines make it incredibly hard to read.

hankucz:
Fixes #1159


xDivisionByZerox (dismissed):

ST-DDT:
One regex seems to be failing. Please check the regex for issues.
Is there a validator.js function for that or we could propose it there?

ST-DDT:
Edit I found this "official" regex: https://stackoverflow.com/a/52043851/4573065

hankucz:
> Edit I found this "official" regex: https://stackoverflow.com/a/52043851/4573065
> Maybe use somewhere official is source for it?

According to https://wise.com/gb/swift-codes/bic-swift-code-checker the location code is 0-9 A-Z. I checked the failed bic `YAQHKI1MIMQ` and got results:

* Your 4-digit bank code YAQH was not found, but may be correct.
* Your 2-digit country code KI is for Kiribati
* Your 2-digit location code 1M was not found but may be correct.
* Your 3-digit branch code IMQ was not found but may be correct.

Supported by wikipedia https://en.wikipedia.org/wiki/ISO_9362
But then the german version adds some more rules: https://de.wikipedia.org/wiki/ISO_9362

I found the official (demo) standard https://cdn.standards.iteh.ai/samples/84108/21078408ed25469391d0cf6c5d1b6ca9/ISO-9362-2022.pdf and it also does not say anything about those exclusions mentioned on German wiki.

ST-DDT:
Then use a simple regex. I would even remove the countryCodes from the regex and check them separately (e.g. using substring).

hankucz:
Rebased and issues fixed. Is it ready to merge?

pkuczynski (approved):

test/finance.spec.ts
```diff
@@ -484,11 +484,9 @@ describe('finance', () => {
       describe('bic()', () => {
         it('should return a random yet formally correct BIC number', () => {
           const bic = faker.finance.bic();
+          const countryCodes = ibanLib.iso3166.join('|');
           const expr = new RegExp(
-            `^[A-Z]{4}(${ibanLib.iso3166.join(
-              '|'
-            )})[A-Z2-9][A-NP-Z0-9]([A-Z0-9]{3})?\$`,
-            'i'
+            `^[A-Z]{4}(${countryCodes})[A-Z0-9]{2}([A-Z0-9]{3})?\$`
```
xDivisionByZerox, line 489: I'm sorry to be nitpicky again, but I really can't stand this RegExp. I mean RegExp is cool if you have a complex string you want to validate, with many different branches. In an internal document, we have written that we want to use as less RegExp as possible (we should really publish that soon). 

So could you split the result string into its various parts (institution code, country code, etc) and validate the parts themselves. 

Example:
```
const bic = faker.finance.bic();

const institutionCode = bic.substring(0, 4);
expect(institutionCode).toHaveLength(4);
expect(institutionCode).toMatch(/^[A-Z]*$/);

// other part validation
```

pkuczynski: This would only unnecessarily extend this test. Regexp is perfect for testing patterns and this one is not even complex. I am against the change suggested by @xDivisionByZerox

xDivisionByZerox: OK, so this is an example error message you would get. Please tell me where the error is and tell me again its not complicated.

```
AssertionError: expected 'ECUINLE12' to match /^[A-Z]{4}(AD|AE|AF|AG|AI|AL|AM|AO|AQ|AR|AS|AT|AU|AW|AX|AZ|BA|BB|BD|BE|BF|BG|BH|BI|BJ|BL|BM|BN|BO|BQ|BR|BS|BT|BV|BW|BY|BZ|CA|CC|CD|CF|CG|CH|CI|CK|CL|CM|CN|CO|CR|CU|CV|CW|CX|CY|CZ|DE|DJ|DK|DM|DO|DZ|EC|EE|EG|EH|ER|ES|ET|FI|FJ|FK|FM|FO|FR|GA|GB|GD|GE|GF|GG|GH|GI|GL|GM|GN|GP|GQ|GR|GS|GT|GU|GW|GY|HK|HM|HN|HR|HT|HU|ID|IE|IL|IM|IN|IO|IQ|IR|IS|IT|JE|JM|JO|JP|KE|KG|KH|KI|KM|KN|KP|KR|KW|KY|KZ|LA|LB|LC|LI|LK|LR|LS|LT|LU|LV|LY|MA|MC|MD|ME|MF|MG|MH|MK|ML|MM|MN|MO|MP|MQ|MR|MS|MT|MU|MV|MW|MX|MY|MZ|NA|NC|NE|NF|NG|NI|NL|NO|NP|NR|NU|NZ|OM|PA|PE|PF|PG|PH|PK|PL|PM|PN|PR|PS|PT|PW|PY|QA|RE|RO|RS|RU|RW|SA|SB|SC|SD|SE|SG|SH|SI|SJ|SK|SL|SM|SN|SO|SR|SS|ST|SV|SX|SY|SZ|TC|TD|TF|TG|TH|TJ|TK|TL|TM|TN|TO|TR|TT|TV|TW|TZ|UA|UG|UM|US|UY|UZ|VA|VC|VE|VG|VI|VN|VU|WF|WS|XK|YE|YT|ZA|ZM|ZW)[A-Z2-9][A-NP-Z0-9]([A-Z0-9]{3})?$/i
```

pkuczynski: 1. New regex is much simpler
2. Error is: the last optional part '2' is too short ;)

xDivisionByZerox: Same error with my implementation proposal:
```
expect(bic).toBeTypeOf('string');

const institutionCode = bic.substring(0, 4);
expect(institutionCode).toHaveLength(4);
expect(institutionCode).toMatch(/^[A-Z]*$/);

const countryCode = bic.substring(4, 6);
expect(countryCode).toHaveLength(2);
expect(ibanLib.iso3166).toContain(countryCode);

const locationCode = bic.substring(6, 8);
expect(locationCode).toHaveLength(2);
expect(locationCode).toMatch(/^[A-Z0-9]*$/);

const branchCode = bic.substring(8);
expect(
  branchCode.length === 0 || branchCode.length === 3,
  'branch code must be exactly either 0 or 3 characters long'
).toBe(true);
expect(branchCode).toMatch(/^[A-Z0-9]*$/);
```

Throws:
```
AssertionError: branch code must be exactly either 0 or 3 characters long: expected false to be true // Object.is equality
```

hankucz: I will do it tmrw...

ST-DDT: Please give me some more time to think about this (otherwise you might end up having to revert your changes).

import-brain: > OK, so this is an example error message you would get. Please tell me where the error is and tell me again its not complicated.
> 
> ```
> AssertionError: expected 'ECUINLE12' to match /^[A-Z]{4}(AD|AE|AF|AG|AI|AL|AM|AO|AQ|AR|AS|AT|AU|AW|AX|AZ|BA|BB|BD|BE|BF|BG|BH|BI|BJ|BL|BM|BN|BO|BQ|BR|BS|BT|BV|BW|BY|BZ|CA|CC|CD|CF|CG|CH|CI|CK|CL|CM|CN|CO|CR|CU|CV|CW|CX|CY|CZ|DE|DJ|DK|DM|DO|DZ|EC|EE|EG|EH|ER|ES|ET|FI|FJ|FK|FM|FO|FR|GA|GB|GD|GE|GF|GG|GH|GI|GL|GM|GN|GP|GQ|GR|GS|GT|GU|GW|GY|HK|HM|HN|HR|HT|HU|ID|IE|IL|IM|IN|IO|IQ|IR|IS|IT|JE|JM|JO|JP|KE|KG|KH|KI|KM|KN|KP|KR|KW|KY|KZ|LA|LB|LC|LI|LK|LR|LS|LT|LU|LV|LY|MA|MC|MD|ME|MF|MG|MH|MK|ML|MM|MN|MO|MP|MQ|MR|MS|MT|MU|MV|MW|MX|MY|MZ|NA|NC|NE|NF|NG|NI|NL|NO|NP|NR|NU|NZ|OM|PA|PE|PF|PG|PH|PK|PL|PM|PN|PR|PS|PT|PW|PY|QA|RE|RO|RS|RU|RW|SA|SB|SC|SD|SE|SG|SH|SI|SJ|SK|SL|SM|SN|SO|SR|SS|ST|SV|SX|SY|SZ|TC|TD|TF|TG|TH|TJ|TK|TL|TM|TN|TO|TR|TT|TV|TW|TZ|UA|UG|UM|US|UY|UZ|VA|VC|VE|VG|VI|VN|VU|WF|WS|XK|YE|YT|ZA|ZM|ZW)[A-Z2-9][A-NP-Z0-9]([A-Z0-9]{3})?$/i
> ```

I'm with @xDivisionByZerox on this, I would prefer to not have this show up in my terminal

ST-DDT: I have thought about this. IMO we should split this into two separate checks:

````ts
expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?\$/);
expect(ibanLib.iso3166).toContain(bic.substring(4, 6));
````

import-brain: > I have thought about this. IMO we should split this into two separate checks:
> 
> ```ts
> expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?\$/);
> expect(ibanLib.iso3166).toContain(bic.substring(4, 6));
> ```

I like this approach.

hankucz:
The suggested change by @ST-DDT was implemented into the test. Is this ready for merge?

Shinigami92 (approved):

ST-DDT (approved):

PR #1174, opened 2022-07-20
chore(user-agent): fix linter warnings

xDivisionByZerox:
This PR fixes all remaining linter warnings in `user-agent.ts`. 

To achieve this I changed the implementation of the function in a non-breaking way. I mainly simplified the `rnd` function since it had a lot of unnecessary branches. I also converted all internal functions to constants so they don't require any JSDoc.


xDivisionByZerox:
BTW, this PR brings the total number of linter warnings in the project down to  **ZERO** 🚀

src/modules/internet/user-agent.ts
```diff
@@ -44,58 +44,43 @@
 
 import type { Faker } from '../..';
 
+// I don't think this should be exposed, since it can potentially be imported via:
+// import { Arch } from '@faker-js/faker/modules/internet/user-agent';
```
pkuczynski, line 48: Similar thing with `SexType` in the other PR, where we do expos it. I think we should not. But if we expose it in one place, we should expose in the others too I think? Same with `Browser`. @Shinigami92 you were in favour of the export...

src/modules/internet/user-agent.ts
```diff
@@ -44,58 +44,43 @@
...
-        let min = 0;
-        let max = 0;
-        let return_val: string;
+/**
+ * Generates a random user-agent.
+ *
+ * @param faker An existing faker instance.
+ */
+export function generate(faker: Faker): string {
+  const weightedKeyFromObject = <T extends Record<string, number>>(
```
pkuczynski, line 59: I think we should move all of those functions to be statically defined and feed them with `faker` instance. This way they will be way faster, as they don't have to redefine on each execution... Maybe you can also wrap them into a private class if you want...

src/modules/internet/user-agent.ts
xDivisionByZerox: Well the problem here is that it literally doesn't serve any purpose since:
1. we don't use this type anywhere other than this file, in contrast to `SexType` which would be useful as an explicitly typed input value
2. it doesn't get exported from the root of the library (`import { Arch } from '@faker-js/faker'`) unlike `SexType` which we explicitly choose to export

src/modules/internet/user-agent.ts
xDivisionByZerox: I wanted to touch the original source code as little as possible (which I somewhat failed by refactoring `rnd` to `weightedKeyFromObject`), because I'm not sure at which point it's not the original source code anymore (licensing).

xDivisionByZerox: TBH I would happily rewrite this think and dismiss the weights in the browser/os versions. IMO we don't need this feature.

pkuczynski: Yeah, I don't really see a point of this either...

src/modules/internet/user-agent.ts
ST-DDT: I think we dont have to export this as it isnt possible to use it with any public facing API.
`SexType` is public facing API (relevant).

src/modules/internet/user-agent.ts
```diff
@@ -44,58 +44,43 @@
...
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-      `Invalid arguments passed to rnd. (${b ? `${a}, ${b}` : a})`
-    );
-  }
+    return return_val;
+  };
 
-  function randomLang(): string | number {
-    return rnd([
+  const randomLang = (): string => {
```
ST-DDT, line 82: This could be moved to faker directly (next to the country code method?).
(Separate PR)

src/modules/internet/user-agent.ts
ST-DDT: How about moving this method to faker.helpers?
(Separate PR)

src/modules/internet/user-agent.ts
xDivisionByZerox: I thought so myself. But, different PR as you said.

src/modules/internet/user-agent.ts
xDivisionByZerox: > How about moving this method to faker.helpers?
> (Separate PR)

I'm against this, since the only use case of this function is the internet module.
Also, `helpers` is for transformation util that doesn't generate data themself (unlike this fn).

ST-DDT: This `weightedKeyFromObject` function doesn't generate data either.
I will create an issue for it, when I have time.

Shinigami92: @xDivisionByZerox I assume @ST-DDT mean the `weightedKeyFromObject` function
It does not generate new stuff but is the same as `objectKey`, `objectValue` or `arrayElement`
But you just provide the weights

src/modules/internet/user-agent.ts
```diff
@@ -44,58 +44,43 @@
...
+    return faker.helpers.arrayElement([
```
Shinigami92, line 83: As this is already an inline lambda function, consider remove the `{`, `return` and `}`

src/modules/internet/user-agent.ts
Shinigami92, line 48: These comments should be removed

I don't think they serve a value purpose

src/modules/internet/user-agent.ts
```diff
@@ -44,58 +44,43 @@
...
-  ): string | number {
-    //calling rnd() with no arguments is identical to rnd(0, 100)
-    a = a || 0;
-    b = b || 100;
-
-    if (typeof b === 'number' && typeof a === 'number') {
-      // 9/2018 - Added faker random to ensure mersenne and seed
-      return faker.datatype.number({ min: a, max: b });
-    }
+type Browser = 'chrome' | 'iexplorer' | 'firefox' | 'safari' | 'opera';
```
Shinigami92, line 51: Export both or none
If these types can somehow be useful for users, then reexport them in `src/index.ts` otherwise consider not to export them at all

ST-DDT: IMO we should not export them.

src/modules/internet/user-agent.ts
xDivisionByZerox: Oh, I see. I thought you were talking about the entire `generate` function. My bad. In that case, we could implement a similar function in `helpers` but as I said earlier: I don't see the point. Just refactor the object keys to a plain array and apply `arrayElement` and we fine.

src/modules/internet/user-agent.ts
xDivisionByZerox: I wanted to add them so they can be discussed in the PR. At least raise attention for it.

src/modules/internet/user-agent.ts
xDivisionByZerox: Would not exposing them (removing the `export` from `Arch`) be considered a breaking change? Since someone could potentially use this type and by removing the export we would introduce errors in their codebase (due to missing type).

src/modules/internet/user-agent.ts
pkuczynski: I also do not see a need to have weighted randoms...

src/modules/internet/user-agent.ts
xDivisionByZerox: If you are strict about this formatting there is a linter rule for that^^
[arrow-body-style](https://eslint.org/docs/latest/rules/arrow-body-style)

src/modules/internet/user-agent.ts
xDivisionByZerox: > You might also want to rename some methods (e.g. browser and os -> browser and arch).

Instead of renaming the function, what about renaming the type to `OS`? TBH I have no idea what `arch` is supposed to mean in this context.

src/modules/internet/user-agent.ts
ST-DDT: IMO In this case not because you cannot do anything with it.

src/modules/internet/user-agent.ts
```diff
@@ -193,80 +176,113 @@ export function generate(faker: Faker): string {
...
+      }[browser]
+    );
 
-    return [browser, rnd(os[browser])];
-  }
+    return [browser, os];
+  };
 
-  function randomProc(arch: Arch): string | number {
+  const randomProc = (arch: OS): string => {
```
xDivisionByZerox, line 201: Does anyone know what `Proc` is supposed to mean here? I would like to have a more descriptive name here.

xDivisionByZerox:
I can't reproduce the error described in the failing pipeline. Anyone may help?

ST-DDT:
> I can't reproduce the error described in the failing pipeline. Anyone may help?

This is a relatively new error that only happens randomly.
I only managed to reproduce it once or twice locally, but I don't know what's causing this.
We could increase the timeout for that test, but I'm not sure whether that's the actual issue or whether the imports in there examples get resolved strangely and thus appear to hang for a "short" moment.
This might or might not be related to the long test start times that a few users have reported which we then recommended to use the locale specific faker instance.

I ran the test again and now it is working.

src/modules/internet/user-agent.ts
ST-DDT: This might refer to processor architectures or related OS "builds".

ST-DDT (dismissed):

xDivisionByZerox:
> This might or might not be related to the long test start times that a few users have reported which we then recommended to use the locale specific faker instance.

I see. Can this be due to all modules importing `Faker` from the `src/index.ts` file and creating a circular reference in the process (`ModuleX` => importing `Faker` from `index.ts` => index.ts import `ModuleX`)? This is just a guess, I have no idea how dependency resolution works in JS.

src/modules/internet/user-agent.ts
xDivisionByZerox: To move back to the original suggestion:

> I think we should move all of those functions to be statically defined

I will not do this in this PR, as I think I have already done more than the PR title would let suspect. I will gladly work on this in the future and will consider this suggestion then.

ST-DDT (approved):

xDivisionByZerox:
@ST-DDT the `accepted` label is described for accepted features or confirmed bugs. Not for chores 🤔

pkuczynski (approved):

Shinigami92 (approved):

PR #1180, opened 2022-07-22
chore: lint switch case blocks

Shinigami92:
This PR introduces the eslint rule [padding-line-between-statements](https://eslint.org/docs/latest/rules/padding-line-between-statements). This rule should apply to each case of a switch statement for now.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,14 +26,31 @@ module.exports = defineConfig({
   },
   plugins: ['@typescript-eslint', 'prettier'],
   rules: {
+    eqeqeq: ['error', 'always', { null: 'ignore' }],
     // We may want to use this in the future
     'no-useless-escape': 'off',
-    eqeqeq: ['error', 'always', { null: 'ignore' }],
+    'padding-line-between-statements': [
+      'error',
+      {
+        blankLine: 'never',
+        prev: ['case', 'default'],
+        next: '*',
+      },
+    ],
     'prefer-template': 'error',
 
     '@typescript-eslint/ban-ts-comment': 'warn',
     '@typescript-eslint/consistent-type-imports': 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'error',
+    '@typescript-eslint/naming-convention': [
+      'error',
+      {
+        format: ['PascalCase'],
+        selector: ['class', 'interface', 'typeAlias', 'enumMember'],
+        leadingUnderscore: 'forbid',
+        trailingUnderscore: 'forbid',
+      },
+    ],
     '@typescript-eslint/no-inferrable-types': [
       'error',
       { ignoreParameters: true },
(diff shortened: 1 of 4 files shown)


ST-DDT (commented):
Not sure about the style itself, but I like to do it consistent.

ST-DDT (approved):

.eslintrc.js
```diff
@@ -26,14 +26,31 @@ module.exports = defineConfig({
...
   plugins: ['@typescript-eslint', 'prettier'],
   rules: {
+    eqeqeq: ['error', 'always', { null: 'ignore' }],
     // We may want to use this in the future
     'no-useless-escape': 'off',
-    eqeqeq: ['error', 'always', { null: 'ignore' }],
+    'padding-line-between-statements': [
+      'error',
+      {
+        blankLine: 'never',
```
pkuczynski, line 35: I am against this change. I really like those blank lines. They nicely separate multiline cases.

.eslintrc.js
```diff
@@ -26,14 +26,31 @@ module.exports = defineConfig({
...
+        prev: ['case', 'default'],
```
pkuczynski, line 36: What does it do? Put default as first?

pkuczynski (changes requested):

.eslintrc.js
Shinigami92: There is a full documentation about this rule
Copied the content from there
Please read it

.eslintrc.js
Shinigami92: I would like to have full consistency over the whole code base, so never or always but not mixed

.eslintrc.js
ST-DDT: If you refer to something that others should read, please link it for simplicity sake.

Shinigami92: That is currently a bit hard from mobile
If really interested make a 1s google or wait for ~24h until I have a laptop again with internet

ST-DDT (changes requested):
The PR description is missing information about the source of the config and the available options aka why it is written/configured exactly like that.
Making it harder to understand the PR without looking at the changes and any external documentaion.

.eslintrc.js
xDivisionByZerox: Im sorry, but if it is THAT hard for you to provide additional information of the things you have don't or link do additional source, it it that hard for you to wait 24h with opening this PR?

xDivisionByZerox: To provide information to the original question:

This configuration requires a new line before each `case` and `default` keyword.

.eslintrc.js
pkuczynski: There are cases when new line is good, and there are when its not. Thats how I would like to see formatting:

```js
switch (value) {
  case 'A': return 'a'
  case 'B': return 'b'
}

switch (value) {
  case 'A':
    console.log(value)
    next = some + operation
    break;

  case 'B':
    console.log(value)
    next = some + operation
    break;
}
```

This is more readable when some spacing is added between larger blocks of code. I would not like to be forced to glue them together. So we either leave it as it is or you provide config which supports such cases.

.eslintrc.js
pkuczynski: I think this configuration prevents new line. And I know where its coming from: @hankucz did it with new lines in her PR and @Shinigami92 didnt like it...

Shinigami92:
Okay, sorry seems to be my fault that I thought it would be a good idea to stop discussions about code style with an automated rule. But seems you prefer inconsistency 🤷

Also it was my fault to assume it would be a good idea to open a PR within my vacation and try to help with that. But yeah I was very wrong.
I think the best for me and my mental health is to close this and take another week of silence off from faker. Bye 👋

pkuczynski:
@Shinigami92 don't take it like this! It's not fair, we all try to contribute, and everybody values your input a lot! Stay positive and let's talk about it :)

Do you see my point why it's sometimes ok to have line break in multiline case statements? Or you totally disagree?

PR #1181, opened 2022-07-22
fix: remove prohibited char `i` in bitcoinAddress generation

import-brain:
According to [the Bitcoin wiki](https://en.bitcoinwiki.org/wiki/Bitcoin_address), Bitcoin addresses "consist of random digits and uppercase and lowercase letters, with the exception that the uppercase letter "O", uppercase letter "I", lowercase letter "i", and the number "0" are never used to prevent visual ambiguity."

I noticed that when generating bitcoin addresses, `finance.bitcoinAddress()` excludes the characters `O`, `I`, and `0` successfully, but it does not exclude the lowercase character `i`.
https://github.com/faker-js/faker/blob/895799b23e8224eb1b623ef1d0ffd508f101389f/src/modules/finance/index.ts#L218-L221

I deleted the character `i` from the generation so that it doesn't show up in our bitcoin addresses.

### DIFF
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -217,7 +217,7 @@ export class Finance {
 
     for (let i = 0; i < addressLength - 1; i++)
       address += this.faker.helpers.arrayElement(
-        '123456789abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ'.split('')
+        '123456789abcdefghjkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ'.split('')
       );
 
     return address;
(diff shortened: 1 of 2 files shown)


pkuczynski (approved):

ST-DDT (changes requested):
The refernced wiki page contains an example that contains an `i`:

> Bitcoin address example:
>
> > 14qViLJfdGaP4EeHnDyJbEGQysnCpwk3gd

ST-DDT:
Maybe we should just match the given validation method:
https://github.com/validatorjs/validator.js/blob/master/src/lib/isBtcAddress.js

import-brain:
> The refernced wiki page contains an example that contains an `i`:
> 
> > Bitcoin address example:
> > > 14qViLJfdGaP4EeHnDyJbEGQysnCpwk3gd

I just reread the wiki, I completely misread that text section I referenced, Bitcoin addresses do allow the lowercase character `i`. Looking back over the generation code, the function already excludes all prohibited characters, my bad.

Issue #1111, opened 2022-06-27
Add support for https://placeholder.com as image provider


Shinigami92:
### Clear and concise description of the problem

Add support for https://placeholder.com

### Suggested solution

Create random generated image URLs based on deterministic and seed via https://placeholder.com

### Additional context

I want to create simple grey or coloured placeholder images for my company's project.
I will used them as thumbnails for a video. Also at least the aspect ratio is important to be controlled.

Devdre1909:
Hi @import-brain.
Can I work on this if anyone isn’t working on it?, and is there a proposed implementation?

import-brain:
> Hi @import-brain.
> 
> Can I work on this if anyone isn’t working on it?, and is there a proposed implementation? 

Sure! I don't believe there is a proposed implementation, but I'll assign this to you now.

@Shinigami92, just checking, this is open for contributions, right?

xDivisionByZerox:
This has been implemented in #1186.

PR #1186, opened 2022-07-24
feat(image): add image via.placeholder provider

Devdre1909:
Issue  #1111 
Add support for https://placeholder.com/ 

width, height, text, format, background color, and text color supported


src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image. Defaults to `640`.
+   * @param height The height of the image. Defaults to `width`.
+   * @param text The text of the image
+   * @param format The file format of the image. Supports `png` `jpeg` `png` `gif` `webp`
+   * @param backgroundColor The background color of the placeholder. Supports HEX CODE format
+   * @param textColor The text color of the placeholder. Requires `backgroundColor` Supports HEX CODE format
```
xDivisionByZerox, line 28: [Our JSDoc rules](https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md#jsdocs) mentions:
> Sentences should always end with a period.

So could you please end any sentence here with one?

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
...
+    format?: 'png' | 'jpeg' | 'jpg' | 'gif' | 'webp',
+    backgroundColor?: string,
+    textColor?: string
+  ): string {
+    width = width || 640;
+    height = height || width;
+
+    let url = 'https://via.placeholder.com';
+
+    url += `/${width}/${height}`;
```
xDivisionByZerox, line 43: These lines could be combined. Or did you have any particular thoughts on separating them?

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
+import type { Faker } from '../../..';
+
+/**
+ * Module to generate links to images on `https://via.placeholder.com/`.
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
+
+  /**
+   * Generates a new placeholder image url
```
xDivisionByZerox, line 10: Missing period.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * @param width The width of the image. Defaults to `640`.
+   * @param height The height of the image. Defaults to `480`.
+   * @param text The text of the image
```
xDivisionByZerox, line 14: Missing period.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Module to generate links to images on `https://via.placeholder.com/`.
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
+
+  /**
+   * Generates a new placeholder image url
+   *
+   * @param width The width of the image. Defaults to `640`.
+   * @param height The height of the image. Defaults to `480`.
```
import-brain, line 13:
```suggestion
   * @param width The width of the image (in pixels). Defaults to `640`.
   * @param height The height of the image (in pixels). Defaults to `480`.
```

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  image(width?: number, height?: number, text?: string): string {
+    return this.imageUrl(width, height, text);
+  }
+
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image. Defaults to `640`.
+   * @param height The height of the image. Defaults to `width`.
```
import-brain, line 24:
```suggestion
   * @param width The width of the image (in pixels). Defaults to `640`.
   * @param height The height of the image (in pixels). Defaults to `width`.
```

ejcheng (commented):
We should specify that `width` and `height` are in pixels.

src/modules/image/providers/placeholder.ts
Devdre1909: Okay, @xDivisionByZerox. Done.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Module to generate links to images on `https://via.placeholder.com/`.
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
+
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image (in pixels). Defaults to `640`.
+   * @param height The height of the image (in pixels). Defaults to `480`.
```
ST-DDT, line 13: The default is not the same as the one used in the called method.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @param backgroundColor The background color of the placeholder. Supports HEX CODE format.
+   * @param textColor The text color of the placeholder. Requires `backgroundColor` Supports HEX CODE format.
+   */
+  imageUrl(
+    width?: number,
+    height?: number,
+    text?: string,
+    format?: 'png' | 'jpeg' | 'jpg' | 'gif' | 'webp',
+    backgroundColor?: string,
+    textColor?: string
```
ST-DDT, line 36: Maybe we should use `options: {...}` here instead of this many parameters.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image (in pixels). Defaults to `640`.
+   * @param height The height of the image (in pixels). Defaults to `width`.
+   * @param text The text of the image.
+   */
+  image(width?: number, height?: number, text?: string): string {
+    return this.imageUrl(width, height, text);
+  }
```
ST-DDT, line 18: IMO this method does nothing extra to the `imageUrl` method and there is no other method that does use the `imageUrl` method, so maybe you can merge them into a single method?

ST-DDT: On the other hand it would be nice, if there was a way to generate an image with random colors and texts, so you can just call it and get unique texts.

test/image.spec.ts
```diff
@@ -237,6 +237,100 @@ describe('image', () => {
...
+          'https://via.placeholder.com/100/200/000000/FFFFFF.jpg?text=I+love+food'
+        );
+      });
+    });
+
+    describe('image()', () => {
+      it('should return a searching image url with text', () => {
+        const imageUrl = faker.image.placeholder.image(100, 200, 'I love food');
+        expect(imageUrl).toBe(
+          'https://via.placeholder.com/100/200?text=I+love+food'
```
ST-DDT, line 328: This image is not `100x200` it is `100x100` and uses the color `200`.

ST-DDT (changes requested):

src/modules/image/providers/placeholder.ts
Devdre1909: Okay, I thought of this but I also thought it better to follow conversion from the other modules. I will use object instead.

xDivisionByZerox: I would suggest to keep it with this many arguments for now so all image helper are consistent. At least for this PR. In the future we should refactor to a single optuons argument.

Devdre1909: Okay

src/modules/image/providers/placeholder.ts
Devdre1909: Hi, @ST-DDT I added this method `randomUrl` in my late update but I'm not sure how to write the test for it since most of the values are generated at random except the width and height. Any suggestions, please?

test/image.spec.ts
Devdre1909: This is resolved since the method has been removed

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (format != null) {
+      url += `.${format}`;
+    }
+
+    const text = this.faker.lorem.text();
+    const urlParam = new URLSearchParams({ text });
+    url += `?${urlParam.toString()}`;
+
+    return url;
```
xDivisionByZerox, line 87: Why can't wrong about:
```ts
return this.imageUrl(
  width,
  height,
  this.faker.lorem.text(),
  format,
  this.faker.color.rgb({
    casing: 'upper',
    prefix: '',
  }),
  this.faker.color.rgb({ casing: 'upper', prefix: '' }),
);
```
?

Devdre1909: Smart, didn't think of this.
I will implement this now.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
+
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image (in pixels). Defaults to `640`.
+   * @param height The height of the image (in pixels). Defaults to `width`.
+   * @param text The text of the image.
```
ST-DDT, line 14: Maybe add a hint what is shown if no/an empty text is selected

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
...
+   * @param format The file format of the image. Supports `png` `jpeg` `png` `gif` `webp`.
+   * @param backgroundColor The background color of the placeholder. Supports HEX CODE format.
+   * @param textColor The text color of the placeholder. Requires `backgroundColor` Supports HEX CODE format.
```
ST-DDT, line 17: Should we add a hint regarding the default color, if nothing is specified?

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
...
+    let url = 'https://via.placeholder.com';
+    url += `/${width}/${height}`;
+
+    if (backgroundColor != null) {
+      url += `/${backgroundColor.replace('#', '').toUpperCase()}`;
+    }
+
+    if (textColor != null && backgroundColor !== null) {
+      url += `/${textColor.replace('#', '').toUpperCase()}`;
+    }
```
ST-DDT, line 39: IMO if you write it like this, then it's easier to understand:

```suggestion
    if (backgroundColor != null) {
      url += `/${backgroundColor.replace('#', '').toUpperCase()}`;

      if (textColor != null) {
        url += `/${textColor.replace('#', '').toUpperCase()}`;
      }
    }
```

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
...
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
+
+  /**
+   * Generates a new placeholder image url.
+   *
+   * @param width The width of the image (in pixels). Defaults to `640`.
+   * @param height The height of the image (in pixels). Defaults to `width`.
+   * @param text The text of the image.
+   * @param format The file format of the image. Supports `png` `jpeg` `png` `gif` `webp`.
```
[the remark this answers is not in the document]
ST-DDT: Maybe add a hint regarding the default format?

ST-DDT (changes requested):

src/modules/image/providers/placeholder.ts
ST-DDT: Maybe split on `/` and check whether the respective parts match some simple criteria. It doesn't have to be a fancy test.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
+import type { Faker } from '../../..';
+
+/**
+ * Module to generate links to images on `https://via.placeholder.com/`.
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {}
```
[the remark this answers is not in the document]
Devdre1909: I'm not sure what I'm supposed to do on this.

src/modules/image/providers/placeholder.ts
Devdre1909: Should I include an example like this?
image

src/modules/image/providers/placeholder.ts
Devdre1909: There is no default color placeholder.com handles the background and text color is none is specified.

src/modules/image/providers/placeholder.ts
ST-DDT: All methods should have an example section.

src/modules/image/providers/placeholder.ts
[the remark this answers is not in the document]
ST-DDT:
```suggestion
  constructor(private readonly faker: Faker) {
    // Bind `this` so namespaced is working correctly
    for (const name of Object.getOwnPropertyNames(_Date.prototype)) {
      if (name === 'constructor' || typeof this[name] !== 'function') {
        continue;
      }
      this[name] = this[name].bind(this);
    }
  }
```

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,77 @@
...
+    height?: number,
+    text?: string,
+    format?: 'png' | 'jpeg' | 'jpg' | 'gif' | 'webp',
+    backgroundColor?: string,
+    textColor?: string
+  ): string {
+    width = width || 640;
+    height = height || width;
+
+    let url = 'https://via.placeholder.com';
```
Minozzzi, line 30: Does it make sense to have a class-scoped variable as a constant instead this? For me this url is immutable

src/modules/image/providers/placeholder.ts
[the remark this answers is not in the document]
Devdre1909: Hi, @ST-DDT I committed this change, but I don't think it's right. Other image modules didn't do this and there is no `_Date`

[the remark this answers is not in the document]
xDivisionByZerox: You have `_Date` as prototype class. I guess that's an error?

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,85 @@
+import type { Faker } from '../../..';
+
+/**
+ * Module to generate links to images on `https://via.placeholder.com/`.
+ */
+export class Placeholder {
+  constructor(private readonly faker: Faker) {
+    // Bind `this` so namespaced is working correctly
+    for (const name of Object.getOwnPropertyNames(_Date.prototype)) {
```
ST-DDT, line 9:
```suggestion
    for (const name of Object.getOwnPropertyNames(Placeholder.prototype)) {
```

src/modules/image/providers/placeholder.ts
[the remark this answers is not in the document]
ST-DDT: Yeah, sorry thats my bad. I copied it from the date module.
I replaced it with Placeholder (the current class).

ST-DDT:
Please update the tests so the dont fail anymore.

src/modules/image/providers/placeholder.ts
ST-DDT: But it is only used in a single location, so there is no real benefit there.
Currently the other modules also don't specify a constant for that, so for consistency reasons we can stick to the current variant.

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  randomUrl(
+    width?: number,
+    height?: number,
+    format?: 'png' | 'jpeg' | 'jpg' | 'gif' | 'webp'
+  ): string {
+    return this.imageUrl(
+      width,
+      height,
+      this.faker.lorem.text(),
```
ST-DDT, line 86: This can generate very long phrases:
https://via.placeholder.com/640x640/DA03F2/B62C6C?text=Error+iusto+necessitatibus+amet+voluptatem.+Dolorum+officia+ratione+ad+a.+Expedita+accusamus+earum+rerum+vel+repellendus+ea.+Ipsum+perspiciatis+sed+dolorum.+Non+consequatur+beatae+et+et+aut+incidunt+impedit.%0AVeritatis+exercitationem+accusamus+laudantium+soluta.+Illo+nostrum+debitis+tempore.+Provident+sit+quaerat+facere+eius+officiis+quidem+aut+vel.+Velit+quaerat+magni.+Ea+optio+illum+porro+natus+quia+corporis.%0AUt+ut+inventore+consequatur+est+quisquam.+Maiores+odit+autem+ut+quis+amet+aut+fuga.+Numquam+vel+temporibus+sed+atque

Maybe a single word is enough?

src/modules/image/providers/placeholder.ts
```diff
@@ -0,0 +1,95 @@
...
+    return url;
+  }
+
+  /**
+   * Generate a new placeholder image with random colors and text.
+   *
+   * @param width The width of the image (in pixels). Defaults to `640`.
+   * @param height The height of the image (in pixels). Defaults to `width`.
+   * @param format The file format of the image. Supports `png` `jpeg` `png` `gif` `webp`.
+   */
```
ST-DDT, line 77: Please add `@examples`.

ST-DDT:
@Devdre1909 May I go ahead and fix these things to get this merged or do you wish to do this yourselves?

Devdre1909:
> @Devdre1909 May I go ahead and fix these things to get this merged or do you wish to do this yourselves?

I will.
Thank you.

ST-DDT (approved):

Shinigami92 (approved):

lk77:
> Issue #1111 Add support for https://placeholder.com/
> 
> width, height, text, format, background color, and text color supported

Hi @Devdre1909 ,

it seems to me that site does not exist anymore

matthewmayer:
See https://github.com/faker-js/faker/issues/2477

While the homepage of the site now promotes am unrelated business the images appear to continue to work for now.

carusoray:
It seems that the via.placeholder.com images API is not longer active. placeholder.com is still alive; however, via.placeholder.com does not resolve. Image URL within the test JSON are no longer valid.

ST-DDT:
Thanks for the report. We have deprecated it:

- https://github.com/faker-js/faker/pull/3341

The deprecation will be released in a few days.

- https://github.com/faker-js/faker/pull/3362

As a workaround, you can use:

https://github.com/faker-js/faker/blob/bddce72b499fc0b62fc527720e6804ca72884643/src/modules/image/index.ts#L307

PR #1196, opened 2022-07-26
test: remove JSON.stringify wrapper around faker.seed() in tests descriptions

pkuczynski:
I am not sure why we had those `JSON.stringify` @Shinigami92?


xDivisionByZerox:
They were wrapped due to `faker.seed` being able to return an array of numbers if one is used as input value. But we don't do that so this should be fine.

xDivisionByZerox (approved):

ST-DDT (approved):
I think it's okay.

pkuczynski:
> They were wrapped due to `faker.seed` being able to return an array of numbers if one is used as input value. But we don't do that so this should be fine.

1. `seed(seed?: number): number;`
2. `console.log(`${[1, 2, 3]}`) // '1,2,3'

:)

xDivisionByZerox:
> console.log(`${[1, 2, 3]}`) // '1,2,3'

And that's why `JSON.stringif` was used for arrays. Because this output this sh*t. I mean it works but doesn't provide clarity. You could also have written ```console.log(`${seed1}, ${seed2}, ${seed3}`)```.
One the other side, JSON gives me a nicely formatted array `console.log(JSON.stringify([1,2,3])) // [1,2,3]`.

But it doesn't matter since we know now from the signature overload if the return value will be an array or not and can act accordingly.

pkuczynski:
From the signature we know it's never an array :)

xDivisionByZerox:
https://github.com/faker-js/faker/blob/5ea8252f727e2e577c2adca9650ac8f24a171632/src/faker.ts#L242-L244

It literally has `number` **OR** `number[]` as return type.

pkuczynski:
Indeed! I haven't notice the overload because of jsdoc, sorry :)

import-brain (approved):

PR #1218, opened 2022-08-01
fix(lorem.word): return word with maximum possible length when expected length is too large

### LINKED ISSUE #1131: `faker.lorem.word` can return `undefined` despite being typed as `string`
### Describe the bug

When calling `faker.lorem.word` with a length that doesn't have any matching words, it will return `undefined` while TypeScript thinks it's a `string`.

### Reproduction

```ts
const nonexistentWord = faker.lorem.word(9999) // <-- typed as `string`
console.log(nonexistentWord) // <-- will output `undefined`
```

### Additional Info

_No response_

hankucz:
Fixes #1131

### DIFF
--- a/src/modules/lorem/index.ts
+++ b/src/modules/lorem/index.ts
@@ -18,20 +18,36 @@ export class Lorem {
    * Generates a word of a specified length.
    *
    * @param length length of the word that should be returned. Defaults to a random length.
+   * If a word with the specified length does not exist return a word with maximum possible length.
    *
    * @example
    * faker.lorem.word() // 'temporibus'
    * faker.lorem.word(5) // 'velit'
+   * faker.lorem.word(9999) // 'necessitatibus'
    */
   word(length?: number): string {
-    const hasRightLength = (word: string) => word.length === length;
     let properLengthWords: readonly string[];
+
     if (length == null) {
       properLengthWords = this.faker.definitions.lorem.words;
     } else {
-      properLengthWords =
-        this.faker.definitions.lorem.words.filter(hasRightLength);
+      let maxLength = 0;
+
+      properLengthWords = this.faker.definitions.lorem.words.filter((word) => {
+        if (word.length > maxLength) {
+          maxLength = word.length;
+        }
+
+        return word.length === length;
+      });
+
+      if (properLengthWords.length === 0) {
+        properLengthWords = this.faker.definitions.lorem.words.filter(
+          (word) => word.length === maxLength
+        );
+      }
     }
+
     return this.faker.helpers.arrayElement(properLengthWords);
   }
 
--- a/test/lorem.spec.ts
+++ b/test/lorem.spec.ts
@@ -38,8 +38,15 @@ describe('lorem', () => {
           expect(faker.definitions.lorem.words).toContain(actual);
         });
 
-        // INFO @Shinigami92 2022-02-11: Seems there are only words with a max length of 14 characters
-        it.each(times(14))(
+        const maxLength = faker.definitions.lorem.words.reduce((len, word) => {
+          if (word.length > len) {
+            len = word.length;
+          }
+
+          return len;
+        }, 0);
+
+        it.each(times(maxLength))(
           'should return random value from word array with a max length of %i characters',
           (length) => {
             const actual = faker.lorem.word(length);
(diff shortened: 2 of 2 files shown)


pkuczynski (approved):

src/modules/lorem/index.ts
```diff
@@ -18,20 +18,36 @@ export class Lorem {
...
-        this.faker.definitions.lorem.words.filter(hasRightLength);
+      let maxLength = 0;
+
+      properLengthWords = this.faker.definitions.lorem.words.filter((word) => {
+        if (word.length > maxLength) {
+          maxLength = word.length;
+        }
+
+        return word.length === length;
+      });
```
Shinigami92, line 42: Beside that this is very costy, you already used `reduce` in test file, but why not here?

src/modules/lorem/index.ts
```diff
@@ -18,20 +18,36 @@ export class Lorem {
    * Generates a word of a specified length.
    *
    * @param length length of the word that should be returned. Defaults to a random length.
+   * If a word with the specified length does not exist return a word with maximum possible length.
```
ST-DDT, line 21: This does not fit my expectations regarding this parameter.

If the length of the word doesn't matter for me, then I wouldn't specify it.
The very fact, that I specify a specific length expresses intend that I actually want a word of that particular length.

If I just want a very long word, I dig into the data pick a suitable length and use that (Or maybe use a range parameter to narrow it down).

So this method should throw an error, if it cannot satisfy the requested data or should have an optional param/option that allows for a "closest fit". Not necessarily the longest. e.g. requesting a word with length 1 should not return `Rindfleischetikettierungsüberwachungsaufgabenübertragungsgesetz` (which is the name of an actual German law until 2013) according to https://de.babbel.com/de/magazine/das-laengste-wort.

ST-DDT (changes requested):

src/modules/lorem/index.ts
hankucz: Filtering was already part of the implementation before with `word.length === length`. So this is no change. Adding calculating max would not increase the cost of filtering much. So don't think this is a major problem.

src/modules/lorem/index.ts
hankucz: This is not for general words, this is for `lorem`, which is the same for all languages (although it does change the alphabet for some, like jp).

I can change implementation to throw an error, but when I had a brief look, we don't do that in other places. For example, `faker.word.*` would return a word of any length, when length can not be satisfied: https://github.com/faker-js/faker/blob/main/src/modules/word/index.ts#L11

So if we want to stay consistent, that's fine for me. I can change it in any other place. The question is: should faker be strict or be as close to the expectation as possible?

src/modules/lorem/index.ts
Shinigami92: Previously you ran only once through all words, now you do it twice

hankucz: It will only run a second time when the length is not satisfied. I don't see any other way of implementing this unless we want to throw, as @ST-DDT suggests, but I don't think it's a good idea.

@Shinigami92, any other suggestions welcome...

src/modules/lorem/index.ts
```diff
@@ -18,20 +18,36 @@ export class Lorem {
...
+      properLengthWords = this.faker.definitions.lorem.words.filter((word) => {
+        if (word.length > maxLength) {
+          maxLength = word.length;
+        }
+
+        return word.length === length;
+      });
+
+      if (properLengthWords.length === 0) {
+        properLengthWords = this.faker.definitions.lorem.words.filter(
```
Shinigami92, line 45: Something looks very odd here 🤔 
How can this even work in TypeScript?
`properLengthWords` is declared as `readonly` but assigned twice here (L36 and L45)

pkuczynski: `readonly` is only on the values, and not on the variable. So reassigning different arrays without modifying later content, would work.

src/modules/lorem/index.ts
Shinigami92: I now looked a bit deeper into the implementation and tried to think about what your proposed implementation does.
And I'm more and more on @ST-DDT side (comment above here) that this current proposed PR is not what I, @ST-DDT or a user would expect.
We might need to discuss this in a team meeting, but I would currently reject your proposed solution, sorry 🙁

Shinigami92 (changes requested):

src/modules/lorem/index.ts
pkuczynski: Yeah, let's discuss this at the meeting, as different functions behave differently. As @hankucz pointed out, `faker.word.*` would ignore length if it can not be satisfied. So if we want to throw an error, we would need to change the behavior of those other methods too, to stay consistent.

ST-DDT:
See here for our team decision: https://github.com/faker-js/faker/issues/1131#issuecomment-1205405540

ST-DDT:
This PR is stale and does not match our team decision.
Closing for now so someone else can tackle this.

PR #1224, opened 2022-08-03
docs: fixed inconsitency in @see usage in jsdoc

hankucz:


src/definitions/finance.ts
```diff
@@ -13,7 +13,7 @@ export type FinanceDefinitions = LocaleEntry<{
    * The pattern by (lowercase) issuer name used to generate credit card codes.
    * `L` will be replaced by the check bit.
    *
-   * @see Helpers.replaceCreditCardSymbols()
```
import-brain, line 16: Not too sure about this change, @Shinigami92, is there any special reason for it to be `Helpers` and not `faker.helpers`? I don't think so, but just making sure. Other than this, this PR LGTM

ejcheng (approved):

src/definitions/finance.ts
ST-DDT: I like this change since it makes it consistent.

ST-DDT (commented):
I'm not sure whether all methods need the empty `()`.

src/definitions/finance.ts
import-brain: I like it too, just making sure there isn't any reason why it would be `Helpers`.

Shinigami92: No I don't know why `Helpers` was written here, change looks legit to me

Shinigami92:
> I'm not sure whether all methods need the empty `()`.

How did we do it in the rest of the codebase and if it's mixed please let us decide one, document that and use it consistently

src/definitions/finance.ts
import-brain: > No I don't know why `Helpers` was written here, change looks legit to me

Ah okay, thanks Shini

ST-DDT:
I will write a test after this PR is merged.

ST-DDT (approved):

PR #1225, opened 2022-08-03
feat(locale): add directions and directions abbr to pl

hankucz:


src/locales/pl/address/direction.ts
```diff
@@ -0,0 +1,10 @@
+export default [
+  'Północ',
```
pkuczynski, line 2: * add source: https://pl.wikipedia.org/wiki/Strony_świata
* sort alphabetically
* use lower case

import-brain (approved):
Translations verified using Google Translate, lgtm

src/locales/pl/address/direction_abbr.ts
```diff
@@ -0,0 +1 @@
+export default ['N', 'E', 'S', 'W', 'NE', 'NW', 'SE', 'SW'];
```
pkuczynski, line 1: There are polish abbreviations: pn., pd., ...

See: https://pl.wikipedia.org/wiki/Strony_świata

pkuczynski (changes requested):

src/locales/pl/address/direction_abbr.ts
hankucz: Done...

pkuczynski (dismissed):

src/locales/pl/address/direction.ts
hankucz: Done...

import-brain (dismissed):

ST-DDT (changes requested):
The order of the elements is somewhat important:
https://github.com/faker-js/faker/blob/04c9d36cc44fa647f401919c708e550dbe42040d/src/modules/address/index.ts#L411-L453
0-3 => cardinal direction
4-7 => ordinal direction

pkuczynski:
Wow! I would never expect this and it's really hard to catch on the review. It would be way better if TS could catch it automatically...

Should we change the definition to look more like:

```js
interface Directions {
  cardinal: string[]
  ordinal: string[] // or intercardinal ?
  secondaryIntercardinal?: string[]
}

type DirectionType = keyof Directions
```

And then both `directions` and `directions_abbr` should use this type?

Additionally, I think it makes sense to simplify the usage and change `direction` signature to:

```
direction ({ abbr: boolean, type: DirectionType, types: DirectionType[]})

@examples
faker.address.direction() // 'north'
faker.address.direction({ abbr: true }) // 'N'
faker.address.direction({ type: 'ordinal' }) // 'north west'
faker.address.direction({ abbr: true, type: 'ordinal' }) // 'NW'
faker.address.direction({ abbr: true, types: ['ordinal', 'secondaryIntercardinal']  }) // 'NNW'
```

And then deprecate the other methods? We might also skip `secondaryIntercardinal`...

ST-DDT:
@pkuczynski Please create a separate issue for that.

pkuczynski:
> @pkuczynski Please create a separate issue for that.

Sure! https://github.com/faker-js/faker/issues/1228

ST-DDT:
@hankucz Could you please restore the original order? (N,E,S,W, NE, SE, SW, SE)
That way we can merge this and can easily automatically transform it once we change the implementation.

hankucz:
@ST-DDT original order restored

pkuczynski (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #1226, opened 2022-08-03
locale(pl): add extra colors and sort lines

hankucz:


import-brain (approved):

pkuczynski (approved):

ST-DDT (approved):

ST-DDT:
Should we use `locale(pl): add extra colors and sort lines` as a way to name these locale only PRs?
That way it would be easier to spot in the changelogs.

Shinigami92:
> Should we use `locale(pl): add extra colors and sort lines` as a way to name these locale only PRs? That way it would be easier to spot in the changelogs.

It's needed to be allowed here: https://github.com/faker-js/faker/blob/main/.github/semantic.yml

PR #1227, opened 2022-08-03
feat(locale): add horses to pl

hankucz:


src/locales/pl/animal/horse.ts
```diff
@@ -0,0 +1,242 @@
+export default [
```
pkuczynski, line 1: It would be good to add a source, where this information comes from...

pkuczynski: And the list should be sorted alphabetically...

pkuczynski (changes requested):

ST-DDT:
@hankucz Could you please perform the requested changes so we can merge this?

Shinigami92:
@xDivisionByZerox did you progress on a PR for checking alphabetically sorted locales?
I think we should test it and add a manual filtering step to define if we want to list some specific cases where sorting doesn't make sense (like direction)

src/locales/pl/animal/horse.ts
hankucz: Done

hankucz:
@ST-DDT feedback applied

pkuczynski (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #1230, opened 2022-08-04
ci: allow all commit keywords in PR titles and add locale keyword

ST-DDT:
Currently we cannot use `locale` commit keywords, however some of our PRs contain that.
We even have a tag for that.
It would be nice if we could highlight these PRs in changelogs/release notes, so the user can quickly either search for them or mentally discard them.

Recommended usage: `locale(pl): add/update xyz`


Shinigami92:
We need to investigate why only these types are allowed, maybe the configuration in our project is wrong?
image

pkuczynski (dismissed):

pkuczynski:
Great idea!

ST-DDT:
We can only use `workflow` after this PR is merged.

Shinigami92:
> We can only use `workflow` after this PR is merged.

Why that? Did you see my comment above?

ST-DDT:
It is part of our commit keywords and after this PR is merged, it is also part of the config.

Shinigami92:
But this PR is about `locale` and not `workflow`
I assume the whole config is not taken at all! And it falls back to the default config.

ST-DDT:
I renamed the title. Is it better now?

Shinigami92:
> I renamed the title. Is it better now?

Oh nonono 😱, I highly assume you totally misunderstood me
I can explain you via voice chat tomorrow if you like to

Shinigami92 (approved):
Oh fucking hell whats up with my GitHub?!?!?!! There are cache issues! You already changed the files 👀 
I needed to retrigger several times a hard reload and now I see that you already commited it, I'm sorry

import-brain (approved):

PR #1231, opened 2022-08-04
test: validate @see contents

ST-DDT:
Checks that all `@see faker` method references to contain `(...)`.

**Invalid:**  `@see faker.helpers.arrayElement`
**Valid:**  `@see faker.helpers.arrayElement()`

Note: I'm not sure whether empty `()` are the correct way to have. Especially, if the method requires parameters.


Shinigami92 (approved):
For the start I think this is good enough, but it might be that we want to enhance the check in the future if we encounter some weird tag content ^^

Like `@see faker is (a really cool) tool`

pkuczynski (approved):

pkuczynski (approved):

import-brain (approved):

PR #1253, opened 2022-08-10
docs: searchable api

Shinigami92:
This is a draft to make the API searchable like https://vuejs.org/api/

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -85,6 +85,11 @@ export default defineConfig({
     nav: [
       { text: 'Guide', link: '/guide/' },
       {
+        text: 'API',
+        activeMatch: `^/api/`,
+        link: '/api/',
+      },
+      {
         text: 'Ecosystem',
         items: [{ text: 'StackBlitz ', link: 'https://fakerjs.dev/new' }],
       },
--- a/docs/api/ApiIndex.vue
+++ b/docs/api/ApiIndex.vue
@@ -51,7 +51,7 @@ function slugify(text: string): string {
       .replace(/\-{2,}/g, '-')
       // Remove prefixing and trailing separators
       .replace(/^\-+|\-+$/g, '')
-      // ensure it doesn't start with a number (#121)
+      // ensure it doesn't start with a number (like #123)
       .replace(/^(\d)/, '_$1')
     // // lowercase
     // .toLowerCase()
--- a/docs/api/ApiIndex.vue
+++ b/docs/api/ApiIndex.vue
@@ -1,0 +1,247 @@
+<script setup lang="ts">
+import apiIndex from './api-data.json';
+import { computed, ref } from 'vue';
+
+interface APIHeader {
+  anchor: string;
+  text: string;
+}
+
+interface APIGroup {
+  text: string;
+  items: {
+    text: string;
+    link: string;
+    headers: APIHeader[];
+  }[];
+}
+
+const query = ref('');
+const normalize = (s: string) => s.toLowerCase().replace(/-/g, ' ');
+
+const filtered = computed(() => {
+  const q = normalize(query.value);
+  const matches = (text: string) => normalize(text).includes(q);
+
+  return apiIndex
+    .map((section) => {
+      // section title match
+      if (matches(section.text)) {
+        return section;
+      }
+
+      // filter groups
+      const matchedGroups = section.items
+        .map((item) => {
+          // group title match
+          if (matches(item.text)) {
+            return item;
+          }
+          // filter headers
+          const matchedHeaders = item.headers.filter(
+            ({ text, anchor }) => matches(text) || matches(anchor)
+          );
+          return matchedHeaders.length
+            ? { text: item.text, link: item.link, headers: matchedHeaders }
+            : null;
+        })
+        .filter((i) => i);
+
+      return matchedGroups.length
+        ? { text: section.text, items: matchedGroups }
+        : null;
+    })
+    .filter((i) => i) as APIGroup[];
+});
+
+// same as vitepress' slugify logic
+function slugify(text: string): string {
+  return (
+    text
+      // Replace special characters
+      .replace(/[\s~`!@#$%^&*()\-_+=[\]{}|\\;:"'<>,.?/]+/g, '-')
+      // Remove continuous separators
+      .replace(/\-{2,}/g, '-')
+      // Remove prefixing and trailing separators
+      .replace(/^\-+|\-+$/g, '')
+      // ensure it doesn't start with a number (#121)
+      .replace(/^(\d)/, '_$1')
+    // // lowercase
+    // .toLowerCase()
+  );
+}
+</script>
+
+<template>
+  <div id="api-index">
+    <div class="header">
+      <h1>API Reference</h1>
+      <div class="api-filter">
+        <label for="api-filter">Filter</label>
+        <input
+          type="search"
+          placeholder="Enter keyword"
+          id="api-filter"
+          v-model="query"
+        />
+      </div>
+    </div>
+
+    <div v-for="section of filtered" :key="section.text" class="api-section">
+      <h2 :id="slugify(section.text)">{{ section.text }}</h2>
+      <div class="api-groups">
+        <div v-for="item of section.items" :key="item.text" class="api-group">
+          <h3>{{ item.text }}</h3>
+          <ul>
+            <li v-for="h of item.headers" :key="h.anchor">
+              <a :href="item.link + '.html#' + slugify(h.anchor)">{{
+                h.anchor
+              }}</a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    </div>
+
+    <div v-if="!filtered.length" class="no-match">
+      No API matching "{{ query }}" found.
+    </div>
+  </div>
+</template>
+
+<style scoped>
+#api-index {
+  max-width: 1024px;
+  margin: 0px auto;
+  padding: 64px 32px;
+}
+
+h1,
+h2,
+h3 {
+  font-weight: 600;
+  line-height: 1;
+}
+
+h1,
+h2 {
+  letter-spacing: -0.02em;
+}
+
+h1 {
+  font-size: 38px;
+}
+
+h2 {
+  font-size: 24px;
+  color: var(--vt-c-text-1);
+  margin: 36px 0;
+  transition: color 0.5s;
+  padding-top: 36px;
+  border-top: 1px solid var(--vt-c-divider-light);
+}
+
+h3 {
+  letter-spacing: -0.01em;
+  color: var(--vt-c-green);
+  font-size: 18px;
+  margin-bottom: 1em;
+  transition: color 0.5s;
+}
+
+.api-section {
+  margin-bottom: 64px;
+}
+
+.api-groups a {
+  font-size: 15px;
+  font-weight: 500;
+  line-height: 2;
+  color: var(--vt-c-text-code);
+  transition: color 0.5s;
+}
+
+.dark api-groups a {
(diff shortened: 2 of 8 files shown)


docs/api/ApiIndex.vue
```diff
@@ -0,0 +1,247 @@
...
+      // Replace special characters
+      .replace(/[\s~`!@#$%^&*()\-_+=[\]{}|\\;:"'<>,.?/]+/g, '-')
+      // Remove continuous separators
+      .replace(/\-{2,}/g, '-')
+      // Remove prefixing and trailing separators
+      .replace(/^\-+|\-+$/g, '')
+      // ensure it doesn't start with a number (#121)
+      .replace(/^(\d)/, '_$1')
+    // // lowercase
+    // .toLowerCase()
```
Shinigami92, line 70: The Vue docs had all values in lowercase, but we use anchors like `buildingNumber`

docs/api/ApiIndex.vue
```diff
@@ -0,0 +1,247 @@
...
+    </div>
+
+    <div v-if="!filtered.length" class="no-match">
+      No API matching "{{ query }}" found.
+    </div>
+  </div>
+</template>
+
+<style scoped>
+#api-index {
```
Shinigami92, line 113: The styles are currently just copy pasted but don't work like in vuejs docs
Maybe we are missing global styles or we just need to write them on our own

docs/api/api-data.json
```diff
@@ -0,0 +1,23 @@
+[
```
Shinigami92, line 1: This file need to be auto-generated

docs/api/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: API Reference
+sidebar: false
+page: true
+footer: false
```
Shinigami92, line 5: somehow the "suggest edit" still gets rendered? don't know how to disable this 🤷

docs/api/api-data.json
```diff
@@ -0,0 +1,23 @@
...
+  {
+    "text": "Module API",
```
Shinigami92, line 3: Not sure if we want multiple sections like Module API and e.g. General Manual API or other details
Like we could combine stuff like call `seed()` and so on

docs/.vitepress/config.ts
```diff
@@ -6,6 +7,57 @@ const description =
   'Generate massive amounts of fake (but reasonable) data for testing and development.';
 const image = 'https://fakerjs.dev/social-image.png';
 
+export const sidebar: DefaultTheme.Sidebar = {
```
Shinigami92, line 10: This could be reverted, this was just required by the vuejs docs example but we might generate our own data and not based on the sidebar
(https://github.com/vuejs/docs/blob/ebab2325755f3c6e29cf0b13b32f7d2936593cc6/src/api/api.data.ts#L5)

docs/api/index.md
ST-DDT: I think it is called `edit`.

docs/api/api-data.json
ST-DDT: Note this could reuse the generated api docs.ts files.

Shinigami92: There is no such file
No `api/docs.ts` or `api-docs.ts` or whatever
I found a `api/typedoc.json`, but unsure if I really want to process this dynamically on client
I also found multiple `api/<module>.ts` files, but not sure if I want to dynamically collect them all together and process them dynamically on client

docs/api/index.md
Shinigami92: Sadly doesn't work 🙁

ST-DDT: https://github.com/faker-js/faker/blob/8cb6027087fbc3fd038c4063b78f283d9aa48959/scripts/apidoc/apiDocsWriter.ts#L18

docs/api/api-data.json
ST-DDT: Maybe we should decide what to search for. E.g. search in the name only or include the description, tags, parameters, ...

Shinigami92:
<details>
<summary>Current State</summary>

image
</details>

ST-DDT:
Amazing! Do we want to search for only the names or do we also want to search for some (hidden) aliases/tags?

Shinigami92:
> Amazing!

thx :3

> Do we want to search for only the names or do we also want to search for some (hidden) aliases/tags?

Not sure, I would start with simple solution
But I think I need to normalize some stuff (like _Date => Date)

ST-DDT:
I think we already have the normalization somewhere for the api docs themselves.

https://github.com/faker-js/faker/blob/e5951c05055e511532d098df403c7d4066b37cf3/scripts/apidoc/moduleMethods.ts#L37

Maybe we should extract it to a separate method (normalizeModuleName).

For method names we have that to a certain extend already:

https://github.com/faker-js/faker/blob/e5951c05055e511532d098df403c7d4066b37cf3/scripts/apidoc/signature.ts#L30

Shinigami92:
Okay, I think I need to generate at build-time a data json for the search page, because I would like to prevent the need of recreating the search data object everytime someone visit the search page on their client

scripts/apidoc/apiDocsWriter.ts
```diff
@@ -1,6 +1,9 @@
 import { writeFileSync } from 'node:fs';
 import { resolve } from 'node:path';
+import type { ProjectReflection } from 'typedoc';
+import { ReflectionKind } from 'typedoc';
 import type { Method } from '../../docs/.vitepress/components/api-docs/method';
+import type { APIGroup } from '../../docs/api/api-types';
```
ST-DDT, line 6: This import seems to be missing from the commit.

scripts/apidoc/apiDocsWriter.ts
```diff
@@ -133,3 +141,31 @@ export function writeApiPagesIndex(pages: PageIndex): void {
...
+  moduleApiSection.items = modules.map((module) => ({
+    text: module.name,
+    link: module.name.toLowerCase(),
+    headers: module.getChildrenByKind(ReflectionKind.Method).map((child) => ({
+      anchor: child.name,
+      text: child.name,
+    })),
+  }));
+
+  writeFileSync(pathDocsApiSearchIndex, JSON.stringify(apiIndex));
```
ST-DDT, line 170: The direct methods are missing (fake, unique).

ST-DDT (changes requested):
Blocked by https://github.com/faker-js/faker/pull/1272 and then it needs to be rebased to use the new methods so we have consistent api docs and search index.

scripts/apidoc/apiDocsWriter.ts
Shinigami92: Might have done it correct, please reevaluate, but we might need to meet together and refactor some parts to e.g. extract helper functions and simplify stuff
But lets do that when we find time in discord

ST-DDT (changes requested):

scripts/apidoc/apiDocsWriter.ts
ST-DDT: Looks good to me.
As for the refactoring/helper methods, I'm looking forward to it.

docs/api/ApiIndex.vue
```diff
@@ -0,0 +1,238 @@
...
+function slugify(text: string): string {
+  return (
+    text
+      // Replace special characters
+      .replace(/[\s~`!@#$%^&*()\-_+=[\]{}|\\;:"'<>,.?/]+/g, '-')
+      // Remove continuous separators
+      .replace(/\-{2,}/g, '-')
+      // Remove prefixing and trailing separators
+      .replace(/^\-+|\-+$/g, '')
+      // ensure it doesn't start with a number (#121)
```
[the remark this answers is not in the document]
Shinigami92: That number was generated by GitHub copilot 😆

docs/api/api-types.ts
```diff
@@ -0,0 +1,15 @@
...
+  text: string;
+}
+
+export interface APIItem {
+  text: string;
+  link: string;
+  headers: APIHeader[];
+}
+
+export interface APIGroup {
```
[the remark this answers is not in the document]
Shinigami92: Yeah, maybe even rename them to our own use, they were originally copy&pasted from vuejs docs
https://github.com/vuejs/docs/blob/f54aab387042460e8e24c1ff9a82d5fe8ac6b656/src/api/api.data.ts#L7-L19

docs/api/ApiIndex.vue
```diff
@@ -0,0 +1,238 @@
...
+            ? { text: item.text, link: item.link, headers: matchedHeaders }
+            : null;
+        })
+        .filter((i) => i);
+
+      return matchedGroups.length
+        ? { text: section.text, items: matchedGroups }
+        : null;
+    })
+    .filter((i) => i) as APIGroup[];
```
import-brain, line 41: Not too sure if I'm not following here, but what is this line doing? Wouldn't it just return the original array?

ST-DDT: It omits null/undefined entries from the above mappimg method.

ST-DDT: Maybe add an explicit null check to reduce the mental complexity?
I also had to look add the code what happened.

docs/api/api-types.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface APIItem {
+  text: string;
+  link: string;
+  headers: APIHeader[];
+}
+
+/**
+ * Represents a whole section within the API search index page.
+ */
+export interface APIGroup {
```
ST-DDT, line 21: Currently we only have one of these, what others do you think we will have?

docs/api/ApiIndex.vue
Shinigami92: Yeah, it's the same as calling `if (i) { /**/ }`
Should we then also send a PR to VueJS's docs?
https://github.com/vuejs/docs/blob/f9d52a3a46408ed1ee5b1df520bc6b6f29bf5bd7/src/api/ApiIndex.vue#L41

docs/api/api-types.ts
Shinigami92: I didn't thought about any of that, it's 99% a copy-paste of VueJS docs
These interfaces are just containers that describes the schema of the data
If you like, I can rename them to `Section`, `SubSection` and `Element` or something like that, but keep in mind that they are totally not important and just describe the data schema

ST-DDT: I dont care about the name at this point.
We have one group. I dont see us ever having more than one group. So why should we have that part of the datastructure?
Or its important for you to the same style/structure as VueJs?

Shinigami92: It is someone important for me to align the structure to what VueJS is doing, as for example they could theoretically implement a component in the future to allow such a search index page, like what they have done with the Team page

Also it makes it much more easier, as if they update their code and we somehow get the info that they updated something, we can just copy paste their code, so we don't need to care that much

ST-DDT: Could you then please add some links in our sources that link to their code counterparts so that we can easier upgrade our code even without their notice?

docs/api/ApiIndex.vue
import-brain: > It omits null/undefined entries from the above mappimg method.

Ah okay, thanks for the clarification.

Shinigami92:
<details>
<summary>Current State</summary>

image
</details>

Anything else to get that merged? Or could we maybe enhance it in later PRs if needed and serve now the doc-user UX?

ST-DDT (approved):

ejcheng (approved):

pkuczynski (approved):

Issue #1250, opened 2022-08-09
Use wrapped functions instead of classes for modules


Shinigami92:
### Clear and concise description of the problem

Currently we need to bind `this` in the constructor of each module to support stuff like
```ts
const color = faker.color
color.rgb()
// and
const rgb = faker.color.rgb
rgb()
```

### Suggested solution

Instead of using js classes we could instead use wrapped functions like e.g. composables are working in Vue / VueUse

### Additional context

This might be a breaking change as with that you would not create a new class via `new Faker(...)` anymore
Also we need to benchmark if there would be performance regressions or benefits :thinking: 

I would like to setup a PR if this proposal gets accepted by the team

ST-DDT:
**Team Decision**

We will think about this again when we actually start the tree shaking.

ST-DDT:
I wrote a proposal for a design for this one in #2667 (because it is quite long)

- #2667

ST-DDT:
Superseded by #2667

- #2667

PR #1254, opened 2022-08-10
feat!: use wrapped functions

Shinigami92:
Draft for #1250 

... it could be that this is only an internal soft breaking change :thinking: 

---

So this PR has two benefits:

1. We get rid of the constructor hack
```ts
    for (const name of Object.getOwnPropertyNames(<Module>.prototype)) {
        continue;
      }
    }
  }
```
2. We are save in the future for misusing `this` and bindings of `this`

New downsides for now: it gets a little bit more complex for new contributors to add a new function
as they would need to declare the TypeScript definition in the module's interface and implement it in the module wrapped-fucntion.


src/modules/git/index.ts
```diff
@@ -62,23 +29,79 @@ export class Git {
...
+    '9',
+    'a',
+    'b',
+    'c',
+    'd',
+    'e',
+    'f',
+  ];
+
+  function branch(): string {
```
ST-DDT, line 82: Is there a way to avoid these warnings?

Shinigami92: For sure, these should also be done in this PR
We just need to reconfigure out config

Minozzzi:
I like the idea of applying the interface to the `git` export function.
Something like `git export function(faker: Faker): Git`. Therefore, an error will be displayed when there are not all methods.
I know it is already being applied to `faker.ts`. But I think it would be more of a security in development.
What do you think about it?

xDivisionByZerox:
I don't like this approach at all. TBH I don't see the benefits you described when looking at the extra declaration work. 

Furthermore, what do you change at all with this approach? Isn't that the same thing that would happen for a js class under the hood? 

I think using classes is perfectly fine for us and so is using the "constructor hack", IMO. I guess the real problem is with our modularization strategy.

Shinigami92:
> I don't like this approach at all. TBH I don't see the benefits you described when looking at the extra declaration work.
> 
> Furthermore, what do you change at all with this approach? Isn't that the same thing that would happen for a js class under the hood?
> 
> I think using classes is perfectly fine for us and so is using the "constructor hack", IMO. I guess the real problem is with our modularization strategy.

Yeah, you raise immediately all the questions I answered in the last meeting :smile: 

So first: re-read the PRs description text (I updated it a bit)

1. the constructor hack is removed
2. we can now use "private"/"not exported" functions inside modules
3. this is theoretically a pre-phase towards moving completely to only exported functions
so in the (no so near) future we would like to have something like this:

```
src/
  modules/
    git/
      index.ts
      branch.ts
      commitEntry.ts
      <...>.ts
```

where index.ts is only:

```ts
import { branch } from './branch'
import { commitEntry } from './commitEntry'

export interface Git {
  ...
}

export const git: Git = {
  branch,
  commitEntry,
  ...
}

export {
  branch,
  commitEntry,
  ...
}
```

and the functions would be extracted to their specific files

The benefit of that would be that we then make use of tree-shaking

BUT: we need to find a way of how we pass the faker instance around and reusing the same seed sequence (not solved yet)
@ST-DDT proposed a solution using factories for the functions, but we didn't tested that out

Any further questions right now?

xDivisionByZerox:
> 3. this is theoretically a pre-phase towards moving completely to only exported functions
so in the (no so near) future we would like to have something like this:

Ok, this is exactly what I had in mind. All methods are exported by themself and as a collection/module from the index.ts file. You guys work just like me. 

I still don't see why this step is required now, but it doesn't introduce any breaking changes, so it should be fine.

xDivisionByZerox:
>export {
  branch,
  commitEntry,
  ...
}

I would not suggest exporting specific functions from the `index.ts` file since that can lead to side-effects and could potentially break tree-shaking.

Just wanted to leave that here.

xDivisionByZerox:
> 2. we can now use "private"/"not exported" functions inside modules

We can do this already. Not private, but I don't see a problem with the approach we have so far.

import-brain:
> New downsides for now: it gets a little bit more complex for new contributors to add a new function
> as they would need to declare the TypeScript definition in the module's interface and implement it in the module wrapped-fucntion.

If we do go this route, we'll definitely need documentation on how to implement a function in the new system.

Shinigami92:
> If we do go this route, we'll definitely need documentation on how to implement a function in the new system.

Absolutely, but as I now dig into it and start experimenting/implementing/working with it, I start to feel more and more like @xDivisionByZerox 
We only solve something real with it if we reach tree-shaking support (not bundled together modules that the user can use)
otherwise it's only a "Disimprovement" ("Verschlimmbesserung") for the source code

Shinigami92:
Closing these experiments for now

ST-DDT:
@Shinigami92 Do you still need the branch? If not, please delete it.

Commit-ID: 41db9e5727e1f5d275e887ee796219cbea8d1003

Shinigami92:
Yes, please do not delete them right now so we have a copy and look into it if needed
I just wanted to clean the PR section as we would need to rewrite them anyway as they got to stale

PR #1264, opened 2022-08-12
ci: verify semantic pull requests

Shinigami92:
In our last meeting we discussed that we want to allow only specific defined scopes so the [CHANGELOG.md](https://github.com/faker-js/faker/blob/main/CHANGELOG.md) will get better updated automatically

This PR will add a guideline that explains which are good commit/PR titles and which are not (with historical real world examples and how to write them instead)

Also it will only allow a subset of `types` and `scopes` so the PR title must follow the new convention

Please note that the `verifyCommit` script doesn't need to be absolutely exact, as only the PR title is important at the end when merging the PR because we always use **squash-merge rebase**

Tip: You can check how we wrote commit messages in the history with following command:
```node
git log --pretty=format:%s > commits.md
```


.github/workflows/semantic-pull-request.yml
```diff
@@ -20,17 +20,43 @@ jobs:
...
+            mersenne
+            music
+            name
+            phone
+            random
+            science
+            system
+            unique
+            vehicle
+            word
```
ST-DDT, line 62: I don't think we should have a limited set of scopes here, because then PRs for new modules would be impossible.
(require a pre-change to the pipeline)

.github/workflows/semantic-pull-request.yml
```diff
@@ -20,17 +20,43 @@ jobs:
           types: |
             feat
             fix
-            docs
-            dx
+            chore
             refactor
```
ST-DDT, line 24: IMO this is confusing, because it won't show up in the changelogs, but in our case it is kind of important.

.github/workflows/semantic-pull-request.yml
```diff
@@ -20,17 +20,43 @@ jobs:
...
             test
-            locale
-            workflow
-            build
             ci
-            chore
-            types
-            wip
+            build
+            infra
```
ST-DDT, line 29: IMO we should decide on one of `ci`, `build` or `infra`

ST-DDT (changes requested):

ST-DDT:
> We need to discuss if we want a type `revert` or a scope `revert` (or both)?

IMO we should allow type revert. Although this might be connected to a bugfix usually.

.github/workflows/semantic-pull-request.yml
Shinigami92: Yes and no, I already had the same thought
But maybe if we introduce/remove a new/old module, this list just needs to be updated in the exact same PR
We as maintainers just need to keep care of that and tell the contributor
I think introducing a new module is relatively rarely (like the color module)

.github/workflows/semantic-pull-request.yml
Shinigami92: IMO they all have their meanings
- ci: touches e.g. github workflow yaml files, ...
- build: touches build scrips, changing node target, ...
- infra: update issue template, ...

.github/workflows/semantic-pull-request.yml
ST-DDT: Tested by renaming this PR. It uses the args from main.

CONTRIBUTING.md
```diff
@@ -225,3 +225,73 @@ pnpm run docs:serve # Serve docs from /dist
...
+> Please note that the PR title should not include a suffix of e.g. `(#123)` as this will be done automatically by GitHub while merging
+
+Some examples of valid pull request titles:
+
+```shell
+feat: add casing option
+fix: lower target to support Webpack 4
+docs: remove unused playground
+ci: allow breaking change commits
+test: validate @see contents
```
ST-DDT, line 258: Please add an example for deprecations somewhere. Plus every supported type.

.github/workflows/semantic-pull-request.yml
Shinigami92: That's odd 😕 
So we need to somehow guide contributors to good PR titles to move maintenance away from us
Maybe with a github PR template

CONTRIBUTING.md
```diff
@@ -225,3 +225,73 @@ pnpm run docs:serve # Serve docs from /dist
...
+
+# A release PR will look like this
+chore(release): 7.4.0
+
+# Renovate automatically generates these
+chore(deps): update devdependencies
+chore(deps): update typescript-eslint to ~5.33.0
+```
+
+Historically real world examples that could be written in a better way:
```
import-brain, line 272:
```suggestion
Previous pull request titles that could have been written in a better way:
```

CONTRIBUTING.md
```diff
@@ -225,3 +225,73 @@ pnpm run docs:serve # Serve docs from /dist
 ## Deploying documentation
 
 See the [netlify.toml](netlify.toml) for configuration.
+
+## Committing
+
+Pull Request titles need to follow our semantic convention.
+
+The configured `types` and `scopes` can be found in [.github/workflows/semantic-pull-request.yml](https://github.com/faker-js/faker/blob/main/.github/workflows/semantic-pull-request.yml).
```
xDivisionByZerox, line 233: I think conventionally `'type'` is named `'topic'` but I might be wrong. Nevertheless I prefer `'topic'`.

.github/workflows/semantic-pull-request.yml
Shinigami92: I have an idea!
If it needs to be just in the main branch before merging the PR, we could create an additional `ci` PR that just adds the new module name

CONTRIBUTING.md
Shinigami92: please re-evaluate and make a new review suggestion if needed

CONTRIBUTING.md
Shinigami92: https://www.conventionalcommits.org/en/v1.0.0/#summary:~:text=structured%20as%20follows%3A-,%3Ctype%3E,-%5Boptional%20scope%5D%3A%20%3Cdescription

Nope

.github/workflows/semantic-pull-request.yml
ST-DDT: The issue with that is that the user sees a ci failure but doesnt know how to fix it himself.
Also PRs with red build failure markers are less likely to be reviewed.
If we add the marker asap, zhen we might end up with months of unused scopes which might lead to confusion as well. If the new module PR gets stale/dies then we have a dead scope in our list that we have to remember to remove.

CONTRIBUTING.md
```diff
@@ -225,3 +225,78 @@ pnpm run docs:serve # Serve docs from /dist
...
+- feat: `datatype.hexadecimal` signature change
++ feat(datatype): hexadecimal signature change
+  datatype is one of our modules and can be used as scope
+
+- feat(image): add image via.placeholder provider
++ feat(image): add via.placeholder provider
+  image was redundant in the subject
+
+- feat(system.networkInterface): add networkInterface faker
++ feat(system): add method networkInterface
```
ST-DDT, line 289:
```suggestion
+ feat(system): add networkInterface method
```

ST-DDT (changes requested):

.github/workflows/semantic-pull-request.yml
Shinigami92: Yeah I see the problem :confused: 
Okay so I need to move these allowed scopes to a documentation / pr template and tell the @faker-js/members to help us fix wrongly named PR titles

Shinigami92:
Cannot name it infra yet, as it's not in the main branch :sweat_smile:

CONTRIBUTING.md
```diff
@@ -225,3 +225,93 @@ pnpm run docs:serve # Serve docs from /dist
 ## Deploying documentation
 
 See the [netlify.toml](netlify.toml) for configuration.
+
+## Committing
+
+Pull Request titles need to follow our semantic convention.
+
+The configured `types` and `scopes` can be found in [.github/workflows/semantic-pull-request.yml](https://github.com/faker-js/faker/blob/main/.github/workflows/semantic-pull-request.yml).
```
ST-DDT, line 233: The scopes aren't listed there anymore.
So I would move this to the **types** section.

scripts/verifyCommit.ts
```diff
@@ -12,7 +12,7 @@ const msg = readFileSync(msgPath, 'utf-8').trim();
 
 const releaseRE = /^v\d/;
 const commitRE =
-  /^(revert: )?(feat|fix|docs|dx|refactor|perf|test|locale|workflow|build|ci|chore|types|wip|release|deps)(\(.+\))?\!?: .{1,50}/;
+  /^(revert: )?(feat|fix|docs|dx|refactor|perf|test|locale|workflow|build|ci|chore|types|wip|release|deps|infra)(\(.+\))?\!?: .{1,50}/;
```
ST-DDT, line 15: IMO we should remove the commit names that aren't valid PR titles, because they will never be in main.

ST-DDT (changes requested):

scripts/verifyCommit.ts
Shinigami92: We could also think about to completely remove the verifyCommit script as we have the Semantic Pull Request action and only the PR title is important
the commits inside the PR can be written absolutely chaotically it doesn't matter

ST-DDT (changes requested):
Please also add an example commit message for a changed locale file.

scripts/verifyCommit.ts
ST-DDT: Theoretically yes, but I use the commit names as PR titles and sometimes I missspell them or forget about the prefix then the check reminds me to add them. Also I hope that it makes the commits easier to review because you can easier find a particular change if the commits follow a certain naming scheme.
I once worked in a project where someone used `.` as commit messages and it was really hard to check the history for relevant changes. It doesn't really matter if you only have a single commit, but the more it gets, the more important it becomes.

ST-DDT (dismissed):

.github/pull_request_template.md
```diff
@@ -0,0 +1,3 @@
+<!-- Please first read https://github.com/faker-js/faker/blob/verify-semantic-pull-requests/CONTRIBUTING.md -->
+
+<!-- Help us by directly write a correct PR title following this guide: https://github.com/faker-js/faker/blob/verify-semantic-pull-requests/CONTRIBUTING.md#committing -->
```
import-brain, line 3:
```suggestion
<!-- Help us by writing a correct PR title following this guide: https://github.com/faker-js/faker/blob/verify-semantic-pull-requests/CONTRIBUTING.md#committing -->
```

ST-DDT (approved):

ejcheng (approved):

PR #1266, opened 2022-08-12
feat!: another experiment

Shinigami92:
Alternative draft for #1250

This is another solution and would be what was initially planned as second step after #1254


Minozzzi:
I liked this approach, in my opinion keeps the code separate and organized.

Shinigami92:
Closing these experiments for now

ST-DDT:
@Shinigami92 Do you still need the branch? If not, please delete it.

Commit-ID: c66ddf2659da215f069c7e22592a9d18dcd85f3c

Shinigami92:
Yes, please do not delete them right now so we have a copy and look into it if needed
I just wanted to clean the PR section as we would need to rewrite them anyway as they got to stale

PR #1270, opened 2022-08-13
docs: properly handle multiple `@see` tags

ST-DDT:
Fixes the support for multiple `@see` tags.

# Single

## Previously

grafik

## Now

grafik

# Multiple

## Previously

grafik

## Now

grafik


docs/.vitepress/components/api-docs/method.vue
```diff
@@ -32,12 +32,14 @@ function seeAlsoToUrl(see: string): string {
...
       <h3>See Also</h3>
-      <div v-for="seeAlso of props.method.seeAlsos" :key="seeAlso">
-        <a :href="seeAlsoToUrl(seeAlso)" v-if="seeAlso.startsWith('faker.')">
-          <p>{{ seeAlso }}</p>
-        </a>
-        <p v-else>{{ seeAlso }}</p>
-      </div>
+      <ul>
+        <li v-for="seeAlso of props.method.seeAlsos" :key="seeAlso">
+          <a :href="seeAlsoToUrl(seeAlso)" v-if="seeAlso.startsWith('faker.')">
```
Shinigami92, line 37:
```suggestion
          <a v-if="seeAlso.startsWith('faker.')" :href="seeAlsoToUrl(seeAlso)">
```

This just help is readability, because v-if and v-else is more easily scan able by eyes underneath

scripts/apidoc/utils.ts
```diff
@@ -105,11 +106,29 @@ export function extractRawExamples(signature?: SignatureReflection): string[] {
...
+export function extractSeeAlsos(signature?: SignatureReflection): string[] {
+  return extractTagContent('@see', signature, (tag) => {
+    const content = tag.content;
+    if (content.length === 1) {
+      return joinTagContent(tag);
+    } else {
+      return tag.content
+        .filter((_, index) => index % 3 === 1) // ['-', 'content', '\n']
+        .map((part) => part.text);
+    }
```
Shinigami92, line 123: You can omit this `else` and just return as you have a return in your `if`

```suggestion
    }
    return tag.content
      .filter((_, index) => index % 3 === 1) // ['-', 'content', '\n']
      .map((part) => part.text);
```

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -113,7 +114,7 @@ describe('examples and deprecations', () => {
       }
 
       // Verify @see tag
-      extractTagContent('@see', signature).forEach((link) => {
+      extractSeeAlsos(signature).forEach((link) => {
         if (link.startsWith('faker')) {
```
Shinigami92, line 118: Above I read something of `faker.` instead of `faker` without dot :thinking: 

Maybe this makes the check a bit more safe? :shrug:

ST-DDT:
```suggestion
        if (link.startsWith('faker.')) {
```

Shinigami92 (dismissed):

scripts/apidoc/utils.ts
```diff
@@ -105,11 +106,28 @@ export function extractRawExamples(signature?: SignatureReflection): string[] {
   );
 }
 
+/**
+ * Extracts the all see also references from the jsdocs separately.
```
xDivisionByZerox, line 110: This sentence doesn't seem right to me. Did you wanted to write this?
```suggestion
 * Extracts all the see also references from the jsdocs separately.
```

But I don't know why the "also" is needed (for the function name as well).

scripts/apidoc/utils.ts
```diff
@@ -105,11 +106,28 @@ export function extractRawExamples(signature?: SignatureReflection): string[] {
...
+      .map((part) => part.text);
+  });
+}
+
 /**
  * Joins the parts of the given jsdocs tag.
  */
-export function joinTagContent(tag: CommentTag): string {
-  return joinTagParts(tag?.content);
+export function joinTagContent(tag: CommentTag): string[] {
```
xDivisionByZerox, line 129: It seems weird to me that a function that is literally named `join*` returns an array. 
Is this ONLY used in line 118? If so, what is the value of extracting this "logic" in a separate function?

Shinigami92: I thought the same, but thought it's not so important :shrug: 
But now were someone else raised the exact same... yeah, that should be handled differently

scripts/apidoc/utils.ts
Shinigami92:
```suggestion
 * Extracts all the `@see` references from the jsdocs separately.
```

scripts/apidoc/utils.ts
xDivisionByZerox: > I thought the same, but thought it's not so important 🤷

Code review in a nutshell, lol.

ST-DDT: Any recommendation?

ST-DDT: > Is this ONLY used in line 118?

https://github.com/faker-js/faker/blob/32223cd259de01fa56b289a9a4d5f7ce2de2bc9a/scripts/apidoc/utils.ts#L93

Shinigami92 (approved):
The name of the function is not that important to hold back this PR
I approve

scripts/apidoc/utils.ts
Shinigami92: resolve for now, to move on

pkuczynski (approved):

PR #1279, opened 2022-08-18
refactor(locale): move company.name formats to locales

Minozzzi:


src/locales/en/company/index.ts
```diff
@@ -8,7 +8,7 @@ import bs_adjective from './bs_adjective';
 import bs_noun from './bs_noun';
 import bs_verb from './bs_verb';
 import descriptor from './descriptor';
-import name from './name';
+import name_ from './name';
```
Minozzzi, line 11: After running `pnpm run generate:locales` name became name_. Why can't it be `name`?

Shinigami92: It might be historical, but VSCode hat conflicts with import / exports of constants named `name` and marked it as automatically deprecated 
Also `type` has problems and confuses TypeScript so we use `type_`

src/modules/company/index.ts
```diff
@@ -35,20 +35,15 @@ export class Company {
...
     if (format != null) {
       deprecated({
         deprecated: 'faker.company.name(format)',
-        proposed: 'faker.company.name()',
+        proposed: 'faker.company.name() or faker.fake(format)',
         since: '7.4',
         until: '8.0',
       });
     }
+    const formats = this.faker.definitions.company.name;
```
Shinigami92, line 46: This is a full breaking change for Faker instances not using en as fallback ⚠️

src/definitions/company.ts
```diff
@@ -29,6 +29,11 @@ export type CompanyDefinitions = LocaleEntry<{
    */
   descriptor: string[];
 
+  /**
+   * Company names.
```
ST-DDT, line 33: These aren't names. These are patterns to generate names.

ST-DDT (changes requested):

src/modules/company/index.ts
Minozzzi: This makes sense. What I need to do?

Shinigami92: We could leave the static patterns and only override them with provided definitions if they are defined

In next major we can then fully remove it (could fit to one of our next planned milestones)

src/definitions/company.ts
```diff
@@ -29,6 +29,11 @@ export type CompanyDefinitions = LocaleEntry<{
    */
   descriptor: string[];
 
+  /**
+   * A list of patterns used to generate company names.
+   */
+  name: string[];
```
Shinigami92, line 35: Should we rename this to `name_patterns`?

Minozzzi: I agree with this change, I will rename it

src/modules/company/index.ts
ST-DDT: I wouldn't do that.
If you don't want any breakage wait for the next major release. 
If a little bit is okay, keep it like this. There is no point in a temp fallback if you have a breaking change.

src/definitions/company.ts
ST-DDT: Please also rename the existing locale files.

src/modules/company/index.ts
Minozzzi: I agree with @Shinigami92. We could keep the static patterns and only override them if it has patterns in the current locale.
Could we going to this direction?

ST-DDT: IMO the and in the pattern makes it explicitly English. Currently English ist our fallback language anyway so it would be set. If someone explicitly removes the English fallback, then they dont want them, so it use it as a fallback here either.
Note, we are planning to automatically throw an error in that case in the future.

Minozzzi: I understand, so in this case are we going to just merge it into the next major version?

ST-DDT: I think there is no need to wait.
Its like adding a new locale data for some locales, I dont consider it breaking at all.

Shinigami92: @ST-DDT so we don't do **semver** anymore?

ST-DDT (dismissed):

src/modules/company/index.ts
```diff
@@ -35,20 +35,15 @@ export class Company {
...
-      '{{name.lastName}} {{company.companySuffix}}',
-      '{{name.lastName}} - {{name.lastName}}',
-      '{{name.lastName}}, {{name.lastName}} and {{name.lastName}}',
-    ];
-
     if (format != null) {
       deprecated({
         deprecated: 'faker.company.name(format)',
-        proposed: 'faker.company.name()',
+        proposed: 'faker.company.name() or faker.fake(format)',
```
Shinigami92, line 41:
```suggestion
        proposed: 'faker.company.name() or faker.helpers.fake(format)',
```

src/modules/company/index.ts
ST-DDT: Why do you consider this a breaking change?

Shinigami92 (changes requested):

src/modules/company/index.ts
Shinigami92: https://github.com/faker-js/faker/pull/1279#discussion_r948644676

ST-DDT: I think we are running in circles now.

@Minozzzi so lets just do us more work and use the existing format array as fallback for now.
Please also add a TODO comment there that the fallback should be removed eventually.
After the PR is merged we will then create an issue for it, so it doesn't get lost.

src/modules/company/index.ts
Minozzzi: Thanks!

src/modules/company/index.ts
```diff
@@ -35,21 +35,24 @@ export class Company {
...
       deprecated({
         deprecated: 'faker.company.name(format)',
-        proposed: 'faker.company.name()',
+        proposed: 'faker.company.name() or faker.helpers.fake(format)',
         since: '7.4',
         until: '8.0',
       });
     }
 
+    // TODO: This `staticFormats` pattern should be removed in the future. It is only used to maintain backwards compatibility.
```
ST-DDT, line 47: TODO: Create an issue for this on merge.

ST-DDT (dismissed):

src/modules/company/index.ts
Minozzzi: Can I create the issue?

Minozzzi:
Guys, I've problems with rebase. I'will close this PR and open another

PR #1283, opened 2022-08-18
feat(locale): add city names (de)

atheck:
Added city names for local `de`.


src/locales/de/address/city_name.ts
```diff
@@ -0,0 +1,202 @@
...
+  'Neuss',
+  'Neustadt an der Weinstraße',
+  'Neuwied',
+  'Norderstedt',
+  'Nordhorn',
+  'Nürnberg',
+  'Oberhausen',
+  'Offenbach am Main',
+  'Offenburg',
+  'Oldenburg (Oldenburg)|Oldenburg',
```
ST-DDT, line 147: This one looks odd to me.

ST-DDT:
Do you have a source for these (in case we need to need to update it)?

Shinigami92:
Please use the new PR title naming convention: https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md#committing

src/locales/de/address/city_name.ts
atheck: That's right. I'll fix that.

atheck:
I took the first 200 cities from here: https://de.wikipedia.org/wiki/Liste_der_Gro%C3%9F-_und_Mittelst%C3%A4dte_in_Deutschland

Shinigami92 (approved):

pkuczynski (approved):

PR #1289, opened 2022-08-19
feat(name): add sexType method

Shinigami92:
Requested by @ST-DDT via https://github.com/faker-js/faker/issues/776#issuecomment-1220384286

This function is just a helper for users that want to use faker instead of plain TS

### DIFF
--- a/src/modules/name/index.ts
+++ b/src/modules/name/index.ts
@@ -285,6 +285,16 @@ export class Name {
   }
 
   /**
+   * Returns a random sex type.
+   *
+   * @example
+   * faker.name.sexType() // Sex.Female
+   */
+  sexType(): SexType {
+    return this.faker.helpers.arrayElement(Object.values(Sex));
+  }
+
+  /**
    * Returns a random name prefix.
    *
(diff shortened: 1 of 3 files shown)


src/modules/name/index.ts
```diff
@@ -284,6 +284,16 @@ export class Name {
...
   }
 
+  /**
+   * Returns a random sex type.
+   *
+   * @example
+   * faker.name.sexType() // Sex.Female
+   */
+  sexType(): SexType {
+    return this.faker.helpers.arrayElement(Object.values(Sex));
```
ST-DDT, line 294: maybe use: https://fakerjs.dev/api/helpers.html#objectvalue

test/name.spec.ts
```diff
@@ -20,6 +20,7 @@ describe('name', () => {
       'gender',
       'prefix',
       'sex',
+      'sexType',
```
ST-DDT, line 23: This doesn't take any arguments as the test tries to check.

ST-DDT (changes requested):

ST-DDT (approved):

import-brain (approved):

Issue #1179, opened 2022-07-22
Localize format patterns for company.name


ST-DDT:
### Clear and concise description of the problem

https://github.com/faker-js/faker/blob/fbddaa39e830e1ac44e7821d219ca7859af79e95/src/modules/company/index.ts#L37-L41

Currently the patterns are hardcoded to match English company names.

### Suggested solution

The patterns should be localized so other locales can define custom patterns.

### Additional context

There seems to be patterns/files for that already, but these needs to be checked/defined and referenced:
https://github.com/faker-js/faker/blob/fbddaa39e830e1ac44e7821d219ca7859af79e95/src/locales/en/company/name.ts#L1-L5
but there is nothing for that in: https://github.com/faker-js/faker/blob/fbddaa39e830e1ac44e7821d219ca7859af79e95/src/definitions/company.ts#L6

There are also other PRs that work on the method so this should only be done after both:

- https://github.com/faker-js/faker/pull/1166
- https://github.com/faker-js/faker/issues/1178

ST-DDT:
@Shinigami92 Do you consider this a breaking change?

Minozzzi:
Hello guys. I woud like contribute with this task.Can you guys assign to me?

ST-DDT:
There is https://github.com/faker-js/faker/pull/1216 already.

Shinigami92:
> There is #1216 already.

But it looks like that it is for the `name` module but this issue is about `company.name` 🤔 is it?

ST-DDT:
Oh yeah, you are right.

Minozzzi:
Alright @ST-DDT . I will try to contribute as best I can to this issue. 😀

PR #1293, opened 2022-08-22
feat(company): move name formats to locales

Minozzzi:
The oldest PR: #1279 
Fixes : #1179


src/modules/company/index.ts
```diff
@@ -35,21 +35,24 @@ export class Company {
...
       deprecated({
         deprecated: 'faker.company.name(format)',
-        proposed: 'faker.company.name()',
+        proposed: 'faker.company.name() or faker.helpers.fake(format)',
         since: '7.4',
         until: '8.0',
       });
     }
 
+    // ToDo: This `staticFormats` pattern should be removed in the future. It is only used to maintain backwards compatibility.
```
Minozzzi, line 47: ToDo: Create an issue for this on merge.

src/modules/company/index.ts
```diff
@@ -35,21 +35,24 @@ export class Company {
...
+    const staticFormats = [
+      '{{name.lastName}} {{company.companySuffix}}',
+      '{{name.lastName}} - {{name.lastName}}',
+      '{{name.lastName}}, {{name.lastName}} and {{name.lastName}}',
+    ];
+    const localeFormats = this.faker.definitions.company.name_patterns;
+    const formats = localeFormats ? localeFormats : staticFormats;
```
Shinigami92, line 54: This can be simplified to 

```suggestion
    const formats = this.faker.definitions.company.name_patterns ?? staticFormats;
```

src/modules/company/index.ts
Shinigami92, line 47:
```suggestion
    // TODO @Minozzzi 2022-08-22: This `staticFormats` pattern should be removed in the future. It is only used to maintain backwards compatibility.
```

Or you should just remove the whole todo comment and just create the issue after the merge 🤔

Shinigami92 (changes requested):

src/modules/company/index.ts
Minozzzi: It might be. I wasn't sure about the return when they don't have locales. But if it is `undefined` or `null` the `??` operator works fine. But if the return is `false` from locales the `??` operator doesn't work correctly.

Shinigami92: It will never be `false` as it's not typed as boolean, and it would fail then later on anyways, because we use it as array

And even then you could use `||` instead of `??` (but not needed here)

src/modules/company/index.ts
```diff
@@ -50,8 +50,7 @@ export class Company {
       '{{name.lastName}} - {{name.lastName}}',
       '{{name.lastName}}, {{name.lastName}} and {{name.lastName}}',
     ];
-    const localeFormats = this.faker.definitions.company.name_patterns;
-    const formats = localeFormats ? localeFormats : staticFormats;
+    const formats = this.faker.definitions.company.name_patterns ?? staticFormats;
```
Shinigami92, line 53: You need to run `pnpm run format` now ^^
The GitHub apply suggestion doesn't do that for us :/

Minozzzi: :/

Shinigami92 (approved):
changing the todo is not so important for me, I will approve now
other maintainers can decide if we want already merge or not

Shinigami92:
do we need to mark this as `feat:`? so it shows up in changelog?

ST-DDT (approved):

ST-DDT:
@Shinigami92 Do you still consider this breaking?

Shinigami92:
> We don't report locale changes either. We could rename this one though.

I thought we swapped explicitly to `feat(locale):` exactly for this reason?

> @Shinigami92 Do you still consider this breaking?

Not anymore as there is now the fallback, the fallback can then safely removed in v8

src/modules/company/index.ts
Minozzzi: Not all locales have patterns.

Minozzzi:
> Not anymore as there is now the fallback, the fallback can then safely removed in v8

@Shinigami92 If we create an issue for this by adding patterns for all locales. Could it be removed before v8?

Shinigami92:
> > Not anymore as there is now the fallback, the fallback can then safely removed in v8
> 
> @Shinigami92 If we create an issue for this by adding patterns for all locales. Could it be removed before v8?

yes

src/modules/company/index.ts
ST-DDT: Issue created: https://github.com/faker-js/faker/issues/1294

PR #1298, opened 2022-08-24
refactor(unique): move to helpers

xDivisionByZerox:
This PR moves the `unique()` method from the `Unique` module into the `Helpers` module. There is no reason for it being a standalone module when it is clearly a utility method.

This was discussed in https://github.com/faker-js/faker/discussions/805 ([direct link](https://github.com/faker-js/faker/discussions/805#discussioncomment-2539551)).


src/modules/unique/index.ts
```diff
@@ -1,11 +1,14 @@
...
-import * as uniqueExec from './unique';
 
 /**
  * Module to generate unique entries.
+ *
+ * @deprecated
  */
 export class Unique {
-  constructor() {
+  constructor(private readonly faker: Faker) {
```
xDivisionByZerox, line 11: Does this change make the entire PR "breaking"?

Shinigami92: We could be in luck here, as we do not yet export the Module constructors on their own

src/modules/helpers/index.ts
```diff
@@ -1,5 +1,7 @@
 import type { Faker } from '../..';
 import { FakerError } from '../../errors/faker-error';
+import type { RecordKey } from '../unique/unique';
+import * as uniqueExec from '../unique/unique';
```
ST-DDT, line 4: IMO these should be moved the helpers module (folder) too.

src/modules/unique/index.ts
```diff
@@ -52,13 +59,12 @@ export class Unique {
...
-    const { maxTime = 50, maxRetries = 50 } = options;
-    return uniqueExec.exec(method, args, {
-      ...options,
-      startTime: new Date().getTime(),
-      maxTime,
-      maxRetries,
-      currentIterations: 0,
+    deprecated({
+      deprecated: 'faker.fake()',
+      proposed: 'faker.helpers.fake()',
```
Shinigami92, line 64: `fake`?! what? I thought this is `unique` 😆

Shinigami92 (changes requested):

src/modules/unique/index.ts
xDivisionByZerox: Done. Was copying the deprecation from `fake` for convenience reasons^^

Shinigami92 (approved):

ST-DDT (approved):

PR #1302, opened 2022-08-25
infra: set renovate schedule to weekly

xDivisionByZerox:
This PR changes the interval from "at any time" to a weekly basis every Monday morning via a [cron](https://crontab.guru/crontab.5.html) expression. 

Important reference links:
- [Renovate schedule config](https://docs.renovatebot.com/configuration-options/#schedule)
- [Cron expression descriptor (for cron expression validation)](https://crontab.cronhub.io/)

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,9 +3,7 @@
   "extends": ["config:base", "schedule:daily", "group:allNonMajor"],
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
-  "schedule": [
-    "at 12am on Monday"
-  ],
+  "schedule": ["at 12am on Monday"],
   "rangeStrategy": "bump",
   "packageRules": [
     {
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,6 +3,11 @@
   "extends": ["config:base", "schedule:daily", "group:allNonMajor"],
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
+  "schedule": [
+    // We must use the * wildcard for minutes, as Renovate doesn't support minute granularity
+    // Every minute, between 12:00 AM and 12:59 AM, only on Monday (minutes are not suppoted)
+    "* 0 * * MON"
+  ],
   "rangeStrategy": "bump",
   "packageRules": [
     {
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,9 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:base", "schedule:daily", "group:allNonMajor"],
+  "extends": ["config:base", "schedule:earlyMondays", "group:allNonMajor"],
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
-  "schedule": ["at 12am on Monday"],
   "rangeStrategy": "bump",
   "packageRules": [
     {
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -4,9 +4,7 @@
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
   "schedule": [
-    // We must use the * wildcard for minutes, as Renovate doesn't support minute granularity
-    // Every minute, between 12:00 AM and 12:59 AM, only on Monday (minutes are not suppoted)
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -3,6 +3,11 @@
   "extends": ["config:base", "schedule:daily", "group:allNonMajor"],
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
+  "schedule": [
+    // We must use the * wildcard for minutes, as Renovate doesn't support minute granularity
+    // Every minute, between 12:00 AM and 12:59 AM, only on Monday (minutes are not suppoted)
```
xDivisionByZerox, line 8: I just wanted to leave this here so it's clear why the expression is the way it is. Is this a valid thought or should this be removed?

Shinigami92 (dismissed):
Maybe we should use just: https://docs.renovatebot.com/presets-schedule/#scheduleweekly

.github/renovate.json5
ST-DDT: Why every minute?

I also have found this, which might be a bit more readable although I don't know how this could be changed to achieve exactly what we want:

` "after 5pm on the 2nd day of the month"`

xDivisionByZerox: We want weekly, not monthly (as your phrase suggests).
Weekly would be `"at 12am on Monday"`.

xDivisionByZerox: > Why every minute?

Line 7. Renovate requires a '*' (translates to `every`) for the minute position.

xDivisionByZerox:
> Maybe we should use just: https://docs.renovatebot.com/presets-schedule/#scheduleweekly

I don't see why. Cron expressions are a known standard for developers (at least to some degree). This language thing seems nice for readability purposes but has no documentation whatsoever. I also really dislike the idea of expressing what I want in plain English and a machine gets what I want.

I'll not block this, because if this. I just don't see the reason why this would be a required change for a merge on this PR.

Shinigami92:
Everyone using renovate is using this syntax
Its the absolutely normal way of configuring renovate

- https://github.com/vitejs/vite/blob/5df788dfe2d89e541461e166f03afb38c2f1dd7e/.github/renovate.json5#L2
- https://github.com/prettier/plugin-pug/blob/d520345569c50dfa3350588bb835dce6ccf4e008/.github/renovate.json5#L3
- https://github.com/vitest-dev/vitest/blob/88d5764894f62449540209567df19926e46a91f3/.github/renovate.json5#L2

And what do you mean by not documented? I linked you the documentation ^^
You can even read more about it here: https://docs.renovatebot.com/config-presets/#how-to-use-preset-configs
And even more: https://docs.renovatebot.com/configuration-options/#schedule

xDivisionByZerox:
So what does the syntax translates to when I write the following

>Take yourself 3 out times before midnight but only on a full moon in winter

Does this run at all? That's what I mean by documented. You linked just a bunch of examples that are marked as "documentation". If I can just write plain English, what words do actually get picked up for the expression?

But whatever, I will just change the expression.

.github/renovate.json5
```diff
@@ -3,6 +3,7 @@
   "extends": ["config:base", "schedule:daily", "group:allNonMajor"],
```
Shinigami92, line 3:
```suggestion
  "extends": ["config:base", "schedule:earlyMondays", "group:allNonMajor"],
```

.github/renovate.json5
```diff
@@ -3,6 +3,7 @@
...
   "labels": ["c: dependencies"],
   "reviewersFromCodeOwners": true,
+  "schedule": ["at 12am on Monday"],
```
Shinigami92, line 6:
```suggestion
```

Shinigami92 (changes requested):

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox:
If your have such specific standards to workflow configuration or code standards in general please don't ask next time in the meeting who whats to take the task if you insist on doing it your way :)

Shinigami92:
> If your have such specific standards to workflow configuration or code standards in general please don't ask next time in the meeting who whats to take the task if you insist on doing it your way :)

But then I couldn't approve it and would need to ask another reviewer :see_no_evil:

1LISA2:
images (2)
下载 (2)

PR #1305, opened 2022-08-26
fix: set sideEffect to false in package.json

missing1984:
Ran into treeshaking issue in Webpack and found `sideEffect` is not declared in package.json. https://webpack.js.org/configuration/optimization/#optimizationsideeffects

This PR is setting `sideEffect: false` in package.json in order to enable Treeshaking. I've go over the code and it doesn't seems we have side effects so it should be safe to set it to false.


xDivisionByZerox:
I think our `UniqueStore` has side effects. 

See line 143, we store a result in a global store (at least if we use the default store).
https://github.com/faker-js/faker/blob/f78843edb05913c44ebd86535b0d50d22e99fc5e/src/modules/unique/unique.ts#L142-L145

Or am I misinterpreting?

missing1984:
> I think our `UniqueStore` has side effects.
> 
> See line 143, we store a result in a global store (at least if we use the default store).
> 
> https://github.com/faker-js/faker/blob/f78843edb05913c44ebd86535b0d50d22e99fc5e/src/modules/unique/unique.ts#L142-L145
> 
> Or am I misinterpreting?

It's fine as this store is local to the package and does not create sideEffect outside of package scope

xDivisionByZerox:
> It's fine as this store is local to the package and does not create sideEffect outside of package scope

Ok, I see. The scope is per package, my bad.

xDivisionByZerox (approved):

import-brain (approved):

Shinigami92 (approved):
Read a bit about what that is, and yeah, looks like its file for us to set that

@ST-DDT would you like to check independent from me if you are also okay with this?
That way I cannot bias you with my knowledge

ST-DDT:
I have trouble understanding, what this requires exactly.
The documentation isn't that helpful either and some SO comments actually contradict the examples.
I need some more time and probably some tests to determine, whether this is safe to apply.

Shinigami92:
Based on my comment I send here: https://github.com/mitodl/open-discussions/issues/3714#issuecomment-1239102062
I'm not so sure anymore if this PR would help in anyway even if we would merge it...
I'm tempted to close this and forward to our v8 solution which will fix this in a much higher order

PR #1238, opened 2022-08-07
feat: `datatype.hexadecimal` signature change

import-brain:
closes #954

PR #1306, opened 2022-08-26
fix(datatype): unintentional hex breaking change

import-brain:
This PR resolves a BREAKING CHANGE introduced in #1238.

### DIFF
--- a/src/modules/color/index.ts
+++ b/src/modules/color/index.ts
@@ -297,7 +297,9 @@ export class Color {
     let color: string | number[];
     let cssFunction: CSSFunction = 'rgb';
     if (format === 'hex') {
-      color = this.faker.datatype.hexadecimal({ length: includeAlpha ? 8 : 6 });
+      color = this.faker.datatype
+        .hexadecimal({ length: includeAlpha ? 8 : 6 })
+        .slice(2);
       color = formatHexColor(color, options);
       return color;
     }
--- a/src/modules/color/index.ts
+++ b/src/modules/color/index.ts
@@ -297,8 +297,10 @@ export class Color {
     let color: string | number[];
     let cssFunction: CSSFunction = 'rgb';
     if (format === 'hex') {
-      color = this.faker.datatype
-        .hexadecimal({ length: includeAlpha ? 8 : 6, prefix: '' });
+      color = this.faker.datatype.hexadecimal({
+        length: includeAlpha ? 8 : 6,
+        prefix: '',
+      });
       color = formatHexColor(color, options);
       return color;
     }
--- a/src/modules/color/index.ts
+++ b/src/modules/color/index.ts
@@ -298,8 +298,7 @@ export class Color {
     let cssFunction: CSSFunction = 'rgb';
     if (format === 'hex') {
       color = this.faker.datatype
-        .hexadecimal({ length: includeAlpha ? 8 : 6 })
-        .slice(2);
+        .hexadecimal({ length: includeAlpha ? 8 : 6, prefix: '' });
       color = formatHexColor(color, options);
       return color;
     }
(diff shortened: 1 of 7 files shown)


src/modules/color/index.ts
```diff
@@ -297,7 +297,9 @@ export class Color {
     let color: string | number[];
     let cssFunction: CSSFunction = 'rgb';
     if (format === 'hex') {
-      color = this.faker.datatype.hexadecimal({ length: includeAlpha ? 8 : 6 });
+      color = this.faker.datatype
+        .hexadecimal({ length: includeAlpha ? 8 : 6 })
+        .slice(2);
```
xDivisionByZerox, line 302: Use the explicitly an empty prefix? If we have the feature, use it.
```suggestion
        .hexadecimal({ length: includeAlpha ? 8 : 6, prefix: '' })
```

src/modules/database/index.ts
```diff
@@ -69,6 +69,8 @@ export class Database {
    * faker.database.mongodbObjectId() // 'e175cac316a79afdd0ad3afb'
    */
   mongodbObjectId(): string {
-    return this.faker.datatype.hexadecimal({ length: 24, case: 'lower' });
+    return this.faker.datatype
+      .hexadecimal({ length: 24, case: 'lower' })
+      .replace('0x', '');
```
xDivisionByZerox, line 74: You are getting creative with the removes?^^
```suggestion
      .hexadecimal({ length: 24, case: 'lower', prefix: '' });
```

src/modules/datatype/index.ts
```diff
@@ -220,7 +215,7 @@ export class Datatype {
       };
     }
 
-    const { length = 1, prefix = '', case: letterCase = 'mixed' } = options;
```
xDivisionByZerox, line 223: Why did you drop the entire feature? Just revert the default value here to `'0x'` instead of `''`.

xDivisionByZerox (changes requested):

src/modules/git/index.ts
```diff
@@ -92,7 +92,9 @@ export class Git {
    * faker.git.commitSha() // '2c6e3880fd94ddb7ef72d34e683cdc0c47bec6e6'
    */
   commitSha(): string {
-    return this.faker.datatype.hexadecimal({ length: 40, case: 'lower' });
+    return this.faker.datatype
+      .hexadecimal({ length: 40, case: 'lower' })
+      .substring(2);
```
xDivisionByZerox, line 97: Use `prefix` instead of string manipulation.

```suggestion
      .hexadecimal({ length: 40, case: 'lower', prefix: '' });
```

src/modules/git/index.ts
```diff
@@ -102,6 +104,8 @@ export class Git {
    * faker.git.shortSha() // '6155732'
    */
   shortSha(): string {
-    return this.faker.datatype.hexadecimal({ length: 7, case: 'lower' });
+    return this.faker.datatype
+      .hexadecimal({ length: 7, case: 'lower' })
+      .substring(2);
```
xDivisionByZerox, line 109: Use `prefix` instead of string manipulation.

```suggestion
      .hexadecimal({ length: 7, case: 'lower', prefix: '' });
```

xDivisionByZerox (dismissed):

src/modules/git/index.ts
```diff
@@ -92,7 +92,11 @@ export class Git {
    * faker.git.commitSha() // '2c6e3880fd94ddb7ef72d34e683cdc0c47bec6e6'
    */
   commitSha(): string {
-    return this.faker.datatype.hexadecimal({ length: 40, case: 'lower' });
+    return this.faker.datatype.hexadecimal({
```
Shinigami92, line 95: prettier check is asking for a format

Shinigami92 (changes requested):

import-brain:
@xDivisionByZerox Thank you for the refactoring help, I missed those instances of string manipulation.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1307, opened 2022-08-27
fix(random): retry on invalid word generation

### LINKED ISSUE #129: Use word methods in `random.word()`

As a developer, I want `random.word()` to draw from the entire pool of appropriate words.

### Suggested solution

Add the word methods (`word.noun`, `word.verb`, etc.) to the pool from which `random.word()` pulls.

https://github.com/faker-js/faker/blob/9506122a7725d9919e5a5c6ba036484cb608cd83/lib/random.js#L150-L182

### Alternative

These words might be transitively included via some other method, I'm not sure.

### Additional context

_No response_

xDivisionByZerox:
This PR adds more source methods to the `random.word` method. It also sorts the sources alphabetically for better maintainability.

Fixes #129.


ST-DDT (approved):
I'm currently unable to test this, but it looks good to me from the sources.

ST-DDT:
There is however this issue, that I feel would be a better solution.
https://github.com/faker-js/faker/issues/339

import-brain (approved):

src/modules/random/index.ts
```diff
@@ -178,7 +209,12 @@ export class Random {
       // randomly pick from the many faker methods that can generate words
       const randomWordMethod = this.faker.helpers.arrayElement(wordMethods);
 
-      result = randomWordMethod();
+      try {
+        result = randomWordMethod();
+      } catch {
+        // catch missing locale data potentially required by randomWordMethod
```
Shinigami92, line 215: Is this a fix worth mention in changelog?

xDivisionByZerox: Don't think so, since there wasn't a reported issue about this (?). But I'm not sure. If yes, the fixed needs to get merged first in a separate PR. Otherwise test in this PR (without this fix) can potentially fail.

ST-DDT: The more I think about it the more I do want to scrap this method in favor of `faker.word.random`.

Shinigami92: > since there wasn't a reported issue about this

You are thinking the wrong way around ^^
It's not only a fix if it is a reported issue, it can also be discovered by us and just be a fix

xDivisionByZerox: > The more I think about it the more I do want to scrap this method in favor of faker.word.random.

That's the plan after this refactor. Isn't it?

> It's not only a fix if it is a reported issue, it can also be discovered by us and just be a fix

Yes, yes. I know your right, but I'm kinda lazy to split this into 2 separate PRs^^ (I'm sorry for being unprofessional)

Shinigami92: It's not about being unprofessional but having this correctly in the changelogs while auto-generate them
So I will switch this from `refactor` to `fix` so it just show up in changelog

pkuczynski: I still vote for `faker.word.any`, as everything is _random_ in faker :)

ST-DDT:
I will merge this because @xDivisionByZerox requested it, befor the moving it to faker.word.random()

PR #1319, opened 2022-08-29
chore(release): 7.5.0

Shinigami92:


xDivisionByZerox (approved):

ST-DDT (approved):

xDivisionByZerox:
But why do you have all these version changes in the package.json file? This should be done y renovate, shouldn't it?

Shinigami92:
> But why do you have all these version changes in the package.json file? This should be done y renovate, shouldn't it?

Because otherwise while in the release process the pnpm lock file will be regenerated and then it cannot commit it while in release process and then fail

Some of these are also even explicitly omitted by our renovate config (I think)

Shinigami92:
So no, the ignores are not in there anymore
I don't know why tsx doesn't got updated, but I think the node types are somehow really special

1LISA2:
images (3)
下载 (4)

Issue #1333, opened 2022-09-05
Improve default seed initialization


ST-DDT:
To avoid issues like this https://github.com/faker-js/faker/issues/1324#issuecomment-1236897601 where the faker instance is initialized by multiple threads in parallel (thus using the same timestamp in the seed)

https://github.com/faker-js/faker/blob/cc18b22e61fdfcd75db5196433931f00810393c5/src/modules/mersenne/index.ts#L11

The seed initialization should either take a more precise time source or add a random element (in the hope that this returns a different value per thread).

Shinigami92:
So like in https://github.com/faker-js/faker/blob/cc18b22e61fdfcd75db5196433931f00810393c5/src/faker.ts#L243 we could replace it with `Math.ceil(Math.random() * Number.MAX_SAFE_INTEGER)`

shraddhafalane:
Would like to work on it

Shinigami92:
> Would like to work on it

@shraddhafalane Go for it

PR #1334, opened 2022-09-06
fix: improve default seed initialization

shraddhafalane:
closes https://github.com/faker-js/faker/issues/1333


ST-DDT (approved):
Looks good to me.

Is there a way to test whether this fixes the original issue?

Shinigami92:
> Is there a way to test whether this fixes the original issue?

Maybe we could add a test using https://vitest.dev/guide/features.html#running-tests-concurrently 🤔 
But not sure if this really works

import-brain (approved):
Also LGTM, we definitely need to test this though

PR #1335, opened 2022-09-06
docs: update order of headings to be semantically correct

mnigh:
These changes ensures that headings are in a logical order in several files using markdown.

> For example, the heading level following an h1 element should be an h2 element, not an h3 element.

source:  https://dequeuniversity.com/rules/axe/4.4/heading-order

### DIFF
--- a/docs/guide/migration-guide-v5.md
+++ b/docs/guide/migration-guide-v5.md
@@ -34,7 +34,7 @@ If you depend on a specific version of Faker you still can reference the version
 Faker now ships with its own types! Remove `@types/faker` from your `package.json` to avoid conflicts.
 :::
 
-### Tree-shaking
+## Tree-shaking
 
 :::warning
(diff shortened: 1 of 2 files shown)


xDivisionByZerox (approved):

xDivisionByZerox:
Can you provide a description that explains the change you did here? If possible, link to some sources about "semantically correct HTML".

ST-DDT (approved):

docs/guide/migration-guide-v5.md
```diff
@@ -2,7 +2,7 @@
 
 [[toc]]
```
Shinigami92, line 3: Quick question (as I didn't checked it out myself)
But now with VitePress v1 we don't need the `toc` anymore as we have the righthand sidebar as the `toc` 👀 

If so, please remove the `toc`
_applies also for the other file_

ejcheng (approved):

PR #1337, opened 2022-09-06
docs: add `@since` tags to all methods

matthewmayer:
Following on from discussion in https://github.com/faker-js/faker/discussions/1241 i added `@since` jsdoc tags to all existing public methods, and exposed this in the documentation:

image

This is a one-off large diff for this initial integration, in future it would just be necessary to add `@since` tags for any newly added methods.

### DIFF
--- a/scripts/apidoc/utils.ts
+++ b/scripts/apidoc/utils.ts
@@ -154,6 +154,6 @@ export function isDeprecated(signature: SignatureReflection): boolean {
  *
  * @returns the contents of the @since tag
  */
-export function since(signature: SignatureReflection): string {
+export function extractSince(signature: SignatureReflection): string {
   return extractTagContent('@since', signature).join().trim();
 }
--- a/scripts/apidoc/utils.ts
+++ b/scripts/apidoc/utils.ts
@@ -146,3 +146,14 @@ export function joinTagParts(parts?: CommentDisplayPart[]): string | undefined {
 export function isDeprecated(signature: SignatureReflection): boolean {
   return extractTagContent('@deprecated', signature).length > 0;
 }
+
+/**
+ * Extracts the "since" tag from the provided signature.
+ *
+ * @param signature The signature to check.
+ *
+ * @returns the contents of the @since tag
+ */
+export function since(signature: SignatureReflection): string {
+  return extractTagContent('@since', signature).join().trim();
+}
--- a/src/modules/address/index.ts
+++ b/src/modules/address/index.ts
@@ -102,7 +102,6 @@ export class Address {
    * @example
    * faker.address.cityPrefix() // 'East'
    *
-   *
    * @since 2.0.1
    *
    * @deprecated
--- a/src/modules/address/index.ts
+++ b/src/modules/address/index.ts
@@ -28,6 +28,7 @@ export class Address {
    * faker.address.zipCode() // '17839'
    * faker.address.zipCode('####') // '6925'
    *
+   * @since 2.0.1
    */
   zipCode(format?: string): string {
     // if zip format is not specified, use the zip format defined for the locale
@@ -54,6 +55,8 @@ export class Address {
    * @example
    * fakerUS.address.zipCodeByState("AK") // '99595'
    * fakerUS.address.zipCodeByState("??") // '47683-9880'
+   *
+   * @since 5.0.0
    */
   zipCodeByState(state: string): string {
     const zipRange = this.faker.definitions.address.postcode_by_state?.[state];
@@ -70,6 +73,8 @@ export class Address {
    *
    * @example
    * faker.address.city() // 'East Jarretmouth'
+   *
+   * @since 2.0.1
    */
   city(format?: string | number): string {
     if (format != null) {
@@ -97,6 +102,9 @@ export class Address {
    * @example
    * faker.address.cityPrefix() // 'East'
    *
+   *
+   * @since 2.0.1
+   *
    * @deprecated
    * Use `faker.address.city()` instead.
    */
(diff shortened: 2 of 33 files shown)


scripts/apidoc/utils.ts
```diff
@@ -146,3 +146,13 @@ export function joinTagParts(parts?: CommentDisplayPart[]): string | undefined {
 export function isDeprecated(signature: SignatureReflection): boolean {
   return extractTagContent('@deprecated', signature).length > 0;
 }
+/**
```
ST-DDT, line 149:
```suggestion
}

/**
```

src/modules/color/index.ts
```diff
@@ -166,10 +166,11 @@ export class Color {
   }
 
   /**
-   * Returns a random human readable color name.
+   * Returns a random human readable color name
```
ST-DDT, line 169:
```suggestion
   * Returns a random human readable color name.
```

scripts/apidoc/utils.ts
```diff
@@ -146,3 +146,13 @@ export function joinTagParts(parts?: CommentDisplayPart[]): string | undefined {
...
+ * Extracts the "since"
```
xDivisionByZerox, line 150: Not sure about this description.

```suggestion
 * Extracts the "since" tag from the provided signature.
```

ST-DDT (changes requested):
I think I like this.

Please also add a new test case to: https://github.com/faker-js/faker/blob/main/test/scripts/apidoc/signature.example.ts

Also not sure whether there should be an empty line after the `@examples` and before `@since`.

xDivisionByZerox:
> Also not sure whether there should be an empty line after the `@examples` and before `@since`.

There definitely should be IMO. Looks a lot better with all topics grouped. Much easier to scan (as a human) as well.

xDivisionByZerox (changes requested):
Please add spaces around the `@since` tag so it aligns with the other tags.

src/modules/address/index.ts
```diff
@@ -28,6 +28,7 @@ export class Address {
    * faker.address.zipCode() // '17839'
    * faker.address.zipCode('####') // '6925'
    *
+   * @since 2.0.1
```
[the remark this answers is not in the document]
ST-DDT: https://github.com/faker-js/faker/discussions/1241#discussioncomment-3367044

matthewmayer:
> I think I like this.
> 
> 
> 
> Please also add a new test case to: https://github.com/faker-js/faker/blob/main/test/scripts/apidoc/signature.example.ts
> 
> 
> 
> Also not sure whether there should be an empty line after the `@examples` and before `@since`.

Sorry my typescript is not really up to this. Maybe someone else can help me with the tests

matthewmayer:
added the newlines and made the other suggested changes.

src/modules/address/index.ts
```diff
@@ -97,6 +102,9 @@ export class Address {
    * @example
    * faker.address.cityPrefix() // 'East'
    *
+   *
```
ST-DDT, line 105: This is one newline too much.

ST-DDT (commented):
I will address the test later today.

src/modules/address/index.ts
Shinigami92:
```suggestion
```

Shinigami92 (commented):
Really cool 👍 
Count me as approved when all the missing requested/required tasks are finished

ST-DDT (changes requested):
faker.fake and faker.unique are missing the `@since` tags.
The `@since` tags won't be picked up for the color module.
I'm not sure why.

ST-DDT:
The color methods have multiple signatures. You have to add the `@since` either the last one (with jsdocs) or all of them.

xDivisionByZerox (dismissed):

matthewmayer:
In general it seems its only reading the JSDoc for the last version of a method with multiple signatures like `faker.color.rgb()`? So in general should the since tag be only added to the last one, or all of them?

ST-DDT:
I think the last one is enough.

Shinigami92:
Current test results in CI:
image

matthewmayer:
i agree maybe all is better. will manually fix the ones in color

matthewmayer:
@Shinigami92 how can i run those tests locally to make sure I caught everything?

Shinigami92 (dismissed):

src/modules/color/index.ts
```diff
@@ -444,6 +476,10 @@ export class Color {
    *
    * @example
    * faker.color.hwb() // [201, 0.21, 0.31]
+   *
+   * @since 7.0.0
+   *
+   * @since 7.0.0
```
xDivisionByZerox, line 482: Duplicated tag in the same signature.

xDivisionByZerox (changes requested):

src/modules/color/index.ts
```diff
@@ -493,6 +535,9 @@ export class Color {
    * faker.color.hwb({ format: 'decimal' }) // [201, 0.21, 0.31]
    * faker.color.hwb({ format: 'css' }) // hwb(194 0% 0%)
    * faker.color.hwb({ format: 'binary' }) // (8-32 bits x 3)
+   *
+   *
+   * @since 7.0.0
```
ST-DDT, line 540:
```suggestion
   *
   * @since 7.0.0
```

matthewmayer:
fixed those two issues

ST-DDT (approved):

xDivisionByZerox (approved):

pkuczynski (approved):

PR #1375, opened 2022-09-13
test(finance): use validator.js for bitcoin addresses

### LINKED ISSUE #1184: Generated bitcoin addresses should pass validators.isBtcAddress

In #1181 the question came up, whether our [bitcoin generator](https://github.com/faker-js/faker/blob/15e3eed927953446d3e3aa8dcd7fca3f8472ff3e/src/modules/finance/index.ts#L213) produces valid bitcoin addresses.
According to [validators.isBtcAddress](https://github.com/validatorjs/validator.js/blob/1bb14e85dd7fb8cec8a43ff586c0d869977cd684/README.md?plain=1#L104) it doesn't do so all the time.

### Suggested solution

We should

- check whether the validator.js method is correct,
- use it in [our tests](https://github.com/faker-js/faker/blob/15e3eed927953446d3e3aa8dcd7fca3f8472ff3e/test/finance.spec.ts#L295),
- and verify that the tests pass.

### Alternative

_No response_

### Additional context

_No response_

Minozzzi:
Fixes: #1184


pnpm-lock.yaml
```diff
@@ -1452,8 +1452,8 @@ packages:
     engines: {node: '>=10'}
     hasBin: true
     dependencies:
-      JSONStream: 1.3.5
       is-text-path: 1.0.1
+      JSONStream: 1.3.5
```
Shinigami92, line 1456: How did this happen?
If it's not related to the PR, I would suggest to revert it

test/finance.spec.ts
```diff
@@ -289,8 +290,8 @@ describe('finance', () => {
            */
 
           expect(bitcoinAddress).toBeTruthy();
+          expect(isValidBtcAddress(bitcoinAddress)).toBeTruthy();
```
Shinigami92, line 293: 1. This should be below the typeof check as more generic primitive checks should be made first
   Otherwise you already know that it was a `string`
2. I think using `toSatisfiy` is a better fit here

Shinigami92 (changes requested):

test/finance.spec.ts
xDivisionByZerox: > 2. I think using `toSatisfiy` is a better fit her

Just to explain: `toSatisfiy` give a much easier  to understand error message then calling the validator function and check if thee result is truthy.

pnpm-lock.yaml
xDivisionByZerox: I guess an auto formater on save. This way, the entries are sorted correctly (A-Z). Still, not related to this PR.

Minozzzi: It happened because i added the validatorjs but i think it was already added and it just formatted. I can remove if needed

Shinigami92: yeah pls revert the file, it is not depended with the PR itself

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #1378, opened 2022-09-18
feat(finance): branch code option in bic()

### LINKED ISSUE #1158: Branch code option for SWIFT/BIC codes
From [bank.codes](https://bank.codes/guides/swift-codes-guide/): "SWIFT codes are laid out in a standard format. You'll find 8 letters or digits for a standard SWIFT code, or 11 if you also have the branch code added. The last 3 digits which show the specific branch an account is held at are not always necessary."

Currently, `finance.bic()` uses a ternary operator https://github.com/faker-js/faker/blob/df9c650432de2d119b4f17d61a0c554b744debd0/src/modules/finance/index.ts#L412-L414 to randomly decide whether to generate an 8 char or 11 char SWIFT code.

I propose that we add a boolean `branchCode` parameter to `finance.bic()` to give users the ability to pick whether they want a standard SWIFT code or a code with the branch code at the end.

import-brain:
closes #1158


src/modules/finance/index.ts
```diff
@@ -429,25 +429,32 @@ export class FinanceModule {
...
    *
    * @since 4.0.0
    */
-  bic(): string {
+  bic(
+    options: {
+      branchCode?: boolean;
+    } = {}
+  ): string {
+    const { branchCode = false } = options;
```
Shinigami92, line 447: Previously the default was set via `this.faker.datatype.boolean()`
I think we need that back :thinking: 
Now the XXX will never be generated but especially I'm at HASPA they are mostly using `HASP DE HH XXX`

Shinigami92 (changes requested):

src/modules/finance/index.ts
```diff
@@ -429,25 +429,32 @@ export class FinanceModule {
...
+   * faker.finance.bic() // 'WYAUPGX1'
+   * faker.finance.bic({ branchCode: true }) // 'KCAUPGR1432'
+   * faker.finance.bic({ branchCode: false }) // 'XDAFQGT7'
    *
    * @since 4.0.0
    */
-  bic(): string {
+  bic(
+    options: {
+      branchCode?: boolean;
```
ST-DDT, line 444: Maybe use `useBranchCode` or `includeBranchCode`?

import-brain: Hmm, I really like `branchCode`. What does the team think?

src/modules/finance/index.ts
import-brain: Are you saying you want the options destructuring to be like this?
`const { branchCode = this.faker.datatype.boolean() } = options;`

What do you mean by `XXX` would never be generated? It currently generates three capital alphanumeric characters, so there should be a chance of getting `XXX`. I'm probably misunderstanding what you mean, though

Shinigami92: Yes, but `XXX` is a more generic case and should be occur more often

src/modules/finance/index.ts
Shinigami92: From a user perspective I think @ST-DDT is right

Shinigami92:
Tests dont pass

test/finance.spec.ts
```diff
@@ -480,11 +483,21 @@ describe('finance', () => {
...
       describe('bic()', () => {
-        it('should return a random yet formally correct BIC number', () => {
+        it('should return a BIC number', () => {
           const bic = faker.finance.bic();
 
           expect(bic).toBeTypeOf('string');
-          expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}([A-Z0-9]{3})?$/);
+          bic.length === 8
+            ? expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}$/)
+            : expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}[A-Z0-9]{3}$/);
```
import-brain, line 492:
```suggestion
          bic.length === 8
            ? expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}$/)
            : expect(bic).toMatch(/^[A-Z]{6}[A-Z0-9]{2}[A-Z0-9]{3}$/);
```
This is a bit janky, any better way to write this?

ST-DDT: Wrap the last part in parenthesis and add a question mark behind this to mark it as optional.

ST-DDT:
Also: https://github.com/validatorjs/validator.js/blob/master/src/lib/isBIC.js

import-brain:
Seems like `validator.isBIC` does not recognize `XK` (Kosovo) as a country code

ST-DDT:
> Seems like `validator.isBIC` does not recognize `XK` (Kosovo) as a country code, it intermittently fails when `XK` is the country code

Maybe report this to validatorjs for them to fix?

import-brain:
> > Seems like `validator.isBIC` does not recognize `XK` (Kosovo) as a country code, it intermittently fails when `XK` is the country code
> 
> Maybe report this to validatorjs for them to fix?

https://github.com/validatorjs/validator.js/issues/2045

ST-DDT (approved):

PR #1385, opened 2022-09-23
ci: workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   unit-test:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -7,8 +7,13 @@ on:
       - edited
       - synchronize
 
+permissions: {}
 jobs:
   main:
+    permissions:
+      pull-requests: read # to analyze PRs (amannn/action-semantic-pull-request)
+      statuses: write # to mark status of analyzed PR (amannn/action-semantic-pull-request)
+
     runs-on: ubuntu-latest
     name: Semantic Pull Request
     steps:


ST-DDT (approved):
This PR seems legit, I haven't seen that many repositories that have the permissions explicitly set yet though.
A quick search in the js ecosystem didn't result in any explicit permission configurations (vite/vitest/...).
However, I found a PR for java (spring) by the same author.

Here is the GH-TOKEN permission diff:

Before: https://github.com/faker-js/faker/actions/runs/3107588269/jobs/5035863133#step:1:19
After: https://github.com/faker-js/faker/actions/runs/3115625100/jobs/5052991942#step:1:17

import-brain (approved):

sashashura:
> This PR seems legit, I haven't seen that many repositories that have the permissions explicitly set yet though. A quick search in the js ecosystem didn't result in any explicit permission configurations (vite/vitest/...). However, I only found a single PR for java (spring) by the same author.

There are [19488 workflows](https://grep.app/search?q=permissions%3A&filter[path][0]=.github/workflows/&filter[lang][0]=YAML) already using explicit permissions. Though I don't know how many repositories it is.

xDivisionByZerox (approved):

PR #1386, opened 2022-09-25
feat(science): add more units in en locale

import-brain:


import-brain:
Should I reorder these units to be in alphabetical order?

ST-DDT (approved):
I had to lookup these units, since I haven't used some of them before.
Not sure whether we need any more "exotic" units that this (unless there is an explicit need for a particular unit).

import-brain:
> I had to lookup these units, since I haven't used some of them before.
> 
> Not sure whether we need any more "exotic" units that this (unless there is an explicit need for a particular unit).

These units are all from the Wikipedia page on SI units (should I link that in a comment?). I added some of the SI derived units with special names/symbols so that the definitions contain all SI units.

Shinigami92 (approved):

pkuczynski (approved):

PR #1392, opened 2022-09-26
fix(finance): Update Belarusian Ruble to new version

AdriVanHoudt:
I updated the code and symbol according to wikipedia. 
See https://en.wikipedia.org/wiki/Belarusian_rubel#Third_rubel,_2016%E2%80%93present

### DIFF
--- a/src/locales/en/finance/currency.ts
+++ b/src/locales/en/finance/currency.ts
@@ -88,8 +88,8 @@ export default {
     symbol: 'P',
   },
   'Belarussian Ruble': {
-    code: 'BYR',
-    symbol: 'p.',
+    code: 'BYN',
+    symbol: 'Rbl',
   },
   'Belize Dollar': {
     code: 'BZD',


import-brain (approved):

src/locales/en/finance/currency.ts
```diff
@@ -88,8 +88,8 @@ export default {
     symbol: 'P',
   },
   'Belarussian Ruble': {
```
import-brain, line 90:
```suggestion
  'Belarusian Ruble': {
```
Somewhat unrelated, but shouldn't this be "Belarusian" and not "Belarussian"?

AdriVanHoudt: Jup, you want me to add a commit to fix that or add it to the current commit?

xDivisionByZerox: Please provide an additional PR for that fix so its easier to understand in the commit history (and thus in release notes).

xDivisionByZerox (approved):

src/locales/en/finance/currency.ts
AdriVanHoudt: Done https://github.com/faker-js/faker/pull/1395

ST-DDT (approved):

PR #1393, opened 2022-09-26
fix(finance): update Zambian Kwacha to new version

AdriVanHoudt:
I also added a symbol.
See https://en.wikipedia.org/wiki/Zambian_kwacha

### DIFF
--- a/src/locales/en/finance/currency.ts
+++ b/src/locales/en/finance/currency.ts
@@ -672,8 +672,8 @@ export default {
     symbol: 'N$',
   },
   'Zambian Kwacha': {
-    code: 'ZMK',
-    symbol: '',
+    code: 'ZMW',
+    symbol: 'K',
   },
   'Zimbabwe Dollar': {
     code: 'ZWL',


xDivisionByZerox (approved):

src/locales/en/finance/currency.ts
```diff
@@ -672,8 +672,8 @@ export default {
     symbol: 'N$',
   },
   'Zambian Kwacha': {
-    code: 'ZMK',
-    symbol: '',
+    code: 'ZMW',
+    symbol: 'K',
```
ST-DDT, line 676: Maybe ZK?

AdriVanHoudt: If you look at https://en.wikipedia.org/wiki/Zambian_kwacha#New_Kwacha_(2012_series) it talks about KR being used temporarily and then switching back to K. ZK is not mentioned anywhere so I picked K.
I do see ZK in other unofficial places but then something like https://www.boz.zm/currency-family.htm is also using K.
So not 100% sure 😅

ST-DDT (approved):

import-brain (approved):

PR #1396, opened 2022-09-26
fix(image): properly lock loremflickr.com images

wesen:
https://loremflickr.com requires the lock= parameter to be given
to serve an identical image (at least as long as its cache is not
updated).


ST-DDT (approved):

src/modules/image/index.ts
```diff
@@ -111,7 +111,9 @@ export class ImageModule {
     }
 
     if (randomize) {
-      url += `?${this.faker.datatype.number()}`;
+      url += `?lock=${this.faker.datatype.number()}`;
+    } else {
+      url += `?lock=0`;
```
ST-DDT, line 116: This changes the current behavior. Currently, returning a random picture is intentional.
In the next major version we plan to change the default of randomize to true.

ST-DDT (changes requested):

src/modules/image/index.ts
wesen: does this mean that randomize = false still randomizes the image? I'm sorry if I misunderstood the meaning.

ST-DDT: Yes, `randomize === false` => no `lock` url suffix.

src/modules/image/index.ts
```diff
@@ -39,9 +39,9 @@ export class ImageModule {
    * @param randomize Whether to randomize the image or not. Defaults to `false`.
    *
    * @example
-   * faker.image.image() // 'https://loremflickr.com/640/480/city'
-   * faker.image.image(1234, 2345) // 'https://loremflickr.com/1234/2345/sports'
-   * faker.image.image(1234, 2345, true) // 'https://loremflickr.com/1234/2345/nature?56789'
+   * faker.image.image() // 'https://loremflickr.com/640/480/city?lock=0'
+   * faker.image.image(1234, 2345) // 'https://loremflickr.com/1234/2345/sports?lock=0'
```
ST-DDT, line 43: Please adjust the examples accordingly.

ST-DDT (changes requested):

src/modules/image/index.ts
wesen: Updated!

ST-DDT (approved):

import-brain (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1398, opened 2022-09-26
docs: document how to write tests

### LINKED ISSUE #1278: Document the usage of the seeded test (factory)

Currently the usage of the seeded test factory is not documented well.

### Suggested solution

Add a writing tests section to the contributing md and explain the seeded test factory there along with our expectations regarding random seeded tests/junit tests in general.

### Alternative

_No response_

### Additional context

_No response_

ST-DDT:
Fixes #1278


CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
 
 You can view a generated code coverage report at `coverage/index.html`.
 
+### Adding tests for new methods/parameters
+
+All methods should have tests for all their parameters.
+
+Usually, they are will be a test case for each of the following scenarios:
```
Shinigami92, line 52: > they are will be

Reads very wrong :thinking:

ST-DDT: Fixed

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+});
+```
+
+You can update the snapshot files by running `pnpm run test -u`.
+
+### Random Seeded Tests
+
+The random seeded tests return a random result in each iteration.
+They are intended to check for edge cases and general result checks.
+The tests will usually use regex or [validator.js](https://github.com/validatorjs/validator.js) check to ensure the method returns valid results.
```
import-brain, line 119:
```suggestion
They are intended to check for edge cases and function as general result checks.
The tests will usually use regex or [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results.
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+- Special cases
+
+We won't test for arguments that don't match the expected types.
+
+Our tests are separated into two parts:
+
+- Fixed Seeded Tests
+- Random Seeded Tests
+
+### Fixed Seeded Tests
```
xDivisionByZerox, line 66: I feel like this should be a subheading in the "testing" section. So like an h4 in this case. Same goes for `### Random Seeded Tests` in ln 115.

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+
+The fixed seeded tests are used to check that the returned results are matching the users expectations.
```
xDivisionByZerox, line 68: It's not only important for expectations but also for deterministic. We have an intern doc that states:
> A minor version should always have the same deterministic. That means for example all 7.0 versions (7.0.0, 7.0.1, etc) will return the same results when using the same seed.

So we should probably add that here:
```suggestion
The fixed seeded tests are used to check that the returned results are matching the users expectations and are deterministic.
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+They are intended to check for edge cases and function as general result checks.
+The tests will usually use regex or [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results.
+
+```ts
+import { describe, expect, it } from 'vitest';
+import { faker } from '../src';
+
+describe('someModule', () => {
+  describe(`random seeded tests for seed ${faker.seed()}`, () => {
+    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
```
xDivisionByZerox, line 127: Why do we need to repeat these tests? Deterministic is ensured by "Fixed seeded tests". So all this does is artificially bum our test count.

If this has any reason it should probably be explained.

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+});
+```
+
+You can update the snapshot files by running `pnpm run test -u`.
+
+### Random Seeded Tests
+
+The random seeded tests return a random result in each iteration.
+They are intended to check for edge cases and function as general result checks.
+The tests will usually use regex or [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results.
```
xDivisionByZerox, line 119: I can remember that we wanted to use as less RegExp as possible as it can get pretty fast out of hand and the error messages are not always the best to read. Sure validator.js might use RegExp under the hood as well, but I feel like that's a different thing...

import-brain: Maybe add something like "Please use validator.js before using regex"?

ST-DDT:
```suggestion
The tests will usually use regex or preferably [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results.
```

CONTRIBUTING.md
ST-DDT:
```suggestion
#### Fixed Seeded Tests
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+    t.it('noArgs')
+      .it('with param1', true)
+      .it('with param1 and param2', false, 1337);
+  });
+});
+```
+
+You can update the snapshot files by running `pnpm run test -u`.
+
+### Random Seeded Tests
```
ST-DDT, line 115:
```suggestion
#### Random Seeded Tests
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,102 @@ pnpm run coverage
...
+
+The random seeded tests return a random result in each iteration.
+They are intended to check for edge cases and function as general result checks.
+The tests will usually use regex or [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results.
+
```
ST-DDT, line 120:
```suggestion
We repeat these tests a few times to reduce the likelihood of flaky tests caused by the various corner cases that the implementation or the relevant locale data might have. The loop can also be used to steeply increase the test count to trigger rare issues.

```

CONTRIBUTING.md
ST-DDT: Done

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #1401, opened 2022-09-27
fix(locale): change id_ID name to `Indonesian`

fzn0x:
The citizens of Indonesia should be called `Indonesian`

I notice that only Indonesian whos using their country name for calling the citizen that originally from Indonesia, it's should be `Indonesian`


ejcheng (approved):

ST-DDT (approved):

pkuczynski:
Change seems to be correct, but worth to notice, that this is not about _citizens_ of Indonesia, but about locale name. I will update the PR title.

pkuczynski (approved):

xDivisionByZerox (approved):

fzn0x:
Thanks @pkuczynski !

Issue #813, opened 2022-04-08
Add CI steps to test execution of generate scripts


Shinigami92:
### Clear and concise description of the problem

We need to add CI steps for the execution of some generate scripts, so we guarantee that these will not break

### Suggested solution

Add steps

ST-DDT:
See also #389

xDivisionByZerox:
Concept:

Add a job to the [`ci.yaml`](https://github.com/faker-js/faker/blob/508082cbde088d2834b324c15dd75f326cb896c7/.github/workflows/ci.yml) that runs the `generate-locale` and `generate-locale` scripts from our [`package.json`](https://github.com/faker-js/faker/blob/508082cbde088d2834b324c15dd75f326cb896c7/package.json#L60-L61).

ST-DDT:
And it should ensure that there are no changed files afterwards.

sunadoi:
Could you assign to me?

PR #1405, opened 2022-10-02
infra: add CI step to detect generate script diffs

sunadoi:
#813 

After running the `generate:locales` command in the workflow/ci pipeline, check for some uncommitted diffs with the following command.

```
git add .
git diff --cached --exit-code
```

- The `--exit-code` option causes it to fail if there are any diffs.

I tested this CI on my repository.
It seems to work well.
ref: https://github.com/sunadoi/faker/actions/runs/3164634036/jobs/5153063988


ST-DDT (commented):
This does not detect deleted files, but I'm not sure whether this is necessary.

sunadoi:
> This does not detect deleted files

I didn't give it enough thought.
How about checking git diff before and after git add?
The former checks for file updates/deletions, the latter checks for file creation.

```
git diff --exit-code
git add -N .
git diff --exit-code
```

.github/workflows/ci.yml
```diff
@@ -203,3 +203,35 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
+  generate-locales:
```
Shinigami92, line 206: Please add a blank line here

```suggestion

  generate-locales:
```

sunadoi: Sorry for my simple mistake, I fixed it.

sunadoi:
To be able to detect deleted files, the method of checking for diffs has been changed to the following command.

```
git add .
git diff --cached --exit-code
```

.github/workflows/ci.yml
```diff
@@ -203,3 +203,36 @@ jobs:
...
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Generate Locales
+        run: pnpm run generate:locales
+
+      - name: Check diff
+        run: |
+          git add .
+          git diff --cached --exit-code
```
ST-DDT, line 238: Should we write an error message for the user with a hint what to do here?

ST-DDT (changes requested):

.github/workflows/ci.yml
```diff
@@ -203,3 +203,36 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
+
+  generate-locales:
+    runs-on: ubuntu-latest
+    name: 'Locales Diff Detection: node-18, ubuntu-latest'
```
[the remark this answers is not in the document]
Shinigami92: your suggestion makes it a bit better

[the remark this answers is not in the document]
ST-DDT: Then lets go for it.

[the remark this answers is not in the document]
sunadoi: Thank you for your suggestion!

.github/workflows/ci.yml
sunadoi: OK, I'm thinking about doing that using [github-script](https://github.com/actions/github-script),  and prompt the user to run `generate:locales` or `generate:api-docs` command via the comment.
This comment will be deleted if the CI passes.

.github/commentCodeGeneration.js
```diff
@@ -0,0 +1,32 @@
+module.exports = async ({ github, context, isSuccess }) => {
+  const { data: comments } = await github.rest.issues.listComments({
+    owner: context.repo.owner,
+    repo: context.repo.repo,
+    issue_number: context.issue.number,
+  });
+
+  const body = `Uncommitted changes were detected after runnning <code>generate</code> command.\nPlease run <code>pnpm run generate:locales</code> or <code>pnpm run generate:api-docs</code> to generate/update the related files, and commit them.`;
```
sunadoi, line 8: I'm not sure if this error message is correct.
If you have a better message, I would appreciate it if you could let me know.

ST-DDT:
I tried to run it in my fork and it doesn't work:
https://github.com/ST-DDT/faker/actions/runs/3175539763/jobs/5173706263#step:8:17

You might need the PR write permission for that action.
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

sunadoi:
> Please assign it for the lowest possible scope (e.g. only for this single job).

I gave permissions `pull-requests: write` in this job only, and it seems to work in my fork.
https://github.com/sunadoi/faker/actions/runs/3179609661/jobs/5182304903

ST-DDT (dismissed):

.github/commentCodeGeneration.js
```diff
@@ -0,0 +1,32 @@
+module.exports = async ({ github, context, isSuccess }) => {
```
Shinigami92, line 1: I have two minor concerns here:
1. If we just have js instead of ts, could we at least use JSDoc typings for these?
2. Not sure if we should store the script here in `<project-root>/.github/` or somewhere else like in `scripts` folder or if possible `.github/workflows/`

sunadoi: > If we just have js instead of ts, could we at least use JSDoc typings for these?

I added a JSDoc.
The purpose, github-script URL, and `@param` are written.

> Not sure if we should store the script here in `<project-root>/.github/` or somewhere else like in `scripts` folder or if possible `.github/workflows/`

I stored the script in `.github/workflows/`.

.github/workflows/commentCodeGeneration.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Notifies uncommitted changes generated by the `generate` command.
```
ST-DDT, line 2:
```suggestion
 * Notifies the PR about uncommitted changes generated by the `generate:*` commands.
```

.github/workflows/commentCodeGeneration.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param isSuccess A boolean indicating whether the workflow was successful
+ */
+module.exports = async ({ github, context, isSuccess }) => {
+  const { data: comments } = await github.rest.issues.listComments({
+    owner: context.repo.owner,
+    repo: context.repo.repo,
+    issue_number: context.issue.number,
+  });
+
+  const body = `Uncommitted changes were detected after runnning <code>generate</code> command.\nPlease run <code>pnpm run generate:locales</code> or <code>pnpm run generate:api-docs</code> to generate/update the related files, and commit them.`;
```
ST-DDT, line 18:
```suggestion
  const body = `Uncommitted changes were detected after runnning <code>generate:*</code> commands.\nPlease run <code>pnpm run generate:locales</code> and <code>pnpm run generate:api-docs</code> to generate/update the related files, and commit them.`;
```

ST-DDT (commented):
@Shinigami92 Is these Jsdocs what you intended with your request?
Maybe you could give an example of what you would like to have.

.github/workflows/commentCodeGeneration.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Notifies uncommitted changes generated by the `generate` command.
+ *
+ * This script is used by github-script
+ * https://github.com/actions/github-script
+ *
+ * @param github A pre-authenticated octokit/rest.js client with pagination plugins
+ * @param context An object containing the context of the workflow run
+ * @param isSuccess A boolean indicating whether the workflow was successful
```
Shinigami92, line 9: Can we add types here via JSDoc types?

ST-DDT: Please add an example so that it is clearer what you are referring to.

Shinigami92: https://jsdoc.app/tags-type.html

Shinigami92: Sorry I didn't had much time

So I expect something like 
`@param {import('the-GH-type-lib').GitHub} github Description text.`

You can see that it works when you hover over the param names.

sunadoi: Thank you for your example.
Now, I clearly understand what to do.

We need install `@actions/github` to add types via JSDoc types.
I could confirmed that params are typed when hovering the param names.
ref: https://github.com/sunadoi/faker/pull/4/commits/9dac32b170dfbab4c7a8481f3fbc86af84ce7919

On the other hand, I noticed that this library is written in ts, so we can write the script in ts file.
In that case, we need transpile before running the github-script.

```
- name: Transpile script ts file
  run: pnpm exec tsc .github/workflows/commentCodeGeneration.ts --outDir .github/workflows
```

It seems to work well.
ref: https://github.com/sunadoi/faker/pull/5/files

Which is better?

Shinigami92: IMO clearly the TS version is better for us! 👍
For our other scripts we already use `tsx`.

sunadoi: Fixed to TS version!

package.json
```diff
@@ -93,6 +93,7 @@
     ]
   },
   "devDependencies": {
+    "@actions/github": "^5.1.1",
```
Shinigami92, line 96: Please use `~` instead of `^` as this makes it more robust and long term stable
After that you need to regenerate the lock file

Shinigami92 (approved):
Beside that, I think we are good to go :+1: 
So I will already approve so other maintainers know that it counts after you fixed the dep version

.github/workflows/commentCodeGeneration.ts
```diff
@@ -0,0 +1,48 @@
+import { context as ctx, GitHub } from '@actions/github/lib/utils';
```
ST-DDT, line 1:
```suggestion
import type { context as ctx, GitHub } from '@actions/github/lib/utils';
```

ST-DDT (commented):
Do we have to `.gitginore` the transpiled `.js` script?
- If yes, why is the script not failing?
- If not, why not?

ST-DDT:
> Do we have to `.gitginore` the transpiled `.js` script?
> 
>     * If yes, why is the script not failing?
> 
>     * If not, why not?

Ah I see, because it is transpiled only after the check.
Please add all `.js` files in the workflows folder to the gitignore anyway, in case somebody transpiles it locally.

sunadoi:
> (I plan to use your script as a template to automate some more tasks 😻 )

That sounds nice!
I'm so happy to hear that.😘

ST-DDT (approved):

Shinigami92 (approved):

ST-DDT:
@sunadoi It looks like this breaks the main branch.
Could you move the job to a new file e.g. `pr.yml` and run it only for PRs and not main?

PR #1428, opened 2022-10-11
test: extend vitest with (not) toContainDuplicates

Shinigami92:
Improving some tests with https://vitest.dev/api/#expect-extend

Not sure if there is even a better way instead of importing the ts file all over again

### DIFF
--- a/test/vitest-extensions.ts
+++ b/test/vitest-extensions.ts
@@ -1,20 +1,25 @@
 import { expect } from 'vitest';
 
 expect.extend({
-  toBeUnique<T>(received: T[]) {
+  toContainDuplicates<T>(received: T[]) {
+    const { isNot } = this;
+
     const uniques = new Set(received);
     const duplications = received.filter((entry) => !uniques.delete(entry));
     const uniqueDuplication = [...new Set(duplications)];
 
     return {
-      pass: uniqueDuplication.length === 0,
-      message: () => `Duplicated values are: [${uniqueDuplication.join(', ')}]`,
+      pass: uniqueDuplication.length !== 0,
+      message: () =>
+        isNot
+          ? `Duplicated values are [${uniqueDuplication.join(', ')}]`
+          : `No duplicate values in [${received.join(', ')}]`,
     };
   },
 });
 
 interface CustomMatchers {
-  toBeUnique(): void;
+  toContainDuplicates(): void;
 }
 
 declare global {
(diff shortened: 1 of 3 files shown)


ST-DDT (approved):

test/vitest-extensions.ts
```diff
@@ -0,0 +1,28 @@
+import { expect } from 'vitest';
+
+expect.extend({
+  toBeUnique<T>(received: T[]) {
```
ST-DDT, line 4: Maybe `toNotContainDuplicates`?

Shinigami92: I'm kinda against to name it `...Not...` as the usual syntax is more like using `expect(x).not.toBe...`
But using `expect(x).not.toBeContainDuplicates()` is also not that great :thinking: (it's to long?!)

xDivisionByZerox: `toHaveUniqueEntries` (or `toHaveUniqueElements` if those are more preferrable). Works with `not`: 
`expect(array).not.toHaveUniqueEnties()`.

test/vitest-extensions.ts
```diff
@@ -0,0 +1,28 @@
...
+    const uniques = new Set(received);
+    const duplications = received.filter((entry) => !uniques.delete(entry));
+    const uniqueDuplication = [...new Set(duplications)];
+
+    return {
+      pass: uniqueDuplication.length === 0,
+      message: () => `Duplicated values are: [${uniqueDuplication.join(', ')}]`,
```
xDivisionByZerox, line 11: What does an error message of this matcher looks like? Is the matchers name name included? Otherwise it isn't immediately clear that an element was supposed to have unique elements.

test/vitest-extensions.ts
Shinigami92: I'm good with `toHaveUniqueEntries` :+1: 
Does someone want to change it? Otherwise I'm available in 1-2 hours to change it

Shinigami92: For now we committed us on `.not.toContainDuplicates` for better readability
I personally expect that in the future we might get a matcher from vitest directly that is not maintained on our side anymore but their community

test/vitest-extensions.ts
Shinigami92: We experimented a bit around with it, and this is one example:

image

The other case around is like the same with the opposite text

xDivisionByZerox (approved):

PR #1407, opened 2022-10-02
feat: lorem null response fix

ST-DDT:
Fixes #1131

- Extended and exported the filterWordListByLength to add support for length ranges and fallback strategies
- Use the improved filterWordListByLength function for all `word.*` and `lorem.word` methods

PR #1438, opened 2022-10-13
docs: link to next docs and vice versa

ST-DDT:
Currently the https://next.fakerjs.dev/ page does not show any hints that it isn't the current version.

grafik

This PR changes the version number to `Next` and adds a link to the latest released page.

grafik


docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,12 @@
 import { version } from '../../package.json';
 
-export const currentVersion = `v${version}`;
+const isNext = true;
```
Shinigami92: Is it in any way possible that this can be archived dynamically?

ST-DDT: I would like to do something like:

```ts
const hostname = window.location.hostname;
const isNext = hostname === 'next.fakerjs.dev' || hostname === 'localhost';
```

But that isn't possible. Any idea on how to achieve that?
Maybe by using the branch name?

ST-DDT: I added a test that ensures we don't forget it in v8.0.

Shinigami92: I don't know how overkill this is, but I assume the vitepress config is running in nodejs and not at runtime in browser
So we might use anything we want to, e.g. https://stackoverflow.com/questions/62225567/get-current-git-branch-with-node-js

matthewmayer:
This seems really subtle unless you really know what to look for. Would it be possible to add something more blatant like a top banner? (rough mockup attached)

Screenshot 2022-10-15 at 20 52 35

ST-DDT:
@matthewmayer Good point. This PR is more about linking the different versions together/making them accessible from our main website, so I created a new issue (#1456) for that to keep track of it.

docs/.vitepress/versions.ts
ST-DDT: Done

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,25 @@
...
+const isNext = !branchName.match(/^v\d+$/);
+
+export const currentVersion = isNext ? NEXT_TEXT : `v${version}`;
+
+const nextVersion = { version: NEXT_TEXT, link: 'https://next.fakerjs.dev/' };
+const latestVersion = { version: `v${version}`, link: 'https://fakerjs.dev/' };
 
 export const oldVersions = [
+  isNext ? latestVersion : nextVersion,
   { version: 'v6.3.1', link: 'https://v6.fakerjs.dev/' },
```
Shinigami92: Do we already also need to add v7 as we are in v8 now? (I mean in a separate PR)

```suggestion
  isNext ? latestVersion : nextVersion,
  { version: 'v7.6.0', link: 'https://v7.fakerjs.dev/' },
  { version: 'v6.3.1', link: 'https://v6.fakerjs.dev/' },
```

I think this is needed at the time when we switch branches in that way that v7.fakerjs.dev already exists :thinking: 

---

Or can we also automate this behavior somehow as well? :eyes:

ST-DDT: This is not possible, because then once we release v8 it will point to v8, 
but the main branch cannot be used with netlify/and it would still point to the same version.
The test ensures that we don't forget to add v7 once v8 is released.

As for automating this, IMO this is not worth it since it only changes very rarely (once every few months).

Shinigami92: Partly I see what you mean
But I see an issue if we want to fully automate the release process :thinking: But might be solved by just using the right workflow
We might need to discuss about that in person

ST-DDT: The release process could add the required changes, since it also decides the version and the tag that the docs will reference.

Shinigami92: release-it supports hooks, so yeah we can add scripts how ever we want to :thinking: 

But yeah, the more I think about an automated release process, the more I want to have a PR that will be created before the actual release happens, so we have a chance to review it (like we currently do)

ST-DDT: After a nights worth of thoughts, I think I found an easy way to automate this in every docs build.
Let me check first, whether my idea works in reality as well.

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,25 @@
+import { execSync } from 'child_process';
```
Shinigami92: I think `child_process` is one of the core node packages
If possible it is now and always recommended to use the node prefix in such cases

```suggestion
import { execSync } from 'node:child_process';
```

Shinigami92 (changes requested):

docs/.vitepress/versions.ts
ST-DDT: Done

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,88 @@
...
+    return Number(version.split('.')[0].substring(1));
+  } else {
+    return NaN;
+  }
+}
+
+export function pickLatest(version1: string, version2: string): string {
+  const parts1 = version1.split('-', 2);
+  const parts2 = version2.split('-', 2);
+  let compare = parts1[0].localeCompare(parts2[0]);
```
Shinigami92: The npm package [`semver`](https://www.npmjs.com/package/semver) is a widely used one and when you use `pnpm why semver` you can already see that we have it in our dev-dependencies.
So I think it's okay to explicitly add it to the dev-deps and use it and remove complexity out of our code by doing that.

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,88 @@
+import { execSync } from 'node:child_process';
 import { version } from '../../package.json';
 
-export const currentVersion = `v${version}`;
+const NEXT_TEXT = 'Next';
```
Shinigami92: Lets use `next` instead of `Next` because this is the tag of the potential npm package and the branch name and and and ^^

docs/.vitepress/versions.ts
ST-DDT: Done

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,58 @@
...
+    execSync('git branch --show-current').toString('utf8').trim() || NEXT_TEXT
+  );
+}
+
+function readOtherLatestReleaseTagNames(): string[] {
+  const currentMajorVersion = semver.major(version);
+  const latestReleaseTagNames = execSync('git tag -l')
+    .toString('utf8')
+    .split('\n')
+    .filter((tagName) => tagName.startsWith('v'))
```
Shinigami92: I think it's more safe to use `valid` here

```suggestion
    .filter((tag) => semver.valid(tag))
```

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,58 @@
...
+      ) {
+        if (/\d{2,}/.test(tag)) {
+          // if this happens, we have to update the version compare logic
+          throw new Error(`Unsupported tag name: ${tag}`);
+        }
+        acc[`v${majorVersion}`] = pickLatest(acc[majorVersion] || tag, tag);
+      }
+      return acc;
+    }, {});
+  return Object.values(latestReleaseTagNames).sort(semver.rcompare);
```
Shinigami92: I think we can improve it by just grouping all the versions and then use `semver.maxSatisfying`.
This is IMO more understandable in perspective of the code-flow and it might also safe us some nano seconds or performance :stuck_out_tongue_winking_eye: 

```ts
      .reduce<Record<number, string[]>>((acc, tag) => {
        const majorVersion = semver.major(tag);

        if (
          // Only keep track of versions the new Faker Team is responsible for
          majorVersion < 6 ||
          // Ignore current branch as we are already on that branch
          majorVersion === currentMajorVersion
        ) {
          return acc;
        }

        acc[majorVersion] ||= [];
        acc[majorVersion].push(tag);
        return acc;
      }, {});

    const versions = Object.entries(latestReleaseTagNames).map(
      ([major, tags]) => semver.maxSatisfying(tags, `^${major}`)
    );

    return versions;
```

I'm still not quite satisfied with the name `latestReleaseTagNames` :thinking: 
Do you have any better idea?

test/docs/versions.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  currentVersion,
+  NEXT_TEXT,
+  oldVersions,
+} from '../../docs/.vitepress/versions';
+import { version } from '../../package.json';
+
+describe('docs versions', () => {
+  describe('oldVersions', () => {
+    it('should have a complete set of oldVersions', () => {
+      console.log(oldVersions);
```
Shinigami92: This can be removed, can it?

Shinigami92 (changes requested):

ST-DDT:
I created some preview deployments:

- fakerjs.dev: https://634c84829385500009ecdd35--melodious-druid-b57668.netlify.app/ 
  (Please ignore the branchname in the nav bar, this is caused by the way I setup the netlify deployment)
- next.fakerjs.dev: https://634c85fd38583a0009a0502d--melodious-druid-b57668.netlify.app/
- v7.fakerjs.dev: https://634c85871819c70008aea012--melodious-druid-b57668.netlify.app/

ST-DDT:
Not sure whether `latest` or `next` should be first:
grafik

If next is the top one, then it will be in the "correct" version order.

Shinigami92:
Looked into the preview deployments and I like them all 👍 
Good job

docs/.vitepress/versions.ts
```diff
@@ -1,7 +1,58 @@
...
+  const latestReleaseTagNames = execSync('git tag -l')
+    .toString('utf8')
+    .split('\n')
+    .filter((tag) => semver.valid(tag))
+    .reduce<Record<number, string[]>>((acc, tag) => {
+      const majorVersion = semver.major(tag);
+      // Only consider tags for our deployed website versions,
+      // excluding the current major version.
+      if (majorVersion >= 6 && majorVersion !== currentMajorVersion) {
+        (acc[majorVersion] = acc[majorVersion] || []).push(tag);
```
Shinigami92:
```suggestion
        (acc[majorVersion] ||= []).push(tag);
```

Maybe even `??=` 👀

Shinigami92 (commented):
The code is now MUCH better 👍 
Will approve after rebased

docs/.vitepress/versions.ts
ST-DDT: It isn't supported. You told me the same just a few weeks ago.
https://github.com/faker-js/faker/pull/1407#discussion_r985299537
[`09d11f8` (#1407)](https://github.com/faker-js/faker/pull/1407/commits/09d11f8e169b0bf16f0c0105dfc800325afff84f#diff-43806c9d91df883f547b50d72ffc58f389e58df082afa98825ba79217e7bf3ccR17)

ST-DDT: I can use `??` though.

```suggestion
        (acc[majorVersion] = acc[majorVersion] ?? []).push(tag);
```

Shinigami92: > It isn't supported. You told me the same just a few weeks ago. [#1407 (comment)](https://github.com/faker-js/faker/pull/1407#discussion_r985299537) [`09d11f8` (#1407)](https://github.com/faker-js/faker/pull/1407/commits/09d11f8e169b0bf16f0c0105dfc800325afff84f#diff-43806c9d91df883f547b50d72ffc58f389e58df082afa98825ba79217e7bf3ccR17)

The small difference here is, that it is s script on our side and doesn't need to be bundled by esbuild for the end-user

Shinigami92 (dismissed):

Shinigami92 (dismissed):

docs/.vitepress/versions.ts
ST-DDT: Netlify seems to support it.

ST-DDT:
It doesn't pass our CI, so I reverted it.

Shinigami92:
🤣 no f*cking way, out TESTS are blocking it 🤣 
Oh my good, okay, no new syntax for us I guess 😭 That's brutal 😖

Shinigami92 (approved):

ST-DDT:
Well, we could omit the tests... 🤣

Shinigami92:
> Well, we could omit the tests... rofl

Na, lets just upgrade when we move to `^Node16`

ejcheng (approved):

PR #1444, opened 2022-10-13
refactor!: make mersenne internal

### LINKED ISSUE #1348: Mark Mersenne module as internal and hide it from ts and the api-docs

The mersenne module is an internal module that should be called by external users/libraries.

### Suggested solution

- Mark the mersenne module as internal
- Hide the field references/mark them as internal or private
- Hide the Mersenne module from the api-docs

### Alternative

- Remove the mersenne module/merge it with the main Faker class

### Additional context

_No response_

Shinigami92:
closes #1348


src/internal/mersenne/mersenne.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  seed(seed: number | number[]): void;
+}
+
+/**
+ * Generate seed based random numbers.
+ *
+ * @internal
+ */
+export default function mersenne(): Mersenne {
```
xDivisionByZerox: Why did you change the implementation itself? If we have classes all over the place, why use a function here? I know then you basically did the same, but syntactically it differs strongly from the rest of the implementations in the code base.

Preferable I'd like to have the refactor itself in a different PR so it can be discussed individually regardless of changes about making this internal.

ST-DDT: This combines changes for v8.0 with ones that could be considered `v8.2+`. IMO this isn't that much of a problem, but it might be clearer in the history if you do it in two separate PRs.

Shinigami92: @xDivisionByZerox Please tell me what I'm allowed to bundle in a PR together and what should be done in separate PRs

1. Mark mersenne and everything coupled with it as `@internal`
2. Move the module out to the internal folder as it is now completely internal
3. Rewrite mersenne to be less boilerplaty
4. Rewrite mersenne's function to be less verbose (simplify) out code

And I might have missed a mirco-step 🤷 

Side note: I would like not to hang on this for a full week as I also have other tasks and the end result would be the same IMO
The only "improved result" would be history, so what about not using a squash-merge but a normal merge 🤔 would that be an option for you?

ST-DDT: I would separate 1 into its own PR and keep the rest together.
As for not using squash commits: No.

Shinigami92: But I want to hear the decision from @xDivisionByZerox as she raised the discussion 
I would like to prevent double or triple work on my side and therefore make it directly how she thought about it

Shinigami92: Team decision:
- split [1, 2] into one PR
- and [3, 4] into one PR

src/faker.ts
```diff
@@ -79,10 +79,9 @@ export class Faker {
   readonly fake: FakeModule['fake'] = new FakeModule(this).fake;
   readonly unique: UniqueModule['unique'] = new UniqueModule(this).unique;
 
-  /**
-   * @deprecated Internal. Use faker.datatype.number() or faker.seed() instead.
-   */
-  readonly mersenne: MersenneModule = new MersenneModule();
+  /** @internal */
+  private readonly _mersenne: MersenneModule = new MersenneModule();
```
xDivisionByZerox, line 78: Since this is really awkward to deal with for our modules, wouldn't it be better to export a global `MersenneModule` instance from `'./internal/mersenne/mersenne'`?

Shinigami92: That's not possible, due to the `mersenne` generator depends on the faker instance.
Each Faker instance has its own `mersenne`.

ST-DDT (dismissed):

src/faker.ts
xDivisionByZerox: I see.

xDivisionByZerox (approved):

ejcheng (approved):

ST-DDT (approved):

PR #1440, opened 2022-10-13
refactor!: cleanup deprecations

Shinigami92:
closes #1439

PR #1445, opened 2022-10-14
refactor(person)!: rename name module

### LINKED ISSUE #1343: Rename Name module to Person

The name module contains more than just names.

### Suggested solution

- Rename the module to Person so it is clear that this contains person related methods.
- The Name module is degraded to an alias for two major versions

### Alternative

_No response_

### Additional context

_No response_

Shinigami92:
closes #1343


Shinigami92:
Blocked by:
- #1440

Shinigami92:
@ST-DDT I'm not sure if we want to support `faker.helpers.fake('{{name.firstName}} or {{name.first_name}}')`
It might be really hard for us to maintain that in the long run
We also would need to write extra code to support that but in a way so we don't duplicate the name definitions
So you have any ideas? Or should we just make a hard breaking change?

Shinigami92:
Okay, this is now in a _done_ / _ready for review_ state, but I want to first merge the cleanup deprecations as of that will reduce some changes.
So until then, this PR will stay in _draft_.

src/faker.ts
```diff
@@ -103,13 +104,17 @@ export class Faker {
...
+  readonly person: PersonModule = new PersonModule(this);
   readonly phone: PhoneModule = new PhoneModule(this);
   readonly science: ScienceModule = new ScienceModule(this);
   readonly system: SystemModule = new SystemModule(this);
   readonly vehicle: VehicleModule = new VehicleModule(this);
   readonly word: WordModule = new WordModule(this);
 
+  // Aliases
+  /** @deprecated Use {@link person} instead */
+  readonly name: NameModule = this.person;
```
ST-DDT: Use a getter with a call to deprecated helper method here.

src/locales/de/person/name.ts
```diff
@@ -0,0 +1,7 @@
+export default [
+  '{{prefix}} {{first_name}} {{last_name}}',
+  '{{first_name}} {{nobility_title_prefix}} {{last_name}}',
+  '{{first_name}} {{last_name}}',
+  '{{male_first_name}} {{last_name}}',
+  '{{female_first_name}} {{last_name}}',
```
ST-DDT: This file got corrupted: `person.` is missing.
Did you convert these files by hand?

ST-DDT (changes requested):

src/faker.ts
Shinigami92: You are right :+1:

src/locales/de/person/name.ts
Shinigami92: Nice catch, thx :+1:

ST-DDT (dismissed):

ST-DDT:
I don't know why this works, but it works.

test/helpers.spec.ts
```diff
@@ -611,6 +611,15 @@ describe('helpers', () => {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           delete (faker.random as any).special;
         });
+
+        it('should support deprecated aliases', () => {
+          expect(faker.definitions.person.first_name).toContain(
+            faker.helpers.fake('{{person.first_name}}')
```
ST-DDT:
```suggestion
            faker.helpers.fake('{{name.first_name}}')
```

test/helpers.spec.ts
```diff
@@ -611,6 +611,15 @@ describe('helpers', () => {
...
+          );
+          expect(faker.definitions.person.first_name).toContain(
+            faker.helpers.fake('{{person.firstName}}')
```
ST-DDT:
```suggestion
            faker.helpers.fake('{{name.firstName}}')
```

ST-DDT (dismissed):

matthewmayer:
in addition to the new faker.person tests, why not keep the existing tests for the faker.name module ?
```
test/__snapshots__/name.spec.ts.snap
test/name.spec.ts
```

That will help ensure no unexpected regressions/breaking changes.

Shinigami92:
> in addition to the new faker.person tests, why not keep the existing tests for the faker.name module ?
> 
> ```
> test/__snapshots__/name.spec.ts.snap
> test/name.spec.ts
> ```
> 
> That will help ensure no unexpected regressions/breaking changes.

Because it is just an alias, behind it there is exactly the same code, so no need to test it twice.

xDivisionByZerox:
I'm a bit lost, shouldn't this follow our normal [deprecation workflow](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#deprecation-workflow)? As now a user has no time-window to migrate, but have to rework their codebase. Is this desired even if this will be in a breaking change?

scripts/apidoc/moduleMethods.ts
```diff
@@ -38,7 +38,7 @@ export function processModuleMethods(project: ProjectReflection): PageIndex {
 }
 
 export function extractModuleName(module: DeclarationReflection): string {
-  return module.name.replace(/Module$/, '');
+  return (module.escapedName || module.name).replace(/Module$/, '');
```
xDivisionByZerox: This change doesn't seem related to this PR. If so, how?

xDivisionByZerox:
> I'm a bit lost, shouldn't this follow our normal [deprecation workflow](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#deprecation-workflow)? As now a user has no time-window to migrate, but have to rework their codebase. Is this desired even if this will be in a breaking change?

Sorry, too many files. Didn't saw the getter in the Faker class. So if this follows the typical deprecation workflow, we need to keep the tests as @matthewmayer suggested. Even is if is just an alias, how do you confirm that it is that way? The test would also catch breaking changes in the `NameModule` which were designed to only affect the `PersonModule`.

src/index.ts
```diff
@@ -18,8 +18,12 @@ export type {
   LocaleDefinition,
   LoremDefinitions,
   MusicDefinitions,
-  NameDefinitions,
-  NameTitleDefinitions,
+  /** @deprecated Use PersonDefinitions instead */
+  PersonDefinitions as NameDefinitions,
```
xDivisionByZerox, line 22: Exporting it like this means that any change in `PersonDefinition` directly influences the old `NameDefinition` type. This means that no changes to this type can be made in the the entirety of v8 (?).

If we for example add `JobDefinitions` to `PersonDefinitions` I could do the following:

```ts
interface PersonDefinitions {
  // current declaration
  jobNames: string[];
}

const personDefs: NameDefinitions = {
  // [...]
  jobNames: [...],
}
```

Is this desired?

src/locales/ar/name/name.ts
```diff
@@ -1,5 +0,0 @@
-export default [
```
xDivisionByZerox, line 1: Did you deleted this file by accident instead of renaming it?

xDivisionByZerox: Ah ok I see. Git cant track that this is the same as `src/locales/ar/person/name.ts` and only being renamed.

src/index.ts
Shinigami92: > Is this desired?

Yes

scripts/apidoc/moduleMethods.ts
ST-DDT: The `name` attribute is derived/calculated by typedoc.
IMO it is a bug from typedoc, that it uses the alias as the main name (because it is exported first).
If we disable our formatter/linter and swap the order of our imports/exports, then typedoc picks the correct name.
To avoid that issue we use the name that typescript assigns to that type (`escapedName`).
(I don't know why it is called `escapedName` though.)
The `escapedName` can be null (says typescript), but it never is, so basically we always use the `escapedName` instead of `name`. The code is effectively like this:

````ts
return module.escapedName.replace(/Module$/, '');
````

This change is required for the docs to work/generate the correct api docs.

File: src/index.ts (hunk 6 above)
Comment (ST-DDT) [reply src/index.ts:22]: > Exporting it like this means that any change in PersonDefinition directly influences the old NameDefinition type. 

Yes, we renamed the `NameModule` to `PersonModule`.
`faker.name` returns the `PersonModule`. There is no longer a `NameModule`.

> This means that no changes to this type can be made in the the entirety of v8 (?).

This is not true, because like in any other module non breaking changes can be made to it.
We cannot do breaking changes after v8.0 is released till v9.0, but that would still be the case if we didn't rename the module.

> If we for example add JobDefinitions to PersonDefinitions I could do the following:

That is true, but if you try to add it then ts wouldn't suggest the name module anymore.
You can still add it, like you can add any other information, but faker won't use them.
If you try to access it with faker you will get a notice that you should use `faker.definitions.person` instead.

https://github.com/faker-js/faker/blob/60a47793117b5c69f3d75a0cd5b8ffb4e21e1d46/src/faker.ts#L175-L184

Comment (xDivisionByZerox) [reply src/index.ts:22]: Ok, gotcha. Thank you for clearification.

File: scripts/apidoc/moduleMethods.ts (hunk 5 above)
Comment (xDivisionByZerox) [reply scripts/apidoc/moduleMethods.ts]: I still don't see how this is related (and only now required) to renaming the `NameModule` to `PersonModule`. 
So, if the renaming would be reverted this line (<s>`return module.name.replace(/Module$/, '');`</s>`return (module.escapedName || module.name).replace(/Module$/, '');`) would break?

**Comment by ST-DDT** [comment]
> So if this follows the typical deprecation workflow, we need to keep the tests as @matthewmayer suggested. Even is if is just an alias, how do you confirm that it is that way? The test would also catch breaking changes in the `NameModule` which were designed to only affect the `PersonModule`.

I'm not sure whether I get you right.
If we duplicate the tests for `faker.name`, then this would only prevent breaking changes if we freeze the name tests until after v8.0.0.
However, the same applies to the PersonModule.
If we treat it like an existing module and only change things after a deprecation phase, then there is no need for the extra tests.

File: scripts/apidoc/moduleMethods.ts (hunk 5 above)
Comment (xDivisionByZerox) [reply scripts/apidoc/moduleMethods.ts]: I also had a deeper look into that. `escapedName` is marked as `@internal`, should really use it then?

Comment (ST-DDT) [reply scripts/apidoc/moduleMethods.ts]: > I still don't see how this is related (and only now required) to renaming the `NameModule` to `PersonModule`.

This is required because typedoc doesn't assign the correct name.
The change is only relevant because we created an alias for the first time.
I don't consider this permanent code, maybe we can add a comment stating we should remove it later on.

> So, if the renaming would be reverted this line (return module.name.replace(/Module$/, '');return (module.escapedName || module.name).replace(/Module$/, '');) would break?

No, it doesn't. If you flip the imports/exports, then also `name` would show the correct value.

> I also had a deeper look into that. escapedName is marked as @internal, should really use it then?

Yes, the field is internal, but I couldn't find another way to achieve the same result.
(Aside from droping typedoc and using the typescript AST directly)
If you have an idea for a different fix, please tell me.
When I have some time, I will write a minimal reproduction and report this issue to typedoc.

We could write a test, to ensure that `faker.name` doesn't show up and `faker.person` does.

I don't think reversing the import/export order (and disabling the formatter/linter for that) is an acceptable alternative.

Comment (ST-DDT) [reply scripts/apidoc/moduleMethods.ts]: Should we write instead:

````ts
const { name } = module;
if (name === 'NameModule') {
    return 'Person';
} else {
    return name.replace(/Module$/, '');
}
````

Comment (Shinigami92) [reply scripts/apidoc/moduleMethods.ts]: This would be a potential solution, but looking forward to address->location rename we would need it for every alias
Your actual solution works and handle all the cases dynamically

Comment (xDivisionByZerox) [reply scripts/apidoc/moduleMethods.ts]: > Should we write instead:
> 
> ```ts
> const { name } = module;
> if (name === 'NameModule') {
>     return 'Person';
> } else {
>     return name.replace(/Module$/, '');
> }
> ```

The is makes in much easer for me to understand whats happening here.

Comment (ST-DDT) [reply scripts/apidoc/moduleMethods.ts]: > This would be a potential solution, but looking forward to address->location rename we would need it for every alias Your actual solution works and handle all the cases dynamically

Sure, but we have it for the definitions proxy already.
See also my (future) comment below.

**Comment by ST-DDT** [comment]
TODO: Create an issue to remove the special handling in v10.0 after this is merged (to get the final line numbers and commit references).

**Review by ST-DDT — APPROVED** [review]

File: test/helpers.spec.ts [hunk 8]
```diff
@@ -611,6 +611,15 @@ describe('helpers', () => {
...
           delete (faker.random as any).special;
         });
+
+        it('should support deprecated aliases', () => {
+          expect(faker.definitions.person.first_name).toContain(
+            faker.helpers.fake('{{name.first_name}}')
+          );
+          expect(faker.definitions.person.first_name).toContain(
+            faker.helpers.fake('{{name.firstName}}')
+          );
```
xDivisionByZerox, line 621: The seems to be duplicated.

xDivisionByZerox: ok... `name.first_name` !== `name.firstName`. 600 files are too much to keep your head working ^^

Shinigami92: first_name vs firstName

xDivisionByZerox (approved):

ejcheng (approved):
LGTM :)

Issue #627, opened 2022-03-16
Get the ".last()" value generated of a module


wilker-dotgroup:
### Clear and concise description of the problem

As a developer using faker I want to use some method e.g. `faker.internet.url.last()` to get the latest value generated for this call.
If no previous value was generated, this method (or prop) executes like the same `faker.internet.url()` call.
So that helps me to avoid create an auxiliary var just to keep this value to use again in another field in the same scope.

### Suggested solution

For any faker api module, the "`.last()`" returns the latest value generated, regardless of how the module was called.
If undefined, perform the same of no parameter call,
e.g.
     ~`faker.address.city.last() // undefined`~
     `faker.address.city.last() // same result as faker.address.city()`

Shinigami92:
Interesting idea, but why don't you just safe it in a local var in your code?

xDivisionByZerox:
I love the idea of a history. But I don't see the responsibility on our side. Since our responsibility is to generate fake data. Not keep track of returned results IMO.

wilker-dotgroup:
Well, I use some other generators and I like this approach in one of them, that is to get the last password generated. When using faker I missed something like this but there is no problem to use a var to store it.
In my code, I'm using faker to compound some POST and PUT requests and some fields need the same values. This pkg mounts the request chaining the .field('foo','bar') call, and I thought it would be better if some like .last() could be used to avoid the var "far way" from where is (re)used.

ST-DDT:
How about a `fork()` method? That way you can achieve the same.
The fork would create a new independent faker instance with the exact same values (including the seed).
That way, you can reproduce the exact same behavior as `last()`. It can even be used for `last(number)`.

````ts
faker.fork().someMethod() // 'FooBar42'
faker.someMethod() // 'FooBar42'
faker.someMethod() // 'FooBar43'
````

Another alternative would be using a seed `snapshot()` and `restore()` mechanism.

Shinigami92:
> `fork()`
> `snapshot()` and `restore()`.

This actually sounds much more helpful 🤔 
Because with that you could solve some even more complex tasks _I think_

Shinigami92:
We will try to plan something out, but for now I think this is something to target in the _far_ future, so we can focus on the active milestones right now.
In the meanwhile, you are still able safe the value in a local var as suggested.
I personally do not want to quickly implement something, but design a well thought out draft for this feature.

ST-DDT:
**Team Decision**

[...]

PR #1447, opened 2022-10-14
refactor(mersenne): rewrite internal mersenne

### LINKED ISSUE #1348: Mark Mersenne module as internal and hide it from ts and the api-docs

Shinigami92:
Splitted from #1444 

### Changes:
- This changes the Mersenne module to be not anymore a class but a light weight function
- Also the code was rewritten so it is more easily maintainable
- Additionally seed does not throw anymore a FakerError on invalid input -> this has also the benefit of that mersenne is now fully decoupled from faker itself and could theoretically be swapped with a generator, but this could be done in another PR


src/internal/mersenne/mersenne.ts
```diff
@@ -1,81 +1,64 @@
...
+ *
+ * @internal
+ */
+export default function mersenne(): Mersenne {
+  const twister = new Twister();
+
+  twister.initGenrand(Math.ceil(Math.random() * Number.MAX_SAFE_INTEGER));
+
+  return {
+    next(max = 32768, min = 0): number {
```
ST-DDT: Please remove the defaults, as this method is purely internal and we will always set them both.
Maybe also swap min and max.

ST-DDT:
IMO we could remove the mersenneFn wrapper and integrate these functions in the faker class.
We would then ask faker for the next seeded value instead of accessing faker's private mersenne property and doing things with it.

xDivisionByZerox:
> IMO we could remove the mersenneFn wrapper and integrate these functions in the faker class. We would then ask faker for the next seeded value instead of accessing faker's private mersenne property and doing things with it.

I think that's the wrong way around. Currently, Faker is tightly coupled with the Mersenne generator. A way better approach would be to require a generator (probably denied from an interface) on construction, thus making the seed generation dependency injectable (follow the dependency inversion principle). This way I could build myself a Faker instance with the underlying generator I want.

```ts
interface SeedGenerator {
  next(): number;
}

class Faker {
  // properties
  constructor(options: {
    // current faker options
    generator: SeedGenerator; // could default to a mersenne implementation
  }) { 
    // initialization
  }
}

const instance = new Faker({
  ...defaultFakerOptions,
  generator: { next: () => 50 },
]);
```

In my eyes that's a much better approach since it allows for a higher level of abstraction and customization.

src/internal/mersenne/mersenne.ts
```diff
@@ -1,81 +1,64 @@
...
-   * @throws If the seed is not a `number[]`.
-   *
-   * @since 5.5.0
-   */
-  seed_array(A: number[]): void {
-    if (typeof A !== 'object') {
       throw new FakerError(
-        `seed_array(A) must take array of numbers; is ${typeof A}`
+        `seed must take numeric argument(s); is ${typeof seed}`
       );
```
xDivisionByZerox: Do we really need to throw an error here, as the function will return void anyway?

Shinigami92: Same as @ST-DDT wrote: because it's now internal and we always new what the input is, I also agree we don't need this check anymore. At least not here, but maybe in faker.seed 🤔

ST-DDT: We usually dont handle badly typed input so I assume we dont need it at all.

Shinigami92: removed

src/internal/mersenne/mersenne.ts
Shinigami92: changed

src/internal/mersenne/mersenne.ts
```diff
@@ -1,81 +1,53 @@
...
-    if (typeof A !== 'object') {
-      throw new FakerError(
-        `seed_array(A) must take array of numbers; is ${typeof A}`
-      );
-    }
+  return {
+    next({ min, max }): number {
+      if (min > max) {
+        [min, max] = [max, min];
+      }
```
ST-DDT: This isn't needed anymore.

ST-DDT (dismissed):

src/internal/mersenne/mersenne.ts
Shinigami92: removed

ST-DDT (dismissed):

ST-DDT (dismissed):

ST-DDT (approved):

src/internal/mersenne/mersenne.ts
```diff
@@ -1,81 +1,49 @@
...
-   * @since 5.5.0
+   * @param seed The seed to use.
    */
-  seed(S: number): void {
-    if (typeof S !== 'number') {
-      throw new FakerError(
-        `seed(S) must take numeric argument; is ${typeof S}`
-      );
-    }
+  seed(seed: number | number[]): void;
```
xDivisionByZerox, line 23: what is the use-case of `seed: number[]`? Just asking in case we can make the interface "simpler" by only allowing plain numbers.

This question is not blocking!

xDivisionByZerox (approved):

src/internal/mersenne/mersenne.ts
ST-DDT: Mainly legacy reasons. I think about reviving it by adding support for `any` seeds via json.stringify and char[] to number[].

That clould be a useful feature for contextual faker.fork.
Faker.fork allows you to create multiple data without impacting the main seed, so you can add new attributes to a person without changing the next person attributes.
The contextual fork allows the same but enhances it with the ability to generate distinct branches from the same origin.
E.g. if you generate a marriage certificate where you need a e.g. female and a male person, but only want to consume a single seed value from the main faker. Then you can even add a pastor or certifing officer to the document without using more seed values.
The benefit of using the contextual fork over the normal fork is, that you are far away from the original seed without loosing the determinism.
If you just consume a seed value between invocations then you might get repetion in the second firstname etc...

See also https://github.com/faker-js/faker/issues/627#issuecomment-1070650803

Most of this is just a thought in my head, I have to write a PR for it to actually check how useful/feasible that is, but for now I would like to keep it.

ejcheng (approved):

PR #1453, opened 2022-10-15
feat(helpers): fake from array

### LINKED ISSUE #1446: Introduce fakeArrayElement helpers

In a lot of places we have a list of patterns in our locales that should be processed by `fake()`.

So the code usually looks like this:

````ts
const patterns = faker.definitions.address.city;
const pattern = faker.helper.arrayElement(patterns);
return faker.helper.fake(pattern);
````

### Suggested solution

Add a new method that combines these two method calls into a single call:

````ts
return faker.helpers.fakeArrayElement( faker.definitions.address.city );
````

### Alternative

Change the `fake()` method to take both a `string` and `string[]` and if it is an array call `arrayElement` on the array before starting the normal procedure.

### Additional context

- [x] I'm willing to create a PR for this.

### PR DESCRIPTION
**Pull request by ST-DDT**
Fixes #1446 

Example Usage:

````ts
faker.helpers.fake(['A: {{name.firstName}}', 'B: {{name.lastName}}']) // 'A: Barry'
````

<details>
<summary>Docs Preview</summary>

grafik

</details>

### CODE REVIEW

**Review by pkuczynski — DISMISSED** [review]

File: src/modules/helpers/index.ts [hunk 1]
```diff
@@ -494,7 +494,7 @@ export class HelpersModule {
    *
    * It is also NOT possible to use any non-faker methods or plain javascript in such templates.
    *
-   * @param str The template string that will get interpolated. Must not be empty.
+   * @param str The template string that will get interpolated. Must not be empty. If an array is passed, a random element will be picked and interpolated.
```
Shinigami92: I know this will duplicate the whole JSDoc, but can we split the signatures to two overloads?
In that case we can name it str for the first signature and arr for the second.
Also the descriptions can differ.
The arr / second-sig can also have another `@since`

test/helpers.spec.ts
```diff
@@ -572,6 +576,21 @@ describe('helpers', () => {
           );
         });
 
+        it('should be able to pass multiple static templates', () => {
```
Shinigami92, line 667: In line 525 we have a test for calling fake with no arguments (not possible with ts, but anyway)
I would like to have added a test for what happens when it gets called with an empty string, and one for with empty array.
If we already have these two somewhere, let me know the lines.

Shinigami92 (changes requested):

src/modules/helpers/index.ts
ST-DDT: The jsdocs will be tripled (`string`, `string[]`, `string | string[]`). 
What would you write differently for the other signature?

test/helpers.spec.ts
ST-DDT: Added

src/modules/helpers/index.ts
ST-DDT: Is `template` a better name for the parameter?

Shinigami92: So I want to write two things:

1. For me it's the same as with `faker.seed`
   https://github.com/faker-js/faker/blob/0f9b0c64e556349cd02d54cb78a780373b2ddb97/src/faker.ts#L175-L241
   So the upper two _specific_ signatures are the ones that gets shown when you use an IDE e.g. VSCode.
   The signature that bundles the upper signatures into one (because that's the real JS one) has it's JSDoc only used by TSDoc as far as I remember :thinking: If this is wrong or can be improved to not be needed, please let me know and we might be able to clean some JSDocs out of the project.

2. Now as we have nowadays IDE hints, a user would e.g. see
   ```ts
   const result = faker.helpers.fake(str: ['{{person.firstName}}', '{{person.lastName}}'])
   //                                _____
   ```
   where `str: ` is just provided in grey color by the IDE. But that name might be confusing as it is not a string at all but an array.

---

I immediately also thought about to name it `template`, but reading
https://github.com/faker-js/faker/blob/9a38c37968110b284b25752d8cd3fd4c0a2c6ef6/src/modules/helpers/index.ts#L497
might be weird to have `template`, `element` and `str` here and there, and in some contexts it is named `pattern` :thinking: 
We might decide for one name,
But nevertheless it is a single value or an array, the parameter name would likely end with an `s` if it receives an array.

ST-DDT: Lets discuss the parameter name in the next team meeting.

As for the duplicate jsdocs. I don't see added value here.

Shinigami92: > As for the duplicate jsdocs. I don't see added value here.

You as user or you as maintainer?

IMO the maintenance burden on our side to copy&paste a JSDoc block and slightly alter it is comparable be far to having a better UX/DX for 1.3 million (weekly downloads) individuals.
But if you like, we can discuss this in team as well, as you already also want to wait until meeting based on the param rename.

src/modules/helpers/index.ts
```diff
@@ -505,27 +505,163 @@ export class HelpersModule {
    * faker.helpers.fake('Good Morning {{person.firstName}}!') // 'Good Morning Estelle!'
    * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
    * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
+   * faker.helpers.fake(['A: {{person.firstName}}', 'B: {{person.lastName}}']) // 'A: Barry'
    *
    * @since 7.4.0
    */
-  fake(str: string): string {
+  /**
```
Shinigami92: Looks like there was a merge conflict not solved successfully 🤔

Shinigami92 (changes requested):

src/modules/helpers/index.ts
ST-DDT: Fixed

src/modules/helpers/index.ts
```diff
@@ -508,24 +508,116 @@ export class HelpersModule {
...
+   * It is also possible to provide parameters. At first, they will be parsed as json,
+   * and if that isn't possible, we will fall back to string:
+   *
+   * ```js
+   * const message = faker.helpers.fake(`You can call me at {{phone.number(+!# !## #### #####!)}}.')
+   * ```
+   *
+   * Currently it is not possible to set more than a single parameter.
+   *
+   * It is also NOT possible to use any non-faker methods or plain javascript in such patterns.
```
Shinigami92: It feels like this should be slightly rewritten to reflect more the usage specifically for passing the patterns array.

src/modules/helpers/index.ts
```diff
@@ -508,24 +508,116 @@ export class HelpersModule {
...
+   * faker.helpers.fake('This is static test.') // 'This is static test.'
+   * faker.helpers.fake('Good Morning {{person.firstName}}!') // 'Good Morning Estelle!'
+   * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
+   * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
+   * faker.helpers.fake(['A: {{person.firstName}}', 'B: {{person.lastName}}']) // 'A: Barry'
+   *
+   * @since 7.4.0
+   */
+  fake(pattern: string | string[]): string;
+  fake(pattern: string | string[]): string {
```
Shinigami92, line 640: I might remember wrongly, but do we really need this doubled signature?
I thought we just need the specific ones and the last generic signature that bundles the specific ones into JS are just for JS and the VitePress generated docs.

src/modules/helpers/index.ts
```diff
@@ -508,24 +508,116 @@ export class HelpersModule {
...
+      // TODO @ST-DDT 2022-10-15: Remove this check after we fail in `arrayElement` when the array is empty
+      if (pattern == null) {
+        throw new FakerError('Fake template array cannot be empty!');
+      }
+    }
     // if incoming str parameter is not provided, return error message
-    if (typeof str !== 'string' || str.length === 0) {
-      throw new FakerError('string parameter is required!');
+    if (typeof pattern !== 'string' || pattern.length === 0) {
+      throw new FakerError('string/string[] parameter is required!');
```
Shinigami92:
```suggestion
      throw new FakerError('string or array of strings is required as parameter');
```

Maybe that's more readable :thinking: 
Do we want the exclamation mark at the end? Not sure about that

---

Beside that, why do we check the input here? Is it special or can we remove it and depend on TypeScript?

src/modules/helpers/index.ts
ST-DDT: Yes, we do. Because typedoc (we) use the last api signature to generate our api docs. The implementation signature isn't taken into account.

src/modules/helpers/index.ts
ST-DDT: 1) I adjusted it to be closer to other error message.
2) I removed it. I assume this is part of the legacy code.

src/modules/helpers/index.ts
ST-DDT: Fixed. If you want to change it some more, you have to be specific.

src/modules/helpers/index.ts
```diff
@@ -513,24 +513,127 @@ export class HelpersModule {
...
+   * which are faster and strongly typed (if you are using TypeScript),
+   * e.g. ``const address = `${faker.location.zipCode()} ${faker.location.city()}`;``
+   *
+   * This method is useful if you have to build a random string from a static, non-executable source
+   * (e.g. string coming from a user, stored in a database or a file).
+   *
+   * It checks the given string for placeholders and replaces them by calling faker methods:
+   *
+   * ```js
+   * const hello = faker.helpers.fake('Hi, my name is {{person.firstName}} {{person.lastName}}!');
```
Shinigami92: Lets use an array example here as well

src/modules/helpers/index.ts
```diff
@@ -513,24 +513,127 @@ export class HelpersModule {
...
+   * const message = faker.helpers.fake([
+   *   'You can call me at {{phone.number(+!# !## #### #####!)}}.',
+   *   'My email is {{internet.email}}.',
+   * ]);
+   * ```
+   *
+   * It is also possible to use multiple parameters (comma separated).
+   *
+   * ```js
+   * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
```
Shinigami92: And here

xDivisionByZerox (dismissed):
TBH, I don't like this change, as it adds a lot of "unnecessary" complexity on top of an already complex function.
But I don't see anything wrong with the implementation itself.

ST-DDT:
@xDivisionByZerox Would you prefer a separate method? Any suggestions for a name? `fakeOneOf`?

src/modules/helpers/index.ts
```diff
@@ -513,24 +513,127 @@ export class HelpersModule {
...
+   * It checks the given string for placeholders and replaces them by calling faker methods:
+   *
+   * ```js
+   * const hello = faker.helpers.fake(['Hi, my name is {{person.firstName}} {{person.lastName}}!']);
+   * ```
+   *
+   * This would use the `faker.person.firstName()` and `faker.person.lastName()` method to resolve the placeholders respectively.
+   *
+   * It is also possible to provide parameters. At first, they will be parsed as json,
+   * and if that isn't possible, we will fall back to string:
```
Shinigami92:
```suggestion
   * and if that isn't possible, it will fall back to string:
```

src/modules/helpers/index.ts
```diff
@@ -513,24 +513,127 @@ export class HelpersModule {
...
+   * It checks the given string for placeholders and replaces them by calling faker methods:
+   *
+   * ```js
+   * const hello = faker.helpers.fake('Hi, my name is {{person.firstName}} {{person.lastName}}!');
+   * ```
+   *
+   * This would use the `faker.person.firstName()` and `faker.person.lastName()` method to resolve the placeholders respectively.
+   *
+   * It is also possible to provide parameters. At first, they will be parsed as json,
+   * and if that isn't possible, we will fall back to string:
```
Shinigami92:
```suggestion
   * and if that isn't possible, it will fall back to string:
```

xDivisionByZerox:
> @xDivisionByZerox Would you prefer a separate method? Any suggestions for a name? `fakeOneOf`?

A separate method would definitely satisfy the single responsibility principle. But I guess that would negate the overall code improvement you had in mind with this changes? As now every function is required to first check the type and then call the according function:
```ts
function foo(options: string | string[]) {
  return Array.isArray(options) 
    ? this.faker.helpers.fakeOneOf(options) 
    : this faker.helpers.fake(options);
}
```

Shinigami92 (approved):

src/modules/helpers/index.ts
```diff
@@ -541,24 +541,127 @@ export class HelpersModule {
...
+   * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
+   * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
+   * faker.helpers.fake(['A: {{person.firstName}}', 'B: {{person.lastName}}']) // 'A: Barry'
+   *
+   * @since 7.4.0
+   */
+  fake(pattern: string | string[]): string;
+  fake(pattern: string | string[]): string {
+    if (Array.isArray(pattern)) {
+      pattern = this.arrayElement(pattern);
```
xDivisionByZerox, line 642: Filter empty strings.

```suggestion
      pattern = this.arrayElement(pattern.filter((str) => str));
```

As I would expect `faker.helpers.faker(['', '{{firstname}}', ''])` not to fail in any case, as I provided at least one template.

ST-DDT: See https://github.com/faker-js/faker/issues/1457 / https://github.com/faker-js/faker/pull/1679
The empty string will be allowed in the future.

xDivisionByZerox (approved):

src/modules/helpers/index.ts
xDivisionByZerox: Makes sense, good work 👍

ejcheng (approved):

PR #1304, opened 2022-08-26
infra: changelog generation

xDivisionByZerox:
This PR introduces configuration for our CHANGELOG file according to the [`standard-version` package](https://github.com/conventional-changelog/standard-version#configuration).

You can see what commands would be run, without committing to git or updating files by using the dry run mode:
1. switch up project locally (if you haven't already)
2. switch to this branch
3. run `npm run release -- --dry-run`

I will provide a sample below.

Fixes #1244

PR #1455, opened 2022-10-15
infra: setup release-it

### LINKED ISSUE #1096: Setup `release-it` + changelog plugin
### Replace https://github.com/conventional-changelog/standard-version with https://github.com/changesets/changesets

`standard-version` is deprecated and not maintained anymore: see https://www.npmjs.com/package/standard-version
The author suggests to use https://github.com/googleapis/release-please
We could look into that, but I would suggest to have a look into `changesets` as it is [used by pnpm](https://github.com/pnpm/pnpm/blob/47b5e45dd6f5eca91a5cedd059aed0f3a7e73be9/package.json#L6)

As changing the release process might produce a different style of `CHANGELOG.md`, I would prefer to make the move to a new release workflow in a major update of Faker.

`changesets` can definitely handle mono-repos and maybe someday we want to convert this repo to a monorepo, so when suggesting alternative tools or workflows, keep this in mind.

_related to_ https://github.com/faker-js/faker/issues/268

Shinigami92:
closes #1096


Shinigami92:
No matter what I do when testing, no `CHANGELOG.md` is generated :thinking: 
Also it doesn't bump the version automatically to e.g. `v8.0.0`
I assume I do something really wrong, but the documentation or verbose logs are also not very helpful 😟

@webpro Would you be so kind and guide us a bit? :slightly_smiling_face:

webpro:
The `preset` is missing in the plugin config:

```json
"plugins": {
  "@release-it/conventional-changelog": {
    "infile": "CHANGELOG.md",
    "preset": {
      "name": "conventionalcommits",
      "types": [...]
    }
  }
}
```

The issue you were facing is that, without a preset, `conventional-recommended-bump` did not return a recommended bump, in turn making release-it skip the release part (which includes everything you are missing).

However, release-it could certainly help out a little here, e.g. the plugin could throw an exception, but at the very least release-it itself should show an error or warning, so I'll add that.

Shinigami92:
> The `preset` is missing in the plugin config:
> 
> ```json
> "plugins": {
>   "@release-it/conventional-changelog": {
>     "infile": "CHANGELOG.md",
>     "preset": {
>       "name": "conventionalcommits",
>       "types": [...]
>     }
>   }
> }
> ```

Oh my god :facepalm: 
Thank you very much, I was so blind
I put `"types" : [...]` next to `"infile"` and didn't wrapped it into `"preset": { "types": [...] }`

Shinigami92:
:heart_eyes:

Shinigami92:
I made a **real-world** test release via https://github.com/Shinigami92/pug-fmt !
For complete automation, we need to setup a auto-workflow that access GH_TOKEN or if we want to release manually, a maintainer needs the right to push to default branch directly

Shinigami92:
Please feel free to review the PR, but let's not merge it now

.release-it.json
```diff
@@ -0,0 +1,50 @@
...
+        "name": "conventionalcommits",
+        "types": [
+          {
+            "type": "feat",
+            "section": "Features"
+          },
+          {
+            "type": "fix",
+            "section": "Bug Fixes"
+          }
```
ST-DDT, line 45: Does this section needs to be extended to meet our requirements?

ST-DDT (changes requested):
IMO the changelog file should be kept, but empty except for a link to the previous version.
Otherwise the v8.0.0 won't have it.

.release-it.json
Shinigami92: Yes, will be done in
- #1304

Shinigami92:
> IMO the changelog file should be kept, but empty except for a link to the previous version. Otherwise the v8.0.0 won't have it.

The linkage must somehow be done post-firsttime generation, so it doesn't mess with the release-it generator

ST-DDT:
> > IMO the changelog file should be kept, but empty except for a link to the previous version. Otherwise the v8.0.0 won't have it.
> 
> The linkage must somehow be done post-firsttime generation, so it doesn't mess with the release-it generator

Lets check whether this breaks the generation first. Somehow, I believe we will forget about that once we actually do the release.
(Also I assume that the release is then already created and we cannot magically add the file afterwards so its in the release.)

Shinigami92:
I've lost the strength to continue working on this PR
It's open to be claimed by any other contributor

xDivisionByZerox:
I'll continue working on this.

ST-DDT:
**Team Decision**

We don't need this for v8.0.
(We will likely immediately start with v9 after v8.0)

Shinigami92:
Right now this PR gets stale and would need anyway a rebase and so on
As well as I'm currently not willing anymore to spend anymore time from my side into this, I will close this PR for now

@xDivisionByZerox you are free to open a new fresh PR if you want to tackle this
But for now I have this out of my PR filter `author:@me`

PR #1458, opened 2022-10-15
test: fix fake test expectation

ST-DDT:
The test uses the following template with `fake()`:

`'}}hello{{random.alpha}}'`

Starting with v8 the new default for the `casing` option is `mixed`, thus the regex has to be updated.

https://next.fakerjs.dev/api/random.html#alpha


ejcheng (approved):

Shinigami92 (approved):
hu? did I missed that one in #1440 ?

PR #1476, opened 2022-10-22
feat(datatype): introduce probability option to boolean

ST-DDT:
Throughout our code base the have some locations that use probabilities to decide what to do.
This PR merges these occurrences into our `datatype.boolean()` method and thus provides this option to others.

<details>
<summary>Docs Preview</summary>

grafik

</details>

### DIFF
--- a/src/modules/datatype/index.ts
+++ b/src/modules/datatype/index.ts
@@ -192,13 +192,30 @@ export class DatatypeModule {
   /**
    * Returns the boolean value true or false.
    *
+   * @param options Options object or the probability (`[0.00, 1.00]`) of returning `true`. Should be between `0` and `1`. Defaults to `0.5`.
+   * @param options.probability The probability (`[0.00, 1.00]`) of returning `true`. Should be between `0` and `1`. Defaults to `0.5`.
+   *
    * @example
    * faker.datatype.boolean() // false
+   * faker.datatype.boolean(0.9) // true
    *
    * @since 5.5.0
    */
-  boolean(): boolean {
-    return !!this.number(1);
+  boolean(options: number | { probability?: number } = {}): boolean {
+    if (typeof options === 'number') {
+      options = {
+        probability: options,
+      };
+    }
+    const { probability = 0.5 } = options;
+    if (probability <= 0) {
+      return false;
+    }
+    if (probability >= 1) {
+      // This check is required to avoid returning false when float() returns 1
+      return true;
+    }
+    return this.faker.datatype.float({ min: 0, max: 1 }) < probability;
   }
 
   /**
--- a/src/modules/datatype/index.ts
+++ b/src/modules/datatype/index.ts
@@ -192,8 +192,15 @@ export class DatatypeModule {
   /**
    * Returns the boolean value true or false.
    *
-   * @param options Options object or the probability (`[0.00, 1.00]`) of returning `true`. Should be between `0` and `1`. Defaults to `0.5`.
-   * @param options.probability The probability (`[0.00, 1.00]`) of returning `true`. Should be between `0` and `1`. Defaults to `0.5`.
+   * **Note:**
+   * A probability of `0.75` results in `true` being returned `75%` of the calls; likewise `0.3` => `30%`.
+   * If the probability is zero or below, it will always return false.
+   * If the probability is one or above, it will always return true.
+   * The probability is limited to two decimal places.
+   * Please keep in mind that the result depends on the seed and is not random.
+   *
+   * @param options The optional options object or the probability (`[0.00, 1.00]`) of returning `true`. Defaults to `0.5`.
+   * @param options.probability The probability (`[0.00, 1.00]`) of returning `true`. Defaults to `0.5`.
    *
    * @example
    * faker.datatype.boolean() // false
--- a/src/modules/datatype/index.ts
+++ b/src/modules/datatype/index.ts
@@ -198,6 +198,7 @@ export class DatatypeModule {
    * @example
    * faker.datatype.boolean() // false
    * faker.datatype.boolean(0.9) // true
+   * faker.datatype.boolean({ probability: 0.1 }) // false
    *
    * @since 5.5.0
    */
--- a/src/modules/datatype/index.ts
+++ b/src/modules/datatype/index.ts
@@ -197,7 +197,6 @@ export class DatatypeModule {
    * If the probability is `<= 0.0`, it will always return `false`.
    * If the probability is `>= 1.0`, it will always return `true`.
    * The probability is limited to two decimal places.
-   * Please keep in mind that the result depends on the seed and is not random.
    *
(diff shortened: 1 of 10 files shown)


src/modules/datatype/index.ts
```diff
@@ -191,13 +191,31 @@ export class DatatypeModule {
   /**
    * Returns the boolean value true or false.
    *
+   * @param options Options object or the probability (`[0.00, 1.00]`) of returning `true`. Should be between `0` and `1`. Defaults to `0.5`.
```
Shinigami92: Should we explicitly document that if it's higher or lower than the requested values, it's always true / false?

Shinigami92 (commented):
Looking into the test snapshots, it looks like the behavior exactly swapped, true is now false and false is now true :thinking: 
Should we just switch the condition to archive the "old" result(s)?

src/modules/datatype/index.ts
ST-DDT: I do consider them invalid, but not to the point where I would throw.
So I wouldn't specifically mention it.
Same for our length < 0 in other methods.

ST-DDT:
> Looking into the test snapshots, it looks like the behavior exactly swapped, true is now false and false is now true 🤔 Should we just switch the condition to archive the "old" result(s)?

IMO the probability should refer to true and not false. I dont know how to preserve the old behavior that wouldn't result in bad code readability/understandability.
E.g. `1 - float(...) < probability`.
Its a major version upgrade so a predictably flipped random boolean is acceptable IMO.
Also, if we flip the result, then we would have to flip the usages as well (maybe+iban).

src/modules/datatype/index.ts
```diff
@@ -191,13 +191,31 @@ export class DatatypeModule {
   /**
    * Returns the boolean value true or false.
    *
```
Shinigami92, line 194: Should we add a bit more explanation?
Like if less 0 then always false, if greater 1 then always true, prop of 0.6 means that 60% chance of getting true
But in better understandable english of course 😉

Shinigami92 (dismissed):

src/modules/datatype/index.ts
ST-DDT: IMO below zero and above one is undefined behavior, that just works to simplify usage. 0.6 is 60% in a mathematical sense, so I'm not sure its worth explicitly mentioning it.

What do the others think?

xDivisionByZerox: Explicitly mentioning something in documentation is always better than assuming IMO. But the thing itself (range inclusiveness) is documented in the `@param` flag so that works.

xDivisionByZerox (dismissed):

src/modules/datatype/index.ts
ST-DDT: 🤔 

1) Everything outside the 0-1 range is explicitly undefined behavior, so if we come to the conclusion that those are bad values, we can throw an error instead. The current behavior is there so it doesn't throw for some computer math failure resulting in `1.000000000001` or something similar.
2) It already mentions that it is the probability of returning true.
3) `0.6 === 60%`. IMO this is common knowledge in both math and computers. Especially with the given range it should be obvious.

For 1 I could think about changing the description to:

- ``The probability `[0.00, 1.00]` (inclusive) of returning `true`. Defaults to `0.5`.``
- or ``The probability (`[0.00, 1.00]`) of returning `true`. Must be between `0` and `1`. Defaults to `0.5`.``

But I think neither of them would any more clear.

For 3 I will go and ask in a few people around me, if this is common knowledge to them.

Is there anything I have missed?

Shinigami92: I wont like to throw an error for values out of bound in this context.
It's more like if you use this function via a bit more complex math, we can explain that if you pass 120% (overcharge :rocket:) you will always get `true`, same for negative values.
So if a user of the function wants to be sure, they may want to clamp their input before passing it to our function.

But for me this is not blocking, just a nit-pick.

src/modules/datatype/index.ts
ST-DDT: Done

src/modules/datatype/index.ts
ST-DDT: It looks like 0.xy = xy% isn't common knowledge, so I added it as a note to the jsdocs.
Is this better?

src/modules/datatype/index.ts
```diff
@@ -192,13 +192,38 @@ export class DatatypeModule {
...
+   * **Note:**
+   * A probability of `0.75` results in `true` being returned `75%` of the calls; likewise `0.3` => `30%`.
+   * If the probability is zero or below, it will always return false.
+   * If the probability is one or above, it will always return true.
```
Shinigami92:
```suggestion
   * If the probability is `<= 0.0`, it will always return `false`.
   * If the probability is `>= 1.0`, it will always return `true`.
```

src/modules/datatype/index.ts
```diff
@@ -192,13 +192,38 @@ export class DatatypeModule {
...
+   * The probability is limited to two decimal places.
+   * Please keep in mind that the result depends on the seed and is not random.
```
Shinigami92: > Please keep in mind that the result depends on the seed and is not random.

I think this is more confusing then clarifying something.
I'm just sitting here and reading this sentence over and over again and just think: Why is it not random but depends on the seed ... to be random ... ?!

ST-DDT: It is not random. It is deterministic.

My point is, we allow the user to pass a probability, but as long as the seed is of a specific value the result will always be the same.

Shinigami92: Yeah, but what makes this different from any other faker function so that it needs to be explicitly mentioned 🤔

src/modules/datatype/index.ts
```diff
@@ -192,13 +192,37 @@ export class DatatypeModule {
   /**
    * Returns the boolean value true or false.
    *
+   * **Note:**
+   * A probability of `0.75` results in `true` being returned `75%` of the calls; likewise `0.3` => `30%`.
+   * If the probability is `<= 0.0`, it will always return `false`.
+   * If the probability is `>= 1.0`, it will always return `true`.
+   * The probability is limited to two decimal places.
```
xDivisionByZerox, line 199: Is it really tho? I don't see anything in the code that states that.

ST-DDT: That's the default for the float method:

https://github.com/faker-js/faker/blob/4204a6ae925a4e0c761ece10ad21222dac7a9e7e/src/modules/datatype/index.ts#L78

I don't see a need to increase the precision beyond that.

xDivisionByZerox: Oh, mb I thought for some reason that it was the default 20 digit precision^^

xDivisionByZerox (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #1480, opened 2022-10-23
feat(internet): add options to url()

### LINKED ISSUE #1474: add options to faker.internet.url()

faker.internet.url() produces URLs in `http://www.url.com` format (format 1).
There isn't any slash in the end of this URL. If I invoke `new URL('http://www.url.com').toString()`,
I'll get URL `http://www.url.com/` (format 2). A problem is that I use `https://github.com/Urigo/graphql-scalars`
and GraphQLURL resolver transforms URL from format 1 to format 2 by using `new URL()`.
It is a problem to compare format 1 with format 2 after I get a response from the GraphQL server.

### Suggested solution

I think that it needs to implement an argument in faker.internet.url() to generate URLs of format 1 or 2.
Maybe it is necessary to create a common config where we can set common arguments for methods
which will be used when the methods are invoked without any argument. 

### Alternative

_No response_

### Additional context

_No response_

ejcheng:
closes #1474


src/modules/internet/index.ts
```diff
@@ -226,15 +228,28 @@ export class InternetModule {
   }
 
   /**
-   * Generates a random url.
+   * Generates a random http(s) url.
+   *
+   * @param options Optional options object.
+   * @param options.appendSlash Whether to append a slash to the end of the url. Defaults to a random boolean value.
```
```suggestion
   * @param options.appendSlash Whether to append a slash to the end of the url (path). Defaults to a random boolean value.
```

test/internet.spec.ts
```diff
@@ -325,6 +333,23 @@ describe('internet', () => {
...
           expect(url).toSatisfy(validator.isURL);
         });
+
+        it('should return a valid url with slash appended at the end', () => {
+          const url = faker.internet.url({ appendSlash: true });
+
+          expect(url).toBeTruthy();
+          expect(url).toBeTypeOf('string');
+          expect(url).toSatisfy(validator.isURL);
+          expect(url.charAt(url.length - 1)).toBe('/');
```
Shinigami92: Maybe just use `expect(url.endsWith('/')).toBeTruthy();`
Feels like much more readable to me.

ejcheng: Didn't realize there was an `endsWith` method😅

Shinigami92 (approved):

ST-DDT (approved):

Issue #1349, opened 2022-09-06
Standardize function parameters and defaults


ST-DDT:
### Clear and concise description of the problem

Most methods have various ways to specify parameters and have different defaults.

### Suggested solution

All methods should be checked that they have:
- consistent parameter definitions
- consistent parameter names
- consistent parameter defaults

The consistency should span across all modules.

ST-DDT:
Blocked by #1346

ST-DDT:
- [x] [string.hexadecimal doesn't support length without options](https://github.com/faker-js/faker/pull/1155#discussion_r997780989)
- [x] [string.sample doesn't support options](https://github.com/faker-js/faker/pull/1155#discussion_r997769964)

Shinigami92:
- [ ] Merge sentences options https://github.com/faker-js/faker/pull/1486/files#diff-6a52f070da035438588918337a536421fd92d4d78ef4998bf140f6fd925b8974R140-R143

Shinigami92:
- [x] [`date` doesn't support options](https://github.com/faker-js/faker/blob/next/src/modules/date/index.ts)

xDivisionByZerox:
Is the `LoremModule` already considered "standardized"?
All functions already have an object argument, but with a more descriptive name than our standard "options".
For Example:
https://github.com/faker-js/faker/blob/fccd641f02fc0db463d34e1a7cc459ad6bd0b265/src/modules/lorem/index.ts#L106-L119

I'm not sure what to do here.

matthewmayer:
I was going to add a section to the migration guide about these changes, but to clarify is it intended that the non-options-object parameters will be deprecated in v8 and removed in v9? Or will both versions be supported in v8+?

xDivisionByZerox:
TLDR;

So yes, some of the signatures will be deprecated in v8 and removed in v9. So they are usable for v8.x. But most of them simply get standardized into an options object and can be used even after v9.

---

> [...] to clarify is it intended that the non-options-object parameters will be deprecated in v8 and removed in v9

No, not entirely. Primitive arguments are often desired for configuring the parameter which is most likely to be used the most by users. For example:
```ts
faker.number.int(10); // number between 0 and 10
faker.number.int({ max: 10 }); // number between 0 and 10
faker.number.int({ max: 10, min: 10 }); // number between 0 and 10
```
[...]

PR #1486, opened 2022-10-24
feat(helpers): add rangeToNumber method and add range parameters

### LINKED ISSUE #617: Improve methods with a length: number argument to support ranges

Some methods like hexadecimal take only a fixed range as parameter. If you want to allow a dynamic range, you have to combine it with another method first.
`faker.datatype.hexaDecimal(faker.datatype.number(...))`
This reduces the readability of the code.

### Suggested solution

Improve their method signature to allow for ranges:
`hexaDecimal(length: number | {min,max})`

### Alternative

_No response_

### Additional context

What do you think?

ST-DDT:
Fixes #617

* Adds `helpers.rangeToNumber(numberOrRange: number | { min: number; max: number }): number`
* And uses it where applicable


Shinigami92:
I'm feeling a bit unconfy about this change.
Here are just my said out load thoughts but no blocker:

- `toNumber` could be made internal as I do not expect that it is in any kind helpful for an end user
- sometimes it's `*count`, sometimes it's `length`. Lets write a rule somewhere when to use which or lets unify them
- sometimes it's only `length | { min, max }`, sometimes it's `length | { length | min, max }`. I personally prefer the second one as then as a user you can use `meth({ length: 42 })` and have readable code + it's a bit more future proof if another option param gets added (by us or by user)
- there are still parts in the code where an if-condition is needed to convert a length option into a number, I feel like this should also be handled by the helper method

ST-DDT:
> * `toNumber` could be made internal as I do not expect that it is in any kind helpful for an end user

Unless you want to write your own faker functions, which applies to most of our helper functions.

> * sometimes it's `*count`, sometimes it's `length`. Lets write a rule somewhere when to use which or lets unify them

I didnt touch that part in this PR. This is a task for #1349.

> * sometimes it's only `length | { min, max }`, sometimes it's `length | { length | min, max }`.

I can change that (use the range only in options).

> * there are still parts in the code where an if-condition is needed to convert a length option into a number

Could you give me an example?

Shinigami92:
> > * `toNumber` could be made internal as I do not expect that it is in any kind helpful for an end user
> 
> Unless you want to write your own faker functions, which applies to most of our helper functions.

Isn't this the same as you requested in image module? There I will also pass the faker instance just forward to the internal called function

> > * sometimes it's `*count`, sometimes it's `length`. Lets write a rule somewhere when to use which or lets unify them
> 
> I didnt touch that part in this PR. This is a task for #1349.

Ah okay, I already thought that this PR is at least a part of
- #1349

> > * there are still parts in the code where an if-condition is needed to convert a length option into a number
> 
> Could you give me an example?

I was referring to this part (I originally meant not from opt to num but from num to opt):
https://github.com/faker-js/faker/blob/6bb73689684ae32afac9075d1312b1bbcea7d24a/src/modules/string/index.ts#L125-L131

ST-DDT:
> I was referring to this part (I originally meant not from opt to num but from num to opt):

IMO that is a different feature. Maybe `toOptions(...)`? Not sure whether this should be done in this PR.

ST-DDT:
> > Unless you want to write your own faker functions, which applies to most of our helper functions.
> 
> Isn't this the same as you requested in image module? There I will also pass the faker instance just forward to the internal called function

I don't think so or I don't know what you are referring to.

If someone external wants to write his own faker method, that happens to require a range of something e.g. a custom `numericAlpha(length: number | { min: number, max: number}): string` method, then the user can use the `toNumber` function in their implementation.
Similar on how a user would use `helpers.arrayElement` in their custom `fooType(): FooType` method.

Shinigami92:
> > I was referring to this part (I originally meant not from opt to num but from num to opt):
> 
> IMO that is a different feature. Maybe `toOptions(...)`? Not sure whether this should be done in this PR.

I see, and I think this can then be solved on a more generic way, which indeed would then not fit anymore in this PR
Something like
```ts
options = toOptions(options);
const length = toNumber(options.length ?? 1);
```
But reading this `toNumber` feels now a bit like we should call it `transformRangeToNumber`, but thats ugly long 😕 
I hope some other voices can help us with further ideas

ST-DDT:
**Team decision**

* We will name the method `rangeToNumber`
* Create a separate PR adding `these are helper methods that can be used to create your own faker methods` to the helper module jsdocs

ST-DDT:
Done. Ready for review.

Shinigami92 (dismissed):

ST-DDT:
Multiple methods depend on this feature, so I moved it to `v8.0`.

Shinigami92 (dismissed):

test/helpers.spec.ts
```diff
@@ -683,6 +687,18 @@ describe('helpers', () => {
         });
       });
 
+      describe('toNumber()', () => {
```
xDivisionByZerox: Can you use the correct function name here?

```suggestion
      describe('rangeToNumber()', () => {
```

Shinigami92: nice catch :+1:

test/lorem.spec.ts
```diff
@@ -189,6 +219,18 @@ describe('lorem', () => {
...
           }
         );
+
+        it('should return a random amount of sentences', () => {
+          const actual = faker.lorem.sentences({ min: 10, max: 20 });
+
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+
+          const words = actual.split('. ');
```
xDivisionByZerox: I think the variable name `word` here is a bit confusing. Probably from copy-pasting previous test implementations.

xDivisionByZerox: This goes for most of the next tests as well.

ST-DDT: Fixed

src/modules/string/index.ts
```diff
@@ -259,12 +263,17 @@ export class StringModule {
...
+      length?: number | { min: number; max: number };
       casing?: Casing;
       prefix?: string;
     } = {}
   ): string {
-    const { length = 1, casing = 'mixed', prefix = '0x' } = options;
+    const length = this.faker.helpers.rangeToNumber(options.length ?? 1);
+    if (length <= 0) {
+      return '';
+    }
```
xDivisionByZerox: Can we get a test case for this? Other functions already have one.

ST-DDT: Fixed

test/string.spec.ts
```diff
@@ -325,6 +353,28 @@ describe('string', () => {
           expect(hex).toMatch(/^[0-9a-f]*$/i);
           expect(hex).toHaveLength(5);
         });
+
+        it.each([0, -1, -100])(
+          'should return empty string when length is <= 0',
+          (length) => {
+            const actual = faker.string.hexadecimal({ length, prefix: '' });
```
xDivisionByZerox: Why are you setting the prefix explicitly here? Looking at the implementation, the function will ALWAYS return an empty string if the length is <= 0 (no matter the prefix). 
This is expected behavior, no?

src/modules/lorem/index.ts
```diff
@@ -60,62 +60,71 @@ export class LoremModule {
...
+  words(wordCount: number | { min: number; max: number } = 3): string {
+    wordCount = this.faker.helpers.rangeToNumber(wordCount);
+
+    return Array.from({ length: wordCount })
+      .map(() => this.word())
+      .join(' ');
   }
 
   /**
    * Generates a space separated list of words beginning a capital letter and ending with a dot.
```
ejcheng:
```suggestion
   * Generates a space separated list of words beginning with a capital letter and ending with a period.
```

ejcheng: Not related to this PR, but I noticed the phrasing here is suspect

src/modules/lorem/index.ts
```diff
@@ -220,6 +241,8 @@ export class LoremModule {
    * Generates the given number lines of lorem separated by `'\n'`.
    *
    * @param lineCount The number of lines to generate. Defaults to a random number between `1` and `5`.
+   * @param lineCount.min The minimum number of lines to generate.
+   * @param lineCount.max The maximum number of lines to generate.
```
ejcheng:
```suggestion
   * @param lineCount.min The minimum number of lines to generate. Defaults to `1`.
   * @param lineCount.max The maximum number of lines to generate. Defaults to `5`.
```

test/string.spec.ts
ST-DDT: It was a really good idea to add this test, because it uncovered an expectation vs implementation mismatch.

IMO a hex string with prefix `0x` and length zero should be `0x` because the length refers to the hex part not prefix+hex.
For consistency reasons, it will return the same for negative lengths.
When I wrote the test, my brain automatically assumed the prefix to be there.

However the implementation clearly didn't follow that.
So what is the correct result for 
`hexadecimal({ length: 0, prefix: '0x' })`

ejcheng (dismissed):

test/string.spec.ts
xDivisionByZerox: > IMO a hex string with prefix 0x and length zero should be 0x because the length refers to the hex part not prefix+hex.

Oh, that's a good take. I didn't thought about it that way, but it makes sense now that I read it.
So the test should rather describe something like `'should return the prefix when length is <= 0'`.

src/modules/system/index.ts
```diff
@@ -50,23 +50,23 @@ export class SystemModule {
...
    * faker.system.fileName({ extensionCount: 2 }) // 'times_after.swf.ntf'
+   * faker.system.fileName({ extensionCount: { min: 1, max: 2 } }) // 'jaywalk_like_ill.osfpvg'
    *
    * @since 3.1.0
    */
   fileName(
     options: {
-      /**
-       * Define how many extensions the file name should have. A negative number will be treated as `0`. Defaults to `1`.
-       */
```
Shinigami92: https://user-images.githubusercontent.com/7195563/202997904-26f9f436-52a4-4a5f-88b0-3e4fc1eefa1f.mov

I somehow liked the UX for having tooltip info on hover over option properties
IMO we should add this for every option property
But we could do this in a separate PR

Shinigami92 (dismissed):

src/modules/system/index.ts
ST-DDT: I reverted this for now.
Please note, that I will only accept it for the entire project, if there is a test that ensures that the text is the same as the param's and optionally the respective parts are generated automatically.

test/string.spec.ts
ST-DDT: Fixed

Shinigami92 (approved):

src/modules/system/index.ts
xDivisionByZerox: > Screen.Recording.2022-11-21.at.09.11.09.mov 
> I somehow liked the UX for having tooltip info on hover over option properties IMO we should add this for every option property But we could do this in a separate PR

That's what I have been complaining about for like 6 months now... Because of that, I proposed to group common options (like Range) into types or interfaces to not have to write these docs every single time...
Every single time you told me "yeah, our docs cant auto gen thus. that sucks, guess we not doing it".

Sorry, but I needed to get that off my chest.

xDivisionByZerox (approved):

src/modules/system/index.ts
Shinigami92: I must say I'm with @xDivisionByZerox here...
I also start more and more thinking about that just because our VitePress Docs cannot (right now) auto-generate or correctly extract these options is a really bad decision to not serve nice IDE UX.

If we finally just start to write such interfaces we have IDE docs just on one place, like:

```ts
/** ... */
export interface AGoodNameForARangeType {
  /** ... */
  min: number;

  /** ... */
  max: number;
}
```

I will open an issue to finally start adding these interfaces as I'm more and more tempted now that we serve a bad UX for IDE users.
Generating good VitePress Docs are a separate step and needs to be handled by someone who understands the automatisation code behind all of that, _and I assume that is not me_.

ejcheng (approved):

src/modules/system/index.ts
Shinigami92: Created the issue: https://github.com/faker-js/faker/issues/1586

PR #1487, opened 2022-10-25
docs: add version banner

### LINKED ISSUE #1456: Add banner to next.fakerjs.dev

Currently is it hard to spot whether you are on a production branch or our development branch.

Alternatively

### Suggested solution

Show a banner with a hint that the users is currently viewing the development branch.

It might be useful to always show the banner if the hostname is not `fakerjs.dev`.

The banner should contain a link to the production website (https://fakerjs.dev).

### Alternative

_No response_

### Additional context

This originally came up here: https://github.com/faker-js/faker/pull/1438#issuecomment-1279750983

brc-dd:
fixes #1456

TODO:
- [x] allow users to dismiss that

---

~~Note on how `id` works: Each time you change the banner text and want users (those who had dismissed it) to see new banner just update the id to any new random value. You can omit `id` to for a particular channel to don't persist users' preference for banner dismissal.~~

Updated behavior: Dismissal preference is only kept for 24hrs now. There won't be need to add random ids.


Shinigami92 (changes requested):
The banner should always be shown on non-`https://fakerjs.dev/` sites
But they all have the same source, so we need to check `window.location.host !== 'fakerjs.dev'` or something like that

docs/.vitepress/config.ts
```diff
@@ -229,4 +229,12 @@ export default defineConfig({
       }),
     },
   },
+
+  vite: {
+    define: {
+      __MAIN__:
+        process.env.DEPLOY_PRIME_URL ===
```
ST-DDT: In other contexts we used a different property:

https://github.com/faker-js/faker/blob/64b031a9e385d1461b2c158abfa1851ba11a8397/docs/.vitepress/versions.ts#L32-L40

IMO the `CONTEXT` is suited better for this purpose, as it is easier to understand than this cryptic address.
We could re-export it there and reuse it here.

docs/.vitepress/components/Banner.vue
```diff
@@ -0,0 +1,61 @@
...
+      `${height.value + 16}px`
+    );
+  }
+});
+</script>
+
+<template>
+  <div ref="el" class="banner">
+    These docs are of the next (unreleased) version. For docs of the current
+    version visit: <a href="https://fakerjs.dev/">fakerjs.dev</a>
```
ST-DDT: This will also be shown on old pages (e.g. v7 once v8 gets released), so maybe this should be a bit differentiated.

We could theoretically export the text (`the next (unreleased)` for `branch === next` / `a previous release` for `branch ~= /^v\d+\/$/` / `a development` for everything else) in the versions.ts and use it here.

**EDIT:** suggestion for insertion into `docs/.vitepress/versions.ts#L43`

````ts
let bannerInfix_: string | undefined;
if (deployContext === 'production') {
  bannerInfix_ = undefined;
} else if (isReleaseBranch) {
  bannerInfix_ = `a previous release version`;
} else if (branchName === 'next') {
  bannerInfix_ = 'the next (unreleased) version';
} else {
  bannerInfix_ = 'a development version';
}
export const versionBannerInfix = bannerInfix_;
````

But we can adjust the text ourselves, once the banner is working as expected.

File: docs/.vitepress/config.ts (hunk 1 above)
Comment (Shinigami92) [reply docs/.vitepress/config.ts]: Do you have an idea how to expose this `CONTEXT` from the server build step to the client side running version?

Comment (ST-DDT) [reply docs/.vitepress/config.ts]: A simple `export` does the job. The vitepress build steps hardcodes the value into the built page and then can be used there.
It/the condition might even be inlined/resolved at build time.

As you can see [here](https://next.fakerjs.dev/):
grafik

That part is already populated from the build time environment variables.

**Comment by brc-dd** [comment]
This is ready for review guys! 🙏

File: docs/.vitepress/versions.ts [hunk 3]
```diff
@@ -41,6 +41,19 @@ const hiddenLink =
...
+  if (deployContext === 'production') {
+    return null;
+  }
+  if (isReleaseBranch) {
+    return { version: '"an old version"', id: 'fk01' };
+  }
+  if (branchName === 'next') {
+    return { version: '"the next (unreleased) version"', id: 'fk02' };
+  }
+  return { version: '"a development version"', id: 'fk03' };
```
Shinigami92:
```suggestion
    return { version: '"an old version"', id: 'old' };
  }
  if (branchName === 'next') {
    return { version: '"the next (unreleased) version"', id: 'next' };
  }
  return { version: '"a development version"', id: 'dev' };
```

docs/.vitepress/config.ts
```diff
@@ -229,4 +222,30 @@ export default defineConfig({
...
+    `
+(() => {
+  const restore = (key, cls, def = false) => {
+    const saved = localStorage.getItem(key);
+    if (saved ? saved !== 'false' : def) {
+      document.documentElement.classList.add(cls);
+    }
+  };
+  window.__FAKER_BANNER_ID__ = '${versionBannerInfix().id}';
+  restore(\`faker-banner-\${__FAKER_BANNER_ID__}\`, 'banner-dismissed');
```
Shinigami92:
```suggestion
  restore(\`faker-version-banner-\${__FAKER_BANNER_ID__}\`, 'banner-dismissed');
```

docs/.vitepress/components/Banner.vue
```diff
@@ -0,0 +1,108 @@
...
+    document.documentElement.style.setProperty(
+      '--vp-layout-top-height',
+      `${height.value + 16}px`
+    );
+  }
+});
+
+const dismiss = () => {
+  localStorage.setItem(
+    `faker-banner-${(window as any).__FAKER_BANNER_ID__}`,
```
Shinigami92:
```suggestion
    `faker-version-banner-${(window as any).__FAKER_BANNER_ID__}`,
```

Shinigami92:
Example Screenshot of iPhone XS (Opera Mini)

Shinigami92:
I'm thinking about to use the `sessionStorage` instead of `localStorage` :thinking:

brc-dd:
> I'm thinking about to use the sessionStorage instead of localStorage

It won't be persistent then. One will need to dismiss it each time they open the tab.

Shinigami92:
> > I'm thinking about to use the sessionStorage instead of localStorage
> 
> It won't be persistent then. One will need to dismiss it each time they open the tab.

I won't want it for each tab, but for each browser re-open :thinking: 
So next time you visit, it come up again, because after half a year or so, you might missed that it is another version
We could also set a timestamp and then check if older then 1 day or x-days

brc-dd:
> We could also set a timestamp and then check if older then 1 day or x-days

Maybe some approach like this: https://www.sohamkamani.com/javascript/localstorage-with-ttl-expiry/ or maybe cookies 😅?

Shinigami92:
I think we don't need to be that complex, we can use the value directly as expire, as currently it is just `1` or `undefined`
So the check would now just be: `defined && < x days, then don't show, otherwise show`, `on dismiss: set timestamp`

brc-dd:
done

ST-DDT:
I think we could just show it always. So the users get annoyed and update to the latest version.
What do you think?

Shinigami92:
> I think we could just show it always. So the users get annoyed and update to the latest version. What do you think?

No, as on mobile it is to big to not get dismissed :warning:

Shinigami92 (dismissed):
For me it's okay now
even if `8.64e7 * 1` is still `8.64e7` and `Date.now() === new Date().getTime()`

brc-dd:
> Date.now() === new Date().getTime()

Directly adding number with `Date.now()` will simply concat strings instead of adding the number to the time. 😅

> 8.64e7 * 1 is still 8.64e7

I kept `1` there if you wanted to customize the number of days (ignoring DST).

Shinigami92:
> > Date.now() === new Date().getTime()
> 
> Directly adding number with `Date.now()` will simply concat strings instead of adding the number to the time. sweat_smile

`typeof Date.now() === 'number'` :thinking: , anyways, it's a nit-pick :slightly_smiling_face: 

> > 8.64e7 * 1 is still 8.64e7
> 
> I kept `1` there if you wanted to customize the number of days (ignoring DST).

Ah I understand, maybe should add a comment? Currently feels like a magic number for not knowing folks.

brc-dd:
> typeof Date.now() === 'number' 🤔 , anyways, it's a nit-pick 🙂

Ah right, I was doing `new Date() + ...`. Will update that.

Shinigami92 (dismissed):

ST-DDT:
I temporarily deployed these to netlify for easier reviewablilty:

- [prod](https://v7--serene-sprite-f3ef50.netlify.app/)
- [old](https://v6--serene-sprite-f3ef50.netlify.app/)
- [next](https://next--serene-sprite-f3ef50.netlify.app/)
- [dev](https://pr-brc-dd-1487--serene-sprite-f3ef50.netlify.app/)

Last updated for [171f7f9](https://github.com/faker-js/faker/pull/1487/commits/171f7f9335c5e056920e4de12a7db2baf7f9046d)

ST-DDT (dismissed):
This looks good to me. How about evaluating the versionBannerInfix value only once and exporting that?
https://github.com/brc-dd/faker-1/compare/next...faker-js:faker:pr/brc-dd/1487-ST-DDT

ST-DDT (dismissed):

Shinigami92 (dismissed):

ST-DDT (dismissed):

docs/.vitepress/config.ts
```diff
@@ -72,13 +72,6 @@ export default defineConfig({
         content: description,
       },
     ],
-    [
-      'meta',
-      {
-        name: 'description',
-        content: description,
```
Shinigami92: Why was this removed? looks unintended to the PR

netlify.toml
```diff
@@ -1,11 +1,10 @@
 [build.environment]
   NODE_VERSION = "18"
-  NPM_FLAGS = "--version" # prevent Netlify npm install
 
 # Documentation
 [build]
   publish = "docs/.vitepress/dist"
-  command = "npx pnpm i --store=node_modules/.pnpm-store && npm run docs:build:ci"
+  command = "pnpm docs:build:ci"
```
Shinigami92: Would you like to create it's own PR (that we can just merge before this PR, so you can rebase without conflict) to tackle this change separately?
That way we have a specific commit history entry for that change.

Shinigami92 (changes requested):

docs/.vitepress/config.ts
brc-dd: There is already `description` option being passed in the config, there is no need to pass it using meta description. It was resulting in duplicate tags. Validator was throwing warnings.

I can move it to different PR if you want.

Shinigami92: So it's managed by VitePress under the hood?

brc-dd: Yeah, this line is equivalent to that block:

https://github.com/faker-js/faker/blob/09e835664add0a342dd089f31bbae7d880198493/docs/.vitepress/config.ts#L56

Shinigami92 (approved):

ST-DDT:
~~For some reason, this doesn't work for the next branch on netlify anymore.
Do you have an idea why?~~

EDIT: Nevermind. I closed it like 23h ago...

ST-DDT (approved):

PR #1499, opened 2022-10-30
feat: introduce faker.clone and derive

### LINKED ISSUE #627: Get the ".last()" value generated of a module

### LINKED ISSUE #2664: Proposal: Use a single seed value per faker function invocation

Glossary:
- seed value: direct or indirect invocation of randomizer.next() or equivalents thereof

Currently when invoking a faker function it may consume an unknown number of seed values (0-n).
While that isn't bad by itself, it has the side effect, that any change to the implementation affects the user by generating different subsequent values. This is especially relevant for functions that generate multiple values such as multiple and unique, each of the generated elements affect the following elements generated in them and all elements afterwards.

### Suggested solution

Let each method use only a single seed value by deriving it, if it uses more than one.
Each method would be responsible for itself to consume only a single seed value.

````ts
function multiple(fakerCore: Faker(Core), generator: (fakerCore, index) => T, options): T[] {
  // consumes a single seed from the original
  const derived = fakerCore.derive(); 
  // consumes a seed value from the derived
  const count = rangeToNumber(derived, options.count); 
  // each call on the generator consumes another seed from derived
  // if the generator would need more than one single value it would derive by itself
  // even if it doesn't the multiple function upholds its contract and still behaves better than a simple for loop
  return Array.from({length: count}, (_, i) => generator(derived, i));
}
````

Important usage detail, the `fakerCore` instance must be passed on and used by any nested code.

Deriving an instance does come at a performance cost, but we could make it cheap, if we keep that as a priority during the derive implementation and use standalone functions like teased in the code example.

<details>
<summary>E.g. by not re-initializing the twister from scratch, but only copy and transforming the state.</summary>

````ts
derive() {
  let random = this.state.next();
  const stateCopy = this.state.map((old) => old + random + aRandomStaticValue + 0 * (random = old));
  return new Twister(stateCopy);
}
````

(We should measure though how much "re-initializing" vs "copy and transforming the state" actually has of a performance impact)

</details>

<details>
<summary>It also comes at a cost of additional code, we could hide that in our potential meta framework though.</summary>

This section is largely unrelated to this proposal and should just demonstrate, how the derive could be included in the potential meta framework.

````ts
function multiple(fakerCore: Faker(Core), generator: (fakerCore, index) => T, options): T[] {
  const count = rangeToNumber(fakerCore, options.count); 
  return Array.from({length: count}, (_, i) => generator(fakerCore, i));
}

--- Autogenerated same file

declare function boundMultiple(generator: (fakerCore, index) => T, options): T[];
[...]
export const multiple = fakerize<typeof multiple, typeof boundMultiple>(multiple, {derive: true, isCallable: ...});
[...]
````

````ts
function fakerize<TRaw, TBound>(fn: TRaw, options): Fakerized<TRaw, TBound> {
  if (options.derive) {
    fn = (fakerCore, ...args) => fn(fakerCore.derive(), ...args);
  }
  fn.bindTo = (fakerCore) => (...args) => fn(fakerCore, ....args) as TBound;
  fn.isCallable = ...;
  return fn;
}
````

Usage
````ts
multiple(fakerEN, (fakerCore) => firstName(fakerCore), 5);
// or multiple(fakerEN, firstName, 5);

const multipleEN = multiple.bindTo(fakerEN);
multipleEN((fakerCore) => firstName(fakerCore), 5);
// or multipleEN(firstName, 5);

// Why not like this?
const firstNameEN = firstName.bindTo(fakerEN);
multipleEN(() => firstNameEN(), 5);
// or
multiple(fakerCore, () => firstNameEN(), 5);

// Because firstNameEN would consume the seeds directly from the bound fakerEN instance
// and thus bypass most benefits from `derive`.
````

</details>

### Alternative

Don't change the current behavior.

### Additional context

Relevant issues/PRs:

- #1499 (outdated demonstration of the feature)
- #1250 (required for performance reasons)
  - #2667

Potentially impacted issues/PRs:

- #2661 (due to the one seed per invocation feature)

---

We already talked about this feature quite a bit, but I would like to have this issue, to bring everybody on the same level.
And give everybody the chance to comment and react.

### PR DESCRIPTION
**Pull request by ST-DDT**
Fixes #627

* #627

Required for:

* #1491

See also:

* #1413

----------

This PR introduces two new methods for faker.

* `faker.clone()`
* `faker.derive()`

## `clone()`

Can be used to generate repeated values or to allow generating the "last" value again.
This method does not consume seed values.

**Example**

````ts
faker.seed(42);
faker.clone().datatype.number(); // 8234
faker.clone().datatype.number(); // 8234
faker.datatype.number(); // 8234
````

`last()` was originally requested in #627

## `derive()`

Can be used to write methods that don't affect the rest of the generated data when changed.
This method consumes a single seed value in order to initialize the seed of the derived instance.

**Example**

````ts
function generatePerson(): Person {
  const derived = faker.derive();
  return {
    firstName: derived.person.firstName(),
    lastName: derived.person.lastName(),
    // You can add more attributes here without affecting the next person
    ...
  };
}

faker.seed(42);
generatePerson(); // { firstName: 'Devyn', lastName: 'Foo', ... }
generatePerson(); // { firstName: 'John', lastName: 'Bar', ... }
````

I evolved the `fork()` method into `derive()` while thinking about #1491 as it is incredible hard for our users to write extensible methods while retaining their original data. It basically makes `generatePerson` or any similar method an atomic operation regardless how many data are generated.

It can also be used to with nesting.

````ts
function generateMarriageCertificate(faker: Faker): MarriageCertificate {
  const derived = faker.derive();
  return {
    partner1: generatePerson(derived),
    partner2: generatePerson(derived),
    certifyingOffice: derived.location.city(),
  };
}

function generatePerson(faker: Faker): Person {
  const derived = faker.derive();
  return {
    firstName: derived.person.firstName(),
    lastName: derived.person.lastName(),
    // You can add more attributes here without affecting the next person
    ...
  };
}
````

## Alternatives

1) Add a `context` option to `clone()`, that mutates the seed (current seed + context -> new seed).
However, I think the consumption of a seed value is actually expected, so that partner/marriage certificate 1 and 2 are different.

2) Add a `seed` option to `clone` and use it like `faker.fork({seed: faker.datatype.number()})`. 
However that would just be a a more verbose version of the current `derive` method.

---

Food for thoughts: We could use the derive method in our code/methods to ensure each faker method call consumes only a single seed value. However this would probably (slightly) decrease performance.

### DIFF
--- a/docs/guide/usage.md
+++ b/docs/guide/usage.md
@@ -275,3 +275,188 @@ We will update these docs once a replacement is available.
 :::
 
 Congratulations, you should now be able to create any complex object you desire. Happy faking 🥳.
+
+## Create multiple complex objects
+
+Sometimes having a single one of your complex objects isn't enough.
+Imagine having a list view/database of some kind you want to populate:
+
+| ID        | First Name | Last Name |
+| --------- | ---------- | --------- |
+| 6fbe024f… | Tatyana    | Koch      |
+| 862f3ccb… | Hans       | Donnelly  |
+| b452acd6… | Judy       | Boehm     |
+
+The values are directly created using this method:
+
+```ts
+import { faker } from '@faker-js/faker';
+
+function createRandomUser(): User {
+  return {
+    _id: faker.string.uuid(),
+    firstName: faker.person.firstName(),
+    lastName: faker.person.lastName(),
+  };
+}
+
+const users = Array.from({ length: 3 }, () => createRandomUser());
+```
+
+After some time you notice that you need a new column `createdDate`.
+
+You modify the method to also create that:
+
+```ts
+function createRandomUser(): User {
+  return {
+    _id: faker.string.uuid(),
+    firstName: faker.person.firstName(),
+    lastName: faker.person.lastName(),
+    createdDate: faker.date.past(),
+  };
+}
+```
+
+Now let's have a look at our table again:
+
+| ID        | First Name | Last Name | Created Date |
+| --------- | ---------- | --------- | ------------ |
+| 6fbe024f… | Tatyana    | Koch      | 2022-12-28   |
+| 62f3ccbf… | Kacie      | Pouros    | 2023-04-06   |
+| 52acd600… | Aron       | Von       | 2023-05-04   |
+
+Suddenly the second line onwards look different.
+
+Why? Because calling `faker.date.past()` consumes a value from the seed changing all subsequent values.
+
+There are two solutions to that:
+
+1. Set the seed explicitly before creating the data for that row:
+
+```ts
+const users = Array.from({ length: 3 }, (_, i) => {
+  faker.seed(i);
+  return createRandomUser();
+});
+```
+
+Which is very straightforward, but comes at the disadvantage, that you change the seed of your faker instance.
+This might cause issues, if you have lists of groups that contains lists of users. Each group contains the same users because the seed is reset.
+
+2. Derive a new faker instance for each user you create.
+
+```ts
+function createRandomUser(faker: Faker): User {
+  const derivedFaker = faker.derive();
+  return {
+    _id: derivedFaker.string.uuid(),
+    firstName: derivedFaker.person.firstName(),
+    lastName: derivedFaker.person.lastName(),
+    createdDate: derivedFaker.date.past(),
+  };
+}
+
+const users = Array.from({ length: 3 }, () => createRandomUser(faker));
+```
+
+The `faker.derive()` call clones the instance and re-initializes the seed of the clone with a generated value from the original.
+This decouples the generation of the list from generating a user.
+It does not matter how many properties you add to or remove from the `User` the following rows will not change.
+
+::: tip Note
+The following is only relevant, if you want to avoid changing your generated objects as much as possible:
+
+When adding one or more new properties, we recommend generating them last, because if you create a new property in the middle of your object, then the properties after that will still change (due to the extra seed consumption).
+When removing properties, you can continue calling the old method (or a dummy method) to consume the same amount of seed values.
+:::
+
+This also works for deeply nested complex objects:
+
+```ts
+function createLegalAgreement(faker: Faker) {
+  const derivedFaker = faker.derive();
+  return {
+    _id: derivedFaker.string.uuid(),
+    partyA: createRandomUser(derivedFaker),
+    partyB: createRandomUser(derivedFaker),
+  };
+}
+
+function createRandomUser(faker: Faker): User {
+  const derivedFaker = faker.derive();
+  return {
+    _id: derivedFaker.string.uuid(),
+    firstName: derivedFaker.person.firstName(),
+    lastName: derivedFaker.person.lastName(),
+    createdDate: derivedFaker.date.past(),
+    address: createRandomAddress(derivedFaker),
+  };
+}
+
+function createRandomAddress(faker: Faker): Address {
+  const derivedFaker = faker.derive();
+  return {
+    _id: derivedFaker.string.uuid(),
+    streetName: derivedFaker.location.street(),
+  };
+}
+```
+
+::: warning Warning
+Migrating your existing data generators will still change all data once, but after that they are independent.
+So we recommend writing your methods like this from the start.
+:::
+
+::: info Note
+Depending on your preferences and requirements you can design the methods either like this:
+
+```ts
+function createRandomXyz(faker: Faker): Xyz {
+  return {
+    _id: faker.string.uuid(),
+  };
+}
+
+createRandomXyz(faker.derive());
+createRandomXyz(faker.derive());
+createRandomXyz(faker.derive());
+```
+
+or this
+
+```ts
+function createRandomXyz(faker: Faker): Xyz {
+  const derivedFaker = faker.derive();
+  return {
+    _id: derivedFaker.string.uuid(),
+  };
+}
+
+createRandomXyz(faker);
+createRandomXyz(faker);
+createRandomXyz(faker);
+```
+
+The sole difference being more or less explicit about deriving a faker instance (writing more or less code).
+:::
+
+## Create identical complex objects
+
+If you want to create two identical objects, e.g. one to mutate and one to compare it to,
+then you can use `faker.clone()` to create a faker instance with the same settings and seed as the original.
+
+```ts
+const clonedFaker = faker.clone();
+const user1 = createRandomUser(faker);
+const user2 = createRandomUser(clonedFaker);
+expect(user1).toEqual(user2); ✅
+
+subscribeToNewsletter(user1);
+// Check that the user hasn't been modified
+expect(user1).toEqual(user2); ✅
+```
+
+::: info Note
+Calling `faker.clone()` is idempotent. So you can call it as often as you want, it doesn't have an impact on the original faker instance.
+:::
--- a/src/faker.ts
+++ b/src/faker.ts
@@ -430,6 +430,21 @@ export class Faker extends SimpleFaker {
       'This method has been removed. Please use the constructor instead.'
     );
   }
+
+  clone(): Faker {
+    const instance = new Faker({
+      locale: this.rawDefinitions,
+      randomizer: this._randomizer.clone(),
+    });
+    instance.setDefaultRefDate(this._defaultRefDate);
+    return instance;
+  }
+
+  derive(): Faker {
+    const instance = this.clone();
+    instance.seed(this.number.int());
+    return instance;
+  }
 }
 
 export type FakerOptions = ConstructorParameters<typeof Faker>[0];
--- a/test/faker.spec.ts
+++ b/test/faker.spec.ts
@@ -71,12 +71,14 @@ describe('faker', () => {
 
   describe('randomizer', () => {
     it('should be possible to provide a custom Randomizer', () => {
+      const randomizer = {
+        next: () => 0,
+        seed: () => void 0,
+        clone: () => randomizer,
+      };
       const customFaker = new Faker({
         locale: {},
-        randomizer: {
-          next: () => 0,
-          seed: () => void 0,
-        },
+        randomizer,
       });
 
       expect(customFaker.number.int()).toBe(0);
@@ -122,6 +124,60 @@ describe('faker', () => {
 
       const actual = faker.animal.cat();
       expect(actual).toBe('Oriental');
+    });
+  });
+
+  describe('clone()', () => {
+    it('should create a clone that returns the same values as the original', () => {
+      const clone1 = faker.clone();
+      const clone2 = faker.clone();
+      const clone3 = clone1.clone();
+
+      expect(clone1).not.toBe(faker);
+      expect(clone2).not.toBe(faker);
+      expect(clone3).not.toBe(faker);
+      expect(clone1).not.toBe(clone2);
+      expect(clone1).not.toBe(clone3);
+      expect(clone2).not.toBe(clone3);
+
+      const value0 = faker.number.int();
+      expect(clone1.number.int()).toBe(value0);
+      expect(clone2.number.int()).toBe(value0);
+      expect(clone3.number.int()).toBe(value0);
+
+      const value1 = clone1.number.int();
+      expect(faker.number.int()).toBe(value1);
+      expect(clone2.number.int()).toBe(value1);
+      expect(clone3.number.int()).toBe(value1);
+
+      const value2 = clone2.number.int();
+      expect(clone1.number.int()).toBe(value2);
+      expect(faker.number.int()).toBe(value2);
+      expect(clone3.number.int()).toBe(value2);
+
+      const value3 = clone3.number.int();
+      expect(clone1.number.int()).toBe(value3);
+      expect(clone2.number.int()).toBe(value3);
+      expect(faker.number.int()).toBe(value3);
+    });
+  });
+
+  describe('derive()', () => {
+    it("should create a derived faker, that doesn't affect the original", () => {
+      const seed = faker.seed();
+      faker.number.int();
+      const value = faker.number.int();
+
+      faker.seed(seed);
+      const derived = faker.derive();
+
+      expect(derived).not.toBe(faker);
+
+      for (let i = 0; i < derived.number.int(100); i++) {
+        derived.number.int();
+      }
+
+      expect(faker.number.int()).toBe(value);
     });
   });
 
--- a/docs/guide/randomizer.md
+++ b/docs/guide/randomizer.md
@@ -104,13 +104,19 @@ export function generatePureRandRandomizer(
   seed: number | number[] = Date.now() ^ (Math.random() * 0x100000000),
   factory: (seed: number) => RandomGenerator = xoroshiro128plus
 ): Randomizer {
-  const self = {
-    next: () => (self.generator.unsafeNext() >>> 0) / 0x100000000,
-    seed: (seed: number | number[]) => {
-      self.generator = factory(typeof seed === 'number' ? seed : seed[0]);
-    },
-  } as Randomizer & { generator: RandomGenerator };
-  self.seed(seed);
-  return self;
+  function wrapperFactory(generator?: RandomGenerator): Randomizer {
+    const self = {
+      next: () => (self.generator.unsafeNext() >>> 0) / 0x100000000,
+      seed: (seed: number | number[]) => {
+        self.generator = factory(typeof seed === 'number' ? seed : seed[0]);
+      },
+      clone: () => wrapperFactory(self.generator.clone()),
+    } as Randomizer & { generator: RandomGenerator };
+    return self;
+  }
+
+  const randomizer = wrapperFactory();
+  randomizer.seed(seed);
+  return randomizer;
 }
 ```
--- a/src/internal/mersenne.ts
+++ b/src/internal/mersenne.ts
@@ -83,6 +83,22 @@ export class MersenneTwister19937 {
   private mti = this.N + 1; // mti==N+1 means mt[N] is not initialized
 
   /**
+   * Creates a new instance of MersenneTwister19937.
+   *
+   * @param options The required options to initialize the instance.
+   * @param options.mt The state vector to use. The array will be copied.
+   * @param options.mti The state vector index to use.
+   */
+  constructor(options?: { mt: number[]; mti: number }) {
+    if (options != null && 'mt' in options) {
+      this.mt = [...options.mt];
+      this.mti = options.mti;
+    } else {
+      this.initGenrand(Date.now() ^ (Math.random() * 0x100000000));
+    }
+  }
+
+  /**
    * Returns a 32-bits unsigned integer from an operand to which applied a bit
    * operator.
    *
@@ -166,11 +182,11 @@ export class MersenneTwister19937 {
   /**
    * Initialize by an array with array-length.
    *
-   * @param initKey is the array for initializing keys
-   * @param keyLength is its length
+   * @param initKey Is the array for initializing keys.
    */
-  initByArray(initKey: number[], keyLength: number): void {
+  initByArray(initKey: number[]): void {
     this.initGenrand(19650218);
+    const keyLength = initKey.length;
     let i = 1;
     let j = 0;
     let k = this.N > keyLength ? this.N : keyLength;
@@ -239,11 +255,6 @@ export class MersenneTwister19937 {
     if (this.mti >= this.N) {
       // generate N words at one time
       let kk: number;
-
-      // if initGenrand() has not been called a default initial seed is used
-      if (this.mti === this.N + 1) {
-        this.initGenrand(5489);
-      }
 
       for (kk = 0; kk < this.N - this.M; kk++) {
         y = this.unsigned32(
@@ -324,6 +335,10 @@ export class MersenneTwister19937 {
     return (a * 67108864.0 + b) * (1.0 / 9007199254740992.0);
   }
   // These real versions are due to Isaku Wada, 2002/01/09
+
+  clone(): MersenneTwister19937 {
+    return new MersenneTwister19937({ mt: this.mt, mti: this.mti });
+  }
 }
 
 /**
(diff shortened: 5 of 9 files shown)

### CODE REVIEW

File: src/internal/mersenne/mersenne.ts [hunk 1]
```diff
@@ -33,6 +38,15 @@ export default function mersenne(): Mersenne {
 
   twister.initGenrand(Math.ceil(Math.random() * Number.MAX_SAFE_INTEGER));
 
+  return wrap(twister);
+}
+
+/**
+ * Wraps the twister instance to provide a more convenient API.
+ *
+ * @param twister THe twister instance to wrap.
```
Shinigami92:
```suggestion
 * @param twister The twister instance to wrap.
```

src/internal/mersenne/twister.ts
```diff
@@ -318,4 +318,11 @@ export default class MersenneTwister19937 {
     return (a * 67108864.0 + b) * (1.0 / 9007199254740992.0);
   }
   // These real versions are due to Isaku Wada, 2002/01/09
```
Shinigami92: This feels like it's detached now to whatever it was related before

ST-DDT: It related to the above methods above `genrandReal1-3` + ...

Shinigami92 (dismissed):

src/faker.ts
```diff
@@ -300,4 +307,53 @@ export class Faker {
...
+   * @example
+   * faker.seed(42);
+   * faker.datatype.number(10); // 4
+   * faker.datatype.number(10); // 8
+   * faker.seed(42);
+   * const derived = faker.derive();
+   * const firstName = derived.person.firstName(); // 'Lavina'
+   * const lastName = derived.person.lastName(); // 'Kuhic'
+   * // It doesn't matter how many calls to derived are executed
+   * faker.datatype.number(10); // 8 <- This is same as before
```
xDivisionByZerox: I cant comprehend what the difference to `fork` would be here even with the `@see` description. What would be different if I do the same thing:
```ts
faker.seed(42);
faker.datatype.number(10); // 4
faker.datatype.number(10); // 8
faker.seed(42);
const forked = faker.fork(); // what is the difference here?
const firstName = forked.person.firstName(); // 'Lavina' (???)
const lastName = forked.person.lastName(); // 'Kuhic' (???)

faker.datatype.number(10); // 8 <- This is same as before (???)
```

I'm sorry for being stupid but I starred at this code for such a long time and cant figure it out.

xDivisionByZerox: Nvm, I should have looked at the tests first. They helped me to understand the difference. But just to be sure:
- fork => different instances, same seed, same result for same call chains
- derive => different instances, different seeds, different results for same call chains

Did I understand correctly? If yes, why would we need derive instead of "just" calling `new Faker(options)`?

xDivisionByZerox (dismissed):

src/faker.ts
ST-DDT: Faker.fork creates a new instance with the exact same state as the current instance. Thus resulting in the exact same values being returned from both instances when calling the same methods.

Faker.derive creates an new instance by taking a single value from the old instance.
The state of the derived instance is dependent on the original instance's state, thus you can generate an entire person for a given seed while only taking a single value from the original instance (this doesnt change even if you add or remove properties to/from the person). Thus the second and any following generated person remains mostly the same even though some properties have been added/removed.
You will still see changes in each person, but only for those parts that you actually changed (e.g. the last name changes if the middle name is now added), but not because the previous person now has a middle name.

ST-DDT: Btw. I'm no longer sure there being a real usecase for Faker.fork.

xDivisionByZerox: > The state of the derived instance is dependent on the original instance's state, thus you can generate an entire person for a given seed while only taking a single value from the original instance [...]

This was what made it really clear to me. Thank you.

xDivisionByZerox (dismissed):

test/faker.spec.ts
```diff
@@ -195,4 +195,18 @@ describe('faker', () => {
       expect(faker.number.int()).toBe(value);
     });
   });
+
+  describe('defaultRefDate', () => {
+    it('should be a defined', () => {
```
xDivisionByZerox, line 185:
```suggestion
    it('should be defined', () => {
```

xDivisionByZerox: Oh sorry that was there merge change I commented on.

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

src/faker.ts
```diff
@@ -351,6 +399,55 @@ export class Faker {
...
+  fork(): Faker {
+    return new Faker({
+      locale: this.definitions,
+      defaultRefDate: this.defaultRefDate,
+      mersenne: this._mersenne.fork(),
+    });
+  }
+
+  /**
+   * Derives a new Faker instance from the current one.
```
matthewmayer: Can we find a way to render this in the api documentation? At the moment methods on the root don't appear in the docs. Or should that be seperate PR?

ST-DDT: IMO we should create a separate PR to add the Faker (setSeed,...) and potentially utility (mergeLocales) methods to the API docs.

ST-DDT: I created https://github.com/faker-js/faker/pull/1940 for that.

src/faker.ts
```diff
@@ -135,6 +135,16 @@ export class Faker {
...
+    /**
+     * The initial seed to use for this instance.
+     */
+    seed?: number;
+    /**
+     * The default reference date to use for this instance.
+     */
+    defaultRefDate?: string | Date | number | (() => Date);
+    /** @internal */
+    mersenne?: Mersenne;
```
xDivisionByZerox: I'm not a big fan of putting this in the constructor AND marking it as `internal`. Can you explain this to me?

ST-DDT: For `fork()` we have to create a copy of the mersenne instance. Alternatively we need copy/paste state methods on the mersenne interface.

xDivisionByZerox: Ok, but why `@internal`? Can we not simply mark this as a feature, that you can supply your own "RNG" implementation which follows the Mersenne interface?

ST-DDT: Theoretically yes. But then we should rename the Mersenne interface accordingly.

ST-DDT: Probably in a different PR.

xDivisionByZerox:
I'm marking this with `s: needs decision` so we can talk about implementation details in tomorrows meeting.

ST-DDT:
**Team Decision**

Remove the mersenne option from the constructor 
and use something like `newFaker.mersenne.copyState(this.mersenne)` instead.

Shinigami92:
> **Team Decision**
> 
> Remove the mersenne option from the constructor and use something like `newFaker.mersenne.copyState(this.mersenne)` instead.

I was not part of the meeting but agree for now
Yes this makes it a bit more complex on our maintainers side, but I like the way it doesn't bleed internal value inside a constructor 👍

ST-DDT:
Done

src/faker.ts
```diff
@@ -530,6 +530,74 @@ export class Faker {
       'This method has been removed. Please use the constructor instead.'
     );
   }
+
+  /**
+   * Creates a new instance of faker with the same state as this instance.
```
```suggestion
   * Creates a new instance of Faker with the same state as this instance.
```

I would name it with capital F, because Faker is the class and faker would be a variable/instance of Faker

src/internal/mersenne/twister.ts
```diff
@@ -322,4 +322,16 @@ export default class MersenneTwister19937 {
...
+
+  backupState(): { mt: number[]; mti: number } {
```
Shinigami92: I would feel safer if we explicitly use `readonly number[]` (or `ReadonlyArray<number>`)

```suggestion
  backupState(): { mt: readonly number[]; mti: number } {
```

Not sure if then something needs to be adjusted, but I assume no 🙂

src/internal/mersenne/twister.ts
```diff
@@ -322,4 +322,16 @@ export default class MersenneTwister19937 {
...
+    return {
+      mt: this.mt.slice(),
+      mti: this.mti,
+    };
+  }
+
+  restoreState(state: { mt: number[]; mti: number }): void {
```
Shinigami92:
```suggestion
  restoreState(state: { mt: readonly number[]; mti: number }): void {
```

Shinigami92 (commented):
LGTM, but I will have a second look later when I'm not on phone screen

src/faker.ts
```diff
@@ -581,52 +532,71 @@ export class Faker {
...
    * faker.number.int(10); // 4 (1st call)
    * faker.number.int(10); // 8 (2nd call)
+   *
    * faker.seed(42);
+   * // Creates a new instance with a seed generated from the current instance
    * const derived = faker.derive(); // (1st call)
-   * const firstName = derived.person.firstName(); // 'Lavina'
-   * const lastName = derived.person.lastName(); // 'Kuhic'
+   * // The derived instance will produce values dependent on the state of the original instance at the time of the derive call
+   * derived.number.int(10); // 7 (derived 1st call)
```
xDivisionByZerox: I would expect `8` as the "first" derived call here instead of `7`. :thinking:

ST-DDT: No, the derived instance is seeded from the main instance, but this doesn't make the derived instance return the same value as the original.
It is intentional, that the derived instance does not match the original instance with a simple call offset.
Is my explanation understandable?

xDivisionByZerox: Yes, that clarifies it. The thing is, I thought that "the derived instance does match the original instance with a simple call offset". But then I understand.

xDivisionByZerox (dismissed):

xDivisionByZerox:
If we merge this, we should unify all our methods to only use up one seed per call. So either merge and refactor pre-v8 or later. But I think it would be a good idea to release both things (fork/derive + one seed per method call) in the same version.

ST-DDT:
Sounds reasonable.
I'm still not sure whether we should explicitly expose this or only implicitly support this via the tree shaking code. (Because creating a new faker instance is kind of expensive)

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/2092

xDivisionByZerox (dismissed):

xDivisionByZerox:
I'm adding the `do NOT merge yet` label, as I don't want this to accidentally slip into v8.0. I'd much rather have an explicit version that focuses on introducing this feature as already stated in https://github.com/faker-js/faker/pull/1499#issuecomment-1514603969.

ST-DDT:
**Team Decision**

@xDivisionByZerox will do some performance benchmarking regarding internal derivation functions. (1 seed usage per call)
If there is negligible performance impact this will be in v8.x. Otherwise this will be done in v9.0 as part of the treeshaking refactor.

ST-DDT:
**Team decision**

Rename `fork` -> `clone`

ST-DDT:
Ready for review

xDivisionByZerox (dismissed):
Great documentation, @ST-DDT! Really enjoyed reading through it 😄

docs/guide/usage.md
```diff
@@ -261,3 +261,178 @@ We will update these docs once a replacement is available.
...
+function createRandomXyz(faker: Faker): Xyz {
+  const derived = faker.derive();
+  return {
+    _id: derived.string.uuid(),
+  };
+}
+
+createRandomXyz(faker);
+createRandomXyz(faker);
+createRandomXyz(faker);
```
ST-DDT: Which of these two variants should we use for our documentation?

xDivisionByZerox: You made the point about being explicit with deriving the faker instance. SO I'd say the first option.
But to be sure: You talk about documentation at other parts than this?

ST-DDT: > You made the point about being explicit with deriving the faker instance. SO I'd say the first option.

In the future with the tree shaking refactor, you will no longer need to call it explicitly because it will be called automatically.
(Probably)

> But to be sure: You talk about documentation at other parts than this?

Yes.

xDivisionByZerox: > In the future with the tree shaking refactor, you will no longer need to call it explicitly because it will be called automatically.

While that is true for all functions provided by Faker itself, you are still required to call it on the use case you documented. Additionally, custom generator libraries (for features we won't implement) might require it as well.

docs/guide/usage.md
```diff
@@ -261,3 +261,178 @@ We will update these docs once a replacement is available.
...
+  };
+}
+```
+
+:::warning Warning
+Migrating your existing data to either of these variants, will still change all data, but after that they are independent.
+So we recommend writing your methods like this from the start.
+:::
+
+::: info Note
```
Shinigami92: in the one you have no whitespace but the other has a whitespace :thinking:

docs/guide/usage.md
Shinigami92: > Which of these two variants should we use for our documentation?

I instantly thought while reading at line [346](https://github.com/faker-js/faker/pull/1499/files#diff-ebc191ae76e5fef17b9b552067d614e7a84ad748c0d917f2486f9a61b5b8adfdR346) "why do we pass `faker.derive()` instead of calling it inside?"

So yeah, my answer would be: **I personally highly prefer the "call inside" variant**, but then it is okay to point out that you could also pass it manually

docs/guide/usage.md
```diff
@@ -261,3 +261,178 @@ We will update these docs once a replacement is available.
...
+The sole difference being more or less explicit about deriving a faker instance (writing more or less code).
+:::
+
+### Create identical complex objects
+
+If you want to create two identical objects, e.g. one to mutate and one to compare it to,
+then you can use `faker.clone()` to create a faker instance with the exact same settings and seed as the original.
+
+```ts
+const fakerClone = faker.clone();
```
Shinigami92:
```suggestion
const clonedFaker = faker.clone();
```

I think this reads better :thinking:

docs/guide/usage.md
```diff
@@ -261,3 +261,178 @@ We will update these docs once a replacement is available.
...
+const user2 = createRandomUser(fakerClone);
+expect(user1).toEqual(user2); ✅
+
+subscribeToNewsletter(user1);
+// Check that the user hasn't been modified
+expect(user1).toEqual(user2); ✅
+```
+
+:::info Note
+Calling `faker.clone()` is idempotent. So you can call it as often as you want, it doesn't have an impact on the faker instance.
```
Shinigami92:
```suggestion
Calling `faker.clone()` is idempotent. So you can call it as often as you want, it doesn't have an impact on the original faker instance.
```

docs/guide/usage.md
```diff
@@ -261,3 +261,178 @@ We will update these docs once a replacement is available.
...
+
+```ts
+const fakerClone = faker.clone();
+const user1 = createRandomUser(faker);
+const user2 = createRandomUser(fakerClone);
+expect(user1).toEqual(user2); ✅
+
+subscribeToNewsletter(user1);
+// Check that the user hasn't been modified
+expect(user1).toEqual(user2); ✅
```
Shinigami92, line 457: Somehow this confuses me more than it brings value to me :thinking:  
What does this have to do with the example above showing what clone is for?

src/internal/mersenne/mersenne.ts
```diff
@@ -18,8 +18,19 @@ export interface Mersenne {
...
+   * Copies the state of the given mersenne instance to this instance.
+   *
+   * @param other The mersenne instance to copy the state from.
+   */
+  copyStateFrom(other: Mersenne): void;
 }
 
+type MersenneImpl = Mersenne & {
+  twister: Twister;
+};
```
Shinigami92: Why this? Instead of just adding `twister: Twister;` to the `Mersenne` interface?

src/internal/mersenne/twister.ts
```diff
@@ -322,4 +322,16 @@ export default class MersenneTwister19937 {
     return (a * 67108864.0 + b) * (1.0 / 9007199254740992.0);
   }
   // These real versions are due to Isaku Wada, 2002/01/09
+
+  backupState(): { mt: ReadonlyArray<number>; mti: number } {
```
Shinigami92: Do we want to define a `TwisterState` interface for this?

test/faker.spec.ts
```diff
@@ -108,6 +108,60 @@ describe('faker', () => {
...
+      const clone3 = clone1.clone();
+
+      expect(clone1).not.toBe(faker);
+      expect(clone2).not.toBe(faker);
+      expect(clone3).not.toBe(faker);
+      expect(clone1).not.toBe(clone2);
+      expect(clone1).not.toBe(clone3);
+      expect(clone2).not.toBe(clone3);
+
+      const valueOrg = faker.number.int();
```
Shinigami92: This doesn't have something to do with `Organization`, does it?

I think you mean something like `originalValue`, did you?

docs/guide/usage.md
ST-DDT: Fixed

docs/guide/usage.md
ST-DDT: Well this demonstrates exactly what clone is for.
Creating 2 identical instances, (not) mutating one of them and and compare the first with the later.
(I'm not exactly sure what `clone()` is for, except for this case, which I personally wouldn't use/implement like this.)

Do you have a suggestion for a better example?

Shinigami92: I hope I didn't confused you, I just talked about line 431 432 and 433
everything above is enough in my opinion

Line 429 already compares the equality

---

Now as I re-read that line, why even use `expect` and `toEqual` here?
I think this is not bound to tests, so we should use something like `JSON.stringify(user1) === JSON.stringify(user2)` here

docs/guide/usage.md
ST-DDT: Fixed

docs/guide/usage.md
ST-DDT: Sure, you can create two identical objects, but why would you need that?
Obviously you want to do something with it, but what?

src/internal/mersenne/mersenne.ts
ST-DDT: Because Mersenne will eventually become our public RNG interface and then the internal shouldn't be accessible.

src/internal/mersenne/twister.ts
ST-DDT: We could, but I tried to change as little as possible in this file.

docs/guide/usage.md
Shinigami92: I don't understand any from and below `subscribeToNewsletter`

Assuming `subscribeToNewsletter` is implemented in a dump way
```js
function subscribeToNewsletter(user) {
	user.prop = 1;
}
```

The ✅ in the line after that is totally wrong and invalid.
But instead of trying to find something valuable in that statement, why even bother?
It is not in our hands what the dev does or wants to do with the cloned Faker instance and their generated values

src/internal/mersenne/mersenne.ts
Shinigami92: So why not using `@internal` in that case?!

src/internal/mersenne/twister.ts
Shinigami92: @xDivisionByZerox what's your opinion?
I would prefer to have a clean way :thinking: and not a "I'm afraid to touch that" of a source that we already moved multiple times around, formatted, and did other stuff to it.

---

I also don't really understand the argument at all, because if you don't want to touch this, why don't you then move it out into a function like:

```js
backupState(twisterIntance): TwisterState;

restoreState(twisterIntance, state: TwisterState): void
```

and then use `@ts-expect-error: ignore private`?

docs/guide/usage.md
ST-DDT: > Assuming `subscribeToNewsletter` is implemented in a dump way

The tests `expect` the method to not do that. So you would fail the test if you do.
Maybe I could put a different marker there.



> But instead of trying to find something valuable in that statement, why even bother?

If we are unable to find/describe a usecase for that method, we shouldn't add it to faker.

Shinigami92: Yes, I agree, lets do that.
I requested anyway in last meeting, to split this features into two separate PRs :shrug: 

I currently do not see the need for clone, where as derive does absolutely makes sense to me, as it only claims one seed value and your example with the database table makes absolutely sense and even we could use derive in our own codebase later on
But clone? it just confuses me more and more

src/internal/mersenne/twister.ts
ST-DDT: I found a different solution to this issue.

src/internal/mersenne/mersenne.ts
ST-DDT: I found a different solution to this issue.

docs/guide/usage.md
ST-DDT: Let's discuss this in the next team meeting.

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/2284

- https://github.com/faker-js/faker/pull/2284

src/internal/mersenne/twister.ts
ST-DDT: I'm planning on using:

- Mersenne#clone()
- Mersenne#derive()

instead

Shinigami92:
This would solve the problem of copycat: https://github.com/snaplet/copycat#the-problem

ST-DDT:
> This would solve the problem of copycat: [snaplet/copycat#the-problem](https://github.com/snaplet/copycat#the-problem)

I would recommend using a custom `Randomizer` (!=Mersenne) e.g. from pure-random, that has shorter initialization times.
The Mersenne Twister is optimized to generate multiple random values.

This will also (potentially) get much more easier with v9's independent methods that can be called just like this `firstName(fakerCore, ...args)` that doesn't need a full faker instance anymore.

ST-DDT:
Closing as obsolete

PR #1511, opened 2022-11-03
chore: remove broken source urls found in some locales

subbe:
Updated date files to remove the broken source links found in the following locales:

- ar
- az
- cz
- de
- en
- es
- fa
- fr
- fr_CH
- he
- hr
- mk
- nl
- pt_BR
- pt_PT
- ru
- sv
- vi


src/locales/es/date/month.ts
```diff
@@ -1,4 +1,3 @@
-// Sources: https://www.unicode.org/cldr/cldr-aux/charts/28/summary/es.html and https://www.wikilengua.org/index.php/Abreviaciones_en_fechas
```
ST-DDT, line 1: The later link seems to still work.

src/locales/es/date/weekday.ts
```diff
@@ -1,4 +1,3 @@
-// Sources: https://www.unicode.org/cldr/cldr-aux/charts/28/summary/es.html and https://www.wikilengua.org/index.php/Abreviaciones_en_fechas
```
ST-DDT, line 1: The later link seems to still work.

ST-DDT (approved):

ejcheng (approved):

xDivisionByZerox:
I mean, the source comment was there for a reason...if we remove it from the locales the maintainer team should decide on an alternative strategy to provide intel on how faker data is populated (what qualifies to be taken into a source set).

ST-DDT:
@xDivisionByZerox It is not about removing all source links. These links are broken and I don't have any hope of any of them coming back online.

xDivisionByZerox:
> @xDivisionByZerox It is not about removing all source links. These links are broken and I don't have any hope of any of them coming back online.

Ah, ok, my bad. That intend wasn't clear for me from the PRs description.

ST-DDT:
> > @xDivisionByZerox It is not about removing all source links. These links are broken and I don't have any hope of any of them coming back online.
> 
> Ah, ok, my bad. That intend wasn't clear for me from the PRs description.

I updated it to make it more clear.

xDivisionByZerox (approved):

matthewmayer:
Rather than completely remove these comments, perhaps it could just be replaced with 

`Source: Unicode CLDR, version 27 https://unicode.org/Public/cldr/27/`

That keeps the data attributed, and if someone wants to come along and update to a newer version of CLDR later, they can.

pkuczynski:
> Rather than completely remove these comments, perhaps it could just be replaced with
> 
> `Source: Unicode CLDR, version 27 https://unicode.org/Public/cldr/27/`

Good idea!

ST-DDT:
My main problem with that is that it only links to a few zip files and it is not clear which file to actually look into. So at least we have to point to a specific zip with the actual file inside the zip added in parenthesis.
But if we can keep the link working then that would be good.

matthewmayer:
more recent versions of CLDR are in Github, e.g. https://github.com/unicode-org/cldr/blob/main/common/main/de.xml

so someone could probably go cross-check and restore correct line numbers later (or write a script to grab data directly from CLDR).

ST-DDT:
> so someone could probably go cross-check and restore correct line numbers later (or write a script to grab data directly from CLDR).

IMO this should be done immediately using githup perma links.
However, I would be fine for a working link into the old release zips, since the actual data are probably from those versions as well.

But we should keep that idea in mind/create an issue for updating the calendar entries if possible.

matthewmayer:
im pretty sure a lot of the line numbers are wrong anyway, for example many languages have a comment with `en.xml` but it shouldn't be en.xml it should be sv.xml, mk.xml etc depending on the language. Probably some copy-pasting between locales.

pkuczynski:
I would use the links to GitHub and create a ticket to fetch it via script scheduled in CI

matthewmayer:
if someone wants to take that up in future note there's a JSON version generated from the XML versions which is probably easier to consume. https://github.com/unicode-org/cldr-json

Shinigami92 (approved):

PR #1512, opened 2022-11-03
feat(git): add options.refDate to commitEntry

KevinMind:
closes: https://github.com/faker-js/faker/pull/1512

### DIFF
--- a/src/modules/git/index.ts
+++ b/src/modules/git/index.ts
@@ -37,6 +37,8 @@ export class GitModule {
    * 'LF' = '\n',
    * 'CRLF' = '\r\n'
    *
+   * @param options.refDate The date to use as reference point for the commit. Defaults to now.
+   *
    * @example
    * faker.git.commitEntry()
    * // commit fe8c38a965d13d9794eb36918cb24cebe49a45c2
@@ -51,11 +53,13 @@ export class GitModule {
     options: {
       merge?: boolean;
       eol?: 'LF' | 'CRLF';
+      refDate?: string | Date | number;
     } = {}
   ): string {
     const {
       merge = this.faker.datatype.number({ min: 0, max: 4 }) === 0,
       eol = 'CRLF',
+      refDate,
     } = options;
 
     const lines = [`commit ${this.faker.git.commitSha()}`];
@@ -66,7 +70,7 @@ export class GitModule {
 
     lines.push(
       `Author: ${this.faker.person.firstName()} ${this.faker.person.lastName()} <${this.faker.internet.email()}>`,
-      `Date: ${this.faker.date.recent().toString()}`,
+      `Date: ${this.faker.date.recent(1, refDate).toString()}`,
       '',
       `\xa0\xa0\xa0\xa0${this.commitMessage()}`,
       // to end with a eol char
--- a/test/git.spec.ts
+++ b/test/git.spec.ts
@@ -5,6 +5,8 @@ import { seededTests } from './support/seededRuns';
 
 const NON_SEEDED_BASED_RUN = 5;
 
+const refDate = '2020-01-01T00:00:00.000Z';
+
 describe('git', () => {
   afterEach(() => {
     faker.locale = 'en';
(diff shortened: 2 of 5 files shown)


KevinMind:
Seems there is a lagging issue with `faker.date.recent()` still using slightly varying timestamps when given `(1, 'static-date-string')`. I'm not totally sure how to work around this. Will have to dig a bit deeper into the date module.

edit: It seems like the `recent()` method is still using something like current time for the time portion of the returned date.

KevinMind:
Here's an example in a storybook where I tried testing the implementation. The time portion is ticking away.

https://user-images.githubusercontent.com/19595165/199698083-90ae20c0-21af-422e-8421-1e4ea406002d.mov

ST-DDT:
I'm not sure why the story-book has that issue (do you use new `Date()` or new `Date('something-fixed')`, but the tests that you have added show, that it is working as expected.

test/git.spec.ts
```diff
@@ -13,8 +14,11 @@ describe('git', () => {
   seededTests(faker, 'git', (t) => {
     t.itEach('branch', 'commitMessage', 'commitSha', 'shortSha');
 
-    // The timestamp is not fixed, so we can't compare it
-    t.todo('commitEntry');
+    t.describe('commitEntry', (t) => {
+      t.it('with only string refDate', { refDate })
+        .it('with only Date refDate', { refDate: new Date(refDate) })
+        .it('with value', { refDate });
```
ST-DDT: This is the same as the first entry. Looks like a copy and paste error from the date tests, that have an additional `value` parameter.

https://github.com/faker-js/faker/blob/4204a6ae925a4e0c761ece10ad21222dac7a9e7e/test/date.spec.ts#L28

src/modules/git/index.ts
```diff
@@ -37,6 +37,7 @@ export class GitModule {
    * 'LF' = '\n',
    * 'CRLF' = '\r\n'
    *
+   * @param options.refDate The date to use as reference point for the commit. Defaults to now.
```
ST-DDT, line 40: I'm not sure why it doesn't show a hint here, but IMO the empty line should be between the param and example block:

```suggestion
   * @param options.refDate The date to use as reference point for the commit. Defaults to now.
   *
```

test/git.spec.ts
KevinMind: Good catch. I noticed that the method accepts a numeric input as well. So I tested with `new Date(refDate).getTime()` as well. Added that test to the date module as well. Does that Make sense to cover?

KevinMind:
> I'm not sure why the story-book has that issue (do you use new `Date()` or new `Date('something-fixed')`, but the tests that you have added show, that it is working as expected.

Is there any way to publish canaries for faker? I would be able to commit the case where it is not working to share if so.

ST-DDT (dismissed):

ST-DDT:
The code looks good, but it looks like you have to update the test snapshots (and remove the obsolete ones).

ST-DDT (dismissed):

Shinigami92 (dismissed):
Reminder to improve date.recent in
- https://github.com/faker-js/faker/issues/1349

test/git.spec.ts
```diff
@@ -13,8 +14,17 @@ describe('git', () => {
...
-    // The timestamp is not fixed, so we can't compare it
-    t.todo('commitEntry');
+    for (let x = 0; x < NON_SEEDED_BASED_RUN; x++) {
+      console.log(x, faker.git.commitEntry({ refDate }));
+    }
+
+    t.describe('commitEntry', (t) => {
+      t.it('with only string refDate', { refDate })
+        .it('with only Date refDate', { refDate: new Date(refDate) })
+        .it('with only numberic refDate', {
```
ejcheng:
```suggestion
        .it('with only number refDate', {
```

test/git.spec.ts
```diff
@@ -4,6 +4,7 @@ import { faker } from '../src';
 import { seededTests } from './support/seededRuns';
 
 const NON_SEEDED_BASED_RUN = 5;
+const refDate = '2020-01-01T00:00:00.000Z';
```
ejcheng, line 8:
```suggestion

const refDate = '2020-01-01T00:00:00.000Z';
```
I would prefer an empty line here because `NON_SEEDED_BASED_RUN` isn't really related to `refDate`

KevinMind:
@import-brain @ST-DDT @Shinigami92 rebased and updated. If 👍 Then I can merge?

ST-DDT:
Since you changed the PR we have to review it again. Please note that only we can press the merge button.

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #1515, opened 2022-11-03
docs(helpers): extend module jsdocs

ST-DDT:
Extend the helpers module jsdocs with a hint that these methods are intended to be used to build other faker methods.
Not sure about the exact wording.

As decided here: https://github.com/faker-js/faker/pull/1486#issuecomment-1302365569


src/modules/helpers/index.ts
```diff
@@ -6,7 +6,8 @@ import * as uniqueExec from './unique';
 
 /**
  * Module with various helper methods that transform the method input rather than returning values from locales.
- * The transformation process may call methods that use the locale data.
+ * The transformation process may call methods that use the locale data, but the methods themselves are not locale data dependent.
+ * The methods are intended to be used by other methods both in this library and in custom methods.
```
ejcheng:
```suggestion
 * The transformation process may call methods that use locale data, but the methods themselves are not locale-dependent.
 * These methods are intended to be used by other methods both in this library and in custom methods.
```

src/modules/helpers/index.ts
```diff
@@ -6,7 +6,8 @@ import * as uniqueExec from './unique';
 
 /**
  * Module with various helper methods that transform the method input rather than returning values from locales.
- * The transformation process may call methods that use the locale data.
+ * The transformation process may call methods that use locale data, but the methods themselves are not locale-dependent.
```
xDivisionByZerox: I guess that's straight-up wrong, no? At most methods from `BaseFaker` should be called. But I think we can simply remove this lie as it is more or less a duplicated statement of the line aboth.

ST-DDT: I don't think it is wrong.
For the fake method this could be argued about: It only uses locale data, but is not locale-dependent.

> At most methods from BaseFaker should be called.

This would also only be true for all methods except fake. IMO this would be more an implementation detail/hint instead of a hint for users.

I have to think about this some more (e.g. removing the locale hints altogether or just rephrasing them).

> Module with various helper methods providing basic (seed-dependent) operations that are useful for implementing faker methods.

Suggestions are welcome.

PS: The more I think about what this module does, the less I think `fake` belongs here, but I don't think we should move it again any time soon.

ST-DDT: I simplified the text. Please check again.

Shinigami92 (approved):

ejcheng (approved):

Issue #1105, opened 2022-06-24
Weird email and username in Chinese locale package


shtse8:
### Describe the bug

email and username should not using Chinese even in Chinese locale package.
there is no one using Chinese as an email and username even in Chinese.

### Reproduction

# code
```ts
// import { faker } from '@faker-js/faker';
import { faker } from '@faker-js/faker/locale/zh_CN'

export const USERS: User[] = []

export function createRandomUser(): User {
  return {
    userId: faker.datatype.uuid(),
    username: faker.internet.userName(),
    email: faker.internet.email(),
    avatar: faker.image.avatar(),
    password: faker.internet.password(),
    birthdate: faker.date.birthdate(),
    registeredAt: faker.date.past(),
  }
}

Array.from({ length: 1 }).forEach(() => {
  USERS.push(createRandomUser())
})

console.log(USERS)
```

# output
```
[
  {
    userId: '88d30bb6-c783-4e56-8ffc-6778ec6e1c0a',
    username: '钰轩.侯68',
    email: '明杰_彭@gmail.com',
    avatar: 'https://cloudflare-ipfs.com/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/765.jpg',
    password: 'UdVxsDkMWFajEId',
    birthdate: 1964-10-12T19:43:31.378Z,
    registeredAt: 2022-04-27T11:56:33.741Z
  }
]
```

Shinigami92:
https://en.wikipedia.org/wiki/International_email#Email_addresses 🤔

ST-DDT:
@shtse8 Is there a "trivial" way to translate/transcribe chinese words to english letters?

shtse8:
> @shtse8 Is there a "trivial" way to translate/transcribe Chinese words to English/Latin letters?
> 
> Are usernames also in Latin letters? I think I have seen mostly Chinese usernames (display names) in Chinese forums (In the few I have ever visited).

because there is not possible to use Chinese in email and username most of the time on any site. also, it's much easier to enter in English which can be directly from keyboard - one char by one char. 

[...]

PR #1520, opened 2022-11-04
fix: ASCII-fy `domainWord()`

schw4rzlicht:
The `en` locale includes the word "jalapeño" which is the only one with a special char. Imo this is inconsistent and it leads to other problems as well, see https://github.com/faker-js/faker/issues/1105#issuecomment-1186630351 for example.

~This PR removes the word from the locale.~

This PR uses `faker.helpers.slugify()` to remove non-ASCII characters from `domainWord()`.


ST-DDT:
AFAICT: `jalapeño` is a proper English word, isn't it?
For #1105 I think we should rather ansify these characters.

schw4rzlicht:
> AFAICT: `jalapeño` is a proper English word, isn't it?
> For #1105 I think we should rather ansify these characters.

Yes, might make more sense. I altered the Regex that filteres illegal characters in `Internet.domainWord()`, maybe that's what helps?

matthewmayer:
if https://github.com/faker-js/faker/pull/1554 gets merged then you would be able to use the new version of `faker.helpers.slugify` to turn `jalapeño` to `jalapeno` automatically in domainWord (as domainWord seems to be a slight variant of slugify really).

schw4rzlicht:
> if #1554 gets merged then you would be able to use the new version of `faker.helpers.slugify` to turn `jalapeño` to `jalapeno` automatically in domainWord (as domainWord seems to be a slight variant of slugify really).

From what I've seen this is right.

But that would mean:

- in order to fix that our team has to change the usage of faker for ~2000 test cases. And I don't think we're alone with this.
- we cannot call `faker.internet.domainWord()`/`faker.internet.domainName()`/`faker.internet.url()` without using `slugify()` together with it, which imo is counter intuitive, not nice to read and will definitely break in the future again, as at some point someone will forget about it
- it doesn't solve the underlying issue with `domainWord()`

As I said [here](https://github.com/faker-js/faker/issues/1105#issuecomment-1186630351):

> From what I understand, not all TLDs are even accepting internationalized domain names ([wiki](https://en.wikipedia.org/wiki/Internationalized_domain_name#Top-level_domains_accepting_IDN_registration)), so I think it is out of scope for faker to determine which are and keep track of that. Imo, domain words should just not include non-ASCII chars to keep it simple.

Additionally, internationalized domain names are converted to ASCII compatible encoding on the client when using them. They simply don't exist when building servers.

Btw, I noticed I forgot to re-add `jalapeño` after refactoring the regex of `domainWord()`. Will do that once our discussion here moves forward :)

Shinigami92:
@schw4rzlicht I highly assume he meant using it inside faker in this PR, not in your code to wrap faker calls.

matthewmayer:
right could you change domainWord to just be something like

```
domainWord(): string {
    return this.faker.helpers
      .slugify(`${this.faker.word.adjective()}-${this.faker.word.noun()}`)
      .toLowerCase();
  }
```

schw4rzlicht:
> @schw4rzlicht I highly assume he meant using it inside faker in this PR, not in your code to wrap faker calls.

Ah alright, misunderstood that, sorry 🤦

Thanks @matthewmayer, I will update the PR ~today or tomorrow~ :)

EDIT: Wasn't able to do it yet, too much work on other projects. Will do asap though.

src/modules/internet/index.ts
```diff
@@ -358,8 +358,8 @@ export class InternetModule {
    * @since 2.0.1
    */
   domainWord(): string {
-    return `${this.faker.word.adjective()}-${this.faker.word.noun()}`
-      .replace(/([\\~#&*{}/:<>?|\"'])/gi, '')
+    return this.faker.helpers
+      .slugify(`${this.faker.word.adjective()}-${this.faker.word.noun()}`)
       .replace(/\s/g, '-')
       .replace(/-{2,}/g, '-')
```
matthewmayer: I don't think you need these two lines any more as slugify already converts spaces to - for you

matthewmayer (changes requested):

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #1521, opened 2022-11-04
refactor(helpers)!: rewrite shuffle

Shinigami92:
https://github.com/faker-js/faker/pull/714#discussion_r1014247945


src/modules/helpers/index.ts
```diff
@@ -240,29 +240,61 @@ export class HelpersModule {
...
-   * Uses the modern version of the Fisher–Yates algorithm.
+   * @template T The type of the entries to shuffle.
+   * @param list The array to shuffle.
+   * @param options The options to use when shuffling.
+   * @param options.inplace Whether to shuffle the array in place or return a new array. Defaults to `false`.
+   *
+   * @example
+   * faker.helpers.shuffle(['a', 'b', 'c'], { inplace: true }) // [ 'b', 'c', 'a' ]
+   *
+   * @since 2.0.1
```
ST-DDT: Do we have to update the `@since` here?

src/modules/helpers/index.ts
```diff
@@ -240,29 +240,61 @@ export class HelpersModule {
...
+   * faker.helpers.shuffle(['a', 'b', 'c'], { inplace: false }) // [ 'b', 'c', 'a' ]
    *
    * @since 2.0.1
    */
-  shuffle<T>(o?: T[]): T[] {
-    if (o == null || o.length === 0) {
-      return o || [];
+  shuffle<T>(list: T[], options?: { inplace?: boolean }): T[];
+  shuffle<T>(list: T[], options?: { inplace?: boolean }): T[] {
+    const { inplace = false } = options || {};
```
```suggestion
  shuffle<T>(list: T[], options?: { inplace?: boolean }): T[];
  shuffle<T>(list: T[], options: { inplace?: boolean } = {}): T[] {
    const { inplace = false } = options;
```

I'm not sure about the first line.

ST-DDT (dismissed):

ejcheng (dismissed):

src/modules/helpers/index.ts
```suggestion
   * @since 8.0.0
```

ejcheng (approved):

src/modules/helpers/index.ts
```diff
@@ -240,29 +240,61 @@ export class HelpersModule {
...
+   * @param options The options to use when shuffling.
+   * @param options.inplace Whether to shuffle the array in place or return a new array. Defaults to `false`.
    *
    * @example
-   * faker.helpers.shuffle() // []
    * faker.helpers.shuffle(['a', 'b', 'c']) // [ 'b', 'c', 'a' ]
+   * faker.helpers.shuffle(['a', 'b', 'c'], { inplace: false }) // [ 'b', 'c', 'a' ]
    *
    * @since 2.0.1
    */
```
xDivisionByZerox, line 267: I don't see the advantage in duplicating all that JSDocs just to show some "more specific" examples. Just put all 3 examples in the same docs IMO.

Shinigami92: I understand the maintenance burden, but for the end user in VSCode it has a value and therefore IMO it's worth it.
Beside that, the signatures still need to be separated, because otherwise we could not support readonly arrays. And when a user then hovers over one of the signatures that uses a readonly array, the user wont get any JSDoc if we would leave it out.

xDivisionByZerox: > I understand the maintenance burden, but for the end user in VSCode it has a value and therefore IMO it's worth it.

First, as I said, there is no added value for the end user here. Why does it matter that the JsDocs tell you which boolean value `inplace` has (which you did input yourself). Furthermore, there is no visible change in the examples no matter if `inplace` is `true`, `false` or `undefined`.

> Beside that, the signatures still need to be separated, because otherwise we could not support readonly arrays

Ok, I see that. Buy why do we need 3 signatures then?

ST-DDT: > Ok, I see that. Buy why do we need 3 signatures then?

The third one is for the docs-generation. Since we only show one there, it must be the union of all supported ones.

xDivisionByZerox: I see. In this case I resolve this.

xDivisionByZerox (approved):

ST-DDT (approved):

PR #1517, opened 2022-11-04
docs: add upgrading guide for v8

matthewmayer:
fix #1518

PR #1522, opened 2022-11-05
infra: update website redirects (v7)

ST-DDT:
Based on an idea proposed by @matthewmayer in https://github.com/faker-js/faker/pull/1517#discussion_r1014600950.

This PR ensures that 
* our chat and demo page can be accessed in a symmetric way
* we can already use the v7+v8 domains as redirects to the main and next page respectively

This PR does point to main website.

Next Branch PR: https://github.com/faker-js/faker/pull/1523

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,6 +7,20 @@
   publish = "docs/.vitepress/dist"
   command = "npx pnpm i --store=node_modules/.pnpm-store && npm run docs:build:ci"
 
+# Alias for the main page
+[[redirects]]
+  from = "https://v7.fakerjs.dev"
+  to = "https://fakerjs.dev"
+  status = 302
+  force = true
+
+# Alias for the next page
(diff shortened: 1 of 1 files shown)


ejcheng (approved):

Shinigami92:
What happens if a sub-domain collides with one of the branch-sub-domains configured by Netlify?
Which would take precedence?

Shinigami92:
Looks like I can merge this right now without giving a review by another maintainer 👀 
I think we need to configure these branches also as protected with more restrictions

ST-DDT:
> What happens if a sub-domain collides with one of the branch-sub-domains configured by Netlify? Which would take precedence?

The configuration from the prod branch. It is set to force so it will even take effect if there is a page to display.

prisis (approved):

xDivisionByZerox (approved):

PR #1523, opened 2022-11-05
infra: update website redirects (next)

ST-DDT:
Based on an idea proposed by @matthewmayer in https://github.com/faker-js/faker/pull/1517#discussion_r1014600950.

This PR ensures that (once v8 is released)
* our chat and demo page can be accessed in a symmetric way
* we can already use the v8+v9 domains as redirects to the main and next page respectively

This PR doesn't point to main website so I assume, it won't affect our current setup, but I will run some additional tests in my private fork later this weekend.

Main/v7 PR: https://github.com/faker-js/faker/pull/1522


ejcheng (approved):

netlify.toml
```diff
@@ -6,14 +6,40 @@
...
+[[redirects]]
+  from = "https://v8.fakerjs.dev"
+  to = "https://fakerjs.dev"
+  status = 302
+  force = true
+
+# Alias for the next page
+[[redirects]]
+  from = "https://v9.fakerjs.dev"
+  to = "https://next.fakerjs.dev"
```
Shinigami92, line 19: I would like to talk about this in a team meeting
My concerns about that are that it is not automated, we need to always handle it manually on a version update and in this example v8 would point to an unreleased doc right now after we merge this.

ST-DDT: IMO this is not worth automating.
There are and will always be things that you have to manually for major version upgrades.

- Clear changelog
- write/update migration guide
- updating certain configs related to netlify (which branch is prod, which branches to deploy, including domain redirects)
- ...

Should we document them somewhere? Yes. But automate them? No.

The next branch is not the prod branch so it wont get active. (I cannot test that because I dont have a test domain that I could use for that)

xDivisionByZerox (approved):

prisis (approved):

PR #1534, opened 2022-11-07
refactor(date)!: unify signatures

Shinigami92:
related to https://github.com/faker-js/faker/issues/1349#issuecomment-1303820190

PR #1536, opened 2022-11-07
fix(date): don't allow parameters <= 0

Shinigami92:
While working on #1534 I found this bug and extracted it to this PR.

Before this fix, `values === 0` would be treated as _falsy_ value and fallback to `1`.
We want to prevent this behavior directly and now throw an error when the user tries to use values below or equals to 0.


src/modules/date/index.ts
```diff
@@ -47,10 +47,14 @@ export class DateModule {
    * @since 2.0.1
    */
   past(years?: number, refDate?: string | Date | number): Date {
+    if (years <= 0) {
+      throw new FakerError('Years should be greater than 0.');
```
Shinigami92: @import-brain should this be `should` or `must`?

ejcheng: Hmm, I would say `must` but either would work here.

ST-DDT (dismissed):
As an alternative we could treat the `<=0` as zero returning effectively `now`/`refDate` (minus/plus 1 second to be in the past/future).

test/date.spec.ts
```diff
@@ -158,11 +158,11 @@ describe('date', () => {
...
-        it('should return a future date when years 0', () => {
+        it('should throw an error when years = 0', () => {
           const refDate = new Date();
-          const date = faker.date.future(0, refDate.toISOString());
-
-          expect(date).greaterThan(refDate); // date should be after the date given
+          expect(() => faker.date.future(0, refDate.toISOString())).toThrow(
+            new FakerError('Years must be greater than 0.')
+          );
         });
```
ST-DDT, line 166: These tests are missing for recent and soon.

Shinigami92: Good catch, thx
Will fix this late evening or tomorrow

Shinigami92:
> As an alternative we could treat the `<=0` as zero returning effectively `now`/`refDate` (minus/plus 1 second to be in the past/future).

IMO this is not directly part of this issue. I would like to have this fixed to proceed with #1534.
But yeah, we can talk about this in next meeting in general, but would like to prio it down in the agenda so we have more focus on other more important things already on the agenda.

ST-DDT (approved):

ejcheng (approved):

xDivisionByZerox (approved):

PR #1537, opened 2022-11-07
test: use runner on different timezone

Shinigami92:
https://github.com/vitest-dev/vitest/issues/1575


ST-DDT:
IMO there is something else going on here.
Which timezone do you use for your system?

Is there anyone else that has trouble generating/updating the test snapshots files?

ST-DDT (changes requested):
Does not work for me:

````diff
- Date: Tue Dec 31 2019 13:03:15 GMT+0000 (Coordinated Universal Time)
+ Date: Tue Dec 31 2019 13:03:15 GMT+0000 (Koordinierte Weltzeit)
````

**Comment by ST-DDT** [comment]
FFR: This PR broke the tests: https://github.com/faker-js/faker/pull/1512, but the bug existed even before that, we just don't had a test that touched it.

**Comment by Shinigami92** [comment]
Oh yeah you are right, previously we didn't had snapshot tests for that
Okay, lets rewrite the implementation in another PR

**Comment by ST-DDT** [comment]
**Team decision**

We will add a runner instead, that runs in non-UTC and non-EN to highlight these issues in the future.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -57,6 +57,49 @@ jobs:
...
+        os: [ubuntu-latest]
+        node_version: [18]
+      fail-fast: false
+
+    name: 'Timezone Test: node-${{ matrix.node_version }}, ${{ matrix.os }}'
+    steps:
+      - name: Set Timezone
+        uses: szenius/set-timezone@v1.0
+        with:
+          timezoneLinux: 'Asia/Singapore'
```
ST-DDT: Shouldn't there easier ways to set the TZ?

Maybe set the locale to non US as well?

https://github.com/actions/runner-images/issues/762#issuecomment-618507288

ST-DDT (approved):

ejcheng (approved):

PR #1539, opened 2022-11-07
fix(git): adjust commitEntry to match `git log` output

ST-DDT:
Supersedes #1537
Fixes the test snapshots tests for `faker.git.commitEntry` introduced by #1512

- #1512
- #1537

-------

Fixes the current issues with our test snapshots for `faker.git.commitEntry()` by changing the output to match the output of `git log`. That command uses the `en` locale on all platforms (we were able to test on).

To simplify the implementation and to retain the code readability the `commitDate` generation was moved to a separate method.

grafik


src/modules/git/index.ts
```diff
@@ -1,5 +1,17 @@
 import type { Faker } from '../..';
 
+const gitDateFormatBase = new Intl.DateTimeFormat('en-US', {
```
Shinigami92:
```suggestion
const GIT_DATE_FORMAT_BASE = new Intl.DateTimeFormat('en-US', {
```

src/modules/git/index.ts
```diff
@@ -95,6 +114,45 @@ export class GitModule {
...
+   * @example
+   * faker.git.commitDate() // 'Mon Nov 7 14:40:58 2022 +0600'
+   * faker.git.commitDate({ refDate: '2020-01-01' }) // 'Tue Dec 31 05:40:59 2019 -0400'
+   *
+   * @since 8.0.0
+   */
+  commitDate(options: { refDate?: string | Date | number } = {}): string {
+    const { refDate } = options;
+
+    const dateParts = gitDateFormatBase
```
Shinigami92:
```suggestion
    const dateParts = GIT_DATE_FORMAT_BASE.format(
      this.faker.date.recent(1, refDate)
    )
```

src/modules/git/index.ts
```diff
@@ -95,6 +114,45 @@ export class GitModule {
...
+        '+0000',
+        `+${this.faker.datatype
+          .number({ min: 1, max: 11 })
+          .toString()
+          .padStart(2, '0')}00`,
+        `-${this.faker.datatype
+          .number({ min: 1, max: 11 })
+          .toString()
+          .padStart(2, '0')}00`,
+      ])
```
Shinigami92:
```suggestion
      `${this.faker.datatype
        .number({ min: -12, max: 11 })
        .toLocaleString('en', {
          minimumIntegerDigits: 2,
          maximumFractionDigits: 0,
          useGrouping: false,
          signDisplay: 'always',
        })}00`
```

Not sure if `min: -12` is what it is intended

Shinigami92 (commented):
After these changes, we need to update snapshots

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):

PR #1545, opened 2022-11-09
feat(helpers): introduce `multiple` method

ST-DDT:
Adds a new method to generate multiple values from a method.

In our code we have a few of places that generate multiple values from a single method:

- https://github.com/faker-js/faker/blob/8bf716225e9c5b3ce5c968511886b9141d5c918d/src/modules/system/index.ts#L80-L81
- https://github.com/faker-js/faker/blob/02f3967a9960a93a8f6137aebcaaae5754261135/src/modules/string/index.ts#L163-L165
- https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/modules/color/index.ts#L323-L325
- https://github.com/faker-js/faker/blob/838f8369196baabb28529470c27125756e6f671e/src/modules/datatype/index.ts#L301-L303
- https://github.com/faker-js/faker/blob/838f8369196baabb28529470c27125756e6f671e/src/modules/finance/index.ts#L345
- And probably the whole lorem module
- And others that use a loop right now

**Example**

````ts
faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
faker.helpers.multiple(faker.person.firstName, { count: 3, unique: true }) // [ 'Santos', 'Nedra', 'Lavinia' ]
````

**Planned features**
- Dynamic result count / range (requires #1486)
- Generate unique elements per array (requires another PR)

### CODE REVIEW

**Comment by ST-DDT** [comment]
Blocked till #1486 is merged.

- #1486

File: src/modules/internet/index.ts [hunk 1]
```diff
@@ -318,16 +318,9 @@ export class InternetModule {
...
-
-    const result: string[] = [];
-    for (let i = 0; i < 4; i++) {
-      result[i] = randNum();
-    }
-
-    return result.join('.');
+    return this.faker.helpers
+      .multiple(() => this.faker.datatype.number(255), { count: 4 })
+      .join('.');
```
ST-DDT: Should I use `Array.from(...)` here instead?

src/modules/internet/index.ts
```diff
@@ -339,36 +332,17 @@ export class InternetModule {
...
+      .multiple(
+        () =>
+          this.faker.string.hexadecimal({
+            length: 4,
+            casing: 'lower',
+            prefix: '',
+          }),
+        { count: 8 }
+      )
+      .join(':');
```
ST-DDT: Same here

src/modules/internet/index.ts
xDivisionByZerox: I'm a bit lost. Why would we want to use `Array.from` when you just implemented the helper method? Like all over in `color/index.ts`?
Why would we want to use one one time and the other another time?

ST-DDT: My thoughts here are that we use Array.from when using static counts (to reduce inter-code-dependencies) and helpers.multiple when using dynamic or user given (count | range)s.

src/modules/helpers/index.ts
```diff
@@ -690,4 +690,34 @@ export class HelpersModule {
...
+   * Generates an array containing values returned by the given method.
+   *
+   * @param method The method used to generate the values.
+   * @param options The optional options object.
+   * @param options.count The number or range of elements to generate. Defaults to `3`.
+   *
+   * @example
+   * faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
+   * faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple(faker.person.firstName, { count: 3, unique: true }) // [ 'Santos', 'Nedra', 'Lavinia' ]
```
Shinigami92: This is not in yet :thinking:

ST-DDT: Fixed

Shinigami92 (dismissed):

ejcheng (dismissed):

ejcheng (dismissed):

test/helpers.spec.ts
```diff
@@ -854,5 +862,33 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`)
         expect(store).toEqual({ 'with conflict: 0': 'with conflict: 0' });
       });
     });
+
+    describe('multiple()', () => {
+      it('should generate values from the function', () => {
```
xDivisionByZerox:
```suggestion
      it('should generate values from the function with a default length of 3', () => {
```

Just to make sure that the test title specifies exactly what it does.

xDivisionByZerox (dismissed):
I will already approve as I don't think the test name reword is thaaat important.

xDivisionByZerox (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #1551, opened 2022-11-12
refactor(string): rename params

### LINKED ISSUE #1465: Rename `bannedChars` and `bannedDigits` in string module
Extracted out of https://github.com/faker-js/faker/pull/1155#discussion_r922645362

The parameter names `bannedChars` and `bannedDigits` could be unified to e.g.:
- except
- without
- exclude

This needs to be done in v8.0 so we do not make breaking changes later on

ejcheng:
closes #1465


src/modules/random/index.ts
```diff
@@ -211,7 +211,11 @@ export class RandomModule {
     if (typeof options === 'number') {
       return this.faker.string.alpha(options);
     }
-    return this.faker.string.alpha({ ...options, length: options.count });
+    return this.faker.string.alpha({
+      ...options,
```
ST-DDT: IMO you can just put the casing here directly since it is the only property that is left.

src/modules/vehicle/index.ts
```diff
@@ -91,15 +91,15 @@ export class VehicleModule {
     return `${this.faker.string.alphanumeric({
       length: 10,
       casing: 'upper',
-      bannedChars,
+      exclude: bannedChars,
```
ST-DDT: Please also rename the variable here.

ST-DDT (changes requested):
Please remove the obsolete test snapshots.

Also we should adjust the jsdocs and error messages to use the new term.

test/__snapshots__/string.spec.ts.snap
```diff
@@ -24,6 +26,8 @@ exports[`string > 42 > alpha > with length parameter 5`] = `"bcYLR"`;
 
 exports[`string > 42 > alpha > with length, casing and bannedChars 1`] = `"fwzcvwj"`;
```
Shinigami92: Run `pnpm run test run -u` and these will go away

ST-DDT: I think we should add it to our uncommited changes CI.

Shinigami92 (dismissed):

src/modules/string/index.ts
```diff
@@ -98,17 +98,17 @@ export class StringModule {
   /**
    * Generating a string consisting of letters in the English alphabet.
    *
-   * @param options Either the number of characters or an options instance. Defaults to `{ length: 1, casing: 'mixed', bannedChars: [] }`.
+   * @param options Either the number of characters or an options instance. Defaults to `{ length: 1, casing: 'mixed', exclude: [] }`.
    * @param options.length The number of characters to generate. Defaults to `1`.
    * @param options.casing The casing of the characters. Defaults to `'mixed'`.
-   * @param options.bannedChars An array with characters to exclude. Defaults to `[]`.
+   * @param options.exclude An array with characters to exclude. Defaults to `[]`.
```
ST-DDT: This description is phrased slightly different than the others.
Could you please adjust it/the others so it is more uniform?

Other than this LGTM.

ST-DDT (dismissed):

ST-DDT (approved):

Shinigami92 (approved):

Issue #1437, opened 2022-10-13
The format of the email address output in Japanese is not correct.


shu-pf:
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md).
- [X] Read the [docs](https://fakerjs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/faker-js/faker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Faker issue and not related to a combination with another package.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/faker-js/faker/discussions) or join our [Discord Chat Server](https://chat.fakerjs.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.
- [X] I am willing to provide a PR.

## Clear and concise description of the problem

Japanese people do not use multibyte characters in their e-mail addresses because using multibyte characters in some e-mail addresses can cause erroneous transmissions.
Therefore, we generally convert the name to romaji first and use that for the e-mail address.

## Suggested solution

Current Behavior

```
import { faker } from '@faker-js/faker/locale/ja';

const email = faker.internet.email();
console.log(email); // 樹_林@hotmail.com
```

Expected behavior

```
import { faker } from '@faker-js/faker/locale/ja';

const email = faker.internet.email();
console.log(email); // itsuki_hayashi@hotmail.com
```

### Additional Context

It may also be possible to solve this problem by creating a function that converts Japanese names to romaji.

### Environment Info

```shell
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 133.45 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ~/.volta/tools/image/node/16.13.2/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 8.1.2 - ~/.volta/tools/image/node/16.13.2/bin/npm
  Browsers:
    Chrome: 106.0.5249.103
    Firefox: 96.0.1
    Safari: 15.3
  npmPackages:
    @faker-js/faker: ^7.5.0 => 7.5.0
```

[...]

Issue #1211, opened 2022-07-30
Add the concept of order for `firstName` and `lastName` in `name.fullName`


satoc0629:
### Clear and concise description of the problem

The sequence of Japanese names is lastName followed by firstName in the Japanese-speaking world.
ex: YAMADA TARO.
YAMADA is lastName.
TARO is firstName.
I would like to incorporate this into faker as well.

// This sentence relies on machine translation.

### Suggested solution

Add full_name_order to NameDefinitions.
The full_name_order is as follows.
```typescript
export type FullNameOrder = 'normal' | 'reverse';
```

Add the ordering process using this to fullName in src/modules/name/index.ts.
```typescript
    if (this.faker.definitions.name.full_name_order === 'reverse') {
      nameParts.push(lastName);
      nameParts.push(firstName);
    } else {
      nameParts.push(firstName);
      nameParts.push(lastName);
    }
```

### Alternative

I do not understand Faker or programming that deeply, so the best approach may be something else.

### Additional context

In addition to this, we would like to add a collection of first names for Japanese men and a collection of first names for Japanese women.

xDivisionByZerox:
The suggested solution you provided sounds reasonable to me. Other locales could gain value from this as well.
In my opinion, I would call the flag in the definition `default_full_name_order` and would suggest adding an option to `name.fullName()` that allows overriding of the default behavior for the current locale:
```ts
// en version, but that doesn't matter
import { faker } from '@faker-js/faker';

console.log(faker.definitions.name.full_name_order); // 'normal'

// call with no params resolves to the current value of the locale definition (in this case 'normal')
console.log(faker.name.fullName()); // Lexie Bernie

// explicitly set the order, but this doesn't do anything in this case as it's the same as the definitions
console.log(faker.name.fullName({ order: 'normal' })); // Lexie Bernie

// explicitly set the order and override the default value
console.log(faker.name.fullName({ order: 'reverse' })); // Bernie Lexie
```

import-brain:
> The suggested solution you provided sounds reasonable to me. Other locales could gain value from this as well.
> 
> In my opinion, we should call the property in the definition `default_full_name_order`. Additionally, I would suggest adding an option to `name.fullName()` that allows overriding of the default behavior for the current locale:
> 
> ```ts
> // en version, but that doesn't matter
> import { faker } from '@faker-js/faker';
> 
> console.log(faker.definitions.name.full_name_order); // 'normal'
> 
[...]

PR #1554, opened 2022-11-13
fix(internet): userName, email and slugify return only ascii

### LINKED ISSUE #1105: Weird email and username in Chinese locale package

matthewmayer:
tentative fix for #1105 and #1437

before this PR is applied:

```js
Object.keys(faker.locales).forEach(locale=>{faker.setLocale(locale); console.log(`${locale}: ${faker.internet.email()}`)})
```

```
af_ZA: Harvey_Ferreira60@gmail.com
ar: .@yahoo.com
az: Kellie_Hansen@yahoo.com
cz: Krytof9@atlas.cz
de: Lisann_Tsamonikian@yahoo.com
de_AT: Lenja2@gmail.com
de_CH: Marlies29@hotmail.com
el: .@gmail.com
en: Isobel40@yahoo.com
en_AU: Eliza_Edwards@yahoo.com
en_AU_ocker: Oliver46@gmail.com
en_BORK: Vita.Buckridge78@yahoo.com
en_CA: Fausto18@gmail.com
en_GB: Adrienne.Konopelski@yahoo.com
en_GH: person.female_first_name.Kusi@hotmail.com
en_IE: Oswaldo.Dietrich@hotmail.com
en_IN: Baalaaditya15@yahoo.co.in
en_NG: Titi.Christian94@yahoo.com
en_US: Erik83@hotmail.com
en_ZA: Amelia_Connelly33@yahoo.com
es: Esteban93@gmail.com
es_MX: Mayte.Ruiz@nearbpo.com
fa: 72@yahoo.com
fi: Oskari.Hmlinen@hotmail.com
fr: Flavie_Nguyen@hotmail.fr
fr_BE: Freda7@advalvas.be
fr_CA: Daija_Osinski@yahoo.ca
fr_CH: Arion13@hotmail.com
ge: _@posta.ge
he: 14@gmail.com
hr: David.Zdelar48@gmail.com
hu: Dina63@outlook.com
hy: .@gmail.com
id_ID: Paul_OKeefe@yahoo.co.id
it: Igor24@libero.it
ja: 太一.中村73@gmail.com
ko: 71@yahoo.co.kr
lv: Grover_Kshlerin@apollo.lv
mk: 44@hotmail.com
nb_NO: Herman_Strand@yahoo.com
ne: Raju26@gmail.com
nl: Nick.Janssen33@gmail.com
nl_BE: Amy44@gmail.com
pl: Gerald.Urbanowicz44@yahoo.com
pt_BR: Lvia98@yahoo.com
pt_PT: Edgar60@mail.pt
ro: Trenton46@hotmail.com
ru: Seamus.Carter@yahoo.com
sk: Bethany.Parisian@zoznam.sk
sv: Monica.Axelsson@gmail.com
tr: Brbars1@yahoo.com
uk: Hellen_Price34@ukr.net
ur: .@gmail.com
vi: VinhDiu.Mai@yahoo.com
zh_CN: 鑫鹏_宋@gmail.com
zh_TW: 樂駒76@hotmail.com
zu_ZA: Maphikelela.Mabhida@hotmail.com
```

after this PR is applied:

```
af_ZA: Margaret_Human@hotmail.com
ar: da8294716@yahoo.com
az: Tobin.Frami@yahoo.com
cz: Nepomuk_Fiedler70@atlas.cz
de: Mina_Markowski@gmail.com
de_AT: Lucy_Dietzsch@yahoo.com
de_CH: Richard.Huber@hotmail.com
dv: bg650967@yahoo.com
el: bm4452630@hotmail.com
en: Berenice.Swaniawski@gmail.com
en_AU: Jasmine_Greenholt49@yahoo.com
en_AU_ocker: Amelia_Nguyen7@yahoo.com
en_BORK: Blanca.Roob@gmail.com
en_CA: Marge_Goyette@hotmail.com
en_GB: Cierra45@hotmail.com
en_GH: person.female_first_name_Acheampong76@hotmail.com
en_IE: Gianni40@gmail.com
en_IN: Tejas14@gmail.com
en_NG: Fatima52@yahoo.com
en_US: Monserrat83@yahoo.com
en_ZA: Stephen_Hearne72@gmail.com
es: Luz_Collazo@gmail.com
es_MX: Ivanna.Llamas@hotmail.com
fa: bl1424060@yahoo.com
fi: Mika11@gmail.com
fr: Philippine28@yahoo.fr
fr_BE: Salvador98@netbel.be
fr_CA: Josue.Sauer@gmail.com
fr_CH: Eline.Sutermeister94@hotmail.com
ge: md9854879@gmail.com
he: gl1440764@hotmail.com
hr: Slavica29@gmail.com
hu: Lilina.Nmeth47@hotamil.com
hy: an7859723@yahoo.com
id_ID: Anita.Gottlieb@yahoo.co.id
it: Erminia_Laezza@email.it
ja: oq961736@gmail.com
ko: ix4305140@hanmail.net
lv: Amira_Renner@yahoo.com
mk: jn6356977@yahoo.com
nb_NO: Kasper_Pettersen91@yahoo.com
ne: Sanjay.Maharjan94@worldlink.com.np
nl: Anne_Klein@gmail.com
nl_BE: Arne_Goossens82@hotmail.com
pl: Mateusz_Knapik9@hotmail.com
pt_BR: Mariana36@hotmail.com
pt_PT: Beatriz55@hotmail.com
ro: Lorenza58@gmail.com
ru: Gilberto_Murphy@mail.ru
sk: Wyman_Bashirian83@azet.sk
sv: Cecilia92@gmail.com
tr: Bulgak_Erberk86@yahoo.com
uk: Krista_Powlowski@ex.ua
ur: zq6319489@gmail.com
vi: DimTrinh43@yahoo.com
zh_CN: fd7075767@yahoo.com
zh_TW: jn4619221@gmail.com
zu_ZA: Ayize.Mehloluhlaza62@yahoo.com
```


src/modules/internet/index.ts
```diff
@@ -85,6 +85,16 @@ export class InternetModule {
...
+    //In some locales e.g. ja or el, userName contains mostly non-Unicode characters
+    //https://github.com/faker-js/faker/issues/1105
+    //After slugify it becomes a string with only 0-9, _ and .
+    //In that case we generate a purely random local part instead
+    const invalidLocalPart: boolean = /^[0-9\._]*$/.test(localPart);
+    if (invalidLocalPart) {
+      localPart =
+        this.faker.string.alpha({ length: 2, casing: 'lower' }) +
+        this.faker.datatype.number({ min: 10000, max: 9999999 }).toString();
+    }
```
ST-DDT: IMO this should be part of slugify. Is there a way to do that input dependent? Maybe add a todo for that.

matthewmayer: In an ideal world we'd be able to transilterate from names in any script to ascii for example "王东" becomes "Wang Dong", "Владимир Ленин" becomes "Vladimir Lenin" and so on.

But there's not really a good generic way to do that without adding lots of dependencies for different locales (pinyin for Chinese, etc).

So i think either we just fallback to something completely random like this, or (as I suggested at https://github.com/faker-js/faker/issues/1105#issuecomment-1305038744) allow non-ASCII-ish locales to expose a seperate static list of ASCII transliterated firstnames and surnames.

src/modules/internet/index.ts
```diff
@@ -85,6 +85,16 @@ export class InternetModule {
       this.userName(firstName, lastName)
     );
 
+    //In some locales e.g. ja or el, userName contains mostly non-Unicode characters
+    //https://github.com/faker-js/faker/issues/1105
+    //After slugify it becomes a string with only 0-9, _ and .
+    //In that case we generate a purely random local part instead
```
Shinigami92:
```suggestion
    // In some locales e.g. ja or el, userName contains mostly non-Unicode characters
    // https://github.com/faker-js/faker/issues/1105
    // After slugify it becomes a string with only 0-9, _ and .
    // In that case we generate a purely random local part instead
```

test/internet.spec.ts
```diff
@@ -130,19 +130,6 @@ describe('internet', () => {
           expect(faker.definitions.internet.free_email).toContain(suffix);
         });
 
-        it('should return an email with japanese characters', () => {
```
Shinigami92, line 133: There are only removed tests but no new added ones 🤔 
Feels destructive

By the way, as far as I know there can really exists email addresses with non english ascii letters. I don't know how far this goes and I assume mail providers explicitly need to support it, but it's possible.

matthewmayer: if you agree this is the right direction I can add new tests :D

its theoretically possible to have non-ascii email addresses but its exceedingly rare in practice as email clients and servers tend to mess them up. 

it's anecdotal but i lived for 10 years in China and never encountered an email address with Chinese chars.

I think faker should aim to give realistic data where possible, rather than something theoretically possible but unrealistic.

Shinigami92: yes and no
It's also good to have some edge cases, so if you use faker you can discover those edge cases where you as a software writer hasn't thought about

That's my only concern, if other maintainers from Faker agree with you, I'm also okay with that. That was just my thought.

ST-DDT: Let's discuss this in the next team-meeting.

src/modules/internet/index.ts
ST-DDT: -> Team-Meeting

test/internet.spec.ts
matthewmayer: i added a test that in all locales, you get a valid email address. Note this fails for `mk` locale due to #1556

src/modules/internet/index.ts
matthewmayer: note in https://github.com/faker-js/faker/issues/1437#issuecomment-1277513007 @shu-pf also suggested a simular idea of using an ASCII alternative name set for Japanese for some contexts.

matthewmayer: My gut feeling is that the ASCII alternative name set is probably the best long term solution, BUT given that there's a lot of other issues with locales and names (like #547  and #1211 ) I think this current PR might be a good temporary solution for v8 which won't interfere with solving those more thorny issues with names and localization.

matthewmayer:
I added a further refinement to slugify which strips simple diacritics. This helps in languages like French or Vietnamese, where you have names like `Adélaïde` or `Ngân Hà`. Previously those accented letters would get removed completed, leading to email addresses like:

`adlade23@gmail.com`
`ngn.h23@gmail.com`

now instead you'd get
`adelaide23@gmail.com`
`ngan.ha23@gmail.com`

which seems nicer. I also added some additional tests for slugify.

Shinigami92:
**Team Decision:**

- We wont allow adding a dependency to faker to convert the given parameters to related outputs
  This can be done by the user in front of the method and then just pass the converted args to the method
- We will convert non-ascii chars to ascii chars, in a way that the input is related to the outcome, but not in a literal linguistic way
  How that is done is an implementation detail and in the hand of the PR author
- The method `internet.userName` should only return ascii characters
- A new method `internet.displayName` should have (kinda) the same implementation as `internet.userName` has right now, but we can improve e.g. the options/parameter as we are already making breaking changes in v8.0

matthewmayer:
"We will convert non-ascii chars to ascii chars, in a way that the input is related to the outcome, but not in a literal linguistic way. How that is done is an implementation detail and in the hand of the PR author"

So I tried a new approach in the latest commit, basically reimplementing the userName function with a few strategies:

Ascii names work as before

> faker.internet.userName("Matt","Mayer")
'Matt_Mayer22'

Simple accents are stripped using Unicode NFD

> faker.internet.userName("Hélene","Müller")
'Helene_Muller11'

A simple mapping can be use to transliterate alphabetical languages like Cyrillic (could also add say Greek and Thai).

> faker.internet.userName("Фёдор","Достоевский")
'Fedor.Dostoevskii50'

As a final fallback, just concatenate the hex unicode char codes of each character.

> faker.internet.userName("大羽","陳")
'59277fbd33'

This works tolerably well but leads to some very long usernames in the final fallback scenario

```
af_ZA: Theuns.Visser7@gmail.com
ar: 64662864a644_62764462e64464a64164a@hotmail.com
az: Kolby51@yahoo.com
cz: Pravdomil99@atlas.cz
de: Malte.Scherz@yahoo.com
de_AT: Defne.Hentschke60@yahoo.com
de_CH: Roland_Walter@yahoo.com
dv: 7907a77837a7.7847a67a07b07997a77877aa83@yahoo.com
el: 3a33c93c43b73c13b93bf3c2.3943b13b33ba3bb3b73c2@hotmail.com
en: Darryl.Dare22@hotmail.com
en_AU: Max.Hill3@yahoo.com
en_AU_ocker: Charlotte_Brown@yahoo.com
en_BORK: Asa.Abshire@hotmail.com
en_CA: Rosanna_Lesch43@hotmail.com
en_GB: Freida_Schaden@gmail.com
en_GH: {{person.male_first_name}}_Yirenkyi78@hotmail.com
en_IE: Margaretta94@gmail.com
en_IN: Bankim99@gmail.com
en_NG: Gbeminiyi.Atanda@yahoo.com
en_US: Alan65@hotmail.com
en_ZA: Jamie.Boucher@yahoo.com
es: Rosalia.Casarez39@hotmail.com
es_MX: Sergio_Barragan@corpfolder.com
fa: 6866276446276a973@yahoo.com
fi: Tarja23@hotmail.com
fr: Savin31@gmail.com
fr_BE: Samuel_Ratke@skynet.be
fr_CA: Elias1@hotmail.com
fr_CH: Iliana_Chappuis@outlook.com
ge: 10dc10dd10d310d010e0_10d110d010e010d310d010d510d410da10d810eb10d4@posta.ge
he: 5d25d55df.5d25d15d95e97@yahoo.com
hr: Gradimir_Uzanicki@gmail.com
hu: Benjamin.Orban@outlook.com
hy: 53f561580565576_53f56158056157a56557f57556157644@gmail.com
id_ID: Lindsay.Cummerata28@yahoo.co.id
it: Benigno71@yahoo.com
ja: 7d508863_677e672c31@yahoo.com
ko: 110b116f11ab1112116711bc.1103116111bc@gmail.com
lv: Hollis_Bogan52@mail.lv
mk: Mirche84@gmail.com
nb_NO: Synne42@hotmail.com
ne: Bibek_Limbu@worldlink.com.np
nl: Bas87@gmail.com
nl_BE: Oona24@gmail.com
pl: Apollo85@hotmail.com
pt_BR: Tertuliano_Moreira@hotmail.com
pt_PT: Pilar51@sapo.pt
ro: Gust19@hotmail.com
ru: Brooklyn58@hotmail.com
sk: Jovany58@gmail.com
sv: Emilia.Lind@yahoo.com
tr: Atalan49@gmail.com
uk: Anne.White39@gmail.com
ur: 64563364862f78@hotmail.com
vi: AnhVu.110ang@yahoo.com
zh_CN: 806a506580@hotmail.com
zh_TW: 70ab660e_4faf@yahoo.com
zu_ZA: Maphikelela61@gmail.com
```

ST-DDT (commented):
Impl wise this looks good to me.
Could you add an example to the jsdocs that uses non ascii input?
Also I think it might be better if you move the char mapping to separate file and import it. Not sure whether it should be ts or json. But that is optional for now IMO.

We could do a substring if it gets too long but that's a different issue.

Shinigami92:
We also discusses that we want `displayName` with (for now) the old behavior.
Can we please do this in this PR as well? Otherwise I'm afraid of missing it out.
If you explicitly not want to do this in this PR, please leave a reason and also open a new issue so we can track it and not miss it before v8.0 release.

Shinigami92 (commented):
I think there are missing tests for displayName

src/modules/internet/index.ts
```diff
@@ -122,14 +123,18 @@ export class InternetModule {
   }
 
   /**
-   * Generates a username using the given person's name as base. This will always return a plain ASCII string.
+   * Generates a username using the given person's name as base. This will always return a plain ASCII string. Some basic stripping of accents and transliteration of characters will be done.
    *
    * @param firstName The optional first name to use. If not specified, a random one will be chosen.
-   * @param lastName The optional last name to use. If not specified, a random one will be chosen.
+   * @param lastName The optional last name to use - though the last name is not always used. If not specified, a random one will be chosen.
```
ST-DDT: I wouldn't limit this to the last name. IMO we could have a more general note, that the given name parts might be used either fully or partially or even not at all to generate the username.

Maybe also add an `@see` targettign the `username` method.

src/modules/internet/index.ts
```diff
@@ -122,14 +123,18 @@ export class InternetModule {
...
    *
    * @example
    * faker.internet.userName() // 'Nettie_Zboncak40'
    * faker.internet.userName('Jeanne', 'Doe') // 'Jeanne98'
+   * faker.internet.userName('John', 'Doe') // 'John.Doe'
+   * faker.internet.userName('Hélene', 'Müller') // 'Helene_Muller11'
+   * faker.internet.userName('Фёдор', 'Достоевский') // 'Fedor.Dostoevskii50'
+   * faker.internet.userName("大羽","陳") //hlzp8d.tpv45
```
ST-DDT: All the other examples use single quotes:

```suggestion
   * faker.internet.userName('大羽', '陳') // 'hlzp8d.tpv45'
```

src/modules/internet/index.ts
```diff
@@ -251,6 +187,46 @@ export class InternetModule {
...
+   * @param firstName The optional first name to use. If not specified, a random one will be chosen.
+   * @param lastName The optional last name to use. If not specified, a random one will be chosen.
+   *
+   * @example
+   * faker.internet.displayName() // 'Nettie_Zboncak40'
+   * faker.internet.displayName('Jeanne', 'Doe') // 'Jeanne98'
+   * faker.internet.displayName('John', 'Doe') // 'John.Doe'
+   * faker.internet.displayName('Hélene', 'Müller') // 'Hélene_Müller11'
+   * faker.internet.displayName('Фёдор', 'Достоевский') // 'Фёдор.Достоевский50'
+   * faker.internet.displayName("大羽","陳") //大羽.陳
```
```suggestion
   * faker.internet.displayName('大羽', '陳') // '大羽.陳'
```

src/modules/internet/index.ts
```diff
@@ -155,6 +157,71 @@ export class InternetModule {
         ])}${lastName}${this.faker.number.int(99)}`;
         break;
     }
+
```
ST-DDT, line 162: Instead of creating the base from scratch, should we use `displayName()` as the base?

test/internet.spec.ts
```diff
@@ -41,6 +41,10 @@ describe('internet', () => {
       t.it('noArgs').it('with names', 'Jane', 'Doe');
     });
 
+    t.describe('displayName', (t) => {
+      t.it('noArgs').it('with names', 'Jane', 'Doe');
```
ST-DDT: Please also use a non ascii example here. (Same for username and email)

It is also possible to join them, since I think they will always take the same parameters:

````ts
t.describeEach('displayName', 'username', (t) => ...);
````

File: test/internet.spec.ts [hunk 9]
```diff
@@ -262,6 +253,50 @@ describe('internet', () => {
...
+          const username = faker.internet.userName('Амос', 'Васильев');
+          expect(username).includes('Amos');
+        });
+
+        it('should provide a fallback for Chinese etc', () => {
+          const username = faker.internet.userName('大羽', '陳');
+          expect(username).includes('hlzp8d');
+        });
+      });
+      describe('displayName()', () => {
```
ST-DDT:
```suggestion
      });

      describe('displayName()', () => {
```

test/helpers.spec.ts
```diff
@@ -180,9 +180,21 @@ describe('helpers', () => {
...
+        it('removes non-word characters from strings except . and -', () => {
+          expect(faker.helpers.slugify('foo bar')).toBe('foo-bar');
+          expect(faker.helpers.slugify('Faker is cool')).toBe('Faker-is-cool');
+          expect(faker.helpers.slugify('super*star')).toBe('superstar');
           expect(faker.helpers.slugify("d'angelo.net")).toBe('dangelo.net');
+          expect(faker.helpers.slugify('hello你好')).toBe('hello');
+        });
+      });
+
+      describe('slugify()', () => {
```
ST-DDT: I think this describe block can be merged with the one above.

ST-DDT (changes requested):

src/modules/internet/index.ts
matthewmayer: personally i would keep them seperate, it makes the logic easier to read and they might diverge in future.

ST-DDT (dismissed):

src/modules/internet/char-mappings.ts
```diff
@@ -0,0 +1,285 @@
+//selectively sourced from https://github.com/sindresorhus/transliterate/blob/main/replacements.js (under MIT license)
```
ejcheng:
```suggestion
// selectively sourced from https://github.com/sindresorhus/transliterate/blob/main/replacements.js (under MIT license)
```
Please add a space after the `//` on every comment. Otherwise, this LGTM

ejcheng (dismissed):

ST-DDT (dismissed):

src/modules/internet/char-mappings.ts
```diff
@@ -0,0 +1,285 @@
+// selectively sourced from https://github.com/sindresorhus/transliterate/blob/main/replacements.js (under MIT license)
```
Shinigami92: Maybe use a permalink for this 🤔

src/modules/internet/index.ts
Shinigami92: I'm with @matthewmayer 
Sure we have tests, but IMO `displayName()` solves another use-case and can diverge in future

src/modules/internet/index.ts
```diff
@@ -155,6 +159,73 @@ export class InternetModule {
...
+    //There may still be non-ascii characters in the result.
+    //First remove simple accents etc
+    result = result
+      .normalize('NFKD') //for example è decomposes to as e +  ̀
+      .replace(/[\u0300-\u036f]/g, ''); // removes combining marks
+
+    result = result
+      .split('')
+      .map(function (char) {
```
Shinigami92: Please use inline lambdas, not `function (char) {`

```suggestion
      .map((char) => {
```

Not sure if you need to reformat after apply suggestion

src/modules/internet/index.ts
```diff
@@ -155,6 +159,73 @@ export class InternetModule {
...
+        //if we have a mapping for this character, (for Cyrillic, Greek etc) use it
```
Shinigami92:
```suggestion
        // If we have a mapping for this character, (for Cyrillic, Greek etc) use it
```

src/modules/internet/char-mappings.ts
Shinigami92: As I already suggest some things, please start such comments with capital sentences

`// Selecti...`

src/modules/internet/index.ts
```diff
@@ -155,6 +159,73 @@ export class InternetModule {
...
+        if (char.charCodeAt(0) < 0x80) {
+          //keep ascii characters
+          return char;
+        }
+        //final fallback return the Unicode char code value for Chinese, Japanese, Korean etc, base-36 encoded
+        return char.charCodeAt(0).toString(36);
+      })
+      .join('');
+
+    //remove spaces and '
```
Shinigami92:
```suggestion
    // Remove spaces and `'`
```

or

```suggestion
    // Remove spaces and ' (single-quotes)
```

Shinigami92: Or even just don't describe the code, because we all can read code ^^
Normally comments should explain the why, not what it do

Shinigami92 (changes requested):

ST-DDT (approved):

Shinigami92 (approved):

PR #1564, opened 2022-11-15
fix: strongly type ethereum address with a `0x${string}` type

RobbyUitbeijerse:
Hello!

In our application we would like to strongly type token addresses for ethereum, and instead of type `string` we would like to provide a type that guarantees us with the format 0x${string} throughout our applications

We currently typecast the value that faker.finance.ethereumAddress() provides us to it, that works perfectly fine but it seems like a nice addition to have a stronger type for this coming from faker.

Lmk if you would like to see any changes.

### DIFF
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -353,11 +353,12 @@ export class FinanceModule {
    *
    * @since 5.0.0
    */
-  ethereumAddress(): string {
+  ethereumAddress(): `0x${string}` {
     const address = this.faker.string.hexadecimal({
       length: 40,
       casing: 'lower',
     });
+
     return address;
   }
 
--- a/src/modules/string/index.ts
+++ b/src/modules/string/index.ts
@@ -78,6 +78,7 @@ export type NumericChar =
 
 export type AlphaChar = LowerAlphaChar | UpperAlphaChar;
 export type AlphaNumericChar = AlphaChar | NumericChar;
+export type StringLiteral<T> = T extends `${string & T}` ? T : string;
 
 const SAMPLE_MAX_LENGTH = Math.pow(2, 20);
 
@@ -257,14 +258,18 @@ export class StringModule {
    *
    * @since 8.0.0
    */
-  hexadecimal(
+  hexadecimal<T = '0x'>(
     options: {
       length?: number;
       casing?: Casing;
-      prefix?: string;
+      prefix?: StringLiteral<T>;
     } = {}
-  ): string {
-    const { length = 1, casing = 'mixed', prefix = '0x' } = options;
+  ): `${StringLiteral<T>}${string}` {
+    const {
+      length = 1,
+      casing = 'mixed',
+      prefix = '0x' as StringLiteral<T>,
+    } = options;
 
     let wholeString = '';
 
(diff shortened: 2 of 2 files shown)


src/modules/string/index.ts
```diff
@@ -295,11 +300,8 @@ export class StringModule {
       ]);
     }
 
-    if (casing === 'upper') {
-      wholeString = wholeString.toUpperCase();
-    } else if (casing === 'lower') {
-      wholeString = wholeString.toLowerCase();
-    }
+    wholeString =
+      wholeString[casing === 'upper' ? 'toUpperCase' : 'toLowerCase']();
```
Shinigami92: This doesn't look correct to me ⚠️ 
You missed the `mixed` casing!

Shinigami92 (changes requested):

src/modules/string/index.ts
RobbyUitbeijerse: @Shinigami92 good catch, didn't notice that casing type. I have simply reverted the change for converting the string

src/modules/finance/index.ts
```diff
@@ -353,11 +353,12 @@ export class FinanceModule {
    *
    * @since 5.0.0
    */
-  ethereumAddress(): string {
+  ethereumAddress(): `0x${string}` {
     const address = this.faker.string.hexadecimal({
       length: 40,
       casing: 'lower',
     });
+
```
Shinigami92, line 361: Nothing changed here, I think we can omit this blank line from this PR

```suggestion
```

Shinigami92 (commented):
We need to decide in a team meeting if we want this change

ST-DDT:
Looks like we forgot to publish our team decision. Sorry for the wait.

**Team decision**

- We think that this change is not worth the extra maintainance/code readability costs, mainly because the later part of `0x{string}` matches strings that aren't etherium addresses at all e.g. `'0xftgugloöiho!'`.
- There are potentially better ways to ensure that you only get etherium addresses instead of just strings. e.g. `type EtheriumAddress = string & {ZZ_IGNORE_ME: 'EtheriumAddress'}` (@Shinigami92 please correct me if I'm wrong and please add the proper name for that construct). If you have such a type in your sources, then you can only pass `EtheriumAddress`es to your methods, a plain string wouldn't work. That way you can ensure that you have to check (and cast) the user given string before using it anywhere. You can overwrite faker's etheriumAddress method type definitions locally, so that the compiler doesn't complain in your tests (@Shinigami92 could you please add an example for that as well?), alternatively write a wrapper method that casts the result accordingly. Since these types are specific to your application, we cannot cast/provide it in our code.

````ts
type EtheriumAddress = string & {zzz_ignoreMe: 'EtheriumAddress'};

function validateEtheriumAddress(input: string): EtheriumAddress {
    // actual validation
    return input as EtheriumAddress;
};
function transfer(source: EtheriumAddress, target: EtheriumAddress, amount: number): void {
    // ...
}
````

**Comment by Shinigami92** [comment]
@ST-DDT both your suggestions are slightly not correct but have the right intend

---

The first you are suggesting are branded / nominal types

- https://github.com/Microsoft/Typescript/issues/202

As you can see the issue number is just 3 digits, it's one of the oldest requested features that are sadly still not nativly supported and therefore should only be used in closed context and caution
I wont suggest to export a nominal type from a lib like Faker

There are some libs that can help to construct some branded types and I use them myself for branded IDs in my frontends, but I never export them from a lib to be used outside context

---

The second you are talking of are [type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

```ts
type EtheriumAddress = whatever;

function isEtheriumAddress(value: any): value is EtheriumAddress {
	//                                  ^^^^^^^^ this is a type-guard
	return typeof value === 'string' && /the-regexp/.test(value)
}
```

PR #1569, opened 2022-11-19
feat(helpers): add fromRegExp method

### LINKED ISSUE #1359: allow letter range for `helpers.regexpStyleStringParse`
This issue represents a todo from our codebase:

https://github.com/faker-js/faker/blob/5ae93a7f51d48b2f2f4999f5cfbd2b4a4a5bb272/src/modules/helpers/index.ts#L244-L245

Currently, it is possible to generate a random number in a string with the `regexpStyleStringParse` functions. For example `faker.helpers.regexpStyleStringParse('[500-15000]')` will return a string representing a number between 500 and 15000 (for example `'8375'`).

This should also be possible for characters.
For example:
- `faker.helpers.regexpStyleStringParse('[A-Z]')` will return an uppercase letter between A and Z (`'K'`)
- `faker.helpers.regexpStyleStringParse('[s-w]')` will return an uppercase letter between s and w (`'u'`)
- `faker.helpers.regexpStyleStringParse('[A-Df-t]')` will return an uppercase letter between A and D or f and t (`'s'`)

wooneusean:
Fixes #1359 

Example:
```javascript
faker.helpers.regexpStyleStringParse('Test[a-d1-4]{2,4}'); // 'Testda3'
faker.helpers.regexpStyleStringParse('Test[^a-zA-Z0-8]{2}'); // 'Test99'
faker.helpers.regexpStyleStringParse('Test[-a-z]'); // Test-
```


src/modules/helpers/index.ts
```diff
@@ -173,21 +173,122 @@ export class HelpersModule {
    * faker.helpers.regexpStyleStringParse() // ''
    * faker.helpers.regexpStyleStringParse('#{5}') // '#####'
    * faker.helpers.regexpStyleStringParse('#{2,9}') // '#######'
-   * faker.helpers.regexpStyleStringParse('[500-15000]') // '8375'
+   * faker.helpers.regexpStyleStringParse('[1-7]') // '5'
```
ST-DDT: Why did you remove/cahnge that example?
Doesn't it work anymore?

wooneusean: Yes and no, it will work but it is not the same outcome. RegEx will see [100-200] as 1, 0, 0-2, 0, 0. Which means it will accept 0, 1 and 2.

ST-DDT: So now it is generating a string for proper regexes?
You have to adjust the jsdocs though because it still references the number ranges.
However I think it might be better to introduce it as a new method to the string module e.g. `faker.string.forRegex` instead of breaking the behavior of the existing method.
What do you think?

wooneusean: > So now it is generating a string for proper regexes?

yep
> You have to adjust the jsdocs though because it still references the number ranges.

Will do

> However I think it might be better to introduce it as a new method to the string module e.g. faker.string.forRegex instead of breaking the behavior of the existing method.
What do you think?

Indeed that would be good for not breaking stuff, but if I were to create a new function, it will practically do the same thing, and users might not immediately know the difference and might cause confusion. I feel, for correctness and also for less confusion, to update the current function with the correct implementation.

ST-DDT: > Indeed that would be good for not breaking stuff, but if I were to create a new function, it will practically do the same thing, and users might not immediately know the difference and might cause confusion. I feel, for correctness and also for less confusion, to update the current function with the correct implementation.

An alternative for that would be deprecating the old implementation when introducing the new method.
We will discuss the best way in a team meeting.

ST-DDT (dismissed):
Looks good to me.

However, we have to decide whether we want to introduce this as a new method (e.g. `faker.string.forRegex`) because it changes the behavior of the method.

ST-DDT (dismissed):

Shinigami92 (changes requested):
Yeah I'm with @ST-DDT, but I would maybe name it `string.fromRegExp`
`from` instead of `for` because it relates to the input parameter
and `RegExp` because that is how the class is named in JS

ST-DDT:
> and `RegExp` because that is how the class is named in JS

Good idea, then it should accept actual Regexp types as well. `( pattern: RegExp | string)`

````ts
if (pattern instanceof RegExp) {
    pattern = pattern.toString();
}
````

it should also be prepared for regexp options such as case (in)sensitive.

**Comment by wooneusean** [comment]
> Good idea, then it should accept actual Regexp types as well. ( pattern: RegExp | string)

Looking around online, I came across another library called [RandExp.js](https://github.com/fent/randexp.js) that does exactly this. We can just use this library to facilitate that functionality. What do you guys think?

**Comment by Shinigami92** [comment]
`@faker-js/faker` is a dependency-free project right now, just to support this for a function that is used by <1% of our users might be really overkill
I would then more be of the favor of deprecating the function and log a warning to directly use their solution
But our current solution depends on `this.faker.datatype.number`-call and therefore depends on a deterministic seed
I think the `RandExp.js` solution does not provide that, does it?

**Comment by wooneusean** [comment]
> But our current solution depends on this.faker.datatype.number-call and therefore depends on a deterministic seed
I think the RandExp.js solution does not provide that, does it?

Hmm you have a point. I'm new to open source, so tell me if this sounds crazy or not: I copy their implementation and use calls to faker's random number generation. That way we can have their functionality, and also keep the deterministic nature of the function.

This also sounds like overkill but its more realistic than writing one from scratch, that may have errors and bugs all over the place.

**Comment by ST-DDT** [comment]
IMO we should use the current implementation as a basic implementation. We can add a comment in the jsdocs that for complete regex support the users can use that library and explain how you can use it with faker (e.g. `randexp.randInt = (min, max) => faker.datatype.number({min, max})`) but not much more.

EDIT: That library is no longer maintained, so I don't think we should explicitly link it.

**Comment by Shinigami92** [comment]
For now my first thought would that this is definitely to much code (in terms of `loc`)
But in a later version of faker > 8.0 we want to restructure the project and move every single function to it's own file
Then it would not be anymore such a huge problem as everything is just one file that depends to this regex things

IMO even the current PR looks really long for just one function

Not sure what I can suggest you now to proceed or maybe just wait, but this could take months :confused:

**Comment by wooneusean** [comment]
I guess there's no rush right? As you mentioned <1% of users probably use this functionality anyways. For now, I'll just wait 😐

**Comment by ST-DDT** [comment]
Sorry for the long wait.

**Team decision**

- For now we don't change the existing method
- Instead we will add a new method `string.fromRegExp(string | RegExp): string`
- We will create a separate issue regarding deprecation of the old method

**Comment by wooneusean** [comment]
Alright noted.
 
This seems like it has become a more complex problem as tokenising RegEx is needed to properly generate correct text (according to the previously mentioned library). If I were to try to implement it it might be a huge change.

Whats the course of action now?

**Comment by ST-DDT** [comment]
IMO We dont require full regex support.
Just mention which features are supported and add a text like `if you need full regexp support, then search for compatible third party libraries.`
If neccessary we can extend our featureset in the future.

**Comment by wooneusean** [comment]
Alright noted.

As my changes stand now, do you guys think it is enough? Or are there other specific things that are needed to be supported?

**Comment by ST-DDT** [comment]
> As my changes stand now, do you guys think it is enough? Or are there other specific things that are needed to be supported?

I think it is enough, if we need more features we can do that in a later PR.

**Comment by wooneusean** [comment]
I reverted the `regexpStyleStringParse` function but I copied over my previous changes with that function and renamed it as I still needed the other functionalities.

File: src/modules/helpers/index.ts [hunk 2]
```diff
@@ -238,6 +238,181 @@ export class HelpersModule {
...
+   * faker.helpers.fromRegExp() // ''
+   * faker.helpers.fromRegExp('#{5}') // '#####'
+   * faker.helpers.fromRegExp('#{2,9}') // '#######'
+   * faker.helpers.fromRegExp('[1-7]') // '5'
+   * faker.helpers.fromRegExp('#{3}test[1-5]') // '###test3'
+   * faker.helpers.fromRegExp('[0-9a-dmno]') // '5' | 'c' | 'o'
+   * faker.helpers.fromRegExp('[^a-zA-Z0-8]') // '9'
+   * faker.helpers.fromRegExp('[a-d0-6]{2,8}') // 'a0' | 'd6' | 'a0dc45b0'
+   * faker.helpers.fromRegExp('[-a-z]{5}') // 'a-zab'
+   * faker.helpers.fromRegExp(new RegExp('[A-Z0-9]{4}-[A-Z0-9]{4}')) // 'BS4G-485H'
```
ST-DDT: Maybe use the regex syntax directly.

```suggestion
   * faker.helpers.fromRegExp(/[A-Z0-9]{4}-[A-Z0-9]{4}/) // 'BS4G-485H'
```

src/modules/helpers/index.ts
```diff
@@ -238,6 +238,181 @@ export class HelpersModule {
...
+   *
+   * @since 5.0.0
```
```suggestion
   * @since 8.0.0
```

test/helpers.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('helpers', () => {
...
+    t.describe('fromRegExp', (t) => {
+      t.it('noArgs')
+        .it('only symbols', '#{3}test[1-5]')
+        .it('some string', 'Hello !#{3}test[1-5]')
+        .it('character class', 'Hello !#{3,5}test[1-5a-g]{10}[5-8]{2,6}')
+        .it('dash', '!#{3,5}test[-1-5a-g]{10}[-5-8]{2,6}')
+        .it('negate', '!#{3,5}test[^1-5a-g]{10}[^-5-8]{2,6}')
+        .it('negate and dash', '!#{3,5}test[^-1-5a-g]{10}[^-5-8]{2,6}')
+        .it('swap range if start is greater than end', '[z-a]')
+        .it('with RegExp object', new RegExp('[A-D0-9]{4}-[A-D0-9]{4}'));
```
ST-DDT:
```suggestion
        .it('with RegExp object', /[A-D0-9]{4}-[A-D0-9]{4}/);
```

plus one with regex options such as case insensitive.

ST-DDT:
Should we move it to the string module?

wooneusean:
Will address those changes you suggested.

> Should we move it to the string module?

I'm not sure, but I guess since it almost does the same thing as `regexpStyleStringParse` maybe it should just be left there?

Also, there seems to be this failed test and I am not sure how to fix it

```
 ❯ test/scripts/apidoc/examplesAndDeprecations.spec.ts:106:27
    104|       } else {
    105|         for (const spy of consoleSpies) {
    106|           expect(spy).not.toHaveBeenCalled();
       |                           ^
    107|         }
    108|       }
```

EDIT:
The failed test fixed itself somehow, so nevermind that.

wooneusean:
New changes a bit messy, but its the only way I could think of adding case insensitive safely.

Also, I found out splicing with negative values remove from the back of the list. So I have to do some ugly code to address that.

src/modules/helpers/index.ts
```diff
@@ -238,6 +238,237 @@ export class HelpersModule {
...
+      }
+
+      const generatedString = Array.from(Array(repetitions)).reduce(
+        (acc: string) =>
+          acc +
+          String.fromCharCode(
+            rangeCodes[this.faker.number.int(rangeCodes.length - 1)]
+          ),
+        ''
+      );
```
ST-DDT: You can probably use `faker.helpers.multiple` + `String.fromCharCode(faker.helpers.arrayElement(...))` + `join('')` for that.

ST-DDT (commented):
This method implementation looks longer than expected.

Please note, that we will probably not or be slower to review this PR again during the holidays.

wooneusean:
> This method implementation looks longer than expected.

I'll try to shorten it

Shinigami92 (dismissed):
I assume @ST-DDT also want some non seeded test, does he?
Otherwise, it's a start :+1:

wooneusean:
Non-seeded tests are referring to the "Random Seeded Tests" right?

ST-DDT (changes requested):
The tests are failing. Please update the PR and fix the errors.

wooneusean:
> The tests are failing. Please update the PR and fix the errors.

Will add unseeded tests and fix errors soon.

wooneusean:
I see a v8 branch, do I merge that into this branch or do I still merge next?

I ask this because currently I merged next into my branch and ran tests but it keeps getting stuck like this
image

ST-DDT:
The next branch is the branch for continous development.
The vX branches are for completed release versions.
So the next branch is the correct branch to target.

For some reason, the massive logs created by deprecated methods causes the tests to freeze up locally. CI doesnt seem to have that problem. We are currently investigating that problem.

The easiest workaround for that problem is testing singular test files with: `pnpm run test test/some.spec.ts`.

wooneusean:
Also, this message shows up even though I have never changed my git config. No idea why this is happening.

image

Shinigami92:
> Also, this message shows up even though I have never changed my git config. No idea why this is happening.
> 
> image

Please run `rm -rf .git/hooks`

pnpm-lock.yaml
```diff
@@ -53,21 +53,21 @@ devDependencies:
   '@algolia/client-search': 4.14.3
   '@types/glob': 8.0.0
   '@types/markdown-it': 12.2.3
-  '@types/node': 18.11.17
+  '@types/node': 18.11.18
```
ST-DDT: Please revert these changes.

wooneusean: Will update soon.

wooneusean: This change is done by `pnpm install` though, and I am not sure how to forcibly downgrade without changing `package.json` because in that file, the version for `@types/node` is `~18.11.17`

Shinigami92: Try
```
git checkout origin/next pnpm-lock.yaml
pnpm install
```

wooneusean: got it, thanks.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,208 @@ export class HelpersModule {
...
+   * Supported patterns:
+   * - `.{times}` => Repeat the character exactly `times` times.
+   * - `.{min,max}` => Repeat the character `min` to `max` times.
+   * - `[x-y]` => Randomly get a character between `x` and `y` (inclusive).
+   * - `[x-y]{times}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `times` times.
+   * - `[x-y]{min,max}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `min` to `max` times.
+   * - `[^...]` => Randomly get a character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character)
+   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used . (e.g. `[^-0-9]` will not get any numeric characters or dashes)
+   *
+   * @param pattern The template string to to parse.
```
```suggestion
   * @param pattern The template string/RegExp to generate a matching string for.
```

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,208 @@ export class HelpersModule {
     return string;
   }
 
+  /**
+   * Replaces the regex like expressions in the given string with matching values.
```
```suggestion
   * Generates a string matching the given regex like expressions.
```

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,208 @@ export class HelpersModule {
...
+   * faker.helpers.fromRegExp('[0-9a-dmno]') // '5' | 'c' | 'o'
+   * faker.helpers.fromRegExp('[^a-zA-Z0-8]') // '9'
+   * faker.helpers.fromRegExp('[a-d0-6]{2,8}') // 'a0' | 'd6' | 'a0dc45b0'
+   * faker.helpers.fromRegExp('[-a-z]{5}') // 'a-zab'
+   * faker.helpers.fromRegExp(/[A-Z0-9]{4}-[A-Z0-9]{4}/) // 'BS4G-485H'
+   * faker.helpers.fromRegExp(/[A-Z]{5}/i) // 'pDKfh'
+   *
+   * @since 8.0.0
+   */
+  fromRegExp(pattern: string | RegExp = ''): string {
```
ST-DDT: Not sure about the empty string default.

test/helpers.spec.ts
```diff
@@ -47,6 +47,26 @@ describe('helpers', () => {
...
 
+    t.describe('fromRegExp', (t) => {
+      t.it('noArgs')
+        .it('only symbols', '#{3}test[1-5]')
+        .it('some string', 'Hello !#{3}test[1-5]')
+        .it('character class', 'Hello !#{3,5}test[1-5a-g]{10}[5-8]{2,6}')
+        .it('dash', '!#{3,5}test[-1-5a-g]{10}[-5-8]{2,6}')
+        .it('negate', '!#{3,5}test[^1-5a-g]{10}[^-5-8]{2,6}')
+        .it('negate and dash', '!#{3,5}test[^-1-5a-g]{10}[^-5-8]{2,6}')
+        .it('swap range if start is greater than end', '[z-a]')
```
ST-DDT: Could you please remove the parts that are not relevant for the individual test cases?

wooneusean: This is necessary as it still needs to conform to the functionality of `regexpStyleStringParse()`.

What can be done in `regexpStyleStringParse()` should be able to be done in `fromRegExp()` as well is what I mean.

src/modules/helpers/index.ts
wooneusean: What is your concern?

ST-DDT: If there is no regex that the result should match, then there isn't a need to generate anything in the first place.
(the default doesn't add any value to the method)

test/helpers.spec.ts
ST-DDT: > This is necessary as it still needs to conform to the functionality of `regexpStyleStringParse()`.
> 
> What can be done in `regexpStyleStringParse()` should be able to be done in `fromRegExp()` as well is what I mean.

I disagree. Both methods are independent of each other.
One is for a custom (non-)regex, the other is for a subset of regex.

wooneusean: I see, noted. Will remove those tests that you marked.

src/modules/helpers/index.ts
wooneusean: Ah I see, so what you want me to add is an early return when `pattern` is an empty string?

ST-DDT: No, disallow calling `fromRegExp()` (without any arguments).

wooneusean: Alright noted, will disallow calling without arguments.

ST-DDT:
Please fix the failing tests.

wooneusean:

Weird, testing locally doesn't show that error. Will address it.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,210 @@ export class HelpersModule {
...
+   */
+  fromRegExp(pattern: string | RegExp): string {
+    let isCaseInsensitive = false;
+    if (pattern instanceof RegExp) {
+      isCaseInsensitive = pattern.flags.includes('i');
+      pattern = pattern.toString();
+      pattern = pattern.match(/\/(.+?)\//)?.[1] ?? ''; // Remove frontslash from front and back of RegExp
+    }
+
+    if (pattern == null) return '';
```
ST-DDT: AFAICT pattern should never be null here, if a pattern is given.
And since it is required, this check should be obsolete.

wooneusean: That is what I thought as well but the failing tests show the pattern is undefined when it is being used with `.match()` a few lines below this. The test that failed isn't written by me though, I think its some automated test thing but I'm not sure.

ST-DDT: One of your examples doesn't have an argument thus causes the illegal invocation:

https://github.com/wooneusean/faker/blob/9b81969f5499f002775892535cd25fc725532b18/src/modules/helpers/index.ts#L268

If you remove that line, you can also remove this one.

(We have tests that ensure that the jsdocs examples work or at least don't throw an error^^)

wooneusean: > (We have tests that ensure that the jsdocs examples work or at least don't throw an error^^)

Wow, I learn something new everyday. Will remove that example.

ST-DDT: Please also remove the null check here.

ST-DDT (dismissed):
Looks good to me. I'm just ensure whether this should be in the `string` module instead.

Shinigami92 (dismissed):

test/helpers.spec.ts
```diff
@@ -427,6 +439,57 @@ describe('helpers', () => {
...
 
+      describe('fromRegExp()', () => {
+        it('deals with range repeat', () => {
+          const string = faker.helpers.fromRegExp(/#{5,10}/);
+          expect(string.length).toBeLessThanOrEqual(10);
+          expect(string.length).toBeGreaterThanOrEqual(5);
+          expect(string).toMatch(/^\#{5,10}$/);
+        });
+
+        it('flips the range when min > max', () => {
```
xDivisionByZerox: Is this really a desired behavior? Wouldn't the use-case be something like:
```ts
const regex = validator.validationRegExp;
const value = faker.helpers.fromRegExp(regex);
// [...]
const isValid = validator.isValid(value);
```

Since we are working with real `RegExp` here I think an invalid RegExp should be treated as such.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,207 @@ export class HelpersModule {
...
+   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used . (e.g. `[^-0-9]` will not get any numeric characters or dashes)
+   *
+   * @param pattern The template string/RegExp to to generate a matching string for.
+   *
+   * @example
+   * faker.helpers.fromRegExp('#{5}') // '#####'
+   * faker.helpers.fromRegExp('#{2,9}') // '#######'
+   * faker.helpers.fromRegExp('[1-7]') // '5'
+   * faker.helpers.fromRegExp('#{3}test[1-5]') // '###test3'
+   * faker.helpers.fromRegExp('[0-9a-dmno]') // '5' | 'c' | 'o'
```
xDivisionByZerox: We don't have multiple example result for a single example in any other place (please correct me if I'm wrong). I see the benefit but would suggest to stick with one example result here as well.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,207 @@ export class HelpersModule {
...
 
+  /**
+   * Generates a string matching the given regex like expressions.
+   *
+   * This function doesn't provide full support of actual `RegExp`.
+   * Features such as grouping, anchors and character classes are not supported.
+   * If you are looking for a library that randomly generates strings based on
+   * `RegExp`s, see [randexp.js](https://github.com/fent/randexp.js)
+   *
+   * Supported patterns:
```
xDivisionByZerox, line 324: Should the list include the use of case insensitivity via an actual `RegExp`?

test/helpers.spec.ts
ST-DDT: > Uncaught SyntaxError: numbers out of order in {} quantifier.

Since it isn't supported by js, we shouldn't support it either.

wooneusean: What should I do here? Leave it as it is or remove the flipping?

ST-DDT: Remove the flipping and throw an error instead.

src/modules/helpers/index.ts
```diff
@@ -335,11 +336,9 @@ export class HelpersModule {
           const rangeMinMax = range[0].split('-').map((x) => x.charCodeAt(0));
           min = rangeMinMax[0];
           max = rangeMinMax[1];
-          // switch min and max
+          // throw error if min larger than max
```
xDivisionByZerox, line 436: From our [docs/contribution guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#jsdocs)

> If a function can throw an error (FakerError) you have to include the @throws tag with an explanation when an error could be thrown

So could you please:
1. Throw a `FakerError` as well as test for that to be thrown ([example](https://github.com/faker-js/faker/blob/795ca8cc52b357e03447a282469c6d1fb44fc960/test/faker.spec.ts#L16-L20))
2. Add a `@throws` param as well a description to the JSDocs

I should have been more clear on the first comment. Thats my bad.

test/helpers.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('helpers', () => {
...
+    t.describe('fromRegExp', (t) => {
+      t.it('with RegExp object', /[A-D0-9]{4}-[A-D0-9]{4}/)
+        .it(
+          'with negated RegExp object with case insensitive flag',
+          /[^a-t0-9]{4}/i
+        )
+        .it(
+          'with RegExp object and case insensitive flag',
+          /[A-D0-9]{4}-[A-D0-9]{4}/i
+        );
```
ST-DDT:
```suggestion
      t.it('with static string', 'Hello World!')
        .it('with static RegExp', /Hello World!/)
        .it('with dynamic string', '[A-D0-9]-[A-D0-9]-A')
        .it('with dynamic RegExp', /[A-D0-9]-[A-D0-9]-A/)
        .it('with wildcard character', /./)
        .it('with optional character', /A?-[A-D0-9]?-[a-d0-4]?/)
        .it('with optional repetition', /A*-[A-D0-9]*-[a-d0-4]*/)
        .it('with required repetition', /A+-[A-D0-9]+-[a-d0-4]+/)
        .it('with quantifier', /A{2}-[A-D0-9]{4}-[a-d0-4]{6}/)
        .it('with quantifier ranges', /A{2,6}-[A-D0-9]{4,6}-[a-d0-4]{6,8}/)
        .it('with case insensitive flag', /[A-D0-9]{10}/i)
        .it('with negation and case insensitive flag', /[^a-t0-7]{10}/i)
        .it('with negation', /[^A-Za-y0-9]{10}/);
```

Could you add support for the missing cases as well?

ST-DDT (commented):
I'm quite unsure which features should be supported in the MVP.

test/helpers.spec.ts
wooneusean: I could, it will take some time and it will make the function larger and more complex than it currently is which I assume is a concern from previous reviews.

If you don't mind the size and complexity then only time is the problem as I have not much of it nowadays.

From glancing over the above suggestions, I have a few concerns but besides that I think most of the cases are already covered. Particularly the optional and required repetition, I am assuming there should be some kind of limit to how long the repetition can be. Do you guys have any preference?

Also, for the sake of simplicity, I'm thinking the wildcard character should contain characters `[A-Za-z0-9]` only. Is that ok?

ST-DDT: @faker-js/maintainers @faker-js/members `@anyone`: Any preferences?

xDivisionByZerox: I wouldn't actually put any restrictions on the repetition part. I think something like `while (faker.datatype.boolean()) { [...] }` instead. But a static number that gets generated upfront might be easier to handle.

For the wildcard, all functions from the string module (including special characters) can be used IMO.

ST-DDT: We could use an option to overwrite what `.` is and it can be used as a base for `[^something]`.

wooneusean: > `while (faker.datatype.boolean()) { [...] }`

Wont this make it so that higher repetitions are rarer?

> But a static number that gets generated upfront might be easier to handle.

I agree but the concern with that is like what I stated above, what should be the maximum number of repetitions? If it randomly generates some large number, it will still work but its kinda unreasonable, no?

> For the wildcard, all functions from the string module (including special characters) can be used IMO.

I see a function `faker.string.sample()`, is this function suitable for generating wildcard characters?
There is also another concern, supposedly `.` matches all characters except `\n`. Chinese, Cyrillic etc characters also need to be included but to include those is also a huge pain, and I have no idea where to start with those. Can I sweep this concern under the rug?

> We could use an option to overwrite what `.` is and it can be used as a base for `[^something]`.

I think this is not appropriate. If the user wants to give their own definition, they can just do it in normal character sets, i.e `/[a-z0-9]/`.

xDivisionByZerox: > Wont this make it so that higher repetitions are rarer?

Yes, you are right. I wanted to include the current interaction count as a `probability` parameter in the function (but forgot to actually write it). This would then look something like this: `while (faker.datatype.boolean(1 / currentIteration)) { [...] }` which equates to a logarithmic distribution of repetitions.

> I agree but the concern with that is like what I stated above, what should be the maximum number of repetitions? If it randomly generates some large number, it will still work but its kinda unreasonable, no?

If this is way easyer for you to implement for you - go for it and hard code the max number to an arbitrary number (like 25). Don't think to much about it. Altho, one could argue that since it is a wildcard, an infinite number of repetitions is expected by design.

> I see a function faker.string.sample(), is this function suitable for generating wildcard characters?
There is also another concern, supposedly . matches all characters except \n. Chinese, Cyrillic etc characters also need to be included but to include those is also a huge pain, and I have no idea where to start with those. Can I sweep this concern under the rug?

Yes, this is perfect. Altho `.` matches anything I would argue that it is simple enough to return any character from our string methods. Keep in mind that this should be a helper function, not a full blown library on its own^^

> I think this is not appropriate. If the user wants to give their own definition, they can just do it in normal character sets, i.e /[a-z0-9]/.

I'm with you. RegExp lets you do some fantastic things. If you want a specific char set, don't use a wildcard matcher.

wooneusean:
I suddenly get this error when running tests

image

ST-DDT:
Please run `pnpm install`. If that doesn't help merge `next` and run `pnpm install` again.

wooneusean:
Sorry for the delay guys, I just got past a busy period. Will start back on this soon.

src/modules/helpers/index.ts
```diff
@@ -254,14 +254,21 @@ export class HelpersModule {
...
+   * - `x{times}` => Repeat the `x` exactly `times` times.
+   * - `x{min,max}` => Repeat the `x` `min` to `max` times.
    * - `[x-y]` => Randomly get a character between `x` and `y` (inclusive).
    * - `[x-y]{times}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `times` times.
    * - `[x-y]{min,max}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `min` to `max` times.
-   * - `[^...]` => Randomly get a character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character)
-   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used . (e.g. `[^-0-9]` will not get any numeric characters or dashes)
-   * - `/[x-y]/i` => Randomly gets an uppercase or lowercase character between `x` and `y` (inclusive)
+   * - `[^...]` => Randomly get a character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character).
+   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used . (e.g. `[^-0-9]` will not get any numeric characters or dashes).
```
```suggestion
   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used (e.g. `[^-0-9]` will not get any numeric characters or dashes).
```

pnpm-lock.yaml
```diff
@@ -52,47 +52,47 @@ devDependencies:
   '@algolia/client-search': 4.14.3
   '@types/glob': 8.0.1
   '@types/markdown-it': 12.2.3
-  '@types/node': 18.11.18
+  '@types/node': 18.11.19
```
ST-DDT: Please revert all changes to this file:

````sh
git fetch --all
git reset origin/next -- pnpm-lock.yaml
````

(Not tested)

File: src/modules/helpers/index.ts [hunk 19]
```diff
@@ -245,6 +245,302 @@ export class HelpersModule {
...
+      if (quantifierSymbol) {
+        switch (quantifierSymbol) {
+          case '?': {
+            repetitions = this.faker.datatype.boolean() ? 0 : 1;
+            break;
+          }
+
+          case '*': {
+            repetitions = 0;
+            while (this.faker.datatype.boolean(1 / (repetitions + 1))) {
```
ST-DDT: AFAICT this will always be true on the first run (1 / (0 + 1)= => 1 / 1 => 1 => true.
IMO just ``while( .......boolean())`` should be enough.

Also this favors very small numbers. I suggest using:

````ts
let limit = 1;
while (boolean()) {
 limit *= 2;
}
repetitions = int(min: 0, max: limit);
````

(if the user doesn't want large strings, then they shouldn't use */+ in the first place, because I as a tester love to destroy applications with this kind of loose validation)

File: src/modules/helpers/index.ts [hunk 20]
```diff
@@ -245,6 +245,302 @@ export class HelpersModule {
...
+              repetitions++;
+            }
+
+            break;
+          }
+
+          case '+': {
+            repetitions = 1;
+            while (this.faker.datatype.boolean(1 / repetitions)) {
```
ST-DDT: Same here

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,302 @@ export class HelpersModule {
...
+              repetitions++;
+            }
+
+            break;
+          }
+
+          // Technically would never happen
+          default:
+            repetitions = 1;
```
ST-DDT: Then lets throw an error here.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,302 @@ export class HelpersModule {
...
+            while (this.faker.datatype.boolean(1 / repetitions)) {
+              repetitions++;
+            }
+
+            break;
+          }
+
+          // Technically would never happen
+          default:
```
ST-DDT: This is the same as above.

src/modules/helpers/index.ts
```diff
@@ -245,6 +245,302 @@ export class HelpersModule {
...
+   * If you are looking for a library that randomly generates strings based on
+   * `RegExp`s, see [randexp.js](https://github.com/fent/randexp.js)
+   *
+   * Supported patterns:
+   * - `x{times}` => Repeat the `x` exactly `times` times.
+   * - `x{min,max}` => Repeat the `x` `min` to `max` times.
+   * - `[x-y]` => Randomly get a character between `x` and `y` (inclusive).
+   * - `[x-y]{times}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `times` times.
+   * - `[x-y]{min,max}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `min` to `max` times.
+   * - `[^...]` => Randomly get a character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character).
```
ST-DDT: Please add a hint that this currently refers to ascii characters.

Maybe like this:

```suggestion
   * - `[^...]` => Randomly get a ascii number or letter character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character).
```

?

ST-DDT (approved):

ST-DDT (changes requested):
Ups, I pressed the wrong button

pnpm-lock.yaml
wooneusean: How do I prevent this from happening again, cause it automatically updates this when I run `pnpm install`.

wooneusean:
Should I put the code to get the repetitions based on the symbol into an inner function within the `fromRegExp` function? Or is this considered sacrilegious? 

something like:
```typescript
 fromRegExp(pattern: string | RegExp): string {
    function getRepetitionsFromQuantifier(quantifierSymbol: string) {
      switch (quantifierSymbol) {
        case '?': {
          return this.faker.datatype.boolean() ? 0 : 1;
        }

        case '*': {
          let limit = 1;
          while (this.faker.datatype.boolean()) {
            limit *= 2;
          }
          return this.faker.number.int({ min: 0, max: limit });
        }

        case '+': {
          let limit = 1;
          while (this.faker.datatype.boolean()) {
            limit *= 2;
          }
          return this.faker.number.int({ min: 1, max: limit });
        }

        default:
          throw new FakerError('Unknown quantifier symbol provided.');
      }
    }
...
```

ST-DDT:
> Should I put the code to get the repetitions based on the symbol into an inner function within the `fromRegExp` function?

If you wrap that logic into a function that will reduce code repetition and thus help maintaining code readability. There are two solutions that I would be fine with.

- A big lambda expression inside the method
- A separate function above the class definition that also takes a faker instance as parameter. See also https://github.com/faker-js/faker/blob/11a5f51ddb4ab1163d770c4b12f5fe4711b8e11d/src/modules/person/index.ts#L22

src/modules/helpers/index.ts
```diff
@@ -247,6 +290,253 @@ export class HelpersModule {
...
+      } else if (quantifierMin && quantifierMax) {
+        repetitions = this.faker.number.int({
+          min: parseInt(quantifierMin),
+          max: parseInt(quantifierMax),
+        });
+      } else if (quantifierMin && !quantifierMax) {
+        repetitions = parseInt(quantifierMin);
+      } else {
+        repetitions = 1;
+      }
```
ST-DDT: This part also looks very much like the other one.

src/modules/helpers/index.ts
```diff
@@ -247,6 +290,253 @@ export class HelpersModule {
...
```
ST-DDT: This one

ST-DDT (commented):
Could you please move the marked code to a separate method/move it to the already extracted method as well?

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -4,6 +4,74 @@ import { luhnCheckValue } from './luhn-check';
...
+        }
+
+        repetitions = faker.number.int({ min: 1, max: limit });
+        break;
+      }
+
+      default:
+        throw new FakerError('Unknown quantifier symbol provided.');
+    }
+  } else if (quantifierMin && quantifierMax) {
```
Shinigami92: Is e.g. `quantifierMin == 0` allowed here?

wooneusean: I would assume yes because `faker.number.int` will handle the error checking if the user makes the `quantifierMin` more than the `quantiferMax` or other problems related to it.

ST-DDT: Do we then have to use 

```suggestion
  } else if (quantifierMin != null && quantifierMax != null) {
```

instead?

wooneusean: Oh yeah that makes sense since `(quantifierMin && quantifierMax)` will be false if both of them are 0 which is wrong. I will update this soon.

src/modules/helpers/index.ts
```diff
@@ -4,6 +4,72 @@ import { luhnCheckValue } from './luhn-check';
...
+ *
+ * @since 8.0.0
+ */
+function getRepetitionsBasedOnQuantifierParameters(
+  faker: Faker,
+  quantifierSymbol: string,
+  quantifierMin: string,
+  quantifierMax: string
+) {
+  let repetitions = 0;
```
ST-DDT: Note to self: Check if `0` is the correct fallback here. `1` might be more correct. Maybe throwing an error would be best?

wooneusean: After thinking about it, `1` seems more reasonable, because now it just removes the character completely if there's a problem.

I'll change it to `1`.

src/modules/helpers/index.ts
```diff
@@ -247,6 +313,235 @@ export class HelpersModule {
...
+   * If you are looking for a library that randomly generates strings based on
+   * `RegExp`s, see [randexp.js](https://github.com/fent/randexp.js)
+   *
+   * Supported patterns:
+   * - `x{times}` => Repeat the `x` exactly `times` times.
+   * - `x{min,max}` => Repeat the `x` `min` to `max` times.
+   * - `[x-y]` => Randomly get a character between `x` and `y` (inclusive).
+   * - `[x-y]{times}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `times` times.
+   * - `[x-y]{min,max}` => Randomly get a character between `x` and `y` (inclusive) and repeat it `min` to `max` times.
+   * - `[^...]` => Randomly get an ACSCII number or letter character that is not in the given range. (e.g. `[^0-9]` will get a random non-numeric character).
```
matthewmayer: ACSCII -> ASCII

src/modules/helpers/index.ts
```diff
@@ -247,6 +313,235 @@ export class HelpersModule {
...
+   * - `[-...]` => Include dashes in the range. Must be placed after the negate character `^` and before any character sets if used (e.g. `[^-0-9]` will not get any numeric characters or dashes).
+   * - `/[x-y]/i` => Randomly gets an uppercase or lowercase character between `x` and `y` (inclusive).
+   * - `x?` => Randomly decide to include or not include `x`.
+   * - `[x-y]?` => Randomly decide to include or not include characters between `x` and `y` (inclusive).
+   * - `x*` => Repeat `x` 0 or more times.
+   * - `[x-y]*` => Repeat characters between `x` and `y` (inclusive) 0 or more times.
+   * - `x+` => Repeat `x` 1 or more times.
+   * - `[x-y]+` => Repeat characters between `x` and `y` (inclusive) 1 or more times.
+   * - `.` => returns a wildcard character that can be any number, character or symbol. Can be combined with quantifiers as well.
```
matthewmayer: maybe also mention that this is ASCII only?

matthewmayer (changes requested):

ST-DDT (approved):

Shinigami92 (approved):

matthewmayer (approved):

PR #1578, opened 2022-11-21
feat(finance): Add ethereum checksummed address faking

### LINKED ISSUE #1567: Adding checksummed ethereum addresses

When developing crypto-related applications, wallet addresses are of great importance and they actually exist in a number of different formats. One is a non-checksummed address and one is checksummed.

Non Checksum ( lowercase ): `0x4bbeeb066ed09b7aed07bf39eee0460dfa261520`
Checksum ( caps letters ): `0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520`

We have a use case in our e2e tests for testing checksummed and non-checksummed addresses.

### Suggested solution

I would propose a solution to implement a new finance method which simply is a checksummed version of the string.

Current: `faker.finance.ethereumAddress()`

New:
1. `faker.finance.ethereumAddress()`
2. `faker.finance.ethereumAddressChecksum()`

### Alternative

_No response_

### Additional context

_No response_

RareBodhi:
Closes https://github.com/faker-js/faker/issues/1567

Adds eth checksum as optional parameter with test to ensure that a "capital" letter isincluded which would indicate it's a checksum address

NOT READY FOR REVIEW

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,8 @@
     "vite": "~3.2.3",
     "vitepress": "1.0.0-alpha.28",
     "vitest": "~0.25.1",
-    "vue": "~3.2.45"
+    "vue": "~3.2.45",
+    "web3-utils": "^1.8.1"
   },
   "packageManager": "pnpm@7.15.0",
   "engines": {
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -350,13 +350,14 @@ export class FinanceModule {
    *
    * @example
    * faker.finance.ethereumAddress() // '0xf03dfeecbafc5147241cc4c4ca20b3c9dfd04c4a'
+   * faker.finance.ethereumAddress(options: { type: 'checksum' }) // '0xf03dFeeCbAFc5147241Cc4c4cA20b3c9Dfd04C4A'
    *
    * @since 5.0.0
    */
-  ethereumAddress(): string {
+  ethereumAddress(options: { type: 'non-checksum' | 'checksum' } = {}): string {
     const address = this.faker.string.hexadecimal({
       length: 40,
-      casing: 'lower',
+      casing: options.type === 'checksum' ? 'mixed' : 'lower',
     });
     return address;
   }
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -354,7 +354,7 @@ export class FinanceModule {
    *
    * @since 5.0.0
    */
-  ethereumAddress(options: { type: 'non-checksum' | 'checksum' } = {}): string {
+  ethereumAddress(options: { type?: 'non-checksum' | 'checksum' } = {}): string {
     const address = this.faker.string.hexadecimal({
       length: 40,
       casing: options.type === 'checksum' ? 'mixed' : 'lower',
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -1,3 +1,4 @@
+import web3 from 'web3-utils';
 import type { Faker } from '../..';
 import { FakerError } from '../../errors/faker-error';
 import iban from './iban';
@@ -348,17 +349,29 @@ export class FinanceModule {
   /**
    * Generates a random Ethereum address.
    *
+   * @param options Ethereum address options.
+   * @param options.type Output a checksummed or non-checksummed ( lowercase ) address.
+   *
    * @example
    * faker.finance.ethereumAddress() // '0xf03dfeecbafc5147241cc4c4ca20b3c9dfd04c4a'
    * faker.finance.ethereumAddress(options: { type: 'checksum' }) // '0xf03dFeeCbAFc5147241Cc4c4cA20b3c9Dfd04C4A'
    *
    * @since 5.0.0
    */
-  ethereumAddress(options: { type?: 'non-checksum' | 'checksum' } = {}): string {
+  ethereumAddress(
+    options: {
+      type?: 'non-checksum' | 'checksum';
+    } = {}
+  ): string {
     const address = this.faker.string.hexadecimal({
       length: 40,
-      casing: options.type === 'checksum' ? 'mixed' : 'lower',
+      casing: 'lower',
     });
+
+    if (options.type === 'checksum') {
+      return web3.toChecksumAddress(address);
+    }
+
     return address;
   }
 
--- a/test/finance.spec.ts
+++ b/test/finance.spec.ts
@@ -1,5 +1,6 @@
 import isValidBtcAddress from 'validator/lib/isBtcAddress';
 import { afterEach, describe, expect, it } from 'vitest';
+import web3 from 'web3-utils';
 import { faker } from '../src';
 import { FakerError } from '../src/errors/faker-error';
 import ibanLib from '../src/modules/finance/iban';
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -140,7 +140,8 @@
     "vite": "~3.2.3",
     "vitepress": "1.0.0-alpha.28",
     "vitest": "~0.25.1",
-    "vue": "~3.2.45"
+    "vue": "~3.2.45",
+    "web3-utils": "^1.8.1"
```
Shinigami92, line 144: Right now we are a direct dependency free project and I would prefer to not add dependencies in the future.

This also would not work right now as you added it to the devDependencies ⚠️ 

Please find a solution that does not depend on a dependency

src/modules/finance/index.ts
```diff
@@ -348,16 +349,29 @@ export class FinanceModule {
...
    * @example
    * faker.finance.ethereumAddress() // '0xf03dfeecbafc5147241cc4c4ca20b3c9dfd04c4a'
+   * faker.finance.ethereumAddress(options: { type: 'checksum' }) // '0xf03dFeeCbAFc5147241Cc4c4cA20b3c9Dfd04C4A'
    *
    * @since 5.0.0
    */
-  ethereumAddress(): string {
+  ethereumAddress(
+    options: {
+      type?: 'non-checksum' | 'checksum';
```
Shinigami92, line 363: Do we only have these two options?
If so, maybe we could go with options.checksummed: boolean

But I even think about just to always return the checksummed variant as it looks like the non checked sum is just the `value.toLowerCase()`
IMO this could be done by the user outside of the function and we just return always checked addresses.

What do others think?

ST-DDT: > Do we only have these two options?
> If so, maybe we could go with options.checksummed: boolean

If yes, then yes.


> But I even think about just to always return the checksummed variant as it looks like the non checked sum is just the `value.toLowerCase()`

@RareBodhi Are they two distinct things? e.g. you cannot use the checksummed version as address and vice versa?
If yes, then we should have an option for it. If no, then yes we could return the checksummed variant all the time, but I wouldn't mind having the parameter anyway.

FFR: https://github.com/validatorjs/validator.js/blob/master/src/lib/isEthereumAddress.js

package.json
RareBodhi: My bad, for some reason I saw no dependencies and a bunch in dev deps and just went for it. 

I'll keep it dep free and implement the logic itself. Though more verbose.

src/modules/finance/index.ts
RareBodhi: There are basically just those two options. 

Checksum is a specific multi case formatting. Non checksum is typically a full lowercase value but not necessarily. 

I think it would be safe to assume if checksum option is false that the string would be a lowercase hex value. 

I would also lean towards returning checksum as default and then having the non checksum be the opt in. 

What do you think to this before I go further with the pr round 2.

Shinigami92: I vote for: only return the checkedsum, no options

Let's see what the others say

ST-DDT: If checksum is the default, then I also tend to no options.

RareBodhi: Well I won't implement this until tomorrow but looks like only returning checksum could be the way forward. Will see if there is any other input in the meantime.

RareBodhi:
@ST-DDT Since we won't be planning to adding external dependencies and the keccak256 hash algorithm is not yet implemented in Node's core cryptographic module I see no choice but to close this issue, we can not accurate generate a checksummed hash without the algorithm and porting an entire cryptographic lobrary to this repo seems like a bad idea.

I suppose in the meantime we can leave it to users to handle this how they see fit with a third party dep like `web3` or `ethers` packages. 

Unforunately I think we have to close it here until node adopts keccak support.

Shinigami92:
IMO you don't need to close it as the implementation looks really easy

https://github.com/web3/web3.js/blob/316bb6ca0d8aaea56c824efb517c2db1286bba46/packages/web3-utils/src/index.js#L306-L313

RareBodhi:
> IMO you don't need to close it as the implementation looks really easy
> 
> https://github.com/web3/web3.js/blob/316bb6ca0d8aaea56c824efb517c2db1286bba46/packages/web3-utils/src/index.js#L306-L313

How do you propose getting the address keccak hash `addressHash` in that implementation with the sha3 function without an external dependency?

`const addressHash = web3.sha3(address).replace(/^0x/i, '');`

Shinigami92:
Oh I overlooked the `addressHash` variable 🤔
What happens if you just use `address[i]` instead of `addressHash[i]`?

RareBodhi:
> Oh I overlooked the `addressHash` variable 🤔 What happens if you just use `address[i]` instead of `addressHash[i]`?

Original address: `0xd20ea4eb0d1dfebbe94c6da28fc4cdadb8ef8abe`
SHA3 address hash: `8a0c0aa5d8af6542a05dcb238ff150175d43ba8a1e5c128d789c86159403a0b7`

As you can see it ends up being a completely different result once hashed. And the resulting address looks like `0x24F0FF879E698F4D8FE07D6E9A02AEDAA0FE4FD7` which is not a correct Checksum unforuntatey.

Shinigami92:
> > Oh I overlooked the `addressHash` variable 🤔 What happens if you just use `address[i]` instead of `addressHash[i]`?
> 
> 
> 
> Original address: `0xd20ea4eb0d1dfebbe94c6da28fc4cdadb8ef8abe`
> 
> SHA3 address hash: `8a0c0aa5d8af6542a05dcb238ff150175d43ba8a1e5c128d789c86159403a0b7`
> 
> 
> 
> As you can see it ends up being a completely different result once hashed. And the resulting address looks like `0x24F0FF879E698F4D8FE07D6E9A02AEDAA0FE4FD7` which is not a correct Checksum unforuntatey.

I see 🙁
I will look a bit deeper later when I'm not on a mobile screen

RareBodhi:
> I see 🙁 I will look a bit deeper later when I'm not on a mobile screen

Sure, if you find a way that would be great, but I see no good option at the moment without introducing dependencies.

Shinigami92:
Yeah, now I understand
It goes deeper and deeper
Looks like there is no way to support it really
Maybe we could enhance the docs and just say `if need a checked address, wrap the call with web3-utils`

ST-DDT:
> Looks like there is no way to support it really

Should we close this PR as "impossible" then? (See also the next part)

> Maybe we could enhance the docs and just say `if need a checked address, wrap the call with web3-utils`

IMO if you use etherium addresses you are aware of the tools required to convert it accordingly.
Maybe we could add a hint to the jsdocs that it returns a non-check-summed address.

RareBodhi:
> > Looks like there is no way to support it really
> 
> Should we close this PR as "impossible" then? (See also the next part)
> 
> > Maybe we could enhance the docs and just say `if need a checked address, wrap the call with web3-utils`
> 
> IMO if you use etherium addresses you are aware of the tools required to convert it accordingly. Maybe we could add a hint to the jsdocs that it returns a non-check-summed address.

I agree it will be obvious that it needs to be transformed. I agree this PR itself it essentially impossible without node supporting keccak or importing a dependency.

It may be worth mentioning that it is a non-checksummed address but I don't think it's a major thing. I will leave it to you to close this if you're happy with the resolution of this conversation.

ST-DDT:
We decided to just amment the jsdocs for now.
Thanks for your time anyway.

PR #1602, opened 2022-11-25
refactor(string)!: swap `allowLeadingZeros` default to `true`

### LINKED ISSUE #1592: Swap `allowLeadingZeros` default value to `true`

Often we have a call like `faker.number.int(9)` and then convert it to a string

### Suggested solution

Implement a new method `faker.string.digit()` which returns a digit between `[0, 9]` and directly returns it as `string`

### Alternative

_No response_

### Additional context

Found in
- #1122

ejcheng:
closes #1592


src/modules/helpers/index.ts
```diff
@@ -523,7 +523,7 @@ export class HelpersModule {
    * It is also possible to use multiple parameters (comma separated).
    *
    * ```js
-   * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.')
+   * const message = faker.helpers.fake('Your pin is {{string.numeric(4)}}.')
```
Shinigami92: This invalidated the wanted example
> It is also possible to use multiple parameters (comma separated).

src/modules/random/index.ts
```diff
@@ -262,15 +262,15 @@ export class RandomModule {
    *
    * @param length The number of digits to generate. Defaults to `1`.
    * @param options The options to use. Defaults to `{}`.
-   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `false`.
+   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `true`.
```
Shinigami92:
```suggestion
   * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
```

I think we should add the `or not` at the end: https://dictionary.cambridge.org/grammar/british-grammar/whether

src/modules/string/index.ts
```diff
@@ -317,15 +317,15 @@ export class StringModule {
    *
    * @param options Either the number of characters or the options to use.
    * @param options.length The number or range of digits to generate. Defaults to `1`.
-   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `false`.
+   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `true`.
```
Shinigami92:
```suggestion
   * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
```

test/string.spec.ts
```diff
@@ -427,10 +427,9 @@ describe('string', () => {
           expect(actual).toMatch(/^[1-9][0-9]+$/);
         });
 
-        it('should allow leading zeros via option', () => {
+        it('should allow leading zeros by default', () => {
           const actual = faker.string.numeric({
             length: 15,
-            allowLeadingZeros: true,
           });
```
Shinigami92: I think this should be swapped. The test was not to test the default but to test passing the option property

Shinigami92 (changes requested):

matthewmayer:
Could you include a change to `docs/guide/upgrading.md` as part of this PR for documenting this behavior change for anyone upgrading from v7 to v8?

src/modules/helpers/index.ts
```diff
@@ -537,7 +537,7 @@ export class HelpersModule {
    * faker.helpers.fake('Good Morning {{person.firstName}}!') // 'Good Morning Estelle!'
    * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
    * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
-   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1, {"allowLeadingZeros": true})}}') // 'I rolled the dice and got: 6'
+   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1)}}') // 'I rolled the dice and got: 6'
```
ST-DDT: This example is about showing the ability to pass multiple arguments:

```suggestion
   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1, {"allowLeadingZeros": false})}}') // 'I rolled the dice and got: 6'
```

src/modules/random/index.ts
```diff
@@ -262,15 +262,15 @@ export class RandomModule {
    *
    * @param length The number of digits to generate. Defaults to `1`.
    * @param options The options to use. Defaults to `{}`.
-   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `false`.
+   * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
```
ST-DDT, line 265: Should we retain the old default for the old method?
I think we should, because it is an unnecessary breaking change.
Also the upgrading docs suggest that the predecessor uses a different default:

> unlike `faker.string.numeric`'s predecessor, `faker.random.numeric`.

ST-DDT (changes requested):
Should we add a link from string.numeric -> number.int (and vice versa)?

ejcheng:
> Should we add a link from string.numeric -> number.int (and vice versa)?

What do you mean by link? A jsdoc @link tag?

ST-DDT:
> > Should we add a link from string.numeric -> number.int (and vice versa)?
> 
> What do you mean by link? A jsdoc @link tag?

I'm referring to `@see`, so the user is aware that they can get a number (within a range) or just a bunch of digits depending on their requirements.

docs/guide/upgrading.md
```diff
@@ -84,6 +84,10 @@ The `faker.address.*` methods will continue to work as an alias in v8 and v9, bu
 | `faker.address.streetPrefix`        | _Removed_                            |
 | `faker.address.streetSuffix`        | _Removed_                            |
 
+## `allowLeadingZeros` behavior change in `faker.string.numeric`
+
+The `allowLeadingZeros` boolean parameter in `faker.string.numeric` (in the new `string` module) now defaults to `true`, unlike `faker.string.numeric`'s predecessor, `faker.random.numeric`. `faker.string.numeric` will now generate numeric strings that could have leading zeros by default.
```
ST-DDT: Could you change it somewhat like this to avoid repeating the fqmn 3 times?

```suggestion
The `allowLeadingZeros` boolean parameter in `faker.string.numeric` (in the new `string` module) now defaults to `true`, unlike its predecessor, `faker.random.numeric`. `faker.string.numeric` will now generate numeric strings that could have leading zeros by default.
```

src/modules/number/index.ts
```diff
@@ -26,6 +26,8 @@ export class NumberModule {
    *
    * @throws When options define `max < min`.
    *
+   * @see faker.string.numeric if you would like to generate a string of digits
```
```suggestion
   * @see faker.string.numeric() If you would like to generate a `string` of digits with a given length (range).
```

src/modules/string/index.ts
```diff
@@ -317,15 +317,17 @@ export class StringModule {
    *
    * @param options Either the number of characters or the options to use.
    * @param options.length The number or range of digits to generate. Defaults to `1`.
-   * @param options.allowLeadingZeros If true, leading zeros will be allowed. Defaults to `false`.
+   * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
    * @param options.exclude An array of digits which should be excluded in the generated string. Defaults to `[]`.
    *
+   * @see faker.number.int() if you would like to generate a number (within a range)
```
ST-DDT:
```suggestion
   * @see faker.number.int() If you would like to generate a `number` (within a range).
```

Would make `within a value range` make it more distinct from the `length range`?

src/modules/number/index.ts
Shinigami92: Woopsy, didn't thought of that this is not my PR :facepalm: 
I will leave the rest to @import-brain

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -537,7 +537,7 @@ export class HelpersModule {
    * faker.helpers.fake('Good Morning {{person.firstName}}!') // 'Good Morning Estelle!'
    * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
    * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
-   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1, {"allowLeadingZeros": true})}}') // 'I rolled the dice and got: 6'
+   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1, {"allowLeadingZeros": false})}}') // 'I rolled the dice and got: 6'
```
matthewmayer: This is a weird example. For rolling a dice I'd expect to use number.int with min 1 and max 6. And a one-digit numeric string doesn't have any leading zeros, so the value of the allowLeadingZeros is irrelevant?

matthewmayer (changes requested):

src/modules/helpers/index.ts
Shinigami92: IMO we should maybe just use another new example.
For context: this PR changed the default to `true`, so it doesn't make anymore sense to use `, {"allowLeadingZeros": false}`

ST-DDT: IMO the example makes sense, because most dices don't have a `0` on them.

matthewmayer: most dice also don't have 7 8 and 9 on them :D

I think the main thing is to show how to pass a options object inline

How about
```
    faker.helpers.fake('Your PIN number is : {{string.numeric(4, {"exclude": ["0"]})}}') // 'Your PIN number is 4834'
```

src/modules/helpers/index.ts
```diff
@@ -537,7 +537,7 @@ export class HelpersModule {
    * faker.helpers.fake('Good Morning {{person.firstName}}!') // 'Good Morning Estelle!'
    * faker.helpers.fake('You can call me at {{phone.number(!## ### #####!)}}.') // 'You can call me at 202 555 973722.'
    * faker.helpers.fake('I flipped the coin and got: {{helpers.arrayElement(["heads", "tails"])}}') // 'I flipped the coin and got: tails'
-   * faker.helpers.fake('I rolled the dice and got: {{string.numeric(1, {"allowLeadingZeros": false})}}') // 'I rolled the dice and got: 6'
+   * faker.helpers.fake('Your PIN number is : {{string.numeric(4, {"exclude": ["0"]})}}') // 'Your PIN number is 4834'
```
```suggestion
   * faker.helpers.fake('Your PIN number is: {{string.numeric(4, {"exclude": ["0"]})}}') // 'Your PIN number is: 4834'
```

ST-DDT (approved):

matthewmayer (approved):

Shinigami92 (approved):

ST-DDT:
@import-brain Looking at the now failing PRs, it looks like we missed some tests.

PR #1611, opened 2022-11-27
docs: Display actual deprecation message in docs

### LINKED ISSUE #1607: Display actual deprecation message in api docs
I noticed if you write like `@deprecated some message` then `some message` does not appear in the online docs. Perhaps that should also show in the yellow deprecation banner for more context?

_Originally posted by @matthewmayer in https://github.com/faker-js/faker/issues/1606#issuecomment-1328061513_

matthewmayer:
fix #1607

looks like this:

Screenshot 2022-11-27 at 17 40 54

### DIFF
--- a/docs/.vitepress/components/api-docs/method.ts
+++ b/docs/.vitepress/components/api-docs/method.ts
@@ -5,8 +5,7 @@ export interface Method {
   readonly parameters: MethodParameter[];
   readonly returns: string;
   readonly examples: string; // HTML
-  readonly deprecated: boolean;
-  readonly deprecatedMessage: string;
+  readonly deprecated?: string;
   readonly since: string;
   readonly seeAlsos: string[];
 }
--- a/docs/.vitepress/components/api-docs/method.ts
+++ b/docs/.vitepress/components/api-docs/method.ts
@@ -6,6 +6,7 @@ export interface Method {
   readonly returns: string;
   readonly examples: string; // HTML
   readonly deprecated: boolean;
+  readonly deprecatedMessage: string;
   readonly since: string;
   readonly seeAlsos: string[];
 }
--- a/scripts/apidoc/signature.ts
+++ b/scripts/apidoc/signature.ts
@@ -19,12 +19,11 @@ import type {
 import vitepressConfig from '../../docs/.vitepress/config';
 import { faker } from '../../src';
 import {
-  extractDeprecatedMessage,
+  extractDeprecated,
   extractRawExamples,
   extractSeeAlsos,
   extractSince,
   formatTypescript,
-  isDeprecated,
   joinTagParts,
   pathOutputDir,
 } from './utils';
@@ -169,7 +168,7 @@ export function analyzeSignature(
 
   const prettyMethodName = prettifyMethodName(methodName);
   const code = '```';
-
+  const deprecatedMessage = extractDeprecated(signature);
   return {
     name: methodName,
     title: prettyMethodName,
@@ -178,9 +177,8 @@ export function analyzeSignature(
     since: extractSince(signature),
     returns: typeToText(signature.type),
     examples: mdToHtml(`${code}ts\n${examples}${code}`),
-    deprecated: isDeprecated(signature),
-    deprecatedMessage: mdToHtml(extractDeprecatedMessage(signature)),
     seeAlsos,
+    ...(deprecatedMessage && { deprecated: mdToHtml(deprecatedMessage) }),
   };
 }
 
--- a/scripts/apidoc/signature.ts
+++ b/scripts/apidoc/signature.ts
@@ -19,10 +19,10 @@ import type {
 import vitepressConfig from '../../docs/.vitepress/config';
 import { faker } from '../../src';
 import {
+  extractDeprecatedMessage,
   extractRawExamples,
   extractSeeAlsos,
   extractSince,
-  extractDeprecatedMessage,
   formatTypescript,
   isDeprecated,
   joinTagParts,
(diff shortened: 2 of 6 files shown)


Shinigami92:
You need to run `pnpm run format` or format the touched files with prettier

docs/.vitepress/components/api-docs/method.ts
```diff
@@ -6,6 +6,7 @@ export interface Method {
   readonly returns: string;
   readonly examples: string; // HTML
   readonly deprecated: boolean;
+  readonly deprecatedMessage: string;
```
ST-DDT: Since we only deprecate with a message you can remove the boolean.

scripts/apidoc/utils.ts
```diff
@@ -147,6 +147,19 @@ export function isDeprecated(signature: SignatureReflection): boolean {
   return extractTagContent('@deprecated', signature).length > 0;
 }
 
+/**
+ * Extracts the content of the deprecated tag
```
ST-DDT, line 151:
```suggestion
 * Extracts the content of the `@deprecated` tag.
```

scripts/apidoc/utils.ts
```diff
@@ -147,6 +147,19 @@ export function isDeprecated(signature: SignatureReflection): boolean {
...
+ *
+ * @param signature The signature to check.
```
ST-DDT:
```suggestion
 * @param signature The signature to extract it from.
```

scripts/apidoc/utils.ts
```diff
@@ -147,6 +147,19 @@ export function isDeprecated(signature: SignatureReflection): boolean {
...
+ *
+ * @returns the contents of the deprecated tag
+ */
+export function extractDeprecatedMessage(
+  signature: SignatureReflection
+): string {
+  return extractTagContent('@deprecated', signature).join().trim();
```
ST-DDT: Maybe only return a string if the method is actually deprecated?
Otherwise return undefined.

scripts/apidoc/utils.ts
```diff
@@ -147,6 +147,19 @@ export function isDeprecated(signature: SignatureReflection): boolean {
   return extractTagContent('@deprecated', signature).length > 0;
 }
 
+/**
+ * Extracts the content of the deprecated tag
+ *
+ * @param signature The signature to check.
+ *
+ * @returns the contents of the deprecated tag
```
ST-DDT: AFAICT the other methods dont have a return tag and IMO it doesnt add much value to the jsdocs in this case.

ST-DDT (changes requested):

scripts/apidoc/utils.ts
matthewmayer: It kept complaining that this method itself was deprecated when I left the @ in 😀

scripts/apidoc/utils.ts
matthewmayer: switched this to string | undefined

scripts/apidoc/signature.ts
```diff
@@ -177,8 +177,8 @@ export function analyzeSignature(
     since: extractSince(signature),
     returns: typeToText(signature.type),
     examples: mdToHtml(`${code}ts\n${examples}${code}`),
-    deprecated: isDeprecated(signature),
     seeAlsos,
+    ...(deprecatedMessage && { deprecated: mdToHtml(deprecatedMessage) }),
```
matthewmayer, line 181: this is a bit ugly but not sure if there's a nicer way of doing this in JS, basically "if deprecatedMessage is undefined, don't add anything to the dictionary, otherwise add `{deprecated: mdToHtml(deprecatedMessage) }`

ST-DDT: As an alternative you could use:

````ts
let deprecated = extractDeprecated(signature);
if (deprecated) {
  deprecated = mdToHtml(deprecated);
}
````

Or change `mdToHtml` to return instantly if the input is undefined.

File: scripts/apidoc/utils.ts [hunk 7]
```diff
@@ -147,6 +147,20 @@ export function isDeprecated(signature: SignatureReflection): boolean {
...
+ *
+ * @param signature The signature to extract it from.
+ *
+ * @returns the contents of the deprecated tag, or undefined if no tag is found
+ */
+export function extractDeprecated(
+  signature: SignatureReflection
+): string | undefined {
+  const deprecated = extractTagContent('@deprecated', signature).join().trim();
+  return deprecated.length === 0 ? undefined : deprecated;
```
ST-DDT, line 161:
```suggestion
  const deprecated = extractTagContent('@deprecated', signature);
  return deprecated.length === 0 ? undefined : deprecated.join().trim();
```

test/scripts/apidoc/__snapshots__/signature.spec.ts.snap
```diff
@@ -98,7 +96,6 @@ exports[`signature > analyzeSignature() > literalUnionParamMethod 1`] = `
 
 exports[`signature > analyzeSignature() > methodWithDeprecated 1`] = `
 {
-  "deprecated": true,
```
ST-DDT, line 101: This method should be/stay marked as deprecated.

ST-DDT (changes requested):

test/scripts/apidoc/__snapshots__/signature.spec.ts.snap
matthewmayer: hmm the problem is this method is marked as deprecated in the JSDoc but with an empty message (which is impossible in reality because that will fail the test that all @deprecated markers need a message).

ST-DDT: It is impossible, then please fix/merge the example?

matthewmayer: having thought about this a bit more, I think it might actually be better to revert back to boolean and (optional string) for deprecated. I think it makes it clearer the difference between a @deprecated with and without message

So in Method.ts you would have

deprecated:Boolean
deprecatedMessage:string | undefined 

(undefined if deprecated is false, empty string if deprecated is true but message is empty)

matthewmayer:
going to rethink this

PR #1622, opened 2022-12-01
docs: improving readability of UUID string

### LINKED ISSUE #1466: Improve readability of `string.uuid`
Extracted from https://github.com/faker-js/faker/pull/1155#discussion_r967532808

In the string module's uuid method we have a code snippet using bit operations

```ts
const value = placeholder === 'x' ? random : (random & 0x3) | 0x8;
```

We want to get rid of these as they are hard to understand
This might result in a deterministic generated value change, but that's okay

Kcops11:
Updated comment to be more closely in line with semantic style guide.

#1466


ST-DDT (changes requested):
Sorry for the late response.
We looked over this, and sadly we feel that this is not what the issue is about.

The issue is about the actual implementation of the method using bit operations as described here:
https://github.com/faker-js/faker/issues/1466#issue-1412189473
In [this comment](https://github.com/faker-js/faker/issues/1466#issuecomment-1281505929) we opted for adding a comment to the implementation explaining what the nit magic actually does, so that a future reader knows exactly what happens.

As for the additions to the jsdocs.
Most of the added docs don't add relevant information for the user.
Except for the first sentence that it "is randomly generated".

Please revert the other documentation additions and add a comment to the bit operations in the implementation (don't refer to bits, only values/value ranges).

Kcops11:
Okay, I will make changes and get back to you.

Kcops11:
I have updated the code with a comment and removed most of the doc comment I previously added. Please let me know if you would like me to make any additional changes or if my changes do not reflect what you are looking for.

src/modules/string/index.ts
```diff
@@ -437,6 +439,7 @@ export class StringModule {
     const RFC4122_TEMPLATE = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
     const replacePlaceholders = (placeholder: string) => {
       const random = this.faker.number.int(15);
+      //x value is set to any random hex and y value is set to random within {8, 9, A, B}
```
ST-DDT:
```suggestion
      // 'x's are set to any random hex and 'y's are set to a random value within {8, 9, a, b}
```

src/modules/string/index.ts
```diff
@@ -428,6 +428,8 @@ export class StringModule {
   /**
    * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
    *
+   * A version 4 UUID is randomly generated.
```
Shinigami92: I think this is obvious and not needed.

The PR should aim to just add a comment to the body of the function, not the JSDoc itself

src/modules/string/index.ts
```diff
@@ -437,6 +439,7 @@ export class StringModule {
     const RFC4122_TEMPLATE = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
     const replacePlaceholders = (placeholder: string) => {
       const random = this.faker.number.int(15);
+      // 'x's are set to any random hex and 'y's are set to a random value within {8, 9, a, b}
```
Shinigami92: Somehow I find this not explicit enough
Maybe something like

```
// `x`'s will be replaced with random hex characters and the `y` will be replaced with `8`, `9`, `a` or `b`
```

but additionally maybe we need also a **how** and **why** :thinking: 
especially the bit-and and bit-or should be explained

src/modules/string/index.ts
Kcops11: Okay I will remove it

src/modules/string/index.ts
Kcops11: I was told to avoid talking about the bits, but if you would like me to add something to that effect I can do so.

src/modules/string/index.ts
ST-DDT: IMO that adds useful information.

src/modules/string/index.ts
ST-DDT: I'm a bit conflicted here @Shinigami92.
On one hand you dont want to mention that the uuid v4 is random in the description, but then here you want a lot of details.
IMO the why is perfectly clear aka the RFC that is mentioned three lines above this line requires a special character/subset for that place.

Why: Because the RFC_xyz states so.
Where: At the location the rfc specifies.
What: Thats what the comment needs to clarify and does IMO.
How: Using the string replace that you are in and refering to in the comment

IMO we can expect the user/dev to actually read the few lines sourounding that method, because if they dont there is no need to write a comment in the first place.

I'm fine with the rephrasing.

ST-DDT (approved):

src/modules/string/index.ts
Shinigami92: @ST-DDT @Kcops11 Okay let me explain again from the start
The initial issue https://github.com/faker-js/faker/issues/1466#issue-1412189473 was about the bit operations

image

After a meeting, we came to the conclusion that we actually do not want to get rid of it by replacing it with e.g. code we already use for finance iban, like:
```ts
    return [
      this.faker.string.numeric(8),
      this.faker.string.numeric(4),
      `4${this.faker.string.numeric(3)}`,
      `${this.faker.helpers.arrayElement([
        '8',
        '9',
        'a',
        'b',
      ])}${this.faker.string.numeric(3)}`,
      this.faker.string.numeric(12),
    ].join('-')
```
but leave the bit operator as is

But now for other contributors we need a developer to developer comment to explain what's going on and why the bit operations are better (or whatever why it is there)

This comment has nothing to do with a JSDoc facing to users of faker at all

src/modules/string/index.ts
Shinigami92: 1. Everything in faker is randomly generated
2. `Returns a UUID v4` and the next sentence is `A version 4 UUID [...]`
   IMO this is redundant 🤷

ST-DDT: > Everything in faker is randomly generated

The comment refers not to a Faker property, but a property of the uuid v4.
Some uuids are generated time based, or system based. This method returns random generated Uuids.
Sure, I could get that information from visiting the link, but that addition sums up the relevant information unless you really want to know in depth about uuids.

> Returns a UUID v4 and the next sentence is A version 4 UUID [...]
IMO this is redundant 🤷

Sure, do you have a suggestion on how to rephrase it/merge it into a single sentence.
Maybe `UUID v4 (randomly generated; Link)`?

Shinigami92: Oh the point that I mis-interpreted the intent of the JSDoc description shows that it is phrased wrongly.
So yeah, revert or rephrase it.
For now I will let @Kcops11 think about it to come up with an idea.

src/modules/string/index.ts
ST-DDT: > to explain what's going on

IMO that it perfectly clear from the comment+context

>  and why the bit operations are better (or whatever why it is there)

They aren't. That's just how it is implemented. The comment explains what it does.
The difference is only on level of personal preference (which I don't have in this case).

- The performance difference is negligible
  (probably a really minuscule penalty, because it is an additional operation even if extremely fast).
- but the bits correlate stronger to the specification, than number ranges would
- but not everyone can read the bits, so we need a comment

We could change the impl to this without any issues:

````ts
const replacePlaceholders = (placeholder: string) => placeholder === 'x' 
	? this.faker.number.hex()
	: this.faker.number.hex({ min: 8, max: 11 });
````

We could even do it like this:

````ts
const replacePlaceholders = (placeholder: string) => placeholder === 'y' 
	? this.faker.number.hex({ min: 8, max: 11 })
	: this.faker.string.hexadecimal({ length: placeholder.length, casing: 'lower', prefix: '' });
return RFC4122_TEMPLATE.replace(/x+|y/g, replacePlaceholders);
````

Or even:

````ts
const replacePlaceholders = (placeholder: string) => placeholder === 'y' 
	? this.faker.number.hex({ min: 8, max: 11 })
	: this.faker.number.hex((16 ** placeholder.length) - 1);
return RFC4122_TEMPLATE.replace(/x+|y/g, replacePlaceholders);
````

(neither tested)

FYI: `y` contains the UUID variant (not version).

Comment (Shinigami92) [reply src/modules/string/index.ts]: All these suggestions are more welcome from my side, to replace this bit operations
Then after that, we can even toggle on https://eslint.org/docs/latest/rules/no-bitwise

Comment (ST-DDT) [reply src/modules/string/index.ts]: I think the first of [my suggestions](https://github.com/faker-js/faker/pull/1622#discussion_r1043144108) is the easiest to read. (Then we also dont need the code comment anymore) . @faker-js/members any preferences?

**Comment by Kcops11** [comment]
Hello @ST-DDT and @Shinigami92. I coming back to this after finishing my finals. Has a consensus been reached between you regarding the direction this should take?

It would seem as though this is it: https://github.com/faker-js/faker/pull/1622#discussion_r1043213403. Let me know!

**Comment by xDivisionByZerox** [comment]
> Hello @ST-DDT and @Shinigami92. I coming back to this after finishing my finals. Has a consensus been reached between you regarding the direction this should take?
> 
> It would seem as though this is it: [#1622 (comment)](https://github.com/faker-js/faker/pull/1622#discussion_r1043213403). Let me know!

While I'm neither of them I think [the approach](https://github.com/faker-js/faker/pull/1622#discussion_r1043213403) from @ST-DDT is our best shot here.
- It improves readability (the main goal of this PR)
- has negligible performance influence
- allows us to activate the [no-bitwise](https://eslint.org/docs/latest/rules/no-bitwise) eslint rule in error mode

File: src/modules/string/index.ts [hunk 4]
```diff
@@ -437,6 +439,7 @@ export class StringModule {
     const RFC4122_TEMPLATE = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
     const replacePlaceholders = (placeholder: string) => {
```
ST-DDT: Please use this instead:

````ts
const replacePlaceholders = (placeholder: string) => placeholder === 'x' 
	? this.faker.number.hex()
	: this.faker.number.hex({ min: 8, max: 11 });
````

**Review by ST-DDT — CHANGES_REQUESTED** [review]

File: src/modules/string/index.ts (hunk 4 above)
Comment (Kcops11) [reply src/modules/string/index.ts]: All right, I will get started on this soon!

Comment (Kcops11) [reply src/modules/string/index.ts]: Forgive my ignorance but if I make the first suggested change, would this need additional testing? It seems to be implied that it would not.

Comment (ST-DDT) [reply src/modules/string/index.ts]: There are already tests in place that ensure that the method works as expected.
The suggested change should still fulfill the tests and shouldnt change the tests.

Comment (Kcops11) [reply src/modules/string/index.ts]: With this change, the comment I added previously is also removed, correct?

Comment (Kcops11) [reply src/modules/string/index.ts]: Further question I would leave the return statement in correct?

This is what I am thinking for the change:

image

Comment (ST-DDT) [reply src/modules/string/index.ts]: Looks good to me.

**Comment by Kcops11** [comment]
Okay, I have made changes to the method, taking into consideration all of the feedback from the various development team members that have commented here. 

Please let me know if anything needs to be put in the right place or if any additional changes need to be made.

File: src/modules/string/index.ts [hunk 5]
```diff
@@ -516,18 +516,18 @@ export class StringModule {
   /**
    * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
    *
+   * UUID v4 (randomly generated; https://www.rfc-editor.org/rfc/rfc4122#section-3)
```
ST-DDT: This looks too much like the first line, maybe omit it for now and if we have a better idea, then we can add it later.

Kcops11: Okay, ill remove it!

src/modules/string/index.ts
```diff
@@ -518,18 +518,18 @@ export class StringModule {
   /**
    * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
    *
+   * 
+   *
```
ST-DDT: Please remove the redundant empty lines here (keep only one).

Kcops11: Sure can do, ill take care of that now

ST-DDT (dismissed):

ST-DDT:
@Kcops11 Is this PR ready for review? (Because currently it is marked as draft)
IMO it is.

Shinigami92 (commented):
There are lint and snapshot errors

Kcops11:
> @Kcops11 Is this PR ready for review? (Because currently it is marked as draft) IMO it is.

It should be after I try fixing these errors

Kcops11:
I have attempted to fix some of the formatting errors pointed out by Prettier. Ill try fixing any additional errors after I have successfully handled the formatting.

xDivisionByZerox:
> I have attempted to fix some of the formatting errors pointed out by Prettier. Ill try fixing any additional errors after I have successfully handled the formatting.

It seems like you are fixing those errors on your own...
To fix formatting errors simply run `npm run format`
To fix the failing snapshot test simply run `npx vitest -u` (`-u` updates the snapshots)

ST-DDT (dismissed):

matthewmayer (changes requested):
I'm not sure this is that much readable than the original version.

I think a comment would be helpful, maybe:

```
Each x in the template is replaced with a random hex digit from `0` to `f`, and y is replaced with a random hex digit from the range `{8, 9, a, b}` only.
```

Also could consider
- Writing "11" as 0xb instead
- splitting replacePlaceholders into two seperate functions, one for replacing the x's and one for replacing the y's?

ST-DDT:
@matthewmayer Like this? (Prettier gives me a hard time to format this in a readable way)

````ts
  uuid(): string {
    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'
      .replace(/x/g, () => this.faker.number.hex({ min: 0x0, max: 0xf }))
      .replace(/y/g, () => this.faker.number.hex({ min: 0x8, max: 0xb }));
  }
````

I'm not sure about using the hex notation for `min`.

matthewmayer:
Yeah that seems much more readable to me.

matthewmayer:
Sometimes I think writing more code makes it easier to read. 

Like, adding the explicit min and max for the x replacement isn't necessary but makes it easier to understand, because you don't need to go check what the default parameter values are for hex() to understand the line of code. And it also makes it easier to see the difference between that line and the line below.

ST-DDT (approved):

matthewmayer (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1623, opened 2022-12-01
docs: how to use with testing frameworks

### LINKED ISSUE #1324: Add a documentation page for how to use faker with testing frameworks
We would like to have a documentation page that contains examples how to use faker in combination with testing frameworks like vitest and Cypress

brenno263:
Fixes #1324 

Add a documentation page about integrating with testing frameworks.

I sort of wish I had more to add besides examples, but Faker is pretty straightforward to use. There were a number of quirks to document with frameworks, but at the time of writing everything significant has been smoothed over to my knowledge.

Please let me know if there's anything I should add.


ST-DDT:
Please fix the merge conflicts.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+
+Faker can easily be used with a variety of testing frameworks. Here are a few examples with popular frameworks, as well as some edge cases you should be aware of.
+
+## Vitest
+
+Faker works about exactly as you would expect with Vitest. Here's a minimal example:
+
+```ts
+import { assert, test } from 'vitest';
+import { faker } from '@faker-js/faker/locale/en';
```
ST-DDT, line 17: Could you please add a small hint at the top you why use locale/en as an import?
Maybe include a link to the localization page as well.

e.g.

> These examples only use the `en` variants for better performance. If you want to know more visit `localization`.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+
+test('reverse array', () => {
```
ST-DDT: In this repo we mainly use describe/it, should we use it here as well?
Is there a recommendation what to use in the vitest tests?

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+  let title = faker.name.jobTitle();
+  let name = faker.name.fullName();
+  let animal = faker.animal.bear();
+
+  let array = [title, name, animal];
+
+  assert.deepEqual(array.reverse(), [animal, name, title]);
```
ST-DDT: Same here. We mainly use `except`.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+import { faker } from '@faker-js/faker/locale/en';
+
+describe('example tests', () => {
+  test('reverse array', () => {
+    let title = faker.name.jobTitle();
+    let name = faker.name.fullName();
+    let animal = faker.animal.bear();
+
+    let array = [title, name, animal];
+    console.log('1', array);
```
ST-DDT: I think this line is obsolete.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+    cy.get('#password-input').type(password);
+    cy.get('#password-confirm-input').type(password);
+
+    cy.get('#register-submit-input').click();
+
+    //now we try to login with these credentials.
+    cy.visit('https://www.example.com/login');
+
+    cy.get('#email-input').type(email);
+    cy.get('#password-input').type(password);
```
ST-DDT, line 92: Is the `login` click missing here?

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+    console.log('1', array);
+
+    expect(array.reverse()).toMatchObject([animal, name, title]);
+  });
+});
+```
+
+## Cypress
+
+Cypress works too:
```
ST-DDT:
```suggestion
[Cypress](https://www.cypress.io/) works too:
```

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
+# Frameworks
+
+Faker can easily be used with a variety of testing frameworks. Here are a few examples with popular frameworks, as well as some edge cases you should be aware of.
+
+## Vitest
+
+Faker works about exactly as you would expect with Vitest. Here's a minimal example:
```
ST-DDT:
```suggestion
Faker works about exactly as you would expect with [Vitest](https://vitest.dev/). Here's a minimal example:
```

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
...
+
+  let array = [title, name, animal];
+
+  assert.deepEqual(array.reverse(), [animal, name, title]);
+});
+```
+
+## Jest
+
+Jest integrates similarly:
```
ST-DDT:
```suggestion
[Jest](https://jestjs.io/) integrates similarly:
```

docs/guide/frameworks.md
```diff
@@ -0,0 +1,78 @@
+# Frameworks
+
+Faker can easily be used with a variety of testing frameworks. Here are a few examples with popular frameworks, as well as some edge cases you should be aware of.
```
ST-DDT: Which edge cases?

ST-DDT (commented):
Could you also add hints regarding faker.seed and test snapshots?

docs/guide/frameworks.md
brenno263: Vitest's docs are pretty unopinionated. I looked into describe/it though and I do like the descriptiveness, so I went ahead with describe/it.

I didn't know this syntax was available with Vite, just Jest, so now the two snippets are identical. I left a note about that and left it as is, would you like them combined into a 'Vitest & Jest' section?

docs/guide/frameworks.md
```diff
@@ -0,0 +1,111 @@
...
+
+    const array = [title, name, animal];
+
+    expect(array.reverse()).toStrictEqual([animal, name, title]);
+
+    // Expect our value to always match a generated snapshot.
+    expect(array.reverse()).toMatchSnapshot();
+
+    // We might want other tests to *not* be seeded. This will reset the seed to something less predictable.
+    faker.seed();
```
Shinigami92: It is bad practice to set something like this at the end of a test, because if something within the test before failed, this code will never be reached and then the seed it still set to what it was in line at the start.

Instead use e.g. `beforeEach` or `afterEach` to clean stuff up.

Shinigami92 (changes requested):

docs/guide/frameworks.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Cypress
+
+[Cypress](https://www.cypress.io/) integration is fairly straighforward as well:
+
+```ts
+import { faker } from '@faker-js/faker/locale/en';
+
+describe('Testing the application', function () {
+  it('should create an account with username and password', function () {
```
ST-DDT: Is it possible to use lambda expressions here as well? e.g. `() => {}`

docs/guide/frameworks.md
```diff
@@ -0,0 +1,111 @@
...
+
+    // We might want other tests to *not* be seeded. This will reset the seed to something less predictable.
+    faker.seed();
+  });
+});
+```
+
+## Jest
+
+Since it uses similar notation, [Jest's](https://jestjs.io/) integration looks exactly the same:
```
ST-DDT: @faker-js/maintainers , @faker-js/members Should this be phrased to be independent of vitest?

docs/guide/frameworks.md
```diff
@@ -0,0 +1,111 @@
...
+  });
+});
+```
+
+## Cypress
+
+[Cypress](https://www.cypress.io/) integration is fairly straighforward as well:
+
+```ts
+import { faker } from '@faker-js/faker/locale/en';
```
ST-DDT, line 70: Please add the Cypress imports (if needed).

docs/guide/frameworks.md
```diff
@@ -0,0 +1,111 @@
...
+  });
+});
+```
+
+## Jest
+
+Since it uses similar notation, [Jest's](https://jestjs.io/) integration looks exactly the same:
+
+```ts
+import { faker } from '@faker-js/faker/locale/en';
```
ST-DDT: Please add the Jest imports (if needed).

ST-DDT (changes requested):

docs/guide/frameworks.md
Shinigami92: Yes, good catch!

docs/guide/frameworks.md
Shinigami92: Maybe 🤔 

But I think it is a good thing to mention jest explicitly beside vitest as
1. Frameworks like NestJS still depend on jest
2. When searching e.g. via Google, we have one more search-engine hit via combination of jest+faker

docs/guide/frameworks.md
Shinigami92: This is actually correct in most cases
Cypress is normally registered globally within the configured e2e | integration | cypress folder

docs/guide/frameworks.md
Shinigami92: I'm not so sure about that. I think the default for jest is that it is configured globally as well as cypress

Vitest's default is the other way around and you need to set https://vitest.dev/config/#globals explicitly

docs/guide/frameworks.md
ST-DDT: Yes, I agree with mentioning it explicitly, I was referring to 

`Since it uses similar notation Jest integration looks exactly the same:`

Which is a back reference to a different test framework (section).

brenno263: It could be good to phrase seeding and snapshots in the context of Jest as well. Since these features are nearly identical for both platforms, I think I'll combine them into a single section to reuse the code blocks.

It will still mention both frameworks of course, preserving those sweet search-engine hits.

Shinigami92 (dismissed):

docs/guide/frameworks.md
```diff
@@ -0,0 +1,98 @@
...
+
+    const array = [title, name, animal];
+
+    expect(array.reverse()).toStrictEqual([animal, name, title]);
+  });
+});
+```
+
+It can sometimes be useful to do seeded tests, where faker will generate the same random value each time.
+These are especially useful in tests that are meant to be deterministic, such as tests.
```
ST-DDT:
```suggestion
These are especially useful in tests that are meant to be deterministic, such as tests.
```

IMO there is something missing at the end. I don't remember the name of the tests that require the exact same values/result. Reproducible tests? Gui Tests?

docs/guide/frameworks.md
```diff
@@ -0,0 +1,98 @@
+# Frameworks
+
+Faker can easily be used with a variety of testing frameworks. Here are a few examples with popular frameworks.
+
+Note that these examples use only the `en` locale for better performance. For more information visit [localization](./localization.md).
+
+## Vitest and Jest
+
+Since [Vitest](https://vitest.dev/) and [Jest](https://jestjs.io/) use an extremely similar notation, this section will cover both at once. All code snippets have been tested in both Vitest and Jest. The main difference is that testing methods need to be imported in Vitest. Simply crop that line out for a Jest integration.
```
ST-DDT: Please put each sentence in a new line.

> All code snippets have been tested in both Vitest and Jest.

IMO This doesn't have to be mentioned.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,98 @@
...
+    const animal = faker.animal.bear();
+
+    const array = [title, name, animal];
+
+    expect(array.reverse()).toStrictEqual([animal, name, title]);
+  });
+});
+```
+
+It can sometimes be useful to do seeded tests, where faker will generate the same random value each time.
```
ST-DDT: Maybe add a hint that this is only true within the same version of Faker.

docs/guide/frameworks.md
brenno263: You're right. I had given 'snapshot tests' as an example, before moving the link to the next line. Fixed.

docs/guide/frameworks.md
```diff
@@ -0,0 +1,100 @@
...
+```ts
+import { faker } from '@faker-js/faker/locale/en';
+
+describe('Testing the application', () => {
+  it('should create an account with username and password', () => {
+    let username = faker.internet.userName();
+    let password = faker.internet.password();
+    let email = faker.internet.exampleEmail();
+
+    // visit the a webpage and create an account.
```
ST-DDT:
```suggestion
    // Visit the a webpage and create an account.
```

docs/guide/frameworks.md
```diff
@@ -0,0 +1,100 @@
...
+    cy.visit('https://www.example.com/register');
+
+    cy.get('#email-input').type(email);
+    cy.get('#username-input').type(username);
+    cy.get('#password-input').type(password);
+    cy.get('#password-confirm-input').type(password);
+
+    cy.get('#register-submit-input').click();
+
+    //now we try to login with these credentials.
```
ST-DDT:
```suggestion
    // Now, we try to login with these credentials.
```

docs/guide/frameworks.md
```diff
@@ -0,0 +1,100 @@
...
+    cy.visit('https://www.example.com/login');
+
+    cy.get('#email-input').type(email);
+    cy.get('#password-input').type(password);
+
+    cy.get('#login-submit-input').click();
+
+    //we should have logged in successfully to the dashboard page.
```
ST-DDT:
```suggestion
    // We should have logged in successfully to the dashboard page.
```

ST-DDT (commented):
Just some minor nit-picking, other than that this looks good to me.

ST-DDT (approved):

Shinigami92 (approved):

brenno263:
It was a pleasure working with you guys!

PR #1642, opened 2022-12-07
feat(number)!: change float default params

### LINKED ISSUE #1597: Change `number.float` default value from `precision` to `max`
We want to change the default value of `number.float` to `{ min: 0.0, max: 1.0, precision: 0.01 }` (or at least by default two digits after comma precision) (like is right now)

The default parameter just number parameter should change the `max` value, not the `precision`

This should not affect legacy `datatype.float`!

Waiting on:
- #1122

Shinigami92:
closes #1597

### DIFF
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -62,7 +62,7 @@ export class NumberModule {
   /**
    * Returns a single random floating-point number for a given precision or range and precision.
    *
-   * @param options Upper bound or options object. Defaults to `1.0`.
+   * @param options Upper bound or options object. Defaults to `{}`.
    * @param options.min Lower bound for generated number. Defaults to `0.0`.
    * @param options.max Upper bound for generated number. Defaults to `1.0`.
    * @param options.precision Precision of the generated number. Defaults to `0.01`.
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -68,12 +68,12 @@ export class NumberModule {
    * @param options.precision Precision of the generated number. Defaults to `0.01`.
    *
    * @example
-   * faker.number.float() // 51696.36
-   * faker.number.float(3) // 52023.2
-   * faker.number.float({ min: 1000000 }) // 212859.76
-   * faker.number.float({ max: 100 }) // 28.11
-   * faker.number.float({ precision: 0.1 }) // 84055.3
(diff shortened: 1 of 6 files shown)


test/number.spec.ts
```diff
@@ -143,13 +143,14 @@ describe('number', () => {
 
     describe('float', () => {
       it('should return a random float with a default precision of 2 digits after floating point', () => {
-        const number = faker.number.float();
-        expect(number).toBe(Number(number.toFixed(2)));
+        const float = faker.number.float();
```
ST-DDT: In a lot of other test we use `actual` maybe use that here instead?

ST-DDT (dismissed):

ejcheng (dismissed):

src/modules/number/index.ts
```diff
@@ -62,18 +62,18 @@ export class NumberModule {
   /**
    * Returns a single random floating-point number for a given precision or range and precision.
    *
-   * @param options Precision or options object. Defaults to `{}`.
-   * @param options.min Lower bound for generated number. Defaults to `0`.
-   * @param options.max Upper bound for generated number. Defaults to `99999`.
+   * @param options Upper bound or options object. Defaults to `1.0`.
```
xDivisionByZerox: The default value is wrong here. It still is `{}`.

```suggestion
   * @param options Upper bound or options object. Defaults to `{}`.
```

test/__snapshots__/number.spec.ts.snap
```diff
@@ -52,13 +52,13 @@ exports[`number > 1211 > bigInt > with string value 1`] = `24n`;
 
 exports[`number > 1211 > float > with max 1`] = `64.07`;
```
xDivisionByZerox, line 53: This doesn't look right :/
A `float(1)` or `float({ max: 1 })` should not return `64.07`. Or am I reading this wrong?

xDivisionByZerox: I'm stupid. the `max 1` means it is the 1st iteration with `max`. 🤦

Shinigami92: I exactly confused my self as well with this ^^'
It's not 1 as input, but the index 1 snapshot of vitest

xDivisionByZerox (approved):

ST-DDT (approved):

ejcheng (approved):

Issue #1590, opened 2022-11-22
Deprecate datatype module entirely and move all methods into other modules


Shinigami92:
We are adding the number module in #1122 and also #1341

- #1122
- #1341

Now we have some leftover methods in the datatype module but IMO it is not worth it to have a whole module for just these few methods.
I hope we can move these into other modules so we can get rid entirely of the datatype module.

ST-DDT:
How about moving some stuff out of `helpers` as well?

- `faker.object` or `faker.record` (NEW)
  - `sample` (basically `faker.datatype.json()` without jsonifing the output)  <-- Not sure about that
  - `sampleJson` (`datatype.json` or maybe just drop it?)  <-- Not sure about that
  - `keyOf` (`helpers.objectKey`)
  - `valueOf` (`helpers.objectValue`)
- `faker.array` (NEW)
  - `sample` (`datatype.array`)
  - `elementOf` (`helpers.arrayElement`)
  - `elementsOf` (`helpers.arrayElements`)
  - `of` (`helpers.multiple`)
- `faker.date`
  - `sample` (`datatype.datetime`)
- `faker.boolean` (NEW)
  - `sample` (datatype.boolean) <-- Not sure about that

Maybe instead of `___Of` we could use `___From`.

xDivisionByZerox:
> * `faker.object` or `faker.record` (NEW)
> [...]
> * `faker.array` (NEW)

An `Array` is nothing more than a specific `Object` in JS, so no to `faker.array`. Furthermore, all modules (with `helpers` as exception) should be able to generate a value from a specific function without requiring any input. The proposed methods `keyOf ` and `valueOf` always require an object as input. Thats why they ended up in `helpers` in the first place.

> - `faker.boolean`

I see why you propose this. But I have to say I really dislike having a module only for it to host one method. But to be fair, `boolean` is the one I'm unable to find a reasonable location as well.

ST-DDT:
Currently, only two methods are left:

- [ ] [boolean](https://next.fakerjs.dev/api/datatype.html#boolean)
- [ ] [datetime](https://next.fakerjs.dev/api/datatype.html#datetime)

---

`datetime`: We can easily move it to the `date` module.

---

`boolean`: We could:

1. move it to helpers
   - ➖ Hard to find from code
2. create a tiny boolean module
   - ➕ Easier to find from code
   - ➖ Only a single method for the entire module
3. keep the tiny datatype module
   - ➖ Only a single method for the entire module
[...]

PR #1643, opened 2022-12-09
feat(number)!: change int default params

### LINKED ISSUE #1598: Change `number.int` default `max` to `Number.MAX_SAFE_INTEGER`
We want that the `number.int` `max` value is by default `Number.MAX_SAFE_INTEGER`

_We are open for other proposed min or max values_

This wont affect legacy `datatype.number`!

Waiting on:
- #1122

Shinigami92:
closes #1598


src/modules/datatype/index.ts
```diff
@@ -277,7 +277,7 @@ export class DatatypeModule {
     properties.forEach((prop) => {
       returnObject[prop] = this.boolean()
         ? this.faker.string.sample()
-        : this.faker.number.int();
+        : this.faker.number.int(99999);
```
ST-DDT: Why do you explicitly set a new max here and elsewhere?

src/modules/image/index.ts
```diff
@@ -149,7 +149,7 @@ export class ImageModule {
 
     return `https://loremflickr.com/${width}/${height}${
       category != null ? `/${category}` : ''
-    }?lock=${this.faker.number.int()}`;
+    }?lock=${this.faker.number.int(99999)}`;
```
ST-DDT: Have you checked that this is the upper limit?

src/modules/datatype/index.ts
Shinigami92: Because this is the old datatype module and I didn't want to bubble up the change to methods that just call this new internally
I would also prefer to deprecate the whole datatype module anyways in v8.0
- #1590

src/modules/image/index.ts
Shinigami92: No, I just restored the old default max and omitted to update snapshots that are not related to this PR at all

src/modules/datatype/index.ts
ST-DDT: IMO we should just use the default here.

src/modules/image/index.ts
ST-DDT: IMO we should just use the default here .

src/modules/datatype/index.ts
Shinigami92: Maybe do this in separate PR :thinking: 
What do @faker-js/members think?

ST-DDT: > Maybe do this in separate PR 🤔

No, because the method returns any string or int/number, the user should not assume there to be a certain limit and we maintainers shouldn't wonder why there is an arbitrary limit (that we just removed from the `int` method).

src/modules/image/index.ts
xDivisionByZerox: > Have you checked that this is the upper limit?

I just checked, and there is no "upper limit":
 - [Link with `lock` = Number.MAX_SAFE_INTEGER](https://loremflickr.com/320/320?lock=9007199254740991)
 - [Link with `lock` = Number.MAX_SAFE_INTEGER + 1](https://loremflickr.com/320/320?lock=9007199254740992)

Because of that, I would remove the explicit "old" value here.

src/modules/datatype/index.ts
```diff
@@ -302,7 +302,9 @@ export class DatatypeModule {
   ): Array<string | number> {
     return this.faker.helpers.multiple(
       () =>
-        this.boolean() ? this.faker.string.sample() : this.faker.number.int(),
+        this.boolean()
+          ? this.faker.string.sample()
+          : this.faker.number.int(99999),
```
ST-DDT: Please use the default here as well.

Shinigami92:
Want to leave this comment that it's a bit against my personal decision doing these last changes, but as I don't even use these methods myself, I don't care that much :shrug:

ST-DDT (approved):

ejcheng (approved):

Issue #1586, opened 2022-11-21
Provide interfaces for options with JSDocs


Shinigami92:
### Clear and concise description of the problem

As a user I want to have JSDoc tooltips when hovering over option properties in my IDE so I can directly see what the parameter is aiming for.

### Suggested solution

Provide interfaces for e.g. ranges.
Maybe even extract types for every ModuleMethodOption so I as a user can see e.g. default values for each specific property.

### Additional context

https://github.com/faker-js/faker/pull/1486#discussion_r1028049577

---

Automatisating the VitePress generated docs is a non-goal for the first PR. Everything that broke after this change for the docs need to be handled in a separate PR. That is so we can focus on IDE UX support in the first place.

ST-DDT:
Just repeating this here so it doesn't get lost: 

> Please note, that I will only accept it for the entire project, if there is a test that ensures that the text is the same as the param's and optionally the respective parts are generated automatically (When using inline types).

---

I think one of the reasons, why we don't use named types, is that IDEs don't show their actual content and you have to click on it to get all properties. (It may potentially also requires some extra work to work with typedoc's `@param` detection/filtering, not a blocker though). So we should use these named types for only a few select params/options i.e. `NumberOrRange`.

![grafik](https://user-images.githubusercontent.com/1579362/203118047-0ae6965b-29cf-4c46-b785-b9b0f93a3196.png)

ST-DDT:
**Team decision**

We discussed adding jsdocs to each option param directly in addition to the `@param` ones (basically a copy of that)

````ts
int({
  /**
   * The min value.
   */
  min: number,
  /**
   * The max value.
   */
  max: number
})
````

We decided not to use explicit types (e.g. AlphaOptions), because

- some methods have and require different types+descriptions for their attributes (exclude digits vs alpha chars)
- `@param` will no longer be checked
- It reduces the readability of the source code

xDivisionByZerox:
I will close this since this issue has been solved by #1644 as far as I can see. If this is not correct please reopen this issue.

PR #1644, opened 2022-12-09
docs: jsdoc for options

Shinigami92:
ref #1586

### DIFF
--- a/src/modules/color/index.ts
+++ b/src/modules/color/index.ts
@@ -249,9 +249,29 @@ export class ColorModule {
    * @since 7.0.0
    */
   rgb(options?: {
+    /**
+     * Prefix of the generated hex color. Only applied when 'hex' format is used.
+     *
+     * @default '0x'
+     */
     prefix?: string;
+    /**
+     * Letter type case of the generated hex color. Only applied when `'hex'` format is used.
+     *
+     * @default 'mixed'
+     */
     casing?: Casing;
+    /**
+     * Format of generated RGB color.
+     *
+     * @default 'hex'
+     */
     format?: 'hex' | StringColorFormat;
+    /**
+     * Adds an alpha value to the color (RGBA).
+     *
+     * @default false
+     */
     includeAlpha?: boolean;
   }): string;
   /**
@@ -269,7 +289,17 @@ export class ColorModule {
    * @since 7.0.0
    */
   rgb(options?: {
+    /**
+     * Format of generated RGB color.
+     *
+     * @default 'hex'
+     */
     format?: NumberColorFormat;
+    /**
+     * Adds an alpha value to the color (RGBA).
+     *
+     * @default false
+     */
     includeAlpha?: boolean;
   }): number[];
   /**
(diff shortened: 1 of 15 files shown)


Shinigami92:
@ST-DDT @xDivisionByZerox Before I move on copying over all the stuff from one place to another: is this going in the right direction, or do we need to re-phrase something?
Also did I missed something?
This is for now an example for the color module (the animal module didn't had options)

xDivisionByZerox:
> @ST-DDT @xDivisionByZerox Before I move on copying over all the stuff from one place to another: is this going in the right direction, or do we need to re-phrase something? Also did I missed something? This is for now an example for the color module (the animal module didn't had options)

Yes, that would be the approach we discussed in yesterday's meeting. Altho, I would likely write a script that does the copy and past, so we could add it to our toolchain (pre-commit or something).

Shinigami92:
> Altho, I would likely write a script that does the copy and past, so we could add it to our toolchain (pre-commit or something).

@xDivisionByZerox While copy and pasting these lines, I came to the exact same conclusion and would prefer these absolutely
The only sad thing about this: it's definitely out of my salary-class while working on faker :laughing: 
I mean I don't have the :brain: for doing such scripts _or at least not the motivation_

ST-DDT:
@Shinigami92 Maybe we can sit together in the next few days and work on a solution.

src/modules/color/index.ts
```diff
@@ -245,9 +245,29 @@ export class ColorModule {
    * @since 7.0.0
    */
   rgb(options?: {
+    /**
```
EarthyOrange, line 252: According to the following links, these comments can be moved to the `rgb` function's main comment block.
 - https://jsdoc.app/tags-type.html
 - https://jsdoc.app/tags-param.html

The `@param` comment line in the main comment block will look as 👇. The curly braces documents the parameter's type; The following square bracket tells the reader that this param is optional & its contents document the name & default value.
<pre>
@param {string} [options.prefix = '0x'] Prefix of the generated hex color. Only applied when 'hex' format is used.
</pre>

Shinigami92: Thank you for your review but this PR aims for something different then you might have expected.
Please re-read the associated issue.

Your suggested JSDoc is also documentation for JavaScript, not TypeScript. Yes these differ slightly :warning: 
You do not have support for e.g. these `{string}` as TypeScript expects these already in the source code. But I'm not sure if this `[options.prefix = '0x']` is supported and we can try it out instead of writing ``Defaults to `'0x'`.`` at each end. But even this is currently read out by the automation script written by @ST-DDT and so we would need to refactor it for something that is already working.

Therefore for this PR I mark this review-comment as _invalid_/resolved

EarthyOrange: The referenced issue talks about JSDoc. From your response, I am assuming that this project needs to have TSDoc comments.

It is still not clear from the referenced issue why this change is required. Can you or someone give more context on this?

Shinigami92: @EarthyOrange 

> As a user I want to have JSDoc tooltips when hovering over option properties in my IDE so I can directly see what the parameter is aiming for.

ST-DDT (dismissed):

Shinigami92:
@ST-DDT @xDivisionByZerox @import-brain please review/approve this so we can omit further merge conflicts
it's currently for me a non-goal to automate stuff as this could be done in a separate PR if really needed/wanted

this PR just improved the actual DX for faker users and we can manually keep care of reviews in the future when someone adds options and also as well add docs to options if there are still missing ones that wasn't catched

ST-DDT (approved):

xDivisionByZerox (approved):
That was way to much. Should have simply trusted you xD

ST-DDT:
> That was way to much.

Yes

Shinigami92:
> Should have simply trusted you xD

to be fair, I already reviewed what GitHub Co-Pilot wrote onto my screen 🙂

xDivisionByZerox:
The pipeline was canceled - any idea why? I would like to merge this ASAP due to the changes having global effects.

Shinigami92:
> I would like to merge this ASAP due to the changes having global effects.

Okay I will merge right now 4h earlier then preferred

xDivisionByZerox:
> > I would like to merge this ASAP due to the changes having global effects.
> 
> Okay I will merge right now 4h earlier then preferred

That was not what I meant, but appreciated xD I was just wondering if the pipeline was actually failing or canceled on purpose.

PR #1645, opened 2022-12-09
test: run tests shuffled

### LINKED ISSUE #1318: Random seeded tests are not random
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/main/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/main/CONTRIBUTING.md).

### Describe the bug

The random seeded tests are not random.

### Minimal reproduction code

````ts
describe('test', () => {
  describe('a', () => {
    console.log('a0');
    it('test-a', () => {
      console.log('a1');
      expect(true).toBe(true);
    });
  });

  describe(`random seeded tests for seed ${faker.seed()}`, () => {
    console.log('b0');

    it('test-b()', () => {
      console.log('b1');

      expect(true).toBe(true);
    });
  });
});
````

Output:

````txt
stdout | unknown test
a0
b0

stdout | test/word.spec.ts > test > a > test-a
a1

stdout | test/word.spec.ts > test > random seeded tests for seed 2458092977221118 > test-b()
b1
````

If we replace  ``a0`` with the ``fixed seeded tests`` and ``b0`` with the ``random seeded tests``, then we have our current setup.

The test sources currently assume that the test blocks will be run in the following order: a0, a1, b0, b0.

### Additional Context

_No response_

### Environment Info

```shell
TLDR: irrelevant information and the command outputs also wrong information

  System:
    OS: Windows 10 (wrong)
    CPU: irrelevant
    Memory: irrelevant
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.16.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: wrong and unused
    Internet Explorer: I didn't know where this might be installed

----------

The above info are partially wrong as I used win11 when running the command.
Also the browser list is wrong.
Not sure where the command got these versions.
```

- [ ] CJS
- [ ] ESM

### Used Package Manager

pnpm

Shinigami92:
closes #1318 

Waiting on merge of:
- https://github.com/faker-js/faker/pull/1646
- https://github.com/faker-js/faker/pull/1647
- https://github.com/faker-js/faker/pull/1649

This PR configured vitest to run the tests in shuffled mode
https://vitest.dev/config/#sequence-shuffle


ST-DDT:
While this seems to solve the issue of testing always with the same values, it also renders our random seeded tests useless, because the tests are no longer reproducible (by setting the previous/failing seed) in case they uncover a test failure.

ST-DDT:
I assume this still is likely to use seeds values close to our seeded tests (fixed seed + 1 or 2 calls).

xDivisionByZerox (dismissed):

ejcheng (dismissed):

Shinigami92:
Uh damn it, I accidentally pushed it directly to this PR :shrug: 
I wanted to extract it to its own PR, but nevermind anyway.
The new commit fixes the mersenne test and additionally removed a `todo` from a test so it actually gets executed.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #1648, opened 2022-12-09
fix(number): values out of bounds

### LINKED ISSUE #1595: Number functions can generate values out of bound
### Pre-Checks

### Describe the bug

When calling following number methods, they can generate values out of bound.

### Minimal reproduction code

```ts
faker.datatype.number({ min: 2.1, max: 2.9 }) // Should throw error
faker.number.int({ min: 2.1, max: 2.9 }) // Should throw error

faker.datatype.float({ min: 0, max: 1.9, precision: 0.4 }) // Should not generate `2.0`
faker.number.float({ min: 0, max: 1.9, precision: Math.log10(2.5) }) // Should not generate `2.0`
```

### Additional Context

Found while working on #1122 

### Environment Info

```shell
System:
    OS: macOS 13.0.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.65 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.4.0 - ~/.nvm/versions/node/v18.4.0/bin/node
    npm: 8.12.1 - ~/.nvm/versions/node/v18.4.0/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Safari: 16.1
```

- [X] CJS
- [X] ESM

### Used Package Manager

pnpm

ST-DDT:
Fixes: #1595

- #1595

This PR fixes the issue caused by number.int returning values out of bounds, if the given min and max aren't integers themselves.
This also causes number.float to return values out of bounds.

**Case 1:** `int({min: 2.1, max: 2.9})`

Previously -> 2 or 3
Now -> Error

**Case 2:** `int({min: 1.1, max: 2.9})`

Previously -> 1, 2 or 3
Now -> 2

### DIFF
--- a/test/number.spec.ts
+++ b/test/number.spec.ts
@@ -130,15 +130,13 @@ describe('number', () => {
         }).toThrowError(`Max ${max} should be greater than min ${min}.`);
       });
 
-      // TODO @Shinigami92 2022-11-24: https://github.com/faker-js/faker/issues/1595
-      it.todo(
-        'should throw when there is no integer between min and max',
-        () => {
-          expect(() => {
-            faker.number.int({ min: 2.1, max: 2.9 });
-          }).toThrow();
-        }
-      );
+      it('should throw when there is no integer between min and max', () => {
+        expect(() => {
+          faker.number.int({ min: 2.1, max: 2.9 });
+        }).toThrow(
+          new FakerError(`No integer value between 2.1 and 2.9 found.`)
+        );
+      });
     });
 
     describe('float', () => {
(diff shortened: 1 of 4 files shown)


test/number.spec.ts
```diff
@@ -130,15 +130,13 @@ describe('number', () => {
...
-          expect(() => {
-            faker.number.int({ min: 2.1, max: 2.9 });
-          }).toThrow();
-        }
-      );
+      it('should throw when there is no integer between min and max', () => {
+        expect(() => {
+          faker.number.int({ min: 2.1, max: 2.9 });
+        }).toThrow(
+          new FakerError(`No integer value between 2.1 and 2.9 found.`)
```
Shinigami92, line 137: If another member thinks that `No integer value between 2.1 and 2.9 could be generated.` is a better sentence, then lets use that, if not, the just leave as is

Shinigami92 (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #1649, opened 2022-12-09
feat(number): adjust hex default

Shinigami92:
While working on
- #1645 

I found that the hex default max are off by 1
We wanted to generate numbers from 0-f and so on, but the +16 instead of +15 generated 0-10 instead


ST-DDT (approved):
Do we need a hint in the upgrade docs?

Shinigami92:
> Do we need a hint in the upgrade docs?

I think no, as this is a new function in v8 and the old is not touched
When the user already moves to the new one, they can expect these changes already

xDivisionByZerox (approved):

ejcheng (approved):

PR #1652, opened 2022-12-11
infra: setup dev-container

ST-DDT:
Review Usage:

- Install dev container extension in vscode (Docker, ...)
- Right click on the PR -> clone in dev container
- run `pnpm run test`, `pnpm run docs:dev` ... as you would do locally

Later Usage:

- Install dev container extension in vscode (Docker, ...)
- Clone Repo in Dev Container
- run `pnpm run test`, `pnpm run docs:dev` ... as you would do locally

--------------

This is an optional feature, which may (or may not) help external contributors.


xDivisionByZerox:
> Review Usage:
> - Install dev container extension in vscode (Docker, ...)
> - Right click on the PR -> clone in dev container
> - run pnpm run test, pnpm run docs:dev ... as you would do locally

<s>I'm sorry, I can get it to work. I don't have any experience with dev containers. Could you update the description to give a more detailed "guide" on how to review, or link to some resources on that?</s>

I just checked this PR branch out and "Reopen in in container" locally.

xDivisionByZerox (approved):
After allowing the "unsafe" git repo (prompt from VS Code) I could do the following things:

- ✅update code and commit it
- ✅install the dependencies (happens automatically)
- ✅build the project
- ✅build the docs
- ✅serve the docs
- ✅run linter
- ✅run tests

.vscode/extensions.json
```diff
@@ -3,6 +3,7 @@
     "dbaeumer.vscode-eslint",
     "editorconfig.editorconfig",
     "esbenp.prettier-vscode",
-    "redhat.vscode-yaml"
+    "redhat.vscode-yaml",
+    "github.vscode-pull-request-github"
```
Shinigami92, line 7: Not sure about this recommendation
I never used it myself and use other ways, maybe we should only recommend extensions that are directly important for the repo itself
Otherwise we could also recommend versionlense and gitlense and so on, but let the users configure their vscode how they want

Shinigami92 (approved):
I will approve so it's not blocking, but feel free to rethink about vscode recommendation
it could get annoying to recommend extensions that are not really needed at the end of the day

.vscode/extensions.json
ST-DDT: This makes reviewing PRs way easier as you can just right click it (in vscode) to check it out (in a separate dev container if you want).

xDivisionByZerox: I'm curious now. I didn't really get what you meant in your description with "right click on the PR".  Now again (I guess "it" is the PR). How do you right click it? I don't have any PR in vscode.

ST-DDT: > I don't have any PR in vscode.

Install the `GitHub Pull Requests and Issues` (`GitHub.vscode-pull-request-github`) extension.

grafik

Shinigami92: I'm confused how this is related to dev-container

ST-DDT: Without that plugin it is kind of hard to checkout the project since you need to enter the path parts of the repo name and branch everytime you switch between branches.
If I have some free time I'll try to record it as a video.

Shinigami92: You cannot just use `gh pr checkout 1652`?

ST-DDT: Then you have it locally but not in a dev container. Sure you could "reopen" it in a devcontainer, but then you have bad performance due to mounting a windows path into a linux container and still cannot checkout multiple branches at the same time because your checkout points to that branch.

Also the gh util isnt present on everyones computer, but we assume that VSCode is.

Shinigami92: I always expected the terminal is running inside the dev container 👀 

And I still assume that as otherwise `postCreateCommand` is very confusing to me

ST-DDT: Sure once you did the steps to clone it to a dev container you can use whatever you want, but the plugin helps you get a dev container in the first place especially for checking out new PRs so that you can work on one while the other one runs its tests.

xDivisionByZerox: > You cannot just use `gh pr checkout 1652`?

For that you need the GitHub CLI (which I don't have as well). So for me (personal preference) a VS Code Plugin is actually more reasonable than having to install an additional CLI.

> Then you have it locally but not in a dev container. Sure you could "reopen" it in a devcontainer, but then you have bad performance due to mounting a windows path into a linux container [...]

Can confirm this. It took me over an hour to do all things that I've listed [in my comment](https://github.com/faker-js/faker/pull/1652#pullrequestreview-1228347458).
But it might also be because of my low spec, sh*t PC.

Shinigami92:
And updates for this PR?
It has two approvals for several days now
As I won't use this anyways it is not such important for me what is configured in it, so go on and merge as you like

PR #1637, opened 2022-12-06
fix(person): change fullName to use name patterns

matthewmayer:
Fixes #1211

This is work-in-progress tentative draft fix for #1211 which is a simplified version of the ideas in #1216.

- Allows the use of the src/locales/*/person/name.ts legacy patterns
- Allows for locales where first name and last name are reversed e.g. ja, zh_CN
- Allows for prefixes and suffixes in some languages
- Removes gender-specific templates. In particular you don't need to repeat male and female patterns in name.ts any more so I simplified all locales where that was the case
- Since src/locales/en/person/name.ts is defined, don't need a final fallback, we can assume there will always be a name pattern
- Uses the new weightedArrayElement to replicate the old behavior of fullName where prefixes and suffixes are only addded to a small proportion of names https://github.com/faker-js/faker/pull/1654

## Sample output: en, generic

```
'Ramiro Wehner',      'Ernest Smith',
  'Kristopher Gutmann', 'Jason Price',
  'Shannon Kuhn',       'Camille Schoen',
  'Israel Kreiger',     'Mrs. Diane Spinka MD',
  'Jon Shields',        'Luke Boyer',
  'Gerard Parker',      'Mrs. Amber DuBuque IV',
  'Ms. Flora Keebler',  'Orlando Cummerata',
  'Ross Friesen MD',    'Miss Sonya Kuhlman V',
  'Estelle DuBuque',    'Malcolm Weber PhD',
  'Mr. Juan Hamill',    'Ms. Lora Rempel MD'
```
## Sample output: en, male

```
'Morris Cummerata I',
  'Jesse Wolff',
  'Harry Erdman',
  'Rudy Swift',
  'Dwight Feest',
  'Nathan Schumm',
  'Roosevelt Kuhn MD',
  'Shawn Stokes I',
  'Mr. Dexter Wintheiser IV',
  'Leonard Marks Jr.',
  'Lloyd Von',
  'Jon Frami',
  'Mr. Jody Sipes MD',
  'Micheal Ward',
  'Alberto Feest',
  'Terrance Schuppe',
  'Leroy Feest IV',
  'Mr. Duane Senger I',
  'Ismael Walsh',
  'Claude Casper'
```

## Sample output: en, female

```
 'Patty West',            'Hope Lueilwitz',
  'Gwen Skiles',           'Anita Homenick',
  "Lucia O'Reilly DVM",    'Dr. Ashley Aufderhar',
  'Karla Keeling',         'Sonia Corwin',
[...]

PR #1654, opened 2022-12-11
feat(helpers): add new faker.helpers.weightedArrayElement

matthewmayer:
Split from #1637

Adds a new function `faker.helpers.weightedArrayElement` which allows you 
to select a random element from an array, but weighted more to certain 
elements than others. This will be useful for refactoring name patterns, 
where for example you might want to have 

- `FirstName Surname` chosen 90% of the time and
- `FirstName Surname Suffix` chosen 10% of the time


src/modules/helpers/index.ts
```diff
@@ -445,6 +445,49 @@ export class HelpersModule {
...
+   *
+   * @since 8.0.0
+   */
+  weightedArrayElement<T = string>(
+    array: ReadonlyArray<[T, number]> = [] as unknown as ReadonlyArray<
+      [T, number]
+    >
+  ): T {
+    if (array.length === 0) {
+      throw new Error(
```
ST-DDT: Please use FakerError for these errors.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,49 @@ export class HelpersModule {
...
+        'weightedArrayElement expects an array of [value, weight] pairs where weight is a positive number'
+      );
+    }
+    const total = array.reduce((acc, [, weight]) => acc + weight, 0);
+    const random = this.faker.number.int({ min: 0, max: total - 1 });
+    let current = 0;
+    for (const [value, weight] of array) {
+      current += weight;
+      if (random < current) {
```
ST-DDT, line 502: faker.number.int returns inclusive max values, thus you might have to use <=.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,49 @@ export class HelpersModule {
...
+        'weightedArrayElement expects an array of [value, weight] pairs'
+      );
+    }
+    if (!array.every((elt) => typeof elt[1] === 'number' && elt[1] > 0)) {
+      throw new Error(
+        'weightedArrayElement expects an array of [value, weight] pairs where weight is a positive number'
+      );
+    }
+    const total = array.reduce((acc, [, weight]) => acc + weight, 0);
+    const random = this.faker.number.int({ min: 0, max: total - 1 });
```
ST-DDT: I think some users might want to use it with percentages rather than ratios, so we might want to use float here.

test/helpers.spec.ts
```diff
@@ -59,6 +59,14 @@ describe('helpers', () => {
       t.it('noArgs').it('with array', 'Hello World!'.split(''));
     });
 
+    t.describe('weightedArrayElement', (t) => {
+      t.it('with array', [
+        ['sunny', 5],
+        ['rainy', 4],
+        ['snowy', 1],
+      ]);
+    });
```
ST-DDT, line 74: Please also add a test using percentages.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,49 @@ export class HelpersModule {
...
 
+  /**
+   * Returns a weighted random element from the given array. Each element of the array should be an array with two elements: the first is the value, the second is an integer weight.
+   * For example, if there are two elements A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
+   *
+   * @template T The type of the entries to pick from.
+   * @param array Array to pick the value from.
+   *
+   * @example
+   * faker.helpers.weightedArrayElement([['sunny', 5], ['rainy', 4],['snowy', 1]]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```
```suggestion
   * faker.helpers.weightedArrayElement([['sunny', 5], ['rainy', 4], ['snowy', 1]]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```

Or whatever prettier would produce.

ST-DDT (changes requested):
I think we should swap the weight and the value (weight first), because the weight is usually much shorter than the value making it easier to spot in longer lists. What do you/the others think?

````ts
[
  [ 5, 'sunny' ],
  [ 4, 'rainy' ],
  [ 1, 'extremely hazardous blizard storm clouds' ]
]
````

````ts
[
  [ 'sunny', 5 ],
  [ 'rainy', 4 ],
  [ 'extremely hazardous blizard storm clouds', 1 ]
]
````

**Comment by ST-DDT** [comment]
Should we allow weight zero for temporarily disabled or dynamically weighted items or should we expect the user to remove it before passing it to our method?

File: test/helpers.spec.ts [hunk 6]
```diff
@@ -140,6 +148,49 @@ describe('helpers', () => {
...
+          expect(testArray.map((a) => a[0])).toContain(actual);
+        });
+
+        it('should return the only element in the array when there is only 1', () => {
+          const testArray: [string, number][] = [['hello', 10]];
+          const actual = faker.helpers.weightedArrayElement(testArray);
+
+          expect(actual).toBe('hello');
+        });
+        it('should throw if the array is empty', () => {
```
Shinigami92:
```suggestion
        });

        it('should throw if the array is empty', () => {
```

test/helpers.spec.ts
```diff
@@ -140,6 +148,49 @@ describe('helpers', () => {
...
+          expect(() => faker.helpers.weightedArrayElement([])).to.throw();
+        });
+        it('should throw if any weight is zero', () => {
```
Shinigami92:
```suggestion
        });

        it('should throw if any weight is zero', () => {
```

test/helpers.spec.ts
```diff
@@ -140,6 +148,49 @@ describe('helpers', () => {
...
+          const testArray: [string, number][] = [
+            ['hello', 0],
+            ['to', 5],
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).to.throw();
+        });
+        it('should throw if any weight is negative', () => {
```
Shinigami92:
```suggestion
        });

        it('should throw if any weight is negative', () => {
```

test/helpers.spec.ts
```diff
@@ -140,6 +148,49 @@ describe('helpers', () => {
...
+          const testArray: [string, number][] = [
+            ['hello', -1],
+            ['to', 5],
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).to.throw();
+        });
+      });
```
Shinigami92, line 248: The `ReadonlyArray` we are using just indicates that the array passed itself will not be mutated, but not the tuples inside :warning: 

So e.g. it would still be possible to change a value like `array[0][0] = 'override'`

Therefore I would suggest to add a test that freeze the whole array passed to the function and make sure that it doesn't throw

Shinigami92 (changes requested):

test/helpers.spec.ts
matthewmayer: One other thing I wasn't sure about how to test that say
```
faker.helpers.weightedArrayElement([["foo",1, "randomextraelement"]])
```
throws?

Because if I write the test in TypeScript it the compiler wont let me do that, but in JS i can obviously pass in a 3rd parameter.

Shinigami92: Yeah, you can pass this, but I think we should not throw in that case
It's the same as passing an options object to a function that has an additional unused param

matthewmayer:
> Should we allow weight zero for temporarily disabled or dynamically weighted items or should we expect the user to remove it before passing it to our method?

i think allowing zeroes causes more confusion, for example what would you expect to happen if your array had three items, all with weight zero?

test/helpers.spec.ts
matthewmayer: so maybe

```
if (!array.every((elt) => elt.length === 2)) {
      throw new Error(
        'weightedArrayElement expects an array of [value, weight] pairs'
      );
    }
```

should change to
```
if (!array.every((elt) => elt.length >= 2)) {
```

Shinigami92:
> > Should we allow weight zero for temporarily disabled or dynamically weighted items or should we expect the user to remove it before passing it to our method?
> 
> i think allowing zeroes causes more confusion, for example what would you expect to happen if your array had three items, all with weight zero?

All have same chance
so 1/3, 1/3 and 1/3

matthewmayer:
> > > Should we allow weight zero for temporarily disabled or dynamically weighted items or should we expect the user to remove it before passing it to our method?
> > 
> > 
> > i think allowing zeroes causes more confusion, for example what would you expect to happen if your array had three items, all with weight zero?
> 
> All have same chance so 1/3, 1/3 and 1/3

Intuitively yes, but that would require some extra code to deal with that case, which i think adds more complexity and chance for bugs?

matthewmayer:
I note that `faker.helpers.float` defaults to a precision of 0.01, which means this is biased for small percentages. What's the cleanest way to request the max precision for the random float?

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,51 @@ export class HelpersModule {
...
 
+  /**
+   * Returns a weighted random element from the given array. Each element of the array should be an array with two elements: the first is a weight which can be an integer or a float, the second is the value.
+   * For example, if there are two elements A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
+   *
+   * @template T The type of the entries to pick from.
+   * @param array Array to pick the value from.
+   *
+   * @example
+   * faker.helpers.weightedArrayElement([[5, 'sunny'], [4, 'rainy'], [1, 'snowy']]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```
Shinigami92: IMO we should switch the values inside the tuples (but first also lets ask @ST-DDT what he thinks)
So we will have `[['sunny', 5], ['rainy', 4], ['snowy', 1]]`

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,51 @@ export class HelpersModule {
...
+   *
+   * @since 8.0.0
+   */
+  weightedArrayElement<T = string>(
```
Shinigami92: I think we should not use a default here, just let TS infer it

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,51 @@ export class HelpersModule {
...
+    array: ReadonlyArray<[number, T]> = [] as unknown as ReadonlyArray<
+      [number, T]
+    >
```
Shinigami92: Do we really need this `as unknown as`?

```ts
array: ReadonlyArray<WeightedTuple> = [] // This does not work?
```

On the other side, why even use a default value here? I think we can require it always :thinking: 
Otherwise in line 466 we would throw an error anyway :shrug:

Shinigami92 (changes requested):

src/modules/helpers/index.ts
ST-DDT: See my reasons here: https://github.com/faker-js/faker/pull/1654#pullrequestreview-1212662634

Also json records/java maps and a lot of other datastructure use key and then value.

src/modules/helpers/index.ts
ST-DDT: Yes, there shouldnt be a default.

src/modules/helpers/index.ts
ST-DDT: I think so too. What was the original intention when adding this?

src/modules/helpers/index.ts
Shinigami92: Interesting, but I must say I would be confused by first using it without reading the manual

ST-DDT:
> I note that `faker.helpers.float` defaults to a precision of 0.01, which means this is biased for small percentages. What's the cleanest way to request the max precision for the random float?

Maybe open a new issue for that (max precision float)?

ST-DDT:
> > i think allowing zeroes causes more confusion, for example what would you expect to happen if your array had three items, all with weight zero?
> 
> All have same chance so 1/3, 1/3 and 1/3

While we could technically forward to array element then, I think the usual expectation for zero weight is "dont show up". If everything is weight zero this expectation shouldnt change.
We could treat it as an empty array, but maybe we should just disallow it until someone actually has a need for that and then we can discuss the pros and cons in a separate context.

src/modules/helpers/index.ts
ST-DDT: @JessicaSachs I think you showed a method like this in one of our teammettings. Do you remember where that method was from? Do you have a preference here?

src/modules/helpers/index.ts
matthewmayer: sorry my TS is a bit fuzzy, i will need a bit more assistance understanding what i need to change on this line.

src/modules/helpers/index.ts
matthewmayer: sorry my TS is a bit fuzzy, i will need a bit more assistance understanding what i need to change on this line.

matthewmayer:
> > > i think allowing zeroes causes more confusion, for example what would you expect to happen if your array had three items, all with weight zero?
> > 
> > 
> > All have same chance so 1/3, 1/3 and 1/3
> 
> While we could technically forward to array element then, I think the usual expectation for zero weight is "dont show up". If everything is weight zero this expectation shouldnt change. We could treat it as an empty array, but maybe we should just disallow it until someone actually has a need for that and then we can discuss the pros and cons in a separate context.

I think "disallow and throw error until someone actually has a use-case for it" is the best approach :)

src/modules/helpers/index.ts
```suggestion
  weightedArrayElement<T>(
```

src/modules/helpers/index.ts
```suggestion
    array: ReadonlyArray<[number, T]>
```

ST-DDT:
Please run the formatter once.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,53 @@ export class HelpersModule {
...
+    array: ReadonlyArray<[number, T]>
+  ): T {
+    if (array.length === 0) {
+      throw new FakerError(
+        'weightedArrayElement expects an array with at least one element'
+      );
+    }
+    if (!array.every((elt) => elt.length >= 2)) {
+      throw new FakerError(
+        'weightedArrayElement expects an array of [value, weight] pairs'
```
ST-DDT: The error message is wrong. (Type needs to bw flipped).

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,53 @@ export class HelpersModule {
...
+      );
+    }
+    if (!array.every((elt) => typeof elt[0] === 'number' && elt[0] > 0)) {
+      throw new FakerError(
+        'weightedArrayElement expects an array of [value, weight] pairs where weight is a positive number'
```
ST-DDT: Same here.

test/helpers.spec.ts
```diff
@@ -140,6 +148,76 @@ describe('helpers', () => {
...
+
+        it('should return the only element in the array when there is only 1', () => {
+          const testArray: [number, string][] = [[10, 'hello']];
+          const actual = faker.helpers.weightedArrayElement(testArray);
+
+          expect(actual).toBe('hello');
+        });
+
+        it('should throw if the array is empty', () => {
+          expect(() => faker.helpers.weightedArrayElement([])).to.throw();
```
ST-DDT: We usually check the actual error. See here for an example:

https://github.com/faker-js/faker/blob/9a1226172fe23aa763c14f5488131373548bba21/test/helpers.spec.ts#L609-L613

test/helpers.spec.ts
```diff
@@ -140,6 +148,76 @@ describe('helpers', () => {
...
+        });
+
+        it('should throw if any weight is zero', () => {
+          const testArray: [number, string][] = [
+            [0, 'hello'],
+            [5, 'to'],
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).to.throw();
```
ST-DDT: Same here

test/helpers.spec.ts
```diff
@@ -140,6 +148,76 @@ describe('helpers', () => {
...
+        });
+
+        it('should throw if any weight is negative', () => {
+          const testArray: [number, string][] = [
+            [-1, 'hello'],
+            [5, 'to'],
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).to.throw();
```
ST-DDT: And here

test/helpers.spec.ts
```diff
@@ -140,6 +148,76 @@ describe('helpers', () => {
...
+        it('should not throw with a frozen array', () => {
+          const testArray: [number, string][] = [
+            [7, 'ice'],
+            [3, 'snow'],
+          ];
+          const frozenArray = Object.freeze(testArray);
+          expect(() =>
+            faker.helpers.weightedArrayElement(frozenArray)
+          ).to.not.throw();
+        });
```
ST-DDT, line 247: And here

matthewmayer: We expect it to not throw what 😃

ST-DDT: Ups, my bad.

src/modules/helpers/index.ts
ST-DDT: > Interesting, but I must say I would be confused by first using it without reading the manual

I searched the internet and also found variants that use two separate arrays, so I think there are many ways to do the same and this is just one like the other. Amd you have to read the example for 1s anyway to know how to use it.
Maybe this gets even easier to grasp if we move each pair to their own line like prettier might do.

test/helpers.spec.ts
ST-DDT: Could you please add that test here?

ST-DDT (dismissed):

src/modules/helpers/index.ts
xDivisionByZerox: > Also json records/java maps and a lot of other datastructure use key and then value.

I would probably allow a type like `WeightedValue<T> = { value: T; weight: number }` as well. 
This should be more readable than a tuple in a dataset where the value is of type `number`. At least I think so.

Shinigami92: This is more verbose, but has the benefit that users can also decide if the want to use `[{value, weight}]` or `[{weight, value}]`

ST-DDT: Do you want it exclusively or additionally?

matthewmayer: I think allowing additional formats for the weights would be nice, but then we would still need to decide what the best format to use would be when using this method internally like in the full name patterns which I intend to use this for.

matthewmayer: Another option could be to make this function internal to start with. Then if we find the selected format is awkward, it's easier to change it without breaking changes.

Shinigami92: I'm not a fan of making this internal
I prefer @xDivisionByZerox's solution and I think for now we should even only go with the explicit input
If a faker-user then wants the "short" version, they can just write `[['sunny', 5], ['rainy', 4], ['snowy', 1]].map(x => ({ weigth: x[1], value: x[0] }))` and get what is needed as input

matthewmayer: Ok I'll make that change in a few days.

Shinigami92 (dismissed):
Reading the tests, I think it was a really good decision to switch to object instead of tuple array
Yes it is more verbose, but by just reading and looking at the code, it is instant more understandable immediately whats going on 👍

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,53 @@ export class HelpersModule {
     return array[index];
   }
 
+  /**
+   * Returns a weighted random element from the given array. Each element of the array should be an object with two keys: weight which can be an integer or a float, and value.
+   * For example, if there are two elements A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
```
ST-DDT: Please add a hint that the weight describes the chance of returning that element in relation to the sum/total weight.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,53 @@ export class HelpersModule {
...
+   * @template T The type of the entries to pick from.
+   * @param array Array to pick the value from.
+   *
+   * @example
+   * faker.helpers.weightedArrayElement([{ weight: 5, value: 'sunny'}, { weight: 4, value: 'rainy'}, { weight: 1, value: 'snowy'}]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
+   *
+   * @since 8.0.0
+   */
+  weightedArrayElement<T>(
+    array: ReadonlyArray<{ weight: number; value: T }>
```
ST-DDT, line 479: Should we add jsdocs the the nested keys as well? @Shinigami92 has one or two that add them there as well.
https://github.com/faker-js/faker/pull/1644

Shinigami92: #1644 is not reviewed or merged yet, so I assume it's just important what first get merged into next branch

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,55 @@ export class HelpersModule {
...
+    if (array.length === 0) {
+      throw new FakerError(
+        'weightedArrayElement expects an array with at least one element'
+      );
+    }
+    if (!array.every((elt) => elt.value)) {
+      throw new FakerError(
+        'weightedArrayElement expects an array of {weight, value} objects'
+      );
+    }
```
xDivisionByZerox: I'm unsure if we need to check this, as our general agreement is "we only check, what typescript allows". Also, we are not able to tests this, without making typescript angry.

Shinigami92: Yeah maybe the error message is misleading here
Currently it checks indirectly if any value contains `0` which is a `falsy` value

xDivisionByZerox: > Currently it checks indirectly if any value contains 0 which is a falsy value

I totally missed this. Even more reason to remove this, as `0`, `''`, and/or `false` should be viable options for `value`

ST-DDT: Me too. We should also add a test to ensure falsy values are allowed.

matthewmayer: I removed those lines and added one more test.

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,50 @@ export class HelpersModule {
...
+   * Returns a weighted random element from the given array. Each element of the array should be an object with two keys `weight` and `value`.
+   * Each `weight` key should be a number representing the probability of selecting the value, relative to the sum of the weights. Weights can be any positive float or integer.
+   * Each `value` key should be the corresponding value.
+   * For example, if there are two values A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
+   *
+   * @template T The type of the entries to pick from.
+   * @param array Array to pick the value from.
+   *
+   * @example
+   * faker.helpers.weightedArrayElement([{ weight: 5, value: 'sunny'}, { weight: 4, value: 'rainy'}, { weight: 1, value: 'snowy'}]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```
Shinigami92:
```suggestion
   * faker.helpers.weightedArrayElement([{ weight: 5, value: 'sunny' }, { weight: 4, value: 'rainy' }, { weight: 1, value: 'snowy' }]) // 'sunny', 50% of the time, 'rainy' 40% of the time, 'snowy' 10% of the time
```

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,50 @@ export class HelpersModule {
...
+    array: ReadonlyArray<{ weight: number; value: T }>
+  ): T {
+    if (array.length === 0) {
+      throw new FakerError(
+        'weightedArrayElement expects an array with at least one element'
+      );
+    }
+    if (!array.every((elt) => elt.weight > 0)) {
+      throw new FakerError(
+        'weightedArrayElement expects an array of {weight, value} objects where weight is a positive number'
```
Shinigami92:
```suggestion
        'weightedArrayElement expects an array of { weight, value } objects where weight is a positive number'
```

test/helpers.spec.ts
```diff
@@ -145,6 +159,94 @@ describe('helpers', () => {
...
+        it('should throw if any weight is zero', () => {
+          const testArray = [
+            { weight: 0, value: 'hello' },
+            { weight: 5, value: 'to' },
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).toThrowError(
+            new FakerError(
+              'weightedArrayElement expects an array of {weight, value} objects where weight is a positive number'
```
Shinigami92:
```suggestion
              'weightedArrayElement expects an array of { weight, value } objects where weight is a positive number'
```

test/helpers.spec.ts
```diff
@@ -145,6 +159,94 @@ describe('helpers', () => {
...
+        it('should throw if any weight is negative', () => {
+          const testArray = [
+            { weight: -1, value: 'hello' },
+            { weight: 5, value: 'to' },
+          ];
+          expect(() =>
+            faker.helpers.weightedArrayElement(testArray)
+          ).toThrowError(
+            new FakerError(
+              'weightedArrayElement expects an array of {weight, value} objects where weight is a positive number'
```
Shinigami92:
```suggestion
              'weightedArrayElement expects an array of { weight, value } objects where weight is a positive number'
```

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,50 @@ export class HelpersModule {
     return array[index];
   }
 
+  /**
+   * Returns a weighted random element from the given array. Each element of the array should be an object with two keys `weight` and `value`.
+   * Each `weight` key should be a number representing the probability of selecting the value, relative to the sum of the weights. Weights can be any positive float or integer.
+   * Each `value` key should be the corresponding value.
+   * For example, if there are two values A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
```
Shinigami92: When hovering over these JSDoc it is very long due to it is rendered as markdown which would need e.g. double-whitespaces at the end to simulate a newline

But instead of doing that and it will get removed by prettier we should maybe use ul-li for this

```suggestion
   * Returns a weighted random element from the given array. Each element of the array should be an object with two keys `weight` and `value`.
   *
   * - Each `weight` key should be a number representing the probability of selecting the value, relative to the sum of the weights. Weights can be any positive float or integer.
   * - Each `value` key should be the corresponding value.
   * 
   * For example, if there are two values A and B, with weights 1 and 2 respectively, then the probability of picking A is 1/3 and the probability of picking B is 2/3.
```



This makes it much more readable

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -445,6 +445,52 @@ export class HelpersModule {
...
+    });
+    let current = 0;
+    for (const { weight, value } of array) {
+      current += weight;
+      if (random < current) {
+        return value;
+      }
+    }
+    // In case of rounding errors, return the last element
+    return array[array.length - 1].value;
```
xDivisionByZerox, line 508: Can we somehow force this case in a test? I have no idea how sadly.

xDivisionByZerox (approved):

src/modules/helpers/index.ts
Shinigami92: I could be wrong, but I think because of line 472 and 484-486 it is impossible to reach this code
So it is only more like a fallback to satisfy TypeScript, is it?

matthewmayer: https://javascript.plainenglish.io/why-0-1-0-2-0-3-in-javascript-d7e218224a72

Adding floating point numbers in JS is weird

But it's practically not going to happen

matthewmayer: Being certain it never happens requires a fairly good understanding of how JS floating point addition works, and the exact implementation of `faker.number.float` so i prefer to have this line as a fail-safe rather than risk returning undefined in certain pathological cases.

xDivisionByZerox: > Being certain it never happens requires a fairly good understanding of how JS floating point addition works, and the exact implementation of `faker.number.float` so i prefer to have this line as a fail-safe rather than risk returning undefined in certain pathological cases.

I totally get this and think it's a good way of implementing this. I was just wondering if we could somehow force this to actually prove your statement "look JS float is weird, that's why this is in place". 

I'm more afraid that someone in the future will have the same conclusion as @Shinigami92, remove the fall back, and will then run into a failing test. That's why I was asking for a possible test^^

xDivisionByZerox (dismissed):

src/modules/helpers/index.ts
ST-DDT: IMO the likelihood of that is low, because we have an explicit comment why it is there.

Shinigami92 (dismissed):

ST-DDT:
The linter now complains about this.

ST-DDT (approved):

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #1667, opened 2022-12-16
docs(datatype): allow markdown in @see links

matthewmayer:
The backtick in the `@see` causes the link to get broken and the second see also not to appear:

On https://next.fakerjs.dev/api/datatype.html#number

Screenshot 2022-12-16 at 14 15 45

after fix:

Screenshot 2022-12-16 at 14 22 50


Shinigami92:
This is for rendering in VSCode tooltips



Therefore I assume instead of changing the source text, we need to somehow fix the docs automatisation text extractor

matthewmayer:
i guess its this method? https://github.com/faker-js/faker/blob/next/scripts/apidoc/utils.ts#L114 but dont really understand what its doing or why there are two cases for a length of 1 or 3?

matthewmayer:
previous PR where this was implemented: https://github.com/faker-js/faker/pull/1270/files

matthewmayer:
I tried fixing the @see extraction, which works OK, although it renders like this in the docs, maybe would be better if the `` rendered as `<code></code>`

image

ST-DDT:
You probably have to call mdToHtml somewhere.

scripts/apidoc/utils.ts
```diff
@@ -117,9 +117,21 @@ export function extractSeeAlsos(signature?: SignatureReflection): string[] {
     if (content.length === 1) {
       return joinTagContent(tag);
     }
```
ST-DDT: This is no longer needed.

scripts/apidoc/utils.ts
```diff
@@ -117,9 +117,21 @@ export function extractSeeAlsos(signature?: SignatureReflection): string[] {
...
-    return tag.content
-      .filter((_, index) => index % 3 === 1) // ['-', 'content', '\n']
-      .map((part) => part.text);
+    // If the @see tag contains code in backticks, the content is split into multiple parts.
+    // So we join together, split on newlines and filter out empty tags.
+    const links = tag.content
+      .map((c) => c.text)
+      .join('')
```
ST-DDT:
```suggestion
    return joinTagParts(tag.content)
```

scripts/apidoc/utils.ts
```diff
@@ -117,9 +117,21 @@ export function extractSeeAlsos(signature?: SignatureReflection): string[] {
...
+      .split('\n')
+      .map((link) => {
+        link = link.trim();
+        if (link.startsWith('-')) {
+          link = link.slice(1).trim();
+        }
+        return link;
+      })
+      .filter((link) => link.length);
```
ST-DDT:
```suggestion
      .filter((link) => link);
```

An empty string is also falsy. Alternatively, trim after `joinTagParts`.

scripts/apidoc/utils.ts
```diff
@@ -117,9 +117,21 @@ export function extractSeeAlsos(signature?: SignatureReflection): string[] {
...
+    return links;
```
ST-DDT:
```suggestion
```

scripts/apidoc/utils.ts
```diff
@@ -117,9 +117,21 @@ export function extractSeeAlsos(signature?: SignatureReflection): string[] {
...
+    // If the @see tag contains code in backticks, the content is split into multiple parts.
+    // So we join together, split on newlines and filter out empty tags.
+    const links = tag.content
+      .map((c) => c.text)
+      .join('')
+      .split('\n')
+      .map((link) => {
+        link = link.trim();
+        if (link.startsWith('-')) {
+          link = link.slice(1).trim();
```
ST-DDT, line 123: Why do you use `.slice(1).trim()` instead of `.slice(2)`? (Just curious)

ST-DDT (commented):
Thanks for fixing this.
I only have some minor nitpicking, but other than that this is ready to be merged.

scripts/apidoc/utils.ts
matthewmayer: in case in theory it began with `-foo` not `- foo` (maybe not possible in reality)

ST-DDT (dismissed):

xDivisionByZerox (dismissed):

scripts/apidoc/signature.ts
```diff
@@ -80,9 +80,14 @@ function comparableSanitizedHtml(html: string): string {
     .replace(/&quot;/g, '"')
     .replace(/&#39;/g, "'");
 }
-
-function mdToHtml(md: string): string {
-  const rawHtml = markdown.render(md);
+/**
```
Shinigami92, line 84:
```suggestion

/**
```

scripts/apidoc/signature.ts
```diff
@@ -80,9 +80,14 @@ function comparableSanitizedHtml(html: string): string {
...
+ * Converts Markdown to an HTML string and sanitizes it.
+ * @param md The markdown to convert.
+ * @param inline Whether to render the markdown as inline, without a wrapping <p> tag. Default false.
```
Shinigami92:
```suggestion
 * @param inline Whether to render the markdown as inline, without a wrapping `<p>` tag. Defaults to `false`.
```

scripts/apidoc/signature.ts
```diff
@@ -164,7 +169,9 @@ export function analyzeSignature(
     examples += `${exampleTags.join('\n').trim()}\n`;
   }
 
-  const seeAlsos = extractSeeAlsos(signature);
+  const seeAlsos = extractSeeAlsos(signature).map((seealso) =>
+    mdToHtml(seealso, true)
+  );
```
Shinigami92:
```suggestion
  const seeAlsos = extractSeeAlsos(signature).map((seeAlso) =>
    mdToHtml(seeAlso, true)
  );
```

Shinigami92 (dismissed):

ST-DDT (dismissed):

ST-DDT:
The snapshots need updating after the last typedoc/vitepress update

ST-DDT:
The PR is still failing.

ST-DDT (approved):

Shinigami92 (approved):

Issue #1596, opened 2022-11-24
Refactor `number.float` parameter `precision` to `fractionDigits`


Shinigami92:
### Clear and concise description of the problem

We want to refactor the behavior of the `precision` parameter of `number.float` to be adjusted to `Number#toFixed` behavior.

This **wont** affect the _legacy_ `datatype.float` method!

### Suggested solution

- `precision` -> `fractionDigits`
- We then want to have a well defined developer experience and support steps like 0.4 and such
  So we might need an addition parameter like `step(s)` or `precision` and then talk about how this behaves
  - see #1595

### Additional context

Extracted from 
- #1122

Shinigami92:
**Team Decision:**

We want support both parameters: `fractionDigits` and the second one
If they are incompatible throw an `FakerError`

xDivisionByZerox:
I will take this.

ST-DDT:
Please wait for https://github.com/faker-js/faker/pull/1675 to be merged.

ST-DDT:
Continuation of the discussion that came up in https://github.com/faker-js/faker/pull/1855#issuecomment-1435655823

@matthewmayer suggested adding a new method for stepped values to separate it from float.

ST-DDT:
@matthewmayer Do you have a suggestion for a method name for the new `step` method?

Maybe `faker.number.stepped`?

ST-DDT:
**Team decision**

The proposal to move the `precision`/`step` logic into a different method has been rejected.
We will keep this `precision`/`step` option in the float method's single signature and document that the parameters are mutually exclusive.

ST-DDT:
**Team Decision**

We don't need this for v8.0.
(We will likely immediately start with v9 after v8.0)

ST-DDT:
Related issue: https://github.com/faker-js/faker/issues/2186

matthewmayer:
So basically
- this would add a new parameter `fractionDigits` as an alternative to `precision`
- fractionDigits must be an integer
- if set then `precision = 10 ** -fractionDigits`

Would you allow negative fractionDigits? (`multipleOf:10` is probably clearer than `fractionDigits:-1`)

ST-DDT:
> Would you allow negative fractionDigits? (`multipleOf:10` is probably clearer than `fractionDigits:-1`)

I'm not sure whether we should actually forbid badly readable input (unless it has other side effects).

matthewmayer:
[...]

PR #1675, opened 2022-12-19
feat(number)!: default to high precision float

### LINKED ISSUE #1655: change default precision of faker.number.float
> I note that `faker.helpers.float` defaults to a precision of 0.01, which means this is biased for small percentages. What's the cleanest way to request the max precision for the random float?

Maybe open a new issue for that (max precision float)?

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/1654#issuecomment-1345579203_

matthewmayer:
fix #1655

Note in order to reduce unexpected changes, i made a new code path for the default case when precision is not passed which grabs a float from Mersenne directly, while keeping the old logic intact (going via an integer) when a precision is pased.

I also added {precision: 0.01} in a few places in the code where float() was previously used to avoid unnecessary changes.


src/internal/mersenne/mersenne.ts
```diff
@@ -7,14 +7,23 @@ import Twister from './twister';
...
   next(options: { max: number; min: number }): number;
 
+  /**
+   * Generates a random float between `[min, max)`.
+   *
+   * @param options The options to generate a random number.
+   * @param options.min The minimum number.
+   * @param options.max The maximum number.
+   */
+  nextFloat(options: { max: number; min: number }): number;
```
ST-DDT: How about changing this to `next(): [0-1)` and just do the min max math in the int/float module directly? 
That would simplify the implementation/changes for the tree shakeable module functions v8.2 (ref fakerCore)

matthewmayer: tried that. i changed the method name to nextReal() because i think its nicer to introduce a new method than change the behavior of an existing one significantly (and its only used internally).

src/modules/number/index.ts
```diff
@@ -102,14 +102,19 @@ export class NumberModule {
...
     }
-
-    const factor = 1 / precision;
-    const int = this.int({
-      min: min * factor,
-      max: max * factor,
-    });
-
-    return int / factor;
+    if (precision !== undefined) {
```
Shinigami92, line 146:
```suggestion

    if (precision != null) {
```

src/modules/number/index.ts
```diff
@@ -102,14 +102,19 @@ export class NumberModule {
...
+      const factor = 1 / precision;
```
Shinigami92, line 151: I think we need to check for `precision == 0` as otherwise this will result in `Infinity`

docs/guide/upgrading.md
```diff
@@ -103,11 +103,20 @@ The `faker.address.*` methods will continue to work as an alias in v8 and v9, bu
...
+//OLD
+faker.datatype.number({ max: 100 }); //35
+faker.datatype.number({ max: 100, precision: 0.01 }); //35.21
+faker.datatype.float({ max: 100 }); //35.21
+faker.datatype.float({ max: 100, precision: 0.001 }); //35.211
+
+//NEW
+faker.number.int({ max: 100 }); //35
+faker.number.float({ max: 100 }); //35.21092065742612
+faker.number.float({ max: 100, precision: 0.01 }); //35.21
```
```suggestion
// OLD
faker.datatype.number({ max: 100 }); // 35
faker.datatype.number({ max: 100, precision: 0.01 }); // 35.21
faker.datatype.float({ max: 100 }); // 35.21
faker.datatype.float({ max: 100, precision: 0.001 }); // 35.211

// NEW
faker.number.int({ max: 100 }); // 35
faker.number.float({ max: 100 }); // 35.21092065742612
faker.number.float({ max: 100, precision: 0.01 }); // 35.21
```

src/modules/number/index.ts
matthewmayer: Maybe could just check precision>0?

Zero and negative precisions are meaningless.

ST-DDT: Should we throw an error if precision is defined but invalid?

matthewmayer: Honestly I think precision is not a very useful parameter and we should move towards deprecating it. 

For example what would a precision of 0.7 or 3 intuitively mean?

And JavaScript rounding rules means you don't necessarily get the expected number of decimal places if you pass a precision of say 0.001. 

Most use cases would be much better served by calling toFixed on the float

ST-DDT: Well, precision allows for custom steps in the values.

E.g. 0.25 -> 0, 0.25, 0.5, 0.75, ...
Or 4 -> 0, 4 , 8, 12, ...

But I guess having decimal digits or something similar has more uses for users.

Shinigami92: https://github.com/faker-js/faker/issues/1596#issuecomment-1326654346

ST-DDT:
Looks like we broke a test.

matthewmayer:
I don't think that test makes much sense (precision 0)

Probably throwing an error is better

matthewmayer:
Some weird error with codecov on CI

ST-DDT:
> I don't think that test makes much sense (precision 0)
> 
> Probably throwing an error is better

Mhh, a float with a precision of zero is an int.
I would still call it a valid value, but IMO that is not that important.
What do the others think?

matthewmayer:
> > I don't think that test makes much sense (precision 0)
> > Probably throwing an error is better
> 
> Mhh, a float with a precision of zero is an int. I would still call it a valid value, but IMO that is not that important. What do the others think?

no, a float with precision 1 is an int

ST-DDT:
> no, a float with precision 1 is an int

Oh yeah, you are totally right.

Do we have a test that ensures the following

````ts
const actual = faker.number.float({precision: 1, ?});
expect(actual).toSatisfy(Number.isInteger);
````

?

**Review by ST-DDT — DISMISSED** [review]
The `precision = 1` => int test would be nice, but nothing blocking IMO.

File: src/internal/mersenne/mersenne.ts [hunk 5]
```diff
@@ -7,13 +7,9 @@ import Twister from './twister';
...
   /**
-   * Generates a random number between `[min, max)`. The result is already floored.
-   *
-   * @param options The options to generate a random number.
-   * @param options.min The minimum number.
-   * @param options.max The maximum number.
+   * Generates a random float between `[0, 1)`.
    */
-  next(options: { max: number; min: number }): number;
+  nextReal(): number;
```
xDivisionByZerox: Why do we need to change the name of this? Is it related to something mathematical? 
`Math.random` returns a floating point number as well but isn't called `randomReal` (at least that's my thought process).

src/modules/number/index.ts
```diff
@@ -81,8 +81,8 @@ export class NumberModule {
     const mersenne: Mersenne =
       // @ts-expect-error: access private member field
       this.faker._mersenne;
-
-    return mersenne.next({ min: effectiveMin, max: effectiveMax + 1 });
+    const real = mersenne.nextReal();
+    return Math.floor(real * (effectiveMax + 1 - effectiveMin) + effectiveMin);
```
xDivisionByZerox, line 85: Before, `float` depended on the implementation of `int` since `mersenne.next` would return an integer. Since the implementation changed, can we not simply `int` as a more specific call of `float` and get rid of the entire implementation?

```ts
int(options number | { ... } = { }): number {
  // defaulting values
  const { ... } = options;    
  const effectiveMin = Math.ceil(min);
  const effectiveMax = Math.floor(max);
  
  return this.float({
    min: effectiveMin,
    max: effectiveMax,
    precision: 1,
  });
}
```

I'm proposing this since it seems to me that the return value is doing a lot of "mathematic".
It might be hard to understand what is going on for people not so "familiar" with mathematics.

src/modules/number/index.ts
```diff
@@ -143,13 +143,23 @@ export class NumberModule {
       throw new FakerError(`Max ${max} should be greater than min ${min}.`);
     }
 
-    const factor = 1 / precision;
-    const int = this.int({
-      min: min * factor,
-      max: max * factor,
-    });
+    if (precision !== undefined) {
```
xDivisionByZerox, line 146: I'm not a fan of maintaining two code paths here. 
We should simply default `precision` in ln 136 with the other properties and get rid of the old implementation.

src/internal/mersenne/mersenne.ts
matthewmayer: Was trying to avoid a breaking change in behavior to the method (mersenne methods are not public, but they used to be).

src/modules/number/index.ts
matthewmayer: as I mentioned before I think the precision property on float is unwise and should be deprecated on the long run so I wouldn't want to make int dependent on that.

src/modules/number/index.ts
matthewmayer: But the point of this change is to not have a default precision any more.

test/number.spec.ts
```diff
@@ -220,10 +221,16 @@ describe('number', () => {
         }
       });
 
-      it('provides number with a precision 0', () => {
-        const actual = faker.number.float({ precision: 0 });
+      it('throws an error for precision 0', () => {
```
xDivisionByZerox, line 224: Should this behavior be documented in the upgrading docs, since it was working in the previous version?

matthewmayer: I'd say it's just a bug fix. Precision 0 is meaningless.

src/internal/mersenne/mersenne.ts
xDivisionByZerox: So by removing `next` you are not introducing a breaking change?^^
The reason why the function was called `next` so it could potentially be used as an [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol).

src/modules/number/index.ts
xDivisionByZerox: And as it got pointed out before, the `precision` argument will be redefined (since it is more of a step argument) and `fractionalDigits` in its place will be introduced as stated in #1596.
When this happens we can savely swap `precision` arguments to `fractionalDigits`.

Eigther way this can be done at a later point as well.

src/modules/number/index.ts
xDivisionByZerox: Yeah, my bad. I didn't notice that you weren't using `precision` in the other code path at all.
In that case, this can be addressed when implementing #1596.

test/number.spec.ts
xDivisionByZerox: I thought about categorizing this specific case as bug fix as well.
Wanted to have feedback if anyone else would suggest that.

src/internal/mersenne/mersenne.ts
matthewmayer: I'm explicitly making a breaking change to a private method that a consumer would definitely notice, rather than changing the behavior of a method which they might not.

matthewmayer: > The reason why the function was called `next` so it could potentially be used as an [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol).

That's fine could keep it as next, but then should add that naming reason as a comment.

xDivisionByZerox: > I'm explicitly making a breaking change to a private method that a consumer would definitely notice, rather than changing the behavior of a method which they might not.

This is not required since an end user will no longer be able to access `faker.mersenne` anyway. At that point, they would already receive an error. Then they had to explicitly target the private property for `_mersenne` which is not officially exposed by us.

> I'm explicitly making a breaking change to a private method that a consumer would definitely notice, rather than changing the behavior of a method which they might not.

You might be right with that. Guess this wasn't done because it was already marked as `@internal` only. Sill, I'll open a PR do resolve this issue.

test/number.spec.ts
ST-DDT: It returned NaN before, so I consider it a bugfix. I think it doesn't have to be mentioned explicitly.
(This is some kind of rewrite, so these tiny changes in undocumented behavior are to be expected)

src/internal/mersenne/mersenne.ts
matthewmayer: renamed it back and added a comment

matthewmayer:
i don't think this needs the breaking change label? Because this is only applied to `faker.number.float()` which is new in 8.0.0, not the old `faker.datatype.float()`

ST-DDT (dismissed):

xDivisionByZerox:
> i don't think this needs the breaking change label? Because this is only applied to `faker.number.float()` which is new in 8.0.0, not the old `faker.datatype.float()`

[You did call it a breaking change](https://github.com/faker-js/faker/pull/1675#discussion_r1089974605) yourself. Even if only in a private function. Please be consistent whether you want to support JS or not! If yes, this is a breaking change for JS users since the way mersenne returns values was changed.

Edit: Thats exactly why I hate mixing points of interest in PRs! This change could have been done without touching mersenne in any way.

src/internal/mersenne/mersenne.ts
```diff
@@ -7,13 +7,10 @@ import Twister from './twister';
  */
 export interface Mersenne {
   /**
-   * Generates a random number between `[min, max)`. The result is already floored.
-   *
-   * @param options The options to generate a random number.
-   * @param options.min The minimum number.
-   * @param options.max The maximum number.
+   * Generates a random float between `[0, 1)`.
+   * This method is called `next` so that it could be used as a [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol)
```
Shinigami92:
```suggestion
   * This method is called `next` so that it could be used as an [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol)
```

Shinigami92:
> > i don't think this needs the breaking change label? Because this is only applied to `faker.number.float()` which is new in 8.0.0, not the old `faker.datatype.float()`
> 
> [You did call it a breaking change](https://github.com/faker-js/faker/pull/1675#discussion_r1089974605) yourself. Even if only in a private function. Please be consistent whether you want to support JS or not! If yes, this is a breaking change for JS users since the way mersenne returns values was changed.
> 
> Edit: Thats exactly why I hate mixing points of interest in PRs! This change could have been done without touching mersenne in any way.

TBH I understand @xDivisionByZerox points here
I think it is good to mark this as breaking change anyway, because it affects a lot of other method calls inside faker like `color.rgb`.

I also understand that it might sometimes be better to make smaller PRs and split them, but for now I'm fine with the current state (except the english spelling in JSDoc), but then I can approve this 👍

matthewmayer:
There's not really a good definition of breaking change. You have to use your common sense as every change is a breaking change if your workflow is weird enough!

Anyway since this IS landing in a major release, whether it is breaking or not is moot. 

https://xkcd.com/1172/

Shinigami92 (approved):

ST-DDT (approved):

PR #1679, opened 2022-12-23
feat(helpers): allow empty string in fake

### LINKED ISSUE #1457: Allow passing empty string to fake()

Currently, I have to write special handling for empty strings because I cannot pass it to `fake()`.

### Suggested solution

Do not explicitly block empty strings and just return an empty string.

### Alternative

_No response_

### Additional context

JS has the `repeat()` method (and others). 
It is also possible to call it on an empty string or repeat it zero times.
It doesn't do anything, but it still possible.
`fake()` should behave the same.

ST-DDT:
Fixes #1457

- #1457

Allows `faker.helpers.fake('')` => `''`


xDivisionByZerox (dismissed):

ejcheng (dismissed):

test/helpers.spec.ts
```diff
@@ -96,6 +96,7 @@ describe('helpers', () => {
     });
 
     t.describe('fake', (t) => {
+      t.it('with empty string', '');
       t.it('with plain string', 'my test string').it(
```
Shinigami92: Please also add a blank string case (here, below or both), to ensure that the result will not get trimmed
This test is just there to ensure future-proof and nobody accidentally tries do e.g. optimize the function

ST-DDT: I will do that in a separate PR as that has nothing to do with this change: https://github.com/faker-js/faker/pull/1680

Shinigami92 (dismissed):

Shinigami92 (dismissed):

ejcheng (dismissed):

xDivisionByZerox:
Is this a breaking change? I (as a user) could have the following case:
```ts
function foo(unvalidatedInput: string) {
  try {
    return fake(unvalidatedInput);
  } catch {
    // I use the fact, that faker.helpers.fake() can't handle empty strings
    return myDefaultValue;
  }
}
```

Just want to get the labels correct and PR title.

xDivisionByZerox (dismissed):

ST-DDT:
If you are relying on the fact that the method throws on empty strings, then yes it is a breaking change.

But:
- We didn't even have a test for that before https://github.com/faker-js/faker/pull/1453 (merged today)
- We never specified that we would throw in case of an empty string
- We explicitly told the user not to use empty strings

aka it was explicitly undefined behavior before

I'm fine with marking this as breaking or not.

Shinigami92 (approved):

xDivisionByZerox (approved):
> We explicitly told the user not to use empty strings

That's a good point. Then it's not breaking IMO.

Issue #1686, opened 2022-12-25
Try turn on eslint `@typescript-eslint/padding-line-between-statements`


Shinigami92:
To ensure consistent readability across multiple contributors and prevent writing reviews like "pls add a blank line here" we want to try to turn on and configure https://typescript-eslint.io/rules/padding-line-between-statements/

# TODO

- [x] after blocks https://github.com/faker-js/faker/pull/1691
- [ ] before blocks https://github.com/faker-js/faker/pull/1703
- [ ] between cases (switch-case)

ST-DDT:
I'm not against adding the rule, but from their page:

![20221225_170510](https://user-images.githubusercontent.com/1579362/209474906-e9c413a5-e780-4c5b-887c-63d9341d3dc8.jpg)

Shinigami92:
@ST-DDT please note that this message appears for everyone visiting the page and I know why they are saying this
But please provide an alternative proposal for a solution for the issue described in the description of this issue or provide other reasons why you think it is important to explicitly double express this warning message from their site

I know that prettier doesn't have any kind of option to fulfill our needs but also know that I'm tired of writing review comments for this over and over again

ST-DDT:
As I said, I'm not against this rule.

I just dont know what this rule does exactly.

- The linked page does not contain an example
- I dont know how/whether this will impact our code
- I dont remember any instance where that was an issue.

I personally prefer PRs for "do something tiny" over issue + PRs.

Also in context of #1683

Shinigami92:
> The linked page does not contain an example

`@typescript-eslint` reflects many native rules and therefore for examples you need to look into the native one: https://eslint.org/docs/latest/rules/padding-line-between-statements

But this is basically the first paragraph in the [examples](https://typescript-eslint.io/rules/padding-line-between-statements/#examples)

> I personally prefer PRs for "do something tiny" over issue + PRs.

I'm okay with that you personally prefer this and cant do anything against your preference,
but I would like to drive Faker as an Open Source project where anyone is welcome to contribute
It is hard to contribute to a project when you don't know what is currently available to contribute
Issues make this more open to the community

In addition to that, I don't have always the time to instantly tackle something and therefore instead of making my own note block full with anykind of potential things I can do, I like to help myself with issues so I can better remember what I wanted to do all kind of things

EarthyOrange:
[...]

Issue #1636, opened 2022-12-06
Add `pnpm run preflight` command


ST-DDT:
### Clear and concise description of the problem

Currently there are many steps that should be run before committing, but have to be run manually one after the other.

- Formatting
- Linting?
- Update Test Snapshots
- Run Tests
- Generate Locales
- Generate Api-Docs

### Suggested solution

Introduce a command that runs all those commands using a single command. e.g. `pnpn run dev:complete`

ST-DDT:
Partially blocked by: #1635

matthewmayer:
as a relatively new contributor i would like this! Easy to forget a step

matthewmayer:
i would vote for a catchier name though. like `pnpm run checks` or `pnpm run preflight` or `pnpm run checklist` or something like that.

ejcheng:
> i would vote for a catchier name though. like `pnpm run checks` or `pnpm run preflight` or `pnpm run checklist` or something like that.

`preflight` sounds cool

xDivisionByZerox:
I'll move the `needs decision` label from here into the PR #1846 itself.

ST-DDT:
**Team Decision**

We will name this command `preflight`.

Shinigami92:
I started to use this command and when switching branches the `pnpm install` makes now a ton of sense 👍 
I like it

I just also do `git switch X` -> `git pull --rebase` -> `pnpm run preflight`

PR #1691, opened 2022-12-26
chore: turn on padding-line-between-statements

Shinigami92:
related to #1686 

---

For now I only turned on `{ blankLine: 'always', prev: 'block-like', next: '*' }`
This will add blank lines after blocks like `if`, `for` and `switch`
Then I executed `pnpm eslint . --fix`

I also sorted the `@typesscript-eslint/*` rules alphabetically

@faker-js/maintainers @faker-js/members Please let me know if you like this PR and if I should proceed or if we should merge this now and configure every further more strictness in separate PRs so the PR diff doesn't explode

I would like to additionally configure:
- `case` in switch blocks
- maybe a blank line before each `block-like` if somehow possible
- ...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,23 +34,6 @@ module.exports = defineConfig({
     '@typescript-eslint/ban-ts-comment': 'warn',
     '@typescript-eslint/consistent-type-imports': 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'error',
-    '@typescript-eslint/no-inferrable-types': [
-      'error',
-      { ignoreParameters: true },
-    ],
-    '@typescript-eslint/no-unsafe-argument': 'warn',
-    '@typescript-eslint/no-unsafe-assignment': 'off',
-    '@typescript-eslint/no-unsafe-call': 'off',
-    '@typescript-eslint/no-unsafe-member-access': 'off',
-    '@typescript-eslint/no-unsafe-return': 'warn',
-    '@typescript-eslint/restrict-template-expressions': [
-      'error',
-      {
-        allowNumber: true,
-        allowBoolean: true,
-      },
-    ],
-    '@typescript-eslint/unbound-method': 'off',
     '@typescript-eslint/naming-convention': [
       'error',
       {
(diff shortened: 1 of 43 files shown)


ST-DDT:
From looking at the code/changes I wouldnt know how to write it according to the requested style unless I run the fix command.

Shinigami92:
> From looking at the code/changes I wouldnt know how to write it according to the requested style unless I run the fix command.

Luckily eslint reports useful error messages like `Expected blank line before this statement.`

matthewmayer:
Will `pnpm run format` enforce this style?

Shinigami92:
> Will `pnpm run format` enforce this style?

No, sadly not, but `eslint` with `--fix` would do

matthewmayer:
Suggest to wait until https://github.com/faker-js/faker/issues/1636 is done? Then that could be part of the preflight checks. At the moment there are too many different commands to remember to run before committing.

EarthyOrange:
@Shinigami92 The "What about formatting?" linked page, from @ST-DDT 's screenshot in the linked issue, recommends configuring adding blank lines using a formatter instead of a linter. I see that this project uses the Prettier formatter; can you attempt following the recommendation and use the formatter to drive this formatting?

https://github.com/faker-js/faker/issues/1686#issuecomment-1365468038

Shinigami92:
> @Shinigami92 The "What about formatting?" linked page, from @ST-DDT 's screenshot in the linked issue, recommends configuring adding blank lines using a formatter instead of a linter. I see that this project uses the Prettier formatter; can you attempt following the recommendation and use the formatter to drive this formatting?
> 
> [#1686 (comment)](https://github.com/faker-js/faker/issues/1686#issuecomment-1365468038)

1. Please do not write the same thing on two different places
2. => https://github.com/faker-js/faker/issues/1686#issuecomment-1365471441

Shinigami92:
> Suggest to wait until #1636 is done? Then that could be part of the preflight checks. At the moment there are too many different commands to remember to run before committing.

I will remove the `on hold` because of two reasons:

1. currently it looks like no one is assigned and tempted to actually implement the `preflight` command
2. the `--fix` is not a thing we have included in out `scripts` because it can take up to several minutes to run and therefore should only be used with caution and when needed
   therefore this PR does add nothing new to our actual workflow and is just a configuration of out lint rules

Also not that I run manually the `--fix` once in this PR as I know why I run it and only need to run it once, further usage of eslint will show errors in our IDEs like for every other lint error before

matthewmayer:
Seems pretty harmless in practice. I tend to blindly accept any autocorrections suggested by prettier and eslint in VS Code so this wouldn't affect my workflow.

xDivisionByZerox (approved):
I have no problem with this kind of rule. It actually make the code more easy to read as a reviewer (at least for me) as I can consume the lines in chunks instead of the entire block at once.

ejcheng (approved):
I like this too, makes code much easier to read

ST-DDT (approved):

PR #1699, opened 2022-12-30
feat(airline): add airline module

matthewpetro:
Added a new module titled 'airline' for generating data pertaining to airlines.


src/locales/en/airline/airport.ts
```diff
@@ -0,0 +1,6074 @@
+export default [
```
ST-DDT, line 1: Please limit the list to 1000 entries.

src/modules/airline/index.ts
```diff
@@ -0,0 +1,110 @@
...
+   * @param options.aircraftType The aircraft type. Can be one of `narrowbody`, `regional`, `widebody`. Defaults to `narrowbody`.
+   *
+   * @example
+   * faker.airline.seat() // '22C'
+   * faker.airline.seat({ aircraftType: 'regional' }) // '7A'
+   * faker.airline.seat({ aircraftType: 'widebody' }) // '42K'
+   *
+   * @since 8.0.0
+   */
+
```
ST-DDT:
```suggestion
```

src/modules/airline/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  seat(
+    options: {
+      aircraftType?: typeof aircraftTypes[number];
```
ST-DDT: Maybe use a proper enum here and also export it as type similar to how we did it with person.SexType.

https://github.com/faker-js/faker/blob/7f7e4428893284efd2df49405e8c98e0eac4b190/src/modules/person/index.ts#L3-L8

src/modules/airline/index.ts
```diff
@@ -0,0 +1,110 @@
...
+      allowVisuallySimilarCharacters?: boolean;
+    } = {}
+  ): string {
+    const numerics = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
+    const visuallySimilarCharacters = ['0', 'O', '1', 'I', 'L'];
+    const excludedChars = [];
+    if (!options.allowNumerics) {
+      excludedChars.push(...numerics);
+    }
+    if (!options.allowVisuallySimilarCharacters) {
```
ST-DDT: Please change this to deconstruct this at the beginning of the method to make the default value more obvious:

````ts
const { allowNumerics = false, allowVisuallySimilarCharacters = false } = options;
````

File: src/modules/airline/index.ts [hunk 5]
```diff
@@ -0,0 +1,110 @@
...
+   * @since 8.0.0
+   */
+  recordLocator(
+    options: {
+      allowNumerics?: boolean;
+      allowVisuallySimilarCharacters?: boolean;
+    } = {}
+  ): string {
+    const numerics = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
+    const visuallySimilarCharacters = ['0', 'O', '1', 'I', 'L'];
```
ST-DDT: Maybe declare them outside of the class as consts.

src/modules/airline/index.ts
```diff
@@ -0,0 +1,110 @@
...
+   * @since 8.0.0
+   */
+  airportCode(): string {
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.airline.airport
+    );
+  }
+
+  /**
+   * Generates a random record locator.
```
ST-DDT: Please add a wiki or similar link to some extended documentation/definition.

src/locales/en/airline/airport.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Great suggestion, thanks. I've made the update.

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Thanks, that does help with clarity.

ejcheng (commented):
Please update the snapshot tests

src/locales/en/airline/airport.ts
ST-DDT: Just for future reference: How did you filter these?
By size/importance/random?

matthewpetro: I just randomly pulled them out of OpenFlight's data at https://raw.githubusercontent.com/jpatokal/openflights/master/data/airports.dat

README.md
```diff
@@ -78,6 +78,7 @@ The API covers the following modules:
 
 | Module   | Example                                       | Output                                                                                             |
 | -------- | --------------------------------------------- | -------------------------------------------------------------------------------------------------- |
+| Airline  | `faker.airline.recordLocator()`               | KIFRWE
```
Shinigami92: Lets use `airportCode` instead as example as this is more common and understandable

Shinigami92 (commented):
Please rebase and rerun format and fix the lint errors

src/modules/airline/index.ts
```diff
@@ -0,0 +1,116 @@
+import type { Faker } from '../..';
+
+export enum Aircraft {
+  Narrowbody = 'narrowbody',
+  Regional = 'regional',
+  Widebody = 'widebody',
+}
+
+export type AircraftType = `${Aircraft}`;
```
ST-DDT, line 15: Please add a new method that returns a random `AircraftType`:

````ts
aircraftType(): AircraftType {
   return faker.helpers.objectValue(Aircraft);
}
````

(Not tested)

File: README.md (hunk 7 above)
Comment (ST-DDT) [inline README.md]: Why did you choose this example for the readme?
As an "outsider", I don't know what this is and maybe a seat number is more self explanatory as an example.
What do you think?

File: src/modules/airline/index.ts [hunk 9]
```diff
@@ -0,0 +1,116 @@
...
+   * @since 8.0.0
+   */
+  airportCode(): string {
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.airline.airport
+    );
+  }
+
+  /**
+   * Generates a random record locator.
```
ST-DDT: Maybe add the link to the top (our other links are also like this). Also maybe add a less technical term for it so it is easier to search on the website.

```suggestion
   * Generates a random [record locator](https://en.wikipedia.org/wiki/Record_locator) (reservation code).
```

What do you think?

ST-DDT (commented):
Also please fix the lint error(s).

README.md
matthewpetro: OK, sounds good.

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Good suggestion. I added a little more explanatory text as well.

src/modules/airline/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    const row = this.faker.number.int({ min: 1, max: maxRow });
+    const seat = this.faker.helpers.arrayElement(allowedSeats);
+    return `${row}${seat}`;
+  }
+
+  /**
+   * Returns a random aircraft type.
+   *
+   * @example
+   * faker.airline.aircraftType() // AircraftType.Narrowbody
```
ST-DDT:
```suggestion
   * faker.airline.aircraftType() // 'narrowbody'
```

matthewmayer:
Looks cool.
Would it also be possible to add airlines (e.g. British Airways), codes (BA) and flight numbers (BA123)?

I wonder if `aviation` would be a better name than `airline`?

src/modules/airline/index.ts
matthewpetro: Done!

matthewpetro:
Thanks! It feels more accurate to call the module `airline` since the data relates to things that are specific to airline processes. A name like `aviation` would seem more appropriate for data that relates to the process of flying planes.

Airlines and airline codes are a good addition, but real flight numbers change too often to be a part of Faker. If someone had a use case that required them to have fake flight numbers, they could just generate random numbers in the range(s) they need.

matthewmayer:
> Thanks! It feels more accurate to call the module `airline` since the data relates to things that are specific to airline processes. A name like `aviation` would seem more appropriate for data that relates to the process of flying planes.
> 
> 


True, another option might be travel I could imagine this expanding to cover other travel related content in future (trains, hotels, car hire)

> 
> Airlines and airline codes are a good addition, but real flight numbers change too often to be a part of Faker. If someone had a use case that required them to have fake flight numbers, they could just generate random numbers in the range(s) they need.

Faker has lots of examples of methods which are basically trivial (say faker.commerce.price) or wrap other methods (like the PNR generation in this PR) but helpful when you are quickly trying to put up a mock up together for a web application in a specific vertical. So if we also add airline codes I think adding in a flight number too by adding a random number would be helpful. 

Could probably add just the most popular 100 airlines in the world, rather than every single one to avoid bloating the data too much.

matthewpetro:
> > Thanks! It feels more accurate to call the module `airline` since the data relates to things that are specific to airline processes. A name like `aviation` would seem more appropriate for data that relates to the process of flying planes.
> 
> True, another option might be travel I could imagine this expanding to cover other travel related content in future (trains, hotels, car hire)

That sounds good! I'll make that change.
> 
> > Airlines and airline codes are a good addition, but real flight numbers change too often to be a part of Faker. If someone had a use case that required them to have fake flight numbers, they could just generate random numbers in the range(s) they need.
> 
> Faker has lots of examples of methods which are basically trivial (say faker.commerce.price) or wrap other methods (like the PNR generation in this PR) but helpful when you are quickly trying to put up a mock up together for a web application in a specific vertical. So if we also add airline codes I think adding in a flight number too by adding a random number would be helpful.

That's a good point, I'll add that in.
> 
> Could probably add just the most popular 100 airlines in the world, rather than every single one to avoid bloating the data too much.

I'll work on adding airlines as well. I like the idea of adding the most popular airlines as well but it may be impossible to find open-source data that is in a consumable format and includes sortable data like number of passengers carried per year. How important is it for the definition data in Faker to be generated in a scripted and automated way? If it's not that important, I can just build a list of the top 100 airlines by hand based on web searches. If the definition data should be regeneratable via a script, then I could randomly pull 100 airlines out of a dataset that includes all airlines.

Thanks for the feedback!

matthewmayer:
This might be a good starting point (although it doesn't list the two digit codes)

https://www.flightglobal.com/download?ac=73318

For something like airlines I don't think there is a need to make it scriptable. It's supposed to be an illustrative list for making realistic sample data not a complete list. 

If reasonably big airlines are in the list they are probably less likely to rebrand or go bankrupt over time.

matthewmayer:
I'll let the maintainers pipe in on the module name too as I'm not a maintainer 😀

Shinigami92:
`travel` might be a good name, so we can later also add train or bus data and so on
and maybe hotel stuff

ST-DDT:
Do trains, busses and hotels follow some general rules per country?
Trains in Germany do. Busses not so much. Hotels not that I know.

matthewpetro:
I'm not sure about trains, buses, and hotels. I work for an airline, so have more knowledge in that space than in other travel-related domains.

Would it be safest to keep the module name as `airline` at this time and change it if someone wants to add other travel content in the future?

Shinigami92:
> Would it be safest to keep the module name as `airline` at this time and change it if someone wants to add other travel content in the future?

I will leave this decision to @ST-DDT

matthewmayer:
FWIW faker-ruby recently added a similar module as Travel
https://github.com/faker-ruby/faker/pull/2601

matthewmayer:
For hotels you could for example have random fake hotel names following certain patterns, random lists of amenities, room types, etc.

ST-DDT:
> I will leave this decision to @ST-DDT

We will discuss this in our next team meeting as this is something that came up before, but wasn't decided on finally.
(Ref: Submodules + Tree-Shaking; Cloud/AWS)

- faker.travel.airline.airport()
- travel.airline.airport()
- airline.airport()
- airport()

For now, I prefer `faker.airline.x()` as we don't have anything travel related yet.

-------

I do have another question yet though, do you only ever need the airport codes or do you sometimes also need the full name?
E.g. I assume most users don't directly know where `TXL` is and it will probably show up somewhere on the screen anyway.

So should we change the airport method to return an object instead?

````ts
interface Airport {
  name: string;
  code: string;
}
````

Usage: `faker.airline.airport().code`

**Comment by Shinigami92** [comment]
Would like to throw in that `...travel.airline...` was never on my "table"
I understand it as a replacement instead of `airline`, so it would be `faker.travel.airport()`

**Comment by matthewpetro** [comment]
> Would like to throw in that `...travel.airline...` was never on my "table" I understand it as a replacement instead of `airline`, so it would be `faker.travel.airport()`

That's my understanding as well.

For what it's worth, I prefer leaving the module name as `airline` and changing it later if non-airline travel features are added in the future.

**Comment by matthewpetro** [comment]
> I do have another question yet though, do you only ever need the airport codes or do you sometimes also need the full name? E.g. I assume most users don't directly know where `TXL` is and it will probably show up somewhere on the screen anyway.
> 
> So should we change the airport method to return an object instead?
> 
> ```ts
> interface Airport {
>   name: string;
>   code: string;
> }
> ```
> 
> Usage: `faker.airline.airport().code`

It really depends on the use case. Airport codes are needed for a lot of things (boarding pass barcodes are a prime example), but airport names can also be useful. I like the idea of returning an object with an airport name and code, but I think it would be good to reduce the number of airports in the definition file to just the top 100 airports so that the file isn't too large.

**Comment by matthewmayer** [comment]
Same for airlines if we add those. Code and name together would be useful.

**Comment by ST-DDT** [comment]
Yes, yes (please select some airports from around the globe), and yes

File: src/locales/en/airline/airlines.ts [hunk 11]
```diff
@@ -0,0 +1,102 @@
+// Top 100 airlines in 2019 according to https://www.flightglobal.com/download?ac=73318
```
ST-DDT: Which license applies to the original data?

src/locales/en/airline/airplane.ts
```diff
@@ -0,0 +1,243 @@
+// Generated from data at https://raw.githubusercontent.com/jpatokal/openflights/master/data/planes.dat
```
ST-DDT: These data are provided under the terms of the AGPL and incompatible with our/the MIT license.

src/locales/en/airline/airport.ts
```diff
@@ -0,0 +1,121 @@
+// Generated from data at https://raw.githubusercontent.com/jpatokal/openflights/master/data/airports.dat
```
ST-DDT: These data are provided under the terms of the AGPL and incompatible with our/the MIT license.

ST-DDT (changes requested):

src/locales/en/airline/airlines.ts
matthewpetro: The airline information came from https://github.com/jpatokal/openflights/blob/master/data/airlines.dat, which also is under the AGPL. Is data from Wikipedia compatible with the MIT license?

src/locales/en/airline/airport.ts
matthewmayer: I am very dubious that a list of public airports can be copyrighted at all. Basic facts can't be copyrighted.

matthewpetro: That is a good point. The data from OpenFlights probably does have some user edits in it, so the list in its entirety may be covered by the AGPL. But the airport names and codes are just facts that can be obtained from numerous sources.

I believe the same is true for the airplane and airline information.

What are everyone's thoughts?

Shinigami92: I'm slightly okay with copy them over from multiple sources, so it doesn't count as "we took it from exactly there"
Maybe we could also check the list by hand and only take the e.g. top 10 airports of each country? This would filter out such tiny airports that could only come in via a full list with values of slightly unknown small airports 🤔

matthewmayer: I agree in this case it's better to have a human compile a list (referring to multiple sources) rather than completely rely on one source of "worlds largest 100 airports" or something as a collection of facts can sometimes be copyright able even if the facts themselves are not.

matthewpetro: I've updated the airline, airplane, and airport definition files so they are not based on a single collection of information. They now contain a collection of large airlines and airports from around the world and a list of airplanes from major manufacturers.

matthewmayer:
For airports would there also be value in adding a "city" property? For example BKK might show as Bangkok rather than Suvarnabhumi Airport in a UI.

ST-DDT:
If we add a city, then what about the country?
I'm not familiar with that area of expertise so I dont know what data are actually needed.

matthewpetro:
Having the name and IATA code is suitable for many use cases and more data points could be added if users working in the airline domain find a need for them.

matthewmayer:
It's common to just show the city and not the country to avoid political arguments over say what country Taipei Taoyuan Airport is in.

src/definitions/airline.ts
```diff
@@ -0,0 +1,19 @@
+import type { Airline, Airplane, Airport } from '../modules/airline';
+import type { LocaleEntry } from './definitions';
+
+export type AirlineDefinitions = LocaleEntry<{
+  /**
+   * Some airline information
+   */
+  airline: readonly Airline[];
```
ST-DDT: I'm not sure whether the list/array itself should be readonly or only the elements therein.
What do the others think?

src/modules/airline/index.ts
```diff
@@ -0,0 +1,232 @@
...
+   * faker.airline.flightNumber({ length: 3 }) // '425'
+   * faker.airline.flightNumber({ length: { min: 2, max: 3 } }) // '84'
+   *
+   * @since 8.0.0
+   */
+  flightNumber(
+    options: {
+      length?: number | { min: number; max: number };
+      addLeadingZeros?: boolean;
+    } = {}
```
ST-DDT, line 264: Maybe add an option to add the airline code?

`airlineCode: boolean | string` (Defaults to `false`)

````ts
if (typeof airlineCode === 'boolean') {
  airlineCode = airlineCode ? this.airline().code : '';
}
````

File: src/modules/airline/index.ts [hunk 16]
```diff
@@ -0,0 +1,232 @@
...
+   * @since 8.0.0
+   */
+  aircraftType(): AircraftType {
+    return this.faker.helpers.objectValue(Aircraft);
+  }
+
+  /**
+   * Returns a random flight number. Flight numbers are always 1 to 4 digits long. Sometimes they are
+   * used without leading zeros (e.g.: American Airlines flight 425) and sometimes with leading
+   * zeros, often with the airline code prepended (e.g.: AA0425).
```
ST-DDT:
```suggestion
   * Returns a random flight number. Flight numbers are always `1` to `4` digits long. Sometimes they are
   * used without leading zeros (e.g.: `American Airlines flight 425`) and sometimes with leading
   * zeros, often with the airline code prepended (e.g.: `AA0425`).
```

src/modules/airline/index.ts
```diff
@@ -0,0 +1,232 @@
...
+   * @since 8.0.0
+   */
+  flightNumber(
+    options: {
+      length?: number | { min: number; max: number };
+      addLeadingZeros?: boolean;
+    } = {}
+  ): string {
+    const { length = { min: 1, max: 4 } } = options;
+    const { addLeadingZeros = false } = options;
```
ST-DDT: You can merge these into a single statement:

```suggestion
    const { length = { min: 1, max: 4 }, addLeadingZeros = false } = options;
```

(Might require formatting)

src/locales/en/airline/airlines.ts
```diff
@@ -0,0 +1,127 @@
+export default Object.freeze([
```
ST-DDT: Not sure about freezing the list. Maybe only freeze the entries therein?

ST-DDT (commented):
Thanks for replacing the lists with custom ones.

src/definitions/airline.ts
matthewpetro: That makes sense. The underlying array isn't exposed to a user, but the objects within it are, so they should be read only.

You're thinking something like this, right?
```
airline: Readonly<Airline>[];
```

matthewpetro:
> Thanks for replacing the lists with custom ones.

You're welcome!

src/locales/en/airline/airlines.ts
matthewpetro: That does make sense. As mentioned above, the array probably doesn't have to be frozen, but the entries should be read only.

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Thanks for catching that!

src/modules/airline/index.ts
matthewpetro: That feels to me like the function would be doing too much and not saving many lines of code for a user. I definitely can add it if anyone feels strongly about it, though.

src/definitions/airline.ts
ST-DDT: Not exactly, but close:

````ts
interface Airline {

.... readonly string;

}
````

If the array only contains readonly objects, then the readonly is lost when returning from the method because the method returns the type directly instead of the array elements, alternatively we could return `Readonly<Airline>` from the method, but there isn't any usecase for non-readonly `Airline` entries in the first place.

Comment (matthewpetro) [reply src/definitions/airline.ts]: Sounds good, I just made the change.

File: src/locales/en/airline/airlines.ts (hunk 18 above)
Comment (matthewpetro) [reply src/locales/en/airline/airlines.ts]: I removed the `Object.freeze()` calls.

File: src/definitions/airline.ts (hunk 14 above)
Comment (ST-DDT) [reply src/definitions/airline.ts]: You forgot to remove the `readonly` from the `AirlineDefinitions`. Could you please remove them from there as well?

File: src/modules/airline/index.ts (hunk 15 above)
Comment (matthewmayer) [reply src/modules/airline/index.ts:264]: I would like to see an option for the code. That way if say you are making a fake boarding pass it's easy to first get a random airline then pass the code into the flight number method to get an associated random flight number.

Comment (matthewmayer) [reply src/modules/airline/index.ts:264]: But personally I'd expect flight number to already include a random airline code by default.with option to omit it?

Comment (matthewmayer) [reply src/modules/airline/index.ts:264]: If the consensus is not to include codes in this method, it would be good to include that as a seperate example, e.g.


```js
`${faker.airline.airline().iataCode}${faker.airline.flightNumber()}` //AA123
```

src/definitions/airline.ts
matthewpetro: Thanks for catching that!

ST-DDT:
There still seems to be some a lint error.

matthewpetro:
> There still seems to be some a lint error.

Fixed!

src/definitions/airline.ts
```diff
@@ -0,0 +1,19 @@
+import type { Airline, Airplane, Airport } from '../modules/airline';
+import type { LocaleEntry } from './definitions';
+
+export type AirlineDefinitions = LocaleEntry<{
+  /**
+   * Some airline information
+   */
+  airlines: Airline[];
```
ST-DDT: We usually use singular names here, and I understand why you used that alternative:

https://github.com/matthewpetro/faker/blob/e38027d703f1601f87f80b946f7b671f43dc3a63/src/locales/en/airline/index.ts#L10-L11

 but I have to think about that first and discuss with others/the other team members.
Any suggestions?

matthewpetro: Yeah, using a plural here seemed to be a good way around the naming conflict, but if there's a better option I'd definitely be in favor of it.

matthewpetro: I haven't been able to find a good way to use a singular name here, other than renaming the module from `airline` to something else like `airtravel` or maybe just `travel` as had been discussed before.

I think it's best to just keep the plural in the definition here.

matthewmayer: FWIW there are currently two other examples of `faker.module.module`: `faker.vehicle.vehicle()` and `faker.image.image()`

matthewpetro: Thanks! I took a look at them and neither have locale definition files with names that match the module name, so they've been able to avoid the problem I'm running into.

matthewpetro: The only way I can think of to use a singular name is to update the generateLocales.ts script so that it can generate a locale index file that looks like:
```
import type { AirlineDefinitions } from '../../..';
import { default as airlines } from './airline';
import airplane from './airplane';
import airport from './airport';

const airline: AirlineDefinitions = {
  airline: airlines,
  airplane,
  airport,
};

export default airline;
```

Are there opinions as to whether I should make those updates or just leave things as they are?

src/modules/airline/index.ts
matthewpetro: I added an example to the documentation that showed how to combine the airline code and flight number, but I'd be happy to discuss further changes if there are opinions about it.

ST-DDT:
Please fix the failing tests.

matthewpetro:

Done!

src/modules/airline/index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+export interface Airplane {
+  /**
+   * The name of the airplane (e.g. `'Airbus A321'`).
+   */
+  readonly name: string;
+  /**
+   * The IATA code of the airplane (e.g. `'321'`).
+   */
+  readonly iataTypeCode: string;
```
Shinigami92, line 36:
```suggestion
  /**
   * The IATA code of the airplane (e.g. `'321'`).
   */
  readonly iataTypeCode: string;
```

What's the difference between an `iataCode` and an `iataTypeCode`

Should `Type` be reflected in the JSDoc?

src/definitions/airline.ts
ST-DDT: **Team decision**

- We will use `airline.airline` for now.
  - Do you want to adjust the `generate:locales` script to suffix the import with `_` in the case of the name collision (in a separate PR, that will be merged before this one)

https://github.com/faker-js/faker/blob/47b2cfc76b790647e398bf9883368a10b2ff5a68/scripts/generateLocales.ts#L183-L190

----------

We are still unsure regarding `airline.airline{ name, iataCode }` vs `airline.name: string[]`.
How important is having the correct airline code and name is for you vs randomly generated codes?

matthewmayer: I feel this is similar to `faker.science.chemicalElement()` where you want both the code and the full name

ST-DDT: That is exactly what we question right now.
Is it important to have these data together or is a little bit randomness fine?

If we have highly realistic data for most parts of the UI, but some mismatching data (e.g. atomic weight is generated randomly), then wouldn't this cause some kind of uncanny valley effect or an increase thereof?
If the data didn't match/or look random in the first place, then you would say "well it's test data".

So basically this is more a question regarding the scope of this library, than regarding this method in particular.
Do we provide lists (of names) or do we provide some kind of a database for all stuff.

src/modules/airline/index.ts
matthewpetro: `iataTypeCode` is what I should have named the field to begin with, as that's a more descriptive name. I've also updated the JSDocs to match, thanks for catching that!

src/definitions/airline.ts
matthewpetro: > So basically this is more a question regarding the scope of this library, than regarding this method in particular. Do we provide lists (of names) or do we provide some kind of a database for all stuff.

Hmm, that's a deep question. Which answer would make Faker more useful to users of the library?

matthewpetro: > **Team decision**
> 
> * We will use `airline.airline` for now.
>   
>   * Do you want to adjust the `generate:locales` script to suffix the import with `_` in the case of the name collision (in a separate PR, that will be merged before this one)
> 

No problem, I'll start on that.
> https://github.com/faker-js/faker/blob/47b2cfc76b790647e398bf9883368a10b2ff5a68/scripts/generateLocales.ts#L183-L190
> 
> We are still unsure regarding `airline.airline{ name, iataCode }` vs `airline.name: string[]`. How important is having the correct airline code and name is for you vs randomly generated codes?

For me right now, it isn't important. So I'm good leaving the data the way it is now or splitting it up into separate `airline.name: string[]` and airline.iataCode: string[]` functions.

ST-DDT: > > We are still unsure regarding `airline.airline{ name, iataCode }` vs `airline.name: string[]`. How important is having the correct airline code and name is for you vs randomly generated codes?
> 
> For me right now, it isn't important. So I'm good leaving the data the way it is now or splitting it up into separate `airline.name: string[]` and airline.iataCode: string[]` functions.

Lets wait for the decision in:

- #1733

ST-DDT:
```suggestion
  airline: Airline[];
```

matthewpetro: Good catch, thanks! I just updated the code.

ST-DDT (dismissed):
Looks good to me. Thanks for your contribution.

> Lets wait for the decision [regarding complex return objects] in:
>
> Better currencies #1733

matthewpetro:
> Looks good to me. Thanks for your contribution.

You're welcome. Thanks to everyone for the suggestions for improvements!

test/airline.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+        });
+        it('should return a random flight number with 2 to 4 digits and leading zeros', () => {
+          const flightNumber = faker.airline.flightNumber({
+            length: 3,
+            addLeadingZeros: true,
+          });
+          expect(flightNumber).toMatch(/^[0-9]{1,4}$/);
+        });
+      });
+    }
```
ejcheng:
```suggestion
        it('should return a random flight number with 2 to 4 digits and leading zeros', () => {
          const flightNumber = faker.airline.flightNumber({
            length: { min: 2, max: 4 },
            addLeadingZeros: true,
          });
          expect(flightNumber).toMatch(/^[0-9]{1,4}$/);
        });
      });
    }
```

ejcheng (changes requested):

test/airline.spec.ts
matthewpetro: Good catch, thank you!

ejcheng (dismissed):

ST-DDT (dismissed):

src/modules/airline/index.ts
```diff
@@ -0,0 +1,237 @@
...
+   * faker.airline.recordLocator({ allowNumerics: true }) // 'E5TYEM'
+   * faker.airline.recordLocator({ allowVisuallySimilarCharacters: true }) // 'ANZNEI'
+   * faker.airline.recordLocator({ allowNumerics: true, allowVisuallySimilarCharacters: true }) // '1Z2Z3E'
+   *
+   * @since 8.0.0
+   */
+  recordLocator(
+    options: {
+      allowNumerics?: boolean;
+      allowVisuallySimilarCharacters?: boolean;
```
Shinigami92: Please update your branch and then add JSDocs to these option params like in other modules

Here it would be something like

```suggestion
      /**
       * Whether to allow numeric characters.
       *
       * @default false
       */
      allowNumerics?: boolean;
      /**
       * Whether to allow visually similar characters such as '1' and 'I'.
       *
       * @default false
       */
      allowVisuallySimilarCharacters?: boolean;
```

src/modules/airline/index.ts
```diff
@@ -0,0 +1,237 @@
...
+   * @example
+   * faker.airline.seat() // '22C'
+   * faker.airline.seat({ aircraftType: 'regional' }) // '7A'
+   * faker.airline.seat({ aircraftType: 'widebody' }) // '42K'
+   *
+   * @since 8.0.0
+   */
+  seat(
+    options: {
+      aircraftType?: AircraftType;
```
Shinigami92, line 191: Same here

```suggestion
      /**
       * The aircraft type. Can be one of `narrowbody`, `regional`, `widebody`.
       *
       * @default 'narrowbody'
       */
      aircraftType?: AircraftType;
```

src/modules/airline/index.ts
```diff
@@ -0,0 +1,237 @@
...
+   * faker.airline.flightNumber({ addLeadingZeros: true, length: { min: 2, max: 3 } }) // '0624'
+   * faker.airline.flightNumber({ length: 3 }) // '425'
+   * faker.airline.flightNumber({ length: { min: 2, max: 3 } }) // '84'
+   *
+   * @since 8.0.0
+   */
+  flightNumber(
+    options: {
+      length?: number | { min: number; max: number };
+      addLeadingZeros?: boolean;
```
Shinigami92: and here (as prettier would complain anyways, I'm to lazy now to write this as a suggestion)

src/modules/airline/index.ts
matthewpetro: Done! Thanks for the suggestions.

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
matthewpetro: Done!

src/modules/airline/index.ts
```diff
@@ -0,0 +1,268 @@
...
+  readonly iataCode: string;
+}
+
+const numerics = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
+const visuallySimilarCharacters = ['0', 'O', '1', 'I', 'L'];
+const aircraftTypeMaxRows = {
+  regional: 20,
+  narrowbody: 35,
+  widebody: 60,
+};
```
ST-DDT, line 56: Maybe add `satisfies Record<AircraftType, number>`

src/modules/airline/index.ts
```diff
@@ -0,0 +1,268 @@
...
+const aircraftTypeSeats = {
+  regional: ['A', 'B', 'C', 'D'],
+  narrowbody: ['A', 'B', 'C', 'D', 'E', 'F'],
+  widebody: ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'J', 'K'],
+};
```
ST-DDT, line 61: Similarly here: `satifies Record<AircraftType, string[]>`

src/modules/airline/index.ts
```diff
@@ -0,0 +1,268 @@
...
+   * faker.airline.flightNumber({ addLeadingZeros: true, length: 2 }) // '0042'
+   * faker.airline.flightNumber({ addLeadingZeros: true, length: { min: 2, max: 3 } }) // '0624'
+   * faker.airline.flightNumber({ length: 3 }) // '425'
+   * faker.airline.flightNumber({ length: { min: 2, max: 3 } }) // '84'
+   *
+   * @since 8.0.0
+   */
+  flightNumber(
+    options: {
+      length?:
```
Shinigami92, line 246: The `length` param also needs a JSDoc 🙂

matthewpetro: I've added the JSDoc for that param as well. Should the parameters be documented using JSDocs on the parameters or using the `@param` tag? I currently have both in the code, but I'm not sure if I need to keep both.

Shinigami92: both

matthewpetro: Thanks!

src/modules/airline/index.ts
matthewpetro: Would defining the type of `aircraftTypeMaxRows` be better?
```
const aircraftTypeMaxRows: Record<AircraftType, number> = {
```

ST-DDT (approved):

ejcheng (dismissed):

src/locales/en/index.ts
```diff
@@ -28,6 +29,8 @@ import word from './word';
 
 const en: LocaleDefinition = {
   title: 'English',
+  separator: ' & ',
```
ST-DDT: This line got removed from the next branch.
Could you please run:

````sh
pnpm run generate:locales
````

to fix it?

ST-DDT (changes requested):

src/locales/en/index.ts
matthewpetro: Done!

ST-DDT (approved):

Shinigami92:
Should this be merged before we decide if we still want to have "complex" objects in definitions? Or should we wait a bit (could be around 3 weeks due to no meetings)

Here we have `iataCode`s which can be put into global definitions, but airport names are in english and could be translated to their individual locales 🤔 
_I feel not really happy with this design, but it's not an issue by this module itself but faker itself right now_

ST-DDT:
Lets get this merged, and decide later.
We have all the data we need here and can strip unwanted/restructure them later easily.

Shinigami92 (approved):

ejcheng (approved):

Shinigami92:
@ST-DDT there is one unresolved discussion
please reevaluate it and decide if we can merge or need to do further actions on that

ST-DDT:
Thanks for your contribution.

matthewpetro:
You're welcome!

PR #1702, opened 2022-12-30
feat: datatype.number and number `not` parameter introduced

amaster507:
Introduced `not` parameter for `int` and `float` number generators.

See: https://discord.com/channels/929487054990110771/929487054990110777/1058484035875258419

I had a recurring need to generate numbers with a provided parameter to signify what number I did _not_ want generated.

To prevent endless loops in rare cases with limited runway between min and max parameters, I included a try count with a maximum of hardcoded 100 tries if the number generated is the number _`not`_ wanted. If this were to occur, and error will be thrown as documented.

### DIFF
--- a/src/modules/datatype/index.ts
+++ b/src/modules/datatype/index.ts
@@ -22,9 +22,12 @@ export class DatatypeModule {
    * @param options Maximum value or options object.
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `min + 99999`.
+   * @param options.not A number that the generated number cannot be equal to. Defaults to `undefined`.
    * @param options.precision Precision of the generated number. Defaults to `1`.
    *
    * @throws When options define `max < min`.
+   * @throws When options define `max === min === Math.round(not)`.
+   * @throws When tries too many times to generate a number not equal to `options.not`.
    *
    * @see faker.number.int() for the default precision of `1`
    * @see faker.number.float() for a custom precision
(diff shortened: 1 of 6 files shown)


Shinigami92:
Somehow it feels like we could try to archive the same in a different/better way
e.g. with a `helpers` function that could wrap such a call
This way we could have the same also for everything else, like a `faker.person.firstName` call, where the return value can be excluded and then a new try will be made

amaster507:
Not against that idea either. This was my first attempt at this. But if there was a way to apply it more globally that would be helpful too.

ST-DDT:
IMO banning a single value isnt that useful usually I omit certain values or values that match or dont match a certain criteria.

I think a dedicated method might be useful here.

E.g. `faker.helpers.exclude(source: () => T, filter: T[] | Set<T> | (T) => boolean, options: {...})`

Not sure about the exact signature yet.

Shinigami92:
> IMO banning a single value isnt that useful usually I omit certain values or values that match or dont match a certain criteria.
> 
> I think a dedicated method might be useful here.
> 
> E.g. `faker.helpers.exclude(source: () => T, filter: T[] | Set<T> | (T) => boolean, options: {...})`
> 
> Not sure about the exact signature yet.

Yes, this is leading in the right direction as equally as I thought about it with my first comment
My only concern right now would be: is this based on mersenne/seed? Theoretically it looks like it's just a function that could be used from outside faker, could it?

Currently in my head it would look something like this (_not tested at all_):

```js
function exclude(source, filter, options) {
  let success = false
  let value

  do {
    value = source(options)
    success = filter(value)
  } while (!success)

  return value
}
```

but as you can see, this is not bound to faker / merseene / seed at all and can be used for anything in JS world

ST-DDT:
**Team decision**

We decided that we don't want functions that are not seed dependent in our library.
These function are very easy to implement yourself: https://github.com/faker-js/faker/pull/1702#issuecomment-1368203928

Issue #1340, opened 2022-09-06
Remove locale switching support


ST-DDT:
### Clear and concise description of the problem

The locale switching support adds a lot of overhead to the faker instance.
Especially the default instance is huge.

### Suggested solution

- Remove locale switching support from Faker.
- The locales have to be set at creation time `new Faker([ en ])`
- It is possible to provide a single instance as well as a list of locales, that will be merged immediately
- The locale merging behavior isn't changed to the version as it is now, but is exported for users to use themselves (e.g. in build scripts)
- The locale data can be accessed from both cjs and esm using simple imports (Maybe covered by #642)
- The main faker instance is only `en`
- The documentation contains a hint regarding the use of other locales (Maybe handled as part of #1322)
- The provided faker instances use a proper locale hierarchy `fr_CH -> fr -> en`
- Global data such as smiley/iban/mime-types will be moved to their own locale (TBD: global or individual)

xDivisionByZerox:
> Global data such as smiley/iban/mime-types will be moved to their own locale (TBD: global or individual)

Why cant they be scoped to the module?

ST-DDT:
They could be. If we drop support for replacing them.
However, that would be another breaking change that should be it's own ticket for better visibility.
Do we have that already somewhere? I remember discussing this before.

Shinigami92:
I'm **somewhat** interested to be champion of this, but have some other issues I already provide higher prio from my work-amount
So if someone else claims this before I can start, I'm also fine with that

matthewmayer:
Just to clarify, if you're currently using a single locale like this:
```js
import { faker } from '@faker-js/faker';
// or with CJS, 
// const {faker} = require("@faker-js/faker")
faker.locale = "de" //or setLocale
console.log(faker.word.noun()) //Glücksspiel
```
The recommended migration will be to switch to
```js
import { faker } from '@faker-js/faker/locale/de';
// or with CJS, 
// const {faker} = require("@faker-js/faker/locale/de")
console.log(faker.word.noun()) //Glücksspiel
```
?

Shinigami92:
We are not sure yet if we still want to support `import { faker } from '@faker-js/faker/locale/de'` in the long run, but yes.

Our actual plan is to encourage using one of these options (_might be slightly different in real world after we done it_):

```ts
import { faker, fakerDE, localeDE, ... } from '@faker-js/faker'

fakerDE.word.noun()

const myFaker = new Faker({ locale: [localeDE, ...] })
[...]

Issue #893, opened 2022-04-30
Introduce a way to get details about missing locale data when using them


ST-DDT:
### Clear and concise description of the problem

Currently many methods look like this:

````ts
foobar(): string {
    return this.faker.random.arrayElement(this.faker.definitions.foo.bar);
}
````

However, the current locale does not have a value for it, then it will return `'a' | 'b' | 'c'`.
It would be nice, if the method would throw an error, with details about missing data.
It should still be possible to check whether the data exists without throwing and catching an error.

This kind of blocks the removal of the default from `random.arrayElement()`

### Suggested solution

Change the `faker.definitions` to throw an error, if the data are absent.
Introduce `faker.extensions` that would return undefined (the current `faker.definitions` behavior).

(Not sure about the name)

### Alternative

Check the data before using it each time. This reduces the code readability a lot.

### Additional context

IMO `faker.definitions` should change its behavior, instead of adding a new field for it, because I assume, the default expectation is that the data are present. Most of our methods wouldn't change either. Only some of them would, where the data are explicitly optional.

Shinigami92:
Just to prioritize some issues, I would see
- https://github.com/faker-js/faker/issues/823

should be handled before this issue

ST-DDT:
We might also need a way to mark/handle explicitly missing data.
https://github.com/faker-js/faker/pull/1665#discussion_r1048345332 introduces empty arrays for that purpose.

ST-DDT:
**Team decision**

If faker.definitions.category.entry is not defined we will throw an error.

xDivisionByZerox:
What blocks us from "simply" making all definitions required? If we currently don't have any they could be filled with an empty array. Then the next step would be to throw an error in `helpers.arrayElement` when an empty array is passed.

Is that the desired behavior or did I misunderstood the issue?

matthewmayer:
The current behavior is to fallback to the fallback locale ie English en which is defined for nearly everything. 

In which cases would we want to explicitly throw an error?

xDivisionByZerox:
> The current behavior is to fallback to the fallback locale ie English en which is defined for nearly everything.

Yes, that is correct. But this happens when accessing the actual locale data via the `Proxy`:
https://github.com/faker-js/faker/blob/f9e5ba706aa23ce958b9f82b09814a98ce830515/src/faker.ts#L191-L196

Which could simply be replaced by checking against the `entries` length instead of its existence.

> In which cases would we want to explicitly throw an error?

[...]

Issue #1677, opened 2022-12-21
Change file structure for gendered lists (names, prefixes etc)


matthewmayer:
_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/547#issuecomment-1240937734_

Change data-structure:

name.female_first_name ->person.first_name.female

Example:
````ts
type Gendered = {
  female: string[]
  male: string[]
  generic: string[]
};
````

matthewmayer:
Also noting that the first names for the following languages are not split by gender currently, someone with better knowledge could check these:

~de_CH~
~es_MX~
ka_GE
~ko~
~ne~
~zh_TW~

Shinigami92:
> Also noting that the first names for the following languages are not split by gender currently, someone with better knowledge could check these:
> 
> de_CH
> es_MX
> ge
> ko
> ne
> zh_TW

For now I would say put them in `person.first_name.generic` (?) or `person.first_name.index` (?)
So with this change they are not anymore in the file but the folder

ST-DDT:
I prefer `generic` over `index`. (Could you explain why you suggested `index`? I'm unable to guess your reasons/thoughts here.)

matthewmayer:
It would be great to make a list of contributors (even if they are not that regular) who know various languages so we can ask their help for locale specific issues. So if say there's an issue with Korean we can loop them in.

Shinigami92:
> I prefer generic over index. (Could you explain why you suggested index? I'm unable to guess your reasons/thoughts here.)

No I cannot => use `generic`

matthewmayer:
@Stichoza it looks like you originally added the data for Georgian, i wondered if you might be able to help us split the https://github.com/faker-js/faker/blob/next/src/locales/ge/person/first_name.ts file into male and female? This will help the `ge` locale work correctly.

https://github.com/faker-js/faker/commits?author=Stichoza

matthewmayer:
I think this will also require changes to faker.helpers.fake to get it to work with patterns like {{person.first_name}}?

ST-DDT:
No, it doesnt or at least we wont implement it that way. Instead we will change the patterns accordingly.e.g. person.first_name.generic/...

But thanks anyway, this is a good reminder for the actual implementation since we didnt write that anywhere.

matthewmayer:
[...]

PR #1707, opened 2023-01-02
feat(locales): add mergeLocales utility

ST-DDT:
Part 1 of
- #1340
- #893

This PR provides a method that merges multiple locales into a single one.

Future PRs will:
- remove the fallback logic from the `Faker` class (#1340)
- change the way the locale is set in the Faker instances (#1340)
- actually throw for missing locale entries (#893)


src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+ * The first locale that provides a entry for a category will be used for that.
+ * Mutating the category entries in the returned locale will also mutate the entries in the respective source locale.
+ *
+ * @param locales The locales to merge.
+ * @returns The newly merged locale.
+ *
+ * @example
+ * const locale = mergeLocales([ de_CH, de, en ]);
+ */
+export function mergeLocales(locales: LocaleDefinition[]): LocaleDefinition {
```
ST-DDT, line 15: Should I return immediately if only a single locale is passed?

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Merges the given locales into one locale.
+ * The locales are merged in the order they are given.
+ * The first locale that provides a entry for a category will be used for that.
+ * Mutating the category entries in the returned locale will also mutate the entries in the respective source locale.
+ *
+ * @param locales The locales to merge.
+ * @returns The newly merged locale.
+ *
+ * @example
+ * const locale = mergeLocales([ de_CH, de, en ]);
```
ST-DDT: Or should I use `de_CH_with_fallbacks` as a variable name here?

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
+import type { LocaleDefinition } from '..';
+
+/**
+ * Merges the given locales into one locale.
+ * The locales are merged in the order they are given.
+ * The first locale that provides a entry for a category will be used for that.
```
Shinigami92:
```suggestion
 * The first locale that provides an entry for a category will be used for that.
```

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Mutating the category entries in the returned locale will also mutate the entries in the respective source locale.
```
Shinigami92: category => definition?

src/utils/merge-locales.ts
Shinigami92: Could be a good idea 🤔

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ * const locale = mergeLocales([ de_CH, de, en ]);
+ */
+export function mergeLocales(locales: LocaleDefinition[]): LocaleDefinition {
+  const merged: LocaleDefinition = {} as LocaleDefinition;
+
+  for (const locale of locales) {
+    for (const key in locale) {
+      if (merged[key] === undefined) {
+        if (typeof locale[key] === 'object') {
```
Shinigami92, line 21: Attention: maybe we need to use `merged[key] == null` here (nullish-check)

Also `typeof locale[key] === 'object'` => `locale[key]` can still be `null` as `typeof null === 'object'`
Luckily `{ ...null }` is still `{}`, but we should handle it anyways

test/utils/merge-locales.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(merged).toEqual({
+      title: 'a',
+      animal: { cat: ['b'] },
+      color: { human: ['c'] },
+      location: { city: ['a'] },
+      person: { first_name: ['a'], last_name: ['b'] },
+    });
+  });
+});
```
Shinigami92, line 59: We need a test where a prop holds a nested object, like:
```js
{
  title: 'a',
  abc: { def: { ghi: 'j', klm: ['n'] } },
}
```

src/utils/merge-locales.ts
ST-DDT: We need clearer definitions for our terms.

IMO:

- Definitions: typeof faker.definitions....
- Category/Entry: faker.definitions.category.entry
- Module/Method: faker.module.method

src/utils/merge-locales.ts
ST-DDT: I will think about it and create a test for it.

test/utils/merge-locales.spec.ts
ST-DDT: I will extend our existing tests accordingly.

ST-DDT: Done

src/utils/merge-locales.ts
ST-DDT: How about using `null` for explicitly marking an entry as absent (instead of `[]`) and thus preventing falling back to another locale?
That way it can be used for all entries the same (simple strings, arrays and complex objects).

src/utils/merge-locales.ts
xDivisionByZerox: I don't think that is required as the performance gain is marginal.

src/utils/merge-locales.ts
xDivisionByZerox: > How about using null for explicitly marking an entry as absent (instead of []) and thus preventing falling back to another locale?

Do I understand that correctly?

```ts
const localeOne: LocaleDefinition = {
  location: {
    city: null,
    [...],
  },
  [...],
};

const localeTwo: LocaleDefinition = {
  location: {
    city: [
      // some cities
    ],
    [...],
  },
  [...],
};

const merged: LocaleDefinition = mergeLocales([localeOne, localeTwo]);
```

In this case `merged` will not have `location.city` even tho `localeTwo` has some values for it. 
Just want to understand correctly.

test/utils/merge-locales.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      person: { firstName: ['c'] },
+      finance: { credit_card: {} },
+    };
+
+    const merged = mergeLocales([locale1, locale2, locale3]);
+
+    expect(merged).toEqual({
+      title: 'a',
+      person: { firstName: ['a'] },
+      finance: { credit_card: { visa: ['a'] } },
```
xDivisionByZerox, line 28: Wouldn't the expected `finance` property be `{ credit_card: { visa: ['a'], mastercard: ['b'] } }`?

src/utils/merge-locales.ts
Shinigami92: This actually could work 👍 
It's also much more explicit then just defining an empty array, as this feels more like someone forgot to add some values

test/utils/merge-locales.spec.ts
ST-DDT: No, because the locale's category's entries won't be merged otherwise we would have to merge city names as well.
(This is also in line with our current implementation)

Ref: faker.definitions.category.entry

Missing categories (finance) and entries(credit_card) are added, but below that everything (visa/mastercard) gets replaced/skipped.

src/utils/merge-locales.ts
ST-DDT: Mhh, currently it isn't possible to set `null` there at all.

`{ animal: null }`

> Type 'null' is not assignable to type 'AnimalDefinitions | undefined'.

`{ animal: { cat: null } }`

> Type 'null' is not assignable to type 'string[] | undefined'.

test/utils/merge-locales.spec.ts
ST-DDT: That's one of the reasons why we want to do https://github.com/faker-js/faker/issues/1677 in the first place.

src/utils/merge-locales.ts
xDivisionByZerox: So we figure that out in the future or in the next meeting?

ST-DDT: Yes, I think changing one thing at a time would be the best.

But I could also use the nullish check for now, if you think that would be better.

xDivisionByZerox: Currently, none of our definitions can be `null` due to typescript, If I'm not mistaken. Correct? If that's the case, we should probably "propose" bullish definition in a new PR entirely.

xDivisionByZerox (dismissed):

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+          merged[key] = { ...locale[key] };
+        } else {
+          merged[key] = locale[key];
+        }
+      } else {
+        if (typeof locale[key] === 'object') {
+          merged[key] = { ...locale[key], ...merged[key] };
+        } else {
+          // Do nothing
```
Shinigami92:
```suggestion
          // Do not override primitive, because [please fill this]
```

test/utils/merge-locales.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  it('should overwrite locales', () => {
+    const locale1: LocaleDefinition = {
+      title: 'a',
+      person: { firstName: ['a'] },
+      finance: { credit_card: { visa: ['a'] } },
+    };
+    const locale2: LocaleDefinition = {
+      title: 'b',
+      person: { firstName: ['b'] },
+      finance: { credit_card: { mastercard: ['b'] } },
```
Shinigami92, line 15: please add a case like `finance: { currency: { 'Netherlands Antillian Guilder': { code: 'ANG', symbol: 'ƒ' }, ... } }`

src/utils/merge-locales.ts
ST-DDT: IMO this doesn't add any value, because there is no real reason to this, besides "there is nothing to merge here".
Aka "do not merge because there isn't a way/reason to merge it".
(Please note that this is the **merge** part of the code, and not the **set** part, there is no **overwrite** part)

test/utils/merge-locales.spec.ts
ST-DDT: IMO there is no need for this. The source code has two loops (`locale` and `category`) and we have a test that ensures we don't extend data in `entry`s directly (which would be the third loop). So any test that checks the nested data inside the `entry.data`s (4th+ loop) doesn't add any value/coverage.

src/utils/merge-locales.ts
Shinigami92: https://medium.com/codex/clean-code-comments-833e11a706dc#:~:text=Explanation%20of%20Intent

It is best practice to explain a **why** instead of only a **what**

As I was looking into the nested `if` code I needed some minutes to understand what's going on and/or why decisions were made
Please do not always look from your own perspective but help the next developer how could potentially cross your code

xDivisionByZerox: Can't we simply get completely rid of the `else` block here?

Shinigami92: That was the other option I was thinking of

ST-DDT: I edited the comment (because it seems to be required to be able to understand what the code does), but I created https://github.com/faker-js/faker/pull/1731 + one future PR to make the ` if (typeof xyz[key] === 'object') {` check obsolete.

src/utils/merge-locales.ts
```diff
@@ -0,0 +1,37 @@
...
+        if (typeof locale[key] === 'object') {
+          merged[key] = { ...locale[key] };
+        } else {
+          merged[key] = locale[key];
+        }
+      } else {
+        if (typeof locale[key] === 'object') {
+          merged[key] = { ...locale[key], ...merged[key] };
+        } else {
+          // Primitive values cannot be not merged
```
Shinigami92: Is the double negotiable really necessary?
I'm even not sure if this is correct, but a typo 🤨

ST-DDT:
```suggestion
          // Primitive values cannot be merged
```

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1708, opened 2023-01-02
feat(number): add binary and octal random number generation

pladreyt:
This PR implements binary and octal random number generation.

**Issue:** https://github.com/faker-js/faker/issues/184

PR that implements this feature in string module: https://github.com/faker-js/faker/pull/1710


ST-DDT (commented):
Could you please move the methods above the hex method so that the methods are a bit more in order.
Binary(2), Octal(8), Hex(16)

Should we also create these methods in the stri g module to generate them in a specific length?

test/number.spec.ts
```diff
@@ -348,5 +350,71 @@ describe('number', () => {
         );
       });
     });
+
+    describe('binary', () => {
+      it('generates single binary character when no additional argument was provided', () => {
+        const binary = faker.number.binary();
+        expect(binary).toBeTypeOf('string');
+        expect(binary).toHaveLength(1);
+        expect(/^[01]+$/.test(binary)).toBe(true);
```
ST-DDT: Please use expect(binary).toMatch here and elsewhere.

test/number.spec.ts
```diff
@@ -348,5 +350,71 @@ describe('number', () => {
...
+
+    describe('octal', () => {
+      it('generates single octal character when no additional argument was provided', () => {
+        const octal = faker.number.octal();
+        expect(octal).toBeTypeOf('string');
+        expect(octal).toHaveLength(1);
+        expect(/^[0-7]+$/.test(octal)).toBe(true);
+      });
+
+      it('generates a random octal string', () => {
```
ST-DDT: The title is duplicate and doesnt reflect the purpose of the test.
Can you please adjust it to something like.
Generates a random octal value with a custom max

test/number.spec.ts
```diff
@@ -348,5 +350,71 @@ describe('number', () => {
...
+
+    describe('binary', () => {
+      it('generates single binary character when no additional argument was provided', () => {
+        const binary = faker.number.binary();
+        expect(binary).toBeTypeOf('string');
+        expect(binary).toHaveLength(1);
+        expect(/^[01]+$/.test(binary)).toBe(true);
+      });
+
+      it('generates a random binary string', () => {
```
ST-DDT: Same here

test/number.spec.ts
pladreyt: Sure!

test/number.spec.ts
pladreyt: True. Will update

pladreyt:
> Could you please move the methods above the hex method so that the methods are a bit more in order. Binary(2), Octal(8), Hex(16)
> 
> Should we also create these methods in the stri g module to generate them in a specific length?

Sure, will reorder the methods. Should I create another PR to add them in the string module as well ?

ST-DDT:
> Should I create another PR to add them in the string module as well ?

Yes, I think that would be the easiest solution.
The PR that gets merged later will have to add `@see` links to both though. I leave the decision to you whether you want to create the other PR now or later.

test/number.spec.ts
```diff
@@ -11,7 +11,9 @@ describe('number', () => {
   seededTests(faker, 'number', (t) => {
     t.describeEach(
       'int',
-      'hex'
+      'hex',
+      'binary',
+      'octal'
```
ST-DDT: Please adjust the methods here as well.

test/number.spec.ts
```diff
@@ -240,6 +242,72 @@ describe('number', () => {
       });
     });
 
+    describe('binary', () => {
+      it('generates single binary character when no additional argument was provided', () => {
+        const binary = faker.number.binary();
+        expect(binary).toBeTypeOf('string');
+        expect(binary).toHaveLength(1);
+        expect(binary).toMatch(/^[01]+$/);
```
ST-DDT:
```suggestion
        expect(binary).toMatch(/^[01]$/);
```

test/number.spec.ts
```diff
@@ -240,6 +242,72 @@ describe('number', () => {
...
+        }).toThrowError(`Max ${max} should be greater than min ${min}.`);
+      });
+    });
+
+    describe('octal', () => {
+      it('generates single octal character when no additional argument was provided', () => {
+        const octal = faker.number.octal();
+        expect(octal).toBeTypeOf('string');
+        expect(octal).toHaveLength(1);
+        expect(octal).toMatch(/^[0-7]+$/);
```
ST-DDT:
```suggestion
        expect(octal).toMatch(/^[0-7]$/);
```

Shinigami92 (dismissed):
Please solve @ST-DDT suggestions, but already a good-to-go from my side

test/number.spec.ts
pladreyt: Sure !

test/number.spec.ts
pladreyt: > > Should I create another PR to add them in the string module as well ?
> 
> Yes, I think that would be the easiest solution. The PR that gets merged later will have to add `@see` links to both though. I leave the decision to you whether you want to create the other PR now or later.

Will create the other PR probably later today and add the `@see` links

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #1709, opened 2023-01-02
chore(test): skip console messages containing 'deprecated'

matthewmayer:
running `pnpm run test` logs an enormous number of deprecation warnings. This makes it hard to review the output locally and in CI. 
https://github.com/vitest-dev/vitest/issues/1700 shows a way to filter out specific logs, in this case anything containing "deprecated".


ST-DDT:
Please fix the lint errors.

ejcheng (dismissed):

ST-DDT (dismissed):
This seems to improve the execution speed in CI (3.5mins -> 2.5mins), but it doesn't have that much of an impact for me locally.

The logs have been reduced from almost 30k logs to ~200 which is good, but I'm not sure this a good long term solution.

vite.config.ts
```diff
@@ -9,5 +9,8 @@ export default defineConfig({
       reporter: ['clover', 'cobertura', 'lcov', 'text'],
       include: ['src'],
     },
+    onConsoleLog(log) {
+      if (log.includes('deprecated')) return false;
+    },
```
Shinigami92:
```suggestion
    onConsoleLog(log, type) {
      if (
        type === 'stderr' &&
        log.includes('[@faker-js/faker]:') &&
        log.includes('deprecated')
      ) {
        return false;
      }
    },
```

Shinigami92 (commented):
Tested before and after my changes
Both still don't really work better on my machine
The tests are absolutely bloated and run into timeouts

matthewmayer:
I don't know enough about how the tests works to figure out why they are so slow, however this change at least makes it possible to read the log output on GitHub.com easily.

Shinigami92 (approved):

Shinigami92:
> [...], however this change at least makes it possible to read the log output on GitHub.com easily.

This is indeed a huge benefit 👌

ejcheng (approved):

ST-DDT (approved):

PR #1710, opened 2023-01-02
feat(string): add binary and octal random string generation

pladreyt:
This PR implements binary and octal random string generation.

Issue: https://github.com/faker-js/faker/issues/184

PR that implements this feature in number module: https://github.com/faker-js/faker/pull/1708

### DIFF
--- a/src/modules/string/index.ts
+++ b/src/modules/string/index.ts
@@ -256,8 +256,6 @@ export class StringModule {
    * faker.string.binary({ length: { min: 5, max: 10 } }) // '0b11101011'
    * faker.string.binary({ prefix: '0b' }) // '0b1'
    * faker.string.binary({ length: 10, prefix: 'bin_' }) // 'bin_1101011011'
-   * faker.string.binary({ prefix: '' }) // '1'
-   * faker.string.binary({ length: 10, prefix: '0b' }) // '0b1101011011'
    *
    * @since 8.0.0
    */
@@ -297,8 +295,6 @@ export class StringModule {
    * faker.string.octal({ length: { min: 5, max: 10 } }) // '0o15263214'
    * faker.string.octal({ prefix: '0o' }) // '0o7'
    * faker.string.octal({ length: 10, prefix: 'oct_' }) // 'oct_1542153414'
-   * faker.string.octal({ prefix: '' }) // '2'
-   * faker.string.octal({ length: 10, prefix: '0o' }) // '0o1526216210'
    *
    * @since 8.0.0
    */
--- a/src/modules/string/index.ts
+++ b/src/modules/string/index.ts
@@ -242,6 +242,93 @@ export class StringModule {
   }
 
   /**
+   * Returns a [binary](https://en.wikipedia.org/wiki/Binary_number) string.
+   *
+   * @param options The optional options object.
+   * @param options.length The number or range of characters to generate after the prefix. Defaults to `1`.
+   * @param options.prefix Prefix for the generated number. Defaults to `'0x'`.
+   *
+   * @example
+   * faker.string.binary() // '0x1'
+   * faker.string.binary({ length: 10 }) // '0x1101011011'
+   * faker.string.binary({ length: { min: 5, max: 10 } }) // '0x11101011'
+   * faker.string.binary({ prefix: '0x' }) // '0x1'
+   * faker.string.binary({ length: 10, prefix: '#' }) // '#1101011011'
+   * faker.string.binary({ prefix: '' }) // '1'
+   * faker.string.binary({ length: 10, prefix: '0x' }) // '0x1101011011'
+   *
+   * @since 8.0.0
+   */
+  binary(
+    options: {
+      length?: number | { min: number; max: number };
+      prefix?: string;
+    } = {}
+  ): string {
+    const { prefix = '0x' } = options;
+    const length = this.faker.helpers.rangeToNumber(options.length ?? 1);
+    if (length <= 0) {
+      return prefix;
+    }
+
+    let binaryString = '';
+
+    for (let i = 0; i < length; i++) {
+      binaryString += this.faker.helpers.arrayElement(['0', '1']);
+    }
+
+    return `${prefix}${binaryString}`;
+  }
+
+  /**
+   * Returns an [octal](https://en.wikipedia.org/wiki/Octal) string.
+   *
+   * @param options The optional options object.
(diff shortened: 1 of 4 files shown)


src/modules/string/index.ts
```diff
@@ -241,6 +241,93 @@ export class StringModule {
     ).join('');
   }
 
+  /**
+   * Returns a [binary](https://en.wikipedia.org/wiki/Binary_number) string.
+   *
+   * @param options The optional options object.
+   * @param options.length The number or range of characters to generate after the prefix. Defaults to `1`.
+   * @param options.prefix Prefix for the generated number. Defaults to `'0x'`.
```
ST-DDT: `0x` is the hex prefix IIRC, the binary variant usually use the `0b` variant.

src/modules/string/index.ts
```diff
@@ -241,6 +241,93 @@ export class StringModule {
...
+
+    return `${prefix}${binaryString}`;
+  }
+
+  /**
+   * Returns an [octal](https://en.wikipedia.org/wiki/Octal) string.
+   *
+   * @param options The optional options object.
+   * @param options.length The number or range of characters to generate after the prefix. Defaults to `1`.
+   * @param options.prefix Prefix for the generated number. Defaults to `'0x'`.
```
ST-DDT: `0x` is the hex prefix IIRC, the octal variant usually use the `0o` variant.

ST-DDT (commented):
Especially octal and binary seem to have a multitude of common prefixes, should we use `''` as the default prefix for all of them?

Only one of our methods uses a non-empty prefix (etheriumAddress), all other 5 methods use ''.

src/modules/string/index.ts
Shinigami92: Yes, `0x` is actually wrong here! `0b` is the correct form
Same applies for `0x` to `0o` below, ~~but maybe just need to be `0` as prefix in JS~~

~~See:~~



Nevermind: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_octal

`0o10 === 8` as example is correct

pladreyt: Oh, good catch ! Thanks !. I am fixing it now

src/modules/string/index.ts
pladreyt: Thanks !

pladreyt:
> Especially octal and binary seem to have a multitude of common prefixes, should we use `''` as the default prefix for all of them?
> 
> Only one of our methods uses a non-empty prefix (etheriumAddress), all other 5 methods use ''.

Also the `hexadecimal` method uses a non-empty prefix, I was using it as a base. Was keeping the prefix for consistency with that method, but I can set it to `''`

src/modules/string/index.ts
```diff
@@ -241,6 +241,93 @@ export class StringModule {
...
+   * @param options The optional options object.
+   * @param options.length The number or range of characters to generate after the prefix. Defaults to `1`.
+   * @param options.prefix Prefix for the generated number. Defaults to `'0b'`.
+   *
+   * @example
+   * faker.string.binary() // '0b1'
+   * faker.string.binary({ length: 10 }) // '0b1101011011'
+   * faker.string.binary({ length: { min: 5, max: 10 } }) // '0b11101011'
+   * faker.string.binary({ prefix: '0b' }) // '0b1'
+   * faker.string.binary({ length: 10, prefix: '#' }) // '#1101011011'
```
Shinigami92: I think this example doesn't make much sense, but we could swap it out with `bin` for binary 🤔 

```suggestion
   * faker.string.binary({ length: 10, prefix: 'bin' }) // 'bin1101011011'
```

Please let have @ST-DDT also have a look over my suggestion before applying it as I'm not sure

---

Edit: Oh, below in the tests you use already `bin_` and `oct_`!
Please use these here in the examples instead

ST-DDT:
Please add `@see` to and from the number module.

src/modules/string/index.ts
pladreyt: Sure, sounds good!

pladreyt:
> Please add `@see` to and from the number module.

Sure!

src/modules/string/index.ts
```diff
@@ -241,6 +241,97 @@ export class StringModule {
...
+   * @see faker.number.binary() If you would like to generate a `binary number` (within a range).
+   *
+   * @example
+   * faker.string.binary() // '0b1'
+   * faker.string.binary({ length: 10 }) // '0b1101011011'
+   * faker.string.binary({ length: { min: 5, max: 10 } }) // '0b11101011'
+   * faker.string.binary({ prefix: '0b' }) // '0b1'
+   * faker.string.binary({ length: 10, prefix: 'bin_' }) // 'bin_1101011011'
+   * faker.string.binary({ prefix: '' }) // '1'
+   * faker.string.binary({ length: 10, prefix: '0b' }) // '0b1101011011'
```
ST-DDT: These examples are redundant/duplicate (I assume they originally demonstrated the casing property)

src/modules/string/index.ts
```diff
@@ -241,6 +241,97 @@ export class StringModule {
...
+   * @see faker.number.octal() If you would like to generate an `octal number` (within a range).
+   *
+   * @example
+   * faker.string.octal() // '0o3'
+   * faker.string.octal({ length: 10 }) // '0o1526216210'
+   * faker.string.octal({ length: { min: 5, max: 10 } }) // '0o15263214'
+   * faker.string.octal({ prefix: '0o' }) // '0o7'
+   * faker.string.octal({ length: 10, prefix: 'oct_' }) // 'oct_1542153414'
+   * faker.string.octal({ prefix: '' }) // '2'
+   * faker.string.octal({ length: 10, prefix: '0o' }) // '0o1526216210'
```
ST-DDT: These examples are redundant/duplicate (I assume they originally demonstrated the casing property)

ST-DDT (commented):
Two minor changes, then we are good to go from my side.

src/modules/string/index.ts
pladreyt: Yes, you are right, will remove them

src/modules/string/index.ts
pladreyt: Yes, you are right, will remove them

ST-DDT (approved):
Thanks for your contribution.

Shinigami92 (approved):

PR #1716, opened 2023-01-08
feat(string): nanoid

### LINKED ISSUE #1426: Add nanoid for uuid alternatives

nanoid is a great alternative for uuid, it is had less complexity and size.

it is commonly used for blog post unique urls.

### Suggested solution

Add nanoid, the logic should based on https://github.com/ai/nanoid

### Alternative

No alternative for https://github.com/ai/nanoid

### Additional context

_No response_

xDivisionByZerox:
Resolves #1426.

- #1426


test.ts
```diff
@@ -0,0 +1,5 @@
+import { faker } from './src';
+
+console.log(faker.string.nanoid());
+console.log(faker.string.nanoid(10));
+console.log(faker.string.nanoid({ min: 13, max: 37 }));
```
ST-DDT: Please remove this debug code.

src/modules/string/index.ts
```diff
@@ -532,6 +532,50 @@ export class StringModule {
...
+   *
+   * @since 8.0.0
+   */
+  nanoid(length: number | { min: number; max: number } = 21): string {
+    length = this.faker.helpers.rangeToNumber(length);
+    if (length <= 0) {
+      return '';
+    } else if (length > 36) {
+      length = 36;
+    }
```
ST-DDT: The original implementation doesn't seem to have this limit:

https://github.com/ai/nanoid/blob/6cd578dca8c5c45946eb15796de3e3c91fa187c4/non-secure/index.js#L23-L32

ST-DDT (changes requested):

test.ts
xDivisionByZerox: ofc, my bad.

src/modules/string/index.ts
xDivisionByZerox: I got that from their [official website](https://zelark.github.io/nano-id-cc/), where `36` is the maximum size of an id.

ST-DDT: > More than 1 quadrillion years needed, in order to have a 1% probability of at least one collision.

I assume that after that the math no longer works out regarding their collision detection.
(that page is specifically for collision probability calculation)

Their source code doesn't have a limit so we shouldn't have one either IMO.

xDivisionByZerox: No problem. I removed the maximum restriction.

ST-DDT (approved):

ejcheng (approved):

src/modules/string/index.ts
```diff
@@ -532,6 +532,48 @@ export class StringModule {
...
+    const generators = [
+      {
+        value: () => this.alphanumeric(1),
+        // a-z is 26 characters
+        // this times 2 for upper & lower case is 52
+        // add all numbers 0-9 (10 in total) you get 62
+        weight: 62,
+      },
+      {
+        value: () => this.faker.helpers.arrayElement(['_', '-']),
```
Shinigami92, line 564: I searched if it's allowed to start with `_|-` but I found nothing :shrug: 
So seems it's okay

Shinigami92 (approved):

PR #1717, opened 2023-01-08
feat(string): generate string from characters

ST-DDT:
Most of our string methods just bundle a set of characters together.
This PR adds a method that simplifies that by providing a specific method for exactly that but with dynamic input characters.


ST-DDT:
Might be useful for https://github.com/faker-js/faker/pull/1716 as well.

src/modules/string/index.ts
```diff
@@ -96,6 +96,44 @@ export class StringModule {
...
+  fromCharacters(
+    characters: string | string[],
+    length: number | { min: number; max: number } = 1
+  ): string {
+    length = this.faker.helpers.rangeToNumber(length);
+    if (length <= 0) {
+      return '';
+    }
+
+    if (typeof characters === 'string') {
```
xDivisionByZerox, line 125: Should we potentially check, that if `characters` is an array, that all "string" elements have `length === 1`?

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/1718

src/modules/string/index.ts
```diff
@@ -96,6 +96,48 @@ export class StringModule {
...
+   * @example
+   * faker.string.fromCharacters('abc') // 'c'
+   * faker.string.fromCharacters(['a', 'b', 'c']) // 'a'
+   * faker.string.fromCharacters('abc', 10) // 'cbbbacbacb'
+   * faker.string.fromCharacters('abc', { min: 5, max: 10 }) // 'abcaaaba'
+   *
+   * @since 8.0.0
+   */
+  fromCharacters(
+    characters: string | string[],
```
Shinigami92: Should we allow `string | readonly string[]`?
Maybe it needs to be `string | ReadonlyArray<string>` which is technically the same but the linter ...

src/modules/string/index.ts
```diff
@@ -96,6 +96,48 @@ export class StringModule {
...
+      characters = characters.split('');
+    }
+
+    if (characters.length === 0) {
+      throw new FakerError(
+        'Unable to generate string: No characters to select from.'
+      );
+    }
+
+    const { multiple, arrayElement } = this.faker.helpers;
```
Shinigami92: :eyes: 
What's this and why here but not also for `rangeToNumber` above?
But tbh, this is a new style of writing code in our code-base
I'm not against it, but we should maybe discuss if and why we prefer that (out of a sudden)

src/modules/string/index.ts
ST-DDT: No, I didn't explicitly checked that because some emojis consist of multiple characters but are displayed as single one.

Should I add a jsdoc hint, that each element in the array is treated as a single character even though their length might be != 1?

src/modules/string/index.ts
ST-DDT: Fixed

src/modules/string/index.ts
ST-DDT: Reverted

Shinigami92 (approved):

src/modules/string/index.ts
ST-DDT: I added a hint to the jsdocs.

xDivisionByZerox: > No, I didn't explicitly checked that because some emojis consist of multiple characters but are displayed as single one.

That's a good hint. I totally forgot that.

xDivisionByZerox (approved):

PR #1718, opened 2023-01-08
test: repair all functional test

ST-DDT:
Fixes various problems with the all functional tests:

- grafik
- fakerAZ.company.suffixes returns `[]` but is marked as valid
- faker.helpers.fake tests often shows up as valid even if the return "value" is not
- It is currently not possible to exclude a method from the tests for all locales
- Obsolete configuration exists and isn't marked for removal


ST-DDT:
Required for https://github.com/faker-js/faker/pull/1717

test/all_functional.spec.ts
```diff
@@ -111,14 +114,24 @@ describe('faker.helpers.fake functional tests', () => {
...
+            const testAssertion = () => {
               faker.locale = locale;
               // TODO ST-DDT 2022-03-28: Use random seed once there are no more failures
               faker.seed(1);
               const result = faker.helpers.fake(`{{${module}.${meth}}}`);
 
               expect(result).toBeTypeOf('string');
-            });
+              expect(result).not.toBe('');
+              expect(result).not.toBe('undefined');
```
xDivisionByZerox, line 148: What about `not.toBe('null')`? We have methods that return objects that could potentially resolve `null` values.

test/all_functional.spec.ts
```diff
@@ -46,7 +47,8 @@ function isWorkingLocaleForMethod(
   meth: string,
   locale: string
 ): boolean {
-  return (BROKEN_LOCALE_METHODS[mod]?.[meth] ?? []).indexOf(locale) === -1;
+  const exclude = BROKEN_LOCALE_METHODS[mod]?.[meth] ?? [];
+  return exclude !== '*' && exclude.indexOf(locale) === -1;
```
xDivisionByZerox: How does `'*'` happen? I don't see that value up in `BROKEN_LOCALE_METHODS`.

test/all_functional.spec.ts
ST-DDT: > We have methods that return objects that could potentially resolve null values.

Which one?

Should I add the check anyway?

test/all_functional.spec.ts
ST-DDT: That is in preparation for https://github.com/faker-js/faker/pull/1717

test/all_functional.spec.ts
xDivisionByZerox: The methods in the `ScienceModule` return an object. So in theory, if something goes horrible wrong, they could resolve to null. This will most likely not happen (looking at the source code), but better safe than sorry I guess.

Since your test check for 'nullish' string values I would include it, just in case.

test/all_functional.spec.ts
xDivisionByZerox: So `BROKEN_LOCALE_METHODS` will be from type `Record<string, Record<string, '*' | string[]>>`, where `'*'` acts as a wildcard for all locales. Correct?
Just just want to make sure, that I understand what you have in mind.

test/all_functional.spec.ts
ST-DDT: Fixed.

test/all_functional.spec.ts
ST-DDT: Yes, that's correct.
I added an explicit type to the data now.

xDivisionByZerox (dismissed):

test/all_functional.spec.ts
```diff
@@ -75,6 +68,35 @@ function modulesList(): { [module: string]: string[] } {
 
 const modules = modulesList();
 
+describe('BROKEN_LOCALE_METHODS test', () => {
+  it('should not contain obsolete configuration (modules)', () => {
+    const moduleNames = Object.keys(modules);
+    const configuredModules = Object.keys(BROKEN_LOCALE_METHODS ?? {});
+    const obsoleteModules = configuredModules.filter(
+      (meth) => !moduleNames.includes(meth)
```
xDivisionByZerox: Wouldn't this be `module` instead of `meth`? I have a hard time understanding what's going on with the current variable name.

ST-DDT: Nicely spotted. Copy and pasting is prone to these errors. I should pay more attention there (but it's so hard).

test/all_functional.spec.ts
```diff
@@ -72,6 +68,35 @@ function modulesList(): { [module: string]: string[] } {
...
 
+describe('BROKEN_LOCALE_METHODS test', () => {
+  it('should not contain obsolete configuration (modules)', () => {
+    const existingModules = Object.keys(modules);
+    const configuredModules = Object.keys(BROKEN_LOCALE_METHODS ?? {});
+    const obsoleteModules = configuredModules.filter(
+      (module) => !existingModules.includes(module)
+    );
+
+    expect(obsoleteModules, 'No obsolete configuration').toEqual([]);
```
xDivisionByZerox, line 78: I'm a bit confused about what you actually test here. Are you simply checking that `BROKEN_LOCALE_METHODS` has no configuration for a module that doesn't exist?
Like, this test would fail if `BROKEN_LOCALE_METHODS` had a key like `'foobar'`. I'm a bit confused why this would be required.

ST-DDT: As you can see from the configs that I removed:
The configs didnt do anything anymore so they should be removed (in the future along with their implementation).

test/all_functional.spec.ts
```diff
@@ -1,13 +1,14 @@
 import { describe, expect, it } from 'vitest';
 import { faker } from '../src';
+import type { KnownLocale } from '../src/locales';
 
 const IGNORED_MODULES = [
   'locales',
-  'locale',
-  'localeFallback',
   'definitions',
   'fake',
```
ST-DDT:
```suggestion
```

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1719, opened 2023-01-08
refactor(company): deprecate suffix methods

ST-DDT:
The company suffixes are just parts of the name but don't carry any use by themselves.

The locale data for the suffixes won't be removed in v9, the suffixes are just no longer directly accessible.


ejcheng (approved):

src/modules/company/index.ts
```diff
@@ -18,12 +19,22 @@ export class CompanyModule {
...
    * @example
    * faker.company.suffixes() // [ 'Inc', 'and Sons', 'LLC', 'Group' ]
    *
    * @since 2.0.1
+   *
+   * @deprecated Use `faker.company.name` instead.
    */
   suffixes(): string[] {
+    deprecated({
+      deprecated: 'faker.company.suffixes',
```
xDivisionByZerox, line 33: Did we drop the `()` after method names in deprecation messages? Since you wrote them in the `@see` tag I suppose you intentionally didn't wrote them?

ST-DDT: Half the methods have `()`, the others don't.

xDivisionByZerox: The example in our [CONTRIBUTING](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#deprecation-workflow) shows both (with and without parenthesis). We should stick with one for consistency reasons.
I'll add that as an bullet point to the meeting notes and resolve this here.

xDivisionByZerox (approved):

Shinigami92 (approved):

Issue #1441, opened 2022-10-13
Add global locale


Shinigami92:
We want to move global locale data out of locale definitions to their related modules.
These data are the same for every language and therefore doesn't need to exist in locale `en` or be override-able by any other locale.

Example: mime-types, emojis, ...

ST-DDT:
I think we should rather introduce a `global` locale.
Like any of our other data there is no reason why we shouldn't make them replaceable.

Shinigami92:
As it doesn't make a difference if we put them in global locale or in the source directly, I'm fine with that, esm/cjs would duplicate them any way
We can only get rid of this duplication in the long run by getting rid of cjs in total :shrug: 

So yes, lets do it that way :+1:

PR #1720, opened 2023-01-08
refactor(location)!: countryCode use ISO-3166-1

xDivisionByZerox:
Scope country code in the location module while satisfying [ISO-3166](https://en.wikipedia.org/wiki/ISO_3166-1).


src/definitions/location.ts
```diff
@@ -111,16 +111,6 @@ export type LocationDefinitions = LocaleEntry<{
...
-  /**
-   * The ISO-3166-1 ALPHA-2 country codes related to this locale.
-   */
-  country_code: string[];
-
-  /**
-   * The ISO-3166-1 ALPHA-3 country codes related to this locale.
-   */
-  country_code_alpha_3: string[];
-
```
ST-DDT, line 123: I'm not sure whether I would want all country codes or only a specific subset related to my country set.
(de => DE, CH, AT)

We might want to have to full list in our global locale and specific lists per other locale.

location. timezone also "only" contains timezones explicitly for that location.

src/modules/location/country-codes.ts
```diff
@@ -0,0 +1,256 @@
+// [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1)
+export const countryCodes: {
+  alpha2: string;
+  alpha3: string;
+  numeric: string;
```
Shinigami92: You added this `numeric` but it is not really accessible, is it?
Also I never saw it or needed the use for such a `numeric` code. So should we just remove it from the data set?

xDivisionByZerox: This is only in there since as kind of a "proposal". It was super easy coping this from the [original source table](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements). If we wish to not ship them, its fine to me - I can remove them if you say so.

Shinigami92: I would say "do not ship them until someone ask for now"
But lets await @ST-DDT response

ST-DDT: I haven't used them yet either. So lets omit them for now.

xDivisionByZerox: I've removed the `numeric` property

Shinigami92 (dismissed):

src/definitions/location.ts
ST-DDT: I would prefer if we keep that (all) data localized.

Shinigami92: Please provide a why

One point for why it is nice to have this not localized is: it can be used for whatever locale you are using, it's always the same content and can even be used if we someday us a LocalizedFaker because then it would be in the BaseFaker

ST-DDT: The reason why I want it localized it that you can replace it with the data you actually care about.

> it's always the same content

No it#s not. If you only need European countries, then you don't need American ones and can limit the results just like any other of our methods.

ST-DDT: (that's what the global locale is for)

Shinigami92: So you want to introduce the global locale first in another PR before merging this one?
I think it would be helpful to open an issue for the global locale so we can keep track of its state.

ST-DDT: No, there is no need to wait for the global locale, for now simply replace/use the en like it was before.

https://github.com/faker-js/faker/issues/1441

ST-DDT:
**Team decision**

We will keep the locale data as localized data, to allow custom overwrites, but we will switch to [{alpha2, alpha3}] locale data and use the ISO data.

Shinigami92 (approved):

src/locales/de_CH/location/country_code.ts
```diff
@@ -1 +1,10 @@
-export default ['CH', 'DE', 'AT', 'US', 'LI', 'HK', 'VN'];
+// [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1)
+export default [
+  { alpha2: 'CH', alpha3: 'CHE' },
```
matthewmayer, line 3: Why Hong Kong, us, Vietnam?

ST-DDT: This should be addressed in a different PR.
But this needs fixing: https://github.com/faker-js/faker/issues/1861

ST-DDT (approved):

ejcheng (approved):

PR #1731, opened 2023-01-11
refactor(locale)!: remove unused separator entry

ST-DDT:
The `separator` is currently unused and doesn't really contain a separator, instead it contains a combinator: `' & '`.

This PR removes the separator from the definitions, the locale data and related scripts.

---------

A following PR will move the `title` to a new `metadata` category to remove any exceptions from the `locale: { [category: string]: { [entry: string]: any }}` data structures. (To be able to remove the special case handling currently required for those fields).


xDivisionByZerox (approved):
This is a breaking change!

Shinigami92:
I'm not sure if this was used for `helpers.fake` to generate text in different locales.

But nowadays there is something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat
which has a `conjunction` and `disjunction` parameter
so maybe this is more likely to be used but has no support for `helpers.fake` anymore

ST-DDT:
> I'm not sure if this was used for `helpers.fake` to generate text in different locales.

It isn't supported there either.

````ts
faker.helpers.fake("{{separator}}")
````

````ts
[Error]: Invalid module method or definition: separator
- faker.separator is not a function
- faker.definitions.separator is not an array
````

We don't use it and the only way to actually access it is via `faker.definitions.separator` directly.

That's why I would like to remove it.

Shinigami92:
I would like to discuss this in a meeting and want to raise the question if we need an alternative
Maybe we broke something from pre-v6 times and just missed something

xDivisionByZerox:
In today meeting we ensured that this entry was never used and can be removed savely.  

I would still suggest to mark this as breaking change.

Shinigami92 (approved):

PR #1735, opened 2023-01-13
refactor!: remove dynamic locale switching support

### LINKED ISSUE #1340: Remove locale switching support

### LINKED ISSUE #1524: Locale de_xy: house numbers are after street name in Europe, not before
### Pre-Checks

### Describe the bug

**Issue**
When generating addresses in Europe (set locale to 'de' or 'de_CH', street addresses should be in the format 'streetName houseNumber'.
Correct: Deutlerweg 12
Wrong: 12 Deutlerweg

**Steps to reproduce**
```
faker.setLocale('de_CH')
const address = faker.address.streetAddress()
```

**Current behaviour**
 "address": "09336 Reinhard Trail",

**Expect behaviour**
 "address": "Reinhard Trail 9336"

### Minimal reproduction code

```
faker.setLocale('de_CH')
const address = faker.address.streetAddress()
```

### Additional Context
To be super correct, the format actually should be in the format of the country that the address is generated in. However this is not possible as generation is not aware of the country at the time of address generation. It makes sense to implement the change, as if I am a Swiss developer, I most likely target the Swiss market and will generate fake data targeted to Switzerland.

### Environment Info

```shell
System:
    OS: macOS 13.0
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 35.21 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.10.0 - ~/node_modules/.bin/npm
  Browsers:
    Brave Browser: 107.1.45.118
    Chrome: 107.0.5304.87
    Firefox: 106.0.3
    Safari: 16.1
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

ST-DDT:
Closes #1340

- #1340 

Removes the ability to change the locale on existing Faker instances.

# New Features

- Access locales directly from the main index file (`en_AU_ocker`)
- Access to all locales at once (`allLocales.en_AU_ocker`)
- Access faker instances directly from the main index file (`fakerEN_AU_ocker`)
- Access to all faker instances at once (`allFakers.en_AU_ocker`)
- Useful locale fallback (`en_AU_ocker` -> `en_AU` -> `en` ( -> `en` if not already present ( -> `global` later))

# Changed Features

- The main `faker` instance is now `en` **only**

# Lost Features

- Changing locales dynamically
- Getting a random locale from the faker instance (and switch to that)
- UsableLocale, UsedLocale, KnownLocale types

# ToDos for later PRs

- Extract `global` locale data into own locale (#1441)
- Actually throw for missing locale entries (#893)
- `findFaker(locale: string): Faker` method **???**


test/locales.spec.ts
```diff
@@ -1,35 +0,0 @@
-import { describe, expect, it } from 'vitest';
```
ST-DDT, line 1: This file was renamed to `locale-data.spec.ts`.

test/internet.spec.ts
```diff
@@ -111,9 +107,8 @@ describe('internet', () => {
         });
 
         it('should return a valid email in every locale', () => {
-          for (const locale of Object.keys(faker.locales)) {
-            faker.setLocale(locale);
-            const email = faker.internet.email();
+          for (const [locale, localeFaker] of Object.entries(allFakers)) {
```
Shinigami92: Maybe we should rewrite this test to use `it.each(Object.entries(allFakers))(...)`

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -85,11 +82,7 @@ describe('examples and deprecations', () => {
       const signature = signatures[signatures.length - 1];
 
       // Extract examples and make them runnable
-      let examples = extractRawExamples(signature).join('').trim() ?? '';
-      examples = examples.replace(
-        /faker([A-Z]{2})\./g,
-        (_, locale: string) => `faker.locale = '${locales[locale]}';\nfaker.`
```
Shinigami92: Line 91 was removed and so it reports that
line 35 is not used anymore

test/support/seededRuns.ts
```diff
@@ -130,7 +130,6 @@ class TestGenerator<
    */
   setup(): void {
     this.faker.seed(this.seed);
-    this.faker.locale = 'en';
```
Shinigami92, line 133: I haven't checked it but now as we do not have a shared faker instance where we switch the locale at runtime, seededRuns that potentially use another locale need to also set the seed manually

It is likely that this is not a problem right now, but should be remembered if it is or gets a case

src/modules/random/index.ts
```diff
@@ -157,19 +157,6 @@ export class RandomModule {
...
-  /**
-   * Returns a random locale, that is available in this faker instance.
-   * You can use the returned locale with `faker.setLocale(result)`.
-   *
-   * @example
-   * faker.random.locale() // 'el'
-   *
-   * @since 3.1.0
-   */
-  locale(): string {
```
Shinigami92, line 182: Not sure about this hard removal without at least a deprecation and/or alternative

Yes we have `countryCode`, but this could be used e.g. to randomly select a locale for e.g. `vue-i18n`
I googled a bit around and we might need something like `ISO 639-1` and/or `ISO 3166`
Maybe this can be implemented in another PR in beforehand and then we can decorate this method `random.locale` to indicate that there is a new alternative

src/faker.ts
```diff
@@ -137,91 +90,20 @@ export class Faker {
...
-  }
-
-  /**
-   * Creates a Proxy based LocaleDefinition that virtually merges the locales.
-   */
-  private initDefinitions(): LocaleDefinition {
-    // Returns the first LocaleDefinition[key] in any locale
-    const resolveBaseData = (key: keyof LocaleDefinition): unknown =>
-      this.locales[this.locale][key] ?? this.locales[this.localeFallback][key];
+  constructor(options: { locale: LocaleDefinition | LocaleDefinition[] }) {
```
Shinigami92: Lets still extract this into `FakerOptions`
We already create very big hard breaking changes with this PR without alternatives, but lets keep it as less as possible

By staying with `FakerOptions` someone can still do:
```ts
const options: FakerOptions = { ... }
// ...
const myFaker = new Faker(options)
```

Otherwise they would be forced to do hard stuff like:
```ts
const options: ConstructorParameters<Faker>[0] = { ... }
// ...
const myFaker = new Faker(options)
```

To me this is less user friendly

src/modules/random/index.ts
ST-DDT: The problem is, that regardless what we do we change the semantic of the method.
Previously, it returned one of the loaded/available locales inside that faker instance, that is no longer possible.

- Returning a country code instead would be an entirely new method, that has nothing in common with the previous one.
- Returning a locale code would also be an entirely new method, that at most returns values that look like the previous ones, but are not
- Returning a supported locale changes the semantics of the method, because previously only "active" locales would be returned. (This would mimic the behavior of the main faker instance, but would violate the behavior for all others)
- Adding a locale metadata to each locale that contains their locale code, would allow this method to return at least one valid result. However, there is no longer a method to use the result in. Users were previously unable to use it as a way to get the active locale, because it most likely returned a random one.

src/faker.ts
ST-DDT: I'll think about it.

Currently, new Faker is the only method in our entire repo that supports that.
So what is so special about that one? I assume, it isn't even the method that is used the most.

test/support/seededRuns.ts
ST-DDT: Not really, the seed is set for the faker instance that is used for the tests (this.faker), so if the seeded tests run in a different locale, then the referenced faker instance would be a different one and thus the set here correctly.

https://github.com/faker-js/faker/blob/788fce048f0233552df47f3407c06d8f83c32b0c/test/support/seededRuns.ts#L70-L77

https://github.com/faker-js/faker/blob/788fce048f0233552df47f3407c06d8f83c32b0c/test/animal.spec.ts#L30-L32

````diff
- seededTests(faker, 'animal', (t) => {
+ seededTests(fakerEN_US, 'animal', (t) => {
````

(Theoretically, it is possible to run the seeded tests for all supported locales, with next to no effort, but that will likely explode our test snapshots file beyond review-ability.)

File: test/scripts/apidoc/examplesAndDeprecations.spec.ts (hunk 3 above)
Comment (ST-DDT) [reply test/scripts/apidoc/examplesAndDeprecations.spec.ts]: Fixed

File: test/support/seededRuns.ts (hunk 4 above)
Comment (matthewmayer) [reply test/support/seededRuns.ts:133]: i would suggest doing this for a very small subset of methods which are commonly localized, e.g. names, email addresses, streets and city names, phone numbers.

File: src/faker.ts [hunk 7]
```diff
@@ -295,13 +177,4 @@ export class Faker {
 
     return seed;
   }
-
-  /**
-   * Set Faker's locale
-   *
-   * @param locale The locale to set (e.g. `en` or `en_AU`, `en_AU_ocker`).
-   */
-  setLocale(locale: UsableLocale): void {
```
matthewmayer, line 328: I think rather than removing the function, you should keep it and throw a helpful error when called (at least for v8) if trying to call faker.setLocale or faker.locale, for example "You can no longer set the locale of faker at runtime, please see the migration guide at ..."

That helps people upgrading to figure out what's wrong with their code if it breaks. 

```
> faker.locale = "de" //silently fails to do anything
'de'
> faker.setLocale("de")
Uncaught TypeError: faker.setLocale is not a function //ok, but not very informative
```

matthewmayer (changes requested):

src/faker.ts
Shinigami92: I 💯% agree with @matthewmayer
Especially in plain JS you wont get just `setLocale is not a function on Faker`, but a helpful message that it is not a bug 

Beside that, IMO we should NEVER just remove a function in only one major bump. That is the reason why we brought `deprecated` function into the project

src/faker.ts
Shinigami92: The only difference I currently see is that this is a class (constructor) not a function, but tbh this is not the reason

I actually always complained about that we use written out option types for everything just to have nice docs and or IDE hints
But when I e.g. use VueUse and they have not exported named options like this for a function it is a bit of a bad developer experience and I always need to work around such things until they introduce it in their repo
In VueUse or Quasar I can also often do things with types like `QTableProps['columns']` to e.g. reuse them in my own component props and forward them, and if they ever got an update or type change, I can see it by checking with `tsc`

src/modules/random/index.ts
Shinigami92: tbh it still would be possible by just taking the 56(?) strings and just paste them into the function, then add a deprecation

For the proposed ISO functions, IMO we can create and open new PRs (and/or issues/feature requests)

src/faker.ts
matthewmayer: Ideally yes, in 8.0 calling 
`faker.locale = "de"` or `faker.setLocale("de")` would throw a deprecation warning, but still work, and the error message would encourage you to switch to the new method for setting locale when you import/create faker. Then in 9.0 it would be completely removed. I understand doing this over two versions is more work (and you need to plan ahead, to ensure you know what the 'new' syntax will be for setting locale) but i think the payoff in terms of easier upgrading is worthwhile.

ST-DDT: There is one problem though.
If we use the key to lookup the locale data, then all of our faker instances require all locale data, regardless of whether we/the user want it or not.
Which effectively renders our v8's tree shaking mood and worse even if you created a small custom faker instance in v7 yourself, that would now blown up to the entire 5MB(?) at runtime without any way to prevent that.

I'm okay with keeping the method and always throw an error in it (which would worsen the DX for TS users), but beyond that everything else needs serious considerations.

ST-DDT: As an alternative, we could put the full locale map inside the main faker instance (and their subsets in the respective localized faker instances) to ship only the previous locales and allow our users to still create smaller instances, by omitting the locale map from their custom instances.

The default instances are then slightly larger than in v7 and depending on the tree shaking process a lot larger than they could be (depending on whether the mergeLocale script is executed at build-time or run-time).

src/modules/random/index.ts
ST-DDT: Returning the string(s) is not the problem, the problem is what you (can) do with the string: https://github.com/faker-js/faker/pull/1735#discussion_r1072008424

Shinigami92: > Returning the string(s) is not the problem, the problem is what you (can) do with the string: [#1735 (comment)](https://github.com/faker-js/faker/pull/1735#discussion_r1072008424)

Maybe don't think so far, return just strings here and throw an error in `setLocale`
That's it
It's still a deprecation with a breaking change

- setLocale for removal
- random.locale for replacement (other PR)

src/faker.ts
matthewmayer: Is there a way to throw an error when faker.locale = "foo" is called (this is the current way that the README recommends setting locales, so you can assume a lot of consumers will have followed that)

image

Shinigami92: > Is there a way to throw an error when faker.locale = "foo" is called

Yes, we can use a `set` accessor for that

src/faker.ts
ST-DDT: How about exporting all options as types:

````ts
export FakerOptions = ConstructorParameters<Faker>[0];
````

and later for each method:

````ts
export StringAlphaOptions = Parameters<Faker>[index of options parameter];
````

(generated)

Comment (Shinigami92) [reply src/faker.ts]: we could try, not sure how that works for overload signatures :thinking:

File: src/faker.ts [hunk 8]
```diff
@@ -137,91 +90,88 @@ export class Faker {
...
+  // This is somehow required for `ConstructorParameters<typeof Faker>[0]` to work
+  constructor(
+    options:
+      | { locale: LocaleDefinition | LocaleDefinition[] }
+      | {
+          locales: Record<string, LocaleDefinition>;
+          locale?: string;
+          localeFallback?: string;
+        }
+  );
```
ST-DDT, line 158: @Shinigami92 Any idea why?

src/faker.ts
ST-DDT: Done

src/faker.ts

test/support/seededRuns.ts
ST-DDT: Moved the discussion to: https://github.com/faker-js/faker/issues/1749

src/modules/random/index.ts
ST-DDT: I pseudo removed the method for now.
(It won't be shown in the api docs either)

scripts/apidoc/moduleMethods.ts
```diff
@@ -20,6 +24,37 @@ export function selectApiModules(
...
+    .filter((method) => !method.flags.isPrivate);
+}
+
+/**
+ * Selects the signature from the method that needs to be documented.
+ *
+ * @param method The method to extract the signature from.
+ * @returns The signature to document.
+ */
+export function selectApiSignature(
```
ST-DDT: TODO: Move to utils and probably its own PR (module private method support for api-docs)

src/modules/random/index.ts
xDivisionByZerox: I'm with @Shinigami92 here. The function itself should still return a locale string from a hardcoded array. Since you can no longer use the string to set the locale I don't see any problem with that. Yes, the use case will most likely be gone, but at least the method itself is still there and doesn't break anything.
We cannot "just" remove a function from our API. 

<details>
<summary>Implementation</summary>

```ts
/**
 * [...]
 */
locale(): string {
  return faker.helpers.arrayElement([
    'af_ZA',
    'ar',
    'az',
    'cz',
    'de',
    'de_AT',
    'de_CH',
    'dv',
    'el',
    'en',
    'en_AU',
    'en_AU_ocker',
    'en_BORK',
    'en_CA',
    'en_GB',
    'en_GH',
    'en_IE',
    'en_IN',
    'en_NG',
    'en_US',
    'en_ZA',
    'es',
    'es_MX',
    'fa',
    'fi',
    'fr',
    'fr_BE',
    'fr_CA',
    'fr_CH',
    'fr_LU',
    'ge',
    'he',
    'hr',
    'hu',
    'hy',
    'id_ID',
    'it',
    'ja',
    'ko',
    'lv',
    'mk',
    'nb_NO',
    'ne',
    'nl',
    'nl_BE',
    'pl',
    'pt_BR',
    'pt_PT',
    'ro',
    'ru',
    'sk',
    'sv',
    'tr',
    'uk',
    'ur',
    'vi',
    'zh_CN',
    'zh_TW',
    'zu_ZA',
  ]);
}
```
</details>

Also, I don't think the "psydo removal" helps anybody here.

test/all_functional.spec.ts
```diff
@@ -29,14 +22,20 @@ const BROKEN_LOCALE_METHODS = {
...
     fromCharacters: '*',
   },
   person: {
     prefix: ['az', 'id_ID', 'ru', 'zh_CN', 'zh_TW'],
     suffix: ['az', 'it', 'mk', 'pt_PT', 'ru'],
   },
-} satisfies Record<string, Record<string, '*' | KnownLocale[]>>;
+} satisfies Record<
+  string,
+  Record<string, '*' | ReadonlyArray<keyof typeof allLocales>>
```
xDivisionByZerox: I like that change 👍 
Should we also make the module key more strict? Something like `Omit<keyof Faker, 'definitions' | 'helpers'>`

src/faker.ts
Shinigami92: I'm not sure if I ever overwrote a constructor like that because I normally do not differ such options that way (outside of Faker)

But I assume the last overload is always needed to implement the real-js signature
Everything above (the actual overloads) are just plain TS and are not real at runtime

ST-DDT: But the implementation signature is the same as this one (duplicate)!?

test/all_functional.spec.ts
ST-DDT: Done but I omitted the omit for readability.

src/faker.ts
Shinigami92: Ah I see what you mean

```
1
2
1|2
real 1|2
```

don't know, feel free to ask in StackOverflow or GitHub https://github.com/microsoft/TypeScript/issues

scripts/apidoc/moduleMethods.ts
ST-DDT: Done

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/1759

scripts/apidoc/typedoc.ts
```diff
@@ -75,7 +75,9 @@ export function selectApiModules(
 export function selectApiMethods(
   module: DeclarationReflection
 ): DeclarationReflection[] {
-  return module.getChildrenByKind(ReflectionKind.Method);
+  return module
+    .getChildrenByKind(ReflectionKind.Method)
+    .filter((method) => !method.flags.isPrivate);
```
ST-DDT, line 99: FFR: @Shinigami92 would like to discuss this in the next team meeting.

Originally posted here: https://github.com/faker-js/faker/pull/1759#discussion_r1082981075

scripts/generateLocales.ts
```diff
@@ -107,19 +114,31 @@ function escapeField(parent: string, module: string): string {
...
 function generateLocaleFile(locale: string): void {
+  const parts = locale.split('_');
+  const locales = [locale];
+
+  for (let i = parts.length - 1; i > 0; i--) {
+    const fallback = parts.slice(0, i).join('_');
+    if (existsSync(resolve(pathLocales, fallback))) {
+      locales.push(fallback);
+    }
+  }
```
ST-DDT, line 126: Export this as a function for our users.

Shinigami92: Please extract this to its own PR and maybe even label it with `s: awaiting more info`

This is not blocking this PR from getting merged and this PR should be merged soon as it already blocks the process of v8.0 as well as some already open PRs

src/index.ts
```diff
@@ -34,7 +31,12 @@ export type {
   WordDefinitions,
 } from './definitions';
 export { FakerError } from './errors/faker-error';
-export type { FakerOptions, UsableLocale, UsedLocales } from './faker';
+export { Faker } from './faker';
+export type { FakerOptions } from './faker';
+export * from './locale';
+export { fakerEN as faker } from './locale';
+export * from './locales';
+export * as allLocales from './locales';
```
Shinigami92, line 39: This export bleeds to a non-treeshaked bundle while e.g. running with vite

I debugged with https://socket.dev/npm/package/vite-bundle-visualizer

We can circumvent this by configuring `sideEffects` in package.json and then move `allLocales` to its own file so it can be reexported from there as standalone without affecting the other ones

We already had once the request for this already

- #1305

We need to _somehow_ test that setting `sideEffects` brings no dark hidden breakings into the project, and I don't know where we could have side effects
So maybe instead of setting it to false, it might be that we need to configure an array of filenames

ST-DDT: Before we start adding flags to the project, we should 
- have an automated test suite that ensures that we don't break this optimization
- or/and preferably understand why there is an issue in the first place.

But thats a different PR/issue from this one.

Shinigami92: How would you like to proceed?

- we cannot really merge this into next branch as it would introduce a bug that is currently not in the next branch
- if you want to find out why this flag is working and what it is doing, I consider searching a bit around on stack overflow or other documentations (I don't know what and why it is doing)
- I'm not aware of how it could be possible to automate a test that checks the emitted bundle size of a bundled external project that use a bundler like rollup to apply tree-shake to its own project
  at least I'm not sure if this is possible to run inside a PR before merging it into main and then automatically test inside CI (without introducing a separate folder for a whole project just to test something that is touched like twice a year)

It is a bit conflicting, because without this PR merged, we do not have this bug and so cannot really test if it breaks tree-shaking, but merging this without fixing the tree-shaking bug it introduces is also not really possible
the only thing this would lead me to approve is: you give me the word that you also fix the following PR to fix the introduced bug
otherwise we would hang in a state were the project is somewhat unreleaseable

ST-DDT: I did some testing with our playground vite-cjs project.

The bundle size does not really change, if I add this change or note (~700KB).
Unless, I change the imports to import directly from the project root.
Regardless of what I import from the project root, the package size immediately jumps to 3.6MB.
So that issue is a preexisting condition that is NOT caused by this PR although it makes it easier to step into.

I even tried the current code on next with sideEffect: false, it didn't decrease the bundle size.
We have to investigate the reason for that, but IMO that is unrelated to this PR.

Shinigami92: Issue extracted to =>
- #1791

:+1:

ST-DDT:
Note to self:

TODO

- Change documentation
- Write migration guide

EDIT: Done

README.md
```diff
@@ -133,18 +132,26 @@ console.log(
...
-faker.locale = 'de';
+import { fakerDE } from '@faker-js/faker';
 ```
 
-See our documentation for a list of [provided languages](https://fakerjs.dev/guide/localization.html#available-locales)
+See our documentation for a list of [provided languages](https://fakerjs.dev/guide/localization.html#available-locales).
+
+Please note: Not every locale provides data for every module. In our pre-made faker instances,
+we fall back to English in such a case as this is the most complete and most commonly used language.
+If you don't want that or prefer a different fallback, you can also build your own instances.
```
Shinigami92, line 150: :heart: :+1:

docs/guide/localization.md
```diff
@@ -3,15 +3,20 @@
...
+See below for a list of available locales.
+
+:::tip Note
+You can also build your own Faker instances, with custom locales/overwrites.
 :::
 
 ## Individual localized packages
 
-By default, Faker will include **all** locale data.  
+Currently, the imports from the main package have a bug and always cause the entire Faker lib to be imported.
```
Shinigami92: Should we extend this sentence with e.g. a link to the issue + that we currently try to solve this but doesn't have the brain-power right now and looking for help?

docs/guide/localization.md
```diff
@@ -39,66 +44,72 @@ However, in most cases, using a specific locale will be beneficial in the long t
...
+| uk          | Ukrainian                 | `fakerUK`          |
+| ur          | Urdu                      | `fakerUR`          |
+| vi          | Vietnamese                | `fakerVI`          |
+| zh_CN       | Chinese                   | `fakerZH_CN`       |
+| zh_TW       | Chinese (Taiwan)          | `fakerZH_TW`       |
+| zu_ZA       | Zulu (South Africa)       | `fakerZU_ZA`       |
 
 <!-- LOCALES-AUTO-GENERATED-END -->
+
+The `Locale` (data) and (Faker) `Import` columns refer to the respective import names:
```
xDivisionByZerox: I'm confused here. `Import` should be `Faker` here to match the header no?

ST-DDT: I forgot to generate and commit the changes to the columns. Should be fixed now.
(Not sure about the names though)

docs/guide/localization.md
ST-DDT: Done, the linking. I think it is clear that we are aware of it and that we have it in mind, but haven't/cannot fix it.
If you open the link, then you get more information and the `help wanted` label.

ST-DDT:
So everything should be done now. Please review. thoroughly.

One thing, that I'm not sure about is:
Whether I should export the locale names (keys) as a separate and standalone const (e.g. `allLocaleNames`).
(Some of our [scripts](https://github.com/faker-js/faker/pull/1735/files#diff-56de87ca2ae95ee7c7a031bf487c87288b5dd206b66fed3ab6f551bf58081645R17) and [tests](https://github.com/faker-js/faker/pull/1735/files#diff-f8cd4ea46aecbba86ee9cba1a851bca523aca5e63bf9fa1866e3604a9d2dec26R6) only need the keys, but the code still loads everything (and then discards the unused parts).)

docs/guide/upgrading.md
```diff
@@ -14,6 +14,36 @@ Not the version you are looking for?
...
+**New (self-built)**
+
+```ts
+import { Faker, de_CH, de, en } from '@faker-js/faker';
+
+export const customFaker = new Faker({
+  locale: [de_CH, de, en],
+});
+```
+
```
matthewmayer: i don't think this is needed for the migration guide, this is advanced usage.

test/helpers.spec.ts
```diff
@@ -904,15 +900,25 @@ describe('helpers', () => {
...
+        it('should support deprecated module aliases', () => {
+          expect(faker.definitions.location.city_name).toContain(
+            faker.helpers.fake('{{address.cityName}}')
           );
           expect(faker.definitions.person.first_name).toContain(
             faker.helpers.fake('{{name.firstName}}')
           );
         });
 
+        // TODO @ST-DDT 2023-01-17: Restore this test when the definitions proxy is restored.
```
Shinigami92: What does this mean?
Should there be opened an issue for that so we do not forget it?

docs/guide/upgrading.md
```diff
@@ -14,6 +14,36 @@ Not the version you are looking for?
...
+In order to facilitate better and easier locale fallback mechanics, we removed the methods to change the locales on existing `Faker` instances.
+
+**Old**
+
+```ts
+faker.setLocale('de_CH');
+faker.locale = 'de_CH';
+faker.fallbackLocale = 'en';
+```
+
```
matthewmayer, line 36: I think we should also explain how to handle a situation where you previously had one faker instance and were changing the locale dynamically at runtime.

For example if you were doing a mix of locales like

```js
for (let user of users) {
  const lang = faker.helpers.arrayElement(["de","en","fr"]) 
  faker.locale = lang
  user.name = faker.person.fullName()
}
```

docs/guide/localization.md
```diff
@@ -39,66 +44,72 @@ However, in most cases, using a specific locale will be beneficial in the long t
...
+| `en_IN`       | English (India)           | `fakerEN_IN`       |
+| `en_NG`       | Nigeria (English)         | `fakerEN_NG`       |
+| `en_US`       | English (United States)   | `fakerEN_US`       |
+| `en_ZA`       | English (South Africa)    | `fakerEN_ZA`       |
+| `es`          | Spanish                   | `fakerES`          |
+| `es_MX`       | Spanish (Mexico)          | `fakerES_MX`       |
+| `fa`          | Farsi                     | `fakerFA`          |
+| `fi`          | Finnish                   | `fakerFI`          |
+| `fr`          | French                    | `fakerFR`          |
+| `fr_BE`       | Français (Belgique)       | `fakerFR_BE`       |
```
matthewmayer, line 95: "French (Belgium)" for consistency?

README.md
```diff
@@ -131,18 +130,26 @@ console.log(
...
+
+```ts
+import { Faker, de, de_CH } from '@faker-js/faker';
 
-Please note: not every locale provides data for every module. In our pre-made locales, we fallback to English in such a case as this is the most complete and most commonly used language.
+export const faker = new Faker({
+  locale: [de_CH, de],
+});
+```
```
matthewmayer: i think this more advanced usage can be covered in localization page only?

test/locale-imports.spec.ts
```diff
@@ -1,21 +1,26 @@
...
     it(`should be possible to directly require('@faker-js/faker/locale/${locale}')`, () => {
       // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const { faker } = require(`../dist/cjs/locale/${locale}`);
+      const { faker } = require(`../dist/cjs/locale/${locale}`) as {
+        faker: Faker;
+      };
 
       expect(faker).toBeDefined();
-      expect(faker.locale).toBe(locale);
+      expect(faker.string.alpha()).toBeTypeOf('string');
```
Shinigami92, line 14: string.alpha generates values just out of algorithm
Maybe we should use a method that uses locales/definitions so we ensure that also that is loaded correctly
We should also make it somehow safe and ensure that the correct locale was loaded (what the previous test case did)

README.md
```diff
@@ -131,18 +130,26 @@ console.log(
...
-The default language locale is set to English.
-
-Setting a new locale is simple:
+The main `faker` instance uses the English locale.
+But you can also import instances using other locales.
 
 ```ts
-// sets locale to de
-faker.locale = 'de';
+import { fakerDE as faker } from '@faker-js/faker';
```
matthewmayer, line 140: can we also include a CommonJS example?

matthewmayer (changes requested):

test/system.spec.ts
```diff
@@ -422,15 +418,10 @@ describe('system', () => {
...
-
       it('#770', () => {
-        faker.seed(5423027051750305);
-        faker.setLocale('sk');
-        faker.system.commonFileName('xml');
-        faker.system.commonFileName('xml');
+        fakerSK.seed(5423027051750305);
+        fakerSK.system.commonFileName('xml');
+        fakerSK.system.commonFileName('xml');
       });
```
Shinigami92, line 425: Not related to this PR, but we should try to find out what this test was purposed for and why it currently doesn't test anything anymore
Currently there is no `expect` call and therefor it could potentially be remove or fixed/reintroduced

README.md
matthewmayer: i guess 
```
const { fakerDE : faker } = require("@faker-js/faker")
```
would be the equivalent?

matthewmayer (dismissed):

docs/guide/localization.md
ST-DDT: That is unrelated to this PR/change. Would you like to create a separate PR for this?

test/helpers.spec.ts
ST-DDT: https://github.com/faker-js/faker/issues/893

test/locale-imports.spec.ts
ST-DDT: This test should only test that there is a faker instance at that import, but I can call some other method as well.

Shinigami92: No, I would also like to have it covered that it is not only a Faker instance but the one that is required/imported
Maybe access the `faker.definition.title`? Or something other that identifies the top-base for the configured locale

docs/guide/upgrading.md
ST-DDT: Lets discuss this in the next team meeting.

docs/guide/upgrading.md
ST-DDT: Not sure whether this is worth mentioning.

````ts
for (let user of users) {
  const faker = faker.helpers.arrayElement([fakerDE, fakerEN, fakerFR ]);
  user.name = faker.person.fullName()
}
````

File: README.md (hunk 21 above)
Comment (ST-DDT) [reply README.md]: Lets discuss this in the next team meeting.

File: docs/guide/upgrading.md (hunk 19 above)
Comment (Shinigami92) [reply docs/guide/upgrading.md:36]: > ```ts
> for (let user of users) {
>   const faker = faker.helpers.arrayElement([fakerDE, fakerEN, fakerFR]);
>   user.name = faker.person.fullName()
> }
> ```

Actually this is genius! Because now this does not (theoretically) load in all the locales but only DE, EN and FR which then comes down to 1-2 mb instead of 5 mb

Comment (matthewmayer) [reply docs/guide/upgrading.md:36]: I think this is a nice example to add. It wasn't obvious to me that this is the idiom you would use. It makes it clear instead of "one faker object you mutate" it's now "multiple faker objects".

File: docs/guide/upgrading.md (hunk 17 above)
Comment (ST-DDT) [reply docs/guide/upgrading.md]: **Team decision**

Mention multiple fallbacks that are now supported.

Add this example to the old to show the diff:

````ts
const customFaker = new Faker({
 locale: 'de_CH'
 fallbackLocale: 'de',
 locales: { de_CH, de, en }
});
const a = customFaker.firstName();
customFaker.locale = 'en'
const b = customFaker.smiley();
````

File: docs/guide/upgrading.md (hunk 19 above)
Comment (ST-DDT) [reply docs/guide/upgrading.md:36]: **Team decision**

Add it to the documentation.

File: README.md (hunk 21 above)
Comment (ST-DDT) [reply README.md]: **Team Decision**

We want to keep this.

**Comment by ST-DDT** [comment]
I tried to use tables same as the Contribution guide, but that doesn't look good at all.

grafik
grafik

File: docs/guide/upgrading.md (hunk 19 above)
Comment (ST-DDT) [reply docs/guide/upgrading.md:36]: Done

File: docs/guide/upgrading.md (hunk 17 above)
Comment (ST-DDT) [reply docs/guide/upgrading.md]: Done

File: docs/guide/localization.md (hunk 20 above)
Comment (ST-DDT) [reply docs/guide/localization.md:95]: Done: https://github.com/faker-js/faker/pull/1884

File: test/system.spec.ts (hunk 24 above)
Comment (ST-DDT) [reply test/system.spec.ts:425]: Moved to own issue: https://github.com/faker-js/faker/issues/1885

File: test/locale-imports.spec.ts [hunk 25]
```diff
@@ -1,21 +1,27 @@
...
+      expect(faker.definitions.title).toBe(allLocales[locale].title);
```
Shinigami92, line 15: > `expect(faker.definitions.title).toBe(allLocales[locale].title);`

at least this is a little compromise
could we add it for the other 3 cases as well?

ST-DDT: Done

Shinigami92 (dismissed):
Lets move on, release new alpha and iterate in further PRs if needed

README.md
```diff
@@ -130,18 +134,30 @@ console.log(
 
 Faker has support for multiple locales.
 
-The default language locale is set to English.
-
-Setting a new locale is simple:
+The main `faker` instance uses the English locale.
+But you can also import instances using other locales.
```
matthewmayer, line 136: Should we add a warning like in #1886 so as not to confuse people who read this on the README but are using the stable v7?

ST-DDT: #1886 is scheduled to arrive after this to include the warnings.

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
...
 
+### Removed ability to change the locale on existing `Faker` instances
+
+In order to facilitate better and easier locale fallback mechanics, we removed the methods to change the locales on existing `Faker` instances.
+Now, we expose specific faker instances for each locale that you can use:
+
+**Old**
+
+```ts
+faker.setLocale('de_CH');
```
matthewmayer, line 31: i think the "Old" should also include the import statement to make the necessary changes clearer

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
 
 ## Breaking changes
 
+### Removed ability to change the locale on existing `Faker` instances
+
+In order to facilitate better and easier locale fallback mechanics, we removed the methods to change the locales on existing `Faker` instances.
```
matthewmayer, line 23: i think we could explicitly mention:

If you are using Faker with just the default `en` locale, you don't need to make any changes to your code.

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
...
+
+```ts
+const customFaker = new Faker({
+  locale: 'de_CH', // the expected locale
+  fallbackLocale: 'de', // ensure we have a German fallbacks for addresses
+  locales: { de_CH, de, en },
+});
+const a = customFaker.firstName();
+customFaker.locale = 'en'; // neither 'de_CH' nor 'de' have smileys
+const b = customFaker.smiley();
```
matthewmayer: what is `faker.smiley()`? you mean `faker.internet.emoji()`?

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
...
+
+**New**
+
+```ts
+import { Faker, de_CH, de, en } from '@faker-js/faker';
+
+// same as fakerDE_CH
+export const customFaker = new Faker({
+  // Now multiple fallbacks are supported
+  locale: [customNames, customAddresses, de_CH, de, en, global],
```
matthewmayer: what is `global` here?

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
...
+  const lang = faker.helpers.arrayElement(['de', 'en', 'fr']);
+  faker.locale = lang;
+  user.name = faker.person.fullName();
+}
+```
+
+**New**
+
+```ts
+for (let user of users) {
```
matthewmayer, line 93: should include the import here

docs/guide/upgrading.md
```diff
@@ -14,6 +14,78 @@ Not the version you are looking for?
...
+});
+const a = customFaker.firstName();
+customFaker.locale = 'en'; // neither 'de_CH' nor 'de' have smileys
+const b = customFaker.smiley();
+```
+
+**New**
+
+```ts
+import { Faker, de_CH, de, en } from '@faker-js/faker';
```
matthewmayer: shouldnt this be fakerDE_CH, fakerDE, fakerEN?

matthewmayer (changes requested):

docs/guide/upgrading.md
Shinigami92: I assume this is a preparation for the followup PR:
- #1748

we just dont want to do too much in one PR as this PR is already really big

docs/guide/upgrading.md
ST-DDT: No, as these refer to the locales not the related faker instances.

README.md
```diff
@@ -56,8 +56,13 @@ npm install --save-dev @faker-js/faker
 ## 🪄 Usage
 
 ```ts
+// ESM
-// import { faker } from '@faker-js/faker/locale/de';
+// import { fakerDE as faker } from '@faker-js/faker';
```
xDivisionByZerox: Since we have the same imports on the "Localization" section I don't see the need to include them here as well.

docs/guide/upgrading.md
matthewmayer: also `customNames`, `customAddresses` only appear in the "New" code so it's unclear what they are doing. They should either be added to Old or removed from both.

docs/guide/upgrading.md
```diff
@@ -14,6 +14,90 @@ Not the version you are looking for?
...
+
+```ts
+import { faker } from '@faker-js/faker';
+
+const customFaker = new Faker({
+  locale: 'de_CH', // the expected locale
+  fallbackLocale: 'de', // ensure we have a German fallbacks for addresses
+  locales: { de_CH, de, en },
+});
+const a = customFaker.person.firstName();
```
matthewmayer: we should probably avoid using faker.person.* in this section of the migration guide, because if you are upgrading from 7.x to 8.x you would actually be using faker.name* previously? So the "old" code doesn't actually run correctly in v7 OR v8.

matthewmayer (changes requested):

docs/guide/upgrading.md
ST-DDT: The idea behind this is to show that you can put your own data there, but I'm not sure how to show that best.

Would an `import custom from test-data.ts` + using only `custom` in the constructor be easier to understand?

matthewmayer: I think examples should ideally be self contained. But then I think this becomes too long to have in the migration guide and should be moved to the Localisation guide. 

Something like 

"If the prebuilt faker instances don't match your needs, for example if you want to add custom data, or not have English as the final fallback language, it is now possible to build a more flexible faker instance. For more details see localisation.html

ST-DDT: I removed it as I already have a link to the localization guide a few lines below this.
I extended the localization with an actual overwrite.

docs/guide/upgrading.md
ST-DDT: Done

Shinigami92 (approved):

xDivisionByZerox (approved):

Shinigami92:
@matthewmayer do you have any more reviews you want to have into the PR before merging?
Otherwise pls approve so your hard blocking `change request` goes away

matthewmayer (approved):

matthewmayer:
> @matthewmayer do you have any more reviews you want to have into the PR before merging? Otherwise pls approve so your hard blocking `change request` goes away

sorry my bad i didnt realise that blocked merge. approved now.

I have some further ideas for improving migration guide but that can be saved for a seperate PR.

Shinigami92:
🎉 

I will try to find some time the next days to address #1748 again
Thanks to all of you to bring this forward
This was the most huge "bloker" / "stone" in our way to proceed to a new future of Faker 🙂

PR #1737, opened 2023-01-15
build: escape locale import names if they match the module name

matthewpetro:
This PR is a result of a discussion located [here](https://github.com/faker-js/faker/pull/1699#discussion_r1068352915) on PR #1699 regarding locale definition imports in locale index files. Prior to this update, locale definition files could not have the same name as their module (e.g.: `src/locales/en/airline/airline.ts` could not be used) because the index.ts files that were generated for the locale would have a name conflict. This PR updates the `generateLocales.ts` script to append an underscore to the imported name of the locale definition and avoid the name conflict.


scripts/generateLocales.ts
```diff
@@ -90,16 +90,16 @@ function removeTsSuffix(files: string[]): string[] {
   return files.map((file) => file.replace('.ts', ''));
 }
 
-function escapeImport(module: string): string {
-  if (['name', 'type', 'switch'].includes(module)) {
+function escapeImport(module: string, name: string): string {
```
ST-DDT: `name` is a bit ambiguous here. Maybe use `parent`?
Also I think it might be a good idea to change the order of the params -> `(parent, module)`.

matthewpetro: Good suggestions, thanks! I've made the changes to the script.

ST-DDT (approved):

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #1746, opened 2023-01-17
fix(internet): fix invalid emails in some locales

### LINKED ISSUE #1739: Generated email addresses are sometimes invalid in certain locales

Sometimes the generated email addresses are invalid in certain locales:

- https://github.com/faker-js/faker/actions/runs/3921031720/jobs/6703049173
- https://github.com/faker-js/faker/actions/runs/3920956127/jobs/6703052613

There are actually two different errors here:

- The email address contains invalid characters
- The email address is too long

### Suggested solution

1. Replace all invalid characters
2. Limit the length of the generated e-mail address

### Alternative

_No response_

### Additional context

(There might be a different issue with the input data as well, but the method should generate valid addresses nevertheless.)

matthewmayer:
fix #1739


src/modules/internet/index.ts
```diff
@@ -84,6 +84,8 @@ export class InternetModule {
       );
 
     let localPart: string = this.userName(firstName, lastName);
+    // The local part of an email address is limited to 64 chars per RFC 3696
+    localPart = localPart.substring(0, 64);
```
ST-DDT: I think we should limit usernames as well.

Also I think we shouldn't use 64 as the max.
I'm not yet whether or not 64 as a fixed max is bad in general (as I have never seen such a long username) or if we should use a ranged max length (e.g. 5-64) to ensure that we have different lengths for all locales.

What do you think?

ST-DDT:
FFR: This fixes only one of two problems in #1739 .

src/modules/internet/index.ts
matthewmayer: 64 is only likely to be hit in very obscure cases, such as Maldivian (dv) and Georgian (ge) where we have longish names in a script with high Unicode char values and no transliterations. I think we could reduce it to say 50 for these exceptional cases (and in the long run add better transliterations).

I ran a quick script to look for invalid emails and emails with very long usernames, it only found `dv` and `ge` plus a few errors in `el` and `fa` which appear to be errors in the original data (names with parentheses inside) which i can fix manually in the source data.

```js
const { faker } = require("@faker-js/faker")
const validator = require("validator")
let allLocales = Object.keys(faker.locales)
for (let l of allLocales) {
    faker.locale = l
    for (let i = 0;i<1000;i++) {
        let email = faker.internet.email()
        if (!validator.isEmail(email) || email.split("@")[0].length > 50) {
            console.log(l, email)
        }
    }
    
}
```

matthewmayer:
> FFR: This fixes only one of two problems in #1739 .

do you have an example of an invalid email address which is still generated?

matthewmayer:
> remove names with () in them

ah looks like that was the problem in `el` which i fixed in this commit https://github.com/faker-js/faker/pull/1746/commits/58dfab49c0fc663a600adeb75c4e9a1b0ac34812

ST-DDT:
Can you also add a filter that removes all invalid characters from the local part, so it still generates a valid email even if the given name/firstname are invalid? Please also add a test for both too long names and invalid characters.

src/modules/internet/index.ts
```diff
@@ -84,6 +84,13 @@ export class InternetModule {
       );
 
     let localPart: string = this.userName(firstName, lastName);
+    // Strip any special characters from the local part of the email address
+    // This could happen if invalid chars are passed in manually in the firstName/lastName
+    localPart = localPart.replace(/[^A-Za-z0-9._+\-\']+/g, '');
+
+    // The local part of an email address is limited to 64 chars per RFC 3696
+    // We limit to 50 chars to be more realistic
+    localPart = localPart.substring(0, 50);
```
ST-DDT, line 100: Should we move this to the username method instead?

ST-DDT (dismissed):

src/modules/internet/index.ts
matthewmayer: i feel the limitation on email addresses length is specific to email, so it should be there. Username could be something like a forum username which could in theory be longer. But i dont have very strong feelings either way.

ejcheng (dismissed):

src/modules/internet/index.ts
```diff
@@ -84,6 +84,13 @@ export class InternetModule {
       );
 
     let localPart: string = this.userName(firstName, lastName);
+    // Strip any special characters from the local part of the email address
+    // This could happen if invalid chars are passed in manually in the firstName/lastName
+    localPart = localPart.replace(/[^A-Za-z0-9._+\-\']+/g, '');
```
xDivisionByZerox: This replacement gets rig of a lot of characters that are valid according to [RFC 3696](https://www.rfc-editor.org/rfc/rfc3696#section-3).

> Without quotes, local-parts may consist of any combination of
 alphabetic characters, digits, or any of the special characters
> ! # $ % & ’ * + - / = ? ^ _ ‘ . { | } ˜

test/internet.spec.ts
```diff
@@ -156,6 +156,25 @@ describe('internet', () => {
...
+          // e.g. ElizabethAlexandraMaryJaneAnnabelVictoria.SmithJon@yahoo.com
+          expect(email).toSatisfy(validator.isEmail);
+          const localPart = email.split('@')[0];
+          expect(localPart.length).toBeLessThanOrEqual(50);
+        });
+
+        it('should return a valid email for names with invalid chars', () => {
+          const email = faker.internet.email('Matthew (Matt)', 'Smith');
+          // should strip invalid chars
+          // e.g. MatthewMatt_Smith@yahoo.com
```
xDivisionByZerox, line 174: A valid value could also be something like this: `"Matthew Matt_Smith"@yahoo.com`.
As stated in [RFC 3696 Section 3](https://www.rfc-editor.org/rfc/rfc3696#section-3):
> Blank spaces may also appear, as in Fred\ Bloggs@example.com [...] "Fred Bloggs"@example.com

What I want to say is, do we support spaces? Do we WANT to support spaces? Since it is in theory valid, I (as a customer) would like to see this edge case pop up in one of my test cases. But that might only be me.

ST-DDT: I have never seen e-mail addresses with spaces.

src/modules/internet/index.ts
matthewmayer: I kept only the chars that realistically appear in email addresses

Other chars, while theoretically allowed by the RFC are frequently rejected by other software, so I think it causes more harm than good to include them.

test/internet.spec.ts
matthewmayer: If you want to test full compliance to the RFC I think there are better ways to do that. Faker should return realistic data not "data we think might break stuff" - like we could have a FirstName which is 10000 characters long to check if you're performing validation on your names. But I think that's out of the scope of faker.

Shinigami92: Even when I would implement a real world app using email, I would forbid using a space inside an email 🙅
Kill it with 🔥 
Totally agree using `_` instead

src/modules/internet/index.ts
xDivisionByZerox: Understood 👍 
In that case, I'd argue that we don't require the apostrophe character (`'`) either. At least I have never seen an email with one.

matthewmayer: I tend to agree. It's theoretically allowed but say gmail doesn't let you create an email with '

https://support.google.com/mail/answer/9211434?hl=en

And it seems to causes issues in office365
https://learn.microsoft.com/en-us/answers/questions/325540/emails-with-an-apostrophe-in-the-address-fail-to-d

xDivisionByZerox: Currently, the RegEx allows `'` to pass. 
```ts
"abcdABCD1234!?:-._'".replace(/[^A-Za-z0-9._+\-\']+/g, '');
// abcdABCD1234-._' <-- still here
```

I'm confused^^ or was that your way of saying "I will remove the apostrophe from the RegEx"?

matthewmayer: Yes, I agree to remove it from the regex

ST-DDT (approved):

xDivisionByZerox (approved):

Shinigami92 (approved):
LGTM

PR #1748, opened 2023-01-17
feat: add base locale

### LINKED ISSUE #1441: Add global locale

Shinigami92:
closes #1441

this adds a global locale that can later be used as the fallback for everything
it contains just generic data not related to a locale specifically, like e.g. css_space which are the same in every locale in the world


Shinigami92:
As we have currently not the possibility to declare multiple stacked `localeFallback`, this PR need to be done in conjunction with

- #1735

there we can finally use e.g. `locale: [en_IN, en, global],`

src/definitions/color.ts
```diff
@@ -13,4 +13,14 @@ export type ColorDefinitions = LocaleEntry<{
...
   space: string[];
+
+  /**
+   * Functions supported by CSS to produce color.
+   */
+  css_function: string[];
+
+  /**
+   * Color space names supported by CSS.
+   */
```
Shinigami92: I'm thinking if we should add a line to indicate that this is intended for the global data
theoretically it is/was possible to call `colorByCSSColorSpace` with `cssSupportedSpace`, but actually `cssSupportedSpace` returns string and therefore needs to be checked manually anyways

Shinigami92 (commented):
This is (IMO) currently blocked by #1735, but @ST-DDT please let me know if this is going in the right direction as you expected

src/locale/global.ts
```diff
@@ -0,0 +1,17 @@
...
+import en from '../locales/en';
+import global from '../locales/global';
+
+export const faker = new Faker({
+  locale: 'global',
+  localeFallback: 'en',
+  locales: {
+    global,
+    en,
+  },
```
Shinigami92: This needs to be
```suggestion
  locale: [global],
```
after #1735 got merged

src/modules/color/index.ts
```diff
@@ -1,34 +1,21 @@
...
+  | 'hwb'
+  | 'cmyk'
+  | 'lab'
+  | 'lch';
+export type CSSSpace =
+  | 'sRGB'
+  | 'display-p3'
+  | 'rec2020'
+  | 'a98-rgb'
+  | 'prophoto-rgb';
```
ST-DDT: AFAICT these types aren't used anywhere.
If we actually need them inside of methods, then I would keep them as an implementation detail and not global data.
Having a matching return type for the other methods option sounds quite useful.

src/definitions/color.ts
ST-DDT: I wouldn't explicitly mark them as such, because then users might feel discouraged to replace them with their own set.
However, a method takes it as an input parameter so I don't think that the method should use the global data.

(same as we don't put SexTypes enum values into the global list.

Shinigami92: So should we convert them to enums? :thinking:

Shinigami92:
Okay will revert the color, and potentially convert the color related things to enums in another PR
Like SexType.

But I can proceed with stuff like emoji and other already locales that are in `en` right now (?)

src/definitions/color.ts
ST-DDT: Sounds like a good idea. But maybe in a different PR.

Shinigami92: yes yes totally in another PR :ok_hand:

Shinigami92:
TBD

Maybe move to global?
- finance->currency
- science->chemicalElement
- science->unit

Maybe move to another locale?
- location->state(_abbr) -> en US

Maybe purge?
- business->credit_card_types
- internet->avatar_uri

ST-DDT (dismissed):

Shinigami92:
@ST-DDT thx for you approval, but before #1735 is not merged, we cannot proceed with this PR as well

ST-DDT:
https://github.com/faker-js/faker/pull/1735 is merged, so this is no longer blocked.

Most/All(?) of the merge conflicts can be solved by running `pnpm run preflight`.

ST-DDT:
TODO:

- [ ] Add `global` to the generated fallback locales in the Faker constructor args and adjust the `generate:locales` script to handle the global locale properly.

src/locale/global.ts
ST-DDT: or just `locale: global`

src/locale/global.ts
```diff
@@ -0,0 +1,17 @@
...
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+
+import { Faker } from '../faker';
+import en from '../locales/en';
+import global from '../locales/global';
+
+export const faker = new Faker({
+  locale: 'global',
```
matthewmayer: This seems backwards. Shouldn't en fall back to global rather than vice versA?

matthewmayer (changes requested):

src/locale/global.ts
ST-DDT: This is a draft that needs to be rebased.
You commented on the old part.

matthewmayer: Ok cool. So every locale will fall back to global at the end?

matthewmayer: ...but you would never actually choose global as your locale?

ST-DDT: > Ok cool. So every locale will fall back to global at the end?

The pre-built faker instances yes.

> ...but you would never actually choose global as your locale?

There will be a fakerGLOBAL instance.

src/modules/random/index.ts
```diff
@@ -133,6 +135,14 @@ export class RandomModule {
...
       } catch {
         // catch missing locale data potentially required by randomWordMethod
+        iteration++;
+
+        if (iteration > 100) {
+          throw new FakerError(
+            'No matching word data available for the current locale'
+          );
+        }
+
```
Shinigami92: TODO: This fix need to be outsourced to its own PR

src/modules/color/index.ts
```diff
@@ -9,7 +11,6 @@ export const CSS_SPACES = [
   'rec2020',
   'a98-rgb',
   'prophoto-rgb',
-  'rec2020',
```
Shinigami92: Do this in another PR

matthewmayer:
i don't think color/space should be moved to global? Those are English names, not identifiers. For example in French you say RVB not RGB https://fr.wikipedia.org/wiki/Rouge_vert_bleu

matthewmayer:
Similarly finance/currency has English names of currencies. A "pound" is a "livre" in French.

matthewmayer:
also science/chemicalElement and science/unit?

Shinigami92:
> Similarly finance/currency has English names of currencies. A "pound" is a "livre" in French.
> also science/chemicalElement and science/unit?

As far as I remember, we discussed in a previous team meeting, that we want to support the third layer of object-values
So we need another PR to e.g. provide the possibility to override `finance.currency.name`

src/modules/color/index.ts
Shinigami92: https://github.com/faker-js/faker/pull/1910

matthewmayer:
> > Similarly finance/currency has English names of currencies. A "pound" is a "livre" in French.
> > also science/chemicalElement and science/unit?
> 
> As far as I remember, we discussed in a previous team meeting, that we want to support the third layer of object-values So we need another PR to e.g. provide the possibility to override `finance.currency.name`

but then you'd have to repeat the codes in the other locale anyway, otherwise you wouldnt know which currency "livre" referred to?

Shinigami92:
> > > Similarly finance/currency has English names of currencies. A "pound" is a "livre" in French.
> > > also science/chemicalElement and science/unit?
> > 
> > 
> > As far as I remember, we discussed in a previous team meeting, that we want to support the third layer of object-values So we need another PR to e.g. provide the possibility to override `finance.currency.name`
> 
> but then you'd have to repeat the codes in the other locale anyway, otherwise you wouldnt know which currency "livre" referred to?

Ah yeah, so like `code` would serve as a "primary key"... which we originally didn't want (as far as I remember)
We have a meeting upcoming thursday

Edit: I added to the meeting notes
But theoretically it might not be an "issue" for globals, but general
So like if we want to define "localized" names for unit/elements/currency and so on, we still need to override the whole object for that locale

ST-DDT:
> As far as I remember, we discussed in a previous team meeting, that we want to support the third layer of object-values
> So we need another PR to e.g. provide the possibility to override `finance.currency.name`

I dont remember that and it would have other side effects that would negatively impact the project.
E.g. you can no longer check whether person.first_name.xyz exists for that locale.

Shinigami92:
> I dont remember that and it would have other side effects that would negatively impact the project.
> E.g. you can no longer check whether person.first_name.xyz exists for that locale.

Ah yes, you are right
I confused my own memory with `[...] we should add support for it in the fake method.`
My memory was just wrong

ST-DDT:
Looks like we didn't publish our team decision here either:

**Team Decision**

- We will keep the names of things as part of the locale data set and won't join different sources to create a single complex object.
- I don't remember and we didn't note our decision regarding the use of English names in the global locale. Does anyone else remember?

ST-DDT:
**Team Decision**

- We don't want to add/move data to the global locale that contain specifically English names/translations such as currency names.

Pro:

- @xDivisionByZerox 
- @ST-DDT 

Contra:

- @Shinigami92

src/locales/global/hacker/abbreviation.ts
matthewmayer, line 1: This file is already localized e.g. in `ru` 
https://github.com/faker-js/faker/blob/next/src/locales/ru/hacker/abbreviation.ts

matthewmayer (changes requested):

src/locales/global/hacker/abbreviation.ts
Shinigami92: globales are always allowed to getting overridden, that is literally the reason why we moved stuff to a global locale instead of using static code

ST-DDT (dismissed):

matthewmayer (changes requested):
Needs documentation in various places
- localisation page
- migration page 

Explain that if you are building custom locale that final should usually be the final fallback

Explain that you shouldn't use the fakerGLOBAL instance alone?

ST-DDT:
> Explain that if you are building custom locale that final should usually be the final fallback

Yes.

> Explain that you shouldn't use the fakerGLOBAL instance alone?

IMO there is no need for it.
If you think using global is a good choice for using it, then you sure can do that.
You will definitely get a very small bundle size when using it.
We just have to clarify that is intended as a fallback. (maybe add an explanation to the bottom of the locales list in the docs.

matthewmayer:
> > Explain that you shouldn't use the fakerGLOBAL instance alone?
> 
> IMO there is no need for it. If you think using global is a good choice for using it, then you sure can do that. You will definitely get a very small bundle size when using it. We just have to clarify that is intended as a fallback. (maybe add an explanation to the bottom of the locales list in the docs.

i think without further explanation/documentation `fakerGLOBAL` sounds like it contains *all* data for all the locales, not "a small subset of data which is conmmon to all locales". Could also consider naming it something different like "base" or "root"?

Shinigami92:
Here is a vote: https://strawpoll.com/polls/e6Z28p39MnN

ST-DDT:
**Poll Decision**

- ``base`` has won

matthewmayer:
Rename the PR title also?

image

ST-DDT (dismissed):

docs/guide/localization.md
```diff
@@ -44,7 +44,7 @@ If our built-in faker instances don't satisfy your needs, you can build your own
```
matthewmayer, line 44: I think we could add a note here about what the base locale is.

src/locales/base/hacker/abbreviation.ts
matthewmayer, line 1: this is localized in some languages e.g. `ru`

docs/guide/localization.md
Shinigami92: Please move on and use the suggestion feature in GitHub

image

src/locales/base/hacker/abbreviation.ts
Shinigami92: That was resolved here: https://github.com/faker-js/faker/pull/1748#discussion_r1140636887

The team agreed with me that everything is correct: the base locale can be overridden if wanted, and so is here in ru locale

docs/guide/localization.md
Shinigami92: @matthewmayer please suggest something or lead me in the right direction

matthewmayer: What's the logic on what information is included in the base locale? My understanding is that it would be only data which is not locale specific. But that seems contradicted by the fact that hacker.abbreviation is localised. Whatever the logic is should be documented here.

Shinigami92: @matthewmayer please have a look into https://github.com/faker-js/faker/pull/1748/commits/f5901e4bafc6ab633d7ecf9a2dbb6d77e9ac94ae and tell me if this satisfies your needs

@ST-DDT please tell me if this is correct what I wrote

matthewmayer: I think that will do for now. I still find it a bit unclear why say `faker.hacker.abbreviation` is includes in `base` but not `faker.lorem.*` if the rule is "Content that should work in all languages, but could be overridden in some cases"

matthewmayer (approved):

docs/guide/localization.md
```diff
@@ -54,10 +54,16 @@ const customLocale: LocaleDefinition = {
...
-  locale: [customLocale, de_CH, de, en, global],
+  locale: [customLocale, de_CH, de, en, base],
 });
 ```
 
+- `customLocale` is your custom locale definition which will override every other fallback definitions (including the base locale if you override something from that).
+- `de_CH` is a specific locale definition that overrides some german definitions.
+- `de` is a generic german locale definition.
+- `en` is a generic english locale definition which will be used if something is not defined in the `de` locales.
+- `base` is the base locale definition which contains definitions that can be used in every language (like e.g. a set of emojis).
```
matthewmayer:
```suggestion
In this example there are 5 locales. Each of these is checked in order, and the first locale which contains the requested data will be used:

- `customLocale` is your custom locale definition which will override all other fallback definitions.
- `de_CH` is a specific locale definition that overrides some German definitions with `CH` (Switzerland) data.
- `de` is a generic `de` (German) locale definition.
- `en` is a generic `en` (English) locale definition. This is our most complete locale, so we add it to fill some gaps. Depending on your needs, you might want or not want to have it as a fallback.
- `base` is the base locale definition which contains definitions that can be used in every language (e.g. emojis).
```

matthewmayer (changes requested):

docs/guide/localization.md
ST-DDT: I would add a newline between the "introduction" and the list.
Also maybe add an additional hint to en like: `This is our most complete locale, so we add it to fill some gaps. Depending on your needs, you might want or not want to have it as a fallback especially for languages with non latin characters.`

matthewmayer: agreed although i dont think the "especially for languages with non latin characters" is needed.

matthewmayer: updated suggestion

docs/guide/localization.md
ST-DDT: I'm also not sure about lorem, but we can adjust that later if needed.

As for `hacker.abbreviation`: I also maintain a repository with Chinese users that also open issues in their native language.
What I noticed is that in most cases IT-components or libraries are still in latin letters making them one of the few parts I can actually read without a translator.

ST-DDT (approved):

matthewmayer (approved):

PR #1750, opened 2023-01-17
refactor(git): commitDate explicitly set default for refDate

xDivisionByZerox:
The default value of the `refDate` option in `faker.git.commitDate()` is documented as `now` (`new Date()`). 
This only happens due to the underlying implementation in `faker.date.recent()`. 
If the default value in `recent()` should change in the future, this change ensures that `commitDate()` still works the same.


xDivisionByZerox:
I'm not sure if this should be labeled as `chore`, since it doesn't really change anything at runtime currently.

ST-DDT (dismissed):

ejcheng (dismissed):

ST-DDT:
There seems to be a merge conflict. Could you please fix it?

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

Issue #815, opened 2022-04-08
Make possible to preview docs of a PR on demand via netlify-preview


Shinigami92:
### Clear and concise description of the problem

Provide the possibility to preview the docs of a PR via netlify preview

### Suggested solution

Ideas:
- GitHub comment message event action (BOT)
- Deploy preview when docs or scripts have changed
- other ideas?

PR #1752, opened 2023-01-18
docs: fix unwrap literal union

ST-DDT:
Restore the unwrapping of literal unions (broken due to the `readonly` array modifier/type operator).

- `value: LiteralUnion<'a' | 'b'>` -> `'a' | 'b' | string`
- `namedValue: LiteralUnion<AB>` -> `AB | string`
- `array: readonly LiteralUnion<'a' | 'b'>[]` -> `readonly ('a' | 'b' | string)[]`
- `namedArray: readonly LiteralUnion<AB>[]` -> `readonly (AB | string)[]`

This also fixes an issue with missing parenthesis in union array types.

I found these issues when testing changes for https://github.com/faker-js/faker/pull/1732, but the PRs are unrelated to each other.

<details>
<summary>Preview Diff</summary>

**Old**

grafik

**New**

grafik

</details>

This mainly has an impact on the `string` module, but also affects `weightedArrayElement` (See also #1753).

<details>
<summary>Preview Diff</summary>

grafik
->
grafik

(We might want to expand this similar to the usual options object in the future)

</details>

Note: We replace `LiteralUnions<X, Y>` with `X | Y` because it is shorter and easier to understand (#811).


scripts/apidoc/signature.ts
```diff
@@ -297,6 +304,8 @@ function typeToText(type_?: Type, short = false): string {
       )}]`;
     case 'literal':
       return formatTypescript(type.toString()).replace(/;\n$/, '');
+    case 'typeOperator':
+      return `${type.operator} ${typeToText(type.target, short)}`;
```
ST-DDT: Is the `readonly` important for the api docs or should we omit it?

scripts/apidoc/signature.ts
```diff
@@ -267,8 +267,15 @@ function typeToText(type_?: Type, short = false): string {
 
   const type = type_ as SomeType;
   switch (type.type) {
-    case 'array':
-      return `${typeToText(type.elementType, short)}[]`;
+    case 'array': {
+      const text = typeToText(type.elementType, short);
+      if (text.includes('|')) {
+        return `(${text})[]`;
```
ST-DDT: Or should we use `Array<X>` instead?

Shinigami92: I personally prefer Array for objects and [] for simple/primitives

So e.g. `Array<{ ... }>` as well as `Array<Type1 | Type2>`
but `Type1[]` as well as `string[]`

There is an eslint rule to ensure this (I even contributed to that: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/array-type.md#combination-matrix **>I<** wrote this matrix)

scripts/apidoc/signature.ts
Shinigami92: I think we can omit it (for now)
In most cases we use readonly arrays to allow to pass in readonly arrays as well as normal ones

So if you can pass a readonly array, you can also pass a normal array but not the other way around

matthewmayer:
Would there be value in having the generated HTML documents (like ./docs/.vitepress/dist) included in the repo? Then when changes like this are made, it would be easy to see the effect of the changes (effectively a snapshot test).

scripts/apidoc/signature.ts
ST-DDT: I will remove it if `short` is `true`.

ST-DDT:
> Would there be value in having the generated HTML documents (like ./docs/.vitepress/dist) included in the repo? Then when changes like this are made, it would be easy to see the effect of the changes (effectively a snapshot test).

- https://github.com/faker-js/faker/issues/815
- https://github.com/faker-js/faker/issues/1754

scripts/apidoc/signature.ts
ST-DDT: Changed

ST-DDT: It looks like I cannot remove the `readonly` just like that.
I need #1754 to avoid breaking stuff unknowingly.

ST-DDT:
Waiting for https://github.com/faker-js/faker/pull/1755
~~Then I will have another look at readonly && short.~~

ST-DDT:
For anybody interested in reviewing this:

<details>
<summary>docs diff</summary>

Documentation changes detected:

- Datatype
  - array
- ~~Date.* + Git.commitDate (#1757)~~
- Helpers
  - arrayElement
  - arrayElements
  - mustache
  - uniqueArray
  - weightedArrayElement
- Random
  - alpha
  - alphaNumeric
  - numeric
- String
  - alpha
  - alphanumeric
  - fromCharacters
  - numeric

(Parameter or return type changed, diff created by #1755)

</details>

(generated for 310c78c)

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):

PR #1753, opened 2023-01-18
docs: unwrap complex array params

ST-DDT:
Currently the `helpers.weightedArrayElement` method does not "properly" display their parameters in our documentation.
This PR displays parameter description for complex array parameters (`Array<{...}>`)

<details>
<summary>Preview Diff</summary>

**Old**

grafik

**New**

grafik

**New in combination with #1752**

grafik

</details>

Found in #1752 .


Shinigami92 (approved):
LGTM

ST-DDT:
**Team decision**

We want this, we just aren't sure about the `array[].value` notation yet, but that isn't a blocker.

src/modules/helpers/index.ts
```diff
@@ -506,14 +506,25 @@ export class HelpersModule {
    *
    * @template T The type of the entries to pick from.
    * @param array Array to pick the value from.
+   * @param array[].weight The weight of the value.
```
xDivisionByZerox, line 509: Is the `array[]` syntax actually required for the signature analyzation? I'm not sure if I read that correctly in `analyzeParameterOptions`.

ST-DDT: No, I made it up to differentiate it from `options.option`.
`[ { ... } ] vs { ... }`

(The jsdocs on method level aren't used here at all by the script, only the inline/in-object array element jsdocs itself are)

xDivisionByZerox: Ok then I did understood correctly what I read. Thank you.

xDivisionByZerox (approved):

Issue #859, opened 2022-04-21
feat(date): introduce global refDate instead of passing it every time


king-in-it:
### Describe the bug

When using faker.seed with a value, e.g. `faker.seed(123)`, faker should always give back the same values for the faker functions, but this is not the case for faker.date.past(). It uses a fixed amount of time to go back, but that's always the current date minus that fixed time. So for example calling faker.date.past() today would give a different result when calling it tomorrow.

The issue is not present when providing the parameters maxYears and refDate, e.g. `faker.date.past(30, new Date(2022,0))`. 

### Reproduction

1. Set the faker seed: `faker.seed(123);`
2. Call `faker.date.past()` (without parameters)
3. Rerunning the same code provides a different value for that same `faker.date.past()`

See https://stackblitz.com/edit/typescript-qiclti for a reproduction.

### Additional Info

@faker-js/faker version used: 6.1.2

Shinigami92:
This is because the `refDate` is default to `Date.now()` as the docs say.
Not sure if this is a bug, just the default params 🤔

king-in-it:
Indeed, it has to do with calling the `toDate` function (https://github.com/faker-js/faker/blob/main/src/date.ts#L47) with the refDate being undefined, which causes the refDate to be a `new Date()` (https://github.com/faker-js/faker/blob/b10f00c94f147c7576472345d0cee0a08dc37f4a/src/date.ts#L13). I'm using the faker.seed to give predictable results (e.g. in conjunction with Jest snapshots), but using faker.date.past() with current implementation is unpredictable...

ST-DDT:
IMO this is not a bug. With a fixed `refDate` it is predictable.

If you pass a `refDate` yourself and do `result.getTime() - refDate.getTime()` then this will always result in the same value for a given seed.
We use `now` as a default to ensure the value actually contains dates that are "soon/future" (relative to now/refDate) whenever you are calling it.
As for the jest snapshots, I think you can avoid the issue with property matchers:
https://jestjs.io/docs/snapshot-testing#property-matchers

king-in-it:
Thanks for the snapshots tip, but I don't want to set this manually in each snapshot whenever a date is involved. That's why the faker.seed is useful. Also the documentation states: "If you want consistent results, you can set your own seed", but if faker.date.past() changes each time, that is not very consistent. When a seed is set, then instead of assigning `new Date()` perhaps the Date can be initialized by other values of the faker lib to have a fixed ref date?

[...]

PR #1755, opened 2023-01-18
infra(docs): add docs diff script

### LINKED ISSUE #1754: Add api docs diff to the PR

Currently it is hard to review where in the api-docs a change takes place.

### Suggested solution

Add a CI step posting a list of changed modules/methods (in the api-docs).

MVP: A simple list

Ideas for extensions:

- Post a visual diff (collapsed by default)
- https://github.com/faker-js/faker/issues/815

### Alternative

Commit all generated docs to git.

### Additional context

This idea came up again in https://github.com/faker-js/faker/pull/1752#issuecomment-1386855952
I remember discussing this somewhere before, but couldn't find an open issue related to it.

ST-DDT:
Fixes #1754

- #1754

Adds CI notification about changed api doc pages.

- [x] Generates a hash for each module and their methods
- [x] Publishes it on our website as (hidden) json
- [x] Diff the local hash with the remote ones
- [ ] ~~Embed into CI~~ Will be done in a later PR


ST-DDT:
Blocked by #859

ST-DDT:
Here a preview for the api hash json: https://docs-diff-notification--serene-sprite-f3ef50.netlify.app/api-diff-index.json

ST-DDT:
Also blocked by https://github.com/faker-js/faker/pull/1759

scripts/apidoc/diff.ts
```diff
@@ -0,0 +1,60 @@
+import type { DocsApiDiffIndex } from './utils';
+import { nameDocsDiffIndexFile, pathDocsDiffIndexFile } from './utils';
+
+const host = 'https://next--serene-sprite-f3ef50.netlify.app';
+// const host = 'https://fakerjs.dev';
```
Shinigami92: Looks like this is currently there for test purposes
Do you think we could pass this per env parameter/variable?

scripts/apidoc/diff.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!res.ok) {
+      throw new Error(`Failed to load remote diff index: ${res.statusText}`);
+    } else {
+      return res.json() as Promise<DocsApiDiffIndex>;
+    }
+  });
+}
+
+async function loadLocal(): Promise<DocsApiDiffIndex> {
+  return import(`file://${pathDocsDiffIndexFile}`).then(
```
Shinigami92: TIL

:eyes:

scripts/apidoc/utils.ts
```diff
@@ -163,3 +195,7 @@ export function isDeprecated(signature: SignatureReflection): boolean {
 export function extractSince(signature: SignatureReflection): string {
   return extractTagContent('@since', signature).join().trim();
 }
+
+export function diffHash(object: unknown): string {
+  return createHash('sha1').update(JSON.stringify(object)).digest('hex');
```
Shinigami92: Should we use `md5` here instead of `sha1`?
IMO `md5` is totally enough for file comparison and `sha1` feels a bit like overkill.

ST-DDT:
```suggestion
  return createHash('md5').update(JSON.stringify(object)).digest('hex');
```

scripts/apidoc/diff.ts
ST-DDT: Sure, but what would you like to pass there (via env)?

This is debug code, later this will point to `https://next.fakerjs.dev` all the time.

Shinigami92: `next--serene-sprite-f3ef50.netlify.app` or `https://next--serene-sprite-f3ef50.netlify.app`

actually in a later PR e.g. `https://next--{what-ever-netlify-needs}.netlify.app`

why: so that we don't need to put it into the code as hard coded but can test stuff against its own branch/fork/whatever

ST-DDT: Ah, I see, I will change it to use parameters instead.

ST-DDT: Done

scripts/diff.ts
```diff
@@ -0,0 +1,30 @@
...
+  throw new Error(
+    `Unable to find local diff index file at: ${pathDocsDiffIndexFile}\n
+    You can run \`pnpm run generate:api-docs\` to generate it.`
+  );
+}
+
+// TODO @ST-DDT 2023-01-20: Remove this url when the diff index is available on next.fakerjs.dev
+diff(
+  `https://docs-diff-notification--serene-sprite-f3ef50.netlify.app/${nameDocsDiffIndexFile}`
+)
```
ST-DDT: ToDo prior to merging:

```suggestion
diff()
```

ST-DDT:
Waiting for https://github.com/faker-js/faker/pull/1757 to be merged.

ST-DDT:
I decided to do the PR integration in a separate PR as this will touch a lot of files currently not affected by this PR.

Shinigami92 (dismissed):

scripts/apidoc/diff.ts
```diff
@@ -0,0 +1,96 @@
...
+ * If the module name is `ADDED` or `REMOVED` it means that the module was added or removed in the local diff index.
+ *
+ * @param remoteDiffIndex The url to the remote diff index. Defaults to the next.fakerjs.dev diff index.
+ * @param localDiffIndex The path to the local diff index. Defaults to the local diff index.
+ */
+export async function diff(
+  remoteDiffIndex = `https://next.fakerjs.dev/${nameDocsDiffIndexFile}`,
+  localDiffIndex = `file://${pathDocsDiffIndexFile}`
+): Promise<Record<string, string[]>> {
+  const remote = await load(remoteDiffIndex);
```
xDivisionByZerox: Can you directly use `loadRemote` and `loadLocale` here? Since you really specifically want to load from a remote or locale system here I don't see how the "abstracting" into a general `load` method helps here, since all those "load" functions are only used in this file anyway.

scripts/apidoc/diff.ts
```diff
@@ -0,0 +1,96 @@
...
+ * The returned object contains the module names as keys and the method names as values.
+ * If the module name is `ADDED` or `REMOVED` it means that the module was added or removed in the local diff index.
+ *
+ * @param remoteDiffIndex The url to the remote diff index. Defaults to the next.fakerjs.dev diff index.
+ * @param localDiffIndex The path to the local diff index. Defaults to the local diff index.
+ */
+export async function diff(
+  remoteDiffIndex = `https://next.fakerjs.dev/${nameDocsDiffIndexFile}`,
+  localDiffIndex = `file://${pathDocsDiffIndexFile}`
+): Promise<Record<string, string[]>> {
```
xDivisionByZerox: This should propably return `Promise<Record<string, LiteralUnion<'ADDED' | 'REMOVED'>[]>>` for better type inference (?).

scripts/apidoc/diff.ts
ST-DDT: It is intended to allow us (me) to do a diff between "major" and "next" remote versions.

scripts/apidoc/diff.ts
ST-DDT: Good idea

scripts/apidoc/diff.ts
xDivisionByZerox: Maybe I set the comment position a bit unfortunate. I'm talking about this AND the next line. Why cant you inline the `loadRemote()` AND `loadLocale()` function calls here. Why do you need the `load()` method (which only checks whether `source` is an URL or not).

ST-DDT: Maybe I should rename the variables to `source` and `target` (as the variables are no longer intended for local and remote anymore)?

xDivisionByZerox: That might also be good.
Still I don't see why you need `load()`. The function is ONLY used in this two lines and each call takes a different path in `load()` itself.

scripts/apidoc/diff.ts
```diff
@@ -0,0 +1,96 @@
...
+  const remote = await load(remoteDiffIndex);
+  const local = await load(localDiffIndex);
```
xDivisionByZerox: I'll try again. Maybe this makes it more clear what I was trying to express [in my previous comment](https://github.com/faker-js/faker/pull/1755/files/b777cae0d065d3c79df903f4e41aafe53bde5f88#r1089773621).
```suggestion
  const remote = await loadRemote(remoteDiffIndex);
  const local = await loadLocal(localDiffIndex);
```

By doing this you can get rid of `load()`.

ST-DDT: I already got that, but the load method is there intentionally to allow `diff`ing different targets e.g. compare two local branches with each other or the "released"/v8 version with the "about to be released"/next version.

I added my previously in-brain-only code to the diff script so I hope it gets more obvious why it is there:

````sh
pnpm run docs:diff [Target] [Source]
pnpm run docs:diff https://docs-diff-notification--serene-sprite-f3ef50.netlify.app/api-diff-index.json
````

File: scripts/diff.ts [hunk 8]
```diff
@@ -0,0 +1,29 @@
+import { existsSync } from 'fs';
+import { diff } from './apidoc/diff';
+import { pathDocsDiffIndexFile } from './apidoc/utils';
+
+const [target, source] = process.argv.slice(2);
```
ST-DDT: Should I specify the defaults here or in (as right now):

https://github.com/faker-js/faker/blob/19e670b6de8bfffdd959b8b0adafed58e39432bd/scripts/apidoc/diff.ts#L63-L64

(and swap the source and target parameters in the source code (excluding reading from argv))

xDivisionByZerox: Yes definitely. Currently the script would fail if none is provided (so a default would not be required at a later point).

scripts/apidoc/diff.ts
xDivisionByZerox: But `target` will always be treated as an URL and source will always be treated as a locale file, or not?

scripts/diff.ts
Shinigami92: I'm not fully sure if this is better, but you could also try this kind of syntax: https://nodejs.org/docs/latest-v18.x/api/process.html#processargv

There is also `parseArgs`: https://nodejs.org/docs/latest-v18.x/api/util.html#utilparseargsconfig
but this is very new and experimental in LTS v18

scripts/apidoc/diff.ts
ST-DDT: What about:
````sh
pnpm run docs:diff https://v8.fakerjs.dev/api-diff-index.json https://next.fakerjs.dev/api-diff-index.json
````

@Shinigami92 (or anybody else): Should I remove the ability to pass both urls and local paths for both parameters?

scripts/diff.ts
ST-DDT: > Currently the script would fail if none is provided

I'm confused what you are referring to.
Theoretically it is possible to run the script without any additional args (`pnpm run docs:diff`), the only reason why it doesn't actually work right now is that this PR isn't merged yet and thus the diff index isn't published in our next docs.

> try this kind of syntax

Are you referring to `import { argv } from 'node:process';`?

> There is also parseArgs:

As well as a ton of other well tested libraries, but it is probably overkill for an effectively `array.slice(2)`.
Thanks for the hint anyway.

scripts/apidoc/diff.ts
Shinigami92: TBH... I dont really care 🤷

scripts/diff.ts
Shinigami92: > Are you referring to import { argv } from 'node:process';?

yes

> As well as a ton of other well tested libraries, but it is probably overkill for an effectively array.slice(2).
> Thanks for the hint anyway.

Exactly that was meant: wanted to give the hint and forward the knowledge that this is currently in improvement states inside Node ecosystem 👍

scripts/apidoc/diff.ts
ST-DDT: I assume I will be the only one using this. 🤷 
@xDivisionByZerox Should I remove it to keep it "simple"?

scripts/diff.ts
xDivisionByZerox: > I'm confused what you are referring to.

My bad. I read line 7 as "no source arg (or something else) => throw error" but it actually is "no source arg AND default doesn't exists => throw error".
Guess I wasn't reading careful enough. Sorry.

scripts/apidoc/diff.ts
xDivisionByZerox: TBH, due to the default values I assumed that `source` was always a locale file path. But checking the code I saw that nothing stops you from using locale or remote URLs for either argument.
Maybe that's why I got so confused over this entire thing - I simply assumed.

> @xDivisionByZerox Should I remove it to keep it "simple"?

Make it work first. Refactor later if required.
This looks fine to me now.

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):
I will approve for now to not block it, but theoretically you could think about to improve this much by using js `URL` class
Even `file://` seems to be supported, and this way you could extract and handle stuff more precise like line 38 `source.startsWith('https://')` could become `source.protocol === 'https:'`

but you could also do this in a new PR and iterate

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

Shinigami92 (approved):

ST-DDT:
Current behavior:

* Type ``[a]`` - ``[z]`` to focus the search input
* Press ``[Esc]`` to exit the input
* Press ``[Esc]`` a second time to clear the input

xDivisionByZerox (approved):

PR #1756, opened 2023-01-18
infra: apply shorter CI timeouts and retry e2e once

ST-DDT:
Some CI runners freeze when executing our CI tests and only timeout after ~6h.
This PR limits all CI pipelines to 10 minutes max (e2e 15 minutes, but the first of two attempts times out after 5 minutes).
(The longest running pipeline only takes 6:30, so there is plenty of buffer before we would hit that limit)


.github/workflows/ci.yml
```diff
@@ -129,11 +132,18 @@ jobs:
       - name: Build docs
         run: pnpm run docs:build:ci
 
-      - name: Run e2e
+      - id: e2e
+        name: Run e2e
+        run: timeout 5m pnpm run docs:test:e2e:run
+        continue-on-error: true
+
+      - name: Run e2e (2nd attempt)
```
xDivisionByZerox, line 140: Is our e2e tests failing randomly or why would we need a 2nd attempt?  This is more of a general question than anything else.

xDivisionByZerox (approved):

.github/workflows/ci.yml
ST-DDT: It hangs randomly (but there isn't a way to differentiate timeout and failure unfortunately).
For some reason, it gets unstuck when receiving the kill command, but then it's too late.

I could also write it like this:

````sh
timeout 5m pnpm run docs:test:e2e:run || pnpm run docs:test:e2e:run
````

Not sure what is easier to understand.

xDivisionByZerox: My question was more aiming at "Do we know why it fails/hangs (is there a reason at all) or does  just randomly happen?". The current syntax is readable IMO.

ST-DDT: > Do we know why it fails/hangs (is there a reason at all) or does just randomly happen?

I don't know the reason. It randomly hangs in CI. I never had that issue locally, but I rarely run the e2e locally.

Shinigami92:
Can we maybe configure just a global CI timeout of 30min?
I wont like to put a scripted retry into the CI, GitHub already has a retry button for that that can be used manually if needed, then the old vs new CI logs can also be compared with each other

Beside that theoretically Cypress has it's own retry https://docs.cypress.io/guides/guides/test-retries#Global-Configuration
But I also wont like to configure that tbh, more like exploring what the cause is and maybe talk to @JessicaSachs?

ST-DDT:
> Can we maybe configure just a global CI timeout of 30min?

If that is possible please do so. I have searched for such an option and haven't found any.

> GitHub already has a retry button for that that can be used manually if needed

And on 1/10 PRs I have to press it. The CI is supposed to reduce my work not add extra to it.
I hope, that the CI repeat fixes or at least reduces the failures.

> Beside that theoretically Cypress has it's own retry

Which is irrelevant here, since cypress or something hangs and does not fail by itself.

> exploring what the cause

That would be the optimal solution, but I don't know how.

xDivisionByZerox:
> > exploring what the cause
> 
> That would be the optimal solution, but I don't know how.

As a compromise, could we add a comment to the CI file that states, that the 2nd attempt is a workaround until we have found why the test hangs and open an issue to investigate further?
Everybody should be happy with that. (?)

Shinigami92 (approved):
After investigation more together with @ST-DDT, there is nothing better we can do right now
We are limited to what GitHub CI can currently be configured to

Shinigami92:
> As a compromise, could we add a comment to the CI file that states, that the 2nd attempt is a workaround until we have found why the test hangs and open an issue to investigate further?

Not needed anymore, I explored with @ST-DDT the GitHub CI config possibilities
The timeouts need to be set per job and there is no default
Actually this PR does two things:
1. Configuring the timeouts for all job (nice thing, I want that) :heavy_check_mark: 
2. Rerunning e2e test on failure (If it helps :shrug:)

Shinigami92:
There is no issue I can write into, so I will do it here

A Cypress CI still failed: https://github.com/faker-js/faker/actions/runs/3982897898/jobs/6827763030
And so https://github.com/faker-js/faker/pull/1756#issuecomment-1397765388 doesn't apply anymore

>> GitHub already has a retry button for that that can be used manually if needed
>
> And on 1/10 PRs I have to press it. The CI is supposed to reduce my work not add extra to it.
I hope, that the CI repeat fixes or at least reduces the failures.

ST-DDT:
> And so [#1756 (comment)](https://github.com/faker-js/faker/pull/1756#issuecomment-1397765388) doesn't apply anymore

Why?

> I hope, that the CI repeat fixes or at least **reduces the failures.**

@Shinigami92 So what is your expectation regarding the next steps and the reason why you posted this?

PR #1757, opened 2023-01-18
feat(date): introduce faker.defaultRefDate and setDefaultRefDate

### LINKED ISSUE #859: feat(date): introduce global refDate instead of passing it every time

ST-DDT:
Fixes #859

- Fixes #859
- Unblocks #1755 

This PR introduces the `faker.setDefaultRefDate` method, that can be used to configure the default `refDate` fallback value.

- fakerjs.dev: grafik (will update when the next release is created)
- next.fakerjs.dev: grafik (will update when the next PR is merged)
- dev: grafik (and it will stay like this)

---

~~I'm not very happy with both the name and implementation.~~

- For the name, I also considered `refDateFallbackSource`.
- For the implementation, I considered adding it directly to the `date` module, but then I would need to pass the fallback to the `toDate` function that is located outside.
  - Adding it directly to the `DateModule` would complicate `Faker.fork` and `Faker.derive`.
  - Adding `toDate` to the module directly causes some issues as well (IIRC)
  - My current approach is adding this as MVP and then move these kind of configs to `fakerCore.config{dateRefDateFallbackSource, helpersUniqueFallbackStore, ...}` or something similar. That way, it can share or not share these things between faker instances, configure it globally or per method (it's bound fakerCore)
- I'm unsure how I can document the ``Defaults to `refDateSource()`.`` in an easier to understand way.

Any feedback or suggestions are welcome!


Shinigami92:
I think `refDateSource` or `defaultRefDate` is a good name
I didn't thought enough yet, but the idea with having a `faker.config` sounds good to me 🤔 (later PR / iteration)

But I would like to have (for now) a `faker.defaultRefDate`
Not sure if we already want to mix it together with "tree-shakable" "global-faker-npm-module" ideas yet

IMO for now it should be next to `faker.seed` like the issue author mentioned
- #859

src/modules/date/index.ts
```diff
@@ -38,7 +42,7 @@ export class DateModule {
    *
    * @param options The optional options object.
    * @param options.years The range of years the date may be in the past. Defaults to `1`.
-   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `new Date()`.
+   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `faker.defaultRefDate()`.
```
xDivisionByZerox, line 45: Technically the default value would be `faker.defaultRefDate()()`. That's why I'm unsure if `faker.defaultRefDate()` shouldn't already return a date instance instead of the function to get one.

src/faker.ts
```diff
@@ -51,6 +51,7 @@ export class Faker {
   locales: UsedLocales;
   private _locale: UsableLocale;
   private _localeFallback: UsableLocale;
+  private _defaultRefDate: () => Date;
```
xDivisionByZerox: I feel like the actual value should be located in the `DateModule`. The `Faker` class itself can still have an API to set and get said value (so should have the `DateModule` itself), but since the value itself is only used for the `DateModule` it should be scope in that as well. 

If you think about `Faker` as a container that simply has modules attached to it one might add or remove modules as they like (not actually possible but as a general concept). If I would create a `Faker` class without a `DateModule`, why should the  value for `defaultRefDate` still be a thing?

I don't know if you understand what I'm trying to say. If you don't understand anything please give me a chance to explain.

Shinigami92: What about the git module?
https://github.com/faker-js/faker/blob/2278a7f6914db3e2ad940f4e2b32e4930c4c33c4/src/modules/git/index.ts#L93

xDivisionByZerox: If it would be used for that as well. I only had a look at the changes files in this PR, which only included the date module. My bad.

src/modules/date/index.ts
ST-DDT: No. Because it's a getter the first parentheses are omited.

src/faker.ts
ST-DDT: Currently it is not required (until  #1750 is merged)

src/modules/date/index.ts
ST-DDT: The method returns a function because all methods (toDate) require it like that.

src/faker.ts
ST-DDT: But I understand what you are refering to.
I also thought about it. IMO these kind of configs should be in a config that contains all user defined states (e.g. seed, unique stores, ref date, potentially default parameters, ...). That is explicitly meant to be shared/copied over. FFR fakerCore (config/seed...) etc.
My first attempt thus was explicitly a global to avoid breaking stuff down the road again.

src/modules/date/index.ts
xDivisionByZerox: Ok, I see. Probably got confused because you set the value via an actual function and get it via a getter.

src/faker.ts
xDivisionByZerox: Alright, I understand. Will resolve this for now. We might dig into that at a later point again (as @ST-DDT stated).

xDivisionByZerox (dismissed):

src/modules/git/index.ts
```diff
@@ -156,12 +156,12 @@ export class GitModule {
...
        * The date to use as reference point for the commit.
        *
-       * @default new Date()
+       * @default faker.defaultRefDate()
        */
       refDate?: string | Date | number;
     } = {}
   ): string {
-    const { refDate = new Date() } = options;
+    const { refDate = this.faker.defaultRefDate() } = options;
```
ST-DDT, line 164: Why cannot I call it like this?
`this.faker.defaultRefDate` returns `() => Date`?

src/faker.ts
```suggestion
  private _defaultRefDate: () => Date = () => new Date();
```

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1758, opened 2023-01-19
feat(company): add realistic Hungarian company names

BenceSzalai:
Hungarian company names were generated using Hungarian last names but using English conjunctions and legal forms (such as LLC, Inc) that are invalid for the Hungarian context.

The PR generate realistic company names that would be expected to be seen in the wild.

I've added a very typical Hungarian thing, which was very popular in the past: to include the number "2000" in the names. While it is out of fashion nowadays, many companies still have it in their names, so it makes the generated names very everyday like to have some of those.


ST-DDT (approved):
Looks good to me.

Can you please add a real world example for all patterns (e.g. a (google) link).
The same for suffixes, alternatively a (wiki) link for verification purposes.

ejcheng (approved):

BenceSzalai:
@ST-DDT 

You can see the list of real company structure abbreviations on [Wikipedia](https://hu.wikipedia.org/wiki/C%C3%A9g#A_c.C3.A9gek_csoportos.C3.ADt.C3.A1sa).
* Note that "Kht." is no longer available since 30/june/2009, it is just listed for historical reasons. I did not add it to the code.

* "Kft." & "Zrt." are analogue to LLC/Ltd. depending on jurisdiction. Both are very popular forms, there are plenty of examples out there, e.g. the state monopoly for gambling: [Szerencsejáték Zrt.](https://www.szerencsejatek.hu/) or the Hungarian branches of the well known [KPMG](https://home.kpmg/hu/hu/home/about/kapcsolat.html) (See KPMG Hungária Kft./ KPMG Tanácsadó Kft. / KPMG Global Services Hungary Kft. in the footer...)
* "Nyrt." means publicly traded company, one of the most prominent example is the Hungarian branch of Deutsche Telekom: [Magyar Telekom Nyrt.](https://www.telekom.hu/jogi-tudnivalok)
* "Bt.", "Kv." are more special forms to Hungary, not sure about analouge forms in other countries, but the former is very common. For example together with 2000, just make a Google search for ["2000 bt"](https://www.google.com/search?q=%222000+bt%22) but ["2000 Kft"](https://www.google.com/search?q=%222000+Kft%22) is very common as well.

* Another pattern "és Tsa." (means "& Co.") has plenty of examples as well on Google ["és Tsa."](https://www.google.com/search?q=%22%C3%A9s+Tsa.%22)

The rest of the generation logic is similar to the default: Use a last name to base the name on with these suffixes and additions.

Note that in many occasions people tend to omit the terminating dot and stylize some of the abbreviations e.g. "NyRT" instead of "Nyrt.", but using the dot and only capitalising the first one is supposed to be the standard way of writing these abbreviations. Although noone really cares.

Shinigami92 (approved):

PR #1759, opened 2023-01-19
refactor: reorganize apidoc scripts and reuse them for tests

ST-DDT:
The api doc scripts have grown over time and now multiple classes contain functions that import/user other files.
This PR reorganizes the methods into different files to keep the methods together.

I also added some new helper methods that simplify certain code parts and improve reuse-ability.


scripts/apidoc/apiDocsWriter.ts
```diff
@@ -139,28 +139,11 @@ export function writeApiSearchIndex(project: ProjectReflection): void {
       const apiSection: APIItem = {
         text: moduleName,
         link: moduleName.toLowerCase(),
-        headers: [],
+        headers: selectApiMethods(module).map((child) => ({
+          anchor: child.name,
+          text: child.name,
+        })),
```
ST-DDT, line 144: Previously this contained the logic for both faker.fake and faker.helpers.fake.

scripts/apidoc/typedoc.ts
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * @param module The module to extract the methods from.
+ * @returns The methods to document.
+ */
+export function selectApiMethods(
+  module: DeclarationReflection
+): DeclarationReflection[] {
+  return module
+    .getChildrenByKind(ReflectionKind.Method)
+    .filter((method) => !method.flags.isPrivate);
```
ST-DDT: This is required for the `faker.random.locale()` pseudo deletion.

scripts/apidoc/apiDocsWriter.ts
```diff
@@ -1,17 +1,17 @@
...
-  formatMarkdown,
-  formatTypescript,
-  pathDocsDir,
-  pathOutputDir,
-} from './utils';
+  extractModuleName,
+  selectApiMethods,
+  selectApiModules,
+} from './typedoc';
+
```
Shinigami92: please remove this blank line and then use _reorganize imports_ (prettier could maybe do that as well)

ST-DDT: Done

scripts/apidoc/typedoc.ts
Shinigami92: Need to wait for next team meeting

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -79,70 +61,74 @@ describe('examples and deprecations', () => {
...
-        `import { faker } from '../../../../../src';\n${examples}`
-      );
-
-      // Run the examples
-      await import(path);
+    it.each(Object.entries(methodsByName))(
+      '%s',
+      async (methodName, signature) => {
+        // Extract examples and make them runnable
+        let examples = extractRawExamples(signature).join('').trim() ?? '';
```
Shinigami92: This looks like `??` is not necessary here :thinking: 
The lhs has no nullish value, otherwise the join or trim would already result in "NPE"
I assume we can omit the rhs

```suggestion
        let examples = extractRawExamples(signature).join('').trim();
```

If this fails, I'm very interested in why :slightly_smiling_face:

Shinigami92 (commented):
Overall LGTM
But please lets wait for next meeting to discuss the `pseudo deletion` as I'm not quite sure right now if this is the right way to go

scripts/apidoc/typedoc.ts
ST-DDT: Reverted

test/scripts/apidoc/examplesAndDeprecations.spec.ts
ST-DDT: Fixed

Shinigami92 (approved):
@ST-DDT reverted the additional change, so if we decide next team meeting if that was the correct way, we can do it in a separate PR :heart: :+1:

xDivisionByZerox (approved):

Issue #1346, opened 2022-09-06
Check modules and methods regarding their name and location


ST-DDT:
### Clear and concise description of the problem

Some methods are misplaced or misnamed. The same is true for some module.

E.g. the string and number methods, or the address and name modules

### Suggested solution

Check every module and method regarding their current name/location.

ST-DDT:
See also
- #805

ST-DDT:
Blocked by
- #1341
- #1342 
- #1343 
- #1344 
- #1350

xDivisionByZerox:
This should no longer be blocked. Correct, @ST-DDT?

ST-DDT:
@xDivisionByZerox Yeah, you are right. Do you wish to tackle this?

xDivisionByZerox:
Checklist from my side:

| Module         | State |
| -------------- | ----- |
| Animal   | ✅    |
| Color    | ✅    |
| Commerce | ✅    |
| Company| <ul><li> `bs` should probably be deprecated #1860</li><li>`bsAdjective` should probably be merged into `catchPhraseAdjective` and deprecate #1860</li><li>`bsBuzz` should probably be called `buzzWord` (is that correctly written?) #1860</li><li>`bsNoun` should probably be merged into `catchPhraseNoun` and deprecated #1860</li></ul>   |
|Database|✅|
|Datatype|#1590|
|Date|<ul><li>I don't really know why we have `between` AND `betweens`</li><li>potentially `month` should be called `monthName`</li></ul>|
|Finance|<ul><li>`account` should probably be called `accountNumber` #2054</li><li>❌ `iban` should perhaps be called `IBAN`</li><li>`mask` should be renamed to `maskedNumber` or be removed #2055</li></ul>|
|Git|<ul><li>✅ `shortSha` should probably be merged with `commitSha` and configurable via an argument</li></ul>|
|Hacker|<ul><li>`ingverb` should probably be called `ingVerb` (or generally a better name)</li></ul>|
|Helpers|✅|
|Image|✅|
|Internet|<ul><li>There is `email` AND `exampleEmail` (what is an example email?)</li></ul>|
[...]

PR #1760, opened 2023-01-21
refactor(location)!: throw error if state is unknown

### LINKED ISSUE #1736: location.zipCodeByState should throw for unknown states

Currently, if the given state is not known, then a random zip code is returned.
Usually, if I pass a value to a method, then I expect it to use it and tell me if the value is wrong.

### Suggested solution

Throw if there is no data for that state.

### Alternative

Keep the method as is.

### Additional context

https://next.fakerjs.dev/api/location.html#zipcodebystate

0916dhkim:
Fixes #1736 

- #1736

### DIFF
--- a/src/modules/location/index.ts
+++ b/src/modules/location/index.ts
@@ -1,4 +1,5 @@
 import type { Faker } from '../..';
+import { FakerError } from '../..';
 import { deprecated } from '../../internal/deprecated';
 
 /**
@@ -65,7 +66,14 @@ export class LocationModule {
     if (zipRange) {
       return String(this.faker.number.int(zipRange));
     } else {
-      throw new Error(`${state} is invalid.`);
+      const stateList = Object.keys(
+        this.faker.definitions.location.postcode_by_state ?? {}
+      );
+      throw new FakerError(
+        `faker.definitions.location.postcode_by_state only contains data for [${stateList.join(
+          ', '
+        )}]`
+      );
     }
   }
 
--- a/src/modules/location/index.ts
+++ b/src/modules/location/index.ts
@@ -64,9 +64,9 @@ export class LocationModule {
     const zipRange = this.faker.definitions.location.postcode_by_state?.[state];
     if (zipRange) {
       return String(this.faker.number.int(zipRange));
+    } else {
+      throw new Error(`${state} is invalid.`);
     }
-
-    return this.zipCode();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/modules/location/index.ts
```diff
@@ -64,9 +64,9 @@ export class LocationModule {
     const zipRange = this.faker.definitions.location.postcode_by_state?.[state];
     if (zipRange) {
       return String(this.faker.number.int(zipRange));
+    } else {
+      throw new Error(`${state} is invalid.`);
```
ST-DDT: Please throw a FakerError instead.
Maybe also add a list with allowed keys and a hint where these keys are comming from.
E.g. `faker.definitions.location.postcode_by_state only contains data for [...]`

ST-DDT (commented):
If you have trouble running the tests locally, you can use `CI=true pnpm run test watch`.

matthewmayer (changes requested):
This documentation needs to be updated to reflect that an error would now be thrown. Also I think it could be make clearer this doesn't work in the default `en` locale, but only in `en_US`.

src/modules/location/index.ts
matthewmayer: you might want to have different errors for the 
1. where the locale doesn't support postcode_by_state at all (e.g. en, de)
1. where the locale does support postcode_by_state, but this is an unknown state (say XX in en_US).

ST-DDT: We will handle the first part centrally: https://github.com/faker-js/faker/issues/893#issuecomment-1397329605

- #893

matthewmayer: hmm i think this is a weird edge case though where the fallback locale (en) doesn't contain the data, which is unusual.

0916dhkim: @ST-DDT Thanks for the review 🙏 
I'll use `FakeError` instead + attach available keys.

0916dhkim: I have fixed the code to throw FakerError and updated the docstring, but I'm not sure how to make the tests pass. It looks like `en` locale is used as default in `seededTests`, and `zipCodeByState` will always throw with `en`.

Shinigami92:
The unresolved conflict brings an issue: the `state` parameter is required and therefore we have two options:

1. We just allow `zipCodeByState({ state: '...', .../* potential other options */ })`
2. We allow `zipCodeByState({ state: '...', .../* potential other options */ })` as well as `zipCodeByState('state', { /* potential other options */ })`

Once we said we only want to have method calls (outside the helpers module) that can always be called without passing any arguments. In this case state is always required otherwise it would always throw an error.

We will discuss this in an upcoming meeting.

ST-DDT:
**Team Decision**

- We want to merge zipCode and zipCodeByState into a single method instead (https://github.com/faker-js/faker/issues/1346#issuecomment-1407728227)
- It will still throw an error if the state doesn't have a config.

---

Thanks for your contribution anyway!

PR #1761, opened 2023-01-22
refactor(internet): remove weigths in userAgent

### LINKED ISSUE #216: RFC: Refactor random User Agents

The current implementation is generating custom user agents based on some old data.

### Suggested solution

The new implementation:

Create a new package "user-agent-list" that is updated on daily basis. With this we could generate realistic random user agents.

Use https://developers.whatismybrowser.com/useragents/explore/ as datasource
Create a user agent string parser to object. This will add the possibility to add a option key on the new user agent function for filtering.

```
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36
```

```
{
  "appName": "Netscape",
  "connection": {
    "downlink": 10,
    "effectiveType": "4g",
    "rtt": 0
  },
  "platform": "Win32",
  "pluginsLength": 3,
  "vendor": "Google Inc.",
  "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36",
  "viewportHeight": 660,
  "viewportWidth": 1260,
  "deviceCategory": "desktop",
  "screenHeight": 800,
  "screenWidth": 1280
}
```

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
Refactor `faker.internet.userAgent`, so that every return value is equally possible. That means that the function will no longer return values based on a "real-world" distribution.
I did this for multiple reasons:

1. They are outdated anyway. Yes, you could update the probabilities, but that would require continuous maintenance (since those numbers change a lot)
2. `Faker` is not a real-world database! We provide data which is reasonable. It is not our goal to mirrow the real world in all tiny details.
3. We don't have a probability of distribution in other functions (and if so, that's most likely not that smart as well). 
For example: In Germany, the last name "Müller" is super popular, yet it has the same return probability as any other last name in the German  locale dataset.

Could be seen as a fix for #216.


src/modules/internet/user-agent.ts
```diff
@@ -179,38 +155,34 @@ export function generate(faker: Faker): string {
...
+      'safari',
+      'opera',
+    ]);
+    const browserToOsMap = {
+      chrome: ['win', 'mac', 'lin'],
+      firefox: ['win', 'mac', 'lin'],
+      opera: ['win', 'mac', 'lin'],
+      safari: ['win', 'mac'],
+      iexplorer: ['win'],
+    } satisfies Record<Browser, OS[]>;
```
ST-DDT: The `browser` list is the same as the keys in `browsersToOsMap`.

ST-DDT: So we could use ObjectKeys to get a random key and then get the values for it.
Or alternatively we introduce `objectEntry` that returns `[key, value]` directly.

xDivisionByZerox: Good point. I'll use `heleprs.objectKey` for now.

ST-DDT: I created #1764 for helpers.objectEntry.

src/modules/internet/user-agent.ts
```diff
@@ -179,38 +155,28 @@ export function generate(faker: Faker): string {
...
     return [browser, os];
   };
 
   const randomProc = (arch: OS): string => {
     const procs = {
       lin: ['i686', 'x86_64'],
-      mac: { Intel: 0.48, PPC: 0.01, 'U; Intel': 0.48, 'U; PPC': 0.01 },
+      mac: ['Intel', 'PPC', 'U; Intel', 'U; PPC'],
       win: ['', 'WOW64', 'Win64; x64'],
     };
```
ST-DDT: Maybe add a `satisfies` here as well.

src/modules/internet/user-agent.ts
```diff
@@ -179,38 +155,28 @@ export function generate(faker: Faker): string {
...
     const archValue = procs[arch];
```
ST-DDT: IMO you can inline this value.

src/modules/internet/user-agent.ts
xDivisionByZerox: `satisfies` has no advantage here (at least for `satisfies Record<OS, string[]>`) since the result type for `procs` will still be `{ lin: string[]; mac: string[]; win: string[]; }` which is the same as explicitly typing the constant. But I can explicitly type `procs` if you wish to do so. But then inlinting makes no sense anymore.

ST-DDT: If you use `satisfies Record<OS, string[]>` then you will get an error if there is a new OS that isn't in that record, don't you?

xDivisionByZerox: Yes, but so does `const procs: Record<OS, string[]> = { [...] };`. You have no advantage from using the `satisfies` operator here is what I'm saying, as you don't care about the actually value at each key.

ST-DDT: OK

src/modules/internet/user-agent.ts
xDivisionByZerox: Done. I did inline everything and made the entire function more compact.

ST-DDT (approved):

matthewmayer:
The whole random_ua functions seem like overkill, and is hard to maintain (e.g. there's no Edge). Could it just be replaced by a random pick from a static array of popular UAs like:

https://gist.github.com/fijimunkii/952acac988f2d25bef7e0284bc63c406?permalink_comment_id=4447237#gistcomment-4447237

xDivisionByZerox:
> The whole random_ua functions seem like overkill, and is hard to maintain [...]

Most definitely. I had some other refactoring changes in mind to make this more maintainable in the future.
But we might as well replace it with a locale entry.

src/modules/internet/user-agent.ts
```diff
@@ -179,39 +155,29 @@ export function generate(faker: Faker): string {
     ]);
 
   const randomBrowserAndOS = (): [Browser, OS] => {
```
Shinigami92, line 157: While reviewing this PR I saw how we handle the typedefs for inline anonymous functions here which did not occurred to me while reviewing https://github.com/faker-js/faker/commit/09df73aa1ab97dfec4ed3a4dfdd8280f7b094a87#diff-3893a8f03afbc8fb65306679b87c097fb96ee06e745c7bc99f7debd3d280c2e5R80-R181

I would like to change these typedefs to improve readability for humans and want to split TS for JS code a bit by doing so

```suggestion
  const randomBrowserAndOS: () => [Browser, OS] = () => {
```

As an example why this help: "`-`" marks JS while "` `" marks TS
Here you can see that the JS code is splitted multiple times with TS code

```ts
const randomProc = (arch: OS): string =>
------------------------    -        ---

const randomProc: (arch: OS) => string = (arch) =>
----------------                      ------------
```

This gets worse for each additional argument and even more worse if it includes defaults

Maybe later I can search and find a lint rule to enforce that

src/modules/internet/user-agent.ts
```diff
@@ -179,39 +155,29 @@ export function generate(faker: Faker): string {
...
-    );
+    const browserToOsMap = {
+      chrome: ['win', 'mac', 'lin'],
+      firefox: ['win', 'mac', 'lin'],
+      opera: ['win', 'mac', 'lin'],
+      safari: ['win', 'mac'],
+      iexplorer: ['win'],
+    } satisfies Record<Browser, OS[]>;
+    const browser: Browser = faker.helpers.objectKey(browserToOsMap);
+    const os: OS = faker.helpers.arrayElement(browserToOsMap[browser]);
```
Shinigami92, line 166: These can be auto-inferred and are checked at the function's return boundary

```suggestion
    const browser = faker.helpers.objectKey(browserToOsMap);
    const os = faker.helpers.arrayElement(browserToOsMap[browser]);
```

src/modules/internet/user-agent.ts
```diff
@@ -179,39 +155,29 @@ export function generate(faker: Faker): string {
...
-      win: ['', 'WOW64', 'Win64; x64'],
-    };
-    const archValue = procs[arch];
-    const proc = Array.isArray(archValue)
-      ? faker.helpers.arrayElement(archValue)
-      : weightedKeyFromObject(archValue);
-
-    return proc;
-  };
+  const randomProc = (arch: OS): string =>
```
Shinigami92, line 171:
```suggestion
  const randomProc: (arch: OS) => string = (arch) =>
```

src/modules/internet/user-agent.ts
```diff
@@ -179,39 +155,29 @@ export function generate(faker: Faker): string {
...
+      (
+        {
+          lin: ['i686', 'x86_64'],
+          mac: ['Intel', 'PPC', 'U; Intel', 'U; PPC'],
+          win: ['', 'WOW64', 'Win64; x64'],
+        } satisfies Record<OS, string[]>
+      )[arch]
+    );
 
   const randomRevision = (dots: number): string => {
```
Shinigami92, line 182:
```suggestion
  const randomRevision: (dots: number) => string = (dots) => {
```

src/modules/internet/user-agent.ts
xDivisionByZerox: yes, it can auto-infer those types, but what if I made an error in the typing of `browserToOsMap` (for example forgot the `satisfied`). Suddenly my `browser` variable becomes `string`. I explicitly lef them in to ensure that I'm working with a `Brwoser` and `OS` type at this point (and if not TypeScript should tell me immediately and not laten when `string is not assignable to Browser`).

src/modules/internet/user-agent.ts
xDivisionByZerox: 1. Since this rule is not enforced, I'll take your comment as what it is - a suggestion.
2. I personally really hate this kind of declaration. IDK where you get "more readability for humans" in that (even with your explanation)

> Here you can see that the JS code is splitted multiple times with TS code

So what? Could you elaborate on why this is important?

Shinigami92: Yeah okay, if I find a lint rule to enforce that kind of style I will apply it in its own PR

src/modules/internet/user-agent.ts
Shinigami92: I could also search for a lint rule for that, so take it as not required for this PR as needed

Shinigami92 (approved):

src/modules/internet/user-agent.ts
xDivisionByZerox: I think you can take a look at [@typescript-eslint/no-inferrable-types](https://typescript-eslint.io/rules/no-inferrable-types/) if you want to enforce this.

Shinigami92: It's already configured :eyes: 
https://github.com/faker-js/faker/blob/6772b00cf19ccdbebefe023dc0f96c6576234b5b/.eslintrc.js#L46-L49

So if this does not complain, I don't complain either :shrug:

PR #1763, opened 2023-01-22
test: use basic reporter

### LINKED ISSUE #1635: Fix `pnpm run test run`
### Pre-Checks

### Describe the bug

`pnpm run test run` fails due to tests timing out.

> test/locale-imports.spec.ts > locale imports > Internal tests to cover `src/locale/*.ts` > should be possible to directly import('../src/locale/af_ZA')
Error: Test timed out in 5000ms.

````txt
  ❯ test/locale-imports.spec.ts (232) 41438ms
   ❯ locale imports (232) 41437ms
     ✓ should be possible to directly require('@faker-js/faker/locale/af_ZA')
     ✓ should be possible to directly import('@faker-js/faker/locale/af_ZA') 1009ms
     ❯ Internal tests to cover `src/locale/*.ts` (2) 5035ms
       ✓ should be possible to directly require('../locale/af_ZA')
       × should be possible to directly import('../src/locale/af_ZA') 5027ms
     ✓ should be possible to directly require('@faker-js/faker/locale/ar')
     ✓ should be possible to directly import('@faker-js/faker/locale/ar') 4299ms
     ✓ Internal tests to cover `src/locale/*.ts` (2) 805ms
     ✓ should be possible to directly require('@faker-js/faker/locale/az')
     ✓ should be possible to directly import('@faker-js/faker/locale/az') 387ms
````

### Minimal reproduction code

- `pnpm run test run`

### Additional Context

These two **previously** worked:

- `pnpm run test`
- `pnpm run test run test/locale-imports.spec.ts`

### Environment Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 Intel(R) Core(TM) i7-7820X CPU @ 3.60GHz
    Memory: 53.93 GB / 63.69 GB
  Binaries:
    Node: 18.12.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.16.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.819.0), Chromium (108.0.1462.42)
    Internet Explorer: 11.0.22621.1
```

- [ ] CJS
- [ ] ESM

### Used Package Manager

pnpm

Shinigami92:
closes #1635 

This fixes running tests locally
The previous default reporter cleared the console log everytime when to much is going on in the console
But this results in issues with to many tests and it hangs
So this new reporter do what would be done when `CI=true` would be configured and will not clear the screen anymore
IMO this is even a better UX

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -9,6 +9,7 @@ export default defineConfig({
       reporter: ['clover', 'cobertura', 'lcov', 'text'],
       include: ['src'],
     },
+    reporters: 'basic',
     onConsoleLog(log, type) {
       if (
         type === 'stderr' &&


ST-DDT:
I checked locally and the original error is no longer there on my current device.
I will try on my other devices as well later.

ST-DDT:
I tried it on my second device as well as dev containers and am no longer able to reproduce the error described in #1635 .
(Even without the fix described in this PR)

ST-DDT (approved):

ST-DDT:
It's okay anyway,

xDivisionByZerox:
This is so great! I can finally run tests locally without specifying the test file ♥

xDivisionByZerox (approved):

PR #1770, opened 2023-01-23
fix(helpers): uniform distribution in helpers.arrayElements

### LINKED ISSUE #1765: Uniform distribution in helpers.arrayElements results

I want faker.helpers.arrayElements to always return each element with the same probability.
However, in certain situations, some indexes are picked far more often than others.

For example, when sampling 10 elements from an array with 1000 elements, indexes ending in 9 are are picked ~20 times more often than indexes ending in 1. And when sampling 9 or fewer elements, indexes ending in 1 seem to be never picked at all.

### Suggested solution

The root cause of the problem is that arrayElements picks array indices using `faker.number.float()` [here](https://github.com/faker-js/faker/blob/5f5be203c0c38a485dd4837ba16281e12de5c4c6/src/modules/helpers/index.ts#L598) with the default precision of 0.01
This works fine with an array of 100 elements, but as the length grows, anomalies begin to appear.

The simplest solution would be to replace `this.faker.number.float({ max: 0.99 })` with `Math.random()`. This would, however, break some deterministic test cases.

### Alternative

We could also use something like `this.faker.number.float({ max: 0.999999999, precision: 0.000000001 })` but I'm not sure what the best number of digits is.
The precision could also be conceivably derived from the length of the given array.

If I get some suggestions from maintainers, I may be able to submit a PR.

### Additional context

Code to reproduce issue:

```
const { faker } = require('@faker-js/faker/locale/en')

const arrayLength = 1000
const ids = Array(arrayLength)

for (let i = 0; i < arrayLength; i++) {
    ids[i] = i
}

const countByMod = new Map()

    for (const id of faker.helpers.arrayElements(ids, 10)) {
        const mod = id % 10
        const count = countByMod.get(mod) ?? 0
        countByMod.set(mod, count + 1)
    }
}

console.log('nines:', countByMod.get(9), 'ones:', countByMod.get(1))
```

Some outputs:
```
nines: 2793 ones: 116
nines: 2760 ones: 126
nines: 2755 ones: 107
```

Both should be much closer to 1000, which is the case when the same code is run on a fixed version of arrayElements:

```
nines: 990 ones: 1098
nines: 1008 ones: 1013
nines: 1023 ones: 943
```

ST-DDT:
Fixes #1765

- #1765


ST-DDT:
FFR: https://jsben.ch/1w9rg

test/helpers.spec.ts
```diff
@@ -292,6 +292,38 @@ describe('helpers', () => {
 
           expect(result).toHaveLength(0);
         });
+
+        it('should be able to return the first element of the array', () => {
+          const input = [0, 1];
+          let found = false;
+
+          for (let i = 0; i < 1000 && !found; i++) {
```
xDivisionByZerox: IDK about this one... This test states that 1:999 is a reasonable distribution. I would change the test implementation to count the appearance of each element and compare them against an equal distribution value plus/minus a margin. 
In this example:
```ts
it('should return each element with a somewhat equal distribution', () => {
  const input = [0, 1];
  let foundZero = 0;
  let foundOne = 0;

    const [result] = faker.helpers.arrayElements(input, 1);
    expect(result).toBeTypeOf('number');
    if (result === 0) {
      foundZero++:
    } else if(result === 1) {
      foundOne++:
    }
  }

  expect(foundZero).toBeGreaterThanOrEqual(350);
  expect(foundZero).toBeLessThanOrEqual(650);

  expect(foundOne).toBeGreaterThanOrEqual(350);
  expect(foundOne).toBeLessThanOrEqual(650);
});
```

ST-DDT: You are right. If we use a distribution metric, then the test gets more expressive.

ST-DDT: Done

ST-DDT:
The new tests show that the measured results are off by less then 20/1000 from the expectation.
(Increased the error margin in the tests a bit to reduce the likelihood of errors)

xDivisionByZerox (dismissed):

test/helpers.spec.ts
```diff
@@ -292,6 +292,43 @@ describe('helpers', () => {
...
           expect(result).toHaveLength(0);
         });
+
+        it('should return each element with a somewhat equal distribution with 2 elements', () => {
+          const input = Array.from({ length: 2 }, (_, i) => i);
+          const occurrences = Array.from({ length: 2 }, () => 0);
+
+          for (let i = 0; i < 1000; i++) {
+            const [result] = faker.helpers.arrayElements(input, 1);
+            expect(result).toBeTypeOf('number');
```
Shinigami92: I think it's not necessary to check every single iteration if the result is valid
The test aims for something different

test/helpers.spec.ts
```diff
@@ -292,6 +292,43 @@ describe('helpers', () => {
...
+
+        it.each([10, 100, 1000])(
+          'should return each element with a somewhat equal distribution with %s elements',
+          (length) => {
+            const input = Array.from({ length }, (_, i) => i % 10);
+            const occurrences = Array.from({ length: 10 }, () => 0);
+
+            for (let i = 0; i < 1000; i++) {
+              const [result] = faker.helpers.arrayElements(input, 1);
+              expect(result).toBeTypeOf('number');
```
Shinigami92: Same here

test/helpers.spec.ts
ST-DDT: This is there because in one of my earlier attempts to fix the method I used an index range that is partially outside of the array and thus it sometimes contained undefined.
And since JS doesn't throw on undefined when you would like it to, I have to check for it explicitly.

`foo[outOfIndex]++ => NaN` and unless it happens too often, the test would still pass without error.

Should I remove it anyway?

Shinigami92: If this is the case, I'm happy to destroy your world :rofl: 

image

ST-DDT: ``result`` cannot be NaN. Only `occurrences[result]++` in line 303 can.

ST-DDT: Fixed

test/helpers.spec.ts
ST-DDT: Fixed

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1772, opened 2023-01-23
feat(helpers): add length range support in `arrayElements`

### LINKED ISSUE #1771: Add support for count ranges in helpers.arrayElements

Most of our methods that allow specifying a length or count allow specifying them as range.

E.g. https://next.fakerjs.dev/api/helpers.html#multiple

But `helpers.arrayElements` does not.

https://next.fakerjs.dev/api/helpers.html#arrayelements

### Suggested solution

Add support for `count: number, {min: number, max: number}` using [`helpers.rangeToNumber()`](https://next.fakerjs.dev/api/helpers.html#rangetonumber).

### Alternative

_No response_

### Additional context

Found in https://github.com/faker-js/faker/issues/1765

ejcheng:
closes #1771


src/modules/helpers/index.ts
```diff
@@ -561,36 +561,50 @@ export class HelpersModule {
...
+           * The maximum number of elements to pick.
+           */
+          max: number;
+        }
   ): T[] {
-    if (typeof count !== 'number') {
-      count =
+    let numElements = this.rangeToNumber(count);
+
+    if (typeof numElements !== 'number') {
```
xDivisionByZerox: This `if` statement should be obsolete - no?

test/helpers.spec.ts
```diff
@@ -281,6 +285,26 @@ describe('helpers', () => {
...
+          expect(subset.length).toBeGreaterThanOrEqual(2);
+          expect(subset.length).toBeLessThanOrEqual(4);
+
+          // Check elements
+          subset.forEach((element) => {
+            expect(testArray).toContain(element);
+          });
+
+          // Check uniqueness
+          expect(subset).toHaveLength(new Set(subset).size);
```
xDivisionByZerox: We have [a custom matcher](https://github.com/faker-js/faker/blob/8bf716225e9c5b3ce5c968511886b9141d5c918d/test/vitest-extensions.ts#L4) for this test use-case which resolves a better error message in case it fails.

```suggestion
          expect(subset).not.toContainDuplicates();
```

src/modules/helpers/index.ts
ejcheng: What if `count` is undefined? What does `rangeToNumber` do in that case?

I tried removing the if statement but the fixed length argument test case started failing.

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,50 @@ export class HelpersModule {
...
+          min: number;
+          /**
+           * The maximum number of elements to pick.
+           */
+          max: number;
+        }
   ): T[] {
-    if (typeof count !== 'number') {
-      count =
+    let numElements = this.rangeToNumber(count);
```
xDivisionByZerox: I think if the `count` parameter is nullish we should provide a custom range from `0` to `array.lenght` because otherwise `faker.number.int` will default to a max of [`Number.MAX_SAFE_INTEGER`](https://github.com/faker-js/faker/blob/250fbb71eb68aa469b6b833a61a4d23fb60ad65d/src/modules/number/index.ts#L63) wich will most likely generate a value outside of the array length and willen then be normalized to the array length itself (in ln 600).

src/modules/helpers/index.ts
xDivisionByZerox: Sorry you were to fast. [I wrote another comment](https://github.com/faker-js/faker/pull/1772#discussion_r1084297976)

src/modules/helpers/index.ts
ejcheng: `let numElements = this.rangeToNumber(count ?? { min: 0, max: array.length });`

Like this?

xDivisionByZerox: Yes, that's what I had in mind.

ST-DDT: The previous default was `1...length`

xDivisionByZerox: Oh yeah, mb. We have to stick to that.

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,56 @@ export class HelpersModule {
...
-      count =
+    if (array.length === 0) {
+      return [];
+    }
+
+    let numElements = this.rangeToNumber(
+      count ?? { min: 1, max: array.length }
+    );
+
+    if (typeof numElements !== 'number') {
```
ST-DDT: AFAICT numElements cannot be undefined at this place.

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,56 @@ export class HelpersModule {
...
+      numElements =
         array.length === 0
           ? 0
           : this.faker.number.int({ min: 1, max: array.length });
-    } else if (count > array.length) {
-      count = array.length;
-    } else if (count < 0) {
-      count = 0;
+    } else if (numElements > array.length) {
+      numElements = array.length;
```
ST-DDT: Maybe replace this with Math.min directly when assingning numElements.

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,56 @@ export class HelpersModule {
...
+    } else if (numElements < 0) {
+      numElements = 0;
```
ST-DDT: Same here. Replace this with Math.max

src/modules/helpers/index.ts
xDivisionByZerox: I would even go as far and return an empty array when length is zero or less and return a copy of the array if the length is greater than or equal to the array length.

ST-DDT: You are right.
For count >= length we can delegate to shuffle.

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,51 @@ export class HelpersModule {
...
+    }
+
+    const numElements = this.rangeToNumber(
+      count ?? { min: 1, max: array.length }
+    );
+
+    if (numElements >= array.length) {
+      return this.shuffle(array);
+    } else if (numElements <= 0) {
+      return [];
```
Shinigami92, line 603: We should add a test for this

src/modules/helpers/index.ts
```diff
@@ -561,36 +561,51 @@ export class HelpersModule {
...
+    if (array.length === 0) {
+      return [];
+    }
+
+    const numElements = this.rangeToNumber(
+      count ?? { min: 1, max: array.length }
+    );
+
+    if (numElements >= array.length) {
+      return this.shuffle(array);
```
Shinigami92, line 601: That's interesting :thinking: 

I was not aware that `arrayElements` only returns each element once
The docs do not indicate this but the implementation seems to be always like that

It's also interesting that we maybe can use just shuffle always and then do just a subslice what is already be done down in line 619
Maybe that way we can reduce the code much

ST-DDT: > The docs do not indicate this but the implementation seems to be always like that

It does, on the `count` parameter:

> When value exceeds array boundaries, it will be limited to stay inside.

ST-DDT: > It's also interesting that we maybe can use just shuffle always and then do just a subslice what is already be done down in line 619
> Maybe that way we can reduce the code much

I think it's a performance thing.

Maybe, we can add a ``shuffleRange`` parameter to the shuffle options, so we can have "the best of both worlds".
I would recommend doing that in a separate PR though.

src/modules/helpers/index.ts
ST-DDT: Yeah, you are right.
We have a test for 
array.length = 0 and count > 0 
and array.length = 0 and count = 0, 
but not array.length > 0 and count = 0

ST-DDT (approved):
Looks good to me.
Thanks for addressing this.

xDivisionByZerox (approved):

src/modules/helpers/index.ts
Shinigami92: >> The docs do not indicate this but the implementation seems to be always like that
>
> It does, on the count parameter:
>
>> When value exceeds array boundaries, it will be limited to stay inside.

But that's the point! It doesn't indicate that duplicate values are not possibly returned
It just says that the count will stay in bounds

---

I wont add a `shuffleRange` yet without someone asking from outside for it
That's not really worth it

xDivisionByZerox: To be honest, I myself was a bit surprised that it wasn't a  combination of `faker.helpers.multiple` and `faker.helpers.arrayElement`.

Shinigami92 (approved):
For now this is an improvement
Later on we could maybe do what I have in my mind and return just random elements out of the given set

PR #1773, opened 2023-01-23
docs: big api button

ST-DDT:
When visiting the fakerjs.dev website I usually want to access the `API` overview page.
However the API button is only very small in the top-bar or via Guide->Api.
This PR adds a big colorful `View API` button to the center of the screen.

grafik

Unfortunately, it isn't possible to use a different color than `Get Started` (and `View On Github`).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,6 +12,9 @@ hero:
     - theme: brand
       text: Get Started
       link: /guide/
+    - theme: brand
+      text: View API
+      link: /api/
     - theme: alt
       text: View on GitHub
       link: https://github.com/faker-js/faker
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,7 +13,7 @@ hero:
       text: Get Started
       link: /guide/
     - theme: alt
-      text: View API Index
+      text: Browse API
       link: /api/
     - theme: alt
       text: View on GitHub
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,7 +12,7 @@ hero:
     - theme: brand
       text: Get Started
       link: /guide/
-    - theme: brand
+    - theme: alt
       text: View API Index
       link: /api/
     - theme: alt
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,7 +13,7 @@ hero:
       text: Get Started
       link: /guide/
     - theme: brand
-      text: View API
+      text: View API Index
       link: /api/
     - theme: alt
       text: View on GitHub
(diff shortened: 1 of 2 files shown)


xDivisionByZerox (dismissed):

docs/index.md
```diff
@@ -12,6 +12,9 @@ hero:
     - theme: brand
       text: Get Started
       link: /guide/
+    - theme: brand
+      text: View API
```
matthewmayer: Maybe "View API Docs" would be more explicit?

ST-DDT:
```suggestion
      text: View API Index
```

docs/index.md
```diff
@@ -12,6 +12,9 @@ hero:
     - theme: brand
       text: Get Started
       link: /guide/
+    - theme: brand
```
Shinigami92: I would set this to `alt` as every other doc site do it like this:
- https://vitejs.dev/
- https://vueuse.org/
- https://vuejs.org/

```suggestion
    - theme: alt
```

ST-DDT: I like the increased visibility the current color offers.

Shinigami92: alternatively we could try something like Vuetify:
image

But for that we need to write custom CSS, I will try

Shinigami92: In theme/index.css you can use

```css
.VPHero .action:not(:last-child) a.VPButton.alt {
  border-color: var(--vp-button-brand-border) !important;
  color: var(--vp-button-brand-border) !important;
}
```

You also need to switch to `alt` so the other (default-)styles for hover and active get applied

ST-DDT: Done. Thanks for the help.

docs/index.md
```diff
@@ -12,6 +12,9 @@ hero:
     - theme: brand
       text: Get Started
       link: /guide/
+    - theme: alt
+      text: View API Index
```
xDivisionByZerox: What is an "API Index"?

ST-DDT: From https://www.google.com/search?q=index+english+definition

> an alphabetical list of names, subjects, etc. with reference to the pages on which they are mentioned.

matthewmayer: Browse API 

Might be another good choice. It emphasises there is lots of content to look through behind the button. 

```suggestion
      text: Browse API
```

xDivisionByZerox: It was more of a sarcastic question since I liked the [original suggestion](https://github.com/faker-js/faker/pull/1773#discussion_r1085284571) from @matthewmayer a lot more since I've never seen it written as "API Index". I'm not a native, but it doesn't sound right to me.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1774, opened 2023-01-23
docs: focus api search immediately

ST-DDT:
When visiting the API page I press either `Ctrl+F` or use the build-in search function to search for a specific method.
By focusing the search input directly the user can start typing immediately to apply the filtering.


docs/api/ApiIndex.vue
```diff
@@ -54,6 +57,7 @@ const filtered = computed(() => {
         <input
           type="search"
           placeholder="Enter keyword"
+          ref="apiFilter"
```
matthewmayer, line 80: can't you just use [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus), no JS required

matthewmayer (changes requested):

docs/api/ApiIndex.vue
xDivisionByZerox: `autofocus` is also much better for accessibility since screen readers get the information and can handle them. They can't do that with JS.

ST-DDT: That property only seems to work when loading the target page directly `http://localhost:5173/api/`
When opening `http://localhost:5173/` and navigating to `api`, then it does not work.

Shinigami92 (dismissed):

ejcheng (dismissed):

matthewmayer:
The downside of this change is that you can no longer easily navigate the page with up arrow/down arrow etc (i often scroll pages with Down or fast-scroll with Cmd+Down)

Shinigami92:
> The downside of this change is that you can no longer easily navigate the page with up arrow/down arrow etc (i often scroll pages with Down or fast-scroll with Cmd+Down)

That is odd :confused: 
Trading one a11y vs another is not really worth it
Does this also apply if only `autofocus` is applied?

ST-DDT:
> Does this also apply if only `autofocus` is applied?

No, even if you click in the field yourself, the navigation gets unavailable.

---

@matthewmayer How about switching to a keypress event.
If the user types a character a-z, then the input field is focussed automatically and the filter is applied.
If you exit the field, then the key-binding gets disabled.

matthewmayer:
Sounds good
> > Does this also apply if only `autofocus` is applied?
> 
> 
> 
> No, even if you click in the field yourself, the navigation gets unavailable.
> 
> 
> 
> ---
> 
> 
> 
> @matthewmayer How about switching to a keypress event.
> 
> If the user types a character a-z, then the input field is focussed automatically and the filter is applied.
> 
> If you exit the field, then the key-binding gets disabled.

ST-DDT:
Current behavior:

* Type ``[a]`` - ``[z]`` to focus the search input
* Press ``[Esc]`` to exit the input
* Press ``[Esc]`` a second time to clear the input

xDivisionByZerox (approved):
I don't think we need to think too hard about this. Our website doesn't have that much trafik and if someone doesn't like the behavior they can open an issue.

matthewmayer (approved):

Shinigami92 (approved):
This needs a maintainer approval to get merged
I don't like it, but I don't want to block it as I don't have better ideas on my side

PR #1778, opened 2023-01-24
fix(person): Improvements to Dutch name generation, in particular regarding affixes

### LINKED ISSUE #1777: Treatment of the "tussenvoegsel", the family name affix in the Dutch locale
### Pre-Checks

### Describe the bug

TL;DR:

1. Dutch affixes are currently broken
2. It doesn't make sense to generate random(affix) + random(family name without affix)
3. Current PR using middle name to track affix has advantages and disadvantages.
4. Proposal: keep affix + family names together, and remove the concept entirely from faker.

Longer version:

I'm following up from a discussion on Discord and creating an issue as requested.

Dutch family names often have an affix, a.k.a. "tussenvoegsel" (https://en.wikipedia.org/wiki/Tussenvoegsel), such as "van", "de", "van der" and a few other variants. It's very common, in this list of top 100 Dutch family names, roughly 30% have an affix (See: https://nl.wikipedia.org/wiki/Lijst_van_meest_voorkomende_achternamen_van_Nederland).

There used to be support for this in faker, but this no longer works. If I set the locale to 'nl' and generate names with `faker.name.fullName()`, you never get any name with an affix. 

The way faker used to do this, wasn't really realistic. Affixes are part of the family name. The way the old version of faker would merge a random affixless family name with a random affix isn't really realistic. "van der Meer" (lit: from the lake) is a valid Dutch name, "Kuipers" (cooper) as well. "van der Kuipers" (from the cooper???) is nonsensical.

You're probably wondering why the affix is important anyway, if they're really just part of the family name. There is one simple answer: sorting. In a list of names, the affix is ignored when sorting. Dutch administrative systems (from governmental systems to the junior football club spreadsheet) track the affix separately, solely for this reason.

New code being prepared (https://github.com/matthewmayer/faker/tree/chore/fullname-name-patterns) treats the affix as a middle name. This approach has a certain advantage: the affix can be extracted separately, which makes it easier to write some custom sorting code according to the Dutch sorting rules. But it also has disadvantages. Firstly, Dutch persons can have middle names too. Secondly, you still have the problem that the random affix doesn't necessarily match the random family name. 

Proposal:

We can take a much simpler route: make the affix part of the family name, put them all together in `locales/nl/person/last_name.ts`. 
I'm preparing a PR to this effect.

Upside is that it's really simple.
Downside is that you need to extract the affix with a regular expression if you want to sort those names.

Alternative:

The fully proper solution is to create an extra field, called 'affix' or 'tussenvoegsel' for Dutch names, and pick them from an array of affix + surname pairs. I think this solution would be overkill for the purposes of Faker. 

### Minimal reproduction code

faker.locale = 'nl';
let affixCount = 0;
for (let i = 0; i < 1000; ++i) {
	const name = faker.name.fullName();
	if (name.includes(' van ') || name.includes(' de ')) { affixCount++; }
}
console.log(affixCount); // -> prints 0

### Additional Context

Tested with faker 7.6.0

### Environment Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    Memory: 18.21 GB / 31.08 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.15.1 - ~/.n/bin/node
    npm: 8.11.0 - ~/.n/bin/npm
  Browsers:
    Chromium: 109.0.5414.74
    Firefox: 109.0
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

amarillion:
Addresses #1777, plus some minor improvements to Dutch name generation. 
I've based this on PR #1637 (fullname name patterns) assuming that will be approved as well.


ST-DDT:
https://github.com/faker-js/faker/pull/1637 has been merged. Please rebase.

amarillion:
Rebased onto next

matthewmayer (dismissed):

ejcheng (dismissed):
Please put the last name list in alphabetical order

amarillion:
I've sorted the list of names ASCIIbetically. NB this is not the Dutch sorting order.

ejcheng:
> I've sorted the list of names ASCIIbetically. NB this is not the Dutch sorting order.

Thank you! @faker-js/maintainers Should we sort the names ASCIIbetically or the Dutch way?

amarillion:
Oh, and I misspoke - it's not ASCIIbetical (because that would imply case-sensitive sorting - Z before a). It's currently sorted with regular English case-insensitive sorting.

So, in a nice example of "be careful what you wish for" we have the following sorting options:

1. case sensitive sorting (ASCIIbetical):  Kok < Visser < de Leeuw
2. case insensitive sorting (English): de Leeuw < Kok < Visser
3. case insensitive sorting (Dutch): Kok < de Leeuw < Visser

The one that is implemented with the latest commit is #2. So, which sorting method should I use? Personally, I don't mind either way.

ST-DDT:
One of the reasons why we want the sorted lists is to aid in the review process to detect similar or identical entries faster.
So I think a case intensive sort is definitely required.
For the start, we should use (2) English locale. We just have to make sure that (our tests/scripts) this also works for non-English users.
@amarillion The "Dutch case insensitive sorting" applies only to names or is the alphabet itself sorted differently from the English one? Everything that is not easily applicable to the entire locale, is a non-goal.

src/locales/nl/person/last_name.ts
```diff
@@ -1,52 +1,120 @@
+// Top 100 most common dutch last names in 2007 (including spelling variants)
+// Source: https://nl.wikipedia.org/wiki/Lijst_van_meest_voorkomende_achternamen_van_Nederland
```
ST-DDT: Can you also add one or two other sources for these names into the list, so we just don't copy the entire list 1:1?
(I would like to avoid license issues: Wiki-Commons or whatever the original list was published under != MIT)

ST-DDT (changes requested):

amarillion:
I've added two more sources for the family names.

Since the purpose of sorting is detecting duplicates, I think the current English sorting is fine and we can leave it at that.

src/locales/nl/person/last_name.ts
amarillion: Done.

ST-DDT:
> I've added two more sources for the family names.

Well, you added more links, but you didn't add any new names, so it is basically still the same list from the first source in the list.
(And thus doesn't solve the legal implications)

[`ff4dee0` (#1778)](https://github.com/faker-js/faker/pull/1778/commits/ff4dee04c2efce1746e1b8ebf2f2594c2f490d9c)

amarillion:
I don't get what you mean by legal implications. It's quite obvious that wikipedia doesn't own this list, and that the original source is in fact from https://www.meertens.knaw.nl/nfb/documenten/top100.pdf, as stated in the article. 

A list of "top 100 most occurring family names in 2007" are facts, and facts are in fact not copyrightable.

amarillion:
However, if it helps I've made a small modification to the list so it's no longer exactly the same as the source.

ST-DDT:
> A list of "top 100 most occurring family names in 2007" are facts, and facts are in fact not copyrightable.

IANAL: The facts itself might not be, but the work to compile it may be or at least, some claim it to be: https://blog.polco.us/copyright-apply-survey

Thanks for your efford to address this, but 1% differnce might not be enough.

matthewmayer:
Maybe go back to the original `nl` last_name list and just make some changes to add some more tussenvoegsel names, using your own knowledge/judgement rather than relying on a single source?

amarillion:
All right, I made a major modification to the list. It's now a fully original compilation of names, a random selection of various sources created by me personally.  Any resemblance with any other list you might find is purely coincidental, of course lists of common names are bound to have overlap.

> Maybe go back to the original nl last_name list and just make some changes to add some more tussenvoegsel names, using your own knowledge/judgement rather than relying on a single source?

Funnily enough, after adding the appropriate tussenvoegsels, nearly every name on the original list is also in the top 100 list. I only detected two differences: 'Klein', which I added, and 'Stichting', which isn't a family name.

ST-DDT (approved):

matthewmayer (approved):

ejcheng (approved):

amarillion:
Nice work guys, I appreciate your attention to detail!

PR #1780, opened 2023-01-25
docs: show source link

### LINKED ISSUE #1695: Add "Go to Source" link in api-docs (per method)

Some users might wish to read the sources of a method instead of the documentation.

### Suggested solution

Add a link to navigate to the source code of a method.

grafik

### Alternative

_No response_

### Additional context

Accessible via `signature.sources[0?].{filename, line, url}`

ST-DDT:
Fixes #1695

- #1695 

This PR ads a `Goto Source` link to each method.

grafik

The source link targets the first line of the last api signature (not the impl signature).
E.g. `https://github.com/faker-js/faker/blob/d16a4d2d/src/modules/date/index.ts#L105`

I haven't found a good way to display it yet.
If we don't find something better, we can just start with this and change the appearance later.

**Things to decide:**

- The exact text/appearance of that link?
- Whether the link targets the correct code location?
- Whether the link should target the branch or commit?
- Should the module have a source link (+since) as well?

**Things to keep in mind:**

- We have to exclude the `sourceLink` from the diff as both the commitId/branchName and the line number are likely to create lots of false positives.


test/scripts/apidoc/signature.spec.ts
```diff
@@ -25,7 +25,13 @@ describe('signature', () => {
     it.each(Object.entries(methods))('%s', (name, signature) => {
       const actual = analyzeSignature(signature, null, name);
 
-      expect(actual).toMatchSnapshot();
+      expect({
+        ...actual,
+        sourceLink: actual.sourceLink.replace(
+          /\/blob\/[0-9a-f]+\//,
+          '/blob/********/'
```
xDivisionByZerox: I know this is just the test, but could we not apply the same logic to the actual source extraction but instead of replacing it with `/blob/********/` we replace it with `/tree/v{extractMajorVersionFromPackage}`? 
Idk if it works like that.

ST-DDT: Sure, we could always refer to the release branch, but then the next docs wouldnt work.

ST-DDT: If we use commit ids here, then this would even work for PRs from external that run `docs:dev`.
If we use branch names, we are limited to ourselves to our own PRs.
If we limit it to vX and maybe next, then PRs no longer have working links.

How about we cut off this entire part before and including the commit id and prefix them only in the Vue/`a` component?

````html
<a :href="baseUrl + props.method.sourceLink" ...
````

Comment (xDivisionByZerox) [reply test/scripts/apidoc/signature.spec.ts]: > If we use commit ids here, then this would even work for PRs from external that run docs:dev.

that's a good point. I did not think about that.

> How about we cut off this entire part before and including the commit id and prefix them only in the Vue/a component?

I have no idea about Vue. What  would that change?

Comment (ST-DDT) [reply test/scripts/apidoc/signature.spec.ts]: > > How about we cut off this entire part before and including the commit id and prefix them only in the Vue/a component?
> 
> I have no idea about Vue. What would that change?

(Not sure how relevant Vue is for `baseUrl + path`)

The json doesn't contain a commit id anymore -> Neither tests nor docs:diff need to erase them.

Comment (ST-DDT) [reply test/scripts/apidoc/signature.spec.ts]: Fixed

**Comment by matthewmayer** [comment]
I think its a bit too prominent. Most people reading the docs won't use it. I'd make it light gray so its less prominent than the method name and description.

**Comment by xDivisionByZerox** [comment]
> I think its a bit too prominent. Most people reading the docs won't use it. I'd make it light gray so its less prominent than the method name and description.

We could also add a small github icon with a tooltip 🤔

**Review by Shinigami92 — DISMISSED** [review]

File: docs/.vitepress/components/api-docs/method.vue [hunk 2]
```diff
@@ -48,5 +47,33 @@ function seeAlsoToUrl(see: string): string {
...
+        :href="sourceBaseUrl + props.method.sourcePath"
+        target="_blank"
+        style="
+          font-weight: bold;
+          font-size: 1.2em;
+          display: flex;
+          align-items: center;
+        "
+      >
+        Source
```
xDivisionByZerox: This differs from the screenshot in the description. Could you update the screenshot?

matthewmayer: looks like this now.

image

Still seems a bit large to me? Maybe remove the 1.2em font size.

Also could the CSS be extracted into a stylesheet rather than have to repeat inline in every definition?

Shinigami92: It is a good idea to extract the CSS for that 👍 
Theoretically it could even be extracted to its own Vue component, than the CSS can be applied there instead of extracting it globally or something like that. (or at least put the CSS in this component it is right now)

The font-size is adjusted to other sub headlines. In the given example it is a bit bad to see as there is no See also included in the screenshot. So maybe have a look into another example

ST-DDT: @Shinigami92 Do you have time to move the styles elsewhere?

---

If you think the splitting into smaller components adds value, then you can do that as well, but I would suggest doing so in a separate PR, as that should be done for the other parts of the Method components consistently.

Shinigami92: Yes I can move CSS in some minutes/hours (currently at company work, but after that)

---

Okay I can split into more components after this got merged

Shinigami92 (dismissed):

docs/.vitepress/components/api-docs/method.vue
ST-DDT: This doesnt address the style change request.
I dont have any preferences regarding this.
@xDivisionByZerox is your thumbs up reagrding the screenshot or the style change?

xDivisionByZerox: > @xDivisionByZerox is your thumbs up reagrding the screenshot or the style change?

Was for both. "Thank you for the screenshot" and "yes its to large".

ST-DDT: @Shinigami92 ^

Shinigami92: But did @xDivisionByZerox already see it in comparison to an "See also" subheadline?

ST-DDT: FFR:

grafik

matthewmayer: Still seems too large to me. I understand you're trying to match the subheadings but none of the other ones are hyperlinked so it stands out too much.

Shinigami92: Okay, please leave a suggestion, not sure if it is helpful when I try out 10 different sizes and send 10 screenshots or more over the wire

So feel free to play around with it a bit on your own and suggest something 🙂

matthewmayer: i'd suggest

```css
a.source-link {
  font-size: 1em;
  display: flex;
  align-items: center;
  color:var(--vp-c-text-2);
}
a.source-link:hover {
  color:var(--vp-c-brand);
}
```
looks like this, with and without hover

Screenshot 2023-01-31 at 09 55 39
Screenshot 2023-01-31 at 09 55 42

Shinigami92: Adding the icon was a design decision to indicate that this link will move you out of the docs (at least open a new tab)

Using `opacity < 1` for text is a no-go, due to a11y issue

Using `color: var(--vp-c-text-2)` is different for this link to any other link, and I would like to decide against doing this

---

Currently I more or less randomly picked with my eye the font-size `em` value of `1.2`, but now I checked and `1.2em == 19px` but `1.25em == 20px == vitepress h3` which is used for all the subheadlines

Maybe we should try wrap the `a` tag with an `h3` and see if this works instead using a custom font size value

I would like to break out of VitePress Design decisions as less as possible, because every custom change would theoretically require us to check if it still works

Shinigami92: I applied my own suggestion and here is an example screenshot:



The `helpers.fake` doc part is a good example as it nearly include everything like also `See also` and `Parameters` 

I assume previously you all compared it against the `Returns`, but this is on another _leven_

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -48,5 +47,42 @@ function seeAlsoToUrl(see: string): string {
         </li>
       </ul>
     </div>
+
+    <div>
+      <h3 v-if="props.method.sourcePath">
```
ST-DDT: Maybe add the if to the containing div?

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -48,5 +47,42 @@ function seeAlsoToUrl(see: string): string {
...
+        <a
+          v-if="props.method.sourcePath"
```
ST-DDT: Duplicate if

ST-DDT (commented):
One of the concerns mentioned is the color.
How about removing the color (but not the link) from the text and only have the color on the icon?

docs/.vitepress/components/api-docs/method.vue
Shinigami92: you are totally right 🤦

Shinigami92:
> One of the concerns mentioned is the color. How about removing the color (but not the link) from the text and only have the color on the icon?

I did not apply anymore any color, it's coming from VitePress now fully

docs/.vitepress/components/api-docs/method.vue
matthewmayer: What about having an unlinked `<h3>Source</h3>` to match the other headings, and then a smaller linked `<a href="">View source</a> ` underneath? Then it won't need any custom styling, and it doesn't seem to be floating in space

Shinigami92: Would like @ST-DDT opinion about that
He wanted first to put it also somehow in the `See Also` section, but that is autogenerated and has its own v-for and stuff, and the if would get totally messy as these are two different things (at least technically)

ST-DDT: Having `Source` and `View Source` directly next to each other looks weird IMO.

grafik

matthewmayer: You could change the link name to View in Github or View code instead if you don't want to repeat "source"

ST-DDT: Lets do that then.
With or without ul>li?

matthewmayer: i think with the bullet, see also also has bullets even when there's only 1 item

ST-DDT: Done

ST-DDT:
<details>
<summary>Preview</summary>

grafik
grafik

</details>

Shinigami92 (dismissed):

ST-DDT:
Why does it keep removing the request for review from @matthewmayer ?

xDivisionByZerox (dismissed):

matthewmayer (approved):

xDivisionByZerox (approved):

Shinigami92 (approved):

Shinigami92:
@ST-DDT if you consider @matthewmayer's review already as member review, then I accept if you already merge right now :slightly_smiling_face: 

@matthewmayer team membership does not mean you need to attempt at meetings but provides you the ability for green approvals (and other stuff)
If you want, just say you want that your approvals count to our green-approval (internal) guide line

> If a PR has 2 green approvals, it needs to wait for 24 hours to get merged, or does get a third green approval and can then directly merged
> Still at least one core maintainer (@ST-DDT or @Shinigami92 ^^) needs to also be one of the approvals (what is mostly always the case anyway)

matthewmayer:
Sure, you can add me on the team if that makes it easier to request reviews etc. Since my Typescript is fairly limited I'll only approve PRs which are more locale-focused or documentation-focused like this one.

PR #1784, opened 2023-01-26
refactor(location): normalize signatures

### LINKED ISSUE #1349: Standardize function parameters and defaults

xDivisionByZerox:
Part of #1349.

Standardize all function signatures in the `LocationModule` to support an options object.


src/modules/location/index.ts
```diff
@@ -31,17 +32,30 @@ export class LocationModule {
...
-    // if zip format is not specified, use the zip format defined for the locale
-    if (format == null) {
-      const localeFormat = this.faker.definitions.location.postcode;
-      if (typeof localeFormat === 'string') {
-        format = localeFormat;
-      } else {
-        format = this.faker.helpers.arrayElement(localeFormat);
-      }
+  zipCode(
+    options:
```
ST-DDT, line 36: IMO we should remove the parameter instead.

src/modules/location/index.ts
```diff
@@ -217,7 +257,26 @@ export class LocationModule {
...
+      | 'alpha-3'
+      | {
+          /**
+           * The code to return.
+           * Can be either `'alpha-2'` (2 letter code)
+           * or `'alpha-3'` (three letter code).
+           *
+           * @default `'alpha-2'`.
+           */
+          alphaCode?: 'alpha-2' | 'alpha-3';
```
ST-DDT: Maybe `variant`?

src/modules/location/index.ts
xDivisionByZerox: If we want to remove the option entirely, I would suggest to do that in a separate PR since this PR aims to standardize all function signatures, not change the current behavior.

xDivisionByZerox:
The PR is now open for review. For simplicity (on my side) I literally took the previous parameter and simply wrapped it in an options object.

I highly encourage all reviewers to discuss the "new" property names in the options argument for each function. If you like a suggestion from an other reviewer, please upvote the comment. I'll leave each comment hanging around for some time before actually changing the discussed object property names.

src/modules/location/index.ts
Shinigami92: Yes, I'm with @xDivisionByZerox on that, lets do this in its own PR
If no one want to work on this right now, open an issue for that so we mark this discussion here as "resolved"

src/modules/location/index.ts
```diff
@@ -31,17 +32,30 @@ export class LocationModule {
...
-        format = this.faker.helpers.arrayElement(localeFormat);
-      }
+  zipCode(
+    options:
+      | string
+      | {
+          /**
+           * The optional format used to generate the the zip code.
+           *
+           * @default
```
Shinigami92: JSDoc tag `@default` should take a real JS snipped as value as this gets rendered in IDEs as that

But before thinking further about what this could be, lets first see if this gets removed as suggested by @ST-DDT

src/modules/location/index.ts
```diff
@@ -52,15 +66,22 @@ export class LocationModule {
...
+   * @param options.state The abbreviation of the state to generate the zip code for.
    *
    * @example
    * fakerUS.location.zipCodeByState("AK") // '99595'
    * fakerUS.location.zipCodeByState("??") // '47683-9880'
    *
    * @since 8.0.0
    */
-  zipCodeByState(state: string): string {
+  zipCodeByState(options: string | { state?: string } = {}): string {
```
Shinigami92: Could you also add JSDoc to the `state` options property itself?

src/modules/location/index.ts
```diff
@@ -217,7 +257,26 @@ export class LocationModule {
...
+    options:
+      | 'alpha-2'
+      | 'alpha-3'
+      | {
+          /**
+           * The code to return.
+           * Can be either `'alpha-2'` (2 letter code)
+           * or `'alpha-3'` (three letter code).
+           *
+           * @default `'alpha-2'`.
```
Shinigami92:
```suggestion
           * @default 'alpha-2'
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
    * @since 8.0.0
    */
-  // TODO @xDivisionByZerox 2022-06-12 this signature should probably be an object for easier maintainability
-  latitude(max: number = 90, min: number = -90, precision: number = 4): number {
+  latitude(
+    options: {
+      /**
+       * The upper bound for the latitude to generate.
+       *
+       * @default `90`.
```
Shinigami92:
```suggestion
       * @default 90
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
+       */
+      max?: number;
+      /**
+       * The lower bound for the latitude to generate.
+       *
+       * @default `-90`.
```
Shinigami92:
```suggestion
       * @default -90
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
+       */
+      min?: number;
+      /**
+       * The number of decimal points of precision for the latitude.
+       *
+       * @default `4`.
```
Shinigami92:
```suggestion
       * @default 4
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
-  // TODO @xDivisionByZerox 2022-06-12 this signature should probably be an object for easier maintainability
   longitude(
-    max: number = 180,
-    min: number = -180,
-    precision: number = 4
+    options: {
+      /**
+       * The upper bound for the longitude to generate.
+       *
+       * @default `180`.
```
Shinigami92:
```suggestion
       * @default 180
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
+       */
+      max?: number;
+      /**
+       * The lower bound for the longitude to generate.
+       *
+       * @default `-180`.
```
Shinigami92:
```suggestion
       * @default -180
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
+       */
+      min?: number;
+      /**
+       * The number of decimal points of precision for the longitude.
+       *
+       * @default `4`.
```
Shinigami92:
```suggestion
       * @default 4
```

src/modules/location/index.ts
```diff
@@ -257,57 +316,123 @@ export class LocationModule {
...
-  direction(useAbbr: boolean = false): string {
+  direction(
+    options:
+      | boolean
+      | {
+          /**
+           * If true this will return abbreviated directions (NW, E, etc).
+           * Otherwise this will return the long name.
+           *
+           * @default `false`.
```
Shinigami92:
```suggestion
           * @default false
```

src/modules/location/index.ts
```diff
@@ -322,17 +447,36 @@ export class LocationModule {
...
-  cardinalDirection(useAbbr: boolean = false): string {
+  cardinalDirection(
+    options:
+      | boolean
+      | {
+          /**
+           * If true this will return abbreviated directions (N, E, etc).
+           * Otherwise this will return the long name.
+           *
+           * @default `false`.
```
Shinigami92:
```suggestion
           * @default false
```

src/modules/location/index.ts
```diff
@@ -347,17 +491,36 @@ export class LocationModule {
...
-  ordinalDirection(useAbbr: boolean = false): string {
+  ordinalDirection(
+    options:
+      | boolean
+      | {
+          /**
+           * If true this will return abbreviated directions (NW, SE, etc).
+           * Otherwise this will return the long name.
+           *
+           * @default `false`.
```
Shinigami92:
```suggestion
           * @default false
```

src/modules/location/index.ts
xDivisionByZerox: Yes. I simply forgot it at that specific location.

src/modules/location/index.ts
xDivisionByZerox: So in this case it should state `@default faker.definitions.location.postcode`. Did I understand you correctly?

Shinigami92: yes

src/modules/location/index.ts
```diff
@@ -207,7 +256,8 @@ export class LocationModule {
   /**
    * Returns a random country code.
    *
-   * @param alphaCode The code to return. Can be either `'alpha-2'` (2 letter code)
+   * @param options The code to return or an options object. Defaults to `{}`.
+   * @param options.alphaCode The code to return. Can be either `'alpha-2'` (2 letter code)
```
```suggestion
   * @param options.variant The variant to return. Can be either `'alpha-2'` (2 letter code)
```

src/modules/location/index.ts
```diff
@@ -257,57 +325,123 @@ export class LocationModule {
...
+   * faker.location.latitude({ max: 10 }) // 5.7225
+   * faker.location.latitude({ max: 10, min: -10 }) // -9.6273
+   * faker.location.latitude({ max: 10, min: -10, precision: 5 }) // 2.68452
    *
    * @since 8.0.0
    */
-  // TODO @xDivisionByZerox 2022-06-12 this signature should probably be an object for easier maintainability
-  latitude(max: number = 90, min: number = -90, precision: number = 4): number {
+  latitude(
+    options: {
```
ST-DDT: This is a breaking signature change, either mark this PR as breaking or add a legacy signature.

src/modules/location/index.ts
```diff
@@ -257,57 +325,123 @@ export class LocationModule {
...
+   * faker.location.longitude({ max: 10, min: -10, precision: 5 }) // -4.03620
    *
    * @since 8.0.0
    */
-  // TODO @xDivisionByZerox 2022-06-12 this signature should probably be an object for easier maintainability
   longitude(
-    max: number = 180,
-    min: number = -180,
-    precision: number = 4
+    options: {
```
ST-DDT: This is a breaking signature change.

ST-DDT (changes requested):

src/modules/location/index.ts
xDivisionByZerox: I would allow both signatures (multi args and options) in this PR and create a separate PR for each breaking change to make them more visible?

ST-DDT: If you allow both options, then its not breaking. So AFAICT you just have to re-add support for the old signature.

ST-DDT (approved):

Shinigami92 (approved):
I assume the deprecation will be done in another PR for better reviewability?

xDivisionByZerox:
> I assume the deprecation will be done in another PR for better reviewability?

Actually, I wanted to deprecate each method in its own PR so it is visible in the changelog.

PR #1786, opened 2023-01-27
test(helpers): widen arrayElements distribution

Shinigami92:
while working on
- #1645 

we run into `arrayElements` expectation failures
this PR just widens the distribution for now as for small numbers the error is much more present as for larger iterations

### DIFF
--- a/test/helpers.spec.ts
+++ b/test/helpers.spec.ts
@@ -370,8 +370,8 @@ describe('helpers', () => {
             }
 
             for (const occurrence of occurrences) {
-              expect(occurrence).toBeGreaterThanOrEqual(70);
-              expect(occurrence).toBeLessThanOrEqual(130);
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):

ejcheng (approved):

xDivisionByZerox (approved):
We might want to simply retry this test internally a set number of times. Since it all comes down to probability, changing the distribution area will not help in the long run as you can always simply get a 1:0 distribution.

I'll approve this for now since it is a quick "fix".

Shinigami92:
@xDivisionByZerox Yes, thanks for thinking about that dedicate on our own
Yesterday, @ST-DDT and I, already run these snippets 10 million times to _ensure_ that these boundaries are considered as _enough_

PR #1787, opened 2023-01-27
test(system): add valid 0 for cron

Shinigami92:
While working on
- #1645 

it reports an failure for https://github.com/faker-js/faker/actions/runs/4025222101/jobs/6918173202#step:7:302

but `0` is a valid digit for cron

### DIFF
--- a/test/system.spec.ts
+++ b/test/system.spec.ts
@@ -407,11 +407,11 @@ describe('system', () => {
           }
         );
 
-        it('should return non-standard cron expressions', () => {
+        it('should be able to return non-standard cron expressions', () => {
           const validResults = [...'0123456789'.split(''), '*', '@'];
           expect(
             faker.system.cron({ includeNonStandard: true })[0],
-            'generated cron, string should contain non-standard cron labels'
+            'generated cron, string should contain standard or non-standard cron labels'
           ).toSatisfy(
             (value) => !!validResults.find((result) => value === result)
           );
--- a/test/system.spec.ts
+++ b/test/system.spec.ts
@@ -408,7 +408,7 @@ describe('system', () => {
         );
 
         it('should return non-standard cron expressions', () => {
-          const validResults = ['1', '2', '3', '4', '5', '6', '*', '@'];
+          const validResults = ['0', '1', '2', '3', '4', '5', '6', '*', '@'];
           expect(
             faker.system.cron({ includeNonStandard: true })[0],
             'generated cron, string should contain non-standard cron labels'
--- a/test/system.spec.ts
+++ b/test/system.spec.ts
@@ -408,7 +408,7 @@ describe('system', () => {
         );
 
         it('should return non-standard cron expressions', () => {
-          const validResults = ['0', '1', '2', '3', '4', '5', '6', '*', '@'];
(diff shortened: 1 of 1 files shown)


ejcheng (dismissed):

test/system.spec.ts
```diff
@@ -408,7 +408,7 @@ describe('system', () => {
         );
 
         it('should return non-standard cron expressions', () => {
-          const validResults = ['1', '2', '3', '4', '5', '6', '*', '@'];
+          const validResults = ['0', '1', '2', '3', '4', '5', '6', '*', '@'];
```
ST-DDT: This can also generate 7-9 (minutes range from 0-59).

https://github.com/faker-js/faker/blob/86ae8b9d698f5b2c75214b0cfef869122a591001/src/modules/system/index.ts#L346
https://github.com/faker-js/faker/blob/86ae8b9d698f5b2c75214b0cfef869122a591001/src/modules/system/index.ts#L366-L369

ST-DDT (dismissed):

test/system.spec.ts
Shinigami92: okay, I will include just all digits

test/system.spec.ts
```diff
@@ -408,7 +408,7 @@ describe('system', () => {
         );
 
         it('should return non-standard cron expressions', () => {
-          const validResults = ['1', '2', '3', '4', '5', '6', '*', '@'];
+          const validResults = [...'0123456789'.split(''), '*', '@'];
```
xDivisionByZerox, line 411: I'm a bit confused. Numbers are considered non-standard cron expressions?

Shinigami92: Uh, yeah the description text now looks weird to me as well :thinking: 
Did this just slipped to the review while this was added? Or is something differently meant by that or the test implementation shifted away from what it was originally?

Maybe @ST-DDT has some more insites to this as I also think he knows better about cron

ST-DDT: IMO the test should be rewritten to actually check that it "can return non standard expressions".

Shinigami92: Okay, what are _non-standard_ cron expressions?

ST-DDT: See the implementation. Thats the `@daily` ones.

Shinigami92:
This PR targeted the wrong goal and was meant to fix the shuffled tests
Now we want to change the whole implementation of the test itself
Therefore it's up to be grepped by someone who wants to fix it

xDivisionByZerox:
I'll provide a follow up PR which addresses this test reimplementation.

xDivisionByZerox:
After checking the source I noticed that the fix for the test was totally valid.
`includeNonStandard` is misleading in that way that it doesn't guarantee that a non standard expression is returned. Thats why we simply have to check whether a number/astrics (standard) or an `@` (non standard) is returned.

The title should probably simply be adjusted to state `'should be able to return non-standard cron expressions'`.

xDivisionByZerox (approved):

ejcheng (approved):

ST-DDT (approved):

PR #1788, opened 2023-01-28
refactor(system): replace mime db with common dataset

### LINKED ISSUE #1317: Remove mime-db and use a small common set of mime-types

The mime-db is very large and most of the extensions are never used by anyone.

Also, not all mime-db entries pass the [validator.js#isMimeType](https://github.com/validatorjs/validator.js/blob/master/src/lib/isMimeType.js#L37) check.

grafik

### Suggested solution

Replace mime-db with a custom/common/useful set of mime types

### Alternative

Filter all mime-db entries to ensure they satisfy the [validator.js#isMimeType](https://github.com/validatorjs/validator.js/blob/master/src/lib/isMimeType.js#L37) check.

### Additional context

_No response_

xDivisionByZerox:
Remove most mime-type definitions to only support common ones based on the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types).
A section regarding this change was documented in the "upgrade" section of the user guides.
Furthermore, remove all other properties other than `extensions` for a specific mime type since that's the only thing that is used.

Unique mime types before: 2279
Unique mime types now: 72

Closes #1317.


xDivisionByZerox:
The methods [`commonFileName`](https://github.com/faker-js/faker/blob/6772b00cf19ccdbebefe023dc0f96c6576234b5b/src/modules/system/index.ts#L137-L139), [`commonFileType`](https://github.com/faker-js/faker/blob/6772b00cf19ccdbebefe023dc0f96c6576234b5b/src/modules/system/index.ts#L109-L113) and [`commonFileExt`](https://github.com/faker-js/faker/blob/6772b00cf19ccdbebefe023dc0f96c6576234b5b/src/modules/system/index.ts#L149-L151) method should be deprecated in a separate PR.

Shinigami92 (dismissed):
IMO from here on we can just add stuff back again on demand
Thx for the work :+1:

docs/guide/upgrading.md
```diff
@@ -118,3 +118,24 @@ For the old `faker.datatype.number` method you should replace with `faker.number
 ### `allowLeadingZeros` behavior change in `faker.string.numeric`
 
 The `allowLeadingZeros` boolean parameter in `faker.string.numeric` (in the new `string` module) now defaults to `true`. `faker.string.numeric` will now generate numeric strings that could have leading zeros by default.
+
+### Remove of `mime-db`
```
xDivisionByZerox: I'm not sure about this section. 

@import-brain Can you at least read read it once since you are native in English?

ST-DDT (dismissed):

docs/guide/upgrading.md
matthewmayer: I think the migration guide should only focus on the changes which affect consumers of the library, not detailed rationales. I think the fact that Faker was using mime-db is an internal implementation detail not really important to end-consumers.

I'd keep this much simpler like

```
### Simplified MIME type data

The functions `faker.system.mimeType`, `faker.system.fileType` and `faker.system.fileExt` now return data from a smaller set of more common MIME types, filetypes and extensions.
```

xDivisionByZerox: Thank you @matthewmayer. I'm always a bit lost when I have to write documentation. 
I'll made the according changes.

Shinigami92 (approved):

xDivisionByZerox:
Can anybody tell me why [the pipeline](https://github.com/faker-js/faker/pull/1788/checks?check_run_id=10956751748) is failing?

Shinigami92:
> Can anybody tell me why [the pipeline](https://github.com/faker-js/faker/pull/1788/checks?check_run_id=10956751748) is failing?

Yes, this is the codecov pipeline and you remove a huge chunk of code, so the diff from prev to now differs "to much"
But as we know what we are doing here, this can be just taken as a warning and should be ignored

xDivisionByZerox:
> Yes, this is the codecov pipeline and you remove a huge chunk of code, so the diff from prev to now differs "to much" But as we know what we are doing here, this can be just taken as a warning and should be ignored

Ah ok, thank you. Never saw this one failing before. That's why I was a bit lost on what to do/how to fix.

ejcheng (approved):

ST-DDT (approved):

PR #1789, opened 2023-01-29
refactor(string): rename `special` to `symbol`

### LINKED ISSUE #1711: Generate random symbol (eg: ! _ #)

Provide a method which would generate random symbol (non alphanumeric character), for example `!#_-/\.$`.

### Suggested solution

```js
faker.random.symbol() // #
``` 

### Alternative

_No response_

### Additional context

_No response_

xDivisionByZerox:
Resolves #1711.

### DIFF
--- a/src/modules/string/index.ts
+++ b/src/modules/string/index.ts
@@ -742,7 +742,7 @@ export class StringModule {
   /**
    * Returns a string containing only special characters from the following list:
    * ```txt
-   * ! " \# $ % & ' ( ) \* \+ , \- . / : ; < = \> ? @ [ \ ] ^ _ ` { | } ~
+   * ! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~
    * ```
    *
    * @param length Length of the generated string. Defaults to `1`.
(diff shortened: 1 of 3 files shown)


src/modules/string/index.ts
```diff
@@ -740,20 +740,52 @@ export class StringModule {
...
+   * - [
+   * - \
+   * - ]
+   * - ^
+   * - _
+   * - `
+   * - {
+   * - |
+   * - }
+   * - ~
```
xDivisionByZerox: If I don't write it like this the "required description" test is [failing](https://github.com/faker-js/faker/actions/runs/4034734413/jobs/6936165971#step:7:299) (I guess due to an parsing error). Any suggestions or opinions?

xDivisionByZerox: [Seems like it's failing either way...](https://github.com/faker-js/faker/actions/runs/4034743278/jobs/6936183958#step:7:299)

ST-DDT (approved):

pkuczynski (approved):

Shinigami92 (approved):

ejcheng (approved):

Issue #1785, opened 2023-01-27
Deprecate helpers.unique for removal


Shinigami92:
In team meeting 2023-01-26, we decided that we do not want to maintain methods anymore that are not directly or indirectly based on _mersenne_, or at least used inside one of other exposed faker methods (e.g. _slugify_ is used in _domainWord_).

This affects `helpers.unique` and we will mark it as deprecated for removal.
There will no alternative given from inside faker, but maybe we can link to another npm package that can be used instead.

IlCallo:
Is there any rationale underneath this decision, except not having enough bandwidth to cope with everything?
It seems to me this is a pretty common use case, I used this library in 2-3 different projects and I ended up using it at least once in all those projects

Shinigami92:
I started to look around for some npm packages and did not found directly something that suites our need as `faker.helpers.unique` does right now.

So I did something what everyone is doing right now 👀 
I ... asked ... ChatGPT 😱

<details>
<summary>ChatGPT chat history</summary>

![image](https://user-images.githubusercontent.com/7195563/215078598-0ac59d17-0860-49c0-a9b6-faa1f9e45146.png)

</details>

TLDR:
```ts
interface Store {
  [key: string]: any;
}

function memoizeUnique<T>(callback: (...args: any[]) => T): (...args: any[]) => T {
  let store: Store = {};
  return function(...args: any[]): T {
    let result: T;
    do {
        result = callback(...args);
        const key = JSON.stringify(args) + JSON.stringify(result);
        if (!store.hasOwnProperty(key)) {
          store[key] = result;
          break;
        }
    } while(true)
    return result;
  };
}
```

With a really simple _this could work_-review, it can be extended to take the store as an additional argument + some addition "stop conditions" and we have something similar to what we have in faker right now.

ChatGPT helped also finding a name for this: `memoizeUnique`

Somehow I'm really shocked that something like this doesn't already exists on npm as far as my researches are going 🤷 
If no one is interested in publishing this as a npm package... I could claim that and do it on my own and make it an Open Source project 👀

xDivisionByZerox:
This is not required for `v8.0`. We can deprecate at any point. 
I'll move this to milestone `v8.x`

Shinigami92:
My actual try of an re-implementation looks like this:

```ts
export interface Store<T> {
  [key: string]: T;
}

export interface MemoizeUniqueOptions<T> {
  /**
   * The time in milliseconds this method may take before throwing an error.
   *
[...]

PR #1790, opened 2023-01-29
refactor(helpers): deprecate unique method

Shinigami92:
For now this is just a preparation for
- #1785 

and marks the method `helpers.unique` as a soft-deprecation
~~This means for now that we wont throw a warning at runtime~~ but just want to indicate in IDEs like VSCode that it is not safe anymore to rely on this function as it will not be maintained by us anymore and we want to remove it in the long run.

_This PR should **not** auto-close_ #1785


ejcheng (dismissed):

ST-DDT:
> This means for now that we wont throw a warning at runtime

Could you please explain the reason why you did it like this instead of our usual deprecation workflow?

Shinigami92:
> > This means for now that we wont throw a warning at runtime
> 
> Could you please explain the reason why you did it like this instead of our usual deprecation workflow?

Yes, it is because I searched for alternatives, but didn't found one yet.
I'm still thinking about to publish a package on my own, but not so sure if I want to do this as I would love to have it more as a community driven open source package. This could still be possible but not sure if my GitHub account is this much known right now.

Secondly I want to deprecate it like this right now so no further maintain request would come in like e.g. `https://github.com/faker-js/faker/issues/456` (_linked like that so I do not link the issues_)

So, until we do not have an alternative we could point to, I wont like to annoy our userbase.

ST-DDT:
> So, until we do not have an alternative we could point to, I wont like to annoy our userbase.

Having a deprecation warning in the codebase could also be considered anoying IMHO.

Shinigami92:
> > So, until we do not have an alternative we could point to, I wont like to annoy our userbase.
> 
> Having a deprecation warning in the codebase could also be considered anoying IMHO.

It will not warn, but just ~~strikes through~~
it would only warn if you have configured an eslint rule

ST-DDT:
IMO we should only consider removing this if there is already replacement for it somewhere.

Shinigami92:
> IMO we should only consider removing this if there is already replacement for it somewhere.

Please suggest a different approach to show users that this function is decided in last team meeting on 2023-01-26 that we as faker maintainer will not maintain this function anymore in the future

ST-DDT:
Other than having an open issue?

I only agree to this if there is an issue blocking v8.0 to document a replacement and the removal of the exclusion customization added by this PR.
I can write the issue for that if you would like me to.

Shinigami92:
> Other than having an open issue?
> 
> I only agree to this if there is an issue blocking v8.0 to document a replacement and the removal of the exclusion customization added by this PR. I can write the issue for that if you would like me to.

No, this is against what we decided here: https://github.com/faker-js/faker/issues/1785#issuecomment-1407468421
The replacement is not blocking v8.0 :shrug:

ST-DDT:
> > I only agree to this if there is an issue blocking v8.0 to document a replacement and the removal of the exclusion customization added by this PR. I can write the issue for that if you would like me to.
> 
> No, this is against what we decided here: [#1785 (comment)](https://github.com/faker-js/faker/issues/1785#issuecomment-1407468421) The replacement is not blocking v8.0 🤷

Yes, currently it wouldn't be a blocking issue, but if you want to merge this PR in v8.0, then it would become one.
IMO: You could directly point to [this comment](https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744) as a replacement and proceed with a normal deprecation.

Shinigami92:
> > > I only agree to this if there is an issue blocking v8.0 to document a replacement and the removal of the exclusion customization added by this PR. I can write the issue for that if you would like me to.
> > 
> > 
> > No, this is against what we decided here: [#1785 (comment)](https://github.com/faker-js/faker/issues/1785#issuecomment-1407468421) The replacement is not blocking v8.0 shrug
> 
> Yes, currently it wouldn't be a blocking issue, but if you want to merge this PR in v8.0, then it would become one. IMO: You could directly point to [this comment](https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744) as a replacement and proceed with a normal deprecation.

ok :ok_hand: I will do so

src/modules/helpers/index.ts
```diff
@@ -902,6 +903,9 @@ export class HelpersModule {
    * faker.helpers.unique(faker.person.firstName) // 'Corbin'
    *
    * @since 7.5.0
+   *
+   * @deprecated Please find a dedicated npm package instead, or even create one on your own if you like to.
+   * More info can be found in issue [faker-js/faker #1785](https://github.com/faker-js/faker/issues/1785).
```
ST-DDT, line 1250: Please add a `@see` as well.

Shinigami92: like this?

ST-DDT: Theoretically yes, but for some reason it won't be shown on the docs page:

grafik
grafik
https://github.com/faker-js/faker/blob/52e9c1c43769d94f3f2c7be8e52504a46aa7cd74/docs/.vitepress/components/api-docs/method.vue#L47
````json
"seeAlsos": [
  "<a href=\"https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744\" target=\"_blank\" rel=\"noreferrer\">https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744</a>"
]
````

Comment (Shinigami92) [reply src/modules/helpers/index.ts:1250]: I'm not sure if this is related to this PR in any way at all :thinking: 

Maybe you should check if there is a general bug in the automation script
Maybe this is the first link containing a href with `#` or anything like that
But `#` are valid URL parts (see https://developer.mozilla.org/en-US/docs/Web/API/URL/hash)

Comment (ST-DDT) [reply src/modules/helpers/index.ts:1250]: Fixed in https://github.com/faker-js/faker/pull/1794

**Review by ST-DDT — DISMISSED** [review]
While I personally don't like this deprecation.
This is what the team has decided on.
The change itself looks good to me.

File: src/modules/helpers/index.ts [hunk 2]
```diff
@@ -933,10 +934,15 @@ export class HelpersModule {
...
    * @param options.store The store of unique entries. Defaults to a global store.
    *
+   * @see https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744
+   *
    * @example
    * faker.helpers.unique(faker.person.firstName) // 'Corbin'
    *
    * @since 7.5.0
+   *
+   * @deprecated Please find a dedicated npm package instead, or even create one on your own if you like to.
```
ejcheng:
```suggestion
   * @deprecated Please find a dedicated npm package instead, or even create one on your own if you want to.
```

ST-DDT (approved):

xDivisionByZerox (approved):
I want to publicly +1 [@ST-DDT's statement](https://github.com/faker-js/faker/pull/1790#pullrequestreview-1274315679). 
I don't like deprecating this in the knowledge that there is no real alternative to this. 
Nevertheless, this was decided in a team meeting and I respect the decisions (and arguments) made there.

Shinigami92:
For future read, please have a look first into #1785 before just visiting and complaining about this PR

- #1785

PR #1793, opened 2023-01-29
infra: configure lint rule array-type

Shinigami92:
This configures the lint rule https://typescript-eslint.io/rules/array-type/

We could decide if we want to configure it with:

`{ default: "array-simple" }` or `{ default: "array-simple", readonly: "generic" }`

The difference would be:

```ts
number[]
Array<Foo & Bar>
readonly number[]
ReadonlyArray<Foo & Bar>
```

vs

```ts
number[]
Array<Foo & Bar>
ReadonlyArray<number> // <- this is different
ReadonlyArray<Foo & Bar>
```


ST-DDT:
> The difference would be:

Please add a hint what the new one is^^

---

I generally agree with adding this rule, I just consider `ReadonlyArray<number>` to be easier to read.

IMO the `readonly` refers to the array and thus should be "close" to the `array`. 
In `readonly number[]` you have the array element type mixed in between.

Is there a reason for the shuffled declaration?

Shinigami92:
> Please add a hint what the new one is ^^

both are new (?!)
we could decide for one of these both option configurations

Shinigami92:
> I generally agree with adding this rule, I just consider ReadonlyArray<number> to be easier to read.
>
> IMO the readonly refers to the array and thus should be "close" to the array.
> In readonly number[] you have the array element type mixed in between.
>
> Is there a reason for the shuffled declaration?

This is exactly what I also thought!
Okay I will switch to the second

src/modules/string/index.ts
```diff
@@ -4,9 +4,13 @@ import type { LiteralUnion } from '../../utils/types';
...
-const UPPER_CHARS: readonly string[] = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
-const LOWER_CHARS: readonly string[] = 'abcdefghijklmnopqrstuvwxyz'.split('');
-const DIGIT_CHARS: readonly string[] = '0123456789'.split('');
+const UPPER_CHARS: ReadonlyArray<string> = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split(
+  ''
+);
+const LOWER_CHARS: ReadonlyArray<string> = 'abcdefghijklmnopqrstuvwxyz'.split(
+  ''
+);
+const DIGIT_CHARS: ReadonlyArray<string> = '0123456789'.split('');
```
ST-DDT, line 13: Are these types even required?

Shinigami92: Yes they are, because they mark these arrays as readonly for compile time

Shinigami92: I tested `const UPPER_CHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('') as const;`, but this gives a compile time error

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #1794, opened 2023-01-29
docs: fix see also for non faker.x references

ST-DDT:
Found in https://github.com/faker-js/faker/pull/1790#discussion_r1090060316

In vue `<template>` tags aren't rendered. So we have to use div instead.

Thanks to @Shinigami92 for helping to fix this issue.

### DIFF
--- a/docs/.vitepress/components/api-docs/method.vue
+++ b/docs/.vitepress/components/api-docs/method.vue
@@ -44,7 +44,7 @@ function seeAlsoToUrl(see: string): string {
             :href="seeAlsoToUrl(seeAlso)"
             v-html="seeAlso"
           />
-          <template v-else v-html="seeAlso" />
+          <div v-else v-html="seeAlso" />
         </li>
       </ul>
     </div>


Shinigami92 (approved):

xDivisionByZerox:
Can you post a screenshot of what changed here for visual comparison? Were "See Also" simply missing in our docs so far?

Documentation on the next branch seems to be working as intended:
image

Shinigami92:
> Can you post a screenshot of what changed here for visual comparison? Were "See Also" simply missing in our docs so far?
> 
> Documentation on the next branch seems to be working as intended: image

In the current docs, there is no example for this yet, but https://github.com/faker-js/faker/pull/1790#discussion_r1090060316 would add one
But you see anyway that the current code is invalid/buggy anyways, so I would appreciate that you approve and trust 🙂 

Maybe read this for more info: https://vuejs.org/guide/essentials/conditional.html#v-if-on-template

@ST-DDT expected that `v-if`/`v-else` is needed in combination with `<template>`, but then applied `v-html` which is not compatible with `<template>` as that is a non rendered block for just code-flow

xDivisionByZerox (approved):

ST-DDT:
I merge this fix now to unblock #1790 .

PR #1808, opened 2023-02-02
test: use cy.request for e2e

Shinigami92:
Instead of visiting and go back on each site, use [cy.request](https://docs.cypress.io/api/commands/request) to just fetch the status and body

This improves the test case from around 30s to just 3s

Only downside right now:
It does not check the header elements but just that the text exists


Shinigami92:
https://github.com/faker-js/faker/actions/runs/4068904881/jobs/7008930568#step:7:49

vs 

https://github.com/faker-js/faker/actions/runs/4073655735/jobs/7017841357#step:7:62

---

Old

```node
  API Test
    ✓ navigates to the api index search (552ms)
    API Reference
      ✓ should at least list more than 7 modules (260ms)
      ✓ should include at least 1 element in each module (26301ms)
      ✓ should not have dead links (86868ms)
```

New

```node
  API Test
    ✓ navigates to the api index search (678ms)
    API Reference
      ✓ should at least list more than 7 modules (272ms)
      ✓ should include at least 1 element in each module (31112ms)
      ✓ should not have dead links (11525ms)
```

---

still "slow" but factor of ~8x

ST-DDT (approved):
Lets give this a try.

xDivisionByZerox (approved):

PR #1809, opened 2023-02-02
feat(finance): currency object

### LINKED ISSUE #1733: Better currencies

This will have mismatched data :)
```
supportedCurrency: {
  code: faker.finance.currencyCode(),
  name: faker.finance.currencyName()
}
```

### Suggested solution

Like `unit()`, it returns `{ name: 'meter', symbol: 'm' }`

It would be nice if `finance.currency()` returned `{ name: 'US Dollar', symbol: '$', code: 'USD' }`

### Alternative

_No response_

### Additional context

_No response_

matthewmayer:
fix #1733


src/modules/finance/index.ts
```diff
@@ -176,7 +176,26 @@ export class FinanceModule {
       this.faker.definitions.finance.transaction_type
     );
   }
-
+  /**
```
ST-DDT, line 595:
```suggestion

  /**
```

src/modules/finance/index.ts
```diff
@@ -176,7 +176,26 @@ export class FinanceModule {
...
+   *
+   * @since 8.0.0
+   */
+  currency(): { name: string; code: string; symbol: string } {
+    const currencies = this.faker.definitions.finance.currency;
+    const name = this.faker.helpers.objectKey(currencies) as string;
+    const code = currencies[name]['code'];
+    const symbol = currencies[name]['symbol'];
+    return { name, code, symbol };
+  }
```
ST-DDT, line 612:
```suggestion
  }

```

ST-DDT:
**Team Decision**

Change the locale data to contain [{ name, code, symbol }]

matthewmayer:
Sorry, had some issues with my branch so had to force-push

matthewmayer:
This was my script to transform the definitions:

```js
'use strict';
const fs = require('fs');
const langs = ['en','el', 'fr_CH', 'fr', 'fa'];
const { faker } = require('@faker-js/faker');
for (let lang of langs) {
  let c = faker.locales[lang].finance.currency;
  let newCurr = Object.keys(c).map((k) => {
    return { name: k, code: c[k].code, symbol: c[k].symbol };
  });
  let str = JSON.stringify(newCurr, null, 2);
  fs.writeFileSync(
    `./src/locales/${lang}/finance/currency.ts`,
    `export default ${str}`
  );
}
```

src/modules/finance/index.ts
```diff
@@ -582,9 +600,9 @@ export class FinanceModule {
    * @since 2.0.1
    */
   currencyCode(): string {
-    return this.faker.helpers.objectValue(
+    return this.faker.helpers.arrayElement(
```
Shinigami92: Just an idea, but we could call `this.currency().code`, could we?

src/locales/fr_CH/finance/currency.ts
```diff
@@ -1,346 +1,432 @@
-export default {
-  Dollar: {
-    symbol: '$',
+export default [
+  {
+    name: 'Dollar',
```
ST-DDT, line 3: Why is this a single line in [`en`](https://github.com/faker-js/faker/pull/1809/files#diff-d2590c7cfefec571e5c844e3da8095c10f090b3eb4d2d522ea66d4d7ab9cfa31R148), but multiple here?

src/modules/finance/index.ts
```diff
@@ -572,6 +572,24 @@ export class FinanceModule {
...
+   * faker.finance.currency() // {code:'USD', name:'US Dollar' , symbol:'$'}
+   *
+   * @since 8.0.0
+   */
+  currency(): { name: string; code: string; symbol: string } {
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.finance.currency
+    );
+  }
   /**
```
```suggestion
  }

  /**
```

src/modules/finance/index.ts
```diff
@@ -572,6 +572,24 @@ export class FinanceModule {
...
+   * faker.finance.currencyCode()
+   * faker.finance.currencyName()
+   * faker.finance.currencySymbol()
+   *
+   * @example
+   * faker.finance.currency() // {code:'USD', name:'US Dollar' , symbol:'$'}
+   *
+   * @since 8.0.0
+   */
+  currency(): { name: string; code: string; symbol: string } {
```
ST-DDT: Should we create a named type for this?
(Like we did in the [airline module](https://github.com/faker-js/faker/blob/next/src/modules/airline/index.ts#L102))

src/modules/finance/index.ts
```diff
@@ -572,6 +572,24 @@ export class FinanceModule {
...
+  /**
+   * Returns a random currency object, containing 'code', 'name' and 'symbol' properties.
+   *
+   * @see
+   * faker.finance.currencyCode()
+   * faker.finance.currencyName()
+   * faker.finance.currencySymbol()
+   *
+   * @example
+   * faker.finance.currency() // {code:'USD', name:'US Dollar' , symbol:'$'}
```
```suggestion
   * faker.finance.currency() // { code: 'USD', name: 'US Dollar', symbol: '$' }
```

src/modules/finance/index.ts
```diff
@@ -572,6 +572,24 @@ export class FinanceModule {
     );
   }
 
+  /**
+   * Returns a random currency object, containing 'code', 'name' and 'symbol' properties.
```
```suggestion
   * Returns a random currency object, containing `code`, `name `and `symbol` properties.
```

src/modules/finance/index.ts
```diff
@@ -596,9 +614,9 @@ export class FinanceModule {
    * @since 2.0.1
    */
   currencyName(): string {
-    return this.faker.helpers.objectKey(
+    return this.faker.helpers.arrayElement(
       this.faker.definitions.finance.currency
-    ) as string;
+    ).name;
   }
```
ST-DDT: Should we deprecate the old methods in favor of the new one?

ST-DDT (changes requested):
Please add a random seeded test for the `currency` object.

src/locales/fr_CH/finance/currency.ts
matthewmayer: i dunno, prettify seems happy with both options?

src/modules/finance/index.ts
matthewmayer: there is one already called FinanceCurrencyEntryDefinitions i can rename that to Currency

matthewmayer: or are the definition types supposed to be different to the named types?

src/modules/finance/index.ts
matthewmayer: i dont know if its necessary. the other methods have been around a long time, are trivial and don't cause a maintenance burden.

src/modules/finance/index.ts
matthewmayer: i tried that in 09b4c330, if that is the wrong approach i can revert

src/modules/finance/index.ts
```diff
@@ -572,6 +572,25 @@ export class FinanceModule {
...
+   * faker.finance.currencyCode()
+   * faker.finance.currencyName()
+   * faker.finance.currencySymbol()
+   *
+   * @example
+   * faker.finance.currency() // { code: 'USD', name: 'US Dollar', symbol: '$' }
+   *
+   * @since 8.0.0
+   */
+  currency(): Currency {
```
ST-DDT, line 608: When we use them as return types then we usually define the type in the module and use it in the definitions. I kind of dont care where, but I would like to have it consistent.

matthewmayer: moved it in https://github.com/faker-js/faker/pull/1809/commits/731e6c0b44c8a269f0e7f57595f1ea546f0b2d3b

matthewmayer:
> add a test of the currency object

added in https://github.com/faker-js/faker/pull/1809/commits/844cb203d76c44f303ccd1c788b02fbc70d770bc

src/locales/fr_CH/finance/currency.ts
matthewmayer: changed in https://github.com/faker-js/faker/pull/1809/commits/56da4ac092f954e77c576483d2491827d96dc4f2

matthewmayer:
i think there are a few "currencies" which could be removed e.g. 

{
    name: 'Codes specifically reserved for testing purposes',
    code: 'XTS',
    symbol: '',
  },

and various metals like Gold, Platinum etc,

Should i include that in this PR or wait and do as a seperate PR

matthewmayer:
> i think there are a few "currencies" which could be removed e.g.
> 
> { name: 'Codes specifically reserved for testing purposes', code: 'XTS', symbol: '', },
> 
> and various metals like Gold, Platinum etc,
> 
> Should i include that in this PR or wait and do as a seperate PR

comparing https://en.wikipedia.org/wiki/List_of_circulating_currencies with the list i'd suggest the removal of

```
[
  'CUC', 'EEK', 'HRK', 'LTL',
  'LVL', 'MRO', 'SLL', 'SVC',
  'XAG', 'XAU', 'XBA', 'XBB',
  'XBC', 'XBD', 'XDR', 'XFU',
  'XPD', 'XPT', 'XTS', 'ZWL'
]
```

and the addition of 

[ 'MRU', 'SLE', 'SSP'] (and possibly 'VED' )

these are a variety of currencies which got replaced (e.g. EEK -> EUR), changed (e.g. MRO -> MRU) or are non-currencies (e.g. XTS, XAU).

src/modules/finance/index.ts
```diff
@@ -2,6 +2,26 @@ import type { Faker } from '../..';
 import { FakerError } from '../../errors/faker-error';
 import iban from './iban';
 
+/**
+ * The possible definitions related to currency entries.
+ */
+export interface Currency {
```
Shinigami92, line 8: We should re-export this interface from src/index.ts so it can be used from outside (Like e.g. ColorFormat or SexType)

ST-DDT:
> i think there are a few "currencies" which could be removed e.g.
> 
> { name: 'Codes specifically reserved for testing purposes', code: 'XTS', symbol: '', },
> 
> and various metals like Gold, Platinum etc,
> 
> Should i include that in this PR or wait and do as a seperate PR

Please do that in a separate PR as the removed entries will be lost within the other changes to the locale definitions.

ST-DDT (dismissed):
Other than the re-export and the deprecation this looks good to me.

src/modules/finance/index.ts
matthewmayer: added the re-export

matthewmayer:
> Other than the re-export and the deprecation this looks good to me.

i added the re-export in 86d1ae09

Not planning to deperecate other methods (per https://github.com/faker-js/faker/pull/1809#discussion_r1111168114 ) unless others feel strongly they should be

Shinigami92 (approved):

ST-DDT:
Lets discuss this in the next team meeting.

ST-DDT:
**Team decision**

We won't deprecate the old methods.

--------

Then this PR is ready for review and merge.

src/modules/finance/index.ts
Shinigami92: => https://github.com/faker-js/faker/pull/1809#issuecomment-1442114654

ST-DDT (approved):

PR #1811, opened 2023-02-02
chore: remove comments in tsconfig.json

bdb2381:
JSON doesn't support comments. This PR removes the comments.

One downstream impact of these comments is Snyk fails to parse the JSON file.  https://jsoneditoronline.org/indepth/parse/json-comments/ for some details


Shinigami92:
Please note that tsconfig.json is officially written in jsonc

- https://github.com/microsoft/TypeScript/issues/4987
- https://github.com/Microsoft/vscode/issues/50974

So comments are officially allowed, based on what TypeScript allows

---

Theoretically we could remove these values anyway, but it might be that we would write other comments into this file anyway in a later phase for any reason

So IMO the problem is not on our side but on `Snyk` as they should use a jsonc / json5 parser to correctly read these kind of files

bdb2381:
That is new learning, jsonc. Thank you for the docs. I'll review so I have a better understanding. 

Agreed it is handy to have comments. And agree it is something Snyk should handle as Faker isn't the only thing Snyk fails to parse. I'll work with an AE at Snyk for it.

xDivisionByZerox:
> Theoretically we could remove these values anyway, but it might be that we would write other comments into this file anyway in a later phase for any reason

I suggest explicitly setting or removing them. Even if comments are technically allowed, the one in question do not serve any purpose currently.

ST-DDT:
I agree with @xDivisionByZerox 's reasoning here.
These should either be activated or removed (or should have a comment why they are there).

Since @Shinigami92 added them in https://github.com/faker-js/faker/pull/651 I would like to hear his original intention with those lines and suggestions whether to enable or remove them.

Shinigami92:
> Since @Shinigami92 added them in #651 I would like to hear his original intention with those lines and suggestions whether to enable or remove them.

It's exactly what the PR title is intended to say: **partially** activate strict mode
At the end it would be very nice to have our project fully `strict` (this is a ts flag that can be set to `true`)
But there were some issues and they can partially get solved one by one, one after another...
And at the end, erase all the partial flags and just set `strict: true`

ST-DDT:
> At the end it would be very nice to have our project fully `strict` (this is a ts flag that can be set to `true`)
> But there were some issues and they can partially get solved one by one, one after another...
> And at the end, erase all the partial flags and just set `strict: true`

I created an issue for each of them:

IMO this PR is now obsolete, since we want to actually turn on those lines.
What do you think?

Shinigami92 (approved):

xDivisionByZerox (approved):
IMO we can still merge this as an appreciation towards @bdb2381 for pointing out those settings.

ST-DDT:
Since @Shinigami92 already started working on them I think this will only cause merge conflicts.

I will close this, but thanks for bringing this to our attention.

Shinigami92:
On the one side, we could merge it, on the other side, it is more or less irrelevant as we wont release something just for this right now and therefore it would take month(s) anyway to land in "prod"

PR #1814, opened 2023-02-06
feat(person): add Portuguese BR Language to western zodiac sign

Minozzzi:

### DIFF
--- a/src/locales/pt_BR/person/index.ts
+++ b/src/locales/pt_BR/person/index.ts
@@ -14,6 +14,7 @@ import prefix from './prefix';
 import sex from './sex';
 import suffix from './suffix';
 import title from './title';
+import western_zodiac_sign from './western_zodiac_sign';
 
 const person: PersonDefinitions = {
   female_first_name,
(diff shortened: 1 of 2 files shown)


matthewmayer (changes requested):
you need to run `pnpm run generate:locales` to update the `src/locales/pt_BR/person/index.ts` file

ST-DDT (approved):
Looks good to me.
We should consider sorting these alphabetically as well (for overall easier review-ability of all locale data), although it might be detrimental for this data set in particular.

matthewmayer:
> Looks good to me.
> 
> We should consider sorting these alphabetically as well (for overall easier review-ability of all locale data), although it might be detrimental for this data set in particular.

As zodiac signs have an natural order (like months) it seems more sensible to keep them in that order.

xDivisionByZerox (approved):

ejcheng (approved):

matthewmayer (approved):

Shinigami92 (approved):

PR #1816, opened 2023-02-08
docs: add matthewmayer to team

Shinigami92:
Add @matthewmayer to team page

### DIFF
--- a/docs/about/team/members.json
+++ b/docs/about/team/members.json
@@ -83,6 +83,13 @@
       ]
     },
     {
+      "avatar": "https://github.com/matthewmayer.png",
+      "name": "Matt Mayer",
+      "title": "Contributor",
+      "desc": "",
+      "links": [{ "icon": "github", "link": "https://github.com/matthewmayer" }]
+    },
+    {
       "avatar": "https://github.com/pkuczynski.png",
       "name": "Piotr Kuczynski",
       "title": "Contributor",


docs/about/team/members.json
```diff
@@ -82,6 +82,13 @@
         { "icon": "github", "link": "https://github.com/xDivisionByZerox" }
       ]
     },
+    {
+      "avatar": "https://github.com/matthewmayer.png",
+      "name": "Matt Mayer",
+      "title": "Contributor",
+      "desc": "",
```
ST-DDT, line 89: This should be filled.

docs/about/team/members.json
```diff
@@ -82,6 +82,13 @@
...
+      "links": [{ "icon": "github", "link": "https://github.com/matthewmayer" }]
```
ST-DDT, line 90: @matthewmayer Any other links to other social media pages you wish to add?

ST-DDT (approved):
Looks good to me.
I didnt add this myself because of the blanks you and I cannot fill.

matthewmayer (approved):
Fine for me. I don't have any social media to promote so can leave blank.

docs/about/team/members.json
Shinigami92: I think the issue is, it's a required field for VitePress but some of us also do not have this filled
So it will just render an empty string instead of breaking the UI / VitePress

Issue #691, opened 2022-03-26
feat(name): add hyphenated patterns to `lastName()`


import-brain:
### Clear and concise description of the problem

The ability to generate hyphenated last names so that it seems just a little bit more realistic. The function only generates single last names currently.

### Suggested solution

Add a boolean option ```hyphenated?```

ST-DDT:
IMO there are three possible solutions.

1) Extend the locales
2) Add parameter
3) Random chance

As I'm not sure whether this actually a thing in all locales, maybe we should add it to the locale files directly?

import-brain:
@ST-DDT Yes, I don't think hyphenated last names are a thing in all locales. On another somewhat related note, what about locales with two or more surnames? Some Spanish names have multiple surnames.

ST-DDT:
> On another somewhat related note, what about locales with two or more surnames? Some Spanish names have multiple surnames.

IMO if you request a single surname, you only get one (hyphenated or not).
We could/should use a name pattern for that.

https://github.com/faker-js/faker/blob/a64cbde7c9cb05d97617708710c572b23a382f6d/src/locales/en/name/name.ts#L2-L7

This is the old format that was never usable.
But I think, we should revive it using `faker.fake` and treat all `{{patterns}}` that don't refer to methods as references to the locale data. (Basically an automatic `faker.random.arrayElement(faker.definitions.<pattern>)`)

That way we can generate `firstname lastname`, `lastname firstname` or `firstname lastname lastname` easily per locale.

import-brain:
> > On another somewhat related note, what about locales with two or more surnames? Some Spanish names have multiple surnames.
> 
> IMO if you request a single surname, you only get one (hyphenated or not). We could/should use a name pattern for that.
> 
> https://github.com/faker-js/faker/blob/a64cbde7c9cb05d97617708710c572b23a382f6d/src/locales/en/name/name.ts#L2-L7
> 
> This is the old format that was never usable. But I think, we should revive it using `faker.fake` and treat all `{{patterns}}` that don't refer to methods as references to the locale data. (Basically an automatic `faker.random.arrayElement(faker.definitions.<pattern>)`)
> 
> That way we can generate `firstname lastname`, `lastname firstname` or `firstname lastname lastname` easily per locale.

Yes, this method would work really well.

ST-DDT:
[...]

PR #1818, opened 2023-02-08
fix(person): avoid repeated double last name in some locales

### LINKED ISSUE #1817: Double last names in some locales
### Pre-Checks

### Describe the bug

#1637 has caused a problem in some locales for example `es` which have name patterns like

`'{{person.firstName}} {{person.lastName}} {{person.lastName}}'`

### Minimal reproduction code
```
const {faker} = require("@faker-js/faker")
faker.locale = "es"
faker.person.fullName() //'Anni Cervantes Cervantes'
```

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) x64 Apple M1
    Memory: 37.86 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.21.2 - /usr/local/bin/node
    npm: 8.19.3 - /usr/local/bin/npm
  Browsers:
    Chrome: 110.0.5481.77
    Firefox: 109.0.1
    Safari: 16.2
```

- [X] CJS
- [ ] ESM

### Used Package Manager

pnpm

matthewmayer:
fix #1817 

Since faker.mustache will always replace the same placeholder with the same name, tried introducing lastName2

If you have any better ideas for this let me know.

Note this affects #691 (the hyphenated patterns discussed there are already used in some locales)

### DIFF
--- a/src/locales/nb_NO/person/name.ts
+++ b/src/locales/nb_NO/person/name.ts
@@ -8,11 +8,11 @@ export default [
     weight: 1,
   },
   {
-    value: '{{person.firstName}} {{person.firstName}} {{person.lastName}}',
+    value: '{{person.firstName}} {{person.firstName}} {{person.lastName2}}',
     weight: 1,
   },
   {
-    value: '{{person.firstName}} {{person.lastName}} {{person.lastName}}',
+    value: '{{person.firstName}} {{person.lastName}} {{person.lastName2}}',
     weight: 1,
   },
   { value: '{{person.firstName}} {{person.lastName}}', weight: 8 },
--- a/src/locales/en_GH/person/name.ts
+++ b/src/locales/en_GH/person/name.ts
@@ -1,7 +1,7 @@
 export default [
   { value: '{{person.firstName}} {{person.lastName}}', weight: 1 },
   {
-    value: '{{person.firstName}} {{person.lastName}}-{{person.lastName}}',
+    value: '{{person.firstName}} {{person.lastName}}-{{person.lastName2}}',
     weight: 1,
   },
 ];
--- a/src/locales/es/person/name.ts
+++ b/src/locales/es/person/name.ts
@@ -1,11 +1,11 @@
 export default [
   {
     value:
-      '{{person.prefix}} {{person.firstName}} {{person.lastName}} {{person.lastName}}',
+      '{{person.prefix}} {{person.firstName}} {{person.lastName}} {{person.lastName2}}',
     weight: 1,
   },
   {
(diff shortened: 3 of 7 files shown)


src/locales/nb_NO/person/name.ts
```diff
@@ -8,11 +8,11 @@ export default [
     weight: 1,
   },
   {
-    value: '{{person.firstName}} {{person.firstName}} {{person.lastName}}',
+    value: '{{person.firstName}} {{person.firstName}} {{person.lastName2}}',
```
ST-DDT, line 11: This pattern only has a single last name.

ST-DDT (commented):
I think this PR tries to fix a problem, without taking the cause of the problem into consideration.

There are two problems here:
1. The patterns seem to express an additional context to the last names that they are unable to express: The person doesn't have a single last name, but two.
2. Some of the patterns seem to modify the exontext of last name to express complex last names such as combined/joined last names.

IMO the real solution to this problem is removing the repeated lastNames from the pattern and actually pass the entire last name to the method directly. (Where would the other last name come from anyway in legal documents? "John" + "Smith" ⇏ "John Smith Foobar")

If there is a locale that actually has two separate last names, then it has to use a different/not yet existing method or parameter set for it. If a locale has some kind of double last name, then it should have them in the locale data itself. Similar to thetussenvoegsel (#1778).

What do you think?

matthewmayer:
I don't think it's quite like tuessenvoegels because there's no "common" double or hyphenated names. In most cultures it's just your mothers and fathers surnames concatenated in some way. So if there are N simple last names there are N^2 compound last names. 

I guess we could add an optional last_name_pattern. Then if provided that would affect both lastName() and fullName()

ST-DDT:
> there's no "common" double or hyphenated names.

It doesn't has to, just have them in our sample set.

> I guess we could add an optional last_name_pattern.

That might be a alternative worth exploring. (We originally planned to remove all optional data access, so we could throw a useful error message instead of returning undefined, but that is something we can keep in mind when implementing this.)

matthewmayer:
> > I guess we could add an optional last_name_pattern.
> 
> That might be a alternative worth exploring. (We originally planned to remove all optional data access, so we could throw a useful error message instead of returning undefined, but that is something we can keep in mind when implementing this.)



i tried out this strategy in https://github.com/faker-js/faker/pull/1819

ST-DDT:
**Team Decision**

Closing in favor of #1819.

PR #1804, opened 2023-02-01
refactor(datatype): standardize arguments

xDivisionByZerox:
Part of #1349.

Standardize all function signatures in the `DatatypeModule` to support an options object.

PR #1821, opened 2023-02-11
refactor(finance): standardize arguments

xDivisionByZerox:
Part of #1349.

Standardize all function signatures in the `FinanceModule` to support an options object (if they have parameters).


xDivisionByZerox:
I have the save issue which I had [here](https://github.com/faker-js/faker/pull/1804#issuecomment-1412628891). This time I also did `pnpm run build` beforehand. Still, no files changed locally.

Deleting `test/scripts/apidoc/temp/` as [@Shinigami92 described](https://github.com/faker-js/faker/pull/1804#issuecomment-1412668421) didn't work for me as well. Neither did deleting the snapshot files itself.

xDivisionByZerox:
Ok, I took a closer look into [the actual errors](https://github.com/faker-js/faker/actions/runs/4152142713/jobs/7182903804#step:7:347) in the pipeline. Due to my locale machine being setup with different localization than the CI one and `faker.finance.amount` being able to [return a localized value](https://github.com/faker-js/faker/blob/49512b93722bdd89b444b25c891bcb5c20ecd225/src/modules/finance/index.ts#L548-L552) the snapshots will always be different.

A quick fix would be to only call parameters for the snapshots which take the "non-localized" code path. But that doesn't solve the real issue behind this.

matthewmayer:
I feel it should call toLocaleString(faker.locale) so it's not dependent on your current environments locale

ST-DDT:
The `faker.locale` as defined right now is unsuitable for that as it contains more of an arbritary name instead of a supported locale code.

ST-DDT:
> Due to my locale machine being setup with different localization than the CI one and `faker.finance.amount` being able to [return a localized value](https://github.com/faker-js/faker/blob/49512b93722bdd89b444b25c891bcb5c20ecd225/src/modules/finance/index.ts#L548-L552) the snapshots will always be different.

What changed so that you only get the error now and not before?
Did you add a test that wasnt there before?

matthewmayer:
> The `faker.locale` as defined right now is unsuitable for that as it contains more of an arbitrary name instead of a supported locale code.

If you replace - with _ you usually have a valid locale string (en, es_MX etc). As a fallback, you could catch a `Uncaught RangeError: Incorrect locale information provided` error and fallback to `en`.

matthewmayer:
> > The `faker.locale` as defined right now is unsuitable for that as it contains more of an arbitrary name instead of a supported locale code.
> 
> If you replace - with _ you usually have a valid locale string (en, es_MX etc). As a fallback, you could catch a `Uncaught RangeError: Incorrect locale information provided` error and fallback to `en`.

It seems all the locales work without errors if you do this replacement:

```js
Object.keys(faker.locales).forEach(fakerLocale => {
    try {
        const [validLocale] = Intl.getCanonicalLocales(fakerLocale.replace(/_/g, '-'));
        console.log(fakerLocale, validLocale, ((1234567.89).toLocaleString(validLocale)))
    } catch (e) {
        console.error("ERROR ON " + fakerLocale +" " +e)
    }
})
```

```txt
af_ZA af-ZA 1 234 567,89
ar ar ١٬٢٣٤٬٥٦٧٫٨٩
az az 1.234.567,89
cz cz 1,234,567.89
de de 1.234.567,89
de_AT de-AT 1 234 567,89
de_CH de-CH 1’234’567.89
dv dv 1,234,567.89
el el 1.234.567,89
en en 1,234,567.89
en_AU en-AU 1,234,567.89
en_AU_ocker en-AU-ocker 1,234,567.89
en_BORK en-Bork 1,234,567.89
en_CA en-CA 1,234,567.89
en_GB en-GB 1,234,567.89
en_GH en-GH 1,234,567.89
en_IE en-IE 1,234,567.89
en_IN en-IN 12,34,567.89
en_NG en-NG 1,234,567.89
en_US en-US 1,234,567.89
en_ZA en-ZA 1 234 567,89
es es 1.234.567,89
es_MX es-MX 1,234,567.89
fa fa ۱٬۲۳۴٬۵۶۷٫۸۹
fi fi 1 234 567,89
fr fr 1 234 567,89
fr_BE fr-BE 1 234 567,89
fr_CA fr-CA 1 234 567,89
fr_CH fr-CH 1 234 567,89
fr_LU fr-LU 1.234.567,89
ge ge 1,234,567.89
he he 1,234,567.89
hr hr 1.234.567,89
hu hu 1 234 567,89
hy hy 1 234 567,89
id_ID id-ID 1.234.567,89
it it 1.234.567,89
ja ja 1,234,567.89
ko ko 1,234,567.89
lv lv 1 234 567,89
mk mk 1.234.567,89
nb_NO nb-NO 1 234 567,89
ne ne १२,३४,५६७.८९
nl nl 1.234.567,89
nl_BE nl-BE 1.234.567,89
pl pl 1 234 567,89
pt_BR pt-BR 1.234.567,89
pt_PT pt-PT 1 234 567,89
ro ro 1.234.567,89
ru ru 1 234 567,89
sk sk 1 234 567,89
sv sv 1 234 567,89
tr tr 1.234.567,89
uk uk 1 234 567,89
ur ur 1,234,567.89
vi vi 1.234.567,89
zh_CN zh-CN 1,234,567.89
zh_TW zh-TW 1,234,567.89
zu_ZA zu-ZA 1,234,567.89
```

xDivisionByZerox:
> What changed so that you only get the error now and not before?
> Did you add a test that wasnt there before?

Yes. In `amount()` the `autoFormatted` parameter was not tested before. [So I added a case for that.](https://github.com/faker-js/faker/pull/1821/files#diff-228b643e5b81a3b42387fe9cdc8f3212927388d9af162a12f1cc2d060aeeda94R55-R61)

ST-DDT:
> > What changed so that you only get the error now and not before?
> > Did you add a test that wasnt there before?
> 
> Yes. In `amount()` the `autoFormatted` parameter was not tested before. [So I added a case for that.](https://github.com/faker-js/faker/pull/1821/files#diff-228b643e5b81a3b42387fe9cdc8f3212927388d9af162a12f1cc2d060aeeda94R55-R61)

Maybe revert that added test and create a separate issue for that bug.
IMO the bug is unrelated to this change and shouldnt be addressed in the same PR.

matthewmayer:
calling toLocaleString with an undefined first parameter just seems like a Bad Idea, because it makes the output of Faker dependent on the *system* locale rather than the Faker locale.

ST-DDT (approved):

Shinigami92 (approved):

PR #1819, opened 2023-02-09
fix(person): avoid repeated last name via new last_name_patterns

matthewmayer:
fix #1817

an alternative strategy which allows for compound and hyphenated last names. The advantage of this is that it works for both `faker.person.lastName()` as well as `faker.person.fullName()`

```js
faker.locale="es"

faker.person.lastName() // 'Garibay Piña'
faker.person.fullName() // 'Marta Menchaca Vergara'
```

<details>
<summary>Here's 100 fullNames for each of the affected locales</summary>

```js
["es","es_MX","en_GH", "lv", "nb_NO", "sv"].forEach(locale => {faker.locale = locale; [...Array(100)].map(x=>console.log(locale, faker.person.fullName()))})
```
```
es Pedro Monroy Navarro
es Sra. Roser Vaca Rendón
es Lorena Valencia Marrero
es Pedro Arteaga Nevárez
es Pablo Duarte Pelayo
es Sra. Maica Cotto Feliciano
es Lorena Urrutia Velasco
es Pedro Juárez Casares
es Sra. Anni Abreu Lucero
es Iván Zelaya Almaráz
es Pablo Palacios Lucio
es Andrea Henríquez Verduzco
es Miguel Olmos Cano
es Iván Pedroza Olivera
es Lorena Estévez Zayas
es Salvador Sierra Mojica
es Roser Muro Navarro
es Ricardo Anaya Medina
es Ángel Abreu Balderas
es María de los Ángeles Salas Rivero
es Anni Carrillo Lemus
es Jennifer Villaseñor Tello
es Miguel Altamirano Laboy
es Laura Cortéz Rolón
es Jordi Gurule Nava
es Sra. Maica Leyva Colunga
es Anni Menchaca Colón
es Maica Lugo Alonso
es Andrea Prieto Lovato
es Jorge Maldonado Medina
es Ricardo Montalvo Armas
es Carles Zepeda Linares
es María de los Ángeles Velasco Quintanilla
es Anni Corona Casares
es Pío Meraz Olivas
es Lorena Gurule Atencio
es Lorena Aguilera Carrero
es Hermenegildo Rosado Alanis
es María de los Ángeles Mercado Negrón
es Jennifer Deleón Muñoz
es Salvador Brito Santana
es Ana Puga Villaseñor
es Jorge Nevárez Serna
es Lorena Garica Lomeli
es Sra. Andrea Quiñónez Osorio
es Pablo Torres Reséndez
es Jennifer Huerta Niño
[...]

PR #1822, opened 2023-02-12
fix(person): fix duplicate last names in full name generation

### LINKED ISSUE #1817: Double last names in some locales

0916dhkim:
Fixes #1817 

#### As Is
`faker.person.fullName` uses mustache templates from each locale to generate a full name. When the template contains multiple last names, the same last name is repeated. Same bug happens to multiple first names. This is because `firstName` and `lastName` are generated before the template replacement unlike `middleName` or `suffix` which are generated on demand.

#### To Be
Make `firstName` and `lastName` get generated on demand similar to `middleName`. Only fallback to `options.firstName` and `options.lastName` when those are supplied.

### DIFF
--- a/src/modules/person/index.ts
+++ b/src/modules/person/index.ts
@@ -183,8 +183,8 @@ export class PersonModule {
   ): string {
     const {
       sex = this.faker.helpers.arrayElement([Sex.Female, Sex.Male]),
-      firstName = this.firstName(sex),
-      lastName = this.lastName(sex),
+      firstName,
+      lastName,
     } = options;
 
     const fullNamePattern: string = this.faker.helpers.weightedArrayElement(
@@ -193,9 +193,9 @@ export class PersonModule {
 
     const fullName = this.faker.helpers.mustache(fullNamePattern, {
       'person.prefix': () => this.prefix(sex),
-      'person.firstName': () => firstName,
+      'person.firstName': () => firstName ?? this.firstName(sex),
       'person.middleName': () => this.middleName(sex),
-      'person.lastName': () => lastName,
+      'person.lastName': () => lastName ?? this.lastName(sex),
       'person.suffix': () => this.suffix(),
     });
     return fullName;
--- a/test/__snapshots__/person.spec.ts.snap
+++ b/test/__snapshots__/person.spec.ts.snap
@@ -6,15 +6,15 @@ exports[`person > 42 > firstName > noArgs 1`] = `"Garnet"`;
 
 exports[`person > 42 > firstName > with sex 1`] = `"Glen"`;
 
-exports[`person > 42 > fullName > noArgs 1`] = `"Sadie Wiegand"`;
+exports[`person > 42 > fullName > noArgs 1`] = `"Dr. Tina Reynolds"`;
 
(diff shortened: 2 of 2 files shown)


ST-DDT:
Could you please share your insights on the last name matter a bit?
Do these languages have two distinct last names (e.g. mothers and fathers last name) or one last name consisting of multiple parts (space or hyphen separated)?
E.g. in Germany you can decide to add your spouses name to your name, but then you have a new single last name that you have use then.
The individual parts are no longer relevant excepts for personal reasons.

madcampos:
Hi @ST-DDT just chiming in to give an insight for Brazilian Portuguese.

It is a tradition that the children get both parents last names, but it varies from place to place and is not a hard rule.

For example, my mother has a super simple name, _Lucrécia Dilda_. When my parents were married, she had my father's last name too.

On the other hand, my father has the same name of my grandfather, _Francisco Aurélio da Silva Campos Filho_. His first name would be "_Francisco Aurélio_", then the rest is his last name, and "_Filho_" means "son" to mark he has his father's name.

I got both last names, so my full name is _Marco Aurélio Dilda Campos_.

So yeah, it can be multiple different names.

ST-DDT:
Thanks @madcampos for your input.

> I got both last names, so my full name is _Marco Aurélio Dilda Campos_.
> 
> So yeah, it can be multiple different names.

Just for clarification, is "Dilda Campos" your last name or "Dilda" AND "Campos"? Is it possible for you to flip them the other way round e.g. "Campos Dilda"?
How is it handled in web forms? Do you have two inputs for the last name or do you put both in one?

madcampos:
@ST-DDT So, to be super semantic here...

In Brazil the whole thing is considered your _surname_, in that case "Dilda Campos", and _last name_ would be your very last name, in this case only "Campos". I cannot flip them around.

In web forms outside of Brazil I use my very first and very last name and omit everything else, so my name would be only "Marco Campos".

For forms in Brazil it is common to have a single field for full name, if it happens to be two fields, then I would input "Marco Aurélio" as first name/given name and "Dilda Campos" as last name/surname

matthewmayer:
Note that the naming of the Faker methods isn't perfect anyway, "lastName" should probably be familyName and "firstName" should be givenName as for example in Chinese the family name comes first.

ST-DDT:
**Team Decision**

Closing in favor of #1819.

PR #1823, opened 2023-02-12
fix(locales): decrease all locales that has excessive number to fixed 1000

### LINKED ISSUE #1802: `hr` locale has excessive number of lastNames
https://github.com/faker-js/faker/blob/next/src/locales/hr/person/last_name.ts has more than 11,000 names and is 166KB in size. This is 5 times larger than any other locale (`it` is next with 2170).

Maybe just randomly pick a 20% subset to keep?

benyaminbeyzaie:
Solves https://github.com/faker-js/faker/issues/1802

I wrote down a script to do that, so tell me if there is a change on the fixed number of 1100 needed!


ST-DDT:
Please fix the issues reported by our ci.

ST-DDT:
> so tell me if there is a change on the fixed number of 1100 needed!

We currently plan to have a limit of 1000 entries/file.

Could you share how you selected the elements your set/share the script that you have used?

matthewmayer (changes requested):
* The entries should be sorted alphabetical order
* Run `pnpm run format` to lint the files (e.g. there should be a newline after each entry)
* Could maybe add a test in https://github.com/faker-js/faker/blob/next/test/locales.spec.ts similar to `expect(data).not.toContainDuplicates();` which checks that each entry is <1000 lines
* I wonder if a filesize limit would make more sense than a simple length limit. 100 long entries make much more difference to the filesize than 1000 short entries?

benyaminbeyzaie:
The file size is better but I guess limit on the length would do the job for now.

test/locales.spec.ts
```diff
@@ -15,11 +15,14 @@ describe('locale', () => {
...
   });
 
-  function checkLocaleData(data: unknown) {
+  function checkLocaleData(data: any) {
     if (Array.isArray(data)) {
       it('should not have duplicate entries', () => {
         expect(data).not.toContainDuplicates();
       });
+      it('should not be more than 1000 entry', () => {
+        expect(data.length).toBeLessThan(1001);
```
Shinigami92: I think there is a less then equals, is it?
That would fit better the description text of this test

test/locales.spec.ts
```diff
@@ -15,11 +15,14 @@ describe('locale', () => {
     });
   });
 
-  function checkLocaleData(data: unknown) {
+  function checkLocaleData(data: any) {
```
Shinigami92: Is it needed to have this any?
I would assume that `Array.isArray(data)` would convert it to `unknown[]` but I might be wrong 🤔

matthewmayer:
Various CI errors
```
❯ test/locales.spec.ts > locale > cz > person > first_name > should not be more than 1000 entry
     → expected 1578 to be less than 1001
   ❯ test/locales.spec.ts > locale > cz > person > last_name > should not be more than 1000 entry
     → expected 1979 to be less than 1001
   ❯ test/locales.spec.ts > locale > de > location > street_name > should not be more than 1000 entry
     → expected 1031 to be less than 1001
   ❯ test/locales.spec.ts > locale > de > word > adverb > should not be more than 1000 entry
     → expected 1028 to be less than 1001
   ❯ test/locales.spec.ts > locale > de_AT > person > first_name > should not be more than 1000 entry
     → expected 1145 to be less than 1001
   ❯ test/locales.spec.ts > locale > en > internet > emoji > person > should not be more than 1000 entry
     → expected 1803 to be less than 1001
   ❯ test/locales.spec.ts > locale > fr > animal > bird > should not be more than 1000 entry
     → expected 1077 to be less than 1001
   ❯ test/locales.spec.ts > locale > fr > word > verb > should not be more than 1000 entry
     → expected 1001 to be less than 1001
   ❯ test/locales.spec.ts > locale > fr_CH > animal > bird > should not be more than 1000 entry
     → expected 1077 to be less than 1001
   ❯ test/locales.spec.ts > locale > fr_CH > word > verb > should not be more than 1000 entry
     → expected 1001 to be less than 1001
   ❯ test/locales.spec.ts > locale > it > person > male_first_name > should not be more than 1000 entry
     → expected 1083 to be less than 1001
   ❯ test/locales.spec.ts > locale > nl_BE > location > city_prefix > should not be more than 1000 entry
     → expected 1100 to be less than 1001
   ❯ test/locales.spec.ts > locale > sk > location > street_name > should not be more than 1000 entry
     → expected 1040 to be less than 1001
   ❯ test/locales.spec.ts > locale > vi > person > first_name > should not be more than 1000 entry
     → expected 1981 to be less than 1001
```
- there do seem to be a bunch of files in the 1000-1100 range, suggest make the limit 1100 for now then.
- you should exclude files from checks which are made by concatenating other files e.g. a lot of the first_name files are a concat of male_first_name and female_last_name
- you should exclude emoji since these are very short?

Also you need to update the snapshot tests - run `pnpm run test` and then u to update

benyaminbeyzaie:
I fixed arrays length on 1000 and edit the test, could you please suggest a way to run a single test locally?

Also I had to change some mergeArrays and took sample size of 1000 of them, we could add loadash to the dependency to took those samples dynamically but I guess this is fine right now.

benyaminbeyzaie:
Also, here is the small script that I wrote down to edit the codes, to run put it in the main folder of the project with name samller.ts, change the type in package.json to commonJs, the run with `ts-node-esm  smaller.ts`

```
import fs from 'fs';
import _ from 'lodash';
import path from 'path';

function requireAll(dir) {
  const files = fs.readdirSync(dir);

  files.forEach((file) => {
    const filePath = path.join(dir, file);
    const stats = fs.statSync(filePath);

    if (stats.isDirectory()) {
      requireAll(filePath);
    } else if (
      stats.isFile() &&
      path.extname(file) === '.ts' &&
      file !== 'index.ts' &&
      file
    ) {
      console.log(filePath);
      import('./' + filePath)
        .then((array) => {
          console.log(array.default.length);
          if (array.default.length > 1000) {
            const newArray = _.sampleSize(array.default, 1000);
            // sort newArray alphabatical
            newArray.sort((a, b) => {
              if (a < b) {
                return -1;
              }
              if (a > b) {
                return 1;
              }
              return 0;
            });
            fs.writeFileSync(
              filePath,
              `export default ${JSON.stringify(newArray)};`
            );
          }
        })
        .catch((error) => {
          console.log(error);
        });
    }
  });
}

requireAll('./src/locales/');
```

benyaminbeyzaie:
Code on the script is a little bit messy, don't mind it! :))

matthewmayer:
Instead of using lodash can use faker itself (faker.helpers.arrayElements). That way you can also set a fixed seed so the result is the same every time?

ST-DDT:
We have to check / mark the merged locales, so we can "remove" them in our #1677 refactor.

benyaminbeyzaie:
Is there any other change that I've to make? Don't know what causes the CI issues.

ST-DDT:
> Is there any other change that I've to make? Don't know what causes the CI issues.

`pnpm run test -u`

benyaminbeyzaie:
result of the `pnpm run test -u`:

FAIL  test/phone.spec.ts [ test/phone.spec.ts ]
Error: No test suite found in file /Users/beyamin/Desktop/University/University07/Web/faker/test/phone.spec.ts

ST-DDT:
That's strange. I fixed all but one failing test (emojis > 1000) for you.
I basically ran these commands:

````sh
git pull
pnpm install
pnpm run build
pnpm run test -u
git add .
git commit -m "test: update snapshots"
pnpm run lint
# Fix lint errors
pnpm run lint
git commit -m "chore: fix lint errors"
pnpm run format
# No format changes to commit
git push
````

Please fix the last remaining test failure.

ST-DDT:
Superseded by #2265

PR #1833, opened 2023-02-13
test(date): add birthdate error test

PeterYinusa:
Adds a birthdate test that throws an error when the min is larger than the max year.

### DIFF
--- a/test/date.spec.ts
+++ b/test/date.spec.ts
@@ -558,6 +558,19 @@ describe('date', () => {
             new Date().getUTCFullYear() - min
           );
         });
+
+        it('should throw an error when the min > max year', () => {
+          const min = 2000;
+          const max = 1990;
+
+          expect(() =>
+            faker.date.birthdate({ min, max, mode: 'year' })
+          ).toThrow(
+            new FakerError(
+              `Max 662515200000 should be larger then min 946771200000.`
+            )
+          );
+        });
       });
 
       describe('deprecated', () => {


test/date.spec.ts
```diff
@@ -558,6 +558,19 @@ describe('date', () => {
...
+          const max = 1990;
+
+          expect(() =>
+            faker.date.birthdate({ min, max, mode: 'year' })
+          ).toThrow(
+            new FakerError(
+              `Max 662515200000 should be larger then min 946771200000.`
+            )
+          );
+        });
```
ejcheng, line 573: LGTM, but should we throw an error using the length of time in the `mode` argument rather than Unix time? Maybe something like `Year {maxYear} should be larger than min {minYear}`.

@faker-js/maintainers @faker-js/members

ejcheng (approved):

test/date.spec.ts
xDivisionByZerox: That might be good for DX, but is unrelated to this specific PR. If you want to touch this or keep track of it open an issue. please.

xDivisionByZerox (approved):
Good catch. Thank you for the contibution!

ST-DDT (approved):

test/date.spec.ts
xDivisionByZerox: I opened #1841 to track this.

ejcheng: > I opened #1841 to track this.

Thank you for creating the issue

PR #1839, opened 2023-02-14
chore: activate noImplicitAny

### LINKED ISSUE #1837: Turn on noImplicitAny in tsconfig

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/tsconfig.json#L15

Related issues:

Shinigami92:
closes #1837

This PR does only build time and no runtime changes on purpose
Otherwise I would suggest to modify some codeflow in later PRs (if really needed)

### DIFF
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -10,7 +10,9 @@ import * as uniqueExec from './unique';
 export class HelpersModule {
   constructor(private readonly faker: Faker) {
     // Bind `this` so namespaced is working correctly
-    for (const name of Object.getOwnPropertyNames(HelpersModule.prototype)) {
+    for (const name of Object.getOwnPropertyNames(
+      HelpersModule.prototype
+    ) as Array<keyof HelpersModule | 'constructor'>) {
       if (name === 'constructor' || typeof this[name] !== 'function') {
         continue;
       }
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -907,7 +907,13 @@ export class HelpersModule {
    *
    * @since 7.5.0
    */
-  unique<Method extends (...parameters: any[]) => RecordKey>(
+  unique<
+    Method extends (
+      // TODO christopher 2023-02-14: This `any` type can be fixed by anyone if they want to.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ...parameters: any[]
+    ) => RecordKey
+  >(
     method: Method,
     args?: Parameters<Method>,
     options: {
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -806,8 +806,10 @@ export class HelpersModule {
 
     // Search for the requested method or definition
     for (const part of parts) {
-      currentModuleOrMethod = currentModuleOrMethod?.[part];
-      currentDefinitions = currentDefinitions?.[part];
+      currentModuleOrMethod =
+        currentModuleOrMethod?.[part as keyof typeof currentModuleOrMethod];
+      currentDefinitions =
+        currentDefinitions?.[part as keyof typeof currentDefinitions];
     }
 
     // Make method executable
@@ -905,7 +907,7 @@ export class HelpersModule {
    *
    * @since 7.5.0
    */
-  unique<Method extends (...parameters) => RecordKey>(
+  unique<Method extends (...parameters: any[]) => RecordKey>(
     method: Method,
     args?: Parameters<Method>,
     options: {
--- a/src/modules/helpers/unique.ts
+++ b/src/modules/helpers/unique.ts
@@ -81,7 +81,13 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * @param options.compare The function used to determine whether a value was already returned. Defaults to check the existence of the key.
  * @param options.store The store of unique entries. Defaults to `GLOBAL_UNIQUE_STORE`.
  */
-export function exec<Method extends (...parameters: any[]) => RecordKey>(
+export function exec<
+  Method extends (
+    // TODO christopher 2023-02-14: This `any` type can be fixed by anyone if they want to.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    ...parameters: any[]
+  ) => RecordKey
+>(
   method: Method,
   args: Parameters<Method>,
   options: {
--- a/src/modules/helpers/unique.ts
+++ b/src/modules/helpers/unique.ts
@@ -81,7 +81,7 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * @param options.compare The function used to determine whether a value was already returned. Defaults to check the existence of the key.
  * @param options.store The store of unique entries. Defaults to `GLOBAL_UNIQUE_STORE`.
  */
-export function exec<Method extends (...parameters) => RecordKey>(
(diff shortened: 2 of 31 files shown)


src/modules/helpers/unique.ts
```diff
@@ -81,7 +81,7 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * @param options.compare The function used to determine whether a value was already returned. Defaults to check the existence of the key.
  * @param options.store The store of unique entries. Defaults to `GLOBAL_UNIQUE_STORE`.
  */
-export function exec<Method extends (...parameters) => RecordKey>(
+export function exec<Method extends (...parameters: any[]) => RecordKey>(
```
xDivisionByZerox: <s>This can be expressed in a more generic way, to avoid explicit any (same in index.ts):

```suggestion
export function exec<Method extends <T>(...parameters: T[]) => RecordKey>(
```

@ST-DDT, can you confirm this? I'm "only" 90% sure that it works as expected. </s>

xDivisionByZerox: NVM, I had massive brain lag or something... this doesn't work as we require the parameters generically as well. MB.

src/utils/types.ts
```diff
@@ -12,7 +12,7 @@ export type LiteralUnion<T extends U, U = string> =
  *
  * For some strange reason this is not the same as `Function`.
  */
-export type Callable = (...args) => unknown;
+export type Callable = (...args: any[]) => unknown;
```
xDivisionByZerox: <s>This can be typed generically to avoid explicit `any`:

```suggestion
export type Callable = <T>(...args: T[]) => unknown;
```
</s>

NVM, doesn't work when you have multiple parameters with different types.
BUT: Since we are only interested in whether a type is a function (callable), this type can be narrowed to `export type Callable = () => void;`, since all functions match this signature.

```suggestion
export type Callable = () => void;
```

The description might need an update as well.

Shinigami92: I'm getting this

image

Please help me further
I suggest checkout the branch and try to handle it on your own with your knowledge

xDivisionByZerox: You are too quick xD. I updated my previous comment.

Shinigami92: Okay I mark these as resolved

xDivisionByZerox: Please read my suggestion again. Guess you only read the first line.

> Since we are only interested in whether a type is a function (callable), this type can be narrowed to `export type Callable = () => void;`

Shinigami92:
@xDivisionByZerox Oh I see, the Callable was just a replacement for `Function` which should not be used in TS as a type as it is a real thing (e.g. like String)

The suggested way is just to use `() => void` instead and therefore I removed the type alias and inlined it as this is the way it should be done and is suggested by TS

This also removed the _bad worded_ comment `For some strange reason ...` which sounds a bit unprofessional

Shinigami92:
The "test"-code failed 👀 
I need to check that later, I will leave my home soon and come back in some hours

src/modules/helpers/index.ts
```diff
@@ -903,7 +907,7 @@ export class HelpersModule {
    *
    * @since 7.5.0
    */
-  unique<Method extends (...parameters) => RecordKey>(
+  unique<Method extends (...parameters: any[]) => RecordKey>(
```
ST-DDT: This has a lint warning.

src/modules/helpers/unique.ts
ST-DDT: Same here

src/utils/types.ts
```diff
@@ -7,17 +7,10 @@ export type LiteralUnion<T extends U, U = string> =
...
- *
- * For some strange reason this is not the same as `Function`.
- */
-export type Callable = (...args) => unknown;
-
 /**
  * Type that represents a single method/function name of the given type.
  */
-export type MethodOf<ObjectType, Signature extends Callable = Callable> = {
+export type MethodOf<ObjectType, Signature extends () => void = () => void> = {
```
ST-DDT: Maybe

```suggestion
export type MethodOf<ObjectType, Signature extends () => unknown = () => unknown> = {
```

Not tested

src/utils/types.ts
```diff
@@ -30,5 +23,5 @@ export type MethodOf<ObjectType, Signature extends Callable = Callable> = {
  */
 export type MethodsOf<
   ObjectType,
-  Signature extends Callable = Callable
+  Signature extends () => void = () => void
```
ST-DDT: Same here

ST-DDT:
> The suggested way is just to use `() => void` instead and therefore I removed the type alias and inlined it as this is the way it should be done and is suggested by TS
> 
> This also removed the _bad worded_ comment `For some strange reason ...` which sounds a bit unprofessional

Soon:

src/modules/helpers/index.ts
Shinigami92: Yes, this is okay for this PR and should be addressed in a separate PR if needed
This PR just aims first to toggle on this TypeScript flag and is not related to eslint

ST-DDT: At least mark them as ignored, otherwise they will be shown in Github in every PR as warnings.

src/utils/types.ts
```diff
@@ -8,11 +8,15 @@ export type LiteralUnion<T extends U, U = string> =
   | (U & { zz_IGNORE_ME?: never });
 
 /**
- * Basically a function that returns a value.
+ * A function that returns a value.
  *
- * For some strange reason this is not the same as `Function`.
+ * This is a workaround for the fact that `Function` is a real JS Object like `String` and therefore should not be used as a type.
```
xDivisionByZerox: That's not the correct for why the linter doesn't allow it. The `Function` type satisfies any function. This includes classes. So while this is correct from a type perspective it not desired most of the time:

```ts
function foo(fn: Function) {
  return fn();
}

foo(() => null); // this is fine and desired

function bar() {  
  return null; 
}
foo(); // this is fine and desired

class A { }
foo(A); // this is fine for typescript but not desired as `A` will be called without the `new` keyword which throws an error at runtime
```

---

Just wanted to point that out, since it has nothing to do with the actual type values being used as a type.

Shinigami92: Would you suggest something different?

Here might be some context for improved text: https://typescript-eslint.io/rules/ban-types

ST-DDT (dismissed):

src/utils/types.ts
xDivisionByZerox: How about this?

```suggestion
 * `Function` cannot be used instead of this, as it doesn't accept accepts class declarations.
 * Those would fail when called, as they are called without the `new` keyword.
```

src/utils/types.ts
```diff
@@ -8,11 +8,16 @@ export type LiteralUnion<T extends U, U = string> =
   | (U & { zz_IGNORE_ME?: never });
 
 /**
- * Basically a function that returns a value.
+ * A function that returns a value.
  *
- * For some strange reason this is not the same as `Function`.
+ * `Function` cannot be used instead of this, as it doesn't accept accepts class declarations.
+ * Those would fail when called, as they are called without the `new` keyword.
```
Shinigami92:
```suggestion
 * `Function` cannot be used instead because it doesn't accept class declarations.
 * These would fail when invoked since they are invoked without the `new` keyword.
```

ST-DDT (approved):

xDivisionByZerox (approved):

Issue #1835, opened 2023-02-13
Turn on strictBindCallApply in tsconfig


ST-DDT:
https://github.com/faker-js/faker/pull/651 partially activated strict mode, this flag was omitted at that time and should probably be turned on.

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/tsconfig.json#L12

Related issues:

- https://github.com/faker-js/faker/pull/651
- https://github.com/faker-js/faker/issues/1834
- https://github.com/faker-js/faker/issues/1836
- https://github.com/faker-js/faker/issues/1837
- https://github.com/faker-js/faker/issues/1838

Shinigami92:
I tried to set `strictBindCallApply: true` but I have a problem in line

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/src/modules/helpers/unique.ts#L140

But as we want to get rid of this anyway
- #1785

I will put no more effort into `strictBindCallApply` until then

ST-DDT:
Is it possible to just replace `this` with `undefined`?

Shinigami92:
> Is it possible to just replace `this` with `undefined`?

Could you explain a bit how you come to this conclusion? I would like to understand what your thought-process is.

ST-DDT:
Since the value isnt bound to anything useful in the first place (aka this isnt bound to anything specific since it isnt even in the helpers module directly and thus doesnt get bound by the constructor). We could just drop it entirely.
Or am I missing something?

Shinigami92:
<img width="800" alt="image" src="https://user-images.githubusercontent.com/7195563/218711706-287a456f-9fcf-4724-b884-fd8b612e50be.png">

I suggest setting `strictBindCallApply` locally and try around a bit on your own
That could speed up the process and doesn't require so much writing here

ST-DDT:
Done: https://github.com/faker-js/faker/pull/1842

ST-DDT:
Blocked by #1842

---

I no longer know how to fix these lines:

````ts
this[name] = this[name].bind(this);
````

For most modules I can simply do:

````ts
this[name] = this[name].bind(this, ...[]);
````

but that doesn't work for the helpers module.

ST-DDT:
Blocked by #2102.

- #2102

xDivisionByZerox:
I will close this as this has been activated implicitly in #2102.

PR #1842, opened 2023-02-14
chore: turn on noImplicitThis

### LINKED ISSUE #1838: Turn on noImplicitThis in tsconfig

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/tsconfig.json#L16

Related issues:

ST-DDT:
Fixes ~~#1835~~, #1838

- ~~#1835~~
- #1838

Both rules need to be changed at the same time, because they cannot be fixed individually.


Shinigami92 (commented):
As already discussed in Discord, but to transparently ask this here again:
Does this still work with something like
```ts
class X {
  y = 1
  a() {
    return this.y
  }
}
const x = new X()
unqiue(x.a) // is y still bound to the class instance?
```

ST-DDT:
In v7.6 and next it outputs undefined.
After this PR it will throw an error.

Neither is correct.

FFR: In v7.6 and next `this` is `{ exec: [Getter] }` which isn't useful at all.

FWIW Both variants work when using `faker.helpers.unique(x.a.bind(x))`.

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #1843, opened 2023-02-14
chore: improve error message for faker.date.birthday

### LINKED ISSUE #1841: Improve error message in `birthday`
## Description

Currently, when `faker.date.birthday` is called with parameters where `min` > `max` an error in the form of `Max {number} should be larger than min {number}.` is thrown. Here, `{number}` isn't the input years (as one would expect), but the UNIX timestamp of the generated Dates.

## Solution

Move the validation to the top of the function and do it against the input parameters. Use the input parameters in the error message.

## Additional Information

Brought up in [#1833](https://github.com/faker-js/faker/pull/1833#discussion_r1105079757).

johnmal-dev:
faker-js/faker#1841

### DIFF
--- a/src/modules/date/index.ts
+++ b/src/modules/date/index.ts
@@ -947,6 +947,10 @@ export class DateModule {
       refDate?: string | Date | number;
     } = {}
   ): Date {
+    if (options.max < options.min) {
+      throw new FakerError(`Max ${options.max} should be larger than or equal to min ${options.min}.`);
+    }
+
     const mode = options.mode === 'age' ? 'age' : 'year';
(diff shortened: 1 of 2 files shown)


ST-DDT:
Looks like you have to run the formatter once:

`pnpm run format`

xDivisionByZerox:
I would categorize this as `chore` instead of `fix`. @Shinigami92 @ST-DDT other opinions?

ST-DDT (approved):

xDivisionByZerox (approved):
Thank you for your contribution! ❤️

PR #1847, opened 2023-02-16
feat(locale): add `state` and `state_abbr` to `cz`

xDivisionByZerox:
Fill the `states` dataset for the locale `cz`.
In the  Czech Republic "states" are called "regions". They are not the same but serve the same purpose for our data.

---

- [source I used](https://en.wikipedia.org/wiki/ISO_3166-2:CZ#Regions)
- [More information on Czech administrative divisions](https://en.wikipedia.org/wiki/Czech_Republic#Administrative_divisions)


xDivisionByZerox:
I was not able to find anything regarding state abbreviations. Should I simply copy the existing dataset in that case?

test/all_functional.spec.ts
```diff
@@ -20,13 +20,13 @@ function isMethodOf(mod: string) {
 }
 
 const BROKEN_LOCALE_METHODS = {
-  // TODO ST-DDT 2022-03-28: these are TODOs (usually broken locale files)
+  // TODO ST-DDT 2022-03-28: these are TODOs (usually broken locale tschechienfiles)
```
matthewmayer: was this line changed by mistake?

matthewmayer (dismissed):

ST-DDT:
> I was not able to find anything regarding state abbreviations. Should I simply copy the existing dataset in that case?

Instead of copying it you can import it and export it afterwards (with a comment why).

test/all_functional.spec.ts
xDivisionByZerox: Yes, probably forgot to focus my browser while copy passing or something 🤔 
I'll remove this.

ST-DDT (approved):

matthewmayer (approved):

Shinigami92 (approved):

PR #1853, opened 2023-02-16
refactor(datatype): deprecate `array`

### LINKED ISSUE #1851: Deprecate datatype.json() and datatype.array() for removal

The methods return data that are most likely useless for the users as it contains arbritary data.

### Suggested solution

Deprecate the method and recommend helpers.multiple (for datatype.array) and a self built method for datatype.json.

### Alternative

_No response_

### Additional context

We decided that we don't want to return random "dynamic" objects from our API unless the user explicitly passed the inputs for the method: https://github.com/faker-js/faker/issues/1733#issuecomment-1433439926

xDivisionByZerox:
Part of #1851.

### DIFF
--- a/docs/guide/upgrading.md
+++ b/docs/guide/upgrading.md
@@ -127,7 +127,7 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
 ### Deprecation of `datatype.array`
 
 The method `faker.datatype.array` has been deprecated and will be removed in v9.
-If you need an array of element you are better off creating your own one using `helpers.multiple`.
+If you need an array of useful values, you are better off creating your own one using `helpers.multiple`.
 
 ### `allowLeadingZeros` behavior change in `faker.string.numeric`
 
--- a/docs/guide/upgrading.md
+++ b/docs/guide/upgrading.md
@@ -124,7 +124,7 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
 | `faker.datatype.float`  | `faker.number.float`                       |
 | `faker.datatype.bigInt` | `faker.number.bigInt`                      |
 
-### Deprecation of `datatype.array`
+### Deprecation of `faker.datatype.array`
 
 The method `faker.datatype.array` has been deprecated and will be removed in v9.
 If you need an array of useful values, you are better off creating your own one using `faker.helpers.multiple`.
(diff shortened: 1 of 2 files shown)


src/modules/datatype/index.ts
```diff
@@ -425,6 +427,13 @@ export class DatatypeModule {
           max: number;
         } = 10
   ): Array<string | number> {
+    deprecated({
+      deprecated: 'faker.datatype.array()',
+      proposed: 'your own function to build complex arrays',
```
ST-DDT, line 441: IMO faker.helpers.multiple would work here better.

xDivisionByZerox: I think `helpers.multiple` and `datatype.array` are used for different use cases. `multiple` is for values of the same kind (multiple names, multiple hex numbers, etc.) while `array` gave you different values per element (well eighter `string` or `number`). Sure that's easy to achieve with `mutiple`, but its not a drop-in replacement which I thought the `proposed` property described.

There should probably be a "bigger" section in the migration guide on how to use `multiple` to mimic this function but for the warning in the code itself it doesn't fit.

At least all of that is how I see this.

ST-DDT: > There should probably be a "bigger" section in the migration guide on how to use `multiple` to mimic this function but for the warning in the code itself it doesn't fit.

Then you can add it as part of this PR. So we don't forget about it.

docs/guide/upgrading.md
```diff
@@ -124,6 +124,11 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
 | `faker.datatype.float`  | `faker.number.float`                       |
 | `faker.datatype.bigInt` | `faker.number.bigInt`                      |
 
+### Deprecation of `datatype.array`
+
+The method `faker.datatype.array` has been deprecated and will be removed in v9.
+If you need an array of element you are better off creating your own one using `helpers.multiple`.
```
```suggestion
If you need an array of useful values, you are better off creating your own one using `helpers.multiple`.
```

ST-DDT (dismissed):

Shinigami92 (dismissed):

docs/guide/upgrading.md
```diff
@@ -124,6 +124,11 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
 | `faker.datatype.float`  | `faker.number.float`                       |
 | `faker.datatype.bigInt` | `faker.number.bigInt`                      |
 
+### Deprecation of `datatype.array`
+
+The method `faker.datatype.array` has been deprecated and will be removed in v9.
+If you need an array of useful values, you are better off creating your own one using `helpers.multiple`.
```
matthewmayer: Using `faker.helpers.multiple`

matthewmayer (dismissed):

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #1854, opened 2023-02-17
docs: Updated first example in README to utilize faker.helpers.multiple

jhechtf:
I'll start by saying that this isn't necessary, but I was looking at the README and noticed that the example for generating fake user data was incorrectly using `Array.from` with a `forEach` and pushing into an array. This uses the initializer parameter so that the array doesn't have to be iterated over twice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,6 @@ npm install --save-dev @faker-js/faker
 import { faker } from '@faker-js/faker';
 // import { faker } from '@faker-js/faker/locale/de';
 
-export const USERS: User[] = [];
-
 export function createRandomUser(): User {
   return {
     userId: faker.datatype.uuid(),
@@ -74,9 +72,7 @@ export function createRandomUser(): User {
   };
 }
 
-Array.from({ length: 10 }).forEach(() => {
-  USERS.push(createRandomUser());
-});
+export const const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
 ```
 
 The above code indicates a basic usage of Faker.
--- a/README.md
+++ b/README.md
@@ -72,7 +72,9 @@ export function createRandomUser(): User {
   };
 }
 
-export const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
+export const USERS: User[] = faker.helpers.multiple(generateRandomUser, {
+  count: 5,
+});
 ```
 
 The above code indicates a basic usage of Faker.
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ export function createRandomUser(): User {
   };
 }
 
-export const const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
+export const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
 ```
 
 The above code indicates a basic usage of Faker.


README.md
```diff
@@ -74,9 +72,7 @@ export function createRandomUser(): User {
   };
 }
 
-Array.from({ length: 10 }).forEach(() => {
-  USERS.push(createRandomUser());
-});
+const USERS = Array.from({ length: 10 }, () => createRandomUser());
```
ST-DDT: USERS was exported previously, was that change intentional?

matthewmayer:
Personally i wasn't even aware Array.from could take a second mapping parameter :D

I feel that as its the first example in the docs we should favour code that newbies wont struggle to understand, maybe a good old fashioned for-loop and plain Javascript so its clear the library works with both TS and JS (more fancy TS features could be shown off in later examples).

e.g.

```js
// import { faker } from '@faker-js/faker/locale/de';

const users = [];

function createRandomUser() {
  return {
    userId: faker.string.uuid(),
    username: faker.internet.userName(),
    avatar: faker.image.avatar(),
    password: faker.internet.password(),
    birthdate: faker.date.birthdate(),
    registeredAt: faker.date.past(),
  };
}
for (let i = 0; i < 10; i++) {
  users.push(createRandomUser());
}
console.dir(users)
```

Shinigami92:
I use for loops as well in my prod code
But USERS should be stay on top before createRandomUser as e.g. you can theoretically use USERS inside that function while generating users

For example:

```ts
import { Location } from 'src/locations/models/location.model';
import { createRandomNode } from './node';

export const LOCATIONS: Location[] = [];

export function createRandomLocation(): Location {
  const parentLocation = faker.helpers.arrayElement(LOCATIONS);

  return new Location({
    ...createRandomNode(),

    name: `Location ${faker.number.int({ max: 100 })}`,
    parentLocationId: parentLocation?.id,
  });
}

for (let index = 0; index < 15; index++) {
  LOCATIONS.push(createRandomLocation());
}
```

That generates a tree structure

jhechtf:
> Personally i wasn't even aware Array.from could take a second mapping parameter :D
> 
> I feel that as its the first example in the docs we should favour code that newbies wont struggle to understand, maybe a good old fashioned for-loop and plain Javascript so its clear the library works with both TS and JS (more fancy TS features could be shown off in later examples).
> 
> e.g.
> 
> ```js
> import { faker } from '@faker-js/faker';
> // import { faker } from '@faker-js/faker/locale/de';
> 
> const users = [];
> 
> function createRandomUser() {
>   return {
>     userId: faker.string.uuid(),
>     username: faker.internet.userName(),
>     email: faker.internet.email(),
>     avatar: faker.image.avatar(),
>     password: faker.internet.password(),
>     birthdate: faker.date.birthdate(),
>     registeredAt: faker.date.past(),
>   };
> }
> for (let i = 0; i < 10; i++) {
>   users.push(createRandomUser());
> }
> console.dir(users)
> ```

@matthewmayer Yeah, the `Array.from` call's second parameter doesn't see a lot of usage because most people do not use it like a generator. 

As for the part about newbies, I have mixed feelings about it. On one hand I remember just learning programming in general and looking at code that was well above my paygrade wondering what any of it said, but at the same time it was how I got introduced to different things (looking at you, first time i ever saw a ternary operator). If someone is using FakerJS, can't we expect _some_ knowledge? If we have to assume zero base knowledge, can't we simply put a note before the example, something like

>**Note:** this example uses `Array.from`'s second parameter, which you can learn more about [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#mapfn)

jhechtf:
@Shinigami92 I agree with you that you _can_ do that, but that was not what the code was actually doing, so I'm not sure why we'd want to take what I thought was supposed to be a quick look into faker and turn it into something that complex. Especially because there are more thorough explanations in the full documentation site.

ST-DDT:
If `Array.from` is not known, how about just using faker.helpers.multiple (maybe with some additional type information for clarity)?

````ts
const users: User[] = faker.helpers.multiple(generateUser, { count: 5 });
````

**Comment by Shinigami92** [comment]
> If `Array.from` is not known, how about just using faker.helpers.multiple (maybe with some additional type information for clarity)?
> 
> ```ts
> const users: User[] = faker.helpers.multiple(generateUser, { count: 5 });
> ```

I think this could be good
Also we didn't had `helpers.multiple` back then

**Comment by jhechtf** [comment]
I mean `Array.from` is known, it's just not many people know about the second parameter of it. Same thing with something like `String.replace` taking a function for the second parameter -- you only learn about it once you need it.

However, the way you did it definitely seems preferable to me as it exposes some helpers out the gate, so I'll update with that and get a new commit up.

**Review by ST-DDT — CHANGES_REQUESTED** [review]
Please revert the unrelated formatting changes.

**Comment by jhechtf** [comment]
That's... weird. I ran the `pnpm format` and that's what came out of it. Let me look a bit more into what happened on that.

**Comment by jhechtf** [comment]
mk so apparently my editor for it's usual setup aggressively formats markdown files, my bad! merged in latest from this repo and reverted anything not related to the code example. Should hopefully be good to go.

**Review by ST-DDT — DISMISSED** [review]

File: README.md [hunk 2]
```diff
@@ -74,9 +72,7 @@ export function createRandomUser(): User {
   };
 }
 
-Array.from({ length: 10 }).forEach(() => {
-  USERS.push(createRandomUser());
-});
+export const const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
```
xDivisionByZerox:
```suggestion
export const USERS: User[] = faker.helpers.multiple(generateRandomUser, { count: 5 });
```

xDivisionByZerox (changes requested):

xDivisionByZerox (dismissed):
> I'll start by saying that this isn't necessary [...]

And yet you still took your time to improve Faker as a project. We appreciate your contribution!

xDivisionByZerox:
Can you please rerun the formatter again locally? Prettier seems to still have a problem.

```bash
pnpm run format
```

jhechtf:
I've re-ran the prettier check but nothing gets modified and it seems fine on my computer?

image

Going to nuke my node_modules and re-run through WSL to see if that changes anything

jhechtf:
that did not appear to change anything either. I'm not sure why the check is failing on CI.

ST-DDT (dismissed):

matthewmayer:
i did get a change when i re-ran prettier. not sure why it wasn't showing up for @jhechtf i went ahead and committed the change so CI is happy

jhechtf:
Thanks Matthew, appreciate it.

ST-DDT (approved):

matthewmayer (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #1855, opened 2023-02-17
feat(number): add parameter `fractionDigits` in float

### LINKED ISSUE #1596: Refactor `number.float` parameter `precision` to `fractionDigits`

xDivisionByZerox:
Fixes #1596.

Since `datatype.float` and `datatype.number` now have significant differences compared to `number.float` I put the legacy implementation for the `datatype` functions directly into that module, without referencing the new implementation of the `NumberModule`.
This allows us to more freely change the new APIs without having a fear of breaking something in the legacy one.


test/number.spec.ts
```diff
@@ -178,14 +178,15 @@ describe('number', () => {
         }
       });
 
-      it('provides numbers with a given precision of 0.5 steps', () => {
+      it.todo('provides numbers with a given precision of 0.5 steps', () => {
```
xDivisionByZerox: I'm a bit lost on how to achieve this with 'fractionDigits' being present. Any math person can help me out?

test/__snapshots__/number.spec.ts.snap
```diff
@@ -26,7 +26,7 @@ exports[`number > 42 > float > with min 1`] = `-25.894775084685534`;
 
 exports[`number > 42 > float > with min and max 1`] = `-0.4260473116301`;
 
-exports[`number > 42 > float > with min, max and precision 1`] = `-0.4261`;
+exports[`number > 42 > float > with min, max and fractionDigits 1`] = `-0.426`;
```
xDivisionByZerox: `fractionDigits` being 4 here but decimal places only being 3 due to the last digit being `0` and thus getting trimmed. Is this desired??

test/number.spec.ts
xDivisionByZerox: Should this feature may be entirely implemented in its own pull request? I would have a pretty hacky solution (locally) to solve this, but I have a feeling it will result in a very big discussion. That's why I'm asking whether `fractionDigits` and `steps` should be 2 separate concerns.

matthewmayer: I don't think "steps" is particularly useful. I would not include it in the new method. 

Assuming we only allow integer values for fraction digits then that's equivalent to precisions of the form 10^x only.

xDivisionByZerox: > Assuming we only allow integer values for fraction digits [...]

I don't think that's the current behavior. There is also no test case for that. What would happen in your opinion if someone passed a float into `fractionDigits`?

matthewmayer: You can't have a number with "two and a half
" decimal places. It should throw an error.

test/__snapshots__/number.spec.ts.snap
matthewmayer: As the return value is a number not a string I think that's unavoidable

test/number.spec.ts
matthewmayer: FWIW toFixed seems to just floor its parameter ie

2.1234567.toFixed(2.99) just acts like 2.1234567.toFixed(2)

src/modules/number/index.ts
```diff
@@ -135,7 +138,7 @@ export class NumberModule {
       };
     }
 
-    const { min = 0, max = 1, precision } = options;
+    const { min = 0, max = 1, fractionDigits = 16 } = options;
```
ST-DDT: IMO we should just return the generated float then instead of converting it to string and then back to number.

test/number.spec.ts
ST-DDT: > I don't think "steps" is particularly useful. I would not include it in the new method.

The `precision` aka `step` feature is intended to be used for or in conjunction with HTML's input step feature:

https://www.w3schools.com/tags/att_input_step.asp

See also: https://github.com/faker-js/faker/issues/1596#issuecomment-1326654346

> I'm a bit lost on how to achieve this with 'fractionDigits' being present.

I think it is not necessary.

> Any math person can help me out?

See also: least common multiple
I assume there also approximation possible using the log function.

> FWIW toFixed seems to just floor its parameter ie

IMO there is no need to go beyond what JS does itself.

src/modules/number/index.ts
matthewmayer: i agree, i would not provide a default and then do something like:

```
if (fractionDigits !== undefined) {
  return parseFloat(real.toFixed(fractionDigits));
} else {
  return real
}
```

Shinigami92:
Adding the new parameter `fractionDigits` should have been an addition, not a replacement

Around 3 month ago, with many discussions and looking deeper into it, we finally came to the conclusion that we want to officially want to support this behavior to achieve "steps"
like 0.4 would result in `0.0, 0.4, 0.8, 1.2, ...`
Using equivalent math with some logarithmic functions resulted in just painfully headaches and are not suitable for persons that are not trained in math

But as https://github.com/faker-js/faker/pull/1855#discussion_r1110977702 pointed to https://github.com/faker-js/faker/issues/1596#issuecomment-1326654346 we want to XOR the parameter, so you will get a `FakerError` when you pass both parameters the same time as they are incompatible

matthewmayer:
Would it be better to make stepped values a seperate method? so faker.number.float takes `{max, min, precision}` and e.g. faker.number.stepped takes `{max, min, step}`. I feel it's trying to stuff too much into one method.

Shinigami92:
> Would it be better to make stepped values a seperate method? so faker.number.float takes `{max, min, precision}` and e.g. faker.number.stepped takes `{max, min, step}`. I feel it's trying to stuff too much into one method.

I mark this for `needs decision` and we can discuss next meeting

src/modules/number/index.ts
xDivisionByZerox: I implemented this without having multiple code pathes/returns.

ST-DDT:
> > Would it be better to make stepped values a seperate method? so faker.number.float takes `{max, min, precision}` and e.g. faker.number.stepped takes `{max, min, step}`. I feel it's trying to stuff too much into one method.
> 
> I mark this for `needs decision` and we can discuss next meeting

Lets continue the discussion in the related issue to give our community the ability to input their suggestions so we can take them into consideration, when deciding this.

https://github.com/faker-js/faker/issues/1596

ST-DDT:
**Team decision**

The proposal to move the `precision`/`step` logic into a different method has been rejected.
We will keep this `precision`/`step` option in the float method's single signature and document that the parameters are mutually exclusive.

ST-DDT:
We will first refactor the method precision -> step.
Then continue with the fractionDigits PR.

matthewmayer:
I'm still somewhat concerned that "step" is poorly defined.

For example what possible values would you expect to be returned for
`{min: 0.6, max: 2.0, step: 0.5}`

Is it 
`{0.6, 1.1, 1.6}`? ie, `min, min+step, min+2*step....`
or `{1.0, 1.5, 2.0}`? ie `step*X for all integer values of X in min-max range`

Shinigami92:
> I'm still somewhat concerned that "step" is poorly defined.
> 
> For example what possible values would you expect to be returned for `{min: 0.6, max: 2.0, step: 0.5}`
> 
> Is it `{0.6, 1.1, 1.6}`? ie, `min, min+step, min+2*step....` or `{1.0, 1.5, 2.0}`? ie `step*X for all integer values of X in min-max range`

**For now** we will go with second, because this is how precision is working right now.
In a later Faker version >=8.1 we can discuss if we want to change something on this behavior or if we want to leave as is. But for v8.0 this will fall out of our milestone plans.

```ts

const result = Array.from({ length: 5 }, () =>
  faker.number.float({ min: 0.6, max: 2.0, precision: 0.5 })
  //                                       ^^^^^^^^^ this will just be renamed
);

console.log(result); // [ 1.5, 1, 2, 2, 1 ]
```

matthewmayer:
Note this is not how the HTML Input element behaves (which you mentioned at https://github.com/faker-js/faker/pull/1855#discussion_r1110977702 was one of the rationales for including it)

`<input type="number" min=0.6 max=2.0 step=0.5>` allows `{0.6, 1.1, 1.6}`

Shinigami92:
> Note this is not how the HTML Input element behaves (which you mentioned at [#1855 (comment)](https://github.com/faker-js/faker/pull/1855#discussion_r1110977702) was one of the rationales for including it)
> 
> `<input type="number" min=0.6 max=2.0 step=0.5>` allows `{0.6, 1.1, 1.6}`

That is good to know, we will bring this into the discussions when discussing this in a team meeting
But please note that this can take some weeks/months because we are currently focusing on v8.0

matthewmayer:
I think it would be better to not rename the parameter or just remove it completely rather than potentially needing another breaking change to its functionality in the future.

Shinigami92:
> I think it would be better to not rename the parameter or just remove it completely rather than potentially needing another breaking change to its functionality in the future.

removing is not an option anymore 🤷
we already agreed from multiple sides that we will NOT remove it
reason: not everyone can handle the math using logarithms to achieve stepping
**period, end of discussion**

renaming it just helps to understand better for what it is considered
we already moved the function into a whole new module (datatype.float -> number.float)
so I don't see a big issue there to directly rename it and just have
```ts
datatype.float({ precision, ... }) => number.float({ steps: precision, ... })
```

matthewmayer:
could it be something like `legacyPrecision` which doesn't imply it works like the step attribute?

Shinigami92:
> could it be something like `legacyPrecision` which doesn't imply it works like the step attribute?

Mh 🤔 I'm not for a `legacy*` prefix, as this is currently used in our code base to indicate parameters that are marked as deprecated.
Do you have alternative names?

matthewmayer:
roundToMultipleOf?

ST-DDT:
`multipleOf` would be fine for me.

matthewmayer:
I think that makes it clearer you can't have arbitrary steps. 

For example you can't use this to return one of

0.25,0.75,1.25,1.75,2.25,2.75

Shinigami92:
Is there at least one test-case with a **combination** of `fractionDigits` + `multipleOf`?

xDivisionByZerox:
> Is there at least one test-case with a **combination** of `fractionDigits` + `multipleOf`?

Now there is. Don't know if it is as expected. We did [decide](https://github.com/faker-js/faker/issues/1596#issuecomment-1326654346) that there should be an error if both parameters are incompatible. What exectly does that mean tho?
Currently I did decide to let `multipleOf` take priority over `fractionDigits`.

src/modules/color/index.ts
```diff
@@ -389,7 +389,7 @@ export class ColorModule extends ModuleBase {
 
     color = Array.from({ length: 3 }, () => this.faker.number.int(255));
     if (includeAlpha) {
-      color.push(this.faker.number.float({ multipleOf: 0.01 }));
+      color.push(this.faker.number.float({ fractionDigits: 2 }));
```
matthewmayer: is there any need to change this? As they do the same thing and neither are deprecated.

src/modules/number/index.ts
```diff
@@ -186,10 +203,21 @@ export class NumberModule extends SimpleModuleBase {
...
+    }
+
     // @ts-expect-error: access private member field
     const randomizer = this.faker._randomizer;
-    const real = randomizer.next();
-    return real * (max - min) + min;
+    let real = randomizer.next() * (max - min) + min;
+
+    if (fractionDigits !== undefined) {
+      real = Number.parseFloat(real.toFixed(fractionDigits));
```
matthewmayer: can we avoid casting to string and back by handling this using the existing precision code and setting precision=10 ** fractionDigits?

src/modules/number/index.ts
```diff
@@ -98,17 +98,23 @@ export class NumberModule extends SimpleModuleBase {
    * This property can be used to limit the result to a specific number of decimal digits.
    * For example `0.01` will round to 2 decimal points.
    * If multipleOf is passed, the upper bound is inclusive.
+   * @param options.fractionDigits The number of digits to appear after the decimal point.
```
matthewmayer: this seems slightly misleading, there won't definitely be fractionDigits after the decimal point (for example if you do {min:0, max:1, fractionDigits:2} then any of 0, 0.2 or 0.23 are possible. Only 0.23 actually has 2 digits after the point.

src/modules/number/index.ts
```diff
@@ -98,17 +98,23 @@ export class NumberModule extends SimpleModuleBase {
...
+   * This parameter is ignored if `multipleOf` is given as well.
```
matthewmayer: i think it would be better if it throws. Providing both options is definitely a programmer error, and it will only happen in newly written code, so better to throw early.

matthewmayer (changes requested):

src/modules/number/index.ts
Shinigami92: On a TypeScript layer we could even provide conditional interfaces. But I fear that right now it is not possible, due to some decisions made in history and the extensive detailed need of JSDoc.
So right now I would also go with @matthewmayer and suggest to throw a runtime error instead.

Maybe even suggest via documentation that you can use `ax+b` style, where `x` is the function call `faker.number.float`.

src/modules/number/index.ts
xDivisionByZerox: Thanks smart. Didn't thought about that.

src/modules/color/index.ts
ST-DDT: I slightly prefer `multipleOf` over `fractionDigits`, but that isn't blocking for me.

src/modules/datatype/index.ts
```diff
@@ -1,5 +1,56 @@
...
+ * The legacy implementation for datatype float.
+ *
+ * @param options An options object.
+ * @param options.min Lower bound for generated number. Defaults to `0`.
+ * @param options.max Upper bound for generated number. Defaults to `min + 99999`.
+ * @param options.faker A faker instance.
+ * @param options.precision Precision of the generated number. Defaults to `0.01`.
+ *
+ */
+function legacyFloatImplementation(options: {
```
ST-DDT: Could you explain why you extracted this method?
Are there some discussions/incompatible changes that I missed?

src/modules/number/index.ts
ST-DDT: Maybe refer to it as max fractionDigits or mention that trailing `0` fraction digits might be hidden during output by JS.

src/modules/datatype/index.ts
xDivisionByZerox: I did write about this in the PR description. Don't know if this is still relevant.

src/modules/color/index.ts
xDivisionByZerox: This would make the current PR less verbose correct? I can apply the change.

src/modules/number/index.ts
```diff
@@ -98,17 +98,23 @@ export class NumberModule extends SimpleModuleBase {
...
+   * Defaults to `16`.
    *
-   * @throws When `min` is greater than `max`.
-   * @throws When `precision` is negative.
+   * @throws If `max` is smaller than `min`.
```
ST-DDT: Int uses the old phrasing here:

https://github.com/faker-js/faker/blob/7a1f5bc50f178c10f6cd335432c0c5f1064688c7/src/modules/number/index.ts#L28

I've also found this variant:

https://github.com/faker-js/faker/blob/1328985f36b9e6670aff7704cb1914fd7a4514d4/src/modules/helpers/index.ts#L379

src/modules/number/index.ts
```diff
@@ -98,17 +98,23 @@ export class NumberModule extends SimpleModuleBase {
    * This property can be used to limit the result to a specific number of decimal digits.
    * For example `0.01` will round to 2 decimal points.
    * If multipleOf is passed, the upper bound is inclusive.
+   * @param options.fractionDigits The number of digits to appear after the decimal point.
+   * This parameter is ignored if `multipleOf` is given as well.
+   * Defaults to `16`.
```
ST-DDT: IMO this should not have a default.

src/modules/number/index.ts
ST-DDT: IMO the user should either use the fractionDigits or the multipleOf. I think we should not include instructions how they should combine the two, as that would go to far into simple math that hopefully everybody can do by themselves. it would also be kind of ugly to read. `faker.number.float({ min: min / multipleOf, max: max / multipleOf, fractionDigits }) * multipleOf and please make sure that multpleOf is an integer otherwise you get more fractionDigits`

src/modules/number/index.ts
ST-DDT: The error message matches the old `@throws` as well.

test/modules/datatype.spec.ts
```diff
@@ -233,12 +233,16 @@ describe('datatype', () => {
...
           expect(() => {
-            faker.datatype.number({ precision: -0.01 });
-          }).toThrow(
-            new FakerError('multipleOf/precision should be greater than 0.')
-          );
+            faker.datatype.number({ min, max, precision: 0 });
+          }).toThrowError(`Max ${max} should be greater than min ${min}.`);
+          expect(() => {
+            faker.datatype.number({ min, max, precision: -1 });
+          }).toThrowError(`Max ${max} should be greater than min ${min}.`);
```
ST-DDT: This test does not match it's description. precision vs min+max

test/modules/number.spec.ts
```diff
@@ -23,10 +23,10 @@ describe('number', () => {
         .it('with min', { min: -42 })
         .it('with max', { max: 69 })
         .it('with min and max', { min: -42, max: 69 })
-        .it('with min, max and precision', {
+        .it('with min, max and fractionDigits', {
           min: -42,
           max: 69,
-          precision: 0.0001,
+          fractionDigits: 4,
         })
```
ST-DDT: IMO this should be a new testcase instead of replacing the old precision one.

test/modules/number.spec.ts
```diff
@@ -281,6 +265,37 @@ describe('number', () => {
...
       });
 
+      it('provides numbers with an exact fractional digits', () => {
+        for (let i = 0; i < 100; i++) {
+          const actual = faker.number.float({
+            min: 0.5,
+            max: 0.99,
+            fractionDigits: 2,
+          });
+          expect(actual).toBe(Number(actual.toFixed(2)));
```
ST-DDT, line 297: Should we use `Number.parseFloat` here instead of this?

src/modules/number/index.ts
matthewmayer: perhaps the *maximum* number of fraction digits to...

test/modules/number.spec.ts
```diff
@@ -233,22 +233,6 @@ describe('number', () => {
...
-                max: 1.5,
-                precision: 0.5,
-              })
-            )
-          ),
-        ].sort();
-
-        expect(results).toEqual([0, 0.5, 1, 1.5]);
-      });
-
```
ST-DDT: IMO we should keep this test till v9.

src/modules/datatype/index.ts
ST-DDT: Me neither, we probably have to check if this breaks/changes anything since the last minor version.

matthewmayer: now we are keeping precision too, probably can get rid of legacyFloatImplementation

test/modules/number.spec.ts
```diff
@@ -317,7 +335,7 @@ describe('number', () => {
         );
       });
 
-      it('throws an error for multipleOf 0', () => {
+      it('throws an error for and0', () => {
```
matthewmayer: looks like accidental change

matthewmayer (changes requested):

src/modules/datatype/index.ts
xDivisionByZerox: I did revert this change.

test/modules/datatype.spec.ts
xDivisionByZerox: Would changing the error message be considered a regression in the first place?

test/modules/number.spec.ts
xDivisionByZerox: I propably did not mean to relace the old one. The last next merge was pretty big. I guess this slipped in there. 
I'll add this case as a separate test case.

test/modules/number.spec.ts
xDivisionByZerox: I previously checked the floating point digits by doing `value.toString().split('.')[1]?.length ?? 0`. Would that be more appropriate here in general?

test/modules/number.spec.ts
xDivisionByZerox: Propably also from the latest next merge. I can see myself skipping over the difference in "multpleOf" and "precision", thus thinking it is the same test. I'll readd it.

src/modules/number/index.ts
xDivisionByZerox: I liked the suggestion from @matthewmayer and did apply it. It was minimal, but included all information required.

src/modules/number/index.ts
xDivisionByZerox: > On a TypeScript layer we could even provide conditional interfaces.

I thought about that as well, but decide against it due to our current limitations regard the documentation auto generation. I would see this as a potential DX improvment in the future tho.

src/modules/number/index.ts
xDivisionByZerox: Removed the default.

src/modules/number/index.ts
xDivisionByZerox: Reverted to the message before and adjusted the other throws tags as well.

src/modules/number/index.ts
```diff
@@ -167,6 +178,24 @@ export class NumberModule extends SimpleModuleBase {
...
+      if (fractionDigits < 0) {
+        throw new FakerError(
+          'The fractional digits count should be greater than 0.'
+        );
+      }
+
+      if (multipleOf === undefined) {
+        multipleOf = 10 ** -fractionDigits;
+      }
+    }
```
xDivisionByZerox: I don't really like this extra if statement here. Would you be down for a strategy pattern (fractionDigit, multipleOf, default) here? That would touch more existing code, tho.

test/modules/datatype.spec.ts
ST-DDT: That is not what I meant:

> ``it('should throw when precision <= 0'``

vs

> ``.toThrowError(`Max ${max} should be greater than min ${min}.`);``

test/modules/number.spec.ts
ST-DDT: No, my concern here is more about consistently using `Number.parseFloat(value)` vs `Number(value)`.

ST-DDT: Well nevermind. We mostly use `Number(value)` in our code.

src/modules/number/index.ts
ST-DDT: You might be able to simplify this if you decontruct the options to 
````ts
const {
  ...
  fractionDigits,
  precision,
  orginalMulipleOf: multipleOf = precision,
  multipleOf = precision ?? fractionDigits == null ? undefined : 10 ** -fractionDigits,
} = options;
````

and then checking only

````ts
if (typeof fractionDigits === 'number' && typeof orginalMulipleOf === 'number') {
  throw new FakerError(
    'multipleOf and fractionDigits cannot exist at the same time.'
  );
}
````

Comment (xDivisionByZerox) [reply src/modules/number/index.ts]: But then you are missing the specific validation for `fractionDigit`. `multipleOf` not allowed to be less than 0, so I expect the same for `fractionDigits`.
Furthermore, currently there is no check that `fractionDigits` is an integer. Should there be validation for that?

File: test/modules/datatype.spec.ts (hunk 11 above)
Comment (xDivisionByZerox) [reply test/modules/datatype.spec.ts]: I see what you mean. 

My question remains: These tests indecate that the behaviour of `datatype.number` (and `datatype.float`) is altering. That would be considered breaking, no? So in theory, there should be no changes in this test files. So in optimal way I reset changes here and fix the current implementation to work as expected by our test. Is that correct?

File: src/modules/number/index.ts (hunk 16 above)
Comment (ST-DDT) [reply src/modules/number/index.ts]: I didn't mean to imply the removal of the other verifications.

Comment (matthewmayer) [reply src/modules/number/index.ts]: Yes we should definitely check fractionDigits is an integer

File: src/modules/number/index.ts [hunk 17]
```diff
@@ -167,6 +185,25 @@ export class NumberModule extends SimpleModuleBase {
...
+    if (
+      typeof fractionDigits === 'number' &&
+      typeof originalMultipleOf === 'number'
+    ) {
+      throw new FakerError(
+        'multipleOf and fractionDigits cannot exist at the same time.'
+      );
+    }
+
+    if (fractionDigits !== undefined) {
```
ST-DDT: This and the check in line 189 is indentical. Maybe merge both branches?

````ts
if (fractionDigits !== undefined) {

  if (typeof originalMultipleOf === 'number') { // or potentially `originalMultipleOf  != null` as a `string` is equally bad
     [...]
  }

[...]
````

File: src/modules/number/index.ts [hunk 18]
```diff
@@ -167,6 +185,25 @@ export class NumberModule extends SimpleModuleBase {
...
+      if (fractionDigits % 1 !== 0) {
```
ST-DDT: Maybe use `Number.isInteger(fractionDigits)` instead?

test/modules/datatype.spec.ts
ST-DDT: AFAICT the current tests for datatype (although slightly changed) test for the same things as before and thus there are currently no breaking changes here. Or am I missing something?

src/modules/number/index.ts
```diff
@@ -167,6 +185,22 @@ export class NumberModule extends SimpleModuleBase {
       throw new FakerError(`Max ${max} should be greater than min ${min}.`);
     }
 
+    if (fractionDigits !== undefined) {
+      if (typeof originalMultipleOf === 'number') {
```
ST-DDT: Should we use `!== undefined` here as well?

Also I just remembered that we mostly do `value != null` checks instead of `!== undefined`. Should we change the checks in this method to be consistent with the rest?

`== null`: 24
`=== undefined`:  8
`!= null`: 47
`!== undefined`: 3 (two of them in float)
(data from next)

src/modules/number/index.ts
```diff
@@ -167,6 +185,22 @@ export class NumberModule extends SimpleModuleBase {
...
+        throw new FakerError(
+          'multipleOf and fractionDigits cannot exist at the same time.'
```
```suggestion
          'multipleOf and fractionDigits cannot be set at the same time.'
```

src/modules/number/index.ts
```diff
@@ -167,6 +185,22 @@ export class NumberModule extends SimpleModuleBase {
...
+        );
+      }
+
+      if (!Number.isInteger(fractionDigits)) {
+        throw new FakerError('fractionDigits should be an integer.');
+      }
+
+      if (fractionDigits < 0) {
+        throw new FakerError('fractionDigits should be greater than 0.');
```
```suggestion
        throw new FakerError('fractionDigits should be greater than or equal to 0.');
```

src/modules/number/index.ts
Shinigami92: I didn't wanted to be picky, but thx that someone else mentioned this 💚

xDivisionByZerox: > I didn't wanted to be picky, but thx that someone else mentioned this 💚

It's fine. In most personall projects I handle `null` and `undefined` explicit. Thats a habit that will slip in every now and then. Sorry about that^^

ST-DDT:
Could please fix the formatting and other CI issues?

xDivisionByZerox:
> Could please fix the formatting and other CI issues?

I'm not at home during the weekend. Feel free to change the PR.

matthewmayer (dismissed):

ST-DDT (dismissed):

matthewmayer (dismissed):

src/modules/number/index.ts
```diff
@@ -97,18 +97,25 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.multipleOf The generated number will be a multiple of this property.
    * This property can be used to limit the result to a specific number of decimal digits.
    * For example `0.01` will round to 2 decimal points.
-   * If multipleOf is passed, the upper bound is inclusive.
+   * If multipleOf is passed, the upper bound is inclusive and the `fractionDigits` option has to be excluded.
```
Shinigami92:
```suggestion
   * If `multipleOf` is passed, the upper bound is inclusive and the `fractionDigits` option has to be excluded.
```

Shinigami92: Maybe better:

```diff
- and the `fractionDigits` option has to be excluded.
+ and the `fractionDigits` option need to be omitted.
```

src/modules/number/index.ts
```diff
@@ -97,18 +97,25 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @param options.fractionDigits The maximum number of digits to appear after the decimal point.
+   * This parameter has to be excluded if `multipleOf` is provided.
```
Shinigami92:
```suggestion
   * This parameter has to be omitted if `multipleOf` is provided.
```

test/modules/datatype.spec.ts
```diff
@@ -316,22 +316,24 @@ describe('datatype', () => {
...
+          const max = 1;
 
           expect(() => {
-            faker.datatype.number({ min, max });
-          }).toThrow(
-            new FakerError(`Max ${max} should be greater than min ${min}.`)
-          );
+            faker.datatype.float({ min, max });
+          }).toThrowError(`Max ${max} should be greater than min ${min}.`);
         });
```
Shinigami92: Why was this test changed?
1. Why the created git-blame for the min/max?
2. Why not checking anymore that a `FakerError` was returned? But just any kind of `Error`?

test/modules/datatype.spec.ts
```diff
@@ -316,22 +316,24 @@ describe('datatype', () => {
...
-        it('should throw when precision is negative', () => {
+        it('should throw when precision <= 0', () => {
+          const min = 1;
+          const max = 2;
+
           expect(() => {
-            faker.datatype.float({ precision: -0.01 });
-          }).toThrow(
-            new FakerError('multipleOf/precision should be greater than 0.')
-          );
```
Shinigami92: Is https://github.com/faker-js/faker/pull/1855/files#diff-7a79dc63c7e78ab49e02d1bb360d4e99f509bfc534e77f659ac3f9895564d6fcR209 not reachable anymore? Or why was this test removed?

test/modules/datatype.spec.ts
```diff
@@ -316,22 +316,24 @@ describe('datatype', () => {
...
+            faker.datatype.float({ min, max, precision: 0 });
+          }).toThrowError('multipleOf/precision should be greater than 0.');
+          expect(() => {
+            faker.datatype.float({ min, max, precision: -1 });
+          }).toThrowError('multipleOf/precision should be greater than 0.');
```
Shinigami92: Maybe check explicitly for `FakerError`s

test/modules/number.spec.ts
```diff
@@ -281,6 +286,44 @@ describe('number', () => {
         expect(results).toEqual([0, 0.4, 0.8, 1.2, 1.6]);
       });
 
+      it('provides numbers with an exact fractional digits', () => {
+        for (let i = 0; i < 100; i++) {
```
Shinigami92: Instead of a for-loop, we could use a it.each here :thinking:  
_Is `eslint-plugin-vitest` still active? Why did it not report that?_

test/modules/datatype.spec.ts
ST-DDT: The test has not been removed only changed (`<` -> `<=`).

test/modules/number.spec.ts
ST-DDT: https://github.com/faker-js/faker/blob/d71bc47a4a7b127b41068016cbed3584a35b47ea/.eslintrc.js#L193

test/modules/datatype.spec.ts
```diff
@@ -316,19 +316,27 @@ describe('datatype', () => {
         });
 
         it('should throw when min > max', () => {
-          const min = 10;
-          const max = 9;
+          const min = 2;
+          const max = 1;
```
ST-DDT:
```suggestion
          const min = 10;
          const max = 9;
```

test/modules/datatype.spec.ts
ST-DDT: number -> float is a typo.
I reverted the min/max change and fixed the FakerError

ST-DDT (approved):

matthewmayer (approved):
i think the documentation to the whole method could still use an overall rewrite, but as its not really specific to this parameter change i'll do that as a seperate PR after this is merged.

Shinigami92 (approved):

PR #1857, opened 2023-02-19
test(number): replace RegExp with validation functions

xDivisionByZerox:
## Description

This PR enhances the type validation in tests for the `NumberModule`. 

---

## Reasoning

> [...] The tests will usually use regex or preferably [validator.js](https://github.com/validatorjs/validator.js) to ensure the method returns valid results. [...]

This statement is from the projects CONTRIBUTING.md. The word of interest here is **preferably**. This is because RexExp can get out of hand really quickly and does not always result in easy-to-understand error messages since they only provide the RegExp and no information on where the error exactly happened.
This problem persist with Validation functions. But, they result in the error message providing the function's name which allows for a quicker evaluation of the actual problem if it is descriptive enough.


test/number.spec.ts
```diff
@@ -48,16 +48,32 @@ describe('number', () => {
...
+
+        expect(actual).toBeGreaterThanOrEqual(0);
+        expect(actual).lessThanOrEqual(Number.MAX_SAFE_INTEGER);
+      });
+
       it('should return a random number given a maximum value as Number', () => {
         const actual = faker.number.int(10);
 
+        expect(actual).toBeTypeOf('number');
+        expect(actual).toSatisfy(Number.isInteger);
```
xDivisionByZerox, line 61: I extended the tests in a way that each case individually checks the return type of a function. I did this because potentially the function could be implemented in a way that returns different types based on the input arguments provided (excessive examples):
- int() => number
- int(number) => still number but potentially float
- int({ min, max }) => suddenly string
- int({ format }) => BigInt

I don't know if this is me overthinking the problem or actually relevant. I would like some feedback on this.

test/number.spec.ts
```diff
@@ -250,23 +300,35 @@ describe('number', () => {
     });
 
     describe('binary', () => {
-      it('generates single binary character when no additional argument was provided', () => {
+      function isBinary(str: string) {
+        return [...str].every((char) => char === '0' || char === '1');
+      }
```
Shinigami92, line 301: We could move something like this into an extension of vitest (https://github.com/faker-js/faker/blob/next/test/vitest-extensions.ts)

test/number.spec.ts
```diff
@@ -142,20 +168,36 @@ describe('number', () => {
     });
 
     describe('float', () => {
-      it('should return a random float', () => {
+      function isFloat(value: number) {
+        return value % 1 !== 0;
+      }
```
Shinigami92, line 169: This feels not safe
When the planets are all aligned in the solar system, this could get theoretically be e.g. 0.0 or 1.0

ST-DDT: After reading the other coditions, I agree with your concern. The test checks and fails for conditions that later checks explicitly allow e.g. `float >= 0`.

test/number.spec.ts
xDivisionByZerox: All those function are only used in this specific use case. Extracting them brings to advantage other than cluttering the vitest API IMO.

test/number.spec.ts
xDivisionByZerox: Yes, that is right but how exactly is `0.0` or `1.0` displayed in JS? Correct the `.0` part is striped and you receive an integer (technically not since all Numbers are floats in JS). This is by design the function should fail for these cases. Tests should provide cases where this can not happen.

I ran the current test 1 Billion times and did not get an error BTW.

ST-DDT (approved):

ejcheng (approved):

PR #1860, opened 2023-02-20
refactor(company)!: rename bs to buzz

Shinigami92:
references #1346


Shinigami92:
https://github.com/faker-js/faker/issues/1346#issuecomment-1407720680

> **Team Decision: Company**
> 
>> `bs` should probably be deprecated
> 
> Yes!

Did we had a reason for just removing it?
What should I write in the `proposed` of `deprecated`?
Or do we just want to rename it to e.g. `buzzWord`?

>> `bsAdjective` should probably be merged into `catchPhraseAdjective` and deprecate
>> `bsBuzz` should probably be called `buzzWord` (is that correctly written?)
>> `bsNoun` should probably be merged into `catchPhraseNoun` and deprecated
> 
> Replace `bs*` with `buzz*` and improve JSDocs (buzz = manager language, catchPhrase = user language).

Open TODO: 
- [x] improve JSDocs (buzz = manager language, catchPhrase = user language)

xDivisionByZerox:
> Did we had a reason for just removing it?
What should I write in the proposed of deprecated?
Or do we just want to rename it to e.g. buzzWord?

It was more important to get rid of the `bs` function as it can be considered unprofessional. Reading it's implementation `buzzWord`makes sense.

src/definitions/company.ts
```diff
@@ -7,17 +7,17 @@ export type CompanyDefinitions = LocaleEntry<{
   /**
    * Business/products related adjectives.
```
ST-DDT: Please also extend the jsdocs here.

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #1872, opened 2023-02-24
feat(locale): add hyphenated names to English locales

### LINKED ISSUE #691: feat(name): add hyphenated patterns to `lastName()`

matthewmayer:
fix #691
- default en locale gets 95% non-hyphenated 5% hyphenated
- other locales without existing last_name_patterns get 100% non-hyphenated
- locales with split male_last_name and  female_last_names get split patterns
- existing locales with last_name_patterns (e.g. lv, en_GB) are not changed.

If additional information about hyphenation rules in other locales can be found, this can be added in a subsequent PR. I kept this one formulaic to make it easy to review.


matthewmayer:
this exposed a bug in the test for commitEntry, the local part of the email addresses was not allowing hyphens. i changed the regex to match the one we use in the emailAddress method https://github.com/faker-js/faker/blob/add2254ca31434af3003c0051064bdb27713992d/src/modules/internet/index.ts

Shinigami92:
> this exposed a bug in the test for commitEntry, the local part of the email addresses was not allowing hyphens. i changed the regex to match the one we use in the emailAddress method [`add2254`/src/modules/internet/index.ts](https://github.com/faker-js/faker/blob/add2254ca31434af3003c0051064bdb27713992d/src/modules/internet/index.ts)

Please lets extract this to its own PR, so
1. it is reviewable in parallel to this PR
2. it gets its own "fix/test" changelog entry

matthewmayer:
I considered that, but the test doesn't fail on the next branch because email addresses in en never contain hyphens.

Shinigami92:
> I considered that, but the test doesn't fail on the next branch because email addresses in en never contain hyphens.

You are free to add static tests that handles special cases in the random seeded test cases (second part of each test file)

Maybe we need to change something here instead of the test
https://github.com/faker-js/faker/blob/11a5f51ddb4ab1163d770c4b12f5fe4711b8e11d/src/modules/git/index.ts#L123

We are already starting to pollute this PR with unrelated discussions, please lets stop this and extract it so we can have dedicated discussions

matthewmayer:
> > I considered that, but the test doesn't fail on the next branch because email addresses in en never contain hyphens.
> 
> You are free to add static tests that handles special cases in the random seeded test cases (second part of each test file)
> 
> Maybe we need to change something here instead of the test
> 
> https://github.com/faker-js/faker/blob/11a5f51ddb4ab1163d770c4b12f5fe4711b8e11d/src/modules/git/index.ts#L123
> 
> We are already starting to pollute this PR with unrelated discussions, please lets stop this and extract it so we can have dedicated discussions

#1876

src/locales/lv/person/female_last_name_patterns.ts
```diff
@@ -1,7 +1 @@
-export default [
-  { value: '{{person.female_last_name}}', weight: 8 },
-  {
-    value: '{{person.female_last_name}}-{{person.female_last_name}}',
-    weight: 2,
-  },
-];
```
ST-DDT: Was this change intentional?

src/locales/lv/person/male_last_name_patterns.ts
```diff
@@ -1,4 +1 @@
-export default [
-  { value: '{{person.male_last_name}}', weight: 8 },
-  { value: '{{person.male_last_name}}-{{person.male_last_name}}', weight: 2 },
-];
```
ST-DDT: Same here

ST-DDT (commented):
These lines are now somewhat mandatory:

https://github.com/faker-js/faker/blob/11a5f51ddb4ab1163d770c4b12f5fe4711b8e11d/src/modules/person/index.ts#L120-L135

So IMO the other code branch could be removed.
Also the PR does more than the PR title suggests IMO.

Not sure whether it is worth splitting into two PRs:
1. Make last_name_patterns required/Generate last_name_patterns for all locales
2. add hyphenated names to English locales

matthewmayer:
It's not mandatory if you create a new locale which doesn't extend en.

src/locales/lv/person/male_last_name_patterns.ts
matthewmayer: reverted

src/locales/lv/person/female_last_name_patterns.ts
matthewmayer: reverted

ST-DDT:
Please fix the merge conflicts.

ST-DDT:
Please adjust the PR title to reflect the current. It also changes non English locales.

matthewmayer:
> Please adjust the PR title to reflect the current. It also changes non English locales.

it doesn't add any hyphenated patterns to any non-English locales, the changes are to prevent hyphenated patterns from propagating to the other non-English locales unexpectedly.

ST-DDT (approved):

matthewmayer:
could append ", add last_name_patterns to all locales" maybe? but makes it a bit long

Shinigami92 (approved):

PR #1873, opened 2023-02-24
refactor: restore en_IND

### LINKED ISSUE #1454: en_IND should be an alias for en_IN in v8
### Pre-Checks

### Describe the bug

https://github.com/faker-js/faker/issues/1354 was fixed in https://github.com/faker-js/faker/pull/1448 where the en_IND locale was renamed to en_IN

However to avoid an unnecessary breaking change, in v8 en_IND should be kept as an alias for en_IN, emitting a warning if used, and only completely removed in v9

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 12.6
    CPU: (8) x64 Apple M1
    Memory: 43.42 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.20.0 - /usr/local/bin/node
    npm: 8.19.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Firefox: 91.0.2
    Safari: 16.0
```

- [X] CJS
- [ ] ESM

### Used Package Manager

npm

Shinigami92:
closes #1454


Shinigami92:
@ST-DDT Do I also need to alias the locales definition and therefore add `'en_IND'` (the string locale **key**)
This might be unnecessary because of
- #1735

ST-DDT:
I'm not sure what you are intent to say. Let's wait for the PR to get merged.

ST-DDT:
#1735 is merged, so this is no longer blocked.

matthewmayer:
As part of this upgrading.html can be updated to reflect this is no longer breaking (remove or move to second section).

Shinigami92:
I will update this PR after

- #1748

was merged

ST-DDT (dismissed):

matthewmayer (changes requested):
upgrading.html can be updated to reflect this is no longer breaking (remove or move to second section).

matthewmayer (approved):

matthewmayer (dismissed):
Doesn't seem to expose the fakerEN_IND object?
```
const {fakerEN_IND, fakerEN_IN} = require("@faker-js/faker")

fakerEN_IN.person.fullName()
>> 'Udit Patel'

fakerEN_IND.person.fullName()
>> Uncaught TypeError: Cannot read property 'person' of undefined
```

Shinigami92:
@ST-DDT what do I need to do here to handle this edge case?

ST-DDT:
Was it ever exposed like that?

ST-DDT:
IMO the sole thing that should work is the original import. We shouldnt add new imports just to deprecate them immediately.

Shinigami92:
> Was it ever exposed like that?

It was only exposed as `@faker-js/faker/locale/en_IND`

So yes you are right
```ts
import { faker } from '@faker-js/faker/locale/en_IND';

console.log(faker.person.firstName());
```
works fine

matthewmayer:
Hmm it kinda feels like a double breaking change if you previously had

```
faker.locale = "en_IND"
faker.foo.bar()
```

You have to figure out to change that to

```
fakerEN_IN.foo.bar()

```

ST-DDT:
> Hmm it kinda feels like a double breaking change if you previously had
> 
> ```
> faker.locale = "en_IND"
> faker.foo.bar()
> ```
> 
> You have to figure out to change that to
> 
> ```
> fakerEN_IN.foo.bar()
> ```

Sure, but in order to know that migration step, you have to go to the migration guide to lookup the new naming convention anyway.
If you don't find the expected import you will probably search for EN_IND in that document and find the information that it has been renamed (or just check the list of available imports via auto completion).
As a compromise, we could add a hint to the migration guide to clarify the new locale/faker import names.

````txt
You can use our pre-built faker instances using the `fakerEN_IN` import.
````

ST-DDT (approved):

matthewmayer:
Sounds good.

xDivisionByZerox (approved):

PR #1874, opened 2023-02-24
refactor(location)!: zip code state

Shinigami92:
references https://github.com/faker-js/faker/issues/1346#issuecomment-1407728227


Shinigami92:
@faker-js/maintainers @faker-js/members Is the current impl good enough? Or should we still throw a `FakerError` for missing `state` and make therefore a breaking change for the deprecated `zipCodeForState`?

matthewmayer:
I think in general breaking changes happening at the same time as deprecating a method is a bad idea. The point of deprecation is "you have one major version where your existing code will continue to work without changes".

Shinigami92:
> I think in general breaking changes happening at the same time as deprecating a method is a bad idea. The point of deprecation is "you have one major version where your existing code will continue to work without changes".

Beside that: I'm asking because I don't see the need to throw a FakerError instead of doing a fallback 🤷
It just needs to be documented so if someone asks, we can point to the documentation 🙂

src/modules/location/index.ts
```diff
@@ -50,6 +62,16 @@ export class LocationModule {
...
     }
 
+    const { state } = options;
+
+    if (state) {
+      const zipRange =
+        this.faker.definitions.location.postcode_by_state?.[state];
+      if (zipRange) {
+        return String(this.faker.number.int(zipRange));
+      }
```
ST-DDT, line 81:
````ts
else throw new FakerError
````

File: src/modules/location/index.ts [hunk 2]
```diff
@@ -88,18 +112,20 @@ export class LocationModule {
...
 
     const { state } = options;
 
-    const zipRange = this.faker.definitions.location.postcode_by_state?.[state];
-    if (zipRange) {
-      return String(this.faker.number.int(zipRange));
-    }
-
-    return this.zipCode();
+    return this.zipCode({ state });
```
ST-DDT, line 140: IMO there is no need to repackage the options, just pass them along as is.

```suggestion
    return this.zipCode(options);
```

src/modules/location/index.ts
```diff
@@ -38,6 +42,14 @@ export class LocationModule {
     options:
       | string
       | {
+          /**
+           * The state to generate the zip code for.
+           *
+           * Only works for locales with postcode_by_state definition. If a locale does not
+           * have a postcode_by_state definition, a random zip code is generated according
+           * to the locale's zip format.
```
ST-DDT: IMO this should throw an error instead.

src/modules/location/index.ts
ST-DDT: We can do that in a separate PR if you want, but if I pass something, then I usually want to actually use it.

Should I reopen: https://github.com/faker-js/faker/issues/1736#issuecomment-1442176970 ?


> **Team Decision**
> [...]
> - It will still throw an error if the state doesn't have a config.

src/modules/location/index.ts
```diff
@@ -50,6 +64,19 @@ export class LocationModule {
...
+
+    if (state) {
+      const zipRange =
+        this.faker.definitions.location.postcode_by_state?.[state];
+
+      if (zipRange) {
+        return String(this.faker.number.int(zipRange));
+      }
+
+      throw new FakerError(`No zip code range found for state "${state}"`);
```
Shinigami92: @ST-DDT Please use GH suggestion to adjust this message to your needs/requirements

test/location.spec.ts
```diff
@@ -113,7 +113,9 @@ describe('location', () => {
     t.describe('zipCode', (t) => {
       t.it('noArgs')
         .it('with string', '###')
-        .it('with format option', { format: '###-###' });
+        .it('with format option', { format: '###-###' })
+        .it('with state option', { state: 'CA' })
+        .it('with options', { state: 'CA', format: '###-###' });
```
Shinigami92: @ST-DDT how do I set the locale for these snapshot tests?

src/modules/location/index.ts
```diff
@@ -38,9 +43,18 @@ export class LocationModule {
     options:
       | string
       | {
+          /**
+           * The state to generate the zip code for.
+           *
+           * Only works for locales with postcode_by_state definition. If a locale does not
+           * have a postcode_by_state definition an error is thrown.
```
ST-DDT: IMO this is redundant:

```suggestion
           * If the currrent locale does not have a corresponding `postcode_by_state` definition, an error is thrown.
```

src/modules/location/index.ts
ST-DDT:
```suggestion
      throw new FakerError(`No zip code definition found for state "${state}"`);
```

test/location.spec.ts
ST-DDT: That is currently unsupported, and I'm not sure how to change that.

test/__snapshots__/location.spec.ts.snap
```diff
@@ -467,9 +455,3 @@ exports[`location > 1337 > zipCode > with format option 1`] = `"251-225"`;
 exports[`location > 1337 > zipCode > with string 1`] = `"251"`;
 
 exports[`location > 1337 > zipCodeByState > noArgs 1`] = `"51225"`;
-
-exports[`location > 1337 > zipCodeByState > with state options 1`] = `"51225"`;
-
-exports[`location > 1337 > zipCodeByState > with string 1 1`] = `"51225"`;
-
-exports[`location > 1337 > zipCodeByState > with string 2 1`] = `"51225"`;
```
ST-DDT, line 475: What happened to these snapshots?

ST-DDT (changes requested):

test/location.spec.ts
Shinigami92: So should I just remove these and ignore snapshot tests for these options?

test/__snapshots__/location.spec.ts.snap
Shinigami92: I have not looked into them yet, because the tests are failing at all
For me the end result of this (currently in draft) PR is relevant

src/modules/location/index.ts
Shinigami92: How important is this change to you and do you have a reason for passing the whole options instead only what is needed?

ST-DDT: Because it is the same, and even if not, it doesn't matter either.
It is not important to me though.

test/location.spec.ts
ST-DDT: I would comment them out and add a comment why, but yes.

src/modules/location/index.ts
```diff
@@ -23,8 +24,12 @@ export class LocationModule {
    * the locale's zip format is used.
    *
    * @param options The format used to generate the the zip code or an options object. Defaults to `{}`.
+   * @param options.state The state to generate the zip code for.
+   * Only works for locales with postcode_by_state definition. If a locale does not
+   * have a postcode_by_state definition an error is thrown.
```
```suggestion
   * If the currrent locale does not have a corresponding `postcode_by_state` definition, an error is thrown.
```

src/modules/location/index.ts
```diff
@@ -64,8 +90,7 @@ export class LocationModule {
    * Generates random zip code from state abbreviation.
    *
    * Only works for locales with postcode_by_state definition. If a locale does not
```
```suggestion
   * If the currrent locale does not have a corresponding `postcode_by_state` definition, an error is thrown.
```

src/modules/location/index.ts
```diff
@@ -64,8 +90,7 @@ export class LocationModule {
...
-   * have a postcode_by_state definition, a random zip code is generated according
-   * to the locale's zip format.
+   * have a postcode_by_state definition an error is thrown.
```
```suggestion
```

src/modules/location/index.ts
```diff
@@ -76,6 +101,8 @@ export class LocationModule {
    * fakerEN_US.location.zipCodeByState("??") // '47683-9880'
    *
    * @since 8.0.0
+   *
+   * @deprecated Use `faker.location.zipCode({ state })` instead.
```
ST-DDT, line 114: Please add an `@see` to `faker.location.zipCode()` (not sure in which order it should be though).

ST-DDT (changes requested):

test/location.spec.ts
Shinigami92: Should I open an issue for tackling this?

test/__snapshots__/location.spec.ts.snap
Shinigami92: Oh, because the implementation now changed (it uses `zipCode()` which now throws an error) so these were removed by vitest because they do not process a result anymore but an error

Shinigami92:
There is an issue that can be better identified by merging

- https://github.com/faker-js/faker/pull/1969

first

ST-DDT (dismissed):

src/modules/location/index.ts
```diff
@@ -71,19 +96,22 @@ export class LocationModule {
   /**
    * Generates random zip code from state abbreviation.
    *
-   * Only works for locales with postcode_by_state definition. If a locale does not
-   * have a postcode_by_state definition, a random zip code is generated according
-   * to the locale's zip format.
+   * If the current locale does not have a corresponding `postcode_by_state` definition, an error is thrown.
```
matthewmayer, line 99: Should this be documented in upgrading.html as a breaking change?

Shinigami92: pls check again

ST-DDT (approved):

test/location.spec.ts
```diff
@@ -156,6 +160,22 @@ describe('location', () => {
 
           expect(zipCode).toMatch(/^[A-Za-z]\d[A-Za-z]\s?\d[A-Za-z]\d$/);
         });
+
+        it.each([
+          ['IL', 60001, 62999],
+          ['GA', 30001, 31999],
+          ['WA', 98001, 99403],
+        ])('returns zipCode valid for state %s', (state, lower, upper) => {
+          const zipCode1 = +fakerEN_US.location.zipCode({ state });
```
xDivisionByZerox, line 169: Maybe `actual`, `result` or `value`? ^^

xDivisionByZerox (approved):
My comment are non blocking. So feel free to resolve them if you don't think they are important.

Shinigami92:
To proceed with this, I will merge it for now
We can extract issues or do several PRs like unifying the test variable names

PR #1875, opened 2023-02-24
test(helpers): mustache special cases

Shinigami92:
fixes https://github.com/faker-js/faker/actions/runs/4251274118/jobs/7393407806#step:7:250


ST-DDT (dismissed):

test/helpers.spec.ts
```diff
@@ -648,11 +648,26 @@ describe('helpers', () => {
         });
 
         it('supports function replace values faker values', () => {
+          const value = faker.string.sample(2);
+
+          const actual = faker.helpers.mustache('1{{value}}3', {
+            value,
+          });
+
+          const expectedLength = { '$&': 11, "$'": 3 }[value] ?? 4;
```
xDivisionByZerox: I don't really like this as it does some "magic" which is covered in a different test case.
Maybe convert `value` to a static string here. That way the expected value could be more expressive as well.

matthewmayer: I admit I am pretty clueless reading this what it's actually supposed to be testing and why those two arbitrary strings are chosen!

Shinigami92: @ST-DDT what is your opinion to this suggestion? (using static value for `value`)

ST-DDT: Since the test basically uses a fixed value for mustache, it is fine to just set it statically.
Another alternative could be using a different generator that is unable to produce active regex components.

test/helpers.spec.ts
```diff
@@ -648,11 +648,26 @@ describe('helpers', () => {
...
+
+          expect(actual).toHaveLength(expectedLength);
+        });
+
+        it.each([
+          ['$&', 11],
+          ["$'", 3],
```
ST-DDT: Maybe add a comment what each of them actually does, so it is easier to understand the test result.

$' escaped ' character

Shinigami92: > $' escaped ' character

Do you know why it does this? I have absolutely no clue about what I did here and just randomly tried thousands of random values with a for loop and just logged the bad ones :shrug: 
So I have no idea what are behind these values

ST-DDT: I found also these:

> $ 1$3
> & 1{{value}}3
> ' 133
> ` 113

Shinigami92: Maybe lets talk later in a voice-call about that, because that confuses me now a bit as these are all not 2-length input values 🤔

ST-DDT: Sure, maybe we should just escape $ ourselves so it ends up there as is was passed to the method.

xDivisionByZerox: **Team decision**

We escape the `$`-sign ourself to prevent unwanted edges from happening.

src/modules/helpers/index.ts
```diff
@@ -399,9 +399,11 @@ export class HelpersModule {
     }
 
     for (const p in data) {
-      const re = new RegExp(`{{${p}}}`, 'g');
-      const value = data[p];
+      const re = new RegExp(`\{\{${p}\}\}`, 'g');
+      let value = data[p];
       if (typeof value === 'string') {
+        // https://stackoverflow.com/a/6969486/6897682
```
ST-DDT: Maybe add a comment what it does?
E.g. escape $.

src/modules/helpers/index.ts
```diff
@@ -399,9 +399,11 @@ export class HelpersModule {
     }
 
     for (const p in data) {
-      const re = new RegExp(`{{${p}}}`, 'g');
-      const value = data[p];
+      const re = new RegExp(`\{\{${p}\}\}`, 'g');
```
ST-DDT: Why did you escape the {} now? It wasnt necessary before.

```suggestion
      const re = new RegExp(`{{${p}}}`, 'g');
```

src/modules/helpers/index.ts
```suggestion
        // escape $, source: https://stackoverflow.com/a/6969486/6897682
```

ST-DDT (approved):

ejcheng (approved):

PR #1878, opened 2023-02-26
fix: correct types field

Shinigami92:
fixing: https://publint.dev/@faker-js/faker@7.6.0

There is no `index.d.ts`
Everything is packaged into the dist folder
This might only affect users using TS + an old npm version that doesn't know the `exports` field

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "MIT",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
-  "types": "index.d.ts",
+  "types": "dist/types/index.d.ts",
   "typesVersions": {
     ">=4.0": {
       "*": [


ST-DDT:
I found this commit, that "reverted" your change:
https://github.com/faker-js/faker/commit/c704e36f692e69bbe23bdf05f4fdcccba4ce8b0f
Any idea why? Is it fixed?

Shinigami92:
> I found this commit, that "reverted" your change: [c704e36](https://github.com/faker-js/faker/commit/c704e36f692e69bbe23bdf05f4fdcccba4ce8b0f) Any idea why? Is it fixed?

I don't remember everything 9 month ago, but see my comment https://github.com/faker-js/faker/pull/979#pullrequestreview-980943370
It looks like only parts of the PRs were correct, but other parts are _wrongly_ reverted

At Vue Amsterdam 2023 I learned a bit more about `package.json` and how to manage it + using https://publint.dev/
So I'm not sure if my knowledge is comparable anymore with my knowledge back then 9 month ago
Thanks a lot to @bluwy

bluwy:
Actually I think before, this _might_ work too as you had the `typesVersion` config that remaps the type definition, but going forward with this PR should still be safe too (and probably less confusing in the future). I find https://github.com/andrewbranch/example-subpath-exports-ts-compat/tree/main/examples/node_modules/types-versions-wildcards to be a great example of understanding how it works.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #1883, opened 2023-02-28
fix(internet): filter banned dots from email addresses

### LINKED ISSUE #1882: email address local part shouldnt start or end with .
Inspired by https://github.com/faker-js/faker/discussions/1881

`faker.internet.email("John*","Doe.") //John_Doe.@yahoo.com`

The code is correctly stripping special chars like * but allowing .

BUT email address local parts are not allowed to start or end with a dot or contain multiple consecutive dots

matthewmayer:
fix #1882


src/modules/internet/index.ts
```diff
@@ -109,6 +109,13 @@ export class InternetModule {
       );
     }
 
+    //local parts may not contain two or more consecutive . characters
+    localPart = localPart.replace(/\.{2,}/g, '.');
+
+    //local parts may not start with or end with a . character
+    localPart = localPart.replace(/^\./, '');
+    localPart = localPart.replace(/\.$/, '');
```
ST-DDT, line 117: This should also catch multiple dots:

```suggestion
    localPart = localPart.replace(/^\.+/, '');
    localPart = localPart.replace(/\.+$/, '');
```

ST-DDT: Alternatively:

```suggestion
    localPart = localPart.replace(/^\.+|\.+$/g, '');
```

test/internet.spec.ts
```diff
@@ -140,6 +140,29 @@ describe('internet', () => {
           expect(faker.definitions.internet.free_email).toContain(suffix);
         });
 
+        it('should not allow an email that starts with a .', () => {
+          const email = faker.internet.email('.Aiden');
```
ST-DDT:
```suggestion
          const email = faker.internet.email('...Aiden');
```

test/internet.spec.ts
```diff
@@ -140,6 +140,29 @@ describe('internet', () => {
           expect(faker.definitions.internet.free_email).toContain(suffix);
         });
 
+        it('should not allow an email that starts with a .', () => {
```
ST-DDT: Maybe add a similar test that ends with multiple dots / or combine them into one by having both.

src/modules/internet/index.ts
matthewmayer: but if we already collapsed multiple dots into one on line 113, there can only be max 1 dot at start or end?

test/internet.spec.ts
matthewmayer: the email function appends numbers etc though, so even if the first name and last name ends with dots, the email normally won't like
`faker.internet.email("Matt...","Mayer....")` // Matt.Mayer.63@gmail.com

src/modules/internet/index.ts
ST-DDT: Yeah, you are right.

src/modules/internet/index.ts
```diff
@@ -109,6 +109,13 @@ export class InternetModule {
       );
     }
 
+    //local parts may not contain two or more consecutive . characters
```
Shinigami92:
```suggestion
    // local parts may not contain two or more consecutive . characters
```

src/modules/internet/index.ts
```diff
@@ -109,6 +109,13 @@ export class InternetModule {
...
+    localPart = localPart.replace(/\.{2,}/g, '.');
+
+    //local parts may not start with or end with a . character
```
Shinigami92:
```suggestion
    // local parts may not start with or end with a . character
```

Shinigami92 (dismissed):

test/internet.spec.ts
ST-DDT: That may be true now, but what if we add a chance that there isn't an appendix to the name?

I think we can test is by changing the following line to:

````ts
const email = faker.internet.email('...Aiden...', '...Doe...');
````
and then test also for:

````ts
expect(prefix).not.toMatch(/\.{2}/);
expect(prefix).not.toMatch(/\.$/);
````

(and adjust the test name accordingly)

File: test/internet.spec.ts [hunk 6]
```diff
@@ -140,6 +140,29 @@ describe('internet', () => {
...
+          const email = faker.internet.email('...Aiden');
+
+          expect(email).toBeTruthy();
+          expect(email).toBeTypeOf('string');
+          expect(email).toSatisfy(validator.isEmail);
+
+          const [prefix] = email.split('@');
+          expect(prefix).not.toMatch(/^(\.)/);
```
ST-DDT: Is there a reason why you put the dot in parenthesis?

```suggestion
          expect(prefix).not.toMatch(/^\./);
```

ST-DDT (changes requested):
See above

ST-DDT (approved):

ST-DDT:
Thank you for fixing this.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #1886, opened 2023-02-28
docs: highlight changes in readme

### LINKED ISSUE #1869: Documentation Update
So I have just use faker again today, and found out that person module has been change to name. Might want to change the docs a bit to help others save a bit of time while creating their faker app.

ST-DDT:
Fixes #1869

Highlight changes since the last major release to make visitors more aware that they might be looking at the wrong branch.


ST-DDT:
Not sure whether this needs to be added to more places in the Readme.

ST-DDT:
Blocked by #1735 to avoid merge conflicts.

xDivisionByZerox:
Just to be clear. This is simply a workaround while in the developing phase of a major update, correct? If we do this  for minor versions as well this could quickly get out of hand since those "warnings" need to be removed on release as well.

ST-DDT:
> This is simply a workaround while in the developing phase of a major update, correct?

Yes, IMO we should keep this for the major release itself (so it will be visible in the v8 branch even after the release), but we will remove it from the next branch soon after. Since then the diff will only be v8.1 <-> v8.0 and there won't be breaking changes anyway. We might want to use it for deprecations (in the Readme's module overview) as well, just to give the users a hint that they should read the documentation.

matthewmayer:
The Templates section also includes a refernence to the person module. Maybe pick a different module entirely to demo this?

e.g.

```
console.log( faker.helpers.fake( 'The {{commerce.productName}} in the {{commerce.department}} department is on sale for  ${{commerce.price}}!' ) );
```

or indeed remove it altogether as we probably want to encourage people to simply use string templates instead.

ST-DDT:
> The Templates section also includes a refernence to the person module.

I marked it accordingly.

> Maybe pick a different module entirely to demo this?
> 
> e.g.
> 
> ```
> console.log( faker.helpers.fake( 'The {{commerce.productName}} in the {{commerce.department}} department is on sale for  ${{commerce.price}}!' ) );
> ```

Why should we change to a different example?

> or indeed remove it altogether as we probably want to encourage people to simply use string templates instead.

The fake function should be used if it isn't possible to use string templates, so IMO they serve different purposes.
And I hope that sufficiently skilled developers (developing an application where it matters) know when to use string templates.
I still added a reference to point users to the template literals.

matthewmayer:
Why should we change to a different example?

> Because then it works the same in both v7 and v8?

ST-DDT:
TODO:

- [x] Mark the changes once #1735 is merged

ST-DDT:
Ready for review

Shinigami92:
As an alternative we could just create a banner at the top that indicates to switch to v7 branch for that documentation

Currently this looks really polluting the README and we could consider instead e.g. to use alternative functions that are compatible in v7 and v8(next) and have the same signature :thinking:

ST-DDT:
> As an alternative we could just create a banner at the top that indicates to switch to v7 branch for that documentation

We already have this. But everybody just seems to scroll to the code.



> we could consider instead e.g. to use alternative functions that are compatible in v7 and v8(next) and have the same signature 🤔

We have breaking changes that prevent that as well as deprecated methods that I wouldnt want to promote anymore.

Shinigami92:
> > As an alternative we could just create a banner at the top that indicates to switch to v7 branch for that documentation
> 
> We already have this. But everybody just seems to scroll to the code.
> 
> > we could consider instead e.g. to use alternative functions that are compatible in v7 and v8(next) and have the same signature thinking
> 
> We have breaking changes that prevent that as well as deprecated methods that I wouldnt want to promote anymore.

Should we then remove the code blocks from the readme and point to the docs? Also that way we would advertise the docs more and make them more relevant.

matthewmayer:
I agree this ends up just looking messy

I think a more prominent banner might work (big text, red), or removing all code snippets and just leaving links to the next and stable versions of the docs.

Shinigami92:
**Team Decision**

We want to cleanup the README and link to the docs website
Maybe we want at least one simple (multiversion-stable) code example, but not for every module

ST-DDT:
I'm not sure which sections to keep, so I'm abandoning this PR.

Shinigami92:
Yeah, I can open another PR (later over the weekend maybe)
and clean it a bit
then we can discuss in the other PR if this cleanup fulfills all our expectations

PR #1888, opened 2023-03-03
docs: Add 'Sourcing data for definitions' to CONTRIBUTING

### LINKED ISSUE #1820: Add section to CONTRIBUTING about how to source data

It comes up fairly frequently e.g. on  https://github.com/faker-js/faker/pull/1778 and https://github.com/faker-js/faker/pull/1699  that it's not clear how to source data for locales. Perhaps we could add a new section to CONTRIBUTING with some tips e.g. 
https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md

### Suggested solution

Add a section of tips, something like

* Faker shouldn't contain copyrighted materials
* Facts cannot be copyrighted, so for example if you are translating a finite known list of things like the names of chemical elements into another language, that's fine
* But if you are compiling list of for example popular personal names or cities, don't copy directly from a single source (Wikipedia, 'most popular lists', government data sites etc) - as compiling a list of facts can be copyrighted https://en.wikipedia.org/wiki/Copyright_in_compilation
* It's best to refer to multiple sources and use your own judgement/knowledge to make a sample list of data

### Alternative

_No response_

### Additional context

_No response_

matthewmayer:
fix #1820


ST-DDT:
I like it. Should we add a section regarding the requirement to be "fit for professional audience"?

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

Shinigami92 (approved):

Shinigami92:
> I like it. Should we add a section regarding the requirement to be "fit for professional audience"?

This is (theoretically) already covered by [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md)
It's also not really related to the source for the data

The only helpful thing would be to co-reference links from and to CoC

PR #1892, opened 2023-03-04
docs: add reproducible results section

### LINKED ISSUE #1870: Faker dates are not reproducible with fixed seed.
### Pre-Checks

### Describe the bug

Faker-generated dates using past and recent functions are not reproducible given a fixed seed.  Seems like there is a slight drift in the regeneration....

Is there a way to fix the generated dates with a given seed?

### Minimal reproduction code

```JS
  // Run multiple times independently
  faker.seed(200)
  console.log('past: '+faker.date.past())
```
```
past: 2022-11-17T17:54:54.896Z
past: 2022-11-17T17:57:42.305Z 
past: 2022-11-17T17:58:07.209Z
past: 2022-11-17T17:58:29.581Z
past: 2022-11-17T17:58:54.245Z 
```

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 10.65 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.volta/tools/image/node/18.13.0/bin/node
    Yarn: 4.0.0-rc.39 - ~/.volta/tools/image/yarn/4.0.0-rc.39/bin/yarn
    npm: 8.19.3 - ~/.volta/tools/image/node/18.13.0/bin/npm
  Browsers:
    Brave Browser: 109.1.47.186
    Chrome: 110.0.5481.100
    Safari: 16.3
  npmPackages:
    @faker-js/faker: 8.0.0-alpha.0 => 8.0.0-alpha.0
```

- [ ] CJS
- [X] ESM

### Used Package Manager

pnpm

matthewmayer:
Improved docs re #1870 to "usage" page


ST-DDT (dismissed):

docs/guide/usage.md
```diff
@@ -84,6 +84,46 @@ In order to have Faker working properly, you need to check if these `compilerOpt
...
+
+// Setting the seed again resets the sequence.
+faker.seed(123);
+
+const secondRandom = faker.number.int();
+
+console.log(firstRandom === secondRandom);
+```
+
+Note that when upgrading to a new version of Faker, you may get different values for the same seed, as the underlying data (lists of names, words etc) may have changed.
```
Shinigami92: Should we actually embed this into a "note" block?

matthewmayer: done

ST-DDT (dismissed):

ejcheng (dismissed):

Shinigami92 (dismissed):

docs/guide/usage.md
```diff
@@ -84,6 +84,48 @@ In order to have Faker working properly, you need to check if these `compilerOpt
...
+const secondRandom = faker.number.int();
+
+console.log(firstRandom === secondRandom);
+```
+
+::: info NOTE
+When upgrading to a new version of Faker, you may get different values for the same seed, as the underlying data (lists of names, words etc) may have changed.
+:::
+
+There are a few relative date methods for which setting a random seed is not sufficient to have reproducible results, specifically: `faker.date.past`, `faker.date.future`, `faker.date.birthdate`, `faker.date.recent` and `faker.date.soon`. This is because these methods default to creating a date before or after "today", and "today" depends on when the code is run. To fix this, you can specify a fixed reference date as a Date or string, for example:
```
xDivisionByZerox: I would not specifically list the methods which fit the criteria since we would need to edit this documentation section based on new features or even changes made in the future. I would say something like "Methods which interact with dates" instead.

If we do want to list all methods `faker.git.commitEntry` is missing. I want to point that out specifically as the current descriptions makes it look like this only affects the `DateModule`.

xDivisionByZerox (changes requested):

docs/guide/usage.md
matthewmayer: i added git.commitEntry and changed the wording to make it not sound like this is an exhaustive list (though it currently is), so in the event another relative date method is added, the docs are not wrong.

xDivisionByZerox (approved):

ejcheng (approved):

ST-DDT (approved):

PR #1897, opened 2023-03-05
feat: git committed file

beaunus:
**This is an early draft, feedback and suggestions welcome, but please no reviews yet.**

---

This PR will eventually make it so `git.commitEntry` includes a variety of file lines. E.g.

```
-	-	{public/img => images/logos}/logo_pixels.png
1	0	images/logos/index.ts
4	0	images/photos/index.ts
```

### DIFF
--- a/src/modules/git/index.ts
+++ b/src/modules/git/index.ts
@@ -246,4 +246,32 @@ export class GitModule {
     });
     return this.commitSha({ length: 7 });
   }
+
+  committedFile({
+    isBinary,
+    nameChange,
+    isNumStat,
+  }: {
+    isBinary?: boolean;
+    nameChange?: 'partial' | 'full';
+    isNumStat?: boolean;
+  } = {}): string {
+    const numLinesAdded = isBinary ? '-' : this.faker.number.int();
+    const numLinesDeleted = isBinary ? '-' : this.faker.number.int();
+
+    const oldFilePath = this.faker.system.directoryPath().substring(1);
+    const oldFileName = this.faker.system.fileName();
+    const newFilePath = this.faker.system.directoryPath().substring(1);
+    const newFileName = this.faker.system.fileName();
+
+    return [
+      numLinesAdded,
+      numLinesDeleted,
+      nameChange === 'full'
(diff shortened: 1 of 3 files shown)


matthewmayer:
FYI you could probably make use of faker.system.commonFileName and other methods in faker.system.* for the paths and filenames

src/modules/git/index.ts
```diff
@@ -246,4 +246,11 @@ export class GitModule {
     });
     return this.commitSha({ length: 7 });
   }
+
+  committedFile({ isBinary = false, isNumStat = false } = {}): string {
```
Shinigami92:
```suggestion
  /**
   * fill this
   */
  committedFile(options: {
    /**
     * fill this
     */
    isBinary?: boolean,
    /**
     * fill this
     */
    isNumStat?: boolean
  } = {}): string {
    const { isBinary = false, isNumStat = false } = options;
```

Shinigami92 (changes requested):

beaunus:
> FYI you could probably make use of faker.system.commonFileName and other methods in faker.system.* for the paths and filenames

Thank you for the pointer!

src/modules/git/index.ts
beaunus: Thank you!

I'll close this PR and re-open when it's ready for review.

I thought that the "Draft" status would not get the attention of the review team. I'll do my best to conform to the CONTRIBUTING doc when it's ready for review.

ST-DDT: Sorry for the trouble. We are currently undecided when and what to review in draft PRs. From my experience, I can tell that the rules vary from repo to repo so it might be best to explicitly mention the own expectations in the description.
Something like:
> This is an early draft, feedback and suggestions welcome, but please no reviews yet.

Shinigami92: Oh that was my bad, it was not my intention to let you close the PR
and setting it to draft is the correct way
I just had the time and will to just write that suggestion/review
And yes, we are a very active repository ^^

ST-DDT:
How is the status on this PR?

ST-DDT:
It looks like this PR got stale. I'm closing this. Feel free to comment/create a new PR when you wish to continue.

PR #1910, opened 2023-03-07
feat: provide enums for color values

Shinigami92:
Extracted from https://github.com/faker-js/faker/pull/1748#discussion_r1128162369


src/modules/color/index.ts
```diff
@@ -205,7 +212,7 @@ export class ColorModule {
    * @since 7.0.0
    */
   cssSupportedFunction(): string {
-    return this.faker.helpers.arrayElement(CSS_FUNCTIONS);
+    return this.faker.helpers.arrayElement(Object.values(CssFunction));
```
ST-DDT:
```suggestion
    return this.faker.helpers.objectValue(CssFunction);
```

src/modules/color/index.ts
```diff
@@ -217,7 +224,7 @@ export class ColorModule {
    * @since 7.0.0
    */
   cssSupportedSpace(): string {
-    return this.faker.helpers.arrayElement(CSS_SPACES);
+    return this.faker.helpers.arrayElement(Object.values(CssSpace));
```
ST-DDT:
```suggestion
    return this.faker.helpers.objectValue(CssSpace);
```

ST-DDT (changes requested):

src/modules/color/index.ts
```diff
@@ -3,32 +3,39 @@ import type { Faker } from '../../faker';
...
-export const CSS_SPACES = [
-  'sRGB',
-  'display-p3',
-  'rec2020',
-  'a98-rgb',
-  'prophoto-rgb',
-  'rec2020',
-] as const;
+export enum CssSpace {
+  SRGB = 'sRGB',
```
xDivisionByZerox, line 7: Should we add additional JSDocs to each key?

src/modules/color/index.ts
xDivisionByZerox: While this works in the current situation, we need to be careful with numeric enums.

```ts
enum Foo {
  BAR = 0;
};

console.log(Foo); // { BAR: 0, "0": "BAR" }, this is the JS object
console.log(faker.helpers.objectValue(Foo)); // 0 | "BAR"
console.log(faker.helpers.objectKey(Foo)); // "0" | "BAR"
```

How do you feel about `enumKey` and `enumValue` as helper function for those use cases?

ST-DDT: @xDivisionByZerox Could you please create a separate issue/PR for that?
Maybe with a hint on how to distinguish between enum keys and values?

xDivisionByZerox: Yes of course. I first wanted to clarify if we do want to be concerned about that topic or if we simply ignore it. 
I'll create an issue for further discussions. Mark my concern as resolved here.

Shinigami92: Do not use Ordinal Enums in TypeScript, ever
Or at least only if you are not a library but an App that uses these enums internally
Enums in TypeScripts are nearly to antipattern and just pollute the emitted code
There are mostly any time a better way instead of using Enums at all in JS

- https://thoughtbot.com/blog/the-trouble-with-typescript-enums
- https://dev.to/dvddpl/whats-the-problem-with-typescript-enums-2okj
- https://www.reddit.com/r/typescript/comments/yr4vv5/typescript_enums_are_terrible_heres_why/
- https://bluepnume.medium.com/nine-terrible-ways-to-use-typescript-enums-and-one-good-way-f9c7ec68bf15
- https://blog.logrocket.com/why-typescript-enums-suck/

Shinigami92: > [...]
> 
> How do you feel about `enumKey` and `enumValue` as helper function for those use cases?

Only implement this, if someone from outside the core members is asking for it

src/modules/color/index.ts
Shinigami92: We also did not for `enum Sex`

I don't think it is necessary

ST-DDT (approved):

src/modules/color/index.ts
ST-DDT: FYI: My comment was more about implementability than actual implementation.

xDivisionByZerox (approved):

PR #1913, opened 2023-03-07
fix(locale): remove invalid city patterns from pt_BR

matthewmayer:
```
> allFakers.pt_BR.location.city()
// 'undefined Hugo'
```

This is because two of the patterns for `city` reference `city_prefix`, but this is empty for pt_PT
https://github.com/faker-js/faker/blob/next/src/locales/pt_BR/location/city_prefix.ts

Removed these patterns

### DIFF
--- a/src/locales/pt_BR/location/city.ts
+++ b/src/locales/pt_BR/location/city.ts
@@ -1,6 +1,4 @@
 export default [
-  '{{location.city_prefix}} {{person.firstName}}{{location.city_suffix}}',
-  '{{location.city_prefix}} {{person.firstName}}',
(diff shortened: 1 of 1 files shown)


ejcheng (approved):

ST-DDT (approved):

xDivisionByZerox:
Should we remove the `city_prefix.ts` file (in this locale) entirely?

ST-DDT:
> Should we remove the `city_prefix.ts` file (in this locale) entirely?

I think the empty array here is a marker that these data are intentionally missing instead of just being missing.

xDivisionByZerox (approved):

PR #1918, opened 2023-03-11
docs: enhance readme

### LINKED ISSUE #1869: Documentation Update

Shinigami92:
closes #1869


ejcheng (dismissed):

README.md
```diff
@@ -88,33 +90,14 @@ Thus, limiting the import to a single locale can speed up startup times.
 ## 💎 Modules
 
 An in-depth overview of the API methods is available in the [documentation](https://fakerjs.dev/guide/).
```
ST-DDT: This should probably point to /api/ and maybe also point to the next version of the docs.

README.md
```diff
@@ -88,33 +90,14 @@ Thus, limiting the import to a single locale can speed up startup times.
...
-| Vehicle  | `faker.vehicle.vehicle()`                     | Lamborghini Camry                                                                                  |
-| Word     | `faker.word.adjective()`                      | adorable                                                                                           |
+The API currently covers the following modules:
+
+| Modules  |          |         |          |         |         |
+| -------- | -------- | ------- | -------- | ------- | ------- |
+| Airline  | Company  | Finance | Image    | Music   | System  |
+| Animal   | Database | Git     | Internet | Person  | Vehicle |
+| Color    | Datatype | Hacker  | Location | Phone   | Word    |
+| Commerce | Date     | Helpers | Lorem    | Science |
```
ST-DDT: Can you please share the your idea/thoughts for using six columns for this?

Should we make this list a bit more usable, by adding links to each of them?

---

IMO the current list does not add any additional value to the feature list above.

Shinigami92: > Can you please share the your idea/thoughts for using six columns for this?

No direct reason for 6 columns
It initially was more like 4 rows
But then I counted them and counted that they are 23 modules right now. Sadly this is a prime number :sweat: 

Do you want another number of rows/columns?

I must also say, that it looked a bit better in the VSCode preview

because there are no gutter lines, so the "headline" `Modules` reads as the whole column-span headline.

> Should we make this list a bit more usable, by adding links to each of them?

Good idea, will do it (after you provided more info about the next question below)

> IMO the current list does not add any additional value to the feature list above.

So you suggest to remove it completely? Or do you want to suggest something else?

ST-DDT: > > IMO the current list does not add any additional value to the feature list above.
> 
> So you suggest to remove it completely? Or do you want to suggest something else?

My suggestion is to keep it as is or to remove it completely.

matthewmayer: I would remove particularly as the module list changes from version to version and the reason for the simplified README is to stop people relying on skimming the wrong README to get an overview of the code.

README.md
```diff
@@ -18,7 +18,9 @@
 
 ## 📙 API Documentation
 
-[Guide - Getting started](https://fakerjs.dev/guide/)
+<h1>⚠️ You are reading the docs for the <a href="https://github.com/faker-js/faker/tree/next">next</a> branch ⚠️</h1>
+
+Please proceed to the [Getting started Guide](https://fakerjs.dev/guide/) for the **stable** release of Faker.
 
 For detailed API documentation, please select the version of the documentation you are looking for.
```
matthewmayer, line 26: I feel the table of versions could also be simplified. Currently there are 5 rows but only 3 actually different versions. And the "website" seems much more important than the Github link for getting people to the right place. Could be simplified to:
```
| Version  | Website                   |
| :-----: | :------------------------ |
| v8 (next)  |  https://next.fakerjs.dev/   |
| v7 (stable)  |   https://fakerjs.dev/   |
| v6 (old)  |   https://v6.fakerjs.dev/   |
```

Shinigami92: Will change that if @ST-DDT gives also his okay to that

Alongside I managed `npm` a bit and added some dist-tags

```bash
> npm dist-tag ls @faker-js/faker

latest: 7.6.0
old: 6.3.1
stable: 7.6.0
next: 8.0.0-alpha.0
alpha: 8.0.0-alpha.0
```

README.md
```diff
@@ -18,7 +18,9 @@
 
 ## 📙 API Documentation
 
-[Guide - Getting started](https://fakerjs.dev/guide/)
+<h1>⚠️ You are reading the docs for the <a href="https://github.com/faker-js/faker/tree/next">next</a> branch ⚠️</h1>
+
+Please proceed to the [Getting started Guide](https://fakerjs.dev/guide/) for the **stable** release of Faker.
```
ST-DDT:
```suggestion
Please proceed to the [Getting Started Guide](https://fakerjs.dev/guide/) for the **stable** release of Faker.
```

README.md
```diff
@@ -87,34 +89,7 @@ Thus, limiting the import to a single locale can speed up startup times.
...
-| Person   | `faker.person.firstName()`                    | Cameron                                                                                            |
-| Phone    | `faker.phone.phoneNumber()`                   | +1 291-299-0192                                                                                    |
-| Science  | `faker.science.unit()`                        | `{ name: 'meter', symbol: 'm' }`                                                                   |
-| System   | `faker.system.directoryPath()`                | /root                                                                                              |
-| Vehicle  | `faker.vehicle.vehicle()`                     | Lamborghini Camry                                                                                  |
-| Word     | `faker.word.adjective()`                      | adorable                                                                                           |
+An in-depth overview of the API methods is available in the [documentation](https://fakerjs.dev/api/).
```
ST-DDT: Should we change all links here to point to the next.fakerjs.dev website?

Since `You are reading the docs for the next branch`

README.md
ST-DDT: OK

README.md
matthewmayer: why not both?

`An in-depth overview of the API methods is available in the documentation for [v7 (stable)](https://fakerjs.dev/api/) and [v8 (next)](https://next.fakerjs.dev/api/).`

ejcheng (approved):

matthewmayer (approved):

xDivisionByZerox (approved):

ST-DDT (approved):
As per team decision

PR #1920, opened 2023-03-11
feat(helpers): new method enumValue

ocodista:
With this method it is possible to get a random value from an enum.

This is very useful for not having to harcode an enum when generating fake data.

<details>
<summary>Docs-Preview</summary>

grafik

</details>


test/__snapshots__/datatype.spec.ts.snap
```diff
@@ -193,6 +201,14 @@ exports[`datatype > 1211 > datetime > with min 1`] = `2065-11-10T19:27:20.915Z`;
 
 exports[`datatype > 1211 > datetime > with min and max 1`] = `1789-03-26T15:44:45.219Z`;
 
+exports[`datatype > 1211 > enum > with default enum 1`] = `"C"`;
+
+exports[`datatype > 1211 > enum > with enum starting from some index 1`] = `"A"`;
+
+exports[`datatype > 1211 > enum > with string enum 1`] = `undefined`;
```
ST-DDT: This doesnt seem right.

test/__snapshots__/datatype.spec.ts.snap
```diff
@@ -323,6 +339,14 @@ exports[`datatype > 1337 > datetime > with min 1`] = `1747-07-16T01:19:54.159Z`;
 
 exports[`datatype > 1337 > datetime > with min and max 1`] = `1669-06-22T01:21:21.236Z`;
 
+exports[`datatype > 1337 > enum > with default enum 1`] = `"A"`;
+
+exports[`datatype > 1337 > enum > with enum starting from some index 1`] = `undefined`;
```
ST-DDT: Neither does this.

ST-DDT (changes requested):

test/__snapshots__/datatype.spec.ts.snap
ocodista: Fixed

test/__snapshots__/datatype.spec.ts.snap
ocodista: You're right, fixed now!

src/modules/datatype/index.ts
```diff
@@ -84,6 +84,34 @@ export class DatatypeModule {
...
   }
 
+  /**
+   * Returns a random value from an Enum object
+   *
+   * @param enumObject
+   *
+   * @example
+   * faker.datatype.enum(enum { A, B, C}) // 'C'
+   * faker.datatype.enum(enum { A, B, C}) // 'A'
```
ST-DDT: These three examples all have the same parameters. Please change them to show different usecases and remove redundant ones.

test/datatype.spec.ts
```diff
@@ -50,6 +50,29 @@ describe('datatype', () => {
...
+
+      enum stringEnum {
+        A = 'A',
+        B = 'B',
+        C = 'C',
+      }
+      t.it('with default enum', defaultEnum)
+        .it('with enum starting from some index', enumStartingFromSomeInt)
+        .it('with string enum', stringEnum);
+    });
```
ST-DDT: Afaik it is also possible to define mixed enums. Could you please define a test for them too?

src/modules/datatype/index.ts
```diff
@@ -84,6 +84,34 @@ export class DatatypeModule {
...
+    if(!enumObject) return '0' as SpecificEnumType
+    const ignoreTypeScriptAddedKeys = (enumKeys: string[]) => {
+      return enumKeys.filter((key) => isNaN(Number(key)));
+    };
+    const keys = Object.keys(enumObject);
+    const enumKeys = ignoreTypeScriptAddedKeys(keys);
+    const { length } = enumKeys;
+    const validIndexes = length - 1;
+    const randomEnumIndex = this.faker.number.int(validIndexes);
+    const randomOption = (enumObject as any)[
```
ST-DDT: You can use helpers.arrayElement(keys) to simplify the code a bit.

ST-DDT (commented):
This method closely resembles how helpers.arrayElement and objectValue work, so we should move it to the helpers module instead.

And dont forget to run `pnpm run preflight` when you are ready.

test/datatype.spec.ts
```diff
@@ -50,6 +50,29 @@ describe('datatype', () => {
...
+        A = 2,
+        B,
+        C,
+      }
+
+      enum stringEnum {
+        A = 'A',
+        B = 'B',
+        C = 'C',
+      }
```
Shinigami92: 1. Please use `PascalCase` for enum names

2. Please use representative example enums like `Color {Red,Green,Blue}` or `Direction {North,East,South,West}` or something like that

test/datatype.spec.ts
```diff
@@ -50,6 +50,29 @@ describe('datatype', () => {
         });
     });
 
+    t.describe('enum', (t) => {
```
Shinigami92: Please also add tests below in the `random seeded tests for seed ...` section

src/modules/datatype/index.ts
xDivisionByZerox: I suggest explicitly providing one enum with numeric and string values each. Something like:
```ts
* @example
* faker.datatype.enum(enum Strings { HELLO: 'world', FOO: 'bar' }) // 'HELLO'
* faker.datatype.enum(enum Numbers { FOO: 0, BAR, 1, BAZ: 2 }) // 'FOO'
```

src/modules/datatype/index.ts
```diff
@@ -84,6 +84,34 @@ export class DatatypeModule {
     return this.faker.number.float({ min, max, precision });
   }
 
+  /**
+   * Returns a random value from an Enum object
```
xDivisionByZerox:
```suggestion
   * Returns a random value from an Enum object.
```

matthewmayer:
How will this method work in JavaScript given that enums are typescript-specific?

src/modules/datatype/index.ts
```diff
@@ -84,6 +84,34 @@ export class DatatypeModule {
...
+   *
+   * @param enumObject
+   *
+   * @example
+   * faker.datatype.enum(enum { A, B, C}) // 'C'
+   * faker.datatype.enum(enum { A, B, C}) // 'A'
+   * faker.datatype.enum(enum { A, B, C}) // 'B'
+   */
+  enum<GenericEnumType extends Object, SpecificEnumType>(
```
xDivisionByZerox: I would suggest moving the implementation into the `HelpersModule` as we plan on deprecating the `DatatypeModule`.

src/modules/datatype/index.ts
xDivisionByZerox: The description says that an enum value is returned, but the current implementation returns a key. So which one is correct?

Shinigami92:
> How will this method work in JavaScript given that enums are typescript-specific?

Funny answer: That's the funny part, it wont

Technical correct answer:
Enums are converted to JS bloatware code constructs like this:
```ts
enum Color { Red, Green, Blue }

var Color;
(function (Color) {
    Color[Color["Red"] = 0] = "Red";
    Color[Color["Green"] = 1] = "Green";
    Color[Color["Blue"] = 2] = "Blue";
})(Color || (Color = {}));
```

So normally you are BY FAR more better suited with setups like `var Color = Object.freeze({ Red: 0, Green: 1, Blue: 2 })`, which are not only provide less bloated code, but also performs much better at runtime as they can be optimized by JS engines.
Additionally you are bringing just JS over TS into JS and TS doesn't need to first transpile it to something JS-usable.

Then there are `const enums` to prevent generating of such bloatware code but it theoretically disables the need in total of enums and you can just use `type` string literals (like e.g. our `SexType`). These then just check at compile time if the type literal is suitable for the intended call/usage and has zero runtime overhead.

TL;DR: enums in TS are a total misconcept and failure and Microsoft / the TS Team knows that and communicate that in the past already (I just sadly don't find a source directly for that).

So there is only few really case for using enums (and the should be used carefully):
- when your programmer audience is not JS native and have also OOP background like Java or C#
- when you are using them only in your own code and want to over-optimize the readability of your TS code base

When using them in a library (like FakerJS), **always ensure** they are 100% optional and you can always call the function in a readable easy way like using string literals.

xDivisionByZerox:
> Funny answer: That's the funny part, it wont

I call BS. An enum as you pointed out is simply converted to a JS object. That means you can use it like a JS object. That means you can use all methods as you would in TS. The catch is that those enum objects need to be handled a little bit differently on our side.

String enums work the same as normal objects would.
```ts
// in js
const myObject = Object.freeze({
  FOO: 'bar',
  HELLO: 'world',
});

faker.helpers.objectKey(myObject); // 'FOO' | 'HELLO'
faker.helpers.enumKey(myObject); // 'FOO' | 'HELLO'

faker.helpers.objectValue(myObject); // 'bar' | 'world'
faker.helpers.enumValue(myObject); // 'bar' | 'world'

// in ts
enum MyEnum {
  FOO: 'bar',
  HELLO: 'world',
}

faker.helpers.objectKey(MyEnum ); // 'FOO' | 'HELLO'
faker.helpers.enumKey(MyEnum); // 'FOO' | 'HELLO'

faker.helpers.objectValue(MyEnum ); // 'bar' | 'world'
faker.helpers.enumValue(MyEnum); // 'bar' | 'world'
```

Numeric enums are working (as you pointed out) a bit differently. You have to know this when working with enums (which sucks, no doubt). So while not working as expected, you can make them work!
```ts
// in js
const myObject = Object.freeze({
  FOO: 0,
  BAR: 1,
  // here is the catch
  0: 'FOO',
  1: 'BAR',
});

faker.helpers.objectKey(myObject); // 'FOO' | 'HELLO' | 0 | 1
faker.helpers.enumKey(myObject); // 'FOO' | 'HELLO'

faker.helpers.objectKey(myObject); // 'FOO' | 'HELLO' | 0 | 1
faker.helpers.enumValue(myObject); // 0 | 1

// in ts
enum MyEnum {
  FOO: 0,
  BAR: 1,
}

faker.helpers.objectKey(MyEnum); // 'FOO' | 'HELLO' | 0 | 1
faker.helpers.enumKey(MyEnum); // 'FOO' | 'HELLO'

faker.helpers.objectKey(MyEnum); // 'FOO' | 'HELLO' | 0 | 1
faker.helpers.enumValue(MyEnum); // 0 | 1
```

By providing the helper functions we essentially provide a bridge that allows users to handle enums as they expect.

> [...] but also performs much better at runtime as they can be optimized by JS engines.

Source? 
While the JS code IS bloaty, it's just an IIFE that assigns the object keys and values. Why should that be sooo underperformed in contrast to "plain JS object creation and freezing"? Unless you talking about some nanosecond BS again...

Shinigami92:
@xDivisionByZerox please be careful in not getting personally and callout terms like "BS"

Yes indeed I'm talking about nano-seconds performance improvements. But also even on a code readability perspective it's possible to confuse folks with ordinal enums more then it helps. `string literal types` on the other hand are understandable, have no hidden side-effects like getting converted to ordinals in the background and other stuff.

xDivisionByZerox:
> @xDivisionByZerox please be careful in not getting personally and callout terms like "BS"

You gave out straight-up wrong information. Why am not allowed to call that out? I wasn't calling YOU out but the information you gave.

Edit:
Oh, I see. You probably are referring to "nanosecond BS". I get it might be a strong word in that context, but nevertheless is negligible.
I still apologize if you felt insulted by that.

Edit2:
"straight-up wrong information" => "Funny answer: That's the funny part, it wont"

matthewmayer:
Could one extend objectKey/objectValue instead to specifically handle the case of it being passed a TS enum?

ST-DDT:
> Could one extend objectKey/objectValue instead to specifically handle the case of it being passed a TS enum?

This would degrade the usability of those methods for user defined objects, that actually use numeric keys. So no.

matthewmayer:
This would be the only Faker method which requires typescript to call into right? I think that would be worthy of calling out in the documentation for this function if accepted.

ST-DDT:
> This would be the only Faker method which requires typescript to call into right? I think that would be worthy of calling out in the documentation for this function if accepted.

AFAICT this will also work with plain json object pseudo enums (it is basically the same as objectValue while ignoring numeric keys).

````js
faker.helpers.enumValue(Color);
````

Works for both:

````ts
enum Color { Red, Green, Blue }
````

and

````js
const Color = { Red: 0, Green: 1, Blue: 2 }
````

And still provides a minuscule readability improvement over `objectValue` by conveying the given value is an "enum" instead of just some magic object.

We might want to add a hint that it ignores numeric keys though (and add an `@see` link to objectValue).

File: src/modules/datatype/index.ts (hunk 3 above)
Comment (ST-DDT) [reply src/modules/datatype/index.ts]: And make sure the examples actually compile.

Comment (ocodista) [reply src/modules/datatype/index.ts]: Fixed, now added the three examples and included similar ones to the tests

File: src/modules/datatype/index.ts (hunk 5 above)
Comment (ocodista) [reply src/modules/datatype/index.ts]: Done.

File: test/datatype.spec.ts (hunk 6 above)
Comment (ocodista) [reply test/datatype.spec.ts]: Done.

File: test/datatype.spec.ts (hunk 7 above)
Comment (ocodista) [reply test/datatype.spec.ts]: Done.

File: src/modules/datatype/index.ts (hunk 9 above)
Comment (ocodista) [reply src/modules/datatype/index.ts]: Agreed, moved.

File: src/modules/datatype/index.ts (hunk 8 above)
Comment (ocodista) [reply src/modules/datatype/index.ts]: Fixed this, now the implementation always returns a value from the enum.

File: src/modules/helpers/index.ts [hunk 10]
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
...
+   *
+   * @example
+   * faker.datatype.enum(enum Color { Red, Green, Blue}) // 'Green'
+   * faker.datatype.enum(enum Direction { North = 'North', South = 'South'}) // 'South'
+   * faker.datatype.enum(enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }) // 200
+   */
+  enumValue<GenericEnumType extends Object, SpecificEnumType>(
+    enumObject: GenericEnumType
+  ): SpecificEnumType {
+    if(!enumObject) return '0' as SpecificEnumType
```
ST-DDT: The parameter is required, no need to handle it being absent.

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
...
+    const ignoreTypeScriptAddedKeys = (enumKeys: string[]) => {
+      return enumKeys.filter((key) => isNaN(Number(key)));
+    };
+    const keys = ignoreTypeScriptAddedKeys(Object.keys(enumObject));
```
ST-DDT: IMO you can inline this since you only use it once:

```suggestion
    const keys = Object.keys(enumObject)
        // ignore numeric keys added by TypeScript
        .filter((key) => isNaN(Number(key)));
```

(Not sure about the formatting)

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
...
+    const randomKey = this.arrayElement(keys)
+    const enumValue = (enumObject as any)[randomKey]
+    return enumValue;
```
ST-DDT: IMO you can just return the value:

```suggestion
    return (enumObject as any)[randomKey]
```

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
...
+   * @param enumObject
+   *
+   * @example
+   * faker.datatype.enum(enum Color { Red, Green, Blue}) // 'Green'
+   * faker.datatype.enum(enum Direction { North = 'North', South = 'South'}) // 'South'
+   * faker.datatype.enum(enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }) // 200
+   */
+  enumValue<GenericEnumType extends Object, SpecificEnumType>(
+    enumObject: GenericEnumType
+  ): SpecificEnumType {
```
ST-DDT: Note to self: Check generic type

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
     return arrayCopy.slice(min);
   }
 
+  /**
+   * Returns a random value from an Enum object.
+   *
+   * @param enumObject
```
ST-DDT: Missing description

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,29 @@ export class HelpersModule {
...
+   *
+   * @example
+   * faker.datatype.enum(enum Color { Red, Green, Blue}) // 'Green'
+   * faker.datatype.enum(enum Direction { North = 'North', South = 'South'}) // 'South'
+   * faker.datatype.enum(enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }) // 200
```
```suggestion
   * faker.helpers.enumValue(enum Color { Red, Green, Blue}) // 'Green'
   * faker.helpers.enumValue(enum Direction { North = 'North', South = 'South'}) // 'South'
   * faker.helpers.enumValue(enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }) // 200
```

+ please make sure these actually compile

ST-DDT (changes requested):
Please run `pnpm run preflight` to get a list of lint and test errors locally before pushing your changes.

test/helpers.spec.ts
```diff
@@ -159,6 +190,48 @@ describe('helpers', () => {
...
+      describe('enumValue', () => {
+        enum ColorValueEnum { Red, Green, Blue }
+        enum ColorValueWithStartIndexEnum { Red = 3, Green, Blue }
+        enum ColorStringEnum { Red = "RED", Green = "GREEN", Blue = "BLUE" }
+        enum FooMixedEnum { FOO = 0, BAR = 1, STR_FOO = 'FOO', STR_BAR = 'BAR' }
+
+        it('should return a value from a numeric enum', () => {
+          const randomColorValue = faker.helpers.enumValue(ColorValueEnum)
+          expect([ColorValueEnum.Red, ColorValueEnum.Green, ColorValueEnum.Blue]).toContain(randomColorValue)
+        })
```
Shinigami92: In line 631-632 the docs says:
```ts
enum Color { Red, Green, Blue }
faker.helpers.enumValue(Color) // 'Green'
```

But this seems to be wrong, because it returns `1` and not `'Green'`.

Shinigami92 (changes requested):

src/modules/helpers/index.ts
```diff
@@ -629,7 +629,7 @@ export class HelpersModule {
    *
    * @example
    * enum Color { Red, Green, Blue }
-   * faker.helpers.enumValue(Color) // 'Green'
+   * faker.helpers.enumValue(Color) // 1
```
matthewmayer: while correct, this example doesn't seem very explanatory. (why 1?). It would be helpful to clarify this equals the value of one of the enum values, something like this?
```
const randomColor = faker.helpers.enumValue(Color) // randomColor is 1 
randomColor === Color.Green // true
```

matthewmayer (dismissed):

src/modules/helpers/index.ts
ST-DDT: IMO we can achieve the same with the comment:

```suggestion
   * faker.helpers.enumValue(Color) // 1 (Green)
```

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,30 @@ export class HelpersModule {
     return arrayCopy.slice(min);
   }
 
+  /**
+   * Returns a random value from an Enum object.
+   *
```
ST-DDT, line 627:
```suggestion
   *
   * This does the same as `objectValue` except that it ignores (the values assigned to) the numeric keys added for typescript enums.
   *
```

ST-DDT (commented):
⚠️ Please fix all formatting, lint and test errors/run `pnpm run preflight` before pushing.

ocodista:
> ⚠️ Please fix all formatting, lint and test errors/run `pnpm run preflight` before pushing.

I fixed all of the warnings but this one related to docs version, any idea how I should fix it?
image

test/helpers.spec.ts
```diff
@@ -159,6 +190,94 @@ describe('helpers', () => {
...
+        });
+
+        it('should compile examples', () => {
+          expect(() => {
+            enum Color {
+              Red,
+              Green,
+              Blue,
+            }
+            faker.helpers.enumValue(Color); // 'Green'
```
Shinigami92: I would expect 1 here, as there are no string literals assigned :thinking:

test/helpers.spec.ts
```diff
@@ -159,6 +190,94 @@ describe('helpers', () => {
...
+          const randomFooValue = faker.helpers.enumValue(FooMixedEnum);
+          expect([
+            FooMixedEnum.Foo,
+            FooMixedEnum.Bar,
+            FooMixedEnum.StrBar,
+            FooMixedEnum.StrFoo,
+          ]).toContain(randomFooValue);
+        });
+
+        it('should compile examples', () => {
```
Shinigami92: I'm not sure if we need this test in this way, as here you are just checking that TypeScript works :thinking: do you?

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,37 @@ export class HelpersModule {
...
+   * This does the same as `objectValue` except that it ignores (the values assigned to) the numeric keys added for typescript enums.
```
Shinigami92: > [...] except that it ignores (the values assigned to) the numeric keys [...]

Could you add one test case for that?
Or in other words: can you please add at least one example where `enumValue` is usable and provide a different result as to `objectValue`?
Currently I'm not sure how to use it and get a different, actual useful result.

Shinigami92 (changes requested):

test/helpers.spec.ts
ST-DDT: Looks like debug code to me.

src/modules/helpers/index.ts
ST-DDT: `faker.helpers.enumValue(Color) // 1 (Green)`

whereas objectValue might also return Red.

````ts
enum Color { Red, Green, Blue }
````
->
````js
var Color;
(function (Color) {
    Color[Color["Red"] = 0] = "Red";
    Color[Color["Green"] = 1] = "Green";
    Color[Color["Blue"] = 2] = "Blue";
})(Color || (Color = {}));
````
->
````js
{ 0: "Red", 1: "Green", 2: "Blue", Red: 0, Green: 1, Blue: 2 }
````
->
````ts
objectValue()  => Red, 0, Green, 1, Blue, 2
enumValue() => 0, 1, 2
````

File: src/modules/helpers/index.ts [hunk 22]
```diff
@@ -622,6 +622,37 @@ export class HelpersModule {
...
+   *
+   * @example
+   * enum Color { Red, Green, Blue }
+   * faker.helpers.enumValue(Color) // 1 (Green)
+   *
+   * enum Direction { North = 'North', South = 'South'}
+   * faker.helpers.enumValue(Direction) // 'South'
+   *
+   * enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }
+   * faker.helpers.enumValue(HttpStatus) // 200
```
```suggestion
   * faker.helpers.enumValue(HttpStatus) // 200 (Ok)
```

ST-DDT:
> With this method it is possible to get a random value from an enum.
> 
> This is very useful for not having to harcode an enum when generating fake data.

Just a question on the side:
Do you need this feature yourself or do you just consider it useful?
What do you use (non-const) ordinal enums for?

src/modules/helpers/index.ts
Shinigami92: A reason more **for me** to 100% prevent usage of ordinal non constant enums

ocodista:
> > With this method it is possible to get a random value from an enum.
> > This is very useful for not having to harcode an enum when generating fake data.
> 
> Just a question on the side: Do you need this feature yourself or do you just consider it useful? What do you use (non-const) ordinal enums for?

Yes, I would use it.

In fact, I'm using it right now together with faker js code, for enums.

The project I work at have a lot of TypeScript enums. like categories, userTypes, statuses, etc

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,37 @@ export class HelpersModule {
...
+   *
+   * enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }
+   * faker.helpers.enumValue(HttpStatus) // 200 (Ok)
+   *
+   * @since 8.0.0
+   */
+  enumValue<
+    GenericEnumType extends Record<string | number, string | number>,
+    SpecificEnumType
+  >(enumObject: GenericEnumType): SpecificEnumType {
```
ST-DDT: I did a bit of fiddling with typescript and found the correct typing for this method:

````ts
  enumValue<
    EnumType extends Record<string | number, string | number>,
    EnumValueType extends
      | {
          [key in KeySearch<EnumType, string>]: `${EnumType[key]}`;
        }[KeySearch<EnumType, string>]
      | {
          [key in KeySearch<EnumType, number>]: EnumType[key];
        }[KeySearch<EnumType, number>]
  >(enumObject: EnumType): EnumValueType {
````

I used named generic types for readability, not sure whether we want them in the final source code.
@Shinigami92: ^

Sadly it requires a small helper type:

````ts
type KeySearch<
  T extends Record<string | number, string | number>,
  S extends string | number
> = {
  [K in keyof T]: T[K] extends S ? K : never;
}[keyof T];

// or shortened

type KeySearch<T extends object, S> = {
  [K in keyof T]: T[K] extends S ? K : never;
}[keyof T];
````

But on the plus side, this doesn't seem to need an `any` cast anymore either.

[TS-Playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwKZQEoEMwBM4C2AkmFABQExgBciYIBARqgE4A0iBWAHrfU6wCUiAN4BYAFCJOVRAF5EAWSxQAFgDoIqGABsKVQQG5J0rt3lKVG4DrhwWFHkZOIW6ECyTK16m3YfeGiw4+ARkwgBUiI7mALQySADUiACMgsLJlGDGEgC+kpKgkLAIiFgswQCeAKI6qASopAA8ACoWAM5QLFTIAHxk5VW0GKhY+GA6lQCCFViVrb2CtG3iUogQCJ2IVLio5gqDc+p1YMhqiL2piAD8KOjYeIQk5AAMHIeVx41nqojxaYhaC8ctI3FAPEgPgBtHZ7AC6OXyEkkjQYiAAwnBbCwAMqiFwjXAWADkhOJbBcAHE3I0SdTUI1yS4AEI6ECoEms9lMvIo-gYrH2ABy+LWhIsKReUreVJpSAUkulLxSFLWXI5CqVACZVUjUQQBdjFKLpOLNdLVdJ6bSFMTrWAedJ1Zy2ageUjJFBKgAHDkAaVQlRxoxYEFUTRcbT2UEauHaiBGGxYuCanW6p0QAB86AxmOxEGmelmcwIWL1LYg8dHY-HCxns-w85JLgpVtIoX7tkgANaBuDARAtOHLDtwxDVvDxvG3Tt8VAAN1YiKhvcq-cHCIKEi9vsQfrxCgDQZDYaaO9Q68x2JxHDrfRy573IsPgeD5VP58vgpYQtvXR6vQPj6-rGi+x7vuGn4Dle9iKH+6b3p6wF7jiKQWEeb6hpBwFfteHCNqwgFIbufpCmhYGYR+OHQd+v4lnmRHbshfqKORe6vie2G+rhsH4bmhE5JIAD0QmVmMkyIGoMDxo++CoO0DpQIgADu9jdmU8ZYK48kgDoSmPjAUDtKgOgDmQEBYCAxnaVg7SbIIxEctU-AAGpYK6LTAU0AAa47cDGk4JqgSYpnexYEfmYUNvxZYti42YiIgHZdux4FYT58EAcOiAAAYACQiN5o65DliC5B2HEQRlBb-qcvRwvFohJZ2sgYZx1URfVtBFX6Y7lW1VXeXxpb1YJTG7gAYnYB6IM5DBuR5XlQYa9g4oxj5TXAz6za57nsp5vpntRK0-utyGbaBO3zXtqAHagR3cTRRqMZIGxgFswDTbQm0zaSqC4MSORvR9dhCt9oMSlKhiICJiBCnASntB4HLKaolSSao0mIHJCnEkprAsPYHBEDZRJ2Q0qOsKgHBSfGym6DoiAIBJnTlPjFT2EW9PnC0eLzgArK9mxKZ9cCKODYuQ9KKRjUU0DwEg+oLeyEZrHNBB3X5AVxkFIWprVyDhTFmX1vRhEVuryu3chE5xi40gJfb0jOyugYpQN6Xq3dJt9Nl+UiF7wGu5UcIlSCzvSP1lWe-w3s1Qh9VO47awR+2q7u9Hp6B76w0Mdl2eoMHm4p87UdpVnsfAbnhENRI-T6gA8owABWwVQLQBdLFdBBW5rbYw6JMDIGA9gcvwrAwBAiCrppuC7ESjDo3dOIQN03pQC4wNKTPFhN630DqDPZCNy3beCL4ugxg4ZCrsIciXNJQpYEKZBCjFN+BukzhrFvrghIQR4LAfFqPURo5AZ7f1BO4Tw45+B7zblCYIjwCBHjHLZbuvdgKIi3L-bgM0lY3TIDBXE39cHbQIa6IhtFSHC0QNwS6FD2RUKNN-IAA)

<details>
<summary>Docs-Preview</summary>

grafik


</details>

File: test/helpers.spec.ts [hunk 24]
```diff
@@ -159,6 +190,94 @@ describe('helpers', () => {
...
+          );
+          expect([
+            ColorValueWithStartIndexEnum.Red,
+            ColorValueWithStartIndexEnum.Green,
+            ColorValueWithStartIndexEnum.Blue,
+          ]).toContain(randomColorVAlue);
+        });
+
+        it('should return a value from a string enum', () => {
+          const randomColorVAlue = faker.helpers.enumValue(ColorStringEnum);
```
ST-DDT: Please fix the typo here and in the other methods:
```suggestion
          const randomColorValue = faker.helpers.enumValue(ColorStringEnum);
```

Alternatively, we sometimes use `actual` as variable name for simplicity reasons.

src/modules/helpers/index.ts
ST-DDT: Never mind, I found an even easier solution:

````ts
  enumValue<
    EnumType extends Record<string | number, string | number>,
    EnumValueType extends {
      [key in keyof EnumType]: EnumType[key] extends string
        ? `${EnumType[key]}`
        : EnumType[key];
    }[keyof EnumType]
  >(enumObject: EnumType): EnumValueType {
````

File: src/modules/helpers/index.ts (hunk 21 above)
Comment (matthewmayer) [inline src/modules/helpers/index.ts]: capitalization: TypeScript

File: test/helpers.spec.ts [hunk 25]
```diff
@@ -159,6 +190,69 @@ describe('helpers', () => {
...
+        });
+
+        it('should return a value from a mixed enum', () => {
+          const actual = faker.helpers.enumValue(FooMixedEnum);
+          expect([
+            FooMixedEnum.Foo,
+            FooMixedEnum.Bar,
+            FooMixedEnum.StrBar,
+            FooMixedEnum.StrFoo,
+          ]).toContain(actual);
```
Shinigami92: To absolutely clarify what the intend of the test is, we should pass the native/raw values to the expected.

```diff
  const actual = faker.helpers.enumValue(ColorValueEnum);
- expect([ColorValueEnum.Red, ColorValueEnum.Green, ColorValueEnum.Blue]).toContain(actual);
+ expect([0, 1, 2]).toContain(actual);
```

That way when reading the test, it is much more clear what the intend of the test is.

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,41 @@ export class HelpersModule {
...
   }
 
+  /**
+   * Returns a random value from an Enum object.
+   *
+   * This does the same as `objectValue` except that it ignores (the values assigned to) the numeric keys added for TypeScript enums.
+   *
+   * @param enumObject Enum to pick the value from.
+   * @param <EnumType> Type of generic enums, automatically inferred by TypeScript.
+   * @param <EnumValueType> Type of enumObject property, also automatically inferred by TypeScript.
```
ST-DDT: We use the `@template` tags for the generic parameters as this has better readability in VSCode.

```suggestion
   * @template EnumType Type of generic enums, automatically inferred by TypeScript.
   * @template EnumValueType Type of enumObject property, also automatically inferred by TypeScript.
   * @param enumObject Enum to pick the value from.
```

ST-DDT (dismissed):
One last change, then this is good to go IMO.

ST-DDT (dismissed):

Shinigami92 (changes requested):
I sadly identified some TypeScript problems: [TS-playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwKZQEoEMwBM4C2AggE4BGMUJWJAngBQExgBciYIBZqJANIgSwAPNhy48AlIgDeAWABQiJYhLoQJJAFksUABYA6ankL0pAKkSNhiALQDmUgNT2wAbgUBfBQtCRYCFHRsYwIASTAoRmZRTm4+AWEY8RIpOUVlJiQAXkRtPX0IVBgAGyiwCXd0pUEhRBy8g2BiuDgSKyEKhWUVNQ1cnUbm1voGwxx8AlNEC3bbF0RnAEYJJxdKr3kfcGh4JBpqWgBRYtQCVAiAHgAVOsQAZypmZAA+ei7lfaxaNgxULHwwMVaKQDtdnu8JGwbmluhAEA9EMxcKhall3t1PrR9CcwMg9IhnohFogAPyBTDjQjhSIABn4mOx5zxujmy0QbBplW6qig6j2JAOAG0kSiALpcxAbLZ+XaIc6cABqWGKIFQF3RiEOYiutAADqg5UIoOdcHdEL84SRcBcHiQnogAD7sWI8fi2+1OsRxZ68DVaxXK1U6-WG414M0w7qIQUAa1QtERSDjtDgwE12r1qFFbH9BGDqFj8dFoZNZvduI13TJAAMACTSXP5wu0UUeauV5Q5jP65vijUeZup9OcfOi96veUEADyZAAVqhoF2R5nIcOCEqVah8zINQB6XeI5BgVoGsQ8GAQRDJs3-ZG4RBkBP5gDKEDtuqgGrhYAR19uM-naB9GvehJwAhcoAkfRgBKY02noZMpCyQkYDuAA5LA0PoNCXTaRCVglJRvwRIwJgAaXjW5MWOU5zkia9OiqHpeT6MC5wgwVSMICiW0QLAzVzDcg0zCUpXkfdbBsGxvHkSdEAAYTgZoSCE1Bcx3dJflwX10gAcVUc4dKUAAhTcdI2OTFOU1SAHVKF0Z8oBoKBwmRIR1MjLTbgAZiMxB9NQQz3lM1VzIUSylNaRy7VxDz3i8nIAHIMEOAARRK-IC85bkS3SUsONCMuCzccuMgAZABVQ4ivkCyxEQAAxFpNBgIRUFwOL0iauBbjpYKaFuRY-Oi7qcoaqcpxqpRouMgakuMogMBqsTiKgPjiV3Cw2Cs1pVPUswDxySdVPoHaVMDNSxAkBRVr4gAmRBNvZBTIvOzc7L0RznNclF9sOuUxBOs7bPsr6SBcvBfqum74TWrBvMerbEAAIhS1LkcdFG8sOAqMadZHyqqjGDtuY6LtO17oqeXNrvkW6sAAFkR57upatqOrEfRRqdVnWva3N9FmkhMeR8apzxlGFowYn-rJzd6F59maZk26yEWbbXr2+qciwRZXEeg84BjGGfzWsg7o16yLo+hynPBn73O1+79Yko2TYRMhvMtqLHlip34Zdg9dRIOAyBOAh3bNhm2EV-n-YZwPEGD0Pw5k3wdgCCBFnoZhdRAKBvbe1UaZkDZ0-8JAIDunOwDzguXqt97QbtiG3JL6Qy+2CvEAgbya7rwuqb9zhUk7mVM4Z-v85j5q+Y5kfS5V7PdakV3jbp6usDu1fDfX3v6Hhnek5DsPThuyfGaP5PT4j+QgA)

So while the runtime code is okay, the actual return type of the function is not real-world usable and therefore lacks the usability for what the function is originally planned for.

Right now I didn't get the types managed myself, hopefully someone of you have more luck.

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,41 @@ export class HelpersModule {
     return arrayCopy.slice(min);
   }
 
+  /**
+   * Returns a random value from an Enum object.
+   *
+   * This does the same as `objectValue` except that it ignores (the values assigned to) the numeric keys added for TypeScript enums.
+   *
+   * @template EnumType Type of generic enums, automatically inferred by TypeScript.
+   * @template EnumValueType Type of enumObject property, also automatically inferred by TypeScript.
```
```suggestion
```

src/modules/helpers/index.ts
```diff
@@ -622,6 +622,41 @@ export class HelpersModule {
...
+      [key in keyof EnumType]: EnumType[key] extends string
+        ? `${EnumType[key]}`
+        : EnumType[key];
+    }[keyof EnumType]
+  >(enumObject: EnumType): EnumValueType {
+    // ignore numeric keys added by TypeScript
+    const keys = Object.keys(enumObject).filter((key) => isNaN(Number(key)));
+    const randomKey = this.arrayElement(keys);
+    return enumObject[randomKey] as EnumValueType;
+  }
```
```suggestion
  enumValue<EnumType extends Record<string | number, string | number>>(
    enumObject: EnumType
  ): EnumType[keyof EnumType] {
    // ignore numeric keys added by TypeScript
    const keys: Array<keyof EnumType> = Object.keys(enumObject).filter((key) =>
      isNaN(Number(key))
    );
    const randomKey = this.arrayElement(keys);
    return enumObject[randomKey];
  }
```

ST-DDT (commented):
Those type errors can be fixed with the following changes.
It slightly reduces the DX for the return types of String enums.

@Shinigami92 Any suggestions on how to modify the tests to catch the type errors (for future refactorings)?

src/modules/helpers/index.ts
Shinigami92: [TS-Playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAUzCAtgNQIYBsTIA8AomugCoCeADsigB5SoAmAzogErIRwBOzhVlF4wwAc0QAfRGQBGyXgBpEQkeKkyM83gD4dACgBQiEyjIB5WQCtuUAFyJSGKrUMBKB04o1kAbQDWyJRwwI5kLsgAuogA3samvMhQILxIaLi4iNjs2GCUANyGAL6GhqgYiADCcLh8OPjIXrHxJlzMii2IAOKJqB2miABCDR0l5ehVNXV4BADqMFAAFgDKUNi8UACSYMzI9E1xA22IALyIAMz9pj3IfZ3DBKNlZJO1vKtqYgedx2cA5BxiAARP5XEw3VCnRB-LqA4gAOVB9waUL+gwAMgBVYhIsYvABicDgAFkYPRkMxvgNCXAoQAGMFDdZQgCMjI+NNR+PM5iRAw+g2Z-0GAEEOLjSoYeGAhFkWYgAPQAKkQDmqb3qBCaSoVUPGmuQ+nV0waXjcUoQsuwACZFSq1VNeAb5ktVustjs9trdWd9TNDcanf6XSs1httrt9mRzdKrec7arXnwPqIvi8dXqyAajY6U+IzRaZVAsgAWBMOGmk8mU9M+swYbOVskUguF2WyFkOjX+ppnbAs-KK3Vwfxt4uya1dk1zBah90Rr0vPvWwcK4ej2Pj85T97CVO9rLnVfrseIWQlitEqstpel4+IEelUCQWAIRAQFn6UTUED2JNB01o1iEpn2geAkAga0vzAH8-0DZ1ZzdcNPSjDA3GAwxQNfCDzmg2CdzzNM0IwrDwPfEs8N-S8SWbGtiJiEopU-ft0LXB8Nygm1WJPCBcOwc5uPYqUKOwEtBMfIA)

I can confirm it fulfills my conditions 👍

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

Shinigami92:
@matthewmayer please also approve or write a new review comment

PR #1921, opened 2023-03-11
feat(locale): add locale TH

munkeawtoast:


docs/guide/localization.md
```diff
@@ -118,6 +118,7 @@ export const customFaker = new Faker({
 | `ru`          | Russian                   | `fakerRU`          |
 | `sk`          | Slovakian                 | `fakerSK`          |
 | `sv`          | Swedish                   | `fakerSV`          |
+| `th`          | TODO: Insert Title for th | `fakerTH`          |
```
matthewmayer: Change to "Thai".

src/locales/th/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import animal from './animal';
+import cell_phone from './cell_phone';
+import color from './color';
+import date from './date';
+import internet from './internet';
+import location from './location';
+import person from './person';
+
+const th: LocaleDefinition = {
+  title: 'TODO: Insert Title for th',
```
matthewmayer: Change to Thai

src/locales/th/person/last_name.ts
```diff
@@ -0,0 +1,172 @@
+export default [
+  'คมคาย',
```
matthewmayer: I know that there aren't really any common Thai surnames, so how did you select these?

src/locales/th/person/index.ts
```diff
@@ -0,0 +1,16 @@
...
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+import type { PersonDefinitions } from '../../..';
+import female_first_name from './female_first_name';
+import last_name from './last_name';
+import male_first_name from './male_first_name';
+
+const person: PersonDefinitions = {
+  female_first_name,
```
matthewmayer, line 12: you should also add a src/locales/th/person/name.ts to give the correct name order for Thai, ie
```
export default [
  { value: '{{person.firstName}} {{person.lastName}}', weight: 1 },
];
```

src/locales/th/cell_phone/formats.ts
```diff
@@ -0,0 +1 @@
+export default ['06 #### ####', '08 #### ####', '09 #### ####'];
```
matthewmayer: this file is deprecated, you can also add src/locales/th/phone_number/formats.ts as well

matthewmayer (changes requested):
ขอบคุณ!

Left a few comments

src/locales/th/person/last_name.ts
munkeawtoast: They are mostly generated from chatGPT. I took time to filter out the unnatural ones.

matthewmayer: that might be problematic, because we need data to be MIT licensed and the output of ChatGPT could contain copyrighted material https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#sourcing-data-for-definitions

An alternative might be to make some fake Thai surnames by taking syllables which commonly appear in Thai names and joining them together?

munkeawtoast: Will try it out!

munkeawtoast:
Oops i renamed the branch and didn't think it would close the PR. Should i make a new PR?

ST-DDT:
If it is easier for you, then yes. But please make sure to fix all comments posted here.

PR #1930, opened 2023-03-14
infra: enable prettier cache

matthewmayer:
Enables prettier cache
https://prettier.io/docs/en/cli.html#--cache

This makes `pnpm run format` orders of magnitude faster after the first run


matthewmayer:
before
`pnpm run format  52.98s user 5.10s system 135% cpu 42.996 total`

after
`pnpm run format  6.97s user 0.93s system 114% cpu 6.894 total`

ejcheng (approved):

xDivisionByZerox (approved):

ST-DDT (approved):
I don't see a drawback here. Where does it write the cache too though?

matthewmayer:
> I don't see a drawback here. Where does it write the cache too though?

By default, Prettier CLI saves the cache file for the --cache option at ./node_modules/.cache/prettier/.prettier-cache

PR #1935, opened 2023-03-15
chore: remove locale definitions which are exact duplicates of en

matthewmayer:
Removes definitions in other locales which are exact duplicates of the `en` locale. This reduces duplication from copy-paste in definition files.

I identified these by running this script:
```js
const fs = require('fs');
const { allLocales } = require('@faker-js/faker');
for (let locale of Object.keys(allLocales).filter((l) => l !== 'en')) {
  let definitions = allLocales[locale];
  for (let module of Object.keys(definitions)) {
    let keys = Object.keys(definitions[module]);
    for (let key of keys) {
      let enValues = allLocales.en[module][key];
      let trValues = allLocales[locale][module][key];
      if (
        enValues &&
        Array.isArray(enValues) &&
        enValues.length === trValues.length &&
        enValues.every((v, i) => v === trValues[i])
      ) {
        console.log(`${locale}.${module}.${key} is dupe of en`);
        fs.unlinkSync(`./src/locales/${locale}/${module}/${key}.ts`);
      }
    }
  }
}
```

then
`pnpm run generate:locales`


ST-DDT:
This would cause issues in combination with:

https://github.com/faker-js/faker/pull/1748/files#diff-fd2caf3156b2fe9a28044b99d27ff58a700ff30462debc02c68f5fc45fe8bc48R128-R131

ST-DDT:
**Team Decision**

We want to remove the `en` locale from the other locales fallbacks, thus this has to take the changed fallbacks into account.
Also if we remove the data we should check our locale data recursively.

matthewmayer:
While I can see the logic in the long run, removing the en fallback warrants further discussion

I imagine at the moment when a new dev starts using faker, they use the default en locale. They write some code using a few methods. Then they realise they want say localised addresses or phone numbers. They look up into the documentation how to do that. Ok easy, replace faker with fakerDE. 

However if the en fallback is removed that means other methods for which there is no de data would now start throwing errors or returning undefined. So now I have to make a custom faker instance which is more complex. 

I think a nicer DX beats the conceptual purity of de not extending en.

ST-DDT:
We will have a dedicated issue for the en removal once @Shinigami92's #1748 is merged. Maybe we can create the issue now and reference it from the TODO comment. We can discuss this in more detail in that issue, the gist behind that idea is that returning erronous data is worse then throwing an error and missing data can be added yourself and will hopefully be completed over time. For de vs en this might be less relevant, but zh vs en has huge differences.

matthewmayer:
> We will have a dedicated issue for the en removal once @Shinigami92's #1748 is merged. Maybe we can create the issue now and reference it from the TODO comment. We can discuss this in more detail in that issue, the gist behind that idea is that returning erronous data is worse then throwing an error and missing data can be added yourself and will hopefully be completed over time. For de vs en this might be less relevant, but zh vs en has huge differences.

agree this needs discussing but this is not the place to discuss it :)

PR #1938, opened 2023-03-15
fix(random): prevent infinite do-while

Shinigami92:
Extracted from
- https://github.com/faker-js/faker/pull/1748#discussion_r1128083370

### DIFF
--- a/src/modules/random/index.ts
+++ b/src/modules/random/index.ts
@@ -125,6 +125,8 @@ export class RandomModule {
     ];
     let result: string;
 
+    let iteration = 0;
+
     do {
       // randomly pick from the many faker methods that can generate words
       const randomWordMethod = this.faker.helpers.arrayElement(wordMethods);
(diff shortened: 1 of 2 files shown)


xDivisionByZerox (changes requested):
Could we get a test that verifies this behavior?

Something like this?
```ts
it('fails if no data are avalible', () => {
  const faker = new Faker();
  faker.definitions = {};

  expect(() => faker.random.word()).toThrowError();
});
```

src/modules/random/index.ts
```diff
@@ -133,6 +135,14 @@ export class RandomModule {
         result = randomWordMethod();
       } catch {
         // catch missing locale data potentially required by randomWordMethod
+        iteration++;
+
+        if (iteration > 100) {
+          throw new FakerError(
```
matthewmayer, line 141: Can you add a test

matthewmayer (dismissed):

xDivisionByZerox (approved):

ejcheng (approved):

ST-DDT (approved):

PR #1939, opened 2023-03-15
chore(random): deprecate random module

ST-DDT:
We have dedicated modules for pseudo words and actual words, so we should redirect the users to use them instead.


ejcheng (dismissed):

src/modules/random/index.ts
```diff
@@ -27,14 +27,26 @@ export class RandomModule {
...
+   *
+   * @see faker.lorem.word()
+   * @see faker.word.sample()
    *
    * @example
    * faker.random.word() // 'Seamless'
    *
    * @since 3.1.0
+   *
+   * @deprecated Use `faker.lorem.word()` or `faker.word.sample()` instead.
```
Shinigami92, line 44: Now every single method in this module is deprecated, so we should also annotate the whole module with `@deprecated`

ST-DDT: Done

ST-DDT:
For some reason, `pnpm run preflight` doesn't fail on lint issues.

ST-DDT:
Is `chore` the correct type? Because chores don't show up in the changelog.

src/faker.ts
```diff
@@ -64,6 +64,10 @@ export class Faker {
   /** @internal */
   private readonly _mersenne: Mersenne = mersenne();
 
+  /**
+   * @deprecated  Use the modules specific to the type of data you want to generate instead.
```
```suggestion
   * @deprecated Use the modules specific to the type of data you want to generate instead.
```

Shinigami92 (approved):

ejcheng (approved):

PR #1945, opened 2023-03-17
chore(location): Deprecate faker.location.county and make alias for faker.location.state

### LINKED ISSUE #1944: Clarify the difference between `faker.location.county` and `faker.location.state`
The top-level administrative regions of a country can be called states (eg. in USA), provinces (e.g. in Canada) or counties (e.g. in UK), or various other names.

Currently there are seperate methods called `faker.location.state()` and `faker.location.county()`

It's clearer to have a single method for this.
The following locales have county definitions

* en, but it's a truncated version of the en_GB list
* el -> rename to state
* en_GB -> rename to state
* en_IE -> rename to state
* he -> contains a comment "Not relevant for Israel"
* ro, but it's a duplicate of the state list

matthewmayer:
fix #1944


ST-DDT:
**Team Decision**

We want to keep `location.county()`.
We should however improve the documentation, that this usually refers to  administrative areas within a state:

Country | State | County | City
--- | -- | --- | ---
USA | Alabama | Baldwin |  Daphne
DE | NRW | Städteregion Aachen | Würselen

Issue #1907, opened 2023-03-07
de_CH street names don't sound very Swiss German


matthewmayer:
i did notice that although #1524 is now fixed, the de_CH street names don't sound very German.

```
> fakerDE.location.streetAddress()
'Wilhelm-Leuschner-Str. 42b'
> fakerDE_CH.location.streetAddress()
'Jäggi Crossroad 3'
```

This is because the de_CH street pattern uses name+street_suffix, but street_suffix falls back to English 
https://github.com/faker-js/faker/blob/next/src/locales/de_CH/location/street.ts

Whereas de street patterns use a fixed list of street_names
https://github.com/faker-js/faker/blob/next/src/locales/de/location/street.ts
https://github.com/faker-js/faker/blob/next/src/locales/de/location/street_name.ts

FYI @Slooti

There are a few solutions:
* Delete https://github.com/faker-js/faker/blob/next/src/locales/de_CH/location/street.ts and let it fall back to the generic `de` list?
* add some de or de_CH street_suffixes like Strasse?
* provide a street_name list for de_CH if they are significantly different to de?

Slooti:
That is a very interesting point! There are a few subtile differences.

• Germans still often use the old eszet: https://en.wikipedia.org/wiki/%C3%9F . Whilst less and less common, “Straẞe” is used and needs to be tested. In Switzerland that is never used. They use “Strasse”.
• Aside “strasse” und “weg” as suffix, Germany and Switzerland has other suffixes that are very distinct.
• The streets have a meaning. The Germans name there streets after different things than Swiss people. Of course there are overlaps, but for most of the street names, you immediately spot that they are Swiss/German.

These are reasons why a dedicated solution for Switzerland would be slick. I would probabely  go for a dedicated street_name, as it also probabely is the most easy for PR and let peope create easy forks for testing their Swiss edge cases.

Kind regards

Martin




Martin

xDivisionByZerox:
Fixed by #1946.

PR #1946, opened 2023-03-18
chore(locale): added dedicated swiss street names

asharnadeem:
fix for #1907 

adds a file for `de_CH`, street_name, that uses a dedicated list of street names that are common to Switzerland. all streets are pulled from state directories and actually exist


ST-DDT (approved):

Shinigami92 (approved):

matthewmayer (approved):

ejcheng (approved):

xDivisionByZerox (approved):
FFR: Faker does not care about whether generated values exist. Many of the provided street names share a common suffix like `weg`, `strasse` or `gasse`. This could potentially be extracted in the future so that the street patterns could include "fake" names.

PR #1947, opened 2023-03-18
infra: try to fix e2e flakyness

### LINKED ISSUE #1943: E2E doc test is very flaky
It gets stuck at least 25% of the time. Any idea why?
example:
https://github.com/faker-js/faker/actions/runs/4444946074/jobs/7803608958?pr=1942

ST-DDT:
Tries to fix:

- #1943 

 Improvements:

- Make HEAD requests instead of GET
- (Potentially) avoid retries due to non 2xx response codes
- Check every url only once (ignoring #part) 270 requests (one per method) -> 15(?) (one per module)
- Don't search the entire page for module methods, only search that module section (bug)

### DIFF
--- a/cypress/e2e/api.cy.ts
+++ b/cypress/e2e/api.cy.ts
@@ -28,19 +28,29 @@ describe('API Test', () => {
     });
 
     it('should not have dead links', () => {
+      const checked = new Set<string>();
       cy.get('.api-group li').each(($el) => {
-        const text = $el.find('a').text();
-        const link = $el.find('a').attr('href');
+        const anchor = $el.find('a');
+        const text = anchor.text();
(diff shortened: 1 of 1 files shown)


ST-DDT:
Failure

ST-DDT:
Especially the last commit will speed up the test by ~30 seconds in CI, so we should definitely merge it eventually.
I'll try to run the CI multiple times maybe I can explore whether either of them actually fixes the issues.

ST-DDT:
I ran both commits 5 times each and didn't have a single failure.

Shinigami92 (approved):

matthewmayer (approved):

xDivisionByZerox (approved):

PR #1958, opened 2023-03-21
docs(color): fix color rgb docs

### LINKED ISSUE #1957: docs are confusing about default parameters of faker.color.rgb
https://next.fakerjs.dev/api/color.html#rgb

options.prefix shows default `0x`

examples show both
```
faker.color.rgb() // => "#cdfcdc"
faker.color.rgb() // '0xffffFF'
```

In the code it actually seems to default to "#"

```
const {
      format = 'hex',
      includeAlpha = false,
      prefix = '#',
      casing = 'lower',
    } = options || {};
```

either
- change the docs to show "#" is default?
- change the code to show "0x" is default?

Shinigami92:
close #1957

### DIFF
--- a/src/modules/color/index.ts
+++ b/src/modules/color/index.ts
@@ -231,7 +231,7 @@ export class ColorModule {
    * Returns an RGB color.
    *
    * @example
-   * faker.color.rgb() // '0xffffFF'
+   * faker.color.rgb() // '#ffffFF'
    *
    * @since 7.0.0
    */
@@ -240,16 +240,16 @@ export class ColorModule {
    * Returns an RGB color.
    *
    * @param options Options object.
-   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `'0x'`.
+   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `'#'`.
    * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'mixed'`.
    * @param options.format Format of generated RGB color. Defaults to `hex`.
    * @param options.includeAlpha Adds an alpha value to the color (RGBA). Defaults to `false`.
    *
    * @example
-   * faker.color.rgb() // '0xffffFF'
-   * faker.color.rgb({ prefix: '#' }) // '#ffffFF'
-   * faker.color.rgb({ casing: 'upper' }) // '0xFFFFFF'
-   * faker.color.rgb({ casing: 'lower' }) // '0xffffff'
+   * faker.color.rgb() // '#ffffFF'
+   * faker.color.rgb({ prefix: '0x' }) // '0xffffFF'
+   * faker.color.rgb({ casing: 'upper' }) // '#FFFFFF'
+   * faker.color.rgb({ casing: 'lower' }) // '#ffffff'
    * faker.color.rgb({ prefix: '#', casing: 'lower' }) // '#ffffff'
    * faker.color.rgb({ format: 'hex', casing: 'lower' }) // '#ffffff'
    * faker.color.rgb({ format: 'css' }) // 'rgb(255, 0, 0)'
@@ -261,7 +261,7 @@ export class ColorModule {
     /**
      * Prefix of the generated hex color. Only applied when 'hex' format is used.
      *
-     * @default '0x'
+     * @default '#'
      */
     prefix?: string;
     /**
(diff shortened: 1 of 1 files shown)


src/modules/color/index.ts
```diff
@@ -315,16 +315,16 @@ export class ColorModule {
    * Returns an RGB color.
    *
    * @param options Options object.
-   * @param options.prefix Prefix of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'0x'`.
+   * @param options.prefix Prefix of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'#'`.
    * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'mixed'`.
```
matthewmayer: This also seems wrong. It defaults to lower

src/modules/color/index.ts
```diff
@@ -240,16 +240,16 @@ export class ColorModule {
...
    * @param options.format Format of generated RGB color. Defaults to `hex`.
    * @param options.includeAlpha Adds an alpha value to the color (RGBA). Defaults to `false`.
    *
    * @example
-   * faker.color.rgb() // '0xffffFF'
-   * faker.color.rgb({ prefix: '#' }) // '#ffffFF'
-   * faker.color.rgb({ casing: 'upper' }) // '0xFFFFFF'
-   * faker.color.rgb({ casing: 'lower' }) // '0xffffff'
+   * faker.color.rgb() // '#ffffFF'
```
matthewmayer: Shall we make it not ffffff? Doesn't feel very random to always get white!

Also lowercase per other comment

matthewmayer (changes requested):

ST-DDT (approved):

xDivisionByZerox (approved):

src/modules/color/index.ts
```diff
@@ -240,34 +240,34 @@ export class ColorModule {
...
-   * faker.color.rgb({ format: 'css' }) // 'rgb(255, 0, 0)'
-   * faker.color.rgb({ format: 'binary' }) // '10000000 00000000 11111111'
+   * faker.color.rgb() // '#8be4ab'
+   * faker.color.rgb({ prefix: '0x' }) // '0x9ddc8b'
+   * faker.color.rgb({ casing: 'upper' }) // '#B8A51E'
+   * faker.color.rgb({ casing: 'lower' }) // '#b12f8b'
+   * faker.color.rgb({ prefix: '#', casing: 'lower' }) // '#eb0c16'
+   * faker.color.rgb({ format: 'hex', casing: 'lower' }) // '#bb9d17'
+   * faker.color.rgb({ format: 'css' }) // 'rgb(216, 17, 192)'
+   * faker.color.rgb({ format: 'binary' }) // '00110010 00001000 01110110'
```
matthewmayer, line 256: can we add some `includeAlpha` examples too?

src/modules/color/index.ts
```diff
@@ -240,34 +240,34 @@ export class ColorModule {
    * Returns an RGB color.
    *
    * @param options Options object.
-   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `'0x'`.
-   * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'mixed'`.
+   * @param options.prefix Prefix of the generated hex color. Only applied when 'hex' format is used. Defaults to `'#'`.
+   * @param options.casing Letter type case of the generated hex color. Only applied when `'hex'` format is used. Defaults to `'lower'`.
```
matthewmayer, line 244: can we explicitly list what the possible values of `casing` and `format` are? (this is similar to issue we had in #1732 where we said we'd list out values of types manually until we are able to get the automatic extraction working)

image

matthewmayer (dismissed):

src/modules/color/index.ts
Shinigami92: This PR just fix the current docs
If we want to add something to the docs additionally, I suggest opening a separate PR

src/modules/color/index.ts
Shinigami92: Somehow `options.sex` for `fullName` can also be correctly extracted: https://next.fakerjs.dev/api/person.html#fullname

So maybe @ST-DDT as the author of the docs gen code need to handle this in a separate PR

ST-DDT: I'll have a look soon.

ST-DDT: I created a PR for that:

- https://github.com/faker-js/faker/pull/1965

But it isn't that simple, so I need some feedback.

src/modules/color/index.ts
ST-DDT: Done: https://github.com/faker-js/faker/issues/1967

PR #1940, opened 2023-03-16
docs: Faker and utility pages

ST-DDT:
This PR adds documentation pages for the `Faker` class and standalone utility methods.

Netlify-Preview: https://deploy-preview-2--serene-sprite-f3ef50.netlify.app/

PR #1960, opened 2023-03-22
docs: extend Faker class jsdocs

ST-DDT:
Extracted from #1940 to make it easier to review.


src/faker.ts
```diff
@@ -33,6 +33,29 @@ import { VehicleModule } from './modules/vehicle';
...
+ *
+ * // faker.seed(1234);
+ *
+ * faker.person.firstName(); // 'John'
+ * faker.person.lastName(); // 'Doe'
+ *
+ * @example
+ * import { Faker, de } from '@fakerjs/faker';
+ * // const { Faker, de } = require('@fakerjs/faker');
+ *
```
matthewmayer, line 53: // create a Faker instance with only `de` data and no `en` fallback

src/faker.ts
```diff
@@ -33,6 +33,29 @@ import { VehicleModule } from './modules/vehicle';
...
+ * const customFaker = new Faker({ locale: [de] });
+ *
+ * customFaker.person.firstName(); // 'Max'
+ * customFaker.person.lastName(); // 'Baumeister'
+ */
```
matthewmayer, line 61:
```
customFaker.music.genre() // throws Error as this data is not available in `de`
```

src/faker.ts
```diff
@@ -50,6 +73,15 @@ export class Faker {
...
    * The function must return a new valid `Date` instance for every call.
    * Defaults to `() => new Date()`.
+   *
+   * @see [Reproducible Results](https://next.fakerjs.dev/guide/usage.html#reproducible-results)
+   * @see faker.seed() for reproducible results.
+   *
+   * @example
+   * faker.seed(1234);
+   * faker.date.past(); // Changes based on the current date
+   * faker.setDefaultRefDate(() => new Date('2020-01-01'));
```
matthewmayer: there's no real need for this to be a function right? Could just be a fixed value. Maybe if we do a function example it could be something actually dynamic?

matthewmayer (changes requested):

src/faker.ts
xDivisionByZerox: I had in mind to review then entire german locale during the next month('s) since I'm from Germany and use the locale frequently.

Could we maybe choose another locale if we want to include a failing example?

src/faker.ts
```diff
@@ -151,12 +205,56 @@ export class Faker {
...
+           * The locale data to use for this instance.
+           * If an array is provided, the first locale that has a definition for a given property will be used.
+           *
+           * @see mergeLocales
+           */
+          locale: LocaleDefinition | LocaleDefinition[];
+        }
+      | {
+          /**
+           * DEPRECATED: The locale data to use for this instance.
```
xDivisionByZerox: Why `DEPRECATED` instead of the commonly use `@deprecated`?

src/faker.ts
matthewmayer: es has names but no music

src/faker.ts
ST-DDT: So that it will be shown in the api docs page.

I will add the `@deprecated` anyway.

xDivisionByZerox: Ah I see the warning card (like [here](https://fakerjs.dev/api/fake.html#fake)) only shows up if an entire method is deprecated (?). Gotcha 👍

src/faker.ts
ST-DDT: Done

src/faker.ts
ST-DDT: Done

src/faker.ts
ST-DDT: Theoretically, I could check if the property is deprecated and add the `DEPRECATED` automatically.
Should I?

xDivisionByZerox: > Theoretically, I could check if the property is deprecated and add the DEPRECATED automatically.

You are referring to documentation only. Yes?
If so that sounds good, but would be overkill for now. At least I think so.

xDivisionByZerox (dismissed):

src/faker.ts
```diff
@@ -33,6 +33,32 @@ import { VehicleModule } from './modules/vehicle';
 import { WordModule } from './modules/word';
 import { mergeLocales } from './utils/merge-locales';
 
+/**
+ * The main Faker class containing all modules that can be used to generate data.
```
Shinigami92: This doesn't sound correct :thinking: 

> The main Faker class containing all modules that can be used to generate data.

Maybe something like:

This is Faker's main class that contains all the modules that can be used to generate data.

ST-DDT: Like this?

```suggestion
 * The is Faker's main class containing all modules that can be used to generate data.
```

Shinigami92 (dismissed):

src/faker.ts
ST-DDT: Done anyway: https://github.com/faker-js/faker/pull/1962

matthewmayer (approved):

Shinigami92 (approved):

PR #1961, opened 2023-03-22
docs(internet): password doesn't create crypto-secure PWs

deltamualpha:
Since this API uses faker.number, which itself is just using a Mersenne Twister implementation, these passwords are not safe for use in cryptographically-sensitive contexts.

### DIFF
--- a/src/modules/internet/index.ts
+++ b/src/modules/internet/index.ts
@@ -1293,7 +1293,8 @@ export class InternetModule {
   }
 
   /**
-   * Generates a random password.
+   * Generates a random password. Since the source of the randomness is not cryptographically secure,
+   * neither is this generator.
    *
    * @param options An options object. Defaults to `{}`.
    * @param options.length The length of the password to generate. Defaults to `15`.
--- a/src/modules/internet/index.ts
+++ b/src/modules/internet/index.ts
@@ -1293,8 +1293,8 @@ export class InternetModule {
   }
 
   /**
-   * Generates a random password. Since the source of the randomness is not cryptographically secure,
-   * neither is this generator.
+   * Generates a random password-like string. Do not use this method for generating actual passwords for users.
+   * Since the source of the randomness is not cryptographically secure, neither is this generator.
    *
    * @param options An options object. Defaults to `{}`.
    * @param options.length The length of the password to generate. Defaults to `15`.


ST-DDT (approved):

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

src/modules/internet/index.ts
```diff
@@ -1293,7 +1293,8 @@ export class InternetModule {
   }
 
   /**
-   * Generates a random password.
+   * Generates a random password. Since the source of the randomness is not cryptographically secure,
```
matthewmayer: I think we could even be a bit stronger?

Generates a random password. Do not use this method for generating real user passwords. Since the source of the randomness is not cryptographically secure, neither is this generator.

deltamualpha: Would be happy to add stronger language; just didn't want to come in guns blazing on this, since most doc strings in the API tend toward the terse.

matthewmayer: We are working on more informative docs, i think these kind of usage notes are particularly helpful, i'll let the maintainers pipe in.

ST-DDT: I think adding that additional sentence is more important than the other one. Thanks for bringing this to my/our attention.
If you are knowlegdable enough to understand what cryptographically secure passwords are, then you probably wouldnt use the method in the first place.
However, if you dont then you might not know what to do with that information alone.

matthewmayer: Agreed, we should flag up "footguns" (ie features that might allow a novice developer to shoot themselves in the foot)

So for example using this for generating real passwords, or sending real emails to the output of faker.internet.email() etc

ST-DDT (changes requested):
Please change the description as mentioned above.

src/modules/internet/index.ts
deltamualpha: I've updated the description to both note that the passwords should not be used, and described the generated output as a "password-like string" as opposed to a "password", which also should drive home the point.

ST-DDT (approved):

xDivisionByZerox (approved):

Shinigami92 (approved):

Issue #1919, opened 2023-03-11
Refactor email generation properties


ST-DDT:
The email address generation function parameters are wrongly named and are lacking in some regards.

- `allowSpecialCharacters` is actually `forceSpecialCharacters`
- A list of allowed special characters might be more useful than a boolean flag for the end user.
- ...

This issue is based on a discussion the team had when talking about:

- https://github.com/faker-js/faker/pull/1876

xDivisionByZerox:
Can this be done without a breaking change? 
Do we introduce the `forceSpecialCharacters` option in v8.0 while deprecating `allowSpecialCharacters`?

matthewmayer:
Depends on your definition of breaking change. 😀

You could argue that removing the parameter wouldn't be breaking, the docs say it "allows" special characters, not that it guarantees special characters.

xDivisionByZerox:
So what you saying is that this is more of a `fix` than a `refactor`? Or did I get you wrong?

matthewmayer:
I'm saying i'm pretty sure that anyone who turned this boolean on will quickly regret it as its going to generate email addresses which will fail most validators in an unpredictable way. I'd be tempted just to deprecate it (without replacement) in v8, maybe point the deprecation message to this issue, and see if anyone complains that thiis is important for their workflow?

ST-DDT:
**Team Decision**

We don't need this for v8.0.
(We will likely immediately start with v9 after v8.0)

xDivisionByZerox:
This issue was disregarded for v8 and not touched been touched during the active development of v9. Since there has been no complain from users for 2 years, I'm going to close this issue.

PR #1962, opened 2023-03-22
docs: mark deprecated parameter options

ST-DDT:
Created based on: https://github.com/faker-js/faker/pull/1960#discussion_r1144596311 and might be useful for #1919 as well.

- #1960
- #1919

Since we have switched from multiple parameters to a single options params, it is likely that we get deprecated options in the future.
This PR adds a `DEPRECATED: ` prefix to deprecated parameter options to increase visibility of these parameters in the docs..

### DIFF
--- a/scripts/apidoc/typedoc.ts
+++ b/scripts/apidoc/typedoc.ts
@@ -247,7 +247,7 @@ export function joinTagParts(parts?: CommentDisplayPart[]): string | undefined {
  *
  * @param reflection The reflection to check.
  *
- * @returns The message explaining the deprecation message if deprecated, otherwise `undefined`.
+ * @returns The message explaining the deprecation if deprecated, otherwise `undefined`.
  */
 export function extractDeprecated(reflection?: Reflection): string | undefined {
   const deprecated = extractTagContent('@deprecated', reflection).join().trim();
--- a/scripts/apidoc/typedoc.ts
+++ b/scripts/apidoc/typedoc.ts
@@ -185,23 +185,23 @@ export function extractSourcePath(reflection: Reflection): string {
  * Extracts the text (md) from a jsdoc tag.
  *
  * @param tag The tag to extract the text from.
- * @param signature The signature to extract the text from.
+ * @param reflection The reflection to extract the text from.
  */
 export function extractTagContent(
   tag: `@${string}`,
-  signature?: SignatureReflection,
+  reflection?: Reflection,
   tagProcessor: (tag: CommentTag) => string[] = joinTagContent
 ): string[] {
-  return signature?.comment?.getTags(tag).flatMap(tagProcessor) ?? [];
+  return reflection?.comment?.getTags(tag).flatMap(tagProcessor) ?? [];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


scripts/apidoc/typedoc.ts
```diff
@@ -185,34 +185,34 @@ export function extractSourcePath(reflection: Reflection): string {
  * Extracts the text (md) from a jsdoc tag.
  *
  * @param tag The tag to extract the text from.
- * @param signature The signature to extract the text from.
+ * @param reflection The reflection to extract the text from.
```
xDivisionByZerox, line 188: Could you quickly explain all those `signature` => `reflection` changes?

test/scripts/apidoc/signature.example.ts
```diff
@@ -261,6 +261,34 @@ export class SignatureTest {
     return 0;
   }
 
+  /**
+   * Test with deprecated option.
+   *
+   * @param option The options.
+   * @param option.a Some deprecated option.
+   * @param option.b Some other options.
```
xDivisionByZerox: `option.c` should also be in here, no?

test/scripts/apidoc/__snapshots__/signature.spec.ts.snap
```diff
@@ -206,6 +207,57 @@ exports[`signature > analyzeSignature() > methodWithDeprecated 1`] = `
...
+    {
+      "default": undefined,
+      "description": "<p>The options.</p>
+",
+      "name": "option",
+      "type": "{ ... }",
+    },
+    {
+      "default": undefined,
+      "description": "<p>DEPRECATED: Some deprecated option.</p>
```
xDivisionByZerox: Could replace this with:
```html
<p>{text_from_description}</p>
<p>DEPRECATED: {text_from_deprecation}</p>
```

Would that make sense?

scripts/apidoc/typedoc.ts
ST-DDT: `Reflection` is the parent class of `Signature(Reflection)` and the `comment` property is defined on `Reflection`.

test/scripts/apidoc/signature.example.ts
ST-DDT:
```suggestion
   * @param option.b Some other deprecated option.
   * @param option.c Some other option.
```

scripts/apidoc/typedoc.ts
xDivisionByZerox: Ok so basically downgrading to the minimal requirements. I like 👍

ST-DDT:
Here a (faked) visual example:

grafik

xDivisionByZerox (dismissed):

matthewmayer:
How about also adding some strikethrough on the parameter name?
image

ST-DDT:
> How about also adding some strikethrough on the parameter name? image

Done

xDivisionByZerox (approved):

matthewmayer (approved):

Shinigami92 (approved):

PR #1966, opened 2023-03-23
fix: revert types field

Shinigami92:
This reverts commit 956a15fff5c3714485e8e612df62034a5f78c0c5.

This had broken the whole package!

---

@bluwy

- If we remove the field it is not working
- If we change the field to something else it is also not working
- Only `"types": "index.d.ts"` is working here

Do you know anything about that? Why did publint reported this as invalid? Or even: what is going on here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "license": "MIT",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.mjs",
-  "types": "dist/types/index.d.ts",
+  "types": "index.d.ts",
   "typesVersions": {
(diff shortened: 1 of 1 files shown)


bluwy:
I suppose it was working because https://github.com/faker-js/faker/pull/1878#issuecomment-1446077911 was true. That the `typesVersion` config is remapping it. I haven't took that into account when linting 😬 Good to know though that this doesn't work.

xDivisionByZerox (approved):

ST-DDT (approved):

PR #1969, opened 2023-03-23
test: separate jsdoc tag tests

Shinigami92:
In the history we had some (mostly newbie) contributors but also myself that got confused by the `examplesAndDeprecations` test.

This PR aims for separating the verification of each JSDoc tag into its own section, so it can faster be identified which problem occurred.

Also this improves the test in that way that for example when a `@param` fails, a `@since` tag will still get validated in the same test run.


test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -1,3 +1,5 @@
+// TODO christopher 2023-03-23: Rename file to verify-jsdoc-tags.spec.ts
```
Shinigami92: Before merging this PR, I would like to rename the file to `verify-jsdoc-tags.spec.ts` to make the intend of the test more clear.
I did not renamed it directly, so the diff does not get confused because of the changes inside the file.

Shinigami92:
Right now the loop should be swapped

yet it is: `... > verify JSDoc tags > CompanyModule > verify @deprecated tag > suffixes`

it should be one of:
1. `... > verify JSDoc tags > CompanyModule > suffixes > verify @deprecated tag`
2. `... > verify JSDoc tags > verify @deprecated tag > CompanyModule > suffixes`

test/scripts/apidoc/examplesAndDeprecations.spec.ts
```diff
@@ -51,48 +53,77 @@ describe('examples and deprecations', () => {
...
+            )} } from '../../../../../src';\n\n${examples}`
+          );
+
+          // Run the examples
+          await import(path);
+        }
+      );
+    });
+
+    describe('verify @deprecated tag', () => {
```
ST-DDT:
```suggestion
    describe('verify @deprecated tag and @examples', () => {
```

Shinigami92: No, I want (try) to achieve that every tag is independent from each other
The `@example` tag is checked in the section above

I'm currently try to rewrite the test so I don't need the `@example` directly but just execute the code. (But I assume I just need the examples for that indirectly)

Shinigami92: Now it might be more clear what is happening
I extracted the write file process out of the tests
What I might need to do here is maybe wrap the `await import(path);` in the example test with a "should not throw" whereas the deprecation test is just running them to execute the potential `console.warn`

ST-DDT (commented):
Maybe add a `beforeAll(deleteTemp);`

ST-DDT (dismissed):

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+ * - has working examples
+ * - and running these do not log anything, unless the method is deprecated
+ */
+
+beforeAll(initMarkdownRenderer);
+
+afterAll(() => {
+  // Remove temp folder
+
+  const tempDir = resolve(__dirname, 'temp');
```
xDivisionByZerox: Can we store this in a variable and reuse this in `resolveDirToModules`?

matthewmayer (dismissed):

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+          const dir = resolveDirToModule(moduleName);
+          mkdirSync(dir, { recursive: true });
+
+          const path = resolvePathToMethodFile(moduleName, methodName);
+          const imports = [...new Set(examples.match(/faker[^\.]*(?=\.)/g))];
+          writeFileSync(
+            path,
+            `import { ${imports.join(
+              ', '
+            )} } from '../../../../../src';\n\n${examples}`
```
xDivisionByZerox, line 67: This import seems kinda magical here. Could we resolve this with a dynamic, "lazy" import?

I'm on phone, so writing code examples  is kinda hard, sorry.

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+            signature,
+            moduleName,
+            methodName
+          ).parameters.forEach((param) => {
+            const { name, description } = param;
+            const plainDescription = description.replace(/<[^>]+>/g, '').trim();
+            expect(
+              plainDescription,
+              `Expect param ${name} to have a description`
+            ).not.toBe('Missing');
```
xDivisionByZerox: Is `'Missing'` the default value of typedoc? If so, can we somehow decouple that or add least add a comment?

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+          });
+        });
+
+        it('verify @see tag', () => {
```
xDivisionByZerox: How does vitest handle `it` without `expect`? I couldn't find anything in their docs. I know that this test would fail in jest for methods with no see tag, since no expectation would be generated.

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+              );
+              expect(link, 'Expect method reference to contain ()').toContain(
+                ')'
+              );
+            }
+          });
+        });
+
+        it('verify @since tag', () => {
+          expect(extractSince(signature), '@since to be present').toBeTruthy();
```
xDivisionByZerox: We can (and should) verify for a semver here, no?

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
Shinigami92: I assume this is related to https://github.com/faker-js/faker/blob/8516bfb109741231a49ec8572176e5a12b4c1a37/scripts/apidoc/signature.ts#L44

Shinigami92: @ST-DDT correct me if I'm wrong

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
Shinigami92: For now, no

This is just the string content of what will be written into a temp file
It would only break if we move the temp file to another location
But as the temp folder will get cleaned after run anyways, I dont care that much of its content (or at least if it is pretty or not)

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
Shinigami92: Right now it works :shrug: 
The check itself might get improved in another PR, but here I just took the already working test-code

ST-DDT:
```suggestion
        it('verify @see tags', () => {
```

ST-DDT (dismissed):

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
xDivisionByZerox: I see. Again, I reviewed this on my phone, so getting everything right was kinda hard there.
Still, a variable named `defaultSummary` exported from the `signature.ts` file could be a good idea to make the assumption less magical.

Shinigami92: @ST-DDT whats your opinion about that?

xDivisionByZerox: This would not be blocking from my side, tho.

ST-DDT: The generated default is `Missing`, but it actually gets markdown parsed.
I guess, it doesn't really hurt to have a constant like that.

scripts/apidoc/signature.ts
```diff
@@ -40,8 +40,10 @@ function prettifyMethodName(method: string): string {
   );
 }
 
+export const DEFAULT_SUMMARY = 'Missing';
```
ST-DDT: This is not really a default. It is missing.

ST-DDT (changes requested):

scripts/apidoc/signature.ts
ST-DDT: Is it a summary? Maybe description?

Shinigami92: Before I try out several names, please suggest one so I don't need to push several commits

ST-DDT: `MISSING_DESCRIPTION`

ST-DDT (approved):

xDivisionByZerox (approved):

PR #1978, opened 2023-03-26
refactor(locale)!: move title to metadata

ST-DDT:
Removes the special handling of `faker.definitions.title` by moving it to `faker.definitions.metadata.title`.
As this makes it way easier to (type) safely implement #893.


ST-DDT:
Blocks https://github.com/faker-js/faker/issues/893

- #893

src/locales/af_ZA/metadata/index.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+import type { MetadataDefinitions } from '../../..';
+import title from './title';
+
+const metadata: MetadataDefinitions = {
+  title,
```
Shinigami92: Would it be easy to just inline the string here?
IMO it's a bit to much to generate a whole new file, load it and use its primitive value, and all that just for a string 🤔

src/locales/de_AT/metadata/title.ts
```diff
@@ -0,0 +1 @@
+export default 'German (Austria)';
```
Shinigami92: Just an idea/note maybe for another PR: we could now split the values, because now we have just a separate metadata object for it.

test/faker.spec.ts
```diff
@@ -33,9 +33,9 @@ describe('faker', () => {
   describe('definitions', () => {
     it('locale definition accessability', () => {
       // Metadata
-      expect(faker.definitions.title).toBeDefined();
+      expect(faker.definitions.metadata.title).toBeDefined();
       // Standard modules
-      expect(faker.definitions.location.city_name).toBeDefined();
+      expect(faker.definitions.location?.city_name).toBeDefined();
```
Shinigami92: Not sure if this change is related to this PR 🤔

src/definitions/definitions.ts
```diff
@@ -54,8 +55,6 @@ export interface Definitions {
...
  */
 export type LocaleDefinition = {
-  /**
-   * The English name of the language (and the specific country, if defined).
-   */
-  title: string;
-} & LocaleEntry<Definitions>;
+  metadata: MetadataDefinitions;
+} & Partial<Definitions> &
+  Record<string, Record<string, unknown>>;
```
Shinigami92: Why is this so different as before?

Why is this not something like
```ts
MetaEntryDefinitions & LocaleEntry<Definitions>;
// or
{ metadata: MetadataDefinitions; } & LocaleEntry<Definitions>;
```

_side note: I have not checked via IDE, I just see plain text code in GitHub Web UI_

matthewmayer:
i would like to see more metadata added, for example:
```
{
title:"German (Austria)",
endonym:"Deutsch (Österreich)",
code:"de_AT",
language:"de",
country:"AT",
ltr:false,
script:"Latin"
}
```

Shinigami92:
> i would like to see more metadata added, for example:
> 
> ```
> {
> title:"German (Austria)",
> endonym:"Deutsch (Österreich)",
> code:"de_AT",
> language:"de",
> country:"AT",
> ltr:false,
> script:"Latin"
> }
> ```

Yes 🙌 
Absolutely, but this PR is just preparation for this

Edit: And this is also why I suggested to not move primitive data into own files
It would just explode in data/files that are all just 1 line long

matthewmayer:
> > i would like to see more metadata added, for example:
> > ```
> > {
> > title:"German (Austria)",
> > endonym:"Deutsch (Österreich)",
> > code:"de_AT",
> > language:"de",
> > country:"AT",
> > ltr:false,
> > script:"Latin"
> > }
> > ```
> 
> Yes 🙌 Absolutely, but this PR is just preparation for this
> 
> Edit: And this is also why I suggested to not move primitive data into own files It would just explode in data/files that are all just 1 line long

agreed, if there are ~7 metadata items like this it should be in one file total

src/locales/af_ZA/metadata/index.ts
ST-DDT: Looks like I cannot remove all special handling related to these data. 😢

src/locales/de_AT/metadata/title.ts
ST-DDT: Sure

test/faker.spec.ts
ST-DDT:
```suggestion
      expect(faker.definitions.location.city_name).toBeDefined();
```

src/definitions/definitions.ts
ST-DDT: The first because that would cause issues because then title and co would remain on the current level, which causes type safety issues.

LocaleEntry is only meant for modules. Using it for the top level would cause type safety issues again. definitions.xxx must be of type `Record<string, unknown>` and must not be `string`...

ST-DDT:
> > Yes 🙌 Absolutely, but this PR is just preparation for this
> > Edit: And this is also why I suggested to not move primitive data into own files It would just explode in data/files that are all just 1 line long
> 
> agreed, if there are ~7 metadata items like this it should be in one file total

I did it like this to avoid having to add special handling for any of these, but I will add that back.

src/locales/af_ZA/metadata/index.ts
Shinigami92: Yes, sorry for that but the meta data are not directly locales (as you already also wrote down below)
they are extra data that enriches / decorates the locale data set
So in my assumption it is totally valid that they need to get handled specially

ST-DDT: Done

src/locales/af_ZA/metadata.ts
```diff
@@ -0,0 +1 @@
+export default { title: 'Afrikaans' };
```
ST-DDT: Should we use explicit type definitions?

```suggestion
export default <MetadataDefinitions> { title: 'Afrikaans' };
```

Shinigami92: Definitely do **not** use this syntax (`<type> value`), ever
It is TS 2.x historical and deprecated
Use `value as type` instead

Beside that: I would use what is used in the other locale definition files

ST-DDT: > It is TS 2.x historical and deprecated

I didn't get a deprecation warning though.

-------

The other files use something like this:

```suggestion
const metadata: MetadataDefinitions = {
  title: 'Afrikaans',
};
export default metadata;
```

ST-DDT: Alternatively we could rewrite all index.ts files to look like this:

```suggestion
export default { title: 'Afrikaans' } as MetadataDefinitions;
```

Which should it be?

Shinigami92: > The other files use something like this:
> 
> 
> 
> ```suggestion
> 
> const metadata: MetadataDefinitions = {
> 
>   title: 'Afrikaans',
> 
> };
> 
> export default metadata;
> 
> ```

This one

ST-DDT: Done

matthewmayer (changes requested):
The available metadata should be documented (either on an explicit /api/metadata.html page or in the localization guide).

matthewmayer (dismissed):
also unless i'm missing something, there's no way to access the metadata from a Faker instance (ref #1908 ). i'd expect to be able to do something like:
```
const {fakerDE} = require("@faker-js/faker")
fakerDE.metadata.title() //German
```

Shinigami92:
> also unless i'm missing something, there's no way to access the metadata from a Faker instance (ref #1908 ). i'd expect to be able to do something like:
> 
> ```
> const {fakerDE} = require("@faker-js/faker")
> fakerDE.metadata.title() //German
> ```

I thought it would be `fakerDE.definitions.metadata.title // German`

matthewmayer:
Ok that works, but it's not documented that you can access the definitions like that.

Shinigami92:
> Ok that works, but it's not documented that you can access the definitions like that.

1. If I remember correctly we dont want to document that so it is not supported officially
2. You can use `fakerDE.helpers.fake('{{metadata.title}}')` as far as I know (that was previously not possible like `fakerDE.helpers.fake('{{title}}')` because then `title` would get parsed as "module" instead of entry/function)

matthewmayer:
> > Ok that works, but it's not documented that you can access the definitions like that.
> 
> 1. If I remember correctly we dont want to document that so it is not supported officially
> 2. You can use `fakerDE.helpers.fake('{{metadata.title}}')` as far as I know (that was previously not possible like `fakerDE.helpers.fake('{{title}}')` because then `title` would get parsed as "module" instead of entry/function)

i think there should be a stable, documented way of accessing the metadata. A faker instance should "know" what metadata is associated with it (you could previously access faker.locale, etc).

ST-DDT:
I'm torn on this one.
On one hand, having a stable way to access these metadata seems useful, but the very structure of the data make them unreliable with our normal way of accessing them. (Especially so when we merge #893, that would throw an Error for missing optional data, that are likely to be added: https://github.com/faker-js/faker/pull/1978#issuecomment-1484592384.)

I can think of two ways to access these data:

- `faker.metadata(): MetadataDefinitions` (might still require the later)
- `faker.raw(Definitions).metadata` (to be introduced, for accessing optional definitions thus avoiding the thrown error, with an explicit hint: Here be dragons) 

EDIT: Since we don't have the API for it yet, I suggest we add it in a separate PR.

Shinigami92:
> * faker.metadata(): MetadataDefinitions

I tend totally to this
The metadata is also something that should always be defined and should not contain fallback data
So I consider metadata as stable and something we want to officially support in the future

ST-DDT:
Since we don't have the API for it yet, I suggest we add it in a separate PR.
When we actually have something to expose there.

ST-DDT:
We will add it in a separate PR so we have an distinct changelog entry.
(and less importantly, this PR refactors the existing structure, whereas that method is a new feature)

ST-DDT:
> We will add it in a separate PR so we have an distinct changelog entry. (and less importantly, this PR refactors the existing structure, whereas that method is a new feature)

Extracted into: https://github.com/faker-js/faker/issues/1984

Shinigami92 (dismissed):

scripts/generateLocales.ts
```diff
@@ -303,10 +272,21 @@ let localizationLocales = '| Locale | Name | Faker |\n| :--- | :--- | :--- |\n';
...
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const metadata: MetadataDefinitions = require(pathMetadata).default;
+    localeTitle = metadata.title;
+    if (!localeTitle) {
+      throw new Error(`No title property found on ${JSON.stringify(metadata)}`);
+    }
+  } catch (e) {
+    console.log(
+      `Failed to load ${pathMetadata}. Please make sure the file exists and exports MetadataDefinitions.`,
+      e
```
xDivisionByZerox: By providing the original error as the second argument it will be joined (with a space) with the previous error message and thus displayed on the same line as first one.
For better readbility we should display it in a new log.

src/definitions/definitions.ts
```diff
@@ -39,6 +39,7 @@ export interface Definitions {
   internet: InternetDefinitions;
   location: LocationDefinitions;
   lorem: LoremDefinitions;
+  metadata: MetadataDefinitions;
```
xDivisionByZerox: Why do we need to declare it in both `Definition` AND `LocaleDefinition`. This doesn't seem right to me.

ST-DDT: I merged both together now.

scripts/generateLocales.ts
ST-DDT: Done.

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

Shinigami92 (approved):

Issue #1977, opened 2023-03-25
Missing faker.location.state data in some locales


matthewmayer:
Various locales are missing states (or equivalent 1st level administrative regions like provinces) so fall back to `en` states.
- af_ZA #1983
- dv
- fi #1986
- ge
- hr #2142
- hu #1990
- mk #2271
- pt_PT #2269
- sr_RS_latin #2270
- tr #2231
- vi #2128 
- zu_ZA #1983

I have already prepared data for some of these locales, but do you want this as 
- a) 1 big PR
- b) 12 small PRs, one per language

ST-DDT:
IMO having one PR per locale group/language (`en_*`) would be best.
That way we can easily forward it to people knowledgeable about that language.

matthewmayer:
going to close this. individual issues/PRs can be created for any remaining states with no data.

PR #1983, opened 2023-03-28
feat(locale): add af_ZA and zu_ZA states

matthewmayer:
part of #1977

Adds South African states (provinces) in Afrikaans and Zulu

References:
https://af.wikipedia.org/wiki/Provinsies_van_Suid-Afrika

https://zu.wikipedia.org/wiki/Izifundazwe_zaseNingizimu_Afrika

https://www.gov.za/about-sa/south-africas-provinces


Shinigami92 (approved):

ST-DDT (approved):

ST-DDT:
They have a strange naming pattern: Their Nord-West province is east of their Nord province. 🤣 

grafik
(Or at least the names sound like that is the case)

PR #1988, opened 2023-03-29
refactor(locale): updates county patterns for en, en_US

matthewmayer:
part of #1944

Currently the `en` faker.location.county data is a truncated list of 6 English counties.

Improve this by
1. Allowing faker.location.county to use faker.fake patterns
2. Changing the default en pattern to be {{person.last_name}} County
3. Changing the en_US data to use the most common US county names from https://en.wikipedia.org/wiki/List_of_the_most_common_U.S._county_names


ST-DDT:
I'm not sure whether we should start faking county names.

src/locales/en/location/county.ts
```diff
@@ -1,8 +1 @@
-export default [
-  'Avon',
-  'Bedfordshire',
-  'Berkshire',
-  'Borders',
-  'Buckinghamshire',
-  'Cambridgeshire',
-];
+export default ['{{person.last_name}} County'];
```
Shinigami92, line 1: Normally I would expect to have County names with famous/memorial names
This now results just in a random name with not so much value as manually calling `faker.person.lastName() + ' County'` myself.
Theoretically this could also result in even stuff like `Виноградов County` when mixing `ru` on top of `en` 👀 
So maybe in this case (for county) we maybe should just remove county from en and lead users to use the specific en_GB/en_US variants.

Shinigami92 (commented):
Looks like you need to run vitest snapshots

src/locales/en/location/county.ts
matthewmayer: I see your point, though
- `Виноградов County` seems better than `Cambridgeshire`, which is what you would get at the moment for `ru`. Eventually all locales could have 1st and 2nd level administrative areas added.
-  Faking counties potentially avoids the problem where people say "but there's not a Marion County, California, that's in Mississippi" ie the 1st and 2nd level admin areas don't match up.
- If we remove the county definitions from `en` completely won't you get 'a', 'b', 'c' in the default locale due to the issue with arrayElement?

Shinigami92: > If we remove the county definitions from en completely won't you get 'a', 'b', 'c' in the default locale due to the issue with arrayElement?

image

- https://github.com/faker-js/faker/issues/581

ST-DDT: actually

Don't forget about:

- #893

matthewmayer: Maybe last_name+" County" is not the best approach, but its similar to how faker.location.city() and faker.location.street() works in `en`. Second level admin areas in most countries are numerous enough that people would not be expected to know all their names (whereas generating completely fake states might seem weirder).

ST-DDT: IMO there are very few counties (and very few usecases for them).
If you ask for one, thrn you probably need one explicitly.
For city and street we have a method that generates them and one that returns existing ones.

matthewmayer: There are 3,006 counties in the US.

ST-DDT: If we are going for generated county names, we should have a separate method for that.
(Similar to city and street names)

The question is, do we need that?

ST-DDT:
Also we should probably add the US county names in a separate PR, than the rewrite/refactor.

src/locales/en/location/county.ts
matthewmayer: No (I find the city/cityNane and street/streetName confusing enough)

matthewmayer: Maybe just make the county definitions for en []

Then you get undefined for now which is less confusing than "b" or "Cambridgeshire" and we can note in the docs that this only works in select locales.

PR #1992, opened 2023-03-29
refactor(finance)!: simplify account implementation

ST-DDT:
We have an explicit method for generating a numbers with a given length, so lets use that instead.

This PR is a precondition to deprecating `helpers.replaceSymbolWithNumber`.

- #1994


src/modules/finance/index.ts
```diff
@@ -140,14 +140,7 @@ export class FinanceModule {
       length = 8;
```
ST-DDT: Not sure about keeping this behavior:

https://github.com/faker-js/faker/blob/7f118474d1cb4521e4c7fe736c20359f0f23f97f/src/modules/finance/index.ts#L139-L141

src/modules/finance/index.ts
```diff
@@ -140,14 +140,7 @@ export class FinanceModule {
...
 
-    let template = '';
-
-    for (let i = 0; i < length; i++) {
-      template += '#';
-    }
-
-    length = null;
-    return this.faker.helpers.replaceSymbolWithNumber(template);
+    return this.faker.string.numeric(length);
```
ST-DDT: Should this allow leading zeros?

xDivisionByZerox: From the description of `helpers.replaceSymbolWithNumbers`:
> Parses the given string symbol by symbol and replaces the placeholders with digits (`0` - `9`).

Since we generate a string which gets completely replaced, I'd say that leading zeros should be allowed.

src/modules/finance/index.ts
xDivisionByZerox: Do we want to introduce a breaking change?

ST-DDT: IMO it doesn't make sense to pass 0 to this method, so I don't see a reason why we should keep that behavior.
While technically it is a breaking change, I don't think anybody is using it.
But I would like to ask for your input on this.

xDivisionByZerox: IMO it does matter whether we think "anybody is [not] using it". 
It doesn't change the fact that it is a breaking change and should be treated as one.
Unless we want to treat this as a bug fix instead. That would work for me.

ST-DDT:
**Team Decision**

We will remove the 0 -> 8 mapping.

Shinigami92 (approved):

Shinigami92:
What a weird code ...
I'm glad we refactored it

xDivisionByZerox (approved):

PR #1999, opened 2023-03-30
test: replace deprecated method call

ST-DDT:
`rmdirSync` is deprecated.

> (node:1909) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead


xDivisionByZerox (approved):

matthewmayer (approved):
rmSync was added in node 14 which matches fakers min version so this is ok

Shinigami92 (approved):

PR #2004, opened 2023-04-01
feat: introduce locale proxy

### LINKED ISSUE #893: Introduce a way to get details about missing locale data when using them

ST-DDT:
Adds #893 

- #893 

## Changes

- Converts `faker.definitions` to a proxy that throws if `faker.definitions.category?.entry` is missing.
- Adds `faker.rawDefinitions` to access optional data

## Types

- `faker.definitions.person.first_name[0]: string`
- `faker.rawDefinitions.person?.first_name?.[0]: string | undefined`

## Usage

- `faker.helpers.arrayElement(faker.definitions.person.first_name)`
- `faker.helpers.arrayElement(faker.rawDefinitions.person?.first_name ?? ['alternatives'])`


src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+      ));
+    },
+
+    set(): never {
+      throw new FakerError('LocaleAccess is read-only.');
+    },
+
+    deleteProperty(): never {
+      throw new FakerError('LocaleAccess is read-only.');
+    },
```
ST-DDT: - [x] Write test

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    ownKeys(): Array<keyof LocaleAccess> {
+      return Object.keys(locale);
+    },
```
ST-DDT: - [x] Write test

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Creates a proxy for LocaleDefinition that throws an error when a property is accessed that is not defined.
+ *
+ * @param locale The locale definition to create the proxy for.
+ */
+export function createLocaleAccess(locale: LocaleDefinition): LocaleAccess {
+  return new Proxy({} as LocaleDefinition, {
+    has(): true {
+      return true;
+    },
```
ST-DDT: When is this used?

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+function createCategoryProxy<T extends Record<string | symbol, unknown>>(
+  categoryName: string,
+  categoryData: T = {} as T
+): Required<T> {
+  return new Proxy({} as Required<T>, {
+    has(_, entryName: keyof T): boolean {
+      const value = categoryData[entryName];
+      return value != null && (!Array.isArray(value) || value.length !== 0);
+    },
```
ST-DDT: When is this method used?

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+    },
+
+    set(): never {
+      throw new FakerError('LocaleAccess is read-only.');
+    },
+
+    deleteProperty(): never {
+      throw new FakerError('LocaleAccess is read-only.');
+    },
```
ST-DDT: - [x] Write test

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    ownKeys(): Array<string | symbol> {
+      return Object.keys(categoryData);
+    },
```
ST-DDT: - [x] Write test

test/all_functional.spec.ts
```diff
@@ -28,6 +29,48 @@ const BROKEN_LOCALE_METHODS = {
     companySuffix: ['az'],
   },
   location: {
+    streetName: [
```
ST-DDT: We should probably add `location.street_name` to our `en` locale.

test/all_functional.spec.ts
```diff
@@ -28,6 +29,48 @@ const BROKEN_LOCALE_METHODS = {
...
+      'pl',
+      'pt_BR',
+      'pt_PT',
+      'ur',
+      'vi',
+      'zh_CN',
+      'zh_TW',
+      'zu_ZA',
+    ],
+    streetAddress: ['fr_CH'],
```
ST-DDT: This looks like an actual bug in our locale data to me.

ST-DDT:
Soft blocked by #2005 

- #2005

test/all_functional.spec.ts
matthewmayer: For methods which only have data in a small number of locales, it might be useful to be able to specify "this is 'broken' in all locales EXCEPT xx, yy and zz.

matthewmayer: (i also ran into this when trying to remove the `en` county definitions, if you remove it then most locales lose their fallbacks, so you'll need to specifiy a long list of locales "except" for say en_US, en_GB, etc.)

test/faker.spec.ts
```diff
@@ -30,6 +30,19 @@ describe('faker', () => {
     }
   });
 
+  describe('rawDefinitions', () => {
+    it('locale definition accessability', () => {
```
matthewmayer:
```suggestion
    it('locale definition accessibility', () => {
```

matthewmayer:
Suppose we decide some definitions are not wanted for `en` for example we remove https://github.com/faker-js/faker/blob/next/src/locales/en/location/county.ts and it would only be available in selected locales like en_US.

What would it be replaced by? Completely remove the file? Or 
```
export default [];
```

Do we intend there for there to be a semantic difference between a missing definition and an empty array?

ST-DDT:
> What would it be replaced by? Completely remove the file? Or
> 
> ```
> export default [];
> ```

I think the `[]` is normally a good solution + a comment that the user should use a country specific variant instead.
But that would cause issues for all other locales 🤔, so maybe some dummy counties (en_US ones as they end in `County`)?

> Do we intend there for there to be a semantic difference between a missing definition and an empty array?

Yes. nullish => Missing. [] => not applicable

https://github.com/faker-js/faker/blob/109ef31da469189b123c6131bb5b9d0b3c379e8d/src/locale-proxy.ts#L70-L80

test/all_functional.spec.ts
ST-DDT: Maybe, but we should best do that in a different PR (and maybe make it a requisite of this one?).

test/all_functional.spec.ts
ST-DDT: Will be fixed in https://github.com/faker-js/faker/pull/2007

matthewmayer (changes requested):
We should document this in migration guide as a breaking change, something like this:

---

Attempting to access a method where Faker has no data will now throw a `FakerError` instead of returning `undefined`. Since the majority of methods have data in at least one fallback language (e.g. English), this error is likely to only happen in two situations

1. Accessing data which is not applicable for the locale, for example Chinese does not use prefixes in names:

```js
import { faker, fakerZH_CN } from '@faker-js/faker';

faker.name.prefix() // 'Mr'
fakerZH_CN.person.prefix() // throws a FakerError
```

2. When using a custom Faker instance rather than one of the prebuilt instances.

```js
import { Faker, fakerES, es } from '@faker-js/faker';

const fakerES_nofallbacks = new Faker({
  locale: [es],
});
fakerES.music.songName() // 'I Want to Hold Your Hand' (fallback from en)
fakerES_nofallbacks.music.songName() // throws a FakerError
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    it('should be possible to access a missing category', () => {
+      expect(locale.category).toBeDefined();
+    });
+
+    it('should not be possible to add a new category', () => {
+      expect(() => {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Shinigami92: Please use ts-expect-error instead as this would throw an error when the ts code is valid again (for whatever reason)

Shinigami92 (commented):
Looks like there should be merged some other PRs before this to reduce the diff
Maybe mark it as draft or label on hold or whatever, so reviewers are notified again when this PR is ready again

ST-DDT:
> We should document this in migration guide as a breaking change

Done

ST-DDT:
> Looks like there should be merged some other PRs before this to reduce the diff
> Maybe mark it as draft or label on hold or whatever, so reviewers are notified again when this PR is ready again

All other things are only soft blocking IMO/don't even exist yet and might not even get implemented anytime soon.
So there is nothing explicitly blocking this PR or what are you referring to specifically?

docs/guide/upgrading.md
```diff
@@ -98,6 +98,44 @@ for (let user of users) {
 
 For more information refer to our [Localization Guide](localization).
 
+### For missing locale data, Faker will now throw instead of returning `undefined` or `a`-`c`
+
+Previously, using `faker.definitions.animal.cat` returned `undefined`, when the data were accessed but missing in that locale, thus `faker.animal.cat()` returned one of `a`-`c` (`arrayElement`'s default value).
+These values aren't expected/useful as a fallback and potentially also violate the method's defined return type definitions (in case it doesn't return a `string`).
```
matthewmayer: The first sentence is a bit confusing
* `faker.definitions` was never documented, so this feels like an internal implementation detail. The breaking changes should focus on the changes in the public API rather than the rationale.
* It's not clear if this is something specific to `faker.animal.cat()` - needs a "for example".
* I think it's useful to note that if you're using the pre-packaged Faker instances this normally doesn't affect you.

matthewmayer (changes requested):

docs/guide/upgrading.md
```diff
@@ -100,7 +100,11 @@ For more information refer to our [Localization Guide](localization).
 
 ### For missing locale data, Faker will now throw instead of returning `undefined` or `a`-`c`
 
-Previously, using `faker.definitions.animal.cat` returned `undefined`, when the data were accessed but missing in that locale, thus `faker.animal.cat()` returned one of `a`-`c` (`arrayElement`'s default value).
+::: note Hinweis
```
matthewmayer: what is Hinweis?

docs/guide/upgrading.md
ST-DDT: Applied

docs/guide/upgrading.md
```diff
@@ -100,7 +100,11 @@ For more information refer to our [Localization Guide](localization).
...
+The following section mostly applies to custom-built Faker instances.
+:::
+
+Previously, for example when `en` doesn't have data for `animal.cat` then `faker.animal.cat()` would have returned one of `a`-`c` (`arrayElement`'s default value).
 These values aren't expected/useful as a fallback and potentially also violate the method's defined return type definitions (in case it doesn't return a `string`).
```
matthewmayer: hmm, think this still needs some more thought, as `animal.cat` DOES exist in `en` and DOES return a String then i think the example maybe does more harm than good. Maybe could use a completely fictional example, like `faker.animal.unicorn()` or `faker.foo.bar()`

matthewmayer (changes requested):

docs/guide/upgrading.md
ST-DDT: Or maybe like this:

````diff
- Previously, for example when `en` doesn't have data for `animal.cat` then `faker.animal.cat()` would have returned one of `a`-`c` (`arrayElement`'s default value).
+ Previously, for example if `en` didn't have data for `animal.cat`, then `faker.animal.cat()` would have returned one of `a`-`c` (`arrayElement`'s default value).
````

?

File: docs/guide/upgrading.md (hunk 12 above)
Comment (matthewmayer) [reply docs/guide/upgrading.md]: I think this is German. Same as Note?

Comment (ST-DDT) [reply docs/guide/upgrading.md]: Yes, should be fixed already.

File: docs/guide/upgrading.md (hunk 13 above)
Comment (matthewmayer) [reply docs/guide/upgrading.md]: better

File: test/location.spec.ts [hunk 14]
```diff
@@ -39,7 +39,10 @@ const NON_SEEDED_BASED_RUN = 5;
 
 describe('location', () => {
   seededTests(faker, 'location', (t) => {
-    t.itEach('street', 'streetName');
+    t.it('street');
+
+    // EN does not have streetName data
+    t.skip('streetName');
```
matthewmayer: maybe should explicitly note in the docstring for streetName that this now throws in the default `en` locale, and in most cases you should be using street() instead?

matthewmayer (changes requested):

docs/guide/upgrading.md
ST-DDT: Changed

matthewmayer:
Maybe we should also mention that this can cause `faker.fake` to throw in certain circumstances?

ST-DDT:
> Maybe we should also mention that this can cause `faker.fake` to throw in certain circumstances?

While technically true, IMO the same is true for any other faker method.

matthewmayer:
```
import { faker, fakerZH_CN } from '@faker-js/faker';

faker.name.prefix() // 'Mr'
fakerZH_CN.person.prefix() // throws a FakerError
```

^ as far as I can see, this actually returns undefined in latest build?

ST-DDT:
> ```
> import { faker, fakerZH_CN } from '@faker-js/faker';
> 
> faker.name.prefix() // 'Mr'
> fakerZH_CN.person.prefix() // throws a FakerError
> ```
> 
> ^ as far as I can see, this actually returns undefined in latest build?

Looks like a side effect of not having https://github.com/faker-js/faker/issues/1677 yet.

- https://github.com/faker-js/faker/issues/1677

But I'm not sure whether I should add some code here to handle that case or implement that feature first.

matthewmayer:
> > ```
> > import { faker, fakerZH_CN } from '@faker-js/faker';
> > 
> > faker.name.prefix() // 'Mr'
> > fakerZH_CN.person.prefix() // throws a FakerError
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ^ as far as I can see, this actually returns undefined in latest build?
> 
> Looks like a side effect of not having #1677 yet.
> 
> * [Change file structure for gendered lists (names, prefixes etc) #1677](https://github.com/faker-js/faker/issues/1677)
> 
> But I'm not sure whether I should add some code here to handle that case or implement that feature first.

for now we could remove it or replace it with a better example. For example, Ghana doesn't have postcodes/ZIP codes. We could remove the incorrect data at https://github.com/faker-js/faker/blob/next/src/locales/en_GH/location/postcode.ts and then use the example of fakerEN_GH.location.zipCode() throwing.

ST-DDT:
> Ghana doesn't have postcodes/ZIP codes. We could remove the incorrect data at [`next`/src/locales/en_GH/location/postcode.ts](https://github.com/faker-js/faker/blob/next/src/locales/en_GH/location/postcode.ts?rgh-link-date=2023-04-04T10%3A26%3A52Z)

Please create a separate issue for that or a PR that fixes it.

---

> as far as I can see, this actually returns undefined in latest build?

Instead of dancing around the issue I think we should just address #1677 and be done with it.

ST-DDT:
Note to self: 
Restore name and address aliases.

docs/guide/upgrading.md
```diff
@@ -97,6 +97,48 @@ for (let user of users) {
 
 For more information refer to our [Localization Guide](localization).
 
+### For missing locale data, Faker will now throw instead of returning `undefined` or `a`-`c`
+
+::: note Note
+The following section mostly applies to custom-built Faker instances.
+:::
+
+Previously, for example if `en` didn't have data for `animal.cat`, then `faker.animal.cat()` would have returned one of `a`-`c` (`arrayElement`'s default value).
```
```suggestion
Previously, for example if `en` didn't have data for `animal.cat`, then `faker.animal.cat()` would have returned one of `a`, `b` or `c` (`arrayElement`'s default value).
```

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+ * A proxy for LocaleDefinitions that marks all properties as required and throws an error when an entry is accessed that is not defined.
+ */
+export type LocaleAccess = Readonly<{
+  [key in keyof LocaleDefinition]-?: Readonly<
+    Required<NonNullable<LocaleDefinition[key]>>
+  >;
+}>;
+
+/**
+ * Creates a proxy for LocaleDefinition that throws an error when a property is accessed that is not defined.
```
Shinigami92:
```suggestion
 * Creates a proxy for LocaleDefinition that throws an error if an undefined property is accessed.
```

This suggestion minimizes the doubled "that"

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+
+    ownKeys(): Array<keyof LocaleAccess> {
+      return Object.keys(locale);
+    },
+  }) as LocaleAccess;
+}
+
+/**
+ * Creates a proxy for a category that throws an error when a property is accessed that is not defined.
```
Shinigami92:
```suggestion
 * Creates a proxy for a category that throws an error when accessing an undefined property.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    it('should be possible to access a missing category', () => {
+      expect(locale.category).toBeDefined();
+    });
+
+    it('should not be possible to add a new category', () => {
+      expect(() => {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-expect-error
```
Shinigami92: You should provide a reason for all `expect-error`s
like `// @ts-expect-error: ... reason ...`

then you should be able to remove the eslint-disable comments

matthewmayer:
> fakerZH_CN.person.prefix()

Instead of the `fakerZH_CN.person.prefix()` example (which doesn't actually throw yet) we could use the example of fakerBASE which does throw for many methods, for example:
```ts
import { faker, fakerBASE } from '@faker-js/faker';

faker.person.fullName() // 'Shelia Renner'
fakerBASE.person.fullName() // throws a FakerError
```

ST-DDT:
**Team Decision**

- We decided not to wait on #1677

src/locale-proxy.ts
xDivisionByZerox: May you explain why this is useful?

src/modules/helpers/index.ts
```diff
@@ -1139,7 +1139,7 @@ export class HelpersModule {
     const parts = method.split('.');
 
     let currentModuleOrMethod: unknown = this.faker;
-    let currentDefinitions: unknown = this.faker.definitions;
+    let currentDefinitions: unknown = this.faker.rawDefinitions;
```
xDivisionByZerox, line 1163: Why is `helpers.fake()` allowed to bypass the proxy?

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    it('should be possible to check for an existing category', () => {
+      expect('airline' in locale).toBe(true);
+    });
+
+    it('should be possible to access the title', () => {
+      expect(locale.metadata.title).toBe('English');
+    });
+
+    it('should be possible to access a missing category', () => {
```
xDivisionByZerox, line 23: How is this different from the first test?

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+
+      return (target[categoryName] = createCategoryProxy(
+        categoryName,
+        locale[categoryName]
+      ));
+    },
+
+    set(): never {
+      throw new FakerError('LocaleAccess is read-only.');
```
xDivisionByZerox: I'm not sure if this error makes sense to the user since they never directly interact with `LocalAccess`. Maybe something more generic like:

```suggestion
      throw new FakerError('Locale data set's are read-only.');
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,147 @@
+import { describe, expect, it } from 'vitest';
+import type { AirlineDefinitions, MetadataDefinitions } from '../src';
+import { en, FakerError } from '../src';
+import { createLocaleAccess } from '../src/locale-proxy';
+
+describe('LocaleAccess', () => {
+  const locale = createLocaleAccess(en);
+  const unavailable = createLocaleAccess({
```
xDivisionByZerox: Maybe scope `unavailable` in the actual test case since it is only used there?
I was a bit confused when reading all test cases as I was expecting both proxies to be used.

test/location.spec.ts
xDivisionByZerox: We normally do not want to mention the locale-specific context in the method description. They are decoupled from each other, so making them aware of each other in the documentation doesn't make sense to me.
(You would need to remember to update the JSDocs for the function when you refactor the locale data set)

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
+import type { LocaleDefinition } from './definitions';
+import { FakerError } from './errors/faker-error';
+
+/**
+ * A proxy for LocaleDefinitions that marks all properties as required and throws an error when an entry is accessed that is not defined.
+ */
+export type LocaleAccess = Readonly<{
```
xDivisionByZerox: The file name is called `locale-proxy`, so why call it `LocaleAccess` instead of `LocaleProxy`?

test/locale-proxy.spec.ts
Shinigami92: `expect(locale.category).toBeDefined()` checks for `!== undefined` where as `expect('category' in locale).toBe(true)` just checks if the key is defined

src/locale-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+  }) as LocaleAccess;
+}
+
+/**
+ * Creates a proxy for a category that throws an error when accessing an undefined property.
+ *
+ * @param categoryName The name of the category.
+ * @param categoryData The module to create the proxy for.
+ */
+function createCategoryProxy<T extends Record<string | symbol, unknown>>(
```
Shinigami92: How about name `T` to `CategoryData`?

src/modules/helpers/index.ts
ST-DDT: Because it should not throw an error. Because it also has to check the module method and has a custom error message.

src/locale-proxy.ts
ST-DDT: This can be used to check if the locale data are present without risk getting an error.
The category is proxied so it is always there.

test/faker.spec.ts
```diff
@@ -1,6 +1,6 @@
 import type { SpyInstance } from 'vitest';
 import { describe, expect, it, vi } from 'vitest';
-import { faker, Faker } from '../src';
+import { Faker, faker } from '../src';
```
Shinigami92: :eyes: :thinking:

ST-DDT:
```suggestion
import { faker, Faker} from '../src';
```

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):

Shinigami92:

matthewmayer:
The second example in the upgrading guide (Chinese prefix) is still incorrect as it does not throw. Suggest replacing with a fakerBASE example.

src/locale-proxy.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Creates a proxy for LocaleDefinition that throws an error if an undefined property is accessed.
+ *
+ * @param locale The locale definition to create the proxy for.
+ */
+export function createLocaleProxy(locale: LocaleDefinition): LocaleProxy {
+  const proxies = {} as LocaleDefinition;
+  return new Proxy(locale, {
+    has(): true {
+      // Categories are always present (proxied), that why we return true.
```
ST-DDT:
```suggestion
      // Categories are always present (proxied), that's why we return true.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        // @ts-expect-error LocaleAccess is read-only.
+        delete locale.category.missing;
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to delete a missing entry in an existing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        locale.airline.airline = [];
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to delete a missing entry in a missing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        locale.airline.missing = {};
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to replace an entry in an existing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        locale.category.missing = {};
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to add a new entry in an existing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        new FakerError(
+          `The locale data for 'airline.airline' aren't applicable to this locale.
+  If you think this is a bug, please report it at: https://github.com/faker-js/faker`
+        )
+      );
+    });
+
+    it('should not be possible to add a new entry in a missing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        delete locale.airline.missing;
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to delete an existing entry in an existing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        delete locale.category;
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to delete an existing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        locale.airline = {};
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to delete a missing category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        locale.category = {};
+      }).toThrowError(
+        new FakerError('You cannot edit the locale data on the faker instance')
+      );
+    });
+
+    it('should not be possible to replace a category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

test/locale-proxy.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+      expect(locale.metadata.title).toBe('English');
+    });
+
+    it('should be possible to access a missing category', () => {
+      expect(locale.category).toBeDefined();
+    });
+
+    it('should not be possible to add a new category', () => {
+      expect(() => {
+        // @ts-expect-error LocaleAccess is read-only.
```
ST-DDT:
```suggestion
        // @ts-expect-error: LocaleProxy is read-only.
```

ST-DDT:
Blocked by #2045 regarding throwing for `fakerZH_CN.person.prefix()` in the migration guide.
But the rest is still reviewable.

src/locale-proxy.ts
```diff
@@ -0,0 +1,91 @@
...
+function createCategoryProxy<
+  CategoryData extends Record<string | symbol, unknown>
+>(
+  categoryName: string,
+  categoryData: CategoryData = {} as CategoryData
+): Required<CategoryData> {
+  return new Proxy(categoryData, {
+    has(target: CategoryData, entryName: keyof CategoryData): boolean {
+      const value = target[entryName];
+      return value != null && (!Array.isArray(value) || value.length !== 0);
```
xDivisionByZerox: What about nested object definitions? Is the value `{}` expected to return `true`? If so I would expect `[]` to return `true` as well.

src/locale-proxy.ts
```diff
@@ -0,0 +1,91 @@
...
+          `The locale data for '${categoryName}.${entryName.toString()}' are missing in this locale.
+  Please contribute the missing data to the project or use a locale/Faker instance that has these data.
+  For more information see https://next.fakerjs.dev/guide/localization.html`
+        );
+      } else if (Array.isArray(value) && value.length === 0) {
+        throw new FakerError(
+          `The locale data for '${categoryName}.${entryName.toString()}' aren't applicable to this locale.
+  If you think this is a bug, please report it at: https://github.com/faker-js/faker`
+        );
+      } else {
```
xDivisionByZerox, line 83: Same here. What about `{}`?

src/locale-proxy.ts
ST-DDT: We dont have those yet, so we havent decided on them either.
🤔

ST-DDT: Then we would also have to decide on `''` and `0` and other special values.
Or is maybe `[]` a bad value for explicitly absent? Would `null` be any better?

xDivisionByZerox: > Would null be any better?

I mean that's literally what `null` is for. From the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null):

> The `null` value represents the intentional absence of any object value.

---

*I might be a bit biased since I'm a `null` enthusiast, but most JS/TS developers are not.

Shinigami92: @xDivisionByZerox That is a really good point :+1: 

However, can we please do this in a separate PR?

Shinigami92: @xDivisionByZerox just as a side note, if we want to use `null` for such cases, we need to (internally) support [this](https://github.com/faker-js/faker/pull/2045/files#diff-3f6b4fc424ba2cedcbf5f612ba256d248f0c05f46314eed1dfa81590bcd429fdR807-R812) `null` check also in the future (only at runtime) and the TODO comment would be obsolete

But yes, I'm fine with that and this would get an agree if we would have a "needs-decision"

ST-DDT: Extracted into: https://github.com/faker-js/faker/pull/2078

Shinigami92: nevermind, seems I have not understand the PR in the first place :facepalm: 
ignore my last comment

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/2078

src/locale-proxy.ts
```diff
@@ -61,25 +61,25 @@ function createCategoryProxy<
...
       entryName: keyof CategoryData
     ): CategoryData[keyof CategoryData] {
       const value = target[entryName];
-      if (value == null) {
+      if (value === null) {
+        throw new FakerError(
+          `The locale data for '${categoryName}.${entryName.toString()}' aren't applicable to this locale.
+      If you think this is a bug, please report it at: https://github.com/faker-js/faker`
+        );
+      } else if (value == null) {
```
xDivisionByZerox: Since we now explicitly check for `null` this check should explicitly be `value === undefined`. 
`value == null` basically is a "shorthand" for `value === undefined || value === null`. It does NOT check for falsely values.

test/all_functional.spec.ts
```diff
@@ -29,47 +29,6 @@ const BROKEN_LOCALE_METHODS = {
     companySuffix: ['az'],
   },
   location: {
-    streetName: [
```
xDivisionByZerox: <s>Why are those no longer broken`?</s>

Seems like the review did not show me the latest version...

src/locale-proxy.ts
ST-DDT:
```suggestion
      } else if (value === undefined) {
```

ST-DDT:
Ready to review.

src/locale-proxy.ts
```diff
@@ -0,0 +1,91 @@
...
+      const value = target[entryName];
+      return value != null;
+    },
+
+    get(
+      target: CategoryData,
+      entryName: keyof CategoryData
+    ): CategoryData[keyof CategoryData] {
+      const value = target[entryName];
+      if (value === null) {
```
Shinigami92, line 72: Can we please add test cases to cover this tripple equals null check as well as the tripple equals undefined check below?
That prevents regression and serves as stability so we can point to the test why this tripple equals are necessary

xDivisionByZerox (approved):

src/locale-proxy.ts
xDivisionByZerox: Is that not covered by the following test?
https://github.com/faker-js/faker/blob/f915d49505f36fa74e6e3045887d9cd407fe67dc/test/locale-proxy.spec.ts#L111-L123

Shinigami92: I checked the last two newest commits by @ST-DDT and nothing new were added in terms of test
So somehow it is not covered or nothing subjective have changed to the behavior of the code

ST-DDT: It is true that nothing substatial has changed since being labelled as on hold previously. Because only merge conflicts and slight adjustments have been adressed.
The existing tests (added earlier in this PR) should cover all cases AFAICT. Please check the test spec if there are cases that are not or insufficiently covered.

Shinigami92: `if (value === null)` on line 72 was already there, so the else-if in line 77 did not get the case for having value === null in any case
thats why everything is working 👍

Shinigami92 (approved):

docs/guide/upgrading.md
```diff
@@ -97,6 +97,48 @@ for (let user of users) {
...
+  locale: [zh_CN],
+});
+
+faker.name.prefix(); // 'Mr'
+// Previously:
+//fakerZH_CN_noFallbacks.person.prefix(); // undefined
+// Now:
+fakerZH_CN.person.prefix(); // throws a FakerError
+```
+
```
matthewmayer, line 141: suggest removing this example as it doesn't work for gendered data yet.

matthewmayer (changes requested):

docs/guide/upgrading.md
ST-DDT: I ran the code this morning and it did throw an error due to the arrayElement PR.

matthewmayer: I think it's a confusing example though as it shows a different error to what would be expected

Better to use fakerBASE or we could add a second example later using the en_HK with missing postcodes?

matthewmayer: Anyway I'll remove my blocking review if you want to merge without changing this

matthewmayer (approved):

docs/guide/upgrading.md
ST-DDT: I'll merge this, now so it unblocks the release progress, but I will open an issue to switch the example once en_HK is merged.

PR #2005, opened 2023-04-01
docs(person): fix formatting issues

ST-DDT:
Fixes some jsdocs formatting issues in the `PersonModule`.


ST-DDT:
@Shinigami92 Any idea why these aren't detected and fixed by prettier in CI?
I only found these by manually saving the file in VS Code.

matthewmayer (approved):

Shinigami92:
> @Shinigami92 Any idea why these aren't detected and fixed by prettier in CI? I only found these by manually saving the file in VS Code.

There could be multiple reasons right now
I have not checked the diff yet, but:
- prettier had some updates recently
- VSCode had an update (and so maybe the prettier VSCode extension might need an update)
- pnpm had a major update (so the prettier plugin loading might have changed or couldn't get picked up)
- and other tools like eslint and so on also had updates

so tldr, right now I feel like many tools have major upgrades around us
and maybe all these components need to be compatible again with each other

Shinigami92 (approved):

PR #2009, opened 2023-04-02
chore: activate eslint no-else-return

Shinigami92:
In team meeting from 09.03.2023 we decided to turn on the eslint rule `no-eslin-return` based on the result of the poll https://main.elk.zone/mas.to/@Shini92/109919131241994227

This aims for having equally written (styled) code over the whole source code and prevent future discussions about how it should be written or not
Now there is just one solution


scripts/apidoc/diff.ts
```diff
@@ -37,9 +37,9 @@ async function loadLocal(path: string): Promise<DocsApiDiffIndex> {
 async function load(source: string): Promise<DocsApiDiffIndex> {
   if (source.startsWith('https://')) {
     return loadRemote(source);
-  } else {
-    return loadLocal(source);
   }
+
+  return loadLocal(source);
```
Shinigami92: @ST-DDT I could also rewrite this to `return source.startsWith('https://') ? loadRemote(source) : loadLocal(source);`
Do you think in the future this code could get a third case?

scripts/apidoc/signature.ts
```diff
@@ -296,9 +296,9 @@ function typeToText(type_?: Type, short = false): string {
       const text = typeToText(type.elementType, short);
       if (text.includes('|') || text.includes('{')) {
         return `Array<${text}>`;
-      } else {
-        return `${text}[]`;
       }
+
+      return `${text}[]`;
```
Shinigami92: This could also be rewritten to e.g.
```ts
const conditionNameTBD = text.includes('|') || text.includes('{');
return conditionNameTBD ? `Array<${text}>`: `${text}[]`;
```

scripts/apidoc/diff.ts
ST-DDT: > Do you think in the future this code could get a third case?

No

> I could also rewrite this to `return source.startsWith('https://') ? loadRemote(source) : loadLocal(source);`

I that still within 80 characters?

scripts/apidoc/signature.ts
ST-DDT: IMO that looks less readable than the current suggestion.

scripts/apidoc/diff.ts
matthewmayer: > @ST-DDT I could also rewrite this to `return source.startsWith('https://') ? loadRemote(source) : loadLocal(source);`
> 


But I thought "Now there is just one solution" 😀

scripts/apidoc/signature.ts
Shinigami92: Really?
IMO this is just because the `conditionNameTBD` didnt get a name yet

But 
```ts
isComplexType ? `Array<${text}>`: `${text}[]`
```
reads just perfect to me 🤔

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2025, opened 2023-04-05
feat(locale): add additional metadata properties

### LINKED ISSUE #2012: More metadata for each locale
follows on from #1978 
related to #1984 

I would like to see more metadata added for each locale, for example: 

```js
{
title:"German (Austria)",
endonym:"Deutsch (Österreich)", //in native language
code:"de_AT", //same as locale key
language:"de", //ISO_639-1
country:"AT", //ISO 3166-1 alpha-2. omitted if no country
rtl:false, //right-to-left
script:"Latn" //ISO_15924
}
```

I would be happy to implement this but maybe can get some feedback on the naming of the keys and values first?

matthewmayer:
fix #2012

* Tried to make locale titles more consistent, so if a country is specified it's always in parentheses. Also favour Slovak and Nepali over Slovakian and Nepalese.
* only put minimal data for base locale
* this probably needs more tests, but not sure exactly what best to test


matthewmayer:
i might need some assistance with the "Make MetadataDefinitions optional and Partial" part, wasn't sure how to do that.

src/definitions/metadata.ts
```diff
@@ -3,4 +3,36 @@ export type MetadataDefinitions = {
    * The English name of the language (and the specific country, if defined).
    */
   title: string;
+  /**
+   * The full code of the locale, including the country code if applicable.
```
ST-DDT, line 12: Maybe add a hint that this follows the bc47(?) standard?

(not included in the suggestions)

src/definitions/metadata.ts
```diff
@@ -3,4 +3,36 @@ export type MetadataDefinitions = {
...
+   */
+  code?: string;
```
ST-DDT: This could be required.

(handled by the suggestions)

ST-DDT (commented):
Here the suggested metadata definition type changes as requested: https://github.com/matthewmayer/faker/pull/1

src/definitions/metadata.ts
matthewmayer: except it doesn't really because they use hyphens not underscores https://en.wikipedia.org/wiki/IETF_language_tag

matthewmayer:
> i might need some assistance with the "Make MetadataDefinitions optional and Partial" part, wasn't sure how to do that.

thanks for the help on this!

src/definitions/metadata.ts
ST-DDT: I don't have a good description for this right now, maybe someone else has a suggestion.
If not we can ignore that for now.

Maybe we should at least describe the expected format like `<locale>[_<country>[_<variant>]]`.

ST-DDT (dismissed):
Thanks for filling in these metadata. The PR looks good to me. 

I will have another look at the endonyms tomorrow just to be sure nothing slipped my checks.

src/definitions/metadata.ts
```diff
@@ -1,6 +1,60 @@
...
+   */
+  language: string;
+  /**
+   * The specific variant of the language. This usually refers to a dialect or slang.
+   */
+  variant?: string;
+  /**
+   * Whether the language is written right-to-left.
+   */
+  rtl: boolean;
```
xDivisionByZerox: I'm a bit confused on this, since the ISO 15924 value from the `script` property already tells us about this behavior.

src/definitions/metadata.ts
```diff
@@ -1,6 +1,60 @@
...
+   *
+   * @see https://en.wikipedia.org/wiki/ISO_15924
+   */
+  script: string;
+};
+
+/**
+ * Metadata for pre-built locales for a specific country.
+ */
+export type PreBuiltMetadataDefinitionsForCountry =
```
xDivisionByZerox, line 46: I'm currently on phone, so I might have missed something. 
Why is this required instead of merging it with `PrebuildMetaDataDefinitions`?

src/locales/sr_RS_latin/metadata.ts
```diff
@@ -1,7 +1,13 @@
-import type { MetadataDefinitions } from '../..';
+import type { PreBuiltMetadataDefinitionsForCountry } from '../../definitions/metadata';
 
-const metadata: MetadataDefinitions = {
-  title: 'Serbian (Latin)',
+const metadata: PreBuiltMetadataDefinitionsForCountry = {
+  title: 'Serbian (Serbia, Latin)',
```
xDivisionByZerox, line 4: Would "Latin" not be the `variant` here?

test/locale-imports.spec.ts
```diff
@@ -29,6 +29,40 @@ describe('locale imports', () => {
...
+          'Grek',
+          'Hans',
+          'Hant',
+          'Hebr',
+          'Jpan',
+          'Kore',
+          'Latn',
+          'Thaa',
+          'Thai',
+        ]).toContain(metadata.script);
```
xDivisionByZerox, line 56: Can we store this in a `validISO15924Scripts` variable which already include all possible scripts?

Same for the `language` property.

test/locale-imports.spec.ts
```diff
@@ -29,6 +29,40 @@ describe('locale imports', () => {
...
+        expect(metadata.endonym).toBeTypeOf('string');
+        expect(metadata.rtl).toBeTypeOf('boolean');
+        if (metadata.country) {
+          expect(metadata.country).toBeTypeOf('string');
+          expect(metadata.country).toMatch(/^[A-Z]{2}$/);
```
xDivisionByZerox, line 62: `toHaveLength`? Can we please check both error messages and choose the more verboten one?

src/definitions/metadata.ts
matthewmayer: It's just a convenience property, for example if you are rendering locale data in HTML you might want to apply the dir="rtl" attribute on elements with right to left text, without having to know this specifically applies in Arabic, Hebrew, etc.

src/definitions/metadata.ts
matthewmayer: @ST-DDT added this, i think it just ensures that the xx_XX locales have the extra country properties?

src/locales/sr_RS_latin/metadata.ts
matthewmayer: i was assuming variant was only used when language, country and script dont distinguish the locale (ie OCKER and BORK only). But could also be used here.

test/locale-imports.spec.ts
matthewmayer: could do, though list is fairly long
https://unicode.org/iso15924/iso15924-codes.html

My feeling was that in the rare event a language with a completly new script is added, it would be sufficient to add it to the list in the tests manually.

test/locale-imports.spec.ts
matthewmayer: toHaveLength would only check its a string of length 2? This also verifies its all caps?

Which error messages do you refer to?

src/definitions/metadata.ts
ST-DDT: We can merge it back if it adds to much overhead.

src/definitions/metadata.ts
ST-DDT: Should we name the property `dir: ltr | rtl` instead?

test/locale-imports.spec.ts
xDivisionByZerox: Good point, I didn't consider that while writing my comment.
Still, I would prefer having two separate verbose expect statements over a single RegExp. So one check against `toHaveLength` and one to verify the casing `expect(value, 'expect all characters to be uppercase').toBe(value.toUpperCase())`.

ST-DDT: There might be something in validatorjs for that as well.

matthewmayer: i feel `/^[A-Z]{2}$/` is quite expressive for 'Exactly two plain ASCII uppercase letters'. Just checking for toUpperCase would allow 'ÁÉ' for example.

matthewmayer: we could use an official list of ISO country codes, but that would exclude things like Kosovo https://en.wikipedia.org/wiki/XK_(user_assigned_code) so i think just checking for two uppercase A-Z chars is OK.

src/locales/sr_RS_latin/metadata.ts
matthewmayer: added variant

src/definitions/metadata.ts
matthewmayer: maybe we could add a paragraph to bottom of https://next.fakerjs.dev/guide/localization.html explaining the structure of locale codes, then we could link that from here?

src/definitions/metadata.ts
matthewmayer: i'd be fine with that. it would also make everything a string which might be more convenient.

xDivisionByZerox: > Should we name the property dir: ltr | rtl instead?

For the sake of extendability, I would favor this version as well.

test/locale-imports.spec.ts
xDivisionByZerox: In the long run, I thought about a `.toSatisfy(isISO15924)` validator.js function. Guess I'll open a feature request on their repo.

src/definitions/metadata.ts
matthewmayer: updated to dir string property

ST-DDT (dismissed):

src/definitions/definitions.ts
```diff
@@ -32,7 +32,7 @@ export type LocaleEntry<T extends Record<string, unknown>> = Partial<T> &
  * that most properties are optional and extra properties are allowed.
  */
 export type LocaleDefinition = {
-  metadata: MetadataDefinitions;
+  metadata?: MetadataDefinitions;
```
Shinigami92, line 35: I havn't checked all the comments yet, but out of curiosity
why and when is it possible to have this undefined?
Even for `base` I would expect base-pseudo data, so this is always defined

src/definitions/metadata.ts
Shinigami92: all good ideas, we should do one of these :+1:

src/definitions/metadata.ts
```diff
@@ -1,6 +1,60 @@
...
+   */
+  language: string;
+  /**
+   * The specific variant of the language. This usually refers to a dialect or slang.
+   */
+  variant?: string;
+  /**
+   * The direction of the language, either 'ltr' (left to right) or 'rtl' (right to left).
+   */
+  dir: string;
```
Shinigami92:
```suggestion
  dir: 'ltr' | 'rtl';
```

test/locale-imports.spec.ts
```diff
@@ -29,6 +29,41 @@ describe('locale imports', () => {
       );
     });
 
+    it(`should have complete metadata values`, () => {
```
Shinigami92:
```suggestion
    it('should have complete metadata values', () => {
```

Shinigami92 (commented):
all in all looks good to me
still just some questions and suggestions

src/definitions/definitions.ts
matthewmayer: So that if you are defining a custom locale you don't have to provide metadata?

Shinigami92: mh... that's not a bad argument :thinking: 
I might need to discuss the pros and cons for this with the others

matthewmayer:
```
const customLocale: LocaleDefinition = {
  metadata: {... blah blah blah.lots of boilerplate...}
  internet: {
    domainSuffix: ['test'],
  },
};

export const customFaker = new Faker({
  locale: [customLocale, de_CH, de, en, base],
});
```

src/definitions/metadata.ts
matthewmayer: https://github.com/faker-js/faker/pull/2030

ST-DDT (approved):

test/locale-imports.spec.ts
xDivisionByZerox: I opened [2215](https://github.com/validatorjs/validator.js/pull/2215) in their repo.
So this comment can be resolved for now.

src/definitions/metadata.ts
xDivisionByZerox: @matthewmayer @ST-DDT This is still in the code. Any info?

matthewmayer: Up to @ST-DDT

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #2032, opened 2023-04-09
fix(location): no leading zero on building number or secondary address

### LINKED ISSUE #2031: Don't allow faker.location.buildingNumber() to start with a 0

Building numbers that start with a zero lead to addresses like '013 Schultz Circles' which look incorrect.

```
faker.helpers.multiple(faker.location.buildingNumber, {count:100}).filter(x=>x.startsWith("0"))
[
  '050',   '01460', '07806',
  '001',   '05217', '0670',
  '07971', '076',   '03212',
  '034',   '062',   '005',
  '08119'
]
```

### Suggested solution

Change the patterns. As this uses replaceSymbolWithNumber we only have the choice of # (0-9) or ! (2-9) as special chars.

We can use a pattern like `['1##', '!##', '1###', '!###', '1####', '!####']` to ensure the numbers can start with any of 1-9, but not 0. This causes a slight side-effect that numbers starting with 1 will be more common. However Benford's law means thats not unrealistic anyway.

### Alternative

_No response_

### Additional context

_No response_

matthewmayer:
fix #2031

Note this only affects default `en` locale. Similar approach could be taken in other locales which have building_number patterns


ST-DDT:
IMO instead of doing this strange pattern duplication stuff we should rather change the implementation to not use replacewithnumber and instead use a simple regex replace with numeric without leading zero or something.

ST-DDT:
**Team Decision**

- We will implement this via `arrayElement(patterns).replace(/#+/g, (m) => string.numeric(m.length)`
- The locale definition documentation will be extended to explicitly mention the `# -> number` transformation
- We will add a test that checks that it will not return leading zeros

See https://github.com/faker-js/faker/issues/2031#issuecomment-1516668522

@matthewmayer Will you update your PR or start a new one?

src/definitions/location.ts
```diff
@@ -72,7 +72,7 @@ export type LocationDefinitions = LocaleEntry<{
   direction_abbr: string[];
 
   /**
-   * The pattern used to generate building numbers.
+   * The pattern used to generate building numbers. Any strings of # will be replaced by with numbers without a leading zero.
```
Shinigami92: @import-brain this doesn't sound right to me, can you check this sentence?

matthewmayer: rewrote it to make it clearer

src/modules/location/index.ts
```diff
@@ -188,9 +188,14 @@ export class LocationModule {
...
+    const format = this.faker.helpers
+      .arrayElement(this.faker.definitions.location.building_number)
+      .replace(/#+/g, (m) =>
+        this.faker.string.numeric({
+          length: m.length,
+          allowLeadingZeros: false,
+        })
+      );
 
     return this.faker.helpers.replaceSymbolWithNumber(format);
```
ST-DDT: This can now be removed.

ST-DDT:
Also the PR name should be adjusted.

src/modules/location/index.ts
matthewmayer: why?

matthewmayer:
Would it also make sense to apply this for secondary address? To avoid "Apt 000" in en or '3/0' in pl etc.

src/modules/location/index.ts
ST-DDT: Because everything this method effectively does is already covered by the above regex.

matthewmayer: sorry maybe i'm misunderstanding. what lines exactly do you want to remove?

matthewmayer: ah right you mean  `return this.faker.helpers.replaceSymbolWithNumber(format);` can just be `return format` because the replaceSymbolWithNumber wont do anything any more.

matthewmayer:
> Would it also make sense to apply this for secondary address? To avoid "Apt 000" in en or '3/0' in pl etc.

in this PR or a seperate PR?

ST-DDT (dismissed):

ST-DDT:
> in this PR or a seperate PR?

Since it does the exact same thing for related stuff, I am fine with doing it in one PR.
Not sure what the others think though.

The PR title needs to be adjusted anyway.

ST-DDT (approved):

xDivisionByZerox (approved):

ST-DDT (approved):

PR #2041, opened 2023-04-12
infra: auto comment on feature requests

xDivisionByZerox:
## Description

This PR adds a workflow that automatically comments on issues that get the `s: waiting for user interest` label assigned. 

This allows us to reduce maintenance resources on feature requests by first checking whether a feature is actually requested by the community instead of a single individual.


xDivisionByZerox:
I tested the implementation in [a new repo](https://github.com/xDivisionByZerox/test-issue-comment-on-label) on my site.

- [Test with a single label assigned on creation](https://github.com/xDivisionByZerox/test-issue-comment-on-label/issues/1)
- [Test with a multiple label assigned on creation](https://github.com/xDivisionByZerox/test-issue-comment-on-label/issues/2)

---

⚠ I will delete this comment if the PR itself gets merged since I will delete the test repo as well.

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Add Comment For User Interest
+        uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'This issue has been marked as "waiting for user interest".\nIf you would like to see this feature being implemented, please react to the original issue with an up-vote (:+1:).',
```
ST-DDT: I don't like the `original issue` part.
Maybe call it `issue/PR description`?

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,24 @@
+name: Comment Issue
+
+on:
+  issues:
+    types:
+      - labeled
```
ST-DDT, line 6: Does this also affect PRs?
IMO it should.

ST-DDT (commented):
I like it. Just some minor things.

ST-DDT:
Can you please create an issue template in the test project so that the external person can set the labels?
Because then the user might not have permissions to post the comment on behalf of the repository.

.github/workflows/comment-issue.yml
xDivisionByZerox: No it doesn't. But I can configure it.

matthewmayer:
honestly, i don't like it. particularly for a new contributor it seems a bit off-putting to be sent a completely canned response by a robot. if there were hundreds of these suggestions sure but it's like 1 or 2 a month?

ST-DDT:
> Can you please create an issue template in the test project so that the external person can set the labels? Because then the user might not have permissions to post the comment on behalf of the repository.

Works as expected

https://github.com/xDivisionByZerox/test-issue-comment-on-label/issues/5#issuecomment-1506455845

.github/workflows/comment-issue.yml
ST-DDT: We should also include some kind of why we are doing that to the message and allow users to post their suggestions regarding this feature in the comments.

ST-DDT:
**Team Decision**

The voting on this issue was ambiguous.
So we are waiting on more input before we decide on this one.

xDivisionByZerox:
> honestly, i don't like it. particularly for a new contributor it seems a bit off-putting to be sent a completely canned response by a robot. if there were hundreds of these suggestions sure but it's like 1 or 2 a month?

@matthewmayer I look at it the other way around. 
By automatically replying to an issue that just got the `s: waiting for user interest` label assigned, the user gets more information on what exactly is happening.
The fact is, every feature request should get this label. We also know that we need to carefully think about which features to support since each additional feature potentially adds a lot of overhead and maintenance. So what this automatic reply does is basically what each of us would do minus we don't need to search for the existing template text.

Please keep in mind this reply ONLY applies if an issue (or maybe PR) gets the `s: waiting for user interest` label assigned. Yes, this includes every issue opened via the feature report issue template (done in #2028). If you want to work around this you are free to use the freestyle issue.

pkuczynski:
For me it is okay. New contributors would get some feedback that we await more interest. We could think of a bit more explanatory and soft message...

xDivisionByZerox:
> We could think of a bit more explanatory and soft message...

Totally agree here. I simply proposed a first message that (at least I thought so) projects what we want to do here.
I'm totally fine going into more detail on:
- Why we waiting for more user interest
- How a contributor can help us achieve their request

damienwebdev:
I think this is a great idea. One of the most difficult decisions that can be made as a maintainer is whether or not the maintenance burden of a feature outweighs the value it provides to your package consumers.

If we can do anything to align this better, I'm all for it!

matthewmayer:
Fair enough. I can help soften the message a little!

ST-DDT:
Maybe we can write something like this:

> Thanks for feature proposal.
>
> We marked it as "waiting for user interest" for now to gather some feedback from our community:
>
> - If you would like to see this feature being implemented, please react to the description with an up-vote (:+1:).
> - If you have a suggestion or want to point out some special cases that needs to be considered, please leave a comment, so we are aware on them.
>
> We would also like to hear about your use cases for the feature to give us a better understanding of potentially implied or explicit requirements.
>
> We will start the implementation based on:
>
> - the number of votes and comments
> - the relevance for the ecosystem
> - availability of alternatives and workarounds
> - and the complexity of the requested feature
>
> We do this because:
>
> - There are plenty of languages/countries out there and we would like to ensure that every method can cover all or almost all of them.
> - Every feature we add to faker has "costs" associated to it:
>   - initial costs: design, implementation, reviews, documentation
>   - running costs: awareness of the feature itself, more complex module structure, increased bundle size, more work during refactors

xDivisionByZerox:
For now, I will swap to [@ST-DDT's suggestion](https://github.com/faker-js/faker/pull/2041#issuecomment-1508150084). We can then discuss single phrases in the code itself.

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Add Comment For User Interest
+        uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `Thanks for feature proposal.
```
xDivisionByZerox:
```suggestion
              body: `**Thank you for your feature proposal.**
```

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,45 @@
...
+
+                We marked it as "waiting for user interest" for now to gather some feedback from our community:
+
+                - If you would like to see this feature being implemented, please react to the description with an up-vote (:+1:).
+                - If you have a suggestion or want to point out some special cases that needs to be considered, please leave a comment, so we are aware on them.
+
+                We would also like to hear about your use cases for the feature to give us a better understanding of potentially implied or explicit requirements.
```
xDivisionByZerox: Shouldn't that already be included in the issue description?

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,45 @@
...
+
+                We will start the implementation based on:
+
+                - the number of votes and comments
```
xDivisionByZerox:
```suggestion
                - the number of votes (:+1:) and comments
```

.github/workflows/comment-issue.yml
ST-DDT: Yes, but only for the person asking for the feature, not others who might want to use it.

```suggestion
                We would also like to hear about other community member's use cases for the feature to give us a better understanding of potentially implied or explicit requirements.
```

matthewmayer: members' not member's

```suggestion
                We would also like to hear about other community members' use cases for the feature to give us a better understanding of potentially implied or explicit requirements.
```

matthewmayer:
Could the bot also add a single upvote to get it started? It's easy to locate it if someone has already added a 👍 otherwise you have to hunt in the emoji menu

xDivisionByZerox:
> Could the bot also add a single upvote to get it started? It's easy to locate it if someone has already added a 👍 otherwise you have to hunt in the emoji menu

Yes, it is possible to add reactions to issues via workflow actions. I need to search up how to do that conditionally but should be easy I guess. Good suggestion 👍 

Should the bot always vote? Like even if the issue already has been upvoted (altho I thought someone is that fast)? Or only react if no upvote is present?

ST-DDT:
> Should the bot always vote?

Yes, to keep it "fair".

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,45 @@
...
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `Thanks for feature proposal.
+
+                We marked it as "waiting for user interest" for now to gather some feedback from our community:
+
+                - If you would like to see this feature being implemented, please react to the description with an up-vote (:+1:).
```
ejcheng:
```suggestion
                - If you would like to see this feature be implemented, please react to the description with an up-vote (:+1:).
```

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,45 @@
...
+                - If you have a suggestion or want to point out some special cases that needs to be considered, please leave a comment, so we are aware on them.
```
ejcheng:
```suggestion
                - If you have a suggestion or want to point out some special cases that need to be considered, please leave a comment so we are aware about them.
```

.github/workflows/comment-issue.yml
ejcheng:
```suggestion
                We would also like to hear about your use cases for the feature to give us a better understanding of your potential implicit or explicit requirements.
```

.github/workflows/comment-issue.yml
xDivisionByZerox: This seems to be a bit more complicated as I thought since:
- PRs and Issues are indeed handled independently from each other (since you can link PRs AND Issues with `#` I thought they are somewhat equal)
- workflows related to PR's require maintainer/member privileges to run

Maybe we can talk about this tomorrow.

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,56 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `**Thank you for your feature proposal.**
+
+              We marked it as "waiting for user interest" for now to gather some feedback from our community:
+
+              - If you would like to see this feature be implemented, please react to the description with an up-vote (:+1:).
+              - If you have a suggestion or want to point out some special cases that need to be considered, please leave a comment, so we are aware about them.
+
+              We would also like to hear about other community members' use cases for the feature to give us a better understanding of your potential implicit or explicit requirements.
```
ejcheng:
```suggestion
              We would also like to hear about other community members' use cases for the feature to give us a better understanding of their potential implicit or explicit requirements.
```

ejcheng (dismissed):

.github/workflows/comment-issue.yml
```diff
@@ -0,0 +1,56 @@
+name: Comment Issue
+
+on:
+  issues:
+    types:
+      - labeled
+
+jobs:
+  add-comment-for-user-interest:
+    if: "github.event.label.name == 's: waiting for user interest'"
```
ST-DDT, line 10: Will this only be executed if the label is added or also when it is removed?

ST-DDT (dismissed):

.github/workflows/comment-issue.yml
xDivisionByZerox: Only when it is added. 
The current triggers are configured to event type `issues` with the active type `labeled`. The active type to remove a label is called `unlabeled`.
You can read more specifications on the [GitHub Actions (Events that trigger workflows) Documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issues)

matthewmayer:
Maybe we could also add a link to an issue search for all issues labelled "awaiting user interest". If a user adds a new issue it might also be nice to know which other issues they consider useful.

xDivisionByZerox:
> [...] If a user adds a new issue it might also be nice to know which other issues they consider useful.

I like that idea. Not only for the user alone but for general awareness. 
I'll suggest a snippet.

.github/workflows/comment-issue.yml
```diff
@@ -41,7 +41,11 @@ jobs:
               - There are plenty of languages/countries out there and we would like to ensure that every method can cover all or almost all of them.
               - Every feature we add to faker has "costs" associated to it:
                 - initial costs: design, implementation, reviews, documentation
-                - running costs: awareness of the feature itself, more complex module structure, increased bundle size, more work during refactors`,
+                - running costs: awareness of the feature itself, more complex module structure, increased bundle size, more work during refactors
+
+              ---
+
+              [View more issues which are waiting for user interest](https://github.com/faker-js/faker/issues?q=is%3Aissue+is%3Aopen+label%3A%22s%3A+waiting+for+user+interest%22+)`,
```
matthewmayer, line 48: looks good to me

I added a couple of issues where there was only an existing PR, but no issue, so they all show up in this list

matthewmayer (approved):

ST-DDT:
Could you please post a screenshot of the result?

xDivisionByZerox:
> Could you please post a screenshot of the result?

image

ST-DDT (approved):

ejcheng (approved):

PR #2044, opened 2023-04-13
feat(datatype)!: reimplement datatype module

Shinigami92:
https://github.com/faker-js/faker/issues/1590#issuecomment-1507328948

This implements all methods that can be compared against the `typeof` operator


src/modules/datatype/index.ts
```diff
@@ -25,71 +25,158 @@ export class DatatypeModule {
...
+    return this.faker.helpers.weightedArrayElement<() => unknown>([
+      {
+        weight: 1,
+        value: () => {
+          // empty function
+        },
+      },
+      {
+        weight: 1,
+        value: () => Promise.resolve(),
```
ST-DDT, line 141: If the purpose of the method is `typeof value === 'function'`, then this is unnecessary.
If you think differently please explain why this is more useful than returing a null, string, number or even an empty object.

src/modules/datatype/index.ts
```diff
@@ -2,13 +2,13 @@ import type { Faker } from '../..';
 import { deprecated } from '../../internal/deprecated';
 
 /**
- * Module to generate various primitive values and data types.
+ * Module to generate values that can be checked against JavaScript's `typeof` operator.
  *
  * ### Overview
  *
- * Most of the methods in this module are deprecated and have been moved to other modules like [`faker.number`](https://next.fakerjs.dev/api/number.html) and [`faker.string`](https://next.fakerjs.dev/api/string.html), see individual entries for replacements.
+ * This module provides methods to generate values that can be checked against JavaScript's `typeof` operator.
```
ST-DDT, line 9: This is redundant with the sentence above. Either exand on this by listing them/their associated methods or remove it.

src/modules/datatype/index.ts
ST-DDT: In a different method chose `new Date()`. Why not return `object()` or `any()`/`unknown()`.

ST-DDT: You could litterally implement this via
`return helpers.objectValue(this)`.

ST-DDT:
Also maybe dont sort the methods for this PR to reduce the diff.
Separate PR.

src/modules/datatype/index.ts
Shinigami92: re-check pls

src/modules/datatype/index.ts
Shinigami92: re-check pls (added below line 9)

Shinigami92:
> Also maybe dont sort the methods for this PR to reduce the diff. Separate PR.

This is done explicitly
We have two sections
1. the typeof in sorted order
2. the old deprecated methods (order not changed)

I suggest to just read/review the new code above the `// deprecated methods` comment
everything below did not change and was not touched

matthewmayer (changes requested):
Any of the methods which previously had a working `options` parameter in 7.4 but are now ignored e.g. array, number should be documented in the Upgrading to v8 guide as breaking changes.

src/modules/datatype/index.ts
```diff
@@ -25,71 +36,234 @@ export class DatatypeModule {
...
-   * @see faker.number.int() for the default precision of `1`
-   * @see faker.number.float() for a custom precision
+   * @since 8.0.0
+   */
+  bigint(): bigint {
+    return this.faker.number.bigInt();
+  }
+
+  /**
+   * Returns a value that will result in `true` when checked against JavaScript's `typeof value === 'boolean'` operator.
```
ST-DDT, line 44: Should we add a hint that this method can also be used for generating booleans as there is no specific boolean module?

src/modules/datatype/index.ts
```diff
@@ -25,71 +36,234 @@ export class DatatypeModule {
...
+   *
+   * **Note:**
+   * A probability of `0.75` results in `true` being returned `75%` of the calls; likewise `0.3` => `30%`.
+   * If the probability is `<= 0.0`, it will always return `false`.
+   * If the probability is `>= 1.0`, it will always return `true`.
+   * The probability is limited to two decimal places.
```
ST-DDT: Is this still true?

src/modules/datatype/index.ts
```diff
@@ -25,71 +36,234 @@ export class DatatypeModule {
...
+   *
+   * @since 8.0.0
+   */
+  function(): () => unknown {
+    return this.faker.helpers.weightedArrayElement<() => unknown>([
+      {
+        weight: 1,
+        value: () => {
+          // empty function
+        },
```
ST-DDT, line 113: This is the same as returning `undefined` / the `undefined` method reference.

src/modules/datatype/index.ts
```diff
@@ -25,71 +36,234 @@ export class DatatypeModule {
...
+              'bigint',
+              'boolean',
+              'function',
+              'number',
+              'object',
+              'string',
+              'symbol',
+              'undefined',
+            ] as const)
+          ](),
```
ST-DDT, line 130: Maybe add a todo comment here, that this list the methods explicitly to avoid the deprecated ones and this should be simplified in v9 to `objectValues(this)()`

test/all_functional.spec.ts
```diff
@@ -5,6 +5,7 @@ import { allFakers, fakerEN } from '../src';
 const IGNORED_MODULES = [
   'definitions',
   'helpers',
+  'datatype',
```
ST-DDT: Why do you ignore the whole module? Instead of only object and undefined?

test/datatype.spec.ts
```diff
@@ -91,145 +85,115 @@ describe('datatype', () => {
...
-                  min: 0,
-                  max: 1.9,
-                  precision: 0.4,
-                })
-              )
-            )
-          ).sort();
-
-          expect(results).toEqual([0, 0.4, 0.8, 1.2, 1.6]);
+      describe('function', () => {
```
ST-DDT: This can probably collapsed into a describe.each or some kind of for-loop.

ST-DDT (changes requested):

src/modules/datatype/index.ts
Shinigami92: I think this is implicit + it was the previous way anyway, so we dont need to "point" folks to that

test/all_functional.spec.ts
Shinigami92: because it failed with other datatype methods like number -> NaN as well

Shinigami92:
For some of the suggestions, lets talk tomorrow when we do the v8.0 dev-weekend

src/modules/datatype/index.ts
ST-DDT: It is going to be the only method that will take arguments in the whole module, so surely this does more than the other methods?!

src/modules/datatype/index.ts
ST-DDT: float() without arguments uses max precision floats don't it?

src/modules/datatype/index.ts
Shinigami92: understood
so I will not point to the method just because, but I will mention that for boolean you have option

src/modules/datatype/index.ts
ST-DDT: This still doesn't explain what is special about `Promise.resolve()`.

docs/guide/upgrading.md
```diff
@@ -217,6 +217,43 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
...
+// NEW
+faker.string.sample(); // 'Zo!.:*e>wR'
+faker.string.sample(5); // '6Bye8'
+```
+
+| Old method              | New method            |
+| ----------------------- | --------------------- |
+| `faker.datatype.string` | `faker.string.sample` |
+
+### Behavior of `faker.datatype` module has changed, as has the scope for which it is to be used
```
matthewmayer: these are hard-breaking changes (ie if you were using the old methods, you need to change your code to get the same behavior) so should be in the top Breaking Changes section

matthewmayer (changes requested):

docs/guide/upgrading.md
ST-DDT: Should we retain support the old options during the deprecation period?

matthewmayer: I think given in faker6 to faker7 various methods were moved from faker.helpers to faker.datatype it's bad if they break again from faker7 to faker8.

Shinigami92: > Should we retain support the old options during the deprecation period?

I would like to vote against this (still open for discussion, it's my personal opinion), because this way we make the intend more clear and tell the user to actually move to the specific module

> these are hard-breaking changes (ie if you were using the old methods, you need to change your code to get the same behavior) so should be in the top Breaking Changes section

good hint :+1: will do

matthewmayer:
The new implementations of these methods seem so different from the old methods that it feels confusing to keep them with the same names. 

I feel if someone proposed "a method which returns a Number, but it could be NaN or POSITIVE_INFINITY so its only useful for checking with typeof" as a new issue it would be tagged as "waiting for user interest". So does it make sense to implement these very limited methods if we dont know if users actually need them?

Shinigami92:
@ST-DDT in the meeting notes we also had defined `unknown(): any of the datatype methods (TS "unknown")`
Can we outsource this to an issue and mark as `awaiting for user interest`?

src/modules/datatype/index.ts
Shinigami92: I enhanced the jsdoc a bit to make more clear that the return value of such a function can be empty, throw an error, return a promise or return a value of any type.
This is helpful when you e.g. just want to generate a function that will be passed to a generic function accepting a callback.
That way within a randomized deterministic test case you will find out via faker that you miss a case for e.g. handling promises and or correctly handle error cases if such a callback would throw something.

I know that users out there should cover such cases on specific base, but for example if you come in to a not so good maintained project and introduce faker, you could e.g. start with such a generic way and from there you can then find out e.g. "oh we missed to handle promise cases".

src/modules/datatype/index.ts
Shinigami92: We should not try to be efficient and fancy here and indirectly knowing that the undefined case surprisingly covers the empty/void case
while reading the internal code it is more clear that we have several cases we want to cover:
1. void -- the simpliest case of all
2. an "unknown" whatever return value
3. an Error
4. a Promise that is successful
5. a Promise that is not successful

test/all_functional.spec.ts
Shinigami92: we added a comment to document why we exclude the whole datatype module

Shinigami92:
> The new implementations of these methods seem so different from the old methods that it feels confusing to keep them with the same names.

We know that, but _sadly_ there is no other option as these methods are explicitly named for what the `typeof` operator supports

> I feel if someone proposed "a method which returns a Number, but it could be NaN or POSITIVE_INFINITY so its only useful for checking with typeof" as a new issue it would be tagged as "waiting for user interest". So does it make sense to implement these very limited methods if we dont know if users actually need them?

Yes this was a team decision on 13. April 2023.

Reasons:
- We dont want to drop the datatype module because we could not find another place for `boolean`
- Then we came up with the idea to make the datatype module valuable for the `typeof` operator

Now it is possible to e.g. use `faker.datatype.number()` to pass to a function that just accepts a `number`.
The test can then check that anykind of number, so also NaN or Infinity, are handled correctly by that function.
This should be mostly (or even only) the case when e.g. you go into a already existing project and introduce faker. Then you can write generic tests with these datatype methods and talk to you project lead and tell them to enhance their error handling or TypeScript definitions (if they are using TypeScript at all).

---

Edit 1:
@matthewmayer now you know the reasons behind, you are very good in pointing out to us when something is missing in the docs
So it would be very welcome if you could help me with that :slightly_smiling_face:

src/modules/datatype/index.ts
ST-DDT: What about a promise that is not yet resolved?

src/modules/datatype/index.ts
```diff
@@ -25,71 +40,237 @@ export class DatatypeModule {
...
+   *
+   * @example
+   * faker.datatype.function() // () => ({})
+   * faker.datatype.function() // () => 55422n
+   * faker.datatype.function() // () => false
+   * faker.datatype.function() // () => Promise.resolve()
+   *
+   * @since 8.0.0
+   */
+  function(): () => unknown {
```
ST-DDT: I think we have to put more thought into this as this isn't really usable with TS.

grafik

````ts
function test() {
  toTest(faker.datatype.function()); // Type-Error
  nestedCall(faker.datatype.unknown()); // Type-Error
}

function toTest(fn: () => number): number {
  return nestedCall(fn());
}

function nestedCall(value: number): number {
  return value + 1;
}
````

We should either reduce this to the most minimal viable implementation: `return () => ?`
Or give this more options, such as should this throw?
What do I expect as result? Passing the result of this method as an argument violates the genericity of the definition and thus behaves the same as using `unknown`.
This might be a viable test case for plain JS applications, but even then the utility of that seems pretty limited to me.
I assume it most cases passing `faker.datatype.number` as a reference would produce better results.

I could imagine that something like this might produce more useful results:

````ts
function<ResultType = unknown, PromiseWrapped extends boolean>( options: {
  errorRate: number;
  returnType: `${ResultType}`
  promise: PromiseWrapped 
}): PromiseWrapped extends true ? Promise<ResultType> : ResultType;
````

But since this is what I came up with in just an hour, there might be more options that should be considered.
For now, we should stick to the minimal implementation or omit the method entirely with some docs stating for `function` you should use the other methods and pass them as a reference directly or build one yourself using them or other methods as basis.
That way we follow the "Open (for extensions and) Closed (for modifications)" principle and breaking any workflows that depend on the temporary behavior.

**Comment by matthewmayer** [comment]
Wouldn't it make the migration path much easier if new names were used for the new methods

For example
faker.datatype.numberType instead of faker.datatype.number?

**Review by ST-DDT — CHANGES_REQUESTED** [review]
I also think, that we might want to move this to v9.0 to let the deprecation sink in for one mayor version, before replacing the methods from one version to another with completely different ones.
(or use new method names as @matthewmayer suggested)

File: src/modules/datatype/index.ts (hunk 5 above)
Comment (Shinigami92) [reply src/modules/datatype/index.ts:113]: > What about a promise that is not yet resolved?

Interesting :thinking:

It looks like `new Promise(() => {})` does return immediately,
so we need a sleep like this
```ts
new Promise((resolve) => {
  setTimeout(resolve, 2**31 - 1);
});
```

_side note_: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value
`setTimeout` has a max delay #TIL

Shinigami92:
I'm thinking about to move this to v10, because in v9 the deprecated methods will be removed. In v10 we could be sure that all users are moved away from these old methods, because they where not available at all anymore.

Another idea would be to define a new module that makes the new aim more clear, like `faker.typeof.number`, `faker.typeof.string` and so on :thinking:

matthewmayer:
I would be in favour of a different module name.

matthewmayer:
Although I think "typeof" is a reserved word so it would have to be something else.

Shinigami92:
> Although I think "typeof" is a reserved word so it would have to be something else.

If I would try to, TS shows:
`'typeof' is not allowed as a variable declaration name. ts(1389)`

src/modules/datatype/index.ts
```diff
@@ -2,82 +2,279 @@ import { deprecated } from '../../internal/deprecated';
...
+        since: '8.0',
+        until: '9.0',
+      });
+    }
+
+    return this.faker.helpers.weightedArrayElement([
+      { weight: 1, value: () => this.faker.number.float() },
+      { weight: 1, value: () => this.faker.number.int() },
+      { weight: 1, value: () => Number.NaN },
+      { weight: 1, value: () => Number.POSITIVE_INFINITY },
```
matthewmayer, line 190: Why two positive infinities?

matthewmayer (changes requested):

matthewmayer:
How about faker.type.*

ST-DDT:
This is probably no longer blocked (but needs a rebase and some thoughts).

PR #2045, opened 2023-04-13
refactor(helpers)!: remove default value from arrayElement

### LINKED ISSUE #219: array-like elements have default arguments which violate genericity
### Describe the bug

I was reviewing https://github.com/faker-js/faker/pull/189/files and I noticed that `randomize` supports generic results, but there's an issue with the arguments having a default.

Consider:

```ts
const a = randomize<number>();
```

This would return a string, which violates the type.

### Reproduction

[TypeScript Playground Link](https://www.typescriptlang.org/play?ssl=7&ssc=16&pln=1&pc=1#code/GYVwdgxgLglg9mABAJwIZgCZwLYwF4CmAPACqIC8iAzlMjGAOYB8AFKsmgJ4BcKBqWMABtOiEgG0AuhUTiA5KjkAaRHIBGy1RDnTUVROADWYOAHckesVICUvMgG8AsAChEfKCGQWOqTuIAMkgDcLgC+Li4QCDSIqDJomDj4xGAg2GoEyKzWIc6R0XBCBAB0QnAMbDlAA)

### Additional Info

_No response_

xDivisionByZerox:
Closes #219.


src/modules/helpers/index.ts
```diff
@@ -788,20 +788,22 @@ export class HelpersModule {
    * @template T The type of the entries to pick from.
    * @param array Array to pick the value from.
    *
+   * @throws If array is empty.
```
matthewmayer: this doesn't show up in the generated API documentation. Either a note about throwing should be added to the main description, or the HTML template should be modified to automatically show `@throws` information.

image

src/modules/helpers/index.ts
```diff
@@ -788,20 +788,22 @@ export class HelpersModule {
...
-    // TODO @Shinigami92 2022-04-30: We want to remove this default value, but currently it's not possible because some definitions could be empty
-    // See https://github.com/faker-js/faker/issues/893
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
     const index =
       array.length > 1 ? this.faker.number.int({ max: array.length - 1 }) : 0;
+    const value = array[index];
+    if (value === undefined) {
+      throw new FakerError('Cannot get value from empty set.');
```
matthewmayer: maybe "array" instead of "set"?

matthewmayer (changes requested):

matthewmayer (changes requested):
regression:
```
faker.helpers.arrayElement(["foo","bar",undefined]) 
```
now throws an error 1/3 of the time.

src/modules/helpers/index.ts
xDivisionByZerox: If this is the case, it is a bigger problem and not related to this PR.
Other methods already have the `@thorws` tag. For example:
- [datatype.number](https://github.com/faker-js/faker/blob/5aa51d8de46e3d9cfbe84ebf7c6130fb13c67846/src/modules/datatype/index.ts#L36)
- [finance.iban](https://github.com/faker-js/faker/blob/5aa51d8de46e3d9cfbe84ebf7c6130fb13c67846/src/modules/finance/index.ts#L960)
- [helpers.fromRegExp](https://github.com/faker-js/faker/blob/5aa51d8de46e3d9cfbe84ebf7c6130fb13c67846/src/modules/helpers/index.ts#L352-L353)

As per our [contribution guidelines (section JSDocs)](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#jsdocs):

> If a function can throw an error (FakerError) you have to include the `@throws` tag with an explanation when an error could be thrown

Shinigami92:
> regression:
> 
> ```
> faker.helpers.arrayElement(["foo","bar",undefined]) 
> ```
> 
> now throws an error 1/3 of the time.

Interesting case we might now have thought about

I think we can use something like this:

image

xDivisionByZerox:
> I think we can use something like this:

Or we simply check the array length on input and throw? Why get that complex...^^

If someone provides an array with `undefined` this will be inferred by the generic and correctly type the return type.

Shinigami92:
> > I think we can use something like this:
> 
> Or we simply check the array length on input and throw? Why get that complex...^^
> 
> If someone provides an array with `undefined` this will be inferred by the generic and correctly type the return type.

Because of 

https://github.com/faker-js/faker/blob/99c16e40c88ae6788eb2bf942fd62c035cbc707e/src/modules/helpers/index.ts#L801-L803

src/modules/helpers/index.ts
ST-DDT: Created a new issue for that:

- https://github.com/faker-js/faker/issues/2048

matthewmayer:
> > > I think we can use something like this:
> > 
> > 
> > Or we simply check the array length on input and throw? Why get that complex...^^
> > If someone provides an array with `undefined` this will be inferred by the generic and correctly type the return type.
> 
> Because of
> 
> https://github.com/faker-js/faker/blob/99c16e40c88ae6788eb2bf942fd62c035cbc707e/src/modules/helpers/index.ts#L801-L803

can't  you just change the condition for throwing from
```js
const value = array[index];
    if (value === undefined) {
      throw new FakerError('Cannot get value from empty set.');
    }
```

to
```js
    if (array.length === 0) {
      throw new FakerError('Cannot get value from empty set.');
    }
const value = array[index];
```

matthewmayer:
To clarify, in Faker 7, the behavior was:

```
> faker.helpers.arrayElement([])
undefined
> faker.helpers.arrayElement()
'b'
```

Do we want it to now throw in both these cases (array is empty, array is undefined)?

xDivisionByZerox:
> can't you just change the condition for throwing from
> 
> ```js
> const value = array[index];
>     if (value === undefined) {
>       throw new FakerError('Cannot get value from empty set.');
>     }
> ```
> 
> to
> 
> ```js
>     if (array.length === 0) {
>       throw new FakerError('Cannot get value from empty set.');
>     }
> const value = array[index];
> ```

Yes and no. This is precisely what I was stating [in my previous comment](https://github.com/faker-js/faker/pull/2045#issuecomment-1508052355). But we have to keep in n mind that this will cause trouble in later versions if we decide to activate more strict typescript features. I'm talking about the [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) rule.

> Do we want it to now throw in both these cases (array is empty, array is undefined)?

Yes, we do want that.
Providing no argument at all doesn't make sense to me. The method is in the helper module so the general consent of "every function has to be callable without arguments" does **NOT** apply here.
If you provide an empty array, typescript correctly infers that the return type is `never`, which basically means "This function never returns anything". This is correct as it throws an error instead f returning.

matthewmayer:
Whether or not it makes sense, passing an undefined array happens in existing places in the code at the moment and ends up with "b". 

We should add tests for various cases:

Undefined array
Empty array
Array which contains undefined as a possible value

Shinigami92:
> Whether or not it makes sense, passing an undefined array happens in existing places in the code at the moment and ends up with "b".
> 
> We should add tests for various cases:
> 
> Undefined array Empty array Array which contains undefined as a possible value

I highly assume that this is the reason why this PR is in "draft", but thanks for also thinking about it 🙂

src/modules/helpers/index.ts
xDivisionByZerox: This method (and thus the error) is also used for:
- helpers.objectValues => object input
- helpers.objectKey => object input
- fake => RegExp input

That's why I decided to use "set". It would be weird to get `Cannot get value from empty set.` if you try to get an object value.

matthewmayer: You could catch and rethrow a more appropriate error in those cases.

xDivisionByZerox: I don't really want to wrap every method with a try-catch-block if I'm honest. Non-centralized error handling is kind of a code smell for me...

matthewmayer: Well I mean that's what we are asking consumers of this method to do now right? 😀

I feel having helpful error messages outweighs making the implementation slightly messier.

xDivisionByZerox: I have to strongly disagree. Even if we try-catch-rethrow, we would have to evaluate that the error is actually the one from `arrayElement`. Else the entire concept of having helpful error messages would be obsolete if we override a fundamental node error with "no empty object allowed". We would need to do that "everywhere". You might know that `arrayElement` is the basic building block for most localized data generation. So yeah, that's a lot of places...

matthewmayer: don't have to rethrow, could have a guard at the top of the `faker.helpers.objectKey` like:
```
objectKey<T extends Record<string, unknown>>(object: T): keyof T {
    if (array.length===0) {
       throw new FakerError('Cannot get key from empty object.');
   }
    return this.arrayElement(array);
  }
```

docs/guide/upgrading.md
```diff
@@ -105,6 +105,17 @@ For more information refer to our [Localization Guide](localization).
...
+
+The methods `faker.helper.arrayElement` and `faker.helper.arrayElements` previously defaulted the `array` argument to a simply string array if non was provided.
+This behavior is no longer supported.
+
+Additionally, by providing an empty array argument (`[]`) the functions previously returned `undefined`.
+This behavior was undesired and has been removed.
+The methods will now throw an `FakerError` instead.
+
+The same thing happens now if you provide an empty object `{}` to `faker.helpers.objectKey` or `faker.helpers.objectValue`.
+
```
matthewmayer, line 119: i think we should have an example of how to fix it if you were reliant on the old behavior.

Something like:
Old
```js
let post = {}
post.tags = faker.helpers.arrayElements(["dogs","cats","fish","horses","sheep"],{min:0, max:3})
post.featuredTag = faker.helpers.arrayElement(post.tags)
```

New
```js
let post = {}
post.tags = faker.helpers.arrayElements(["dogs","cats","fish","horses","sheep"],{min:0, max:3})
post.featuredTag = post.tags.length===0 ? undefined : faker.helpers.arrayElement(post.tags)
```
or
```js
let post = {}
post.tags = faker.helpers.arrayElements(["dogs","cats","fish","horses","sheep"],{min:0, max:3})
try {
  post.featuredTag = faker.helpers.arrayElement(post.tags)
} catch (e) {
  //error thrown, leave post.featuredTag as undefined
}
```

matthewmayer (changes requested):

src/modules/helpers/index.ts
Shinigami92: maybe use dataset instead of array or set to distinguish more from JS Set

test/location.spec.ts
```diff
@@ -39,7 +39,10 @@ const NON_SEEDED_BASED_RUN = 5;
 
 describe('location', () => {
   seededTests(faker, 'location', (t) => {
-    t.itEach('street', 'streetName');
+    t.it('street');
+
+    // todo - missing locale data
```
Shinigami92: vitest has a `t.todo`, has it?
Otherwise I would suggest to follow our TODO pattern, so it is more easy to find when searching TODOs through the whole project

docs/guide/upgrading.md
ST-DDT: Or we define an overload with an option: { allowEmpty: true } that adds `undefined` to the result type.

docs/guide/upgrading.md
```diff
@@ -105,6 +105,49 @@ For more information refer to our [Localization Guide](localization).
 
 The `faker.location.zipCodeByState` method has been deprecated, but will also now throw an error if the current locale does not have a `postcode_by_state` definition.
 
+### Methods will throw on empty data set inputs
+
+The methods `faker.helper.arrayElement` and `faker.helper.arrayElements` previously defaulted the `array` argument to a simply string array if non was provided.
```
matthewmayer: simply to simple, non to none

docs/guide/upgrading.md
```diff
@@ -105,6 +105,49 @@ For more information refer to our [Localization Guide](localization).
...
+The same thing happens now if you provide an empty object `{}` to `faker.helpers.objectKey` or `faker.helpers.objectValue`.
+
+**Old**
+
+```ts
+const allTags = ['dogs', 'cats', 'fish', 'horses', 'sheep'];
+const tags = faker.helpers.arrayElements(allTags, { min: 0, max: 3 });
+// `tags` might be an empty array which was no problem in v7
+const featuredTag = faker.helpers.arrayElement(tags);
+// `featureTag` will will be typed as `string` but could actually be `undefined`
```
matthewmayer: Double "will"

Shinigami92:

docs/guide/upgrading.md
matthewmayer: helper to helpers

matthewmayer (dismissed):

matthewmayer:
Also I feel it still needs more tests. There is no test that actually tests the removal of the default value, or handling arrays containing undefined.

docs/guide/upgrading.md
Shinigami92:
```suggestion
The methods `faker.helpers.arrayElement` and `faker.helpers.arrayElements` previously defaulted the `array` argument to a simple string array if none was provided.
```

docs/guide/upgrading.md
Shinigami92:
```suggestion
// `featureTag` will be typed as `string` but could actually be `undefined`
```

docs/guide/upgrading.md
Shinigami92:
```suggestion
The methods `faker.helpers.arrayElement` and `faker.helpers.arrayElements` previously defaulted the `array` argument to a simple string array if none was provided.
```

xDivisionByZerox:
> There is no test that actually tests the removal of the default value [...]

How would you do this? By removing the default value TypeScript no longer allows calling the function without an argument. This is the expected behavior. We can't unit-test that since TypeScript will already complain.

> [...] or handling arrays containing undefined

Test for nullish values (including `undefined`):
https://github.com/faker-js/faker/blob/909a1e6ea6ac0f2cfa915514e63c3916b74a03a8/test/helpers.spec.ts#L217-L220

matthewmayer:
> > There is no test that actually tests the removal of the default value [...]
> 
> How would you do this? By removing the default value TypeScript no longer allows calling the function without an argument. This is the expected behavior. We can't unit-test that since TypeScript will already complain.
> 
> > [...] or handling arrays containing undefined
> 
> Test for nullish values (including `undefined`):
> 
> https://github.com/faker-js/faker/blob/909a1e6ea6ac0f2cfa915514e63c3916b74a03a8/test/helpers.spec.ts#L217-L220

I'm not sure what the best way is, if you call it from Javascript you don't get a Faker Error you get a TypeError which is a bit unintuitive. Could this be documented better/provide a better error message for JS?

```
faker.helpers.arrayElement()
Uncaught TypeError: Cannot read property 'length' of undefined

> try {faker.helpers.arrayElement([])} catch (e) {console.log(e instanceof FakerError)}
true

> try {faker.helpers.arrayElement()} catch (e) {console.log(e instanceof FakerError)}
false
```

matthewmayer:
> > > There is no test that actually tests the removal of the default value [...]
> > 
> > 
> > How would you do this? By removing the default value TypeScript no longer allows calling the function without an argument. This is the expected behavior. We can't unit-test that since TypeScript will already complain.

How about this:
```ts
it('should throw on an undefined array', () => {
          let undefinedArray: string[];
          expect(() =>
            faker.helpers.arrayElement(undefinedArray)
          ).toThrowError();
        });
```

xDivisionByZerox:
> > > > There is no test that actually tests the removal of the default value [...]
> > > 
> > > 
> > > How would you do this? By removing the default value TypeScript no longer allows calling the function without an argument. This is the expected behavior. We can't unit-test that since TypeScript will already complain.
> 
> How about this:
> 
> ```ts
> it('should throw on an undefined array', () => {
>           let undefinedArray: string[];
>           expect(() =>
>             faker.helpers.arrayElement(undefinedArray)
>           ).toThrowError();
>         });
> ```

That is simply not the expected behavior (on top of being invalid code). TypeScript first. I don't care about JS Users. Sorry for the strong language, but this argument simply annoys me. 

If we want to fully support plain JS, half of the code base requires a refactor, as we only check valid TS types for all method arguments and not the ones that are invalid but still possible in JS.

matthewmayer:
I do care about JavaScript users! I'm not saying we need to add lots of extra tests to handle weird things being passed into methods that typescript prevents. But in this specific case previously if you passed undefined to this method it would return one of a, b or c. Now it doesn't, it throws instead. From a TDD perspective it makes sense to write a test which would fail before this change is merged and pass after it is merged to ensure correctness and prevent future regressions.

src/modules/helpers/index.ts
```diff
@@ -788,16 +788,18 @@ export class HelpersModule {
...
    * faker.helpers.arrayElement(['cat', 'dog', 'mouse']) // 'dog'
    *
    * @since 6.3.0
    */
-  arrayElement<T = string>(
-    // TODO @Shinigami92 2022-04-30: We want to remove this default value, but currently it's not possible because some definitions could be empty
-    // See https://github.com/faker-js/faker/issues/893
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
```
Shinigami92, line 806: NOW I understand what @matthewmayer is talking about! 💡

```suggestion
  arrayElement<T>(array?: ReadonlyArray<T>): T {
```

```ts
if (array == null)
  deprecate({ ... })
```

Shinigami92 (changes requested):

src/modules/helpers/index.ts
ST-DDT: @Shinigami92 I'm confused.
The parameter is now and should be required. You suggestion would allow for invalid usage.

I understand @matthewmayer 's suggestion as adding an explicit test for that case (throw if called without args).

xDivisionByZerox: But this doesn't make sense considering what this PR wants to achieve. I did in fact think they wanted a change like this, but I'll block that decision. 
This IS a breaking change. It is supposed to do precisely that, break your code if you do not adapt. I don't know why we discuss that at all.

docs/guide/upgrading.md
```diff
@@ -105,6 +105,50 @@ For more information refer to our [Localization Guide](localization).
...
 The `faker.location.zipCodeByState` method has been deprecated, but will also now throw an error if the current locale does not have a `postcode_by_state` definition.
 
+### Methods will throw on empty data set inputs
+
+The methods `faker.helpers.arrayElement` and `faker.helpers.arrayElements` previously defaulted the `array` argument to a simple string array if none was provided.
+This behavior is no longer supported, as the default value has been removed.
+You are now required to provide an argument.
+
+Additionally, by providing an empty array argument (`[]`) the functions previously returned `undefined`.
+This behavior was undesired and has been removed.
```
ST-DDT: Maybe mention that this violated the typescript definition regarding the return type.

src/modules/helpers/index.ts
Shinigami92: Why suddenly not make a deprecation for this case, but for all other cases?
Why do we need to make a compile-time breaking change here?

xDivisionByZerox: So we will have them as optional in v8 and make them completely required in v9? I don't this is what @matthewmayer wants. If so, this makes even less sense to me.

Shinigami92: This is what I understand @matthewmayer had requested
If I misunderstood him, yes... I'm with you

I added #2069 and we might want to merge this first, then override/adjust the tests again in this PR
So because we are not releasing a v7.6.1 just for a test, at least it is in the commit history 😆

ST-DDT: We might want to throw a specific error for JS users, that the array is now required.
Because JS users dont note the signature change otherwise and get this "cryptic" error instead:

grafik

Similar to https://github.com/faker-js/faker/blob/4a095dd407e61e8cea4f9953a8886cd87087f54c/src/faker.ts#L508-L512

ST-DDT:
There seems to be some misunderstanding about various peoples intentions.
Lets talk about that in Discord later.

docs/guide/upgrading.md
ST-DDT:
```suggestion
This behavior violated the expected return type of the method.
```

src/modules/helpers/index.ts
```diff
@@ -796,16 +796,24 @@ export class HelpersModule {
    *
    * @param array Array to pick the value from.
    *
+   * @throws If array is empty.
```
```suggestion
   * @throws If `array` is empty.
```

ST-DDT: Or

```suggestion
   * @throws If the given array is empty.
```

src/modules/helpers/index.ts
```diff
@@ -796,16 +796,24 @@ export class HelpersModule {
    *
    * @param array Array to pick the value from.
```
ST-DDT: Optionally:

```suggestion
   * @param array The array to pick the value from.
```

src/modules/helpers/index.ts
```diff
@@ -796,16 +796,24 @@ export class HelpersModule {
...
    * @since 6.3.0
    */
-  arrayElement<T = string>(
-    // TODO @Shinigami92 2022-04-30: We want to remove this default value, but currently it's not possible because some definitions could be empty
-    // See https://github.com/faker-js/faker/issues/893
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array == null) {
+      throw new Error(
```
ST-DDT: Maybe throw a `FakerError`?

src/modules/helpers/index.ts
```diff
@@ -907,6 +913,12 @@ export class HelpersModule {
           max: number;
         }
   ): T[] {
+    if (array == null) {
+      throw new Error(
```
ST-DDT: Maybe throw a `FakerError`?

src/modules/helpers/index.ts
```diff
@@ -796,16 +796,24 @@ export class HelpersModule {
...
    */
-  arrayElement<T = string>(
-    // TODO @Shinigami92 2022-04-30: We want to remove this default value, but currently it's not possible because some definitions could be empty
-    // See https://github.com/faker-js/faker/issues/893
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array == null) {
+      throw new Error(
+        'Calling `faker.helpers.arrayElement()` without arguments is no longer supported. You would have known this if you used TypeScript :)'
```
ST-DDT: While I would love to have this, this could be considered toxic and in violation of our code of conduct.
So please use neutral language for this.

src/modules/helpers/index.ts
```diff
@@ -907,6 +913,12 @@ export class HelpersModule {
           max: number;
         }
   ): T[] {
+    if (array == null) {
+      throw new Error(
+        'Calling `faker.helpers.arrayElements()` without arguments is no longer supported. You would have known this if you used TypeScript :)'
```
ST-DDT: While I would love to have this, this could be considered toxic and in violation of our code of conduct.
So please use neutral language for this.

src/modules/helpers/index.ts
```diff
@@ -907,6 +913,12 @@ export class HelpersModule {
           max: number;
         }
   ): T[] {
+    if (array == null) {
```
ST-DDT, line 918: Should we add a `TODO remove in v9 here`?

src/modules/helpers/index.ts
```diff
@@ -794,18 +794,26 @@ export class HelpersModule {
...
    *
    * @since 6.3.0
    */
-  arrayElement<T = string>(
-    // TODO @Shinigami92 2022-04-30: We want to remove this default value, but currently it's not possible because some definitions could be empty
-    // See https://github.com/faker-js/faker/issues/893
-    array: ReadonlyArray<T> = ['a', 'b', 'c'] as unknown as ReadonlyArray<T>
-  ): T {
+  arrayElement<T>(array: ReadonlyArray<T>): T {
+    if (array == null) {
```
ST-DDT, line 808: Should we add a `TODO remove in v9 here`?

src/modules/helpers/index.ts
```diff
@@ -891,9 +899,7 @@ export class HelpersModule {
    * @since 6.3.0
    */
   arrayElements<T>(
```
ST-DDT, line 902: This doesn't have the `@throws`.

test/helpers.spec.ts
```diff
@@ -206,6 +205,31 @@ describe('helpers', () => {
 
           expect(actual).toBe('hello');
         });
+
+        it('should throw with no arguments', () => {
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
ST-DDT: @Shinigami92 told me that you should use `ts-expect-error` here with a comment why.

test/helpers.spec.ts
```diff
@@ -464,6 +488,16 @@ describe('helpers', () => {
             }
           }
         );
+
+        it('should throw with no arguments', () => {
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
ST-DDT: Same here

test/helpers.spec.ts
```diff
@@ -206,6 +205,31 @@ describe('helpers', () => {
...
+          );
+        });
+
+        describe('should not throw on an array with nullish elements', () => {
+          it.each(['', 0, undefined, null, false])('%s', (nullishValue) => {
+            expect(() =>
+              faker.helpers.arrayElement([nullishValue])
+            ).not.toThrowError();
+          });
+        });
```
ST-DDT, line 231: Please add these two tests also for `arrayElements` if they aren't already there.
(might need some adjustments)

src/modules/helpers/index.ts
xDivisionByZerox: Because the case it currently thows at is not offically supported.

test/helpers.spec.ts
xDivisionByZerox: But `helpers.arrayElements` is not supposed to throw with `[]` as argument. But I can add the second test case.

ST-DDT: > But `helpers.arrayElements` is not supposed to throw with `[]` as argument.

I couldn't find a test for that case, you please add one?

xDivisionByZerox: It already existed before https://github.com/faker-js/faker/blob/f721e8d38ec175f532523ea3a762d44775a8fdf5/test/helpers.spec.ts#L441-L445

ST-DDT (dismissed):

test/helpers.spec.ts
```diff
@@ -206,6 +205,30 @@ describe('helpers', () => {
 
           expect(actual).toBe('hello');
         });
+
+        it('should throw with no arguments', () => {
+          // @ts-expect-error `arrayElement` without arguments is not supported in TypeScript
```
Shinigami92:
```suggestion
          // @ts-expect-error: `arrayElement` without arguments is not supported in TypeScript
```

test/helpers.spec.ts
```diff
@@ -464,6 +487,26 @@ describe('helpers', () => {
             }
           }
         );
+
+        it('should throw with no arguments', () => {
+          // @ts-expect-error `arrayElements` without arguments is not supported in TypeScript
```
Shinigami92:
```suggestion
          // @ts-expect-error: `arrayElements` without arguments is not supported in TypeScript
```

ST-DDT (dismissed):

docs/guide/upgrading.md
```diff
@@ -105,6 +105,50 @@ For more information refer to our [Localization Guide](localization).
 
 The `faker.location.zipCodeByState` method has been deprecated, but will also now throw an error if the current locale does not have a `postcode_by_state` definition.
 
+### Methods will throw on empty data set inputs
+
+The methods `faker.helpers.arrayElement` and `faker.helpers.arrayElements` previously defaulted the `array` argument to a simple string array if none was provided.
+This behavior is no longer supported, as the default value has been removed.
+You are now required to provide an argument.
+
+Additionally, by providing an empty array argument (`[]`) the functions previously returned `undefined`.
```
ejcheng:
```suggestion
Additionally, when passing in an empty array argument (`[]`), the functions previously returned `undefined`.
```

test/helpers.spec.ts
```diff
@@ -886,8 +930,9 @@ describe('helpers', () => {
         });
 
         it('should return undefined if given object is empty', () => {
```
ejcheng:
```suggestion
        it('should throw if given object is empty', () => {
```

test/helpers.spec.ts
```diff
@@ -946,7 +991,7 @@ describe('helpers', () => {
 
         it('does not allow empty array parameters', () => {
```
ejcheng:
```suggestion
        it('should throw with empty array parameters', () => {
```

ejcheng (dismissed):

Shinigami92 (approved):

ST-DDT (approved):

ST-DDT:
> please review again or approve

@Shinigami92 we are not in a hurry.

Shinigami92:
> > please review again or approve
> 
> @Shinigami92 we are not in a hurry.

This was not related with being in a hurry in any case?!
I just wrote a message into the dismiss form that GitHub requires to dismiss stale review blockings

ST-DDT:
> This was not related with being in a hurry in any case?!
> I just wrote a message into the dismiss form that GitHub requires to dismiss stale review blockings

Then please wait with dismissing their review, so they actually have time to review again.
If sufficient time has passed and we actually want to proceed with merging this it is still early enough to dismiss them then.

ejcheng (approved):

PR #2046, opened 2023-04-14
chore: rename generics

Shinigami92:
This PR renamed generics to a convention suggested by one of the TypeScript experts @mattpocock based on the YouTube video: https://www.youtube.com/watch?v=qA65QjWCl60

~~Also based on this poll (https://main.elk.zone/mas.to/@jvhellemond@mastodon.social/110124086297164650) by @jvhellemond `entries` are renamed to `elements` when it is in the context of an array/set~~ _extracted into #2049_

_I only checked `src` folder_

### DIFF
--- a/scripts/generateLocales.ts
+++ b/scripts/generateLocales.ts
@@ -40,9 +40,10 @@ const pathDocsGuideLocalization = resolve(
 );
 
 // Workaround for nameOf<T>
-type PascalCase<S extends string> = S extends `${infer P1}_${infer P2}`
-  ? `${Capitalize<P1>}${PascalCase<P2>}`
-  : Capitalize<S>;
+type PascalCase<TName extends string> =
+  TName extends `${infer Prefix}_${infer Remainder}`
+    ? `${Capitalize<Prefix>}${PascalCase<Remainder>}`
+    : Capitalize<TName>;
 
 type DefinitionType = {
   [key in keyof LocaleDefinition]-?: PascalCase<`${key}Definition`>;
--- a/src/definitions/definitions.ts
+++ b/src/definitions/definitions.ts
@@ -22,8 +22,8 @@ import type { WordDefinition } from './word';
 /**
  * Wrapper type for all definition categories that will make all properties optional and allow extra properties.
  */
-export type LocaleEntry<T extends Record<string, unknown>> = {
-  [P in keyof T]?: T[P] | null;
+export type LocaleEntry<TCategoryDefinition extends Record<string, unknown>> = {
+  [P in keyof TCategoryDefinition]?: TCategoryDefinition[P] | null;
 } & Record<string, unknown>; // Unsupported & custom entries
 
 /**
--- a/src/modules/helpers/unique.ts
+++ b/src/modules/helpers/unique.ts
@@ -57,7 +57,7 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * Generates a unique result using the results of the given method.
  * Used unique entries will be stored internally and filtered from subsequent calls.
  *
- * @template Method The type of the method to execute.
+ * @template TMethod The type of the method to execute.
  *
  * @param method The method used to generate the values.
  * @param args The arguments used to call the method.
@@ -71,14 +71,14 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * @param options.store The store of unique entries. Defaults to `GLOBAL_UNIQUE_STORE`.
  */
 export function exec<
-  Method extends (
+  TMethod extends (
     // TODO @Shinigami92 2023-02-14: This `any` type can be fixed by anyone if they want to.
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ...parameters: any[]
   ) => RecordKey
 >(
-  method: Method,
-  args: Parameters<Method>,
+  method: TMethod,
+  args: Parameters<TMethod>,
   options: {
     startTime?: number;
     maxTime?: number;
@@ -88,7 +88,7 @@ export function exec<
     compare?: (obj: Record<RecordKey, RecordKey>, key: RecordKey) => 0 | -1;
     store?: Record<RecordKey, RecordKey>;
   } = {}
-): ReturnType<Method> {
+): ReturnType<TMethod> {
   const now = new Date().getTime();
 
   const {
@@ -132,7 +132,7 @@ export function exec<
   }
 
   // Execute the provided method to find a potential satisfied value.
-  const result: ReturnType<Method> = method(...args) as ReturnType<Method>;
+  const result: ReturnType<TMethod> = method(...args) as ReturnType<TMethod>;
 
   // If the result has not been previously found, add it to the found array and return the value as it's unique.
   if (compare(store, result) === -1 && exclude.indexOf(result) === -1) {
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -3,9 +3,9 @@
  *
  * @see https://github.com/microsoft/TypeScript/issues/29729#issuecomment-471566609
  */
-export type LiteralUnion<T extends U, U = string> =
-  | T
-  | (U & { zz_IGNORE_ME?: never });
+export type LiteralUnion<TSuggested extends TBase, TBase = string> =
+  | TSuggested
+  | (TBase & { zz_IGNORE_ME?: never });
 
 /**
  * A function that returns a value.
@@ -22,18 +22,18 @@ export type Callable = (
 /**
  * Type that represents a single method/function name of the given type.
  */
-export type MethodOf<ObjectType, Signature extends Callable = Callable> = {
-  [Key in keyof ObjectType]: ObjectType[Key] extends Signature
+export type MethodOf<TObjectType, TSignature extends Callable = Callable> = {
+  [Key in keyof TObjectType]: TObjectType[Key] extends TSignature
     ? Key extends string
       ? Key
       : never
     : never;
-}[keyof ObjectType];
+}[keyof TObjectType];
 
 /**
  * Type that represents all method/function names of the given type.
  */
 export type MethodsOf<
-  ObjectType,
-  Signature extends Callable = Callable
-> = ReadonlyArray<MethodOf<ObjectType, Signature>>;
+  TObjectType,
+  TSignature extends Callable = Callable
+> = ReadonlyArray<MethodOf<TObjectType, TSignature>>;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,6 +48,13 @@ module.exports = defineConfig({
         leadingUnderscore: 'forbid',
         trailingUnderscore: 'forbid',
       },
+      {
+        format: ['PascalCase'],
+        selector: ['typeParameter'],
+        prefix: ['T'],
+        leadingUnderscore: 'forbid',
+        trailingUnderscore: 'forbid',
+      },
     ],
     '@typescript-eslint/no-inferrable-types': [
       'error',
--- a/scripts/apidoc/utils.ts
+++ b/scripts/apidoc/utils.ts
@@ -47,10 +47,10 @@ export function adjustUrls(description: string): string {
   return description.replace(/https:\/\/(next.)?fakerjs.dev\//g, '/');
 }
 
-export function mapByName<T extends { name: string }, V>(
-  input: T[],
-  valueExtractor: (item: T) => V
-): Record<string, V> {
+export function mapByName<TInput extends { name: string }, TValue>(
+  input: TInput[],
+  valueExtractor: (item: TInput) => TValue
+): Record<string, TValue> {
   return input.reduce(
     (acc, item) => ({ ...acc, [item.name]: valueExtractor(item) }),
     {}
--- a/src/locale-proxy.ts
+++ b/src/locale-proxy.ts
@@ -61,21 +61,21 @@ export function createLocaleProxy(locale: LocaleDefinition): LocaleProxy {
  * @param categoryData The module to create the proxy for.
  */
 function createCategoryProxy<
-  CategoryData extends Record<string | symbol, unknown>
+  TCategoryData extends Record<string | symbol, unknown>
 >(
   categoryName: string,
-  categoryData: CategoryData = {} as CategoryData
-): Required<CategoryData> {
+  categoryData: TCategoryData = {} as TCategoryData
+): Required<TCategoryData> {
   return new Proxy(categoryData, {
-    has(target: CategoryData, entryName: keyof CategoryData): boolean {
+    has(target: TCategoryData, entryName: keyof TCategoryData): boolean {
       const value = target[entryName];
       return value != null;
     },
 
     get(
-      target: CategoryData,
-      entryName: keyof CategoryData
-    ): CategoryData[keyof CategoryData] {
+      target: TCategoryData,
+      entryName: keyof TCategoryData
+    ): TCategoryData[keyof TCategoryData] {
       const value = target[entryName];
       if (typeof entryName === 'symbol' || entryName === 'nodeType') {
         return value;
@@ -97,5 +97,5 @@ function createCategoryProxy<
 
     set: throwReadOnlyError,
     deleteProperty: throwReadOnlyError,
-  }) as Required<CategoryData>;
+  }) as Required<TCategoryData>;
 }
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -732,7 +732,7 @@ export class HelpersModule {
   /**
    * Returns the result of the callback if the probability check was successful, otherwise `undefined`.
    *
-   * @template T The type of result of the given callback.
+   * @template TResult The type of result of the given callback.
    *
    * @param callback The callback to that will be invoked if the probability check was successful.
    * @param options The options to use. Defaults to `{}`.
@@ -745,8 +745,8 @@ export class HelpersModule {
    *
    * @since 6.3.0
    */
-  maybe<T>(
-    callback: () => T,
+  maybe<TResult>(
+    callback: () => TResult,
     options: {
       /**
        * The probability (`[0.00, 1.00]`) of the callback being invoked.
@@ -755,7 +755,7 @@ export class HelpersModule {
        */
       probability?: number;
     } = {}
-  ): T | undefined {
+  ): TResult | undefined {
     if (this.faker.datatype.boolean(options)) {
       return callback();
     }
@@ -965,7 +965,7 @@ export class HelpersModule {
    *
    * This does the same as `objectValue` except that it ignores (the values assigned to) the numeric keys added for TypeScript enums.
    *
-   * @template EnumType Type of generic enums, automatically inferred by TypeScript.
+   * @template T Type of generic enums, automatically inferred by TypeScript.
    *
    * @param enumObject Enum to pick the value from.
    *
@@ -981,11 +981,11 @@ export class HelpersModule {
    *
    * @since 8.0.0
    */
-  enumValue<EnumType extends Record<string | number, string | number>>(
-    enumObject: EnumType
-  ): EnumType[keyof EnumType] {
+  enumValue<T extends Record<string | number, string | number>>(
+    enumObject: T
+  ): T[keyof T] {
     // ignore numeric keys added by TypeScript
-    const keys: Array<keyof EnumType> = Object.keys(enumObject).filter((key) =>
+    const keys: Array<keyof T> = Object.keys(enumObject).filter((key) =>
       isNaN(Number(key))
     );
     const randomKey = this.arrayElement(keys);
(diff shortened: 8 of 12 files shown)


src/definitions/definitions.ts
```diff
@@ -22,9 +22,10 @@ import type { WordDefinitions } from './word';
 /**
  * Wrapper type for all definition categories that will make all properties optional and allow extra properties.
  */
-export type LocaleEntry<T extends Record<string, unknown>> = Partial<T> &
-  // Unsupported & custom entries
-  Record<string, unknown>;
+export type LocaleEntry<TRecord extends Record<string, unknown>> =
```
ST-DDT: Note to self:

- Rename `LocaleEntry` to `LocaleCategoryDefinition` or something.

ST-DDT: Maybe name it `TBuiltInCategoryDefinition`? (Maybe something to be reconsidered when renaming the type itself)

ST-DDT (commented):
I'm really torn on this one.

- I like using useful names
- I don't like the TName convention
- I don't think it adds value for methods that only do `T*` -> `T`
- I think `TValue`/`Value` does not transport any additional meaning over `T`

ST-DDT:
I watched the video and he explicitly mentions that `T` is fine as long as you don't have multiple (RSTUV...) ones.

As for TPrefixed: I see what argument he is trying to make, but if you cannot remember that `NotPrefixed` is your generic, then there are probably more important issues such as bad naming or a too complex method.

Shinigami92:
Why might need to activate https://typescript-eslint.io/rules/naming-convention/#enforce-that-type-parameters-generics-are-prefixed-with-t

matthewmayer:
T is prefixed with T :)

I feel this is unnecessary. `<T>` is very commonly understood in both Typescript and from other languages like Java. If i see `<TRecord>` i wonder if you are doing something special and different.

xDivisionByZerox (changes requested):
I sadly can't reject the only request changes. Can I request to change everything back?
All of this is overkill. There are good situations to get verbose with your generic names. The ones changed in this PR do not belong to that category.

Shinigami92:
> I'm really torn on this one.
> 
> * I like using useful names

Me either :+1: 
Even while working on this PR, I now more and more respect this naming convention and I'm tempted to use it for all my projects, because you can directly and easily read what the generic type is aimed for instead of first reading the code and try to find it out.

> * I don't like the TName convention

Do you have a reason for that?

> * I don't think it adds value for methods that only do `T*` -> `T`

We have one example for when `T` is the result/return value of a function and one for when it is for the input.
Now with this change it is directly readable from the code for what the generic type is used for.

Yeah okay we could discuss this specific case and maybe just use better names from the start like `Input` and `Result`, but the `T` prefix makes it better in two ways:
1. it does not conflict with other types like if you would have an interface called `Result` (for what ever reason)
2. you can directly find the generic types vs the other types/interfaces used in "more complex" code

> * I think `TValue`/`Value` does not transport any additional meaning over `T`

_read above_

---

> T is prefixed with T :)
> 
> I feel this is unnecessary. `<T>` is very commonly understood in both Typescript and from other languages like Java. If i see `<TRecord>` i wonder if you are doing something special and different.

1. I'm not tempted to follow old ways of doing something just because it was ever done like that without arguments for pro or con
2. Yes, maybe we need to find a better name instead of `TRecord`, and @ST-DDT already pointed to that: https://github.com/faker-js/faker/pull/2046#discussion_r1166569752

---

> I sadly can't reject the only request changes. Can I request to change everything back? All of this is overkill. There are good situations to get verbose with your generic names. The ones changed in this PR do not belong to that category.

These are really very generic (pun not intended) reasons without any real valuable arguments :confused: 
You say `There are good situations to get verbose with your generic names` but doesn't come up with an example but just the statement.
Please substantiate your statement.

scripts/generateLocales.ts
```diff
@@ -40,9 +40,10 @@ const pathDocsGuideLocalization = resolve(
 );
 
 // Workaround for nameOf<T>
-type PascalCase<S extends string> = S extends `${infer P1}_${infer P2}`
-  ? `${Capitalize<P1>}${PascalCase<P2>}`
-  : Capitalize<S>;
+type PascalCase<TString extends string> =
```
Shinigami92: I'm not sure what `S` was holding before
is it just a "string"? is it a "name"? are there any suggestions how we could make it more clear what this actually holds?

/cc @ST-DDT

src/definitions/definitions.ts
Shinigami92: I'm not totally satisfied with `BuiltIn` as these `CategoryDefinition` could also be filled with custom data, can it?
Did you maybe mean something like `TFakerCategoryDefinition` or just a bit more verbose `TCategoryDefinition`?

src/internal/merge.ts
```diff
@@ -1,12 +1,12 @@
...
  *
- * @template T The type of the elements in the arrays.
+ * @template TArgument The type of the elements in the arrays.
  *
  * @internal
  *
  * @param args The arrays to merge.
  */
-export function mergeArrays<T>(...args: T[][]): T[] {
+export function mergeArrays<TArgument>(...args: TArgument[][]): TArgument[] {
```
Shinigami92: Are these maybe `TElement` instead of `TArgument`?
I'm currently also a bit confused about the `[][]` :eyes:
it might be better read as `Array<TArgument[]>` (or even `ReadonlyArray<TArgument[]>` :thinking:), but this might conflict with `@typescript-eslint/array-type` (_not tested yet_)

Shinigami92:
Addition: This PR aims to reduce the cognitive load of reading the code / using the methods

ST-DDT:
> > * I don't like the TName convention
> 
> Do you have a reason for that?

Sure, it creates an arbitrary rule to name something in a special way inconsistently.
I could flip this around and ask why don't you prefix **both** types and interfaces with `I`?
What makes a method local type different from a file local (not exported) one or exported ones?

**EDIT:**

You could even use `G` as prefix, because it refers to a **G**eneric type.
Using **T** is part of that very convention that you try to abolish.

**EDIT2:**

"prefix **both** types and interfaces with `I`" to avoid conflicts during `type` <-> `interface` refactors.
If the distinction is important, than a separate prefix would be as well.

scripts/generateLocales.ts
xDivisionByZerox: I call cap. If you use a higher order type and don't know what the generic is supposed to be that's bad... like from your side, and you should probably not work with those types in that case.

Furthermore, If we assume you had no idea what `S` was supposed to be, how can you say in all seriousness that now that you named it `TString` (which literally can still mean anything) know what it is supposed be? Renaming the generic doesn't provide any additional context in this example.

If you really wanted a verbose generic name you should go for `StringToMakePascalCase` or something. Then I could at least believe that you now understand the meaning of the generic type.

ST-DDT:
> but the `T` prefix makes it better in two ways:
> 
> 1. it does not conflict with other types like if you would have an interface called `Result` (for what ever reason)

But what if you have an exported type `TResult` "for what ever reason"? Ambiguous naming is prone to collisions.

> 2. you can directly find the generic types vs the other types/interfaces used in "more complex" code

If the generic type is simple you don't need a full name in the first place.
Both generics and external code can get pretty complex, I often put this complexity into the methods generics, so that TS resolves the result instead of displaying `Whatever<T>`. But it is very easy to outsource the generic complexity to a separate type. And if I can outsource or inline it, why would I name it specially?

src/internal/merge.ts
xDivisionByZerox: > I'm currently also a bit confused about the T[][]

That is a personal preference. I for example cannot read `Array<TArgument[]>`.
I read the first one as: "I have a **generic element**; in an array; times 2 => 2-dimensional array".
While The second one is all over the place in my thought process: "I have an array; that consists of **generic elements**; in an array".

xDivisionByZerox:
> I could flip this around and ask why don't you prefix both types and interfaces with I?

This! Excatly this!!!
Why stop there use the correct prefix for **I**nterfaces and **T**ypes.

```ts
export interface IMyInterface {
  // ... definition
}

export type TMyType = // definition
```

I mean it cool, no? You just have to type `I` or `T` and VSCode suggests all possible interfaces or types.

ST-DDT:
> > I feel this is unnecessary. `<T>` is very commonly understood in both Typescript and from other languages like Java. If i see `<TRecord>` i wonder if you are doing something special and different.
> 
> 1. I'm not tempted to follow old ways of doing something just because it was ever done like that without arguments for pro or con

I'm not tempted to drop proven ways of doing something just because of a single video without arguments for pro or con.

Pro simple generics:

- It is a well understood concept
- It doesn't bloat the code, with `TThisIsSomethingReallyUnimportantName` over "hey this is something dynamic called `T`".
- It doesn't require us to teach every new contributor, that **we** have this special rule for generics.

> 2. Yes, maybe we need to find a better name instead of `TRecord`, and @ST-DDT already pointed to that:

If you have a suggestion, we can discuss it. Currently the increased length of the name only takes more time to read till the extends clause that already tells me it is a `Record`. And if you try to guess what a method does without at least having a look at the method name/definition is stupid. If the method or generic is poorly named, then a longer name won't help either.

ST-DDT:
> > I sadly can't reject the only request changes. Can I request to change everything back? All of this is overkill. There are good situations to get verbose with your generic names. The ones changed in this PR do not belong to that category.
> 
> These are really very generic (pun not intended) reasons without any real valuable arguments 😕 You say `There are good situations to get verbose with your generic names` but doesn't come up with an example but just the statement. Please substantiate your statement.

Yes, the reasoning is a bit fussy, but IMO it is a valid point. 
If you want a reason: I assign specific names if the **name** carries value, not just because I can.

xDivisionByZerox:
> These are really very generic (pun not intended) reasons without any real valuable arguments 😕
You say There are good situations to get verbose with your generic names but doesn't come up with an example but just the statement.
Please substantiate your statement.

Sure. Have a look at the implementation of a JS Generator:

```ts
interface Generator<T = unknown, TReturn = any, TNext = unknown> extends Iterator<T, TReturn, TNext> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
    return(value: TReturn): IteratorResult<T, TReturn>;
    throw(e: any): IteratorResult<T, TReturn>;
    [Symbol.iterator](): Generator<T, TReturn, TNext>;
}
```

This makes to have verbose names as we have multiple generics.

Again, as long as you don't have multiple generic values you don't need to get creative. Please check the video again (time 1:59min to 2:01min):

> For simple types like this I will generally keep this as the single letter T

scripts/generateLocales.ts
ST-DDT: `PascalCase` is a notation. If you don't know what `PascalCase` is, then you have other problems (As mentioned earlier).

If you jump in the middle of the definition and don't instantly know what S has, then start from the beginning.
Adding longer names such as `TString extends string` increases the amount I have to read, without adding value, that you would gain anyway, if you look at the code from the start. 

IMO in this case especially, the entire inner type definition is entirely irrelevant, because the name of the type carries all information you need about what the type does.

**EDIT**

Renaming `P` to `Part`/`Prefix`+`Remainer` might add some value if you want to understand how the type does what it does. But then we should also be more explicit with our type name `Camel_Case2PascalCase`. But then again this type isn't exported and only used in a single location and that even outside of the actual project source code.
We can do this rename, if that helps you understand the code better.

xDivisionByZerox:
@mattpocock May you give me/us some insight if this PR is what you wanted to achieve when creating your video [How to Name your Types](https://youtu.be/qA65QjWCl60?t=118)?

src/internal/merge.ts
ST-DDT: Element is better than Argument. But it adds weight to the type, that is entirely unimportant.
You want to merge arrays, is important to have a dedicated name for the elements?
No, the method implementation doesn't really care and just shoves them together. It doesn't even appear in the implementation.
You add 20% more code (105->129 chars) that doesn't tell you anymore about it than `T[]`.

ST-DDT:
> Addition: This PR aims to reduce the cognitive load of reading the code / using the methods



> These are really very generic [...] reasons without any real valuable arguments 😕 You say [...] but doesn't come up with an example but just the statement. Please substantiate your statement.

No seriously: Reading more text for the same amount of information usually doesn't "reduce the cognitive load".

src/definitions/definitions.ts
ST-DDT: > I'm not totally satisfied

Me neither, it was just an attempt in adding some value to the currently proposed "name".

scripts/generateLocales.ts
```diff
@@ -40,9 +40,10 @@ const pathDocsGuideLocalization = resolve(
 );
 
 // Workaround for nameOf<T>
-type PascalCase<S extends string> = S extends `${infer P1}_${infer P2}`
-  ? `${Capitalize<P1>}${PascalCase<P2>}`
-  : Capitalize<S>;
+type PascalCase<TValue extends string> =
+  TValue extends `${infer Prefix}_${infer Remainder}`
+    ? `${Capitalize<Prefix>}${PascalCase<Remainder>}`
+    : Capitalize<TValue>;
```
Shinigami92: Should we load this from https://github.com/sindresorhus/type-fest/blob/9feb8c89be9a0f2f688bf2f497230298a8e2472e/source/pascal-case.d.ts#L36 instead of defining it manually in our code base?
We just would need to keep care that we don't use types that would need to be exported tied together from our `src/` code, because I don't know how this would behave when users use them but `type-fest` is just a `devDependencies` in `@faker-js/faker` and therefore doesn't get published with it.

Shinigami92:
I would like to remember @xDivisionByZerox and @ST-DDT to read our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md) again.

https://github.com/faker-js/faker/blob/5e421b0cdfd272e9bb96468e80f41152a88d7fbf/CODE_OF_CONDUCT.md?plain=1#L5

Please don't blame me down or make assumptions about **my potential** `level of experience`.

https://github.com/faker-js/faker/blob/5e421b0cdfd272e9bb96468e80f41152a88d7fbf/CODE_OF_CONDUCT.md?plain=1#L9-L13

Please try to write a bit more `respectful` comments especially in terms of `differing viewpoints and experiences`.
Using `welcoming and inclusive language` can also help finding a better path to a more constructive base.

> I call cap. If you use a higher order type and don't know what the generic is supposed to be that's bad... like from your side, and you should probably not work with those types in that case.

(https://github.com/faker-js/faker/pull/2046#discussion_r1167310461)

For example we should prevent using exclamations like `I call crap` and just leave it out of the discussion.
Furthermore we could write with more `welcoming and inclusive language` like:
```diff
- I call cap. If you use a higher order type and don't know what the generic is supposed to be that's bad... like from your side, and you should probably not work with those types in that case.
+ If you use a higher order type, the generic template parameter is only supposed to transfer the given type the user is providing through it. However we should keep care on your side to not over-engineer types if they are getting to complex to used and the intention is not clear in the first place.
```

This would have prevented me from getting attacked on a personal level.

---

Please also think about PRs as a Pull **Request** which is indirectly always a proposal and needs to be accepted by the team/maintainers of a project, but can also get rejected / not merged.

@ST-DDT was telling me the last weeks that I should not open issues for something when I directly can open a PR, because (and I agree with that) in a PR we can discuss on a per code-line level.
In addition to that it prevents managing a flood of more and more issues.

---

I checked the template type of the TS definition `Array/ReadonlyArray` and it is indeed just and only `T`.
Also the `interface Generator<T = unknown, TReturn = any, TNext = unknown>` example is a good showcase for a real-world example.

So yes we should discuss two points:
1. where is it okay to just use `T`, because I gave the argument above that `TValue`/`TInput` vs `TResult` makes it more clear to read the intent out of the name of the generic type without reading left or right from it. But in our cases we have these examples only on separate places and therefore not the case where we would have two generics next to each other.
2. Should we follow the suggestion of @mattpocock and use the `T` prefix for template types in general, or should we not use it, but maybe come to a compromise where we use the non-prefixed but more expressive variant:
   ```diff
   - maybe<T>(callback: () => T, options: ...): T | undefined;
   + maybe<Result>(callback: () => Result, options: ...): Result | undefined;
   ```

ST-DDT:
> I would like to remember @xDivisionByZerox and @ST-DDT to read our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md?rgh-link-date=2023-04-15T08%3A05%3A43Z) again.

Thanks for pointing this out. I'll try to improve on this.
But I would also like to ask yourself to do the same regarding your prior rude comments.
Lets talk about this in person as to not further get off topic with the discussion.

Shinigami92:
From yesterdays discussion in Discord voice chat, I found out you are not at all against renaming some of these generic template parameters to some more verbose names, but you are especially more against the `T` prefix itself.

So maybe a good solution is to find the compromise between single letter values and the T prefixed verbose naming as we only want the more verbose naming if it is not instantly obvious what T is (like for an array).
So if we have e.g. an input (that is not an array element), a value, a result etc. we would prefer meaningful names.
We already did good for some cases like instead of using `T` or `C` we use `CategoryData` (https://github.com/faker-js/faker/blob/385d4fae52956778b9d856ef2a6f39a779a7d07f/src/locale-proxy.ts#L59)
This makes it (and maybe I just speck for me as myself, but I'm also one of the contributor of this project that needs to work with the code and mostly read not write it) by far more readable especially when it comes into the body of this function where the generic type is reused below
https://github.com/faker-js/faker/blob/385d4fae52956778b9d856ef2a6f39a779a7d07f/src/locale-proxy.ts#L70

If you would just read `get(_, entryName: keyof T): T[keyof T] {`, yes this is shorter, but you are absolutely loosing the context of what is going on.

The T prefix naming convention helped me to find all these places in the whole code base, not only the `src` code, because it is an automated eslint rule.
Maybe we could redefined this setting to our own needs to accept `T` as a single value, or a at least two length letter PascalCase named typename.
That way, we would get support from eslint for future PR reviews.

ST-DDT:
**Team Decision**

- We want to name our generics `T` or `TPrefixed`.
- We only use `TPrefixed` if this adds value to the declaration, decided on a case by case role
  - if there are multiple generic paramters

Shinigami92:
@faker-js/members @faker-js/maintainers Please write new review comments on each remaining rename with a reason why or why not it should be named T / not T, if it hurts you in your soul

scripts/generateLocales.ts
```diff
@@ -40,9 +40,10 @@ const pathDocsGuideLocalization = resolve(
 );
 
 // Workaround for nameOf<T>
-type PascalCase<S extends string> = S extends `${infer P1}_${infer P2}`
-  ? `${Capitalize<P1>}${PascalCase<P2>}`
-  : Capitalize<S>;
+type PascalCase<TValue extends string> =
```
ST-DDT: Maybe `TName` or just `T`?

src/modules/helpers/unique.ts
```diff
@@ -83,14 +83,15 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`
  * @param options.store The store of unique entries. Defaults to `GLOBAL_UNIQUE_STORE`.
  */
 export function exec<
-  Method extends (
+  TMethod extends (
     // TODO @Shinigami92 2023-02-14: This `any` type can be fixed by anyone if they want to.
+    // TODO christopher 2023-02-14: This `any` type can be fixed by anyone if they want to.
```
ST-DDT:
```suggestion
```

src/modules/person/index.ts
```diff
@@ -20,14 +20,18 @@ export type SexType = `${Sex}`;
  *
  * @returns Definition based on given sex.
  */
-function selectDefinition<T>(
+function selectDefinition<TValue>(
```
ST-DDT: IMO this can just be `T`.

src/utils/types.ts
```diff
@@ -3,9 +3,9 @@
  *
  * @see https://github.com/microsoft/TypeScript/issues/29729#issuecomment-471566609
  */
-export type LiteralUnion<T extends U, U = string> =
-  | T
-  | (U & { zz_IGNORE_ME?: never });
+export type LiteralUnion<TUnion extends TBase, TBase = string> =
```
ST-DDT: Maybe use `TSuggested` instead of `TUnion`, but not sure?

scripts/generateLocales.ts
Shinigami92: I will go with `TName`

src/modules/person/index.ts
Shinigami92: I think that's okay, because this is just internal code anyway

ST-DDT: A single `TValue` without context is the most generic name possible here and does not add any value.

ST-DDT (dismissed):

ST-DDT:
This has gotten stale over the last merges.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2049, opened 2023-04-14
chore: reword to elements

Shinigami92:
Extracted out of #2046 to separate it

Based on this poll (https://main.elk.zone/mas.to/@jvhellemond@mastodon.social/110124086297164650) by @jvhellemond `entries` are renamed to `elements` when it is in the context of an array/set

This is due to an `entry` is mostly known as a `[key, value]` tuple from `Object.entries()`

### DIFF
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -554,7 +554,7 @@ export class HelpersModule {
   /**
    * Takes an array and randomizes it in place then returns it.
    *
-   * @template T The type of the entries to shuffle.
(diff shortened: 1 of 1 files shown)


xDivisionByZerox (approved):
I do not have strong feelings about this. Items, Entries, Elements - I use those interchangeably. If more users lean toward "elements" I see no downside changing those descriptions.

matthewmayer (approved):

ST-DDT (approved):
Same for me.

PR #2050, opened 2023-04-15
docs: document @throws parameter in API docs

### LINKED ISSUE #2048: Document `@throws` in api docs

Currently some methods have jsdocs `@throws` tags. However these don't show up on the api docs page.

### Suggested solution

Add them to the api docs generation.

### Alternative

Don't show them.

### Additional context

Found in: https://github.com/faker-js/faker/pull/2045#discussion_r1166375473

matthewmayer:
fix #2048

Example:

Screenshot 2023-04-15 at 17 08 06

### DIFF
--- a/scripts/apidoc/signature.ts
+++ b/scripts/apidoc/signature.ts
@@ -25,6 +25,7 @@ import {
   extractSeeAlsos,
   extractSince,
   extractSourcePath,
+  extractThrows,
   joinTagParts,
 } from './typedoc';
 import { pathOutputDir } from './utils';
@@ -160,12 +161,16 @@ export function analyzeSignature(
   const deprecated = deprecatedMessage
     ? mdToHtml(deprecatedMessage)
     : undefined;
+  const throwsMessage = extractThrows(signature);
+  const throws = throwsMessage ? mdToHtml(throwsMessage, true) : undefined;
+
   return {
     name: methodName,
     description: mdToHtml(toBlock(signature.comment)),
     parameters: parameters,
     since: extractSince(signature),
     sourcePath: extractSourcePath(signature),
+    throws,
     returns: typeToText(signature.type),
     examples: mdToHtml(`${code}ts\n${examples}${code}`),
     deprecated,
--- a/scripts/apidoc/typedoc.ts
+++ b/scripts/apidoc/typedoc.ts
@@ -287,7 +287,7 @@ export function extractDeprecated(
 }
 
 /**
- * Checks if the given reflection throws.
+ * Extracts the "throws" tag from the provided signature.
  *
  * @param reflection The reflection to check.
  *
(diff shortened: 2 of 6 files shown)


scripts/apidoc/typedoc.ts
```diff
@@ -286,6 +286,18 @@ export function extractDeprecated(
   return deprecated.length === 0 ? undefined : deprecated;
 }
 
+/**
+ * Checks if the given reflection throws.
```
ST-DDT: This doesnt really check whether it throws instead it extract the value.
I missed the docs for extractDeprecation during the refactor.

scripts/apidoc/signature.ts
```diff
@@ -160,12 +161,16 @@ export function analyzeSignature(
   const deprecated = deprecatedMessage
     ? mdToHtml(deprecatedMessage)
     : undefined;
+  const throwsMessage = extractThrows(signature);
+  const throws = throwsMessage ? mdToHtml(throwsMessage, true) : undefined;
```
ST-DDT, line 165: The deprecation doesnt have the true flag.
IMO we should have both consistently. What is the difference in output?

test/scripts/apidoc/signature.example.ts
```diff
@@ -1,4 +1,5 @@
 import type { Casing, ColorFormat } from '../../../src';
+import { FakerError } from '../../../src';
```
Shinigami92: Here in test it doesn't make much difference, but we import `FakerError` from its exact script on all other places as inside the src it can somehow result in bundle errors

```suggestion
import { FakerError } from '../../../src/errors/faker-error';
```

ST-DDT (commented):
Nice.

scripts/apidoc/signature.ts
matthewmayer: depends if you want to have the HTML inline or not. Since the throws is shown like 

`<p> Throws: THROWSMESSAGE</p>` you don't want another wrapping `<p>`

scripts/apidoc/typedoc.ts
matthewmayer: updated comment

scripts/apidoc/typedoc.ts
```diff
@@ -286,6 +286,18 @@ export function extractDeprecated(
   return deprecated.length === 0 ? undefined : deprecated;
 }
 
+/**
+ * Extracts the "throws" tag from the provided signature.
+ *
+ * @param reflection The reflection to check.
+ *
+ * @returns The message explaining the error in the @throws tag, otherwise `undefined`.
```
ST-DDT:
```suggestion
 * @returns The message explaining the conditions when this method throws. Or `undefined` if it does not throw.
```

Not sure about the wording.

ST-DDT (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #2051, opened 2023-04-15
refactor(location)!: rename definition street to street_pattern

### LINKED ISSUE #2020: Merge faker.location.street() and faker.location.streetName()

- https://next.fakerjs.dev/api/location.html#street
- https://next.fakerjs.dev/api/location.html#streetname

It also causes issues when a locale does not have a street_name_pattern or street_name locale file.

### Suggested solution

````ts
faker.location.streetName({ allowGenerated: boolean = true });
````

### Alternative

- Rename the methods to account for them generating real and generated street names respectively.

### Additional context

Part of #1346

- #1346

### PR DESCRIPTION
**Pull request by Shinigami92**
close #2020

### CODE REVIEW

**Comment by Shinigami92** [comment]
1. Is this a breaking change? Do we need to add the `!` in PR title?
2. How should I name this PR/changelog entry?
3. What do we need to do for documentation?

**Comment by Shinigami92** [comment]
> 1. Is this a breaking change? Do we need to add the `!` in PR title?

We decided yes, just to have this at the top in the changelog

> 2. How should I name this PR/changelog entry?

Still unsure how to name this
Especially the internal stuff is not really relevant

> 3. What do we need to do for documentation?

No response yet

File: src/definitions/location.ts [hunk 1]
```diff
@@ -79,7 +79,7 @@ export type LocationDefinitions = LocaleEntry<{
   /**
    * The patterns to generate street names.
    */
-  street: string[];
+  street_pattern: string[];
```
matthewmayer, line 82: other definitions have plural patterns, e.g. company/name_patterns, person/bio_patterns, person/last_name_patterns

matthewmayer (dismissed):

src/definitions/location.ts
Shinigami92: I suggest to name everything consistently to singular because this can better read when using the fake function

`faker.helpers.faker('I'm living in {{location.street_pattern}}')`

We need a separate PR to rename everything to singular (but I think this is obvious ^^)

matthewmayer: You shouldn't be using street_pattern/s directly though, you should just write

faker.helpers.faker('I'm living in {{location.street}}') no?

Shinigami92: Why do write `city` instead of `cities` but `x_patterns` instead of `x_pattern`?

matthewmayer: company/name_patterns started the trend and the rest copied.

Shinigami92: So we will revert that and use singular consistently
Luckily this is currently all only in alpha v8

src/modules/location/index.ts
```diff
@@ -193,18 +193,31 @@ export class LocationModule {
...
    * Returns a random localized street name.
    *
+   * @see faker.location.street()
+   *
    * @example
    * fakerDE.location.streetName() // 'Cavill Avenue'
    *
    * @since 8.0.0
+   *
+   *
```
```suggestion
   *
```

src/locales/th/location/street_pattern.ts
ST-DDT, line 1: This diff confused me, because `src/locales/tr/location/street.ts` (TR) → `src/locales/th/location/street_pattern.ts` (TH), but the TH file is newly created and the moved TR file just looks identical.

But everything is fine.

ST-DDT (commented):
If I checked correctly the following locales have `street_name`s, but no pattern using it.

- az
- es_MX
- he
- ka_GE
- ko
- lv
- nb_NO
- ro
- ru
- sv
- uk

I will create a new issue for this: https://github.com/faker-js/faker/issues/2062

src/definitions/location.ts
Shinigami92: - #2059

ST-DDT (dismissed):

matthewmayer:
why is this marked as breaking in the PR desc?

ST-DDT:
Because it breaks your code if you use `fake("{{location.street}}")`.

matthewmayer:
Won't {{location.street}} grab the method not the definitions?

Shinigami92:
> Won't {{location.street}} grab the method not the definitions?

I expect yes

ST-DDT:
Oh yeah. You are right.

Shinigami92:

matthewmayer:
I need some more time to test this.

xDivisionByZerox (approved):

matthewmayer:
Looks ok but shouldn't be marked breaking if it isn't.

matthewmayer:
On https://next.fakerjs.dev/guide/upgrading.html#faker-address-changed-to-faker-location perhaps the row for streetName should be updated to reflect that if you are upgrading from faker.address.streetName then you should migrate straight to faker.location.street instead of faker.location.streetName?

ST-DDT:
Should we split this PR into two?
Both changes are indepent from each other.
https://github.com/faker-js/faker/pull/2070#issuecomment-1514700947

xDivisionByZerox:
> Should we split this PR into two? Both changes are indepent from each other. [#2070 (comment)](https://github.com/faker-js/faker/pull/2070#issuecomment-1514700947)

Would be a welcome change but not required since we already reviewed it IMO

ST-DDT:
> Would be a welcome change but not required since we already reviewed it IMO

The migration guide needs updating anyway so it doesnt really matter IMO.

Shinigami92:
#2071 should be merged first, so this can be rebased

ST-DDT (approved):

ejcheng (approved):

PR #2053, opened 2023-04-15
refactor(datatype): deprecate datetime

Shinigami92:
Deprecate `datatype.datetime` in favor of `date.between`


src/modules/date/index.ts
```diff
@@ -993,4 +993,61 @@ export class DateModule {
...
+   *    as maximum generated date. Defaults to `4102444800000`.
+   *
+   * @example
+   * faker.date.datetime() // '2089-04-17T18:03:24.956Z'
+   * faker.date.datetime(1893456000000) // '2022-03-28T07:00:56.876Z'
+   * faker.date.datetime({ min: 1577836800000, max: 1893456000000 }) // '2021-09-12T07:13:00.255Z'
+   *
+   * @since 8.0.0
+   */
+  datetime(
```
ST-DDT: Should we name this `sample`?

src/modules/date/index.ts
```diff
@@ -993,4 +993,61 @@ export class DateModule {
 
     return new Date(this.faker.number.int({ min, max }));
   }
+
+  /**
+   * Returns a Date object using a random number of milliseconds since
+   * the [Unix Epoch](https://en.wikipedia.org/wiki/Unix_time) (1 January 1970 UTC).
```
ST-DDT: IMO this description is confusing.

src/modules/date/index.ts
```diff
@@ -993,4 +993,61 @@ export class DateModule {
...
+   *
+   * @param options Max number of milliseconds since unix epoch or options object.
+   * @param options.min Lower bound for milliseconds since base date.
+   *    When not provided or smaller than `-8640000000000000`, `1990-01-01` is considered
+   *    as minimum generated date. Defaults to `631152000000`.
+   * @param options.max Upper bound for milliseconds since base date.
+   *    When not provided or larger than `8640000000000000`, `2100-01-01` is considered
+   *    as maximum generated date. Defaults to `4102444800000`.
```
ST-DDT: Should we adjust these defaults?

src/modules/datatype/index.ts
```diff
@@ -172,6 +172,8 @@ export class DatatypeModule {
    * faker.datatype.datetime({ min: 1577836800000, max: 1893456000000 }) // '2021-09-12T07:13:00.255Z'
    *
    * @since 5.5.0
+   *
+   * @deprecated Use `faker.date.datetime()` instead.
```
ST-DDT: This also needs an `@see`

ST-DDT (commented):
Please include this in the upgrading guide.

src/modules/date/index.ts
Shinigami92: @xDivisionByZerox @matthewmayer what are your opinions for that?

src/modules/date/index.ts
Shinigami92: What exactly is confusing to you?
I did not read it at all but just copy pasted the entire method just over to its new home without touching anything

src/modules/date/index.ts
Shinigami92: I did not read the docs or implementation at all but just copy pasted the entire method just over to its new home without touching anything

I suggest to do such changes in a separate PR, should we? /cc @xDivisionByZerox

src/modules/date/index.ts
matthewmayer: What does this method offer that you can't do with `faker.date.between()`?

ST-DDT: Between requires boundaries, this method doesn't.

src/modules/date/index.ts
ST-DDT: > using a random number of milliseconds since the [Unix Epoch](https://en.wikipedia.org/wiki/Unix_time) (1 January 1970 UTC)

The random number of MS in epoch time are totally irrelevant for the user. 
(1 January 1970 UTC) is the base for `Date` but this method doesn't use it as some kind of limit.

> I did not read it at all but just copy pasted the entire method just over to its new home without touching anything

IMO we should also fix issues with the description/definition/implementation while we are at it.

src/modules/date/index.ts
matthewmayer: faker.datatype.datetime({min, max}) seems to behave identically to faker.date.between({from, to}). 

Could probably change the deprecation method to point to `faker.date.between` instead; enhance the documentation of `between` to clarify that if a number is passed that its milliseconds since Unix epoch; and explain how to set from and to if you want the same no-parameter behavior as old `faker.datatype.datetime`()

matthewmayer: between also works without boundaries, just has different defaults.

ST-DDT:
**Team Decision**

We won't move it over because it behaves the same as `faker.date.between`, but we will adjust the deprecation message accordingly.

docs/guide/upgrading.md
```diff
@@ -230,6 +230,10 @@ faker.number.float({ max: 100, precision: 0.01 }); // 35.21
 The method `faker.datatype.array` has been deprecated and will be removed in v9.
 If you need an array of useful values, you are better off creating your own one using `faker.helpers.multiple`.
 
+### `faker.datatype.datetime` moved to `faker.date.between`
```
```suggestion
### `faker.datatype.datetime` deprecated in favor of `faker.date.between`
```

ST-DDT (dismissed):

xDivisionByZerox (dismissed):

ST-DDT (dismissed):

xDivisionByZerox (dismissed):

ST-DDT:
The PR title + labels needs updating.

src/modules/datatype/index.ts
```diff
@@ -166,12 +166,16 @@ export class DatatypeModule {
...
+   * @see faker.date.between()
+   *
    * @example
    * faker.datatype.datetime() // '2089-04-17T18:03:24.956Z'
    * faker.datatype.datetime(1893456000000) // '2022-03-28T07:00:56.876Z'
    * faker.datatype.datetime({ min: 1577836800000, max: 1893456000000 }) // '2021-09-12T07:13:00.255Z'
    *
    * @since 5.5.0
+   *
+   * @deprecated Use `faker.date.between()` instead.
```
matthewmayer: It should explain how to set the parameters of between if you want the old bounds for datetime.

Shinigami92: I considered this as not to complex and using `faker.date.between(options)` doesn't provide any value
The runtime deprecation warning is already providing more info if necessary, but the deprecation JSDoc tag follows our current "standards" we already have in the project

@ST-DDT what do you say to this?

ST-DDT: Considering how happy you and I were to have someting to basically copy and paste as a replacement for the image categories. I dont see a reason not to be a little more verbose with the deprecating message. And a simple mapping doesnt look excesive to me either.
Also the deprecation message shown in the IDE is more important than some runtime deprecation log as there is no direct context between the calling line and the log message.

ST-DDT (approved):

ejcheng (approved):

Shinigami92:
@matthewmayer @ST-DDT due to the requested reviews from your side to use between instead of its own function, it slipped into the next branch that we dont have a method anymore that just provides a random date but without arguments/parameters
We need to fix that before releasing v8.0

PR #2055, opened 2023-04-15
refactor(finance): rename mask to maskedNumber

Shinigami92:
references https://github.com/faker-js/faker/issues/1346#issuecomment-1407724091

### DIFF
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -217,6 +217,8 @@ export class FinanceModule {
    * @param parens Whether to use surrounding parenthesis. Defaults to `true`.
    * @param ellipsis Whether to prefix the numbers with an ellipsis. Defaults to `true`.
    *
+   * @see faker.finance.maskedNumber()
+   *
    * @example
    * faker.finance.mask() // '(...9711)'
    * faker.finance.mask(3) // '(...342)'
--- a/src/modules/finance/index.ts
+++ b/src/modules/finance/index.ts
@@ -224,8 +224,31 @@ export class FinanceModule {
    * faker.finance.mask(3, false, false) // '298'
    *
    * @since 2.0.1
+   *
+   * @deprecated Use `faker.finance.maskedNumber` instead.
    */
-  mask(length?: number, parens?: boolean, ellipsis?: boolean): string;
+  mask(length?: number, parens?: boolean, ellipsis?: boolean): string {
+    deprecated({
+      deprecated: 'faker.finance.mask',
+      proposed: 'faker.finance.maskedNumber',
+      since: '8.0',
+      until: '9.0',
+    });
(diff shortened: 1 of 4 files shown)


src/modules/finance/index.ts
```diff
@@ -274,18 +292,17 @@ export class FinanceModule {
...
+   * faker.finance.maskNumber(3) // '(...342)'
+   * faker.finance.maskNumber({ length: 3 }) // '(...342)'
+   * faker.finance.maskNumber({ length: 3, parens: false }) // '...236'
+   * faker.finance.maskNumber({ length: 3, parens: false, ellipsis: false }) // '298'
    *
-   * @since 2.0.1
+   * @since 8.0.0
    */
-  mask(
+  maskNumber(
```
ST-DDT: Maybe name this `maskedNumber`? Because it doesn't mask a given number, but masks a new one.

ST-DDT (approved):

ST-DDT (changes requested):
Please include this in the upgrading guide.

src/modules/finance/index.ts
```diff
@@ -197,8 +198,31 @@ export class FinanceModule {
    * faker.finance.mask(3, false, false) // '298'
    *
    * @since 2.0.1
+   *
+   * @deprecated Use `faker.finance.maskedNumber` instead.
```
ST-DDT, line 230: This should also have a `@see` tag.

matthewmayer:
> Please include this in the upgrading guide.

i dont think the upgrading guide has to mention every deprecation, only
- breaking changes
- major module renames like name to person
- unexpected behavior changes

if a method is just changed to a better name within the same module, and the deprecation workflow is used, i think its fine to just mention this as a breaking change in the v9 upgrading guide when it is actually removed.

ST-DDT:
> i dont think the upgrading guide has to mention every deprecation

Not individually, but maybe as a list of renamed methods.

ST-DDT (dismissed):

xDivisionByZerox (dismissed):

Shinigami92:
@matthewmayer @ST-DDT I cannot find it right now, but I thought we have an open TODO/issue to re-evaluate the upgrading guideline in total before releasing v8.0
Then we should check everything over and unify some stuff and so on (in the upgrading.md)

matthewmayer:
It would be good to have some guidelines on what needs to be in upgrading and what doesn't per my other comment.

Shinigami92:
> It would be good to have some guidelines on what needs to be in upgrading and what doesn't per my other comment.

I suggest moving this into its own discussion, discord chat or even issue

matthewmayer:
Let's make an issue

xDivisionByZerox (approved):

ST-DDT (approved):

PR #2056, opened 2023-04-15
chore: consistent jsdocs tag formatting

ST-DDT:
Formatted the code to have:
- exactly one newline between different tags
- exactly zero trailing newlines between same tags

I'm not sure whether there should be a newline between `@template` and `@param`, but we have (at least) one place in the code where `@internal` jumps between the two.

Assumed tag sequence:

````ts
        'jsdoc/sort-tags': [
          'error',
          {
            tagSequence: [
              { tags: ['template'] },
              { tags: ['internal'] },
              { tags: ['param'] },
              { tags: ['returns'] },
              { tags: ['throws'] },
              { tags: ['see'] },
              { tags: ['example'] },
              { tags: ['since'] },
              { tags: ['default'] },
              { tags: ['deprecated'] },
            ],
          },
        ],
````

Requires: https://github.com/gajus/eslint-plugin-jsdoc/pull/1017 to work (slightly buggy ATM).

### DIFF
--- a/src/faker.ts
+++ b/src/faker.ts
@@ -46,7 +46,6 @@ import { mergeLocales } from './utils/merge-locales';
  *
  * faker.person.firstName(); // 'John'
  * faker.person.lastName(); // 'Doe'
- *
  * @example
  * import { Faker, es } from '@faker-js/faker';
  * // const { Faker, es } = require('@faker-js/faker');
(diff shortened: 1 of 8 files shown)


Shinigami92 (approved):

xDivisionByZerox (approved):
Good that you have done this, but it would be even better to automate this. Should we consider opening an issue at [gajus/eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) for the "new line between different tags" and "no new line for same tags" rule we have?

matthewmayer (approved):

ST-DDT:
> Good that you have done this, but it would be even better to automate this. Should we consider opening an issue at [gajus/eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) for the "new line between different tags" and "no new line for same tags" rule we have?

I linked a PR that already adds that in the description. 
I used that to do the changes and provide feedback to the PR author about this feature.

Issue #1834, opened 2023-02-13
Turn on strictNullChecks in tsconfig


ST-DDT:
https://github.com/faker-js/faker/pull/651 partially activated strict mode, this flag was omitted at that time and should probably be turned on.

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/tsconfig.json#L11

Related issues:

- https://github.com/faker-js/faker/pull/651
- https://github.com/faker-js/faker/issues/1835
- https://github.com/faker-js/faker/issues/1836
- https://github.com/faker-js/faker/issues/1837
- https://github.com/faker-js/faker/issues/1838

Shinigami92:
This raised around 200 errors

Mostly coming from `this.faker.definitions.xy` might be `undefined`

So theoretically this is related to
- #893

ST-DDT:
I consider this blocked by #893

xDivisionByZerox:
@Shinigami92 are you working on this? Is this blocked by #2102?

Shinigami92:
> @Shinigami92 are you working on this? Is this blocked by #2102?

Yes and yes 👍

ST-DDT:
I started a PR to fix this: https://github.com/faker-js/faker/pull/2435

- https://github.com/faker-js/faker/pull/2435

PR #2058, opened 2023-04-16
refactor(types): rename locale definition types to singular

ST-DDT:
Rename:

- `LocaleDefinition` -> `LocaleDefinitions`
- `LocaleEntry` -> `LocaleCategoryDefinitions` (cause it is only for categories)

I'm not sure about the new names and whether we should use singular or plural names for these.


xDivisionByZerox (dismissed):

Shinigami92:
Especially `locale: LocaleDefinitions | LocaleDefinitions[]` is a good indicator to me that we should use singular (I'm sorry :shrug:)

ST-DDT:
We should decide this before releasing v8.0

Shinigami92:
> We should decide this before releasing v8.0

I vote for singular

Shinigami92:
@xDivisionByZerox What do you think is the better option?

xDivisionByZerox:
I prefer the singular variant (`LocaleDefinition`, `LocaleCategoryDefinition`). 

You could say that the object, represented by the type, is a collection of keys - similar to a `Map`. 
While you would most likely write the value of a `Map` in the plural form (`const books = new Map();`), I don't see the argument being fitting for our specific case. A locale definition may be a collection of suggested definitions but it still applies to one locale only (so singular).

Thats at least how I see it.

ST-DDT:
Changed to singular

matthewmayer:
If this is considered a breaking change then it should be mentioned in migration guide.

ST-DDT:
> If this is considered a breaking change then it should be mentioned in migration guide.

No, it is not because the old aliases are still exported.

Shinigami92 (dismissed):
LGTM

src/definitions/definitions.ts
```diff
@@ -28,28 +28,26 @@ export type LocaleEntry<T extends Record<string, unknown>> = {
...
+  location?: LocationDefinition;
+  lorem?: LoremDefinition;
+  music?: MusicDefinition;
+  person?: PersonDefinition;
+  phone_number?: PhoneNumberDefinition;
+  science?: ScienceDefinition;
+  system?: SystemDefinition;
+  vehicle?: VehicleDefinition;
+  word?: WordDefinition;
+} & Record<string, Record<string, unknown>>;
```
xDivisionByZerox: Why did you remove the `| undefined` from the custom modules?

ST-DDT: This was not intentional. I noticed an error showing up in the locale-proxy tests that is caused by that `undefined` and tried addressing this but not as part of this PR and more for #1834 / #2102.

Reverted

Shinigami92 (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2060, opened 2023-04-16
refactor(location): merge state and stateAbbr

xDivisionByZerox:
References https://github.com/faker-js/faker/issues/1346#issuecomment-1407728227


src/modules/location/index.ts
```diff
@@ -338,17 +338,91 @@ export class LocationModule {
...
+   * faker.location.state(false) // 'Florida'
+   * faker.location.state(true) // 'AR'
    *
    * @since 8.0.0
    */
-  state(): string {
-    return this.faker.helpers.arrayElement(
-      this.faker.definitions.location.state
-    );
+  state(useAbbreviation?: boolean): string;
```
ST-DDT: IMO we should only use options.

src/modules/location/index.ts
```diff
@@ -338,17 +338,91 @@ export class LocationModule {
...
+  /**
+   * Returns a random localized state, or other equivalent first-level administrative entity for the locale's country such as a province or region.
+   *
+   * @param options An options object. Defaults to `{}`.
+   * @param options.useAbbreviation Whether to return an abbreviation. Defaults to `false`.
+   *
+   * @example
+   * faker.location.state() // 'Mississippi'
+   * faker.location.state({ useAbbreviation: false }) // 'Florida'
```
matthewmayer: `abbreviated: true` would be simpler

src/modules/location/index.ts
```diff
@@ -338,17 +338,91 @@ export class LocationModule {
...
+          useAbbreviation?: boolean;
+        } = {}
+  ): string {
+    if (typeof options === 'boolean') {
+      options = { useAbbreviation: options };
+    }
+
+    const { useAbbreviation = false } = options;
+    const stateDataSet = useAbbreviation
+      ? this.faker.definitions.location.state_abbr
```
matthewmayer, line 366: Not all states have abbreviations. In that case would you expect the parameter to be ignored and it falls back to providing a full state name, or an error to be thrown?

matthewmayer (changes requested):

src/modules/location/index.ts
xDivisionByZerox: It is expected to throw after #2045 has been merged.

ST-DDT: I assume it will also throw after #2004

src/modules/location/index.ts
ST-DDT: I think we use these abbr or similar parameters elsewhere and should rename them there as well.

Shinigami92: - #2061

src/modules/location/index.ts
```diff
@@ -338,17 +338,36 @@ export class LocationModule {
...
    * Returns a random localized state, or other equivalent first-level administrative entity for the locale's country such as a province or region.
    *
+   * @param options An options object. Defaults to `{}`.
+   * @param options.abbreviated Whether to return an abbreviation. Defaults to `false`.
+   *
    * @example
    * faker.location.state() // 'Mississippi'
-   * fakerEN_CA.location.state() // 'Saskatchewan'
-   * fakerDE.location.state() // 'Nordrhein-Westfalen'
+   * faker.location.state(false) // 'Iowa'
```
```suggestion
```

src/modules/location/index.ts
```diff
@@ -338,17 +338,36 @@ export class LocationModule {
...
+       *
+       * @default false
+       */
+      abbreviated?: boolean;
+    } = {}
+  ): string {
+    if (typeof options === 'boolean') {
+      options = { abbreviated: options };
+    }
+
```
```suggestion
```

ST-DDT (changes requested):

src/modules/location/index.ts
```diff
@@ -338,17 +338,31 @@ export class LocationModule {
   /**
    * Returns a random localized state, or other equivalent first-level administrative entity for the locale's country such as a province or region.
    *
+   * @param options An options object. Defaults to `{}`.
+   * @param options.abbreviated Whether to return an abbreviation. Defaults to `false`.
+   *
    * @example
    * faker.location.state() // 'Mississippi'
-   * fakerEN_CA.location.state() // 'Saskatchewan'
-   * fakerDE.location.state() // 'Nordrhein-Westfalen'
```
ST-DDT: I highlight this diff to raise awareness, that while we already mentioned in the docs that this method refers to the `first-level administrative entity`, there might still be some confusion what this might actually be. Should we keep these for reference or should we let the users uncover this themselves?

matthewmayer: I think we should keep. It's helpful to see a variety of  examples from different locales

xDivisionByZerox: I just want to say that I do not see why that would be necessary, but whatever. I revered the change.

matthewmayer (approved):

ST-DDT (approved):

matthewmayer (approved):
Looks good though we should come back and review this works as expected after #2045 and #2004 are done.

Shinigami92 (approved):

PR #2067, opened 2023-04-17
docs: strikethrough deprecated methods on api index

matthewmayer:
Strikes through deprecated methods on API index page so its easier to see the non-deprecated methods

Example:
Screenshot 2023-04-16 at 20 12 59


docs/api/ApiIndex.vue
```diff
@@ -165,6 +169,9 @@ h3 {
   color: var(--vp-c-text-code);
   transition: color 0.5s;
 }
+.api-groups ul a.deprecated {
```
ST-DDT, line 173:
```suggestion

.api-groups ul a.deprecated {
```

ST-DDT (commented):
I somewhat like it, but I'm not sure whether the user easily understands that the methods there are deprecated/what the strike-through should mean.

Also I'm not sure whether this adds any kind of benefit. If you notice a method doing what you search for then you click on it. If it happens to be deprecated, you will read the deprecation hint pointing to the right method.

matthewmayer:
The idea is if you are skimming looking for the right method you can focus on the ones that are non-deprecated.

ST-DDT:
**Team Decision**

- We want this.

xDivisionByZerox (dismissed):

docs/api/ApiIndex.vue
```diff
@@ -101,7 +101,11 @@ onUnmounted(() => window.removeEventListener('keydown', apiSearchFocusHandler));
           </h3>
           <ul>
             <li v-for="h of item.headers" :key="h.anchor">
-              <a :href="item.link + '#' + slugify(h.anchor)">{{ h.text }}</a>
+              <a
+                :href="item.link + '#' + slugify(h.anchor)"
+                :class="h.deprecated ? 'deprecated' : ''"
```
ST-DDT: Should we simplify this to:

```suggestion
                :class="{ deprecated: h.deprecated }"
```

?

ST-DDT (dismissed):
Looks good to me.

docs/api/ApiIndex.vue
Shinigami92: yes, we should

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2069, opened 2023-04-17
test(helpers): add regression prevention test

Shinigami92:
reference https://github.com/faker-js/faker/pull/2045#issuecomment-1511113037

/ping @matthewmayer @xDivisionByZerox 

---

This PR just adds tests cases for v7.6.0's arrayElement for the edge cases

Now we could merge this PR before #2045 and then override/enhance these tests in #2045 afterwards


matthewmayer:
if we go this route it would be good to add the other cases too from #2045 ?

an array with nullish elements
an empty array

Shinigami92:
> if we go this route it would be good to add the other cases too from #2045 ?
> 
> an array with nullish elements an empty array

done

xDivisionByZerox:
I will directly implement this in #2045

PR #2070, opened 2023-04-19
refactor(location): deprecate cityName

### LINKED ISSUE #2021: Merge faker.location.city() and faker.location.cityName()

### Suggested solution

````ts
````

### Alternative

### Additional context

Part of #1346

- #1346

xDivisionByZerox:
Part of #2021.

### DIFF
--- a/docs/guide/upgrading.md
+++ b/docs/guide/upgrading.md
@@ -167,7 +167,7 @@ The `faker.address.*` methods will continue to work as an alias in v8 and v9, bu
 | `faker.address.buildingNumber`      | `faker.location.buildingNumber`      |
 | `faker.address.cardinalDirection`   | `faker.location.cardinalDirection`   |
 | `faker.address.city`                | `faker.location.city`                |
-| `faker.address.cityName`            | `faker.location.cityName`            |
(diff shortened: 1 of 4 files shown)


ST-DDT:
The other PR https://github.com/faker-js/faker/pull/2051 also adjusts the locale data.

xDivisionByZerox:
> The other PR #2051 also adjusts the locale data.

I know but wanted to prevent merge conflicts as well as breaking changes in THIS PR. I had a really hard time actually finding the deprecation itself in the other PR.

matthewmayer:
Do you plan to make city and cityName return the same data in v8.0? Or you just deprecate cityName in v8.0 and merge the locale data in v9?

xDivisionByZerox:
This is the same as #2051. 
`location.city` already is a superset of `location.cityName`. As described in https://github.com/faker-js/faker/issues/2021#issuecomment-1499260452 we simply deprecate the `*Name` implementation for now. We can think about adding a `generated` option parameter (to only get generated or real cities) in the future if the community requests.

Shinigami92 (approved):

ST-DDT (approved):

Issue #1791, opened 2023-01-29
Investigate why importing anything from '@faker-js/faker' directly causes all of faker to be bundled.


ST-DDT:
### Clear and concise description of the problem

Importing anything directly from `@faker-js/faker` causes the bundler to contain all of faker in the resulting bundle.

Reproduction:

https://github.com/faker-js/playground/blob/main/playgrounds/vite-cjs/src/App.vue

Add:

````ts
import { FakerError } from "@faker-js/faker";

const natureImageUrl = ref(faker.image.nature() + new FakerError("test"));
````

That will increase the bundle size from 0.6MB to 3.6MB.

Maybe try with sideEffect: false + refactorings.

Shinigami92:
So to explain what a side effect is:

When a function modifies the state of something outside its scope, then it is declared as a side effect

For example:

https://github.com/faker-js/faker/blob/64f480d214dc0bd173b0921025c8a1ad76d12c0f/src/modules/helpers/unique.ts#L144

https://github.com/faker-js/faker/blob/64f480d214dc0bd173b0921025c8a1ad76d12c0f/src/modules/helpers/unique.ts#L104

https://github.com/faker-js/faker/blob/64f480d214dc0bd173b0921025c8a1ad76d12c0f/src/modules/helpers/unique.ts#L9

The `helpers.unique` function takes `GLOBAL_UNIQUE_STORE` as a default value and this is defined out of scope of the function itself, but is modified inside it in line 144.
So if the default is taken for `unique`, then the modification to `GLOBAL_UNIQUE_STORE` is a side effect.

---

_I'm not aware of any other side effects right now_

ST-DDT:
Would moving this to a class field fix that issue?

ST-DDT:
**Team Decision**

We will try to investigate this for v8.0

- Remove global unique store in favor of a faker (helperModule) instance bound one

Shinigami92:
@xDivisionByZerox pointed made a hint in meeting 2023-04-20 that we could move the variable `allLocales` into the file `src/locales/index.ts` itself

https://github.com/faker-js/faker/blob/bf67a219365d63c9430d4935056752859eaeaa8f/src/index.ts#L41

This should be tested and could work, so that when loading the file, no variable assignment will be made in src/index.ts itself, but only in the underlying file on demand

Shinigami92:
@evanw could you help us with this bug?

Please checkout 
```bash
git clone git@github.com:faker-js/faker.git
[...]

PR #2072, opened 2023-04-20
chore(helpers): move unique store into faker instance

Shinigami92:
references #1791


ST-DDT (commented):
Should we merge this only if it helps with the tree shake-ability or anyway?

Shinigami92:
> Should we merge this only if it helps with the tree shake-ability or anyway?

I'm also not so sure

xDivisionByZerox:
Are these tests still relevant in their current setup (do they still have to be separated)?

https://github.com/faker-js/faker/blob/3fc7bf1b24ed67696b4824abcd9fa14af43cb66d/test/helpers.spec.ts#L1201-L1203

src/modules/helpers/index.ts
```diff
@@ -83,6 +83,22 @@ function getRepetitionsBasedOnQuantifierParameters(
...
+   */
+  private uniqueStore: Record<RecordKey, RecordKey> = {};
+
+  /**
+   * Global exclude list of results.
+   * Defaults to nothing excluded.
+   *
+   * @internal
+   */
+  private uniqueExclude: RecordKey[] = [];
```
ST-DDT: AFAICT this can be inlined because it is readonly.

Shinigami92:
> Are these tests still relevant in their current setup (do they still have to be separated)?
> 
> https://github.com/faker-js/faker/blob/3fc7bf1b24ed67696b4824abcd9fa14af43cb66d/test/helpers.spec.ts#L1201-L1203

I see what you mean, but let's ignore it for now as the method is deprecated anyway
The tests are just executed once instead of multiple times, but the globals are still bound internally to the faker instance

xDivisionByZerox (approved):
> I see what you mean, but let's ignore it for now as the method is deprecated anyway
The tests are just executed once instead of multiple times, but the globals are still bound internally to the faker instance

Yes, but you could instantiate a new Faker object in every test which wasn't possible before.

I don't really like to ignore it, but the fact that it is deprecated seems reasonable.

ST-DDT (approved):
The jsdocs aren't ideal, but I don't know how to phrase them any better.

PR #2073, opened 2023-04-20
fix: move allLocales into all.ts

Shinigami92:
references #1791


ST-DDT:
@Shinigami92 Is this something you are still working on? If not please close this.

PR #2075, opened 2023-04-21
feat(locale): add city names for th

matthewmayer:
Adds city names for `th`. This stops `fakerTH.location.city()` sometimes throwing when #2004 is merged, as `th` had no city patterns, so fell back to en patterns, but sometimes that would select a pattern with a city_prefix, which was empty in `th`.


src/locales/th/location/city.ts
```diff
@@ -0,0 +1 @@
+export default ['{{location.city_name}}'];
```
Shinigami92, line 1: Maybe we should first merge the pattern renaming, and then already have this file named city_pattern

matthewmayer: there's no PR for that yet, but can wait for it if you prefer.

ST-DDT: This PR was first so lets merge it first.

ST-DDT:
This needs updating.

matthewmayer:
Test failure seems unrelated see https://github.com/faker-js/faker/issues/2080

ST-DDT:
> Test failure seems unrelated see #2080

Yeah, one of the rare cases where a float() is an int.

ST-DDT (dismissed):
Are the city prefixes/suffixes not applicable to th in general or did you remove them because we dont need them for the current pattern?

matthewmayer:
> Are the city prefixes/suffixes not applicable to th in general or did you remove them because we dont need them for the current pattern?

Both

ST-DDT:
If it is not applicable, then keeping the data as [] shows to the other developers that this is intentionally absent.
Also the [] does not really add big memory or source code overhead, so I think we should keep them.

matthewmayer:
My assumption was that was only needed for definitions which have public methods. Now that citysuffix and cityprefix are no longer methods, seems like they could just be omitted for locales that don't need them?

xDivisionByZerox:
> [...] Now that citysuffix and cityprefix are no longer methods, seems like they could just be omitted for locales that don't need them?

Yes they are indeed no longer accessible via public methods, BUT if the city suffixes/prefixes simply don't exist in the th locale they should be left in. This allows us to track which data is actually missing and which are not applicable. This will become easier after #2078.

matthewmayer:
Sure, restored

Shinigami92 (dismissed):

Shinigami92 (approved):

xDivisionByZerox (approved):

ST-DDT (approved):

PR #2078, opened 2023-04-21
refactor(locale): use null as not applicable

ST-DDT:
Created based on:

- https://github.com/faker-js/faker/pull/2004#discussion_r1174088271

> I mean that's literally what `null` is for. From the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null):
> 
> > The `null` value represents the intentional absence of any object value.

Instead of relying on magic values such as `[]`, `{}`, `''`, ... we should use `null` as explicitly absent values.


ejcheng (dismissed):

Shinigami92 (dismissed):

xDivisionByZerox:
What is the expected behavior for:

```ts
const result = mergeLocale([
 { 
    location: { 
      city: null,
      country: null,
    },
  },
 { 
    location: { 
      city: ['cityA'],
    },
  },
 { 
    location: { 
      country: ['countyA'],
    },
  }
]);

console.log(result); // ?
```

Is there anything we need to take care of in merging locales and `null` values?

Shinigami92:
```ts
{ 
  location: { 
    city: null,
    country: null,
  },
}
```

:thinking:

ST-DDT:
> ```ts
>  { 
>     location: { 
>       city: null,
>       country: null,
>     },
>   }
> ```

I'll write a test for that tomorrow, if I don't have one of those yet.

xDivisionByZerox:
> [...] if I don't have one of those yet.

Currently the word `null` does not exist in the `merge-locale.spec.ts` file.

ST-DDT:
I added the missing test.

xDivisionByZerox (approved):
Thank you :+1:

Shinigami92 (approved):

src/definitions/person.ts
```diff
@@ -7,21 +7,21 @@ export type PersonDefinitions = LocaleEntry<{
   gender: string[];
   sex: string[];
 
-  prefix?: string[];
-  female_prefix?: string[];
-  male_prefix?: string[];
+  prefix: string[];
```
matthewmayer, line 10: If these are allowed to be null, how come the ? is removed

Shinigami92: I was confused by this first as well

However `?` just stands for _this prop is optional_, which has a slightly other meaning than `| null` or even `| undefined`

But here we just defined the Definition for Person and then wrap it with `Partial` in the upper layer (see `LocalEntry`)

https://github.com/faker-js/faker/blob/d9c8ed29af3972d0084fb5a0d8b7780d7ad4c3a7/src/definitions/definitions.ts#L25-L27

https://github.com/faker-js/faker/blob/d9c8ed29af3972d0084fb5a0d8b7780d7ad4c3a7/src/definitions/person.ts#L6

ST-DDT: The ? originally caused issues during the build:types step. I'm not sure why, but since they are redundant anyway I decided to remove them all.

matthewmayer: ok so practically, all definitions are allowed to be null?

PR #1936, opened 2023-03-15
feat(locale): ro_MD

botzill:
Created new PR after rebase with upstream

Issue #1542, opened 2022-11-07
Add phone.number generation options


ST-DDT:
### Clear and concise description of the problem

Currently the phone.number method generates phone numbers in various formats.
However, the usually the phone number inputs require specific input formats e.g. an international telephone number.

### Suggested solution

It would be nice if it is possible to be more specific which addresses should be generated.

- `type?: 'landline' | 'cell-phone'`
- `format?: 'international' | 'national'`

It is not a goal to generate phone numbers that are normalized beyond the given options. So the user might still have to strip whitespace or other "human" additions.

### Alternative

- https://github.com/faker-js/faker/issues/1541

mcmxcdev:
We actually have a real-world issue related to this:

We use [react-phone-number-input](https://gitlab.com/catamphetamine/react-phone-number-input) with a restriction for `country=US`, but `faker.phone.number()` can't be configured easily to give us US locale only.

Therefore, we encounter this warning within our unit tests:

```
console.error
      [react-phone-number-input] Expected phone number +17427794379 to correspond to country US but in reality it corresponds to country CA.
```

matthewmayer:
> We actually have a real-world issue related to this:
> 
> We use [react-phone-number-input](https://gitlab.com/catamphetamine/react-phone-number-input) with a restriction for `country=US`, but `faker.phone.number()` can't be configured easily to give us US locale only.
> 
> Therefore, we encounter this warning within our unit tests:
> 
> ```
> console.error
>       [react-phone-number-input] Expected phone number +17427794379 to correspond to country US but in reality it corresponds to country CA.
> ```

if you do `faker.locale = 'en_US'`, i think you should only get US area codes, not including 742? https://github.com/faker-js/faker/blob/next/src/locales/en_US/phone_number/area_code.ts

mcmxcdev:
I guess I can just wait for https://github.com/faker-js/faker/pull/1735 to be merged, which makes `en_US` the default and kicks out all other locales?

matthewmayer:
There's a difference between the default locale `en` which is generic for worldwide English, and `en_US` which is specific for the United States, #1735 won't change that.

So `en` if you want a mix of phone numbers from different countries, `en_US` for US and `en_GB` for Great Britain, etc.

mcmxcdev:
Hmm, the problem is we use `faker.phone.number('+1##########')` currently which then doesn't care about the locale set.

[...]

PR #2084, opened 2023-04-23
feat(locale): add ro_MD locale

matthewmayer:
Rebased version of https://github.com/faker-js/faker/pull/1936 by @botzill


Shinigami92 (dismissed):

src/locales/ro_MD/location/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  city,
+  city_name,
+  city_prefix,
+  county,
+  default_country,
+  postcode,
+  secondary_address,
+  state,
+  state_abbr,
+  street,
```
ST-DDT: Should this be `street_pattern` now?

src/locales/ro_MD/location/street.ts
```diff
@@ -0,0 +1 @@
+export default ['{{location.street_prefix}} {{location.street_name_part}}'];
```
Shinigami92: yes, this file needs to be renamed + preflight

ST-DDT:
This PR needs to be updated by running the generate locales script again.

ST-DDT (approved):

src/locales/ro_MD/cell_phone/index.ts
```diff
@@ -0,0 +1,11 @@
...
+ * This file is automatically generated.
+ * Run 'pnpm run generate:locales' to update.
+ */
+import formats from './formats';
+
+const cell_phone = {
+  formats,
+};
+
+export default cell_phone;
```
xDivisionByZerox, line 11: `cell_phone` is not officially supported as locale data. I'm not sure if we should add those since we have the `PhoneDefinition` for stuff like that.

matthewmayer: indeed, but i figured it was better to keep it and then if #1542 is implemented it will be easier to reuse existing patterns for mobile phones

xDivisionByZerox: But should we not store those values in the phone module? Like `phone.cell_phone_formats`? Why have an extra (undocumented) module for those.

matthewmayer: presumably cell_phone was used at some point in the past, so for consistency we're reusing the same name as other locales. Yes, it will need to be moved under phone in the future, but thats out of scope for this PR.

Shinigami92 (approved):

xDivisionByZerox (approved):

Issue #1471, opened 2022-10-18
src/locales/*/address/city.ts is inplausible


ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/en_AU_ocker/address/city.ts#L1

It should most likely point to `city_name` or be deleted.

**Note:** After #1469 is merged these files will then be located under `src/locales/*/location/city.ts`.

ST-DDT:
On hold till #1469 is fixed.

ST-DDT:
Also

https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/en_NG/address/city.ts#L1

ST-DDT:
Also

https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/en_ZA/address/city.ts#L1

ST-DDT:
Also

https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/es/address/city.ts#L1

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/es_MX/address/city.ts#L1

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/fr_BE/address/city.ts#L1-L4

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/nl_BE/address/city.ts#L1-L4

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/sv/address/city.ts#L1

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/zh_CN/address/city.ts#L1

ST-DDT:
https://github.com/faker-js/faker/blob/79858fea203bce7ada9e9bcc7751f6ab25123977/src/locales/zh_TW/address/city.ts#L1

ST-DDT:
I might have missed some files, so the other `city.ts` files should be checked again.

[...]

PR #2094, opened 2023-04-25
refactor(location)!: rename definition city to city_pattern

### LINKED ISSUE #2021: Merge faker.location.city() and faker.location.cityName()

xDivisionByZerox:
Fixes #2021, #1471.


Shinigami92 (approved):

matthewmayer:
this doesn't actually fix #2021 though right? Because the two methods still return different data.

Shinigami92:
> this doesn't actually fix https://github.com/faker-js/faker/issues/2021 though right? Because the two methods still return different data.

- https://github.com/faker-js/faker/issues/2021#issuecomment-1499260452

> We will start without a generated/fictional parameter and will add that back when users request them. (This avoids breaking the API when removing the parameter if unneeded)

matthewmayer:
But i mean even after this PR is merged, say

`fakerEN.location.city()` and `fakerEN.location.cityName()`
`fakerDE.location.city()` and `fakerDE.location.cityName()`

do not behave the same, so there would need to be another PR before https://github.com/faker-js/faker/issues/2021 is considered fixed?

xDivisionByZerox:
> But i mean even after this PR is merged, say
> 
> `fakerEN.location.city()` and `fakerEN.location.cityName()` `fakerDE.location.city()` and `fakerDE.location.cityName()`
> 
> do not behave the same, so there would need to be another PR before #2021 is considered fixed?

No, because `location.city()` already returns a superset of `location.cityName()`. 
I have explained this to you this in https://github.com/faker-js/faker/pull/2070#issuecomment-1514920745

matthewmayer:
ok then maybe #2021 should be renamed? Because the methods are not being merged.

matthewmayer:
in what way does location.city()  return a superset of location.cityName()?

fakerDE.location.cityName() can return Düsseldorf but fakerDE.location.city() will never return Düsseldorf

xDivisionByZerox:
> in what way does location.city() return a superset of location.cityName()?
> 
> fakerDE.location.cityName() can return Düsseldorf but fakerDE.location.city() will never return Düsseldorf

In that case, that's an issue with the `fakerDE` instance. Not the method implementation.

matthewmayer:
Right so these locales will need fixing in a seperate PR if we want city() to also return their real city names

[
'de',    'dv',    'en',
    'en_CA', 'it',    'ka_GE',
    'ko',    'nb_NO', 'ne',
    'ro',    'tr',    'uk',
    'ur'
]

xDivisionByZerox:
The following locales are in fact able to return real city names:
- [dv](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/dv/location/city.ts#L7)
- [en](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/en/location/city.ts#L6)
- [ne](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/ne/location/city.ts#L6)
- [ro](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/ro/location/city.ts#L6)
- [tr](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/tr/location/city.ts#L6)
- [uk](https://github.com/faker-js/faker/blob/4d0458c96071917c8c3bb85fa61544caf8ff1763/src/locales/uk/location/city.ts#L2)

ST-DDT (approved):

PR #2095, opened 2023-04-25
docs: add npm script descriptions to CONTRIBUTING

### LINKED ISSUE #322: Document npm scripts

xDivisionByZerox:
Closes #322.

Original content from @matthewmayer in https://github.com/faker-js/faker/issues/322#issuecomment-1522076156.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,6 +5,19 @@ If you want to make `Faker` a better, please read the following contribution gui
 # Important
 
 Please make sure that you run `pnpm run preflight` before making a PR to ensure that everything is working from the start.
+This is a shorthand for running the following scripts in order:
+
+- `pnpm install` - installs npm packages defined in package.json
+- `pnpm run generate:locales` - generates locale files
+- `pnpm run generate:api-doc` - generates API documentation
+- `pnpm run format` - runs [prettify](https://github.com/prettier/prettier) to format code
+- `pnpm run lint` - runs [ESLint](https://github.com/eslint/eslint) to enforce project code standards
+- `pnpm run build:clean` - removes artifacts from previous builds
+- `pnpm run build:code` - builds the code, both CommonJS and ESM versions
+- `pnpm run build:types` - builds the TypeScript type definitions
+- `pnpm run test:update-snapshots ` - runs all tests, and updates any snapshots if needed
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -5,6 +5,19 @@ If you want to make `Faker` a better, please read the following contribution gui
 # Important
 
 Please make sure that you run `pnpm run preflight` before making a PR to ensure that everything is working from the start.
```
Shinigami92: This will add a breakline

```suggestion
Please make sure that you run `pnpm run preflight` before making a PR to ensure that everything is working from the start.  
```

Shinigami92 (commented):
I'm okay with that, but it might not get updated if we forget it when changing something in the preflight command itself

matthewmayer (dismissed):

CONTRIBUTING.md
ST-DDT: If we use ` /` (or backslash?) then it would be better visible that there is an intentional linebreak here.

Shinigami92: We don't use that nowhere else yet 🤔 
Where as using `/\s{2}$/` search on `*.md` results in 11 hits

4 hits of that already in this exact file

We also configured editorconfig to support that markdown standard

https://github.com/faker-js/faker/blob/3fc7bf1b24ed67696b4824abcd9fa14af43cb66d/.editorconfig#L11-L12

xDivisionByZerox: I always do a simple `<br>`. Didn't even know that double whitespace did the same.

Shinigami92 (approved):

CONTRIBUTING.md
Shinigami92: image

matthewmayer (approved):

CONTRIBUTING.md
xDivisionByZerox: But why would I ever search for that if `<br>` works fine?

Shinigami92: Because double spaces are the defacto standard for markdown but it also supports a subset of html

ejcheng (approved):

PR #2096, opened 2023-04-25
feat(date): introduce anytime

### LINKED ISSUE #2093: date.anytime
Due to a requested reviews to use `between` instead of its own function, it slipped into the next branch that we dont have a method anymore that just provides a random date but without arguments/parameters
We need to fix that before releasing v8.0

https://github.com/faker-js/faker/pull/2053#issuecomment-1520404149

@ST-DDT suggested to add a `date.anytime` for that

ST-DDT:
grafik

### DIFF
--- a/src/modules/date/index.ts
+++ b/src/modules/date/index.ts
@@ -52,6 +52,41 @@ export class DateModule {
   }
 
   /**
+   * Generates a random date that can be either in the past or in the future.
+   *
+   * @param options The optional options object.
+   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `faker.defaultRefDate()`.
+   *
+   * @see faker.date.between() For dates in a specific range.
+   * @see faker.date.past() For dates explicitly in the past.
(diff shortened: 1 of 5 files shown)


Shinigami92 (dismissed):

Shinigami92 (approved):

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):

src/modules/date/index.ts
```diff
@@ -51,6 +51,41 @@ export class DateModule {
     }
   }
 
+  /**
+   * Generates a random date that can be either in the past or in the future.
```
matthewmayer, line 55: Could mention that it's up to one year into past or future?

Shinigami92: We talked about this already yesterday in person and decided not to do this for now, so we can leave it open and update with optional parameters in the future if community requests it

PR #2097, opened 2023-04-26
chore: eslint report-unused-disable-directives

Shinigami92:
https://eslint.org/docs/latest/use/command-line-interface#--report-unused-disable-directives

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "docs:serve": "vitepress serve docs --port 5173",
     "docs:diff": "tsx ./scripts/diff.ts",
     "format": "prettier --cache --write .",
-    "lint": "eslint --cache --cache-strategy content .",
+    "lint": "eslint --cache --cache-strategy content --report-unused-disable-directives .",
     "ts-check": "run-s ts-check:scripts ts-check:tests",
     "ts-check:scripts": "tsc --project tsconfig.check-scripts.json",
     "ts-check:tests": "tsc --project tsconfig.check-tests.json",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "docs:serve": "vitepress serve docs --port 5173",
     "docs:diff": "tsx ./scripts/diff.ts",
     "format": "prettier --cache --write .",
-    "lint": "eslint --cache --cache-strategy content .",
+    "lint": "eslint --cache --cache-strategy content --report-unused-disable-directives .",
```
ST-DDT, line 79: Does this throw or only report?
Is it possible to move this to a config file?

ST-DDT (dismissed):

package.json
Shinigami92: It throws error like any other error thrown by eslint

I did not found a config for that, and I'm the author of eslint-define-config 😅 this arg is totally new to me and I never heard of it until today

ST-DDT: There seems to be an option to enable warnings:

- https://github.com/eslint/eslint/issues/15466

Shinigami92: So indeed there is a https://eslint.org/docs/latest/use/configure/rules#report-unused-eslint-disable-comments and even my eslint-define-config already has it 😆 

But! it only warns 🙁 
So maybe we can turn on just both?! config for in IDE and cli for PRs

ST-DDT (approved):
I havent tested this in the IDE but looks good to me.

xDivisionByZerox (approved):
I like this one.

ejcheng (approved):

Issue #1431, opened 2022-10-11
Set lint warnings to error


ST-DDT:
### Clear and concise description of the problem

We previously fixed all lint warnings in our code, then we accidentally introduced a lint "warning" to our code.

### Suggested solution

Set lint levels to error, so that the build pipeline fails, if these things are added.

### Alternative

Manually check for lint warnings in each PR.

xDivisionByZerox:
I like the idea. We have these rules for a reason, no reason to allow bypassing them.

PR #2098, opened 2023-04-26
test: lint vitest

Shinigami92:
Lint with https://www.npmjs.com/package/eslint-plugin-vitest

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,7 +18,6 @@ module.exports = defineConfig({
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/recommended-requiring-type-checking',
     'plugin:prettier/recommended',
-    'plugin:vitest/recommended',
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
@@ -75,8 +74,6 @@ module.exports = defineConfig({
       { allowNumber: true, allowBoolean: true },
     ],
     '@typescript-eslint/unbound-method': 'off',
-
-    'vitest/valid-expect': ['error', { maxArgs: 2 }],
   },
   overrides: [
     {
@@ -115,8 +112,10 @@ module.exports = defineConfig({
     },
     {
       files: ['test/*.spec.ts'],
+      extends: ['plugin:vitest/recommended'],
       rules: {
         'deprecation/deprecation': 'off',
+
         '@typescript-eslint/restrict-template-expressions': [
           'error',
           {
@@ -125,6 +124,8 @@ module.exports = defineConfig({
             allowAny: true,
           },
         ],
+
+        'vitest/valid-expect': ['error', { maxArgs: 2 }],
       },
     },
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -127,6 +127,7 @@ module.exports = defineConfig({
           },
         ],
 
+        'vitest/expect-expect': 'off',
         'vitest/valid-expect': ['error', { maxArgs: 2 }],
       },
     },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,10 +29,12 @@ module.exports = defineConfig({
   rules: {
     // We may want to use this in the future
     'no-useless-escape': 'off',
-    'deprecation/deprecation': 'error',
     eqeqeq: ['error', 'always', { null: 'ignore' }],
     'no-else-return': 'error',
     'prefer-template': 'error',
+
+    'deprecation/deprecation': 'error',
+
     '@typescript-eslint/array-type': [
       'error',
       { default: 'array-simple', readonly: 'generic' },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = defineConfig({
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/recommended-requiring-type-checking',
     'plugin:prettier/recommended',
+    'plugin:vitest/recommended',
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
@@ -74,6 +75,8 @@ module.exports = defineConfig({
       { allowNumber: true, allowBoolean: true },
     ],
     '@typescript-eslint/unbound-method': 'off',
+
+    'vitest/valid-expect': ['error', { maxArgs: 2 }],
   },
   overrides: [
     {
--- a/test/all_functional.spec.ts
+++ b/test/all_functional.spec.ts
@@ -138,7 +138,7 @@ describe('BROKEN_LOCALE_METHODS test', () => {
   });
 
   Object.keys(modules).forEach((module) => {
-    describe(module, () => {
+    describe(`${module}`, () => {
       it('should not contain obsolete configuration (methods)', () => {
         const existingMethods = modules[module];
         const configuredMethods = Object.keys(
@@ -157,14 +157,14 @@ describe('BROKEN_LOCALE_METHODS test', () => {
 
 describe('functional tests', () => {
   for (const [locale, faker] of Object.entries(allFakers)) {
-    describe(locale, () => {
+    describe(`${locale}`, () => {
       if (locale === 'base') {
         it.skip('base locale is checked by other tests');
         return;
       }
 
       Object.keys(modules).forEach((module) => {
-        describe(module, () => {
+        describe(`${module}`, () => {
           modules[module].forEach((meth) => {
             const testAssertion = () => {
               // TODO @ST-DDT 2022-03-28: Use random seed once there are no more failures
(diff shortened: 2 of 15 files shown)


Shinigami92:
I think this is a good start

But there are more rules to explore later in future PRs like

- [prefer-each](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer-each.md)
- [prefer-lowercase-title](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer-lowercase-title.md)
- [prefer-spy-on](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer-spy-on.md)
- [prefer-to-contain](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer-to-contain.md)

test/all_functional.spec.ts
```diff
@@ -138,7 +138,7 @@ describe('BROKEN_LOCALE_METHODS test', () => {
   });
 
   Object.keys(modules).forEach((module) => {
-    describe(module, () => {
+    describe(`${module}`, () => {
```
ST-DDT, line 141: Why?

test/helpers.spec.ts
```diff
@@ -359,7 +359,7 @@ describe('helpers', () => {
           const frozenArray = Object.freeze(testArray);
           expect(() =>
             faker.helpers.weightedArrayElement(frozenArray)
-          ).to.not.throw();
+          ).not.throw();
```
ST-DDT: Is there maybe a `not.toThrow`?

test/vehicle.spec.ts
```diff
@@ -82,7 +73,24 @@ describe('vehicle', () => {
         });
       });
 
+      describe('color()', () => {
```
ST-DDT, line 76: Why did you shuffle them around?

test/all_functional.spec.ts
Shinigami92: https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/valid-title.md

it is expected that `describe` and `it` always contains (compile-time) strings
in this specific example it is later anyways recommended to use `describe.each(Object.keys(modules))(`%s`, () => {})` via https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer-each.md

test/helpers.spec.ts
Shinigami92: I think yes, will check later

If I remember correctly, `toThrow()` and `throw()` behave slightly differently, but in most cases this is not relevant (but I will keep this in mind when I will check that later)

test/vehicle.spec.ts
Shinigami92: https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/no-identical-title.md

reported that `describe('vin()', ...` was used twice, so I correctly merged them
then while already touching `vin`, I moved it down above the `vrm` so they are near to each other
I recommend (for a later PR) to order all the tests again to the same order as in the modules
And additionally also order module methods in alphabetic or logical understandable order (maybe there is even a lint rule that could help us?)

ST-DDT: > I recommend (for a later PR) to order all the tests again to the same order as in the modules

I think that wont be neccesary since the tree shaking refactor will split them to individual files anyway.

.eslintrc.js
```diff
@@ -112,8 +112,10 @@ module.exports = defineConfig({
     },
     {
       files: ['test/*.spec.ts'],
+      extends: ['plugin:vitest/recommended'],
       rules: {
         'deprecation/deprecation': 'off',
+
```
ST-DDT, line 120: The other configs don't have the white space here.

```suggestion
```

.eslintrc.js
```diff
@@ -122,6 +124,8 @@ module.exports = defineConfig({
             allowAny: true,
           },
         ],
+
```
ST-DDT, line 129:
```suggestion
```

ST-DDT (changes requested):
Please set the lint level to error. Since there is no point in just warning about errors and then ignoring them.

.eslintrc.js
Shinigami92: not directly related to this PR, but I will add a blankline in the other places
the blanklines are separating the plugin groups from each other

Shinigami92:
> Please set the lint level to error. Since there is no point in just warning about errors and then ignoring them.

you might have read something different
there is no 'warn'?!
please point me where you have read that

ST-DDT:
You haven't done explicitly here, but implicitly (or there is a config error somewhere).

As per https://github.com/faker-js/faker/issues/1431 we decided to use error for lint rules to prevent accidentally adding warnings, that don't get fixed.

- https://github.com/faker-js/faker/issues/1431
- https://github.com/faker-js/faker/pull/1714

To prevent cases like these that "pollute" the build logs.

grafik

(If it is important, then it should be fixed. If it isn't important, why clutter the logs with warnings.)

I don't care whether we explicitly configure them to error or consider warnings to be errors in CI.
For `preflight` it might be best, if we fail in the same cases as the CI/when there are warnings.

test/all_functional.spec.ts
ST-DDT: Then let's fix this as intended instead of simply bypassing the error/warning.

Shinigami92:
> You haven't done explicitly here, but implicitly (or there is a config error somewhere).
> 
> As per #1431 we decided to use error for lint rules to prevent accidentally adding warnings, that don't get fixed.
> 
> * [Set lint warnings to error #1431](https://github.com/faker-js/faker/issues/1431)
> * [infra: set lint warnings to error #1714](https://github.com/faker-js/faker/pull/1714)
> 
> To prevent cases like these that "pollute" the build logs.
> 
> grafik
> 
> (If it is important, then it should be fixed. If it isn't important, why clutter the logs with warnings.)
> 
> I don't care whether we explicitly configure them to error or consider warnings to be errors in CI. For `preflight` it might be best, if we fail in the same cases as the CI/when there are warnings.

The issue is now open and requested at upstream

test/all_functional.spec.ts
Shinigami92: As discussed, I will activate `prefer-each` in a separate PR
Reason: minimize the changes already made in this PR (already 15 files changed)

The current change is necessary anyway as already commented above `valid-title`

ST-DDT (dismissed):

test/finance.spec.ts
```diff
@@ -577,15 +571,15 @@ describe('finance', () => {
...
+          ).toBe(1);
         });
 
         it.each(['AA', 'EU'])(
           'throws an error for unsupported country code "%s"',
           (unsupportedCountryCode) =>
             expect(() =>
               faker.finance.iban(false, unsupportedCountryCode)
-            ).toThrowError(
+            ).toThrow(
```
xDivisionByZerox, line 582: According to the [maintainer](https://github.com/vitest-dev/vitest/discussions/2570?sort=top#discussioncomment-4503284) `toThrow` and `toThrowError` are the same.
Why prefer one over the other? If for consistency purposes, why prefer the less verbose one?

test/helpers.spec.ts
```diff
@@ -745,7 +741,7 @@ describe('helpers', () => {
           ]);
           expect(() =>
             faker.helpers.shuffle(input, { inplace: false })
-          ).not.to.throw();
+          ).not.throw();
```
xDivisionByZerox: Why not `.not.toThrow` for consistency purposes?

test/datatype.spec.ts
```diff
@@ -226,7 +226,7 @@ describe('datatype', () => {
 
           expect(() => {
             faker.datatype.number({ min, max });
-          }).toThrowError(`Max ${max} should be greater than min ${min}.`);
+          }).toThrow(`Max ${max} should be greater than min ${min}.`);
```
xDivisionByZerox: This error expectation is different from all others :/

test/finance.spec.ts
Shinigami92: Yes, they are aliases (see https://vitest.dev/api/expect.html#tothrowerror)

@ST-DDT had the argument that we mostly use `toThrow(FakerError` and therefore it already includes `Error`
In the other cases it is not relevant for us if it is an Error or anything else, just something was thrown

test/helpers.spec.ts
Shinigami92: thx, I might just have missed that one 👍

test/finance.spec.ts
xDivisionByZerox: This explanation makes sense to me. Approved 👍

xDivisionByZerox (dismissed):

Shinigami92:
will rebase after the dep updates

ST-DDT (dismissed):

ST-DDT (approved):

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2099, opened 2023-04-26
fix(types): locale proxy

ST-DDT:
The main sources don't show the error, but the test code does:

grafik
(e.g. date.spec.ts)


ST-DDT:
I'm not sure whether this is the correct solution.

xDivisionByZerox:
> The main sources don't show the error, but the test code does [...]

I'm pretty sure that is because the test files are not concluded in the `tsconfig.json`. I don't actually know which config files they use, but they seem to have different TS config settings. If you include `'test/**/*'` in the `tsconfig.json` file the error is gone.

If you then activate the `strictNullChecks` rule, the error is there again and additionally in the source files. So we essentially missing strict configuration settings.

Shinigami92 (approved):

Shinigami92:
I checked if we could use `any` instead of `unknown`, but then we are affected by not being refactor safe :confused: 
So if someone really uses the for custom definitions, they are forced to cast :shrug:

xDivisionByZerox (approved):
We may want to start adding TS compile tests to the project. Just an idea.

Shinigami92:
> We may want to start adding TS compile tests to the project. Just an idea.

No, we just need to start using `strict`
see my already started efforts at
- #2102 

we dont need tests, we need more safe checked TS itself

Issue #1836, opened 2023-02-13
Turn on strictPropertyInitialization in tsconfig


ST-DDT:
https://github.com/faker-js/faker/pull/651 partially activated strict mode, this flag was omitted at that time and should probably be turned on.

https://github.com/faker-js/faker/blob/c46c53948efaf9ccb2597f36954128f914b73179/tsconfig.json#L14

Related issues:

- https://github.com/faker-js/faker/pull/651
- https://github.com/faker-js/faker/issues/1834
- https://github.com/faker-js/faker/issues/1835
- https://github.com/faker-js/faker/issues/1837
- https://github.com/faker-js/faker/issues/1838

Shinigami92:
This is blocked by 
- #1834

As `strictNullChecks` is a precondition

ST-DDT:
Blocked by #2102.

- #2102

xDivisionByZerox:
I will close this as this has been activated implicitly in #2102.

PR #2102, opened 2023-04-27
infra: reduce tsconfig complexity

Shinigami92:
This PR turns around the tsconfig files. Previously we had one tsconfig.base file, from which the others extends (but not the normal tsconfig.json).

~~Now we have one tsconfig.json which serves as the root/base config for all others!~~

~~This tsconfig.json now only configures the minimum amount of base for all the other configs.~~

~~Then we have three specific configs:~~

~~build-types defines what is required to generate the files from src to dist/types
we need to have this because it declares the "include: [src]" which is needed to tell tsc it only takes files from this folder.
I looked already if there is a CLI flag for --include, but sadly no.~~

~~then we have two (right now mostly identically) configs for scripts and tests.
they differ in their "include"
We could discuss if we want to merge them into one file and just use one CI step to check tests and scripts folder, or we could just move the configs into there specific folders
The later option would result in VSCode would read them and process the configs for the folders~~

eslint now takes the tsconfig.json instead of its own lint.config
this results in that we now see even more lint errors that were previously hidden (see https://github.com/faker-js/faker/actions/runs/4822808872/jobs/8590495983?pr=2102)

---

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "target": "ES2019", // the min target for Node v14.17, could potentially also moved into the tsconfig.build-types.json but it could affect eslint
    "moduleResolution": "node", // in TS v5+ we might want to set this to "bundler"
    "module": "ESNext", // the newest stuff
    "noEmit": true, // we dont want to emit any files by default, this will be overwritten in tsconfig.build-types.json
    "strict": true // we want strict by default
  },
  "exclude": ["node_modules", "dist", "locale"] // exclude node_modules and generated content by default
}
```

```jsonc
// tsconfig.build-types.json
{
  "extends": "./tsconfig.json", // extend from base
  "compilerOptions": {
    "noEmit": false, // we want to emit .d.ts files
    "declaration": true, // we want to emit .d.ts files
    "emitDeclarationOnly": true, // we want to emit .d.ts files only
    "stripInternal": true, // we want to strip @internal
    "rootDir": "src",
    "outDir": "dist/types",
    // We need to disable these for now, and need to tackle them in another PR
    "strictNullChecks": false,
    "strictBindCallApply": false
  },
  "include": ["src"] // this is important and the reason why we need this file!
}
```

```jsonc
// tsconfig.check-scripts.json and tsconfig.check-tests.json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "target": "ESNext", // we can use newer code in scripts and tests because it will not be shipped
    "allowSyntheticDefaultImports": true, // we import e.g. prettier.cjs
    "resolveJsonModule": true, // we e.g. import json files
    "skipLibCheck": true, // we access some vitepress and so on which throws invalid types but we dont care
    // We need to disable these for now, and need to tackle them in another PR
    "strictNullChecks": false,
    "strictBindCallApply": false,
    "noImplicitAny": false
  },
  "include": ["scripts"] || "include": ["tests"]
}	}
```

---

Edit:

we now have one tsconfig.json which serves for all files and then one specific for the build process


Shinigami92:
Waiting on #2099 to be merged
- #2099

xDivisionByZerox:
> I looked already if there is a CLI flag for --include, but sadly no.

It's literally the default argument. 
`tsc src/**` is the same as only  `tsc` with `include: ["src/**"]` in the `tsconfig.json`.

Shinigami92:
> > I looked already if there is a CLI flag for --include, but sadly no.
> 
> It's literally the default argument. `tsc src/**` is the same as only `tsc` with `include: ["src/**"]` in the `tsconfig.json`.

This explains why I did not found it :facepalm:

.github/workflows/ci.yml
```diff
@@ -196,7 +196,7 @@ jobs:
           CYPRESS_INSTALL_BINARY: 0
 
       - name: Check scripts
-        run: pnpm run ts-check:scripts
+        run: pnpm run ts-check
```
Shinigami92, line 195: reminder: merge these jobs

CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@ If you want to make `Faker` a better, please read the following contribution gui
 Please make sure that you run `pnpm run preflight` before making a PR to ensure that everything is working from the start.  
 This is a shorthand for running the following scripts in order:
 
+<!-- TODO @Shinigami92 2023-04-28: update below -->
```
Shinigami92: reminder: update this section

src/modules/helpers/index.ts
```diff
@@ -277,7 +277,7 @@ export class HelpersModule {
     let max: number;
     let tmp: number;
     let repetitions: number;
-    let token = string.match(RANGE_REP_REG);
+    let token = RANGE_REP_REG.exec(string);
```
Shinigami92: extract this into own PR

test/helpers.spec.ts
```diff
@@ -791,7 +791,7 @@ describe('helpers', () => {
         });
 
         it('empty array returns empty array', () => {
-          const input = [];
+          const input: never[] = [];
```
Shinigami92: extract this into own PR

test/scripts/apidoc/signature.example.ts
```diff
@@ -94,7 +94,7 @@ export class SignatureTest {
    * @param b The string parameter.
    */
   optionalStringParamMethod(b?: string): number {
-    return +b;
+    return b ? 0 : 1;
```
Shinigami92: extract this into own PR

test/support/seededRuns.ts
```diff
@@ -34,7 +34,7 @@ type NoArgsMethodOf<ObjectType> = MethodOf<ObjectType> &
   {
     [Key in MethodOf<ObjectType, () => unknown>]: ObjectType[Key] extends (
       arg0: string | number | boolean | Record<string, undefined>,
-      ...args
+      ...args: any[]
```
Shinigami92: extract this into own PR

vite.config.ts
```diff
@@ -11,6 +11,7 @@ export default defineConfig({
     setupFiles: ['test/setup.ts'],
     coverage: {
       all: true,
+      provider: 'c8',
```
Shinigami92: extract this into own PR

CONTRIBUTING.md
xDivisionByZerox: What do you want to update exactly? If you provide some requirements I could help with that.

Shinigami92: e.g. the ts-check:* were now merged to one script

xDivisionByZerox: Ok I see. So this can be done in this PR, In that case NVM my request^^

xDivisionByZerox:
For the title: How about `infra: reduce tsconfig complexity`?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
-    "useUnknownInCatchVariables": true,
     "stripInternal": true,
-    "baseUrl": "."
+
+    // We need to disable these for now, and need to tackle them in another PR
+    "strictNullChecks": false,
+    "strictBindCallApply": false,
+    "noImplicitAny": false,
+
+    // These are configs specifically for !build and have to be reverted in the tsconfig.build.json
```
xDivisionByZerox, line 16: If we have cases that negate each other I'm not a fan of the current approach.  

I have multiple questions:
- Do we even need to consider this file as "shared"? 
- Does the `tsconfig.build.json` really need to extend the default config?
- Does the `tsconfig.build.json` need to exist at all? Or can we provide the options in the command directly?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
 {
   "compilerOptions": {
-    "target": "ES2019",
-    "moduleResolution": "Node",
-    "rootDir": "src",
-    "outDir": "dist",
```
xDivisionByZerox, line 6: This (`outDir`) no longer exists in the new config. Why did you remove this?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
+    "target": "ESNext",
+    "moduleResolution": "node",
+    "module": "ESNext",
+    "strict": true,
+    "noEmit": true,
     "declaration": true,
-    "esModuleInterop": true,
```
xDivisionByZerox, line 8: This (`esModuleInterop`) no longer exists in the new config. Why did you remove this?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
-    "allowJs": true,
```
xDivisionByZerox, line 9: This (`allowJs`) no longer exists in the new config. Why did you remove this?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
-    "alwaysStrict": true,
-    "strictFunctionTypes": true,
```
xDivisionByZerox, line 11: This (`strictFunctionTypes`) no longer exists in the new config. Why did you remove this?
It was explicitly enabled in #651 and initialized the mindset for strict TS configuration.

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "useUnknownInCatchVariables": true,
     "stripInternal": true,
-    "baseUrl": "."
+
+    // We need to disable these for now, and need to tackle them in another PR
+    "strictNullChecks": false,
+    "strictBindCallApply": false,
+    "noImplicitAny": false,
```
xDivisionByZerox, line 14: This (`noImplicitAny`) was previously activated. Do we have code outside of the `src` directory which violates this rule?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
...
+    "strict": true,
+    "noEmit": true,
     "declaration": true,
-    "esModuleInterop": true,
-    "allowJs": true,
-    "alwaysStrict": true,
-    "strictFunctionTypes": true,
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "useUnknownInCatchVariables": true,
```
xDivisionByZerox, line 14: This (`useUnknownInCatchVariables`) no longer exists in the new config. Why did you remove this?

tsconfig.json
```diff
@@ -1,20 +1,22 @@
 {
   "compilerOptions": {
```
xDivisionByZerox, line 2: Since you already changed a lot of keys in this config. Could we sort the config keys alphabetically?

tsconfig.json
Shinigami92: It is enabled via `strict` indirectly already

tsconfig.json
Shinigami92: `allowJs` is only necessary if you want to import a js file. This is not the case for any our files, so we can remove it and try to rely fully on ts files

tsconfig.json
Shinigami92: I removed it for now and hoped that we don't need to enable it anymore as we are not relying on really old CJS code
So it is good if we don't need it, but yes we need to check if this does not break the bundled dist output or we might need to compare the dist before and after this PR

But on the other side... maybe this does not affect us at all, because the bundle process is made by esbuild and not tsc for cjs and esm dist

tsconfig.json
Shinigami92: it is configured in tsconfig.build.json line 8
it is not required for the non-build config

tsconfig.json
Shinigami92: that is something we might need to discuss in a meeting as it would take to much space of writing comments

tldr: I would like to have it like this, as it would require very many cli flags and the current way is a really good way

tsconfig.json
Shinigami92: ~~same as `strictFunctionTypes`, it is already set via `strict`~~

nevermind, did not read the exact question 😓 
yes, I think we have that
we should tackle that in separate PR (as comment above is mentioning)

tsconfig.json
Shinigami92: it's part of `strict`

tsconfig.json
Shinigami92: no, I don't like to sort them alphabetically, but logically / groups

e.g. target, module and moduleResolution are grouped together because they indirectly affect each other

Shinigami92: also these configs changes very rarely, so I would suggest to stay with the structure given like a package.json which is also not sorted

tsconfig.json
xDivisionByZerox: Might just be me, but I like to be explicit. Since TypeScript does not follow semantic versioning, every minor update might break the behavior of `strict`.

tsconfig.json
xDivisionByZerox: Yes, but I really like to explicitly set this property to `true` or `false` to provide information on whether JS is allowed in the project.
In our case, I would suggest setting it to `false`.

tsconfig.json
Shinigami92: Everything that got inside `strict` never left it, there are only additional flags that slowly go inside `strict` in a later version

tsconfig.json
xDivisionByZerox: +1 for discussion

tsconfig.json
xDivisionByZerox: I see. That's why `noEmit` is set in this config. Accepted.

tsconfig.json
ST-DDT: Havent you done so already or is there something still missing?

Shinigami92: all of these still need to be checked by me
but in a later PR

.github/workflows/ci.yml
```diff
@@ -214,8 +214,8 @@ jobs:
         env:
           CYPRESS_INSTALL_BINARY: 0
```
Shinigami92: line 193 - 215 needs to be removed

cant grab the lines in GitHub Web UI...

package.json
```diff
@@ -53,16 +53,14 @@
   },
   "files": [
     "CHANGELOG.md",
-    "CHANGELOG_old.md",
     "dist",
-    "locale",
-    "tsconfig.json"
```
ST-DDT, line 59: I dont understand this change.
Is the file useless? If yes, why was it added in the first place?

Shinigami92: > Is the file useless?

I think yes, maybe we should test that :shrug: 
But a file like this is also not included in plain JS packages, is it?

> If yes, why was it added in the first place?

I don't know. Maybe just for debug purposes, same as e.g. changelog and readme is also included

Shinigami92: I can revert this change if you want to

xDivisionByZerox: > Is the file useless?

From the [npm docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files):

> The optional files field is an array of file patterns that describes the entries to be included when your package is installed as a dependency. [...] will make it so that file is included in the tarball when it's packed. Omitting the field will make it default to ["*"], which means it will include all files.

So yes, this file is kinda useless. You can even see that it is currently shipped. 
image
But since the package files are JS the config is not doing anything

---

Also note:

> Certain files are always included, regardless of settings:
> - CHANGES / CHANGELOG / HISTORY
> - [...]
> 
> [...], `CHANGES` can have any case and extension.

tsconfig.json
xDivisionByZerox: I'd like to discuss this. I don't see whats so bad about explicitly listing all configs.

package.json
Shinigami92: I was not aware that we are potentially lacking team experience about fields of the package json 👀 😲 
As it was already found in the comment mentioned here, every field is clearly documented on [docs.npmjs.com](https://docs.npmjs.com/cli/v6/configuring-npm/package-json)

But tbh I highly assume the question was more about the `tsconfig.json` being part of the `files` field itself, not what the `files` field is in general

If that was not the case, I can try to train us a bit about several config options or I can recommend some YouTube/Udemy tutorials

xDivisionByZerox: I provided an explanation about the `field` property from the docs because I thought it answeres the question perfectly. 

The additional note can be read as "Technically we don't need to add the `CHANGELOG.md` file here."
But we should discuss if we favor explicit or implicit configs as I already requested in https://github.com/faker-js/faker/pull/2102#discussion_r1183626141.

Shinigami92:
This PR should be merged before starting:

tsconfig.json
Shinigami92: In meeting 2023-07-13 we discussed that we dont want to explicit set config values that are not necessary

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2104, opened 2023-04-28
refactor(locale): add city_name to city_patterns

### LINKED ISSUE #2021: Merge faker.location.city() and faker.location.cityName()

xDivisionByZerox:
Part of  #2021.

[Comment REF](https://github.com/faker-js/faker/issues/2021#issuecomment-1523870053)


Shinigami92 (dismissed):

src/locales/nb_NO/location/city_pattern.ts
```diff
@@ -1 +1,4 @@
-export default ['{{location.city_name}}{{location.city_suffix}}'];
+export default [
+  '{{location.city_name}}{{location.city_suffix}}',
```
matthewmayer: If city_name is always a real city name it seems it should always stand alone and not be combined with a prefix or suffiX?

src/locales/ko/location/city_pattern.ts
```diff
@@ -1 +1,4 @@
-export default ['{{location.city_name}}{{location.city_suffix}}'];
+export default [
+  '{{location.city_name}}{{location.city_suffix}}',
```
matthewmayer: Ditto

matthewmayer (changes requested):

src/locales/nb_NO/location/city_pattern.ts
xDivisionByZerox: As you may see from the diff, I added the standalone `location.city_name` templates. If those did exist before, there was probably a reason to include them as they are. 

I don't see why we shouldn't include those patterns as they provide more city name variants.
We need to put some work into researching if those patterns are actually harmful to the specific locale.

ST-DDT: IMO we should fix those, but probably in a different PR to keep the PR concise.

If you have a city named Neustadt and just append a city prefix again, then you get bad names.
For street-names we also changed the pattern files to `street_name_part` to separate them from actual names.
We have to check the contents of these files to check if they contain actual city names or parts thereof.

matthewmayer: If you look at the nb_NO city_names they are not actually real city names but fragments of the first half of names, designed to be combined with city_suffix. 

So best would be to rename city_name to city_prefix and use pattern {{city_prefix}}{{city_suffix}} instead?

matthewmayer: This is how sv does it, for example.

matthewmayer: i made a quick script to output what all the different city_pattern look like with examples for each locale:

https://gist.github.com/matthewmayer/f14b59d799be8ed063c3f38cdca38a6a

Shinigami92:
What is the state of this PR? should this still be merged for v8.0?
Is it blocking?

matthewmayer:
> What is the state of this PR? should this still be merged for v8.0? Is it blocking?

i think the `no_NB` and `ko` changes should be reverted and then the rest can be merged. Then those two languages can be better handled in another PR.

xDivisionByZerox:
I reverted the changes for `ko` and `nb_NO`

Shinigami92 (approved):

matthewmayer (approved):

ST-DDT (approved):

PR #2107, opened 2023-04-28
test: prepare test sources for strict mode

Shinigami92:
extracted https://github.com/faker-js/faker/pull/2102

### DIFF
--- a/test/helpers.spec.ts
+++ b/test/helpers.spec.ts
@@ -791,9 +791,8 @@ describe('helpers', () => {
         });
 
         it('empty array returns empty array', () => {
-          const input = [];
           const length = faker.number.int({ min: 1, max: 6 });
-          const unique = faker.helpers.uniqueArray(input, length);
+          const unique = faker.helpers.uniqueArray([], length);
           expect(unique).toHaveLength(0);
         });
 
--- a/test/support/seededRuns.ts
+++ b/test/support/seededRuns.ts
@@ -34,7 +34,7 @@ type NoArgsMethodOf<ObjectType> = MethodOf<ObjectType> &
   {
     [Key in MethodOf<ObjectType, () => unknown>]: ObjectType[Key] extends (
       arg0: string | number | boolean | Record<string, undefined>,
-      ...args
+      ...args: unknown[]
     ) => unknown
       ? Key
       : never;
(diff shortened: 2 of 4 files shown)


test/support/seededRuns.ts
```diff
@@ -34,7 +34,7 @@ type NoArgsMethodOf<ObjectType> = MethodOf<ObjectType> &
   {
     [Key in MethodOf<ObjectType, () => unknown>]: ObjectType[Key] extends (
       arg0: string | number | boolean | Record<string, undefined>,
-      ...args
+      ...args: any[]
```
ST-DDT: I think you have to suppress the warning here.

ST-DDT (dismissed):

ST-DDT:
Please amend the title or description to include at least the intention of this PR before merging this.
So we are inclusive to our community that wasn't present in our private conversation where we talked about this change.

E.g. `This fixes some issues preventing the use of strict for our test sources`
or `prepare test sources for strict mode`

test/support/seededRuns.ts
```diff
@@ -34,7 +34,8 @@ type NoArgsMethodOf<ObjectType> = MethodOf<ObjectType> &
   {
     [Key in MethodOf<ObjectType, () => unknown>]: ObjectType[Key] extends (
       arg0: string | number | boolean | Record<string, undefined>,
-      ...args
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ...args: any[]
```
xDivisionByZerox: I'm very sure this can be `unknown`, so we no longer would have a warning at all.
This is because we don't actually care about any argument except the first one.

[TS Playground](https://tsplay.dev/mZZqom)

test/helpers.spec.ts
```diff
@@ -791,7 +791,7 @@ describe('helpers', () => {
         });
 
         it('empty array returns empty array', () => {
-          const input = [];
+          const input: never[] = [];
           const length = faker.number.int({ min: 1, max: 6 });
           const unique = faker.helpers.uniqueArray(input, length);
```
xDivisionByZerox: Explicit typing could be avoided by directly providing the value as an argument. 

```suggestion
          const length = faker.number.int({ min: 1, max: 6 });
          const unique = faker.helpers.uniqueArray([], length);
```

test/support/seededRuns.ts
Shinigami92: ah yes, and we are safe here anyways because this is just in the `test` folder, so no behavior changes to customer at all

xDivisionByZerox (approved):

ST-DDT (approved):

ejcheng (approved):

PR #2108, opened 2023-04-28
chore: improve regex usage

Shinigami92:
extracted https://github.com/faker-js/faker/pull/2102

Changes the usage of regexp to always follow `RegExp#method(string)`.
As a side effect this fixes some potential null value type issues currently hidden by our tsconfig.


ST-DDT (approved):
:shrug:

Shinigami92:
> 🤷

May I ask what the value of this reaction was intended?

If you want to understand why this PR, please just gentle ask for it

string.match(RegExp) bleeds to an token.index that can potentially undefined
where as RegExp.exec(string) does not
this PR is a preparation and extraction of #2102 and should be merged before we can work further with that other PR

beside that RegExp.test(string) just returns a boolean instead of Result|null, so the thrown away overhead is reduced

ST-DDT:
> May I ask what the value of this reaction was intended?

The intention behind this comment was originally to point out that some of your PR descriptions - especially for those PRs that dont fix an obvious bug - tend to be on a the shorter side and omit the part containing the intend and reasoning.
This makes them harder to understand/more ambigouus than necessary same as the lonely smiley.

I had also hoped that during the talk about this change I had successfully conveijed how unsure I am regarding the two methods and their differences.

xDivisionByZerox (approved):

matthewmayer:
i think this could be split into two. Seems like you are doing two things...

1. Replace `str.match(regex)` with `regex.test(string)` if you only need a boolean "it matches" or "it doesn't match" like inside an if-condition.

2. Replace `str.match(regex)` with `regex.exec(str)` if you do need the return values - but i'm not sure of the reason for this change, as the output seems to be the same if there is no /g flag on the regex - an object if it matches, or null if it doesn't

ST-DDT:
> i think this could be split into two.

I don't think this is necessary, as they are close enough.

> 2\. but i'm not sure of the reason for this change, as the output seems to be the same if there is no /g flag on the regex

There is a type difference there that avoid an issue in combination with the PR mentioned above.
A better description would make this easier to detect.

PR #2116, opened 2023-05-01
fix(test): fix failing latitude test

### LINKED ISSUE #2080: Test failure for latitude
FAIL  test/location.spec.ts > location > random seeded tests for seed 4743915782701865 > latitude() > returns latitude with min and max and default precision

matthewmayer:
fix #2080


ST-DDT (approved):

xDivisionByZerox:
<s>This doesn't really fix the issue, since those test failures come from `number.float` returning a floating point number where all decimal places are exactly `0`. 
For example `faker.number.float({ precision: 0.001 })` CAN return a number like `x.000` which JS trims to `x`. This is not a problem with the `float` implementation, but with how JS handles floating point number.

The fix you provided will have nicer error logs (at least I believe so) but doesn't actually fix the problem.</s>

Ok, I read your comments in #2080 and saw that you were already aware of the actual problem. That made me curious and have an additional look at the chang you provided. I didn't see that the `precision` function was actually used to change the expected value. MB.

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #2118, opened 2023-05-02
docs(location): mention real cities

### LINKED ISSUE #2021: Merge faker.location.city() and faker.location.cityName()

xDivisionByZerox:
As described in https://github.com/faker-js/faker/issues/2021#issuecomment-1523870053.


ST-DDT:
IMO we should phrase the description similar to street or vice versa.

src/modules/location/index.ts
```diff
@@ -140,7 +140,7 @@ export class LocationModule {
   }
 
   /**
-   * Generates a random fictional city name for the locale.
+   * Generates a random fictional or real city name for the locale.
```
xDivisionByZerox: @ST-DDT 
Do you like this?

```suggestion
   * Generates a random localized city name.
```

ST-DDT: I think for now it is okay.
We can think about better descriptions at a later point in time/different PR.

matthewmayer (approved):

Shinigami92 (approved):

ST-DDT (approved):

PR #2119, opened 2023-05-02
fix(location): avoid hyphenated surnames in city patterns

matthewmayer:
Changes city_patterns to use person.last_name instead of person.lastName. This means it uses the definitions directly, not the method, which avoids using hyphenated and double surnames in locales where those are used which were introduced in #1872 

e.g. remove city names like 
Newton Morissette-Kovacekwell
Pollich-Douglasville

Note this doesn't affect default `en` locale as that is already using last_name instead of lastName https://github.com/faker-js/faker/blob/e490ed3e505124585857421cac73d6f8418280aa/src/locales/en/location/city_pattern.ts


ejcheng (dismissed):

src/locales/af_ZA/location/city_pattern.ts
```diff
@@ -2,5 +2,5 @@ export default [
   '{{location.city_prefix}} {{person.firstName}}{{location.city_suffix}}',
   '{{location.city_prefix}} {{person.firstName}}',
   '{{person.firstName}}{{location.city_suffix}}',
-  '{{person.lastName}}{{location.city_suffix}}',
+  '{{person.last_name}}{{location.city_suffix}}',
```
xDivisionByZerox: This locale data includes strings which include spaces. For example: `van de Merwe`. Is this really expected to work as a pattern?

matthewmayer: I think both af_ZA and zu_ZA locale data for cities are suspect and would be better replaced with real cities like en_ZA but I think that's a seperate PR

xDivisionByZerox (dismissed):

ST-DDT (dismissed):

Shinigami92 (dismissed):

ejcheng (dismissed):

xDivisionByZerox (approved):

ejcheng (approved):

ST-DDT (approved):

PR #2120, opened 2023-05-02
fix(locale): remove unrealistic fictional city patterns in ne, ro, tr

matthewmayer:
Removes fictional patterns in three locales where location.city_name is available but there are no native city prefixes/suffixes available.

Avoids weird combinations of ne/ro/tr names with English prefixes/suffixes like

West Vladcester
Bökeshire
Manishacester

### DIFF
--- a/src/locales/ne/location/city_pattern.ts
+++ b/src/locales/ne/location/city_pattern.ts
@@ -1,7 +1,1 @@
-export default [
-  '{{location.city_prefix}} {{person.firstName}}{{location.city_suffix}}',
-  '{{location.city_prefix}} {{person.firstName}}',
-  '{{person.firstName}}{{location.city_suffix}}',
-  '{{person.lastName}}{{location.city_suffix}}',
-  '{{location.city_name}}',
-];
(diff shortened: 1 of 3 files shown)


ST-DDT (approved):
Looks okay to me.

Shinigami92 (approved):

xDivisionByZerox:
> Removes fictional patterns in three locales where location.city_name is available but there are no native city prefixes/suffixes available.

Do those prefixes/suffixes simply not exist in our dataset or is the concept itself non existing in the locale? Do you have any info on that? 

If that should be the case, please export `null` from the according data files. Otherwise, I think we should fill in those data instead of removing variants.

matthewmayer:
I'd imagine most languages would have certain prefixes or suffixes which might appear commonly in city names 

If someone who knows more about the language comes by later they could add back some new fictional patterns with appropriate suffixes and prefixes. 

But for now, just using the real city names seems superior.

xDivisionByZerox (approved):

PR #2123, opened 2023-05-03
feat(helpers): new method `objectEntry`

### LINKED ISSUE #1764: Add faker.helpers.objectEntry
So we could use ObjectKeys to get a random key and then get the values for it.
Or alternatively we introduce `objectEntry` that returns `[key, value]` directly.

What do you think?

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/pull/1761#discussion_r1083471463_

suyashgulati:
For issue #1764, added a new helper method named `faker.helpers.objectEntry` that will return a random key-value pair from the given object.

Changes made:
- Added a new method `faker.helpers.objectEntry`
- Added tests and updated snapshots

### DIFF
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -798,14 +798,14 @@ export class HelpersModule {
   }
 
   /**
-   * Returns a random [key, value] pair from given object or `undefined` if no key could be found.
+   * Returns a random [key, value] pair from given object or throws an Error if no key could be found.
    *
    * @template T The type of the object to select from.
    *
    * @param object The object to be used.
    *
    * @example
-   * faker.helpers.objectEntry({ myProperty: 'myValue' }) // ['myProperty', 'value']
+   * faker.helpers.objectEntry({ prop1: 'value1', prop2: 'value2' }) // ['prop1', 'value1']
    *
    * @since 8.0.0
    */
@@ -813,7 +813,7 @@ export class HelpersModule {
     object: T
   ): [keyof T, T[keyof T]] {
     const key = this.faker.helpers.objectKey(object);
-    return key ? [key, object[key]] : undefined;
+    return [key, object[key]];
   }
 
   /**
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -764,11 +764,13 @@ export class HelpersModule {
   }
 
   /**
-   * Returns a random key from given object or `undefined` if no key could be found.
+   * Returns a random key from given object.
    *
    * @template T The type of the object to select from.
    *
    * @param object The object to be used.
+   *
+   * @throws If the given object is empty.
    *
    * @example
    * faker.helpers.objectKey({ myProperty: 'myValue' }) // 'myProperty'
@@ -781,11 +783,13 @@ export class HelpersModule {
   }
 
   /**
-   * Returns a random value from given object or `undefined` if no key could be found.
+   * Returns a random value from given object.
    *
    * @template T The type of object to select from.
    *
    * @param object The object to be used.
+   *
+   * @throws If the given object is empty.
    *
    * @example
    * faker.helpers.objectValue({ myProperty: 'myValue' }) // 'myValue'
@@ -798,11 +802,13 @@ export class HelpersModule {
   }
 
   /**
-   * Returns a random [key, value] pair from given object or throws an Error if no key could be found.
+   * Returns a random `[key, value]` pair from the given object.
    *
    * @template T The type of the object to select from.
    *
    * @param object The object to be used.
+   *
+   * @throws If the given object is empty.
    *
    * @example
    * faker.helpers.objectEntry({ prop1: 'value1', prop2: 'value2' }) // ['prop1', 'value1']
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -798,6 +798,25 @@ export class HelpersModule {
   }
 
   /**
+   * Returns a random [key, value] pair from given object or `undefined` if no key could be found.
+   *
+   * @template T The type of the object to select from.
+   *
+   * @param object The object to be used.
+   *
+   * @example
+   * faker.helpers.objectEntry({ myProperty: 'myValue' }) // ['myProperty', 'value']
+   *
+   * @since 8.0.0
+   */
+  objectEntry<T extends Record<string, unknown>>(
+    object: T
+  ): [keyof T, T[keyof T]] {
+    const key = this.faker.helpers.objectKey(object);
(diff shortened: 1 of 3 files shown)


src/modules/helpers/index.ts
```diff
@@ -797,6 +797,25 @@ export class HelpersModule {
...
+   * Returns a random [key, value] pair from given object or `undefined` if no key could be found.
+   *
+   * @template T The type of the object to select from.
+   *
+   * @param object The object to be used.
+   *
+   * @example
+   * faker.helpers.objectEntry({ myProperty: 'myValue' }) // ['myProperty', 'value']
+   *
+   * @since 8.0.0
```
suyashgulati, line 816: Need more info here
As I am new I used the current version here, please let me know if in case this needs to be changed.

src/modules/helpers/index.ts
```diff
@@ -797,6 +797,25 @@ export class HelpersModule {
...
+   */
+  objectEntry<T extends Record<string, unknown>>(
+    object: T
+  ): [keyof T, T[keyof T]] {
+    const key = this.faker.helpers.objectKey(object);
+    return key ? [key, object[key]] : undefined;
```
matthewmayer: since this.faker.helpers.objectKey already throws if the object is empty, you don't need to check if key is defined. Can just do 

`return [key, object[key]];`

src/modules/helpers/index.ts
```diff
@@ -797,6 +797,25 @@ export class HelpersModule {
...
 
+  /**
+   * Returns a random [key, value] pair from given object or `undefined` if no key could be found.
+   *
+   * @template T The type of the object to select from.
+   *
+   * @param object The object to be used.
+   *
+   * @example
+   * faker.helpers.objectEntry({ myProperty: 'myValue' }) // ['myProperty', 'value']
```
matthewmayer: make the example have at least two keys and values

src/modules/helpers/index.ts
```diff
@@ -797,6 +797,25 @@ export class HelpersModule {
     return object[key];
   }
 
+  /**
+   * Returns a random [key, value] pair from given object or `undefined` if no key could be found.
```
matthewmayer: for consistency with the other object methods, it should throw if no key could be found, not return undefined

matthewmayer (changes requested):

test/helpers.spec.ts
```diff
@@ -935,6 +939,28 @@ describe('helpers', () => {
...
 
+      describe('objectEntry', () => {
+        it('should return a random key, value pair', () => {
+          const testObject = {
+            hello: 'to',
+            you: 'my',
+            friend: '!',
+          };
+          const actual = faker.helpers.objectEntry(testObject);
+          console.log(actual);
```
matthewmayer: remove the console.log

test/helpers.spec.ts
```diff
@@ -935,6 +939,28 @@ describe('helpers', () => {
...
       });
 
+      describe('objectEntry', () => {
+        it('should return a random key, value pair', () => {
+          const testObject = {
+            hello: 'to',
+            you: 'my',
+            friend: '!',
+          };
+          const actual = faker.helpers.objectEntry(testObject);
```
matthewmayer: can you do 
`const [key, value] = ...`
seems cleaner

test/helpers.spec.ts
```diff
@@ -935,6 +939,28 @@ describe('helpers', () => {
...
+          console.log(actual);
+
+          expect(Object.keys(testObject)).toContain(actual[0]);
+          expect(Object.values(testObject)).toContain(actual[1]);
+          expect(testObject[actual[0]]).toContain(actual[1]);
```
matthewmayer: this should be toEqual, not toContain right?

matthewmayer (changes requested):

src/modules/helpers/index.ts
matthewmayer: i think the `@example` in objectValue and objectKey could also be improved to add multiple key-values, but that can be done in a separate PR.

matthewmayer: that is correct for now, if this doesn't get merged before 8.0.0 we will change this before release.

src/modules/helpers/index.ts
suyashgulati: https://github.com/faker-js/faker/pull/2123/files#diff-3f6b4fc424ba2cedcbf5f612ba256d248f0c05f46314eed1dfa81590bcd429fdR767

Docs for objectKey and objectValue also mention it in the same way, hence maybe I was inclined towards doing that.

Do you want me to create a new PR to fix those as well?

suyashgulati:
@Shinigami92 Can you also assign issue #1764 to me as well.

Shinigami92:
> @Shinigami92 Can you also assign issue #1764 to me as well.

github doesn't let me select you, maybe you need to write a comment there

src/modules/helpers/index.ts
```diff
@@ -797,6 +797,25 @@ export class HelpersModule {
     return object[key];
   }
 
+  /**
+   * Returns a random [key, value] pair from given object or throws an Error if no key could be found.
```
```suggestion
   * Returns a random `[key, value]` pair from given object or throws an Error if no key could be found.
```

src/modules/helpers/index.ts
matthewmayer: Yea that would be good. Docs are definitely wrong for that.

src/modules/helpers/index.ts
matthewmayer: We can add a @throws jsdoc to explain the throwing error condition. See arrayElement for an example of this.

matthewmayer (changes requested):

src/modules/helpers/index.ts
xDivisionByZerox:
```suggestion
   * Returns a random `[key, value]` pair from the given object.
```

The throwing statement should be explained in the `@throw` tag as [@matthewmayer](https://github.com/faker-js/faker/pull/2123/files#r1183649038) pointed out.

src/modules/helpers/index.ts
suyashgulati: I have made those changes within this PR only [here](https://github.com/faker-js/faker/pull/2123/files#diff-3f6b4fc424ba2cedcbf5f612ba256d248f0c05f46314eed1dfa81590bcd429fdR767).
Hope this is OK. If not, I will create a new one.

xDivisionByZerox (approved):

Shinigami92 (approved):

matthewmayer (approved):

ST-DDT (approved):

PR #2121, opened 2023-05-03
infra!: drop NodeJS v14 and v16 support

Shinigami92:
Drop support for NodeJS v14 due to it reached EOL 2023-04-30

This is targeted for v9.0 and can be merged now ✅

PR #2124, opened 2023-05-03
infra: use managed pnpm

Shinigami92:
Instead of pinning the pnpm version, use the given pnpm version set in package.json

Example of that it works: https://github.com/faker-js/faker/actions/runs/4871503438/jobs/8688541770#step:4:16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,8 +93,6 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v2.2.4
-        with:
-          version: 7
 
       - name: Set node version to ${{ matrix.node_version }}
         uses: actions/setup-node@v3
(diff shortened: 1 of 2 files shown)


ST-DDT (approved):

xDivisionByZerox:
This should still stay?

https://github.com/faker-js/faker/blob/744b80fd5910e37d03edd6b3c7aab84e1163b180/.github/workflows/ci.yml#L39-L43

Shinigami92:
> This should still stay?
> 
> https://github.com/faker-js/faker/blob/744b80fd5910e37d03edd6b3c7aab84e1163b180/.github/workflows/ci.yml#L39-L43

exactly, this is to make node v14 support working, because node v14 needs pnpm v7

this will be removed in #2121

xDivisionByZerox (approved):

ejcheng (approved):

PR #2126, opened 2023-05-05
feat: new plant module

### LINKED ISSUE #2022: faker.plant.tree and faker.plant.flower

Similar to animal module, a plant module 

I feel this would be quite useful in its own right but also as a pattern for other modules

for example it’s common in English to name street names after trees eg Oak Steet, Pine Street. 

Trees and flowers are also often used as placeholder names eg naming meeting rooms, classes in a school etc. 

### Suggested solution

Add plant module with two methods to start with

### Alternative

_No response_

### Additional context

_No response_

sarahtang7:
Adds faker.plant.tree and faker.plant.flower as described in #2022 

Changes made:
- Added a new module, `plant`
- Added two new methods, `plant.tree` and `plant.flower` that return random tree and flower names
- Added data for tree and flower names
- Updated `en` locale data
- Added fixed seeded test class `plant.spec.ts` and snapshots


src/locales/en/plant/tree.ts
matthewmayer, line 1: For the trees, i think it would be better to keep the names to simpler, mostly one-word tree names. Then those could be used in other patterns like https://github.com/faker-js/faker/blob/next/src/locales/en/location/street_pattern.ts

For example "Pine Street" or "Ash Drive" is realistic, but "Lavender Twist Redbud Avenue" is not.

matthewmayer (changes requested):

src/index.ts
```diff
@@ -54,6 +54,9 @@ export type {
   PhoneNumberDefinition,
   /** @deprecated Use PhoneNumberDefinition instead */
   PhoneNumberDefinition as PhoneNumberDefinitions,
+  PlantDefinition,
+  /** @deprecated Use AnimalDefinition instead */
+  PlantDefinition as PlantDefinitions,
```
ST-DDT, line 59: This is redundant as this is a new definition that didnt exist before.

src/modules/plant/index.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @example
+   * faker.plant.tree() //'Willow'
+   *
+   * @since 8.0.0
+   */
+  tree(): string {
+    const definition = this.faker.helpers.arrayElement(
+      this.faker.definitions.plant.tree
+    );
+    return this.faker.helpers.fake(definition);
```
ST-DDT, line 33: There is no need to call `fake()` on them since the names dont have templates attached to them.

src/modules/plant/index.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @example
+   * faker.plant.flower() //'Poppy'
+   *
+   * @since 8.0.0
+   */
+  flower(): string {
+    const definition = this.faker.helpers.arrayElement(
+      this.faker.definitions.plant.flower
+    );
+    return this.faker.helpers.fake(definition);
```
ST-DDT, line 48: Same here

ST-DDT (commented):
Thanks for your contribution. Please note that we are approaching the v8.0 release and this PR likely wont make it in there.

Please run `pnpm run preflight`.

src/locales/en/plant/tree.ts
ST-DDT: Maybe we should have a separate method/option for simple names?

ST-DDT:
I also noticed the issue is marked as waiting for user interest.
It currently does not have enough up votes for immediate inclusion.

matthewmayer:
I think a new contributor making a pull request should count as 5 upvotes 🤣

ST-DDT:
It looks like some lint errors have creeped in over time.

ST-DDT:
Close as stale. We can reopen this when required/updated.

PR #2130, opened 2023-05-06
docs: streamline upgrading guide before release

### LINKED ISSUE #2079: Document what should appear in migration guide
Suggestion for new section to be added to CONTRIBUTING. 

---

## Documenting changes for new major versions

Each major version has an upgrading guide eg https://next.fakerjs.dev/guide/upgrading.html

While developing new features and bugfixes for the new release, changes are added to the migration guide to aid developers when the version is released. 

The general principle is to document anything which requires a normal user of the library to change their code which uses Faker when upgrading to the new major version. 

There are two sections
- Breaking changes (user MUST change their code)
- Deprecations and other changes (user SHOULD change their code but it will still work for this major version if they don't)

Not every change needs to be in the migration guide. If it is too long, it becomes hard for users to spot the important changes. 

### Should be in the guide

- Breaking changes, eg removal of methods
- Behavior changes, eg a different default for a parameter, or a parameter becoming required
- Whole modules renaming (eg faker.name to faker.person)
- Locale renames
- Changes to minimum versions eg requiring a new version of Node
- Changes to how Faker is imported

### Doesn't need to be in the guide
- New locales
- Changes to locale data in existing locales
- Bugfixes where it's unlikely anyone was relying on the old behavior (eg broken values in locale files)
- New methods and parameters 
- Straightforward method aliases, eg where a method or parameter is renamed but the old name still works identically. (Runtime warnings will already guide the user in this case)
- Changes to locale definition files which only affect usage via `faker.fake` e.g. if a definition file is renamed, but the public API for the method stays the same

matthewmayer:
re #2079

Not necessary to mention all straightforward method aliases, eg where a 
method or parameter is renamed but the old name still works identically. 
(Runtime warnings will already guide the user in this case)


Shinigami92:
Would you like the whole upgrading guide from top to bottom and maybe clean it up and or order stuff to your wish? That would prevent multiple small PRs. :slightly_smiling_face: 

If this is just the only change you wanted to make, then it's okay and we can think about a real release the next few days :ok_hand:

matthewmayer:
this is the only significant change i would suggest. I could probably do some minor polishing/reordering too. 

Also the "Since v8 has not yet been released, this is a work in progress list of any major and breaking changes in v8." should be removed before release, maybe makes sense to do that in this PR?

Shinigami92:
> this is the only significant change i would suggest. I could probably do some minor polishing/reordering too.
> 
> Also the "Since v8 has not yet been released, this is a work in progress list of any major and breaking changes in v8." should be removed before release, maybe makes sense to do that in this PR?

The docs are theoretically decoupled from the npm dist/package, but it is the v8 branch (https://github.com/faker-js/faker/tree/v8)
I think we want to make the release and then we will clean everything up post-release so the docs will be up-to-date in the v8 branch

it is a bit like a "hen egg"-problem :laughing:

matthewmayer:
i removed it for now, i can restore it if you want.

Once v8 is released, what will be at next.fakerjs.dev ? Will this be docs for say 8.0.1 or will it be already getting ready for v9.0.0?

ST-DDT:
> i removed it for now, i can restore it if you want.

Lets keep it removed.

> Once v8 is released, what will be at next.fakerjs.dev ? Will this be docs for say 8.0.1 or will it be already getting ready for v9.0.0?

We will probably do one v8.1 and then start with v9.0 (After some vacation).

Shinigami92:
- next branch is the next.fakerjs.dev
- v8 branch will be fakerjs.dev and v8.fakerjs.dev (currently fakerjs.dev is v7.fakerjs.dev)

Shinigami92:
docs do not really have a x.0.1 relevance, it is "just" the latest X.*

Shinigami92 (dismissed):

docs/guide/upgrading.md
```diff
@@ -270,14 +268,6 @@ The `faker.address.*` methods will continue to work as an alias in v8 and v9, bu
...
 | `faker.address.streetSuffix`        | _Removed_                            |
 
-### `faker.finance.account` changed to `faker.finance.accountNumber`
-
-The `faker.finance.account` method has been renamed to `faker.finance.accountNumber` to better reflect the data it returns and not to get confused with a user "Account".
-
-### `faker.finance.mask` changed to `faker.finance.maskedNumber`
-
-The `faker.finance.mask` method has been renamed to `faker.finance.maskedNumber` to better reflect its purpose.
-
```
ST-DDT, line 280: Should we add a hint instead, like "Have a look at our deprecations [here](https://github.com/faker-js/faker/issues?q=label%3Adeprecation+is%3Amerged+milestone%3A%22v8.0+-+Module+Re-Shuffling%22) and [here](https://github.com/faker-js/faker/issues?q=label%3Adeprecation+is%3Amerged+milestone%3A%22v8+-+None+milestone+specific+tasks%22)."?

matthewmayer: We could add a sentence at the top of the "## Deprecations and other changes" section like:

This is not an exhaustive list of all deprecations in v8.0.0. Many methods and parameters have been renamed in this release, runtime warnings will guide you to use the new names.

ST-DDT: I would still recommend actual links to the merged deprecation PRs, so it is easier to grasp without running the own code.

matthewmayer: The average consumer of the library doesn't need to grasp the full scope of all renamed methods and parameters though, they just need to know that a bunch of stuff has been renamed, and what affects their code, which is easiest to grasp by running said code and looking at the warnings.

Shinigami92: Theoretically we could also recommend in general to use https://www.npmjs.com/package/eslint-plugin-deprecation

matthewmayer: i added a summary encompassing several of these suggestions

cities.cjs
ST-DDT, line 1: Looks like this was committed by accident.

matthewmayer: yup reverted

docs/guide/upgrading.md
```diff
@@ -214,6 +212,13 @@ Some data definitions, which were only available via the `faker.helpers.fake` me
 
 ## Deprecations and other changes
 
+This is not an exhaustive list of all deprecations in v8.0.0. Many methods and parameters have been renamed in this release. You can:
+
+- use the warnings which are shown at runtime to guide you to the new names
+- use a [suitable plugin](https://www.npmjs.com/package/eslint-plugin-deprecation) to find usages of deprecated code
+- Review the full list of deprecations [here](https://github.com/faker-js/faker/issues?q=label%3Adeprecation+is%3Amerged+milestone%3A%22v8.0+-+Module+Re-Shuffling%22) and [here](https://github.com/faker-js/faker/issues?q=label%3Adeprecation+is%3Amerged+milestone%3A%22v8+-+None+milestone+specific+tasks%22)
+- Ignore the deprecations for now: the old names will continue to work for version 8.x.x, however you will need to make changes before upgrading to 9.x.x.
```
ST-DDT: I'm not sure which is better.

```suggestion
- Ignore the deprecations for now: the old names will continue to work for v8.x.x, however you will need to make changes before upgrading to v9.x.x.
```

ST-DDT (dismissed):
This really a nice addition to the text.

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #2140, opened 2023-05-08
refactor(locale): cs_CZ state_abbr not applicable

xDivisionByZerox:
Set the locale data of `cs_CZ.location.state_abbr` to non-applicable.
The previous locale file even included a comment hinting in that direction.


ST-DDT (approved):

test/all_functional.spec.ts
```diff
@@ -30,7 +30,7 @@ const BROKEN_LOCALE_METHODS = {
   },
   location: {
     state: ['az', 'nb_NO', 'ro_MD', 'sk'],
-    stateAbbr: ['ro_MD', 'sk'],
+    stateAbbr: ['cs_CZ', 'ro_MD', 'sk'],
```
xDivisionByZerox, line 33: Since this is not really broken, I'll suggest an additional PR that splits these locales into broken and non-applicable. Thoughts?

matthewmayer: i agree BROKEN_LOCALE_METHODS is a bad name but not sure how you'd distinguish between "broken" and "non-applicable".

xDivisionByZerox: Broken = not provided from the community or us => undefined => throws error on access
Non Applicable = is provided as `null` => throws an error on 

At least that is what I think.

matthewmayer: It doesn't seem very DRY to have to both declare null in the definition files, and have to explicitly add to a list of "method/locales which return null" though.

xDivisionByZerox: Yes, that is exactly the reason why I would like to refactor this^^ (because currently we need to do this)

matthewmayer: So basically the NULL ones don't have to be explicitly listed here any more?

xDivisionByZerox: Yes, that's what I would want to achieve. How that is specifically don't can be discussed in the other PR.

ST-DDT: IMO we should check if the method throws and if it throws the not applicable error and then not fail.

ST-DDT: IMO we should check if the method throws and if it throws the not applicable error and then not fail.

matthewmayer (approved):

ejcheng (approved):

PR #2141, opened 2023-05-08
fix(locales): add real cities for zu_ZA and af_ZA, update en_ZA city list

matthewmayer:
Changes faker.location.city() to use real city names for af_ZA and zu_ZA, like en_ZA

Not all cities have seperate names in all three languages

Originally discussed at https://github.com/faker-js/faker/pull/2119#discussion_r1185272990


src/locales/zu_ZA/location/city_name.ts
```diff
@@ -0,0 +1,16 @@
+export default [
+  'Polokwane',
+  'eGoli',
```
ST-DDT, line 3: Note to self check these weird names.

matthewmayer: Zulu has some curious capitalization!

https://zu.wikipedia.org/wiki/IRiphabhuliki_yaseNingizimu_Afrika

ST-DDT (approved):

Shinigami92 (approved):

PR #2142, opened 2023-05-09
feat(locale): add states for Croatia (hr)

matthewmayer:
Part of #1977

There are 21 counties

https://en.wikipedia.org/wiki/Counties_of_Croatia
https://hr.wikipedia.org/wiki/Hrvatske_%C5%BEupanije


src/locales/hr/location/state.ts
```diff
@@ -0,0 +1,23 @@
...
+  'Bjelovarsko-bilogorska',
+  'Brodsko-posavska',
+  'Dubrovačko-neretvanska',
+  'Grad Zagreb',
+  'Istarska',
+  'Karlovačka',
+  'Koprivničko-križevačka',
+  'Krapinsko-zagorska',
+  'Ličko-senjska',
+  'Međimurska',
```
ST-DDT, line 11: According to the pages you have linked I could only find: [ Međimurska županija](https://hr.wikipedia.org/wiki/Me%C4%91imurska_%C5%BEupanija) which I would call a county.

grafik

ST-DDT (commented):
It might be best to just call them counties here.
As for states we could use these https://en.wikipedia.org/wiki/Regions_of_Croatia or omit them entirely.

src/locales/hr/location/state.ts
matthewmayer: https://en.wikipedia.org/wiki/ISO_3166-2:HR ISO 3166-2 uses the counties as the 1st level administrative subdivisions, the regions seem to be more historical and not included in a typical address if i am not mistaken, so i feel the counties should be under `state()`

ST-DDT: While technically it is the first level adminstrative subdivision, I would still use them as counties and simply mark state as being not applicable due to them having county as their name.

ST-DDT:
**Team Decision**

We will jump to a definition loophole and will put it in both state (as first level administrative unit) and in county (as county names).

We will create a separate issue to clarify that states ~= `ISO_3166-2`

ST-DDT (changes requested):
Please put these in both state and county. County should import state.

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2143, opened 2023-05-09
feat(metadata): add method to access metadata

### LINKED ISSUE #1984: Add method to access the locale metadata

Currently there is no official way to access the locale metadata provided to the current faker instance.

### Suggested solution

Add `faker.metadata() : MetadataDefinitions`

### Alternative

Promote `faker.raw/definitions.metadata`

### Additional context

Created based on https://github.com/faker-js/faker/issues/1978#issuecomment-1486569973

matthewmayer:
fix #1984


src/faker.ts
```diff
@@ -458,6 +458,21 @@ export class Faker {
     return seed;
   }
 
+  /**
+   * Returns an object with metadata about the current locale
+   *
+   * @returns An object including the locale title, code, language, endonym (native name), direction (left to right, or right to left), script, and optional country.
```
ST-DDT: We usually dont document the return value with jsodcs, do we?

src/faker.ts
```diff
@@ -458,6 +458,21 @@ export class Faker {
...
+   *
+   * @example
+   * import { faker, fakerES_MX } from '@faker-js/faker';
+   * // const { faker, fakerES_MX } = require("@faker-js/faker")
+   * faker.metadata(); // { title: 'English', code: 'en', language: 'en', endonym: 'English', dir: 'ltr', script: 'Latn' }
+   * fakerES_MX.metadata(); // { title: 'Spanish (Mexico)', code: 'es_MX', language: 'es', endonym: 'Español (México)', dir: 'ltr', script: 'Latn', country: 'MX' }
+   */
+  metadata(): MetadataDefinition {
+    return this.rawDefinitions.metadata;
```
ST-DDT:
```suggestion
    return this.rawDefinitions.metadata ?? {};
```

src/faker.ts
```diff
@@ -458,6 +458,21 @@ export class Faker {
...
+   * @returns An object including the locale title, code, language, endonym (native name), direction (left to right, or right to left), script, and optional country.
+   *
+   * @example
+   * import { faker, fakerES_MX } from '@faker-js/faker';
+   * // const { faker, fakerES_MX } = require("@faker-js/faker")
+   * faker.metadata(); // { title: 'English', code: 'en', language: 'en', endonym: 'English', dir: 'ltr', script: 'Latn' }
+   * fakerES_MX.metadata(); // { title: 'Spanish (Mexico)', code: 'es_MX', language: 'es', endonym: 'Español (México)', dir: 'ltr', script: 'Latn', country: 'MX' }
+   */
+  metadata(): MetadataDefinition {
+    return this.rawDefinitions.metadata ?? {};
```
ST-DDT, line 471: We need a test for this method.

ST-DDT (changes requested):

ST-DDT (dismissed):

src/faker.ts
xDivisionByZerox: Yes, we don't. Furthermore, by explicitly listing all available properties we decouple the documentation from the actual implementation. If one changes we need to remember to change the other as well. This is not good.

I would suggest dropping this line.

src/faker.ts
```diff
@@ -458,6 +458,21 @@ export class Faker {
     return seed;
   }
 
+  /**
+   * Returns an object with metadata about the current locale
```
xDivisionByZerox:
```suggestion
   * Returns an object with metadata about the current locale.
```

src/faker.ts
```diff
@@ -458,6 +458,21 @@ export class Faker {
...
+   *
+   * @returns An object including the locale title, code, language, endonym (native name), direction (left to right, or right to left), script, and optional country.
+   *
+   * @example
+   * import { faker, fakerES_MX } from '@faker-js/faker';
+   * // const { faker, fakerES_MX } = require("@faker-js/faker")
+   * faker.metadata(); // { title: 'English', code: 'en', language: 'en', endonym: 'English', dir: 'ltr', script: 'Latn' }
+   * fakerES_MX.metadata(); // { title: 'Spanish (Mexico)', code: 'es_MX', language: 'es', endonym: 'Español (México)', dir: 'ltr', script: 'Latn', country: 'MX' }
+   */
+  metadata(): MetadataDefinition {
```
xDivisionByZerox: The method name and type are conflicting with the rest of the faker class. All subjects (`person`, `definitions`, etc) are properties. Methods are verbs or include prefixes like `get`/`set` (`setDefaultRefDate`). This method breaks this implicit convention and might cause confusion.

xDivisionByZerox (commented):
I'm a bit conflicted about why we extract the metaobject access from the definitions. Currently, you can access them fine just fine via `faker.definitions.metadata`. What is the benefit of the implementation of this PR?

src/faker.ts
matthewmayer: removed the @returns line

matthewmayer:
> I'm a bit conflicted about why we extract the metaobject access from the definitions. Currently, you can access them fine just fine via `faker.definitions.metadata`. What is the benefit of the implementation of this PR?

1. its not documented
2. we want the freedom to be able to rearrange things within `definitions` without it being a breaking change.

src/faker.ts
matthewmayer: so `getMetadata()` ?

xDivisionByZerox: I would suggest doing so. Alternatively, you could refactor the function to a `getter`. I don't have a preference between those  two.

xDivisionByZerox (approved):

ST-DDT (approved):

Shinigami92:
Do not merge yet, so we can release a v8.0.2 the next few days
Generally do not merge feat+ into next, only fix, chore, test and docs

PR #2144, opened 2023-05-09
test: vitest prefer-each

Shinigami92:
references https://github.com/faker-js/faker/pull/2098#issuecomment-1524051531

### TODOs and Notes for myself

- we should check if we need to set the seed in a setup block, because it might be that the description title is not executed again
  - right now I assume the repeats might handle the faker instance differently than the previous for-loop, but I need to check that
- the global unique was/is repeated 🤔 how does it behave if the seed gets reset? does this affect the global store? did we fixed it now accidentally via faker instance store?

---

Helpful docs to understand the .each string syntax: https://vitest.dev/api/#test-each

For debugging purposes it can help quite a log to set `reporters: 'verbose'` in vitest config!


Shinigami92:
@ST-DDT could you have a first look and re-evaluate if I did not made any mistakes?

test/color.spec.ts
```diff
@@ -402,6 +403,7 @@ describe('color', () => {
           expect(color).match(/^([01]{8,32} [01]{8,32} [01]{8,32})$/);
         });
       });
-    }
-  });
+    },
+    { repeats: NON_SEEDED_BASED_RUN }
```
ST-DDT: This is formatted differently then everything else.

test/helpers.spec.ts
```diff
@@ -1214,118 +1215,121 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`)
...
 
-    // This test can be only executed once, because the unique function has a global state.
-    // See: https://github.com/faker-js/faker/issues/371
-    describe('global unique()', () => {
-      it('should be possible to exclude results as array', () => {
-        const internetProtocol = () =>
-          faker.helpers.arrayElement(['https', 'http']);
-        const result = faker.helpers.unique(internetProtocol, [], {
-          exclude: ['https'],
+      describe('multiple()', () => {
```
ST-DDT, line 1220: Why was this method used?

Shinigami92: It moved up
Only the global store stuff was extracted out of the loop

test/color.spec.ts
Shinigami92: Good catch 👌

Shinigami92:
I'm now experimenting a bit with the `repeats` and it looks like that the "string" description title does not get executed again (as assumed).
So we need to move the `seed()` into a `beforeAll` or we need to use `describe.each`.
I think `beforeAll` is more clear (but sadly more verbose) here

Shinigami92:
~~blocked by https://github.com/vitest-dev/vitest/issues/3334~~

- https://github.com/vitest-dev/vitest/issues/3334

we will use the "old" strategy for now

test/locale-data.spec.ts
```diff
@@ -8,11 +8,9 @@ function checkLocaleData(data: unknown) {
       expect(data).not.toContainDuplicates();
     });
   } else if (typeof data === 'object' && data != null) {
-    for (const [nestedKey, nestedData] of Object.entries(data)) {
-      describe(`${nestedKey}`, () => {
-        checkLocaleData(nestedData);
-      });
-    }
+    describe.each(Object.entries(data))('%s', (nestedData) => {
+      checkLocaleData(nestedData);
```
Shinigami92, line 12: @ST-DDT this never runs :thinking:

ST-DDT: Why?

Shinigami92: good question, I'm asking you ^^

ST-DDT: On the `next` branch, it does get checked.

Shinigami92: okay I will check again

test/locale-data.spec.ts
```diff
@@ -8,11 +8,9 @@ function checkLocaleData(data: unknown) {
       expect(data).not.toContainDuplicates();
     });
   } else if (typeof data === 'object' && data != null) {
-    for (const [nestedKey, nestedData] of Object.entries(data)) {
-      describe(`${nestedKey}`, () => {
-        checkLocaleData(nestedData);
-      });
-    }
+    describe.each(Object.entries(data))('%s', (nestedData) => {
```
ST-DDT:
```suggestion
    describe.each(Object.entries(data))('%s', (_, nestedData) => {
```

ST-DDT (changes requested):

test/all_functional.spec.ts
```diff
@@ -137,94 +137,84 @@ describe('BROKEN_LOCALE_METHODS test', () => {
...
-            }
-          });
-        });
+  describe.each(Object.entries(allFakers))('%s', (locale, faker) => {
+    if (locale === 'base') {
+      it.skip('base locale is checked by other tests');
+      return;
+    }
+
+    describe.each(Object.keys(modules))('%s', (module) => {
```
ST-DDT: Maybe

```suggestion
    describe.each(Object.entries(modules))('%s', (module, methods) => {
```

test/all_functional.spec.ts
```diff
@@ -137,94 +137,84 @@ describe('BROKEN_LOCALE_METHODS test', () => {
...
+      modules[module].forEach((meth) => {
```
ST-DDT: Maybe

```suggestion
      methods.forEach((meth) => {
```

test/all_functional.spec.ts
```diff
@@ -137,94 +137,84 @@ describe('BROKEN_LOCALE_METHODS test', () => {
...
-            }
-          });
-        });
+  describe.each(Object.entries(allFakers))('%s', (locale, faker) => {
+    if (locale === 'base') {
+      it.skip('base locale is checked by other tests');
+      return;
+    }
+
+    describe.each(Object.keys(modules))('%s', (module) => {
```
ST-DDT: Same here

test/location.spec.ts
```diff
@@ -363,44 +364,50 @@ describe('location', () => {
...
-              expect(actualDistance).toBeLessThanOrEqual(radius);
+        it.each(
+          times(100).reduce<Array<[{ isMetric: boolean; radius: number }]>>(
+            (prev, curr) => {
+              prev.push([{ isMetric: true, radius: curr }]);
+              prev.push([{ isMetric: false, radius: curr }]);
+              return prev;
+            },
+            []
+          )
```
ST-DDT:
```suggestion
          times(100).flatMap((i) => [
            { isMetric: true, radius: i },
            { isMetric: false, radius: i },
          ])
```

test/mersenne.spec.ts
```diff
@@ -2,51 +2,55 @@ import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
...
-  for (const seed of [...seededRuns, [42, 1, 2], [1337, 1, 2], [1211, 1, 2]]) {
-    describe(`seed: ${JSON.stringify(seed)}`, () => {
-      beforeEach(() => {
-        mersenne.seed(seed);
-      });
+  describe.each(
+    (seededRuns as Array<number | number[]>)
+      .concat([[42, 1, 2]], [[1337, 1, 2]], [[1211, 1, 2]])
+      .map((s) => [s])
+  )('seed: %j', (seed) => {
```
ST-DDT:
```suggestion
  describe.each(
    [...seededRuns, [42, 1, 2], [1337, 1, 2], [1211, 1, 2]].map((v) => [v])
  )('seed: %j', (seed) => {
```

test/mersenne.spec.ts
```diff
@@ -2,51 +2,55 @@ import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
...
+  >((prev) => {
+    prev.push([Math.ceil(Math.random() * 1_000_000_000)]);
+    prev.push([
+      [
+        Math.ceil(Math.random() * 1_000_000_000),
+        Math.ceil(Math.random() * 1_000_000_000),
+      ],
+    ]);
+    return prev;
+  }, []);
```
ST-DDT: Please use flatMap instead.

test/mersenne.spec.ts
ST-DDT: The `.map((v) => [v])` is only needed due to some inconsistency in the handling of %j vs the args,

````txt
stdout | test/mersenne.spec.ts > mersenne twister > seed: 1337 > should return deterministic value for next()
seed:  1337

stdout | test/mersenne.spec.ts > mersenne twister > seed: 42 > should return deterministic value for next()
seed:  [ 42, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1211 > should return deterministic value for next()
seed:  [ 1211, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1211 > should return deterministic value for next()
seed:  1211

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1337 > should return deterministic value for next()
seed:  [ 1337, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 42 > should return deterministic value for next()
seed:  42
````

Comment (ST-DDT) [reply test/mersenne.spec.ts:16]: IMO we should report that to vitest as a bug.

Comment (ST-DDT) [reply test/mersenne.spec.ts]: The `.map((v) => [v])` is only needed due to some inconsistency we otherwise have in the handling of %j vs the args,

````txt
stdout | test/mersenne.spec.ts > mersenne twister > seed: 1337 > should return deterministic value for next()
seed:  1337

stdout | test/mersenne.spec.ts > mersenne twister > seed: 42 > should return deterministic value for next()
seed:  [ 42, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1211 > should return deterministic value for next()
seed:  [ 1211, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1211 > should return deterministic value for next()
seed:  1211

stdout | test/mersenne.spec.ts > mersenne twister > seed: 1337 > should return deterministic value for next()
seed:  [ 1337, 1, 2 ]

stdout | test/mersenne.spec.ts > mersenne twister > seed: 42 > should return deterministic value for next()
seed:  42
````

IMO we should report that to vitest as a bug.

Comment (Shinigami92) [reply test/mersenne.spec.ts]: I'm not sure if this is a bug upstream 🤔 
vitest expect an array of "tuple"s that are spreaded as args.

So `[1337]` becomes `(1337) => {}` but `[1337, 1, 2]` would become `(1337, 1, 2) => {}`
But we want `([1337, 1, 2]) => {}` instead, so we need to wrap them in a tuple with only one first value.

and then `.each` takes an array of tuples so we would have:

```ts
describe.each([
  [42],
  [1337],
  [1211],
  [[42, 1, 2]],
  [[1337, 1, 2]],
  [[1211, 1, 2]],
])...
```

Do you think we can somehow improve the code on our side to achieve the wanted result in a better way?

ST-DDT: > vitest expect an array of "tuple"s that are spreaded as args.

Then why doesnt it throw a type error if called otherwise?

> Do you think we can somehow improve the code on our side to achieve the wanted result in a better way?

IMO my suggestion pretty much handles this. We could shorten it further by mapping the arrays as well. :man_shrugging: 

```suggestion
  describe.each(
    [...seededRuns, ...seededRuns.map((v) => [v, 1, 2])].map((v) => [v])
  )('seed: %j', (seed) => {
```

And maybe extract the `.map((v) => [v])` part in a separate method called tuplize or similar.

test/mersenne.spec.ts
```diff
@@ -2,51 +2,54 @@ import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
...
-  }
+  });
 
   // Create and log-back the seed for debug purposes
-  const seeds = [
-    Math.ceil(Math.random() * 1_000_000_000),
+  const seeds: Array<[seed: number | number[]]> = times(
+    NON_SEEDED_BASED_RUN
+  ).flatMap(() => [
+    [Math.ceil(Math.random() * 1_000_000_000)],
```
ST-DDT: IMO we can shorten this by extracting this in a separate method called `randomSeed()`.
Then it should be short enough to inline `seeds`.

ST-DDT (approved):

ejcheng (approved):

test/company.spec.ts
```diff
@@ -23,8 +24,9 @@ describe('company', () => {
     t.skip('bs').skip('bsAdjective').skip('bsBuzz').skip('bsNoun');
   });
 
-  describe(`random seeded tests for seed ${faker.seed()}`, () => {
-    for (let i = 1; i <= NON_SEEDED_BASED_RUN; i++) {
+  describe.each(times(NON_SEEDED_BASED_RUN).map(() => faker.seed()))(
+    'random seeded tests for seed %i',
```
xDivisionByZerox, line 28: Sorry if this might be a noob question, but why is the difference between `%i`,`%s` and `%j`? 
I know that `%s` takes the first argument of each description as the value, in jest. Is that the same in vitest? `%i` is probably for index (?). But what does `%j` do?

Shinigami92: https://vitest.dev/api/#test-each

xDivisionByZerox: I've searched in the docs for `descrive.each`...mb

xDivisionByZerox (approved):

PR #2148, opened 2023-05-11
feat(location): add ne postcodes

matthewmayer:
Nepali postcodes are 5 digits starting with [1-5]

https://www.gpo.gov.np/Home/Postalcode


ST-DDT (approved):

xDivisionByZerox (approved):
I wasn't really able to figure out the zip codes from the source you provided. 
https://www.nepalitrends.com/list-of-postal-codes-of-nepal/ did help me tho. Additionally, it has a full table of all possible postcodes, as the current templates could generate some invalid zip codes. But it's good enough IMO.

ejcheng (approved):

PR #2156, opened 2023-05-13
feat(location): es province should be county as its a 2nd level administrative area

matthewmayer:
es province should be county as its a 2nd level administrative area

Allows it to be returned by fakerES.location.county()

Currently it would only work via faker.helpers.fake

https://en.wikipedia.org/wiki/Political_divisions_of_Spain
https://en.wikipedia.org/wiki/Provinces_of_Spain


ST-DDT (approved):

ejcheng (approved):

xDivisionByZerox (approved):

xDivisionByZerox:
Should the explanation from the title ("as its a 2nd level administrative") move to the description?

ejcheng:
> Should the explanation from the title ("as its a 2nd level administrative") move to the description?

Yeah, we should keep the title to just `feat(location): es province should be county`

PR #2172, opened 2023-05-23
fix(git): limit need for Intl to specific method

### LINKED ISSUE #2171: Using Faker in react-native: `Can't find variable: Intl` error on android
### Pre-Checks

### Describe the bug

After upgrading from version 7 to version 8, our react-native application crashes on Android devices.

We uses Expo 48.0.17, react-native 18.0.2 and tested on Android Tiramisu

Callstacks points to code here:

https://github.com/faker-js/faker/blob/614fd67f436546bafe1a1cafa579427255144274/src/modules/git/index.ts#L4

Assumption is that `Intl.DateTimeFormat` does not exists for react-native on android. 

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 7.80 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    Yarn: 3.5.0 - ~/.nvm/versions/node/v18.16.0/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
    Watchman: 2023.05.08.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 113.0.5672.126
    Safari: 16.4
```

- [X] CJS
- [ ] ESM

### Used Package Manager

yarn

matthewmayer:
#2171


Shinigami92:
@matthewmayer you already found out how to fake that Intl does not exists, or alternatively you can also just mock it with vitest

Please add tests for this

matthewmayer:
> @matthewmayer you already found out how to fake that Intl does not exists, or alternatively you can also just mock it with vitest
> 
> Please add tests for this

Hmm, the problem is that Intl is checked only once at import time when the library is loaded. So you can't just do something like this:

```js
describe('commitDate', () => {
        it('should return a random commitDate', () => {
          const commitDate = faker.git.commitDate();

          expect(commitDate).toBeTruthy();
          expect(commitDate).toBeTypeOf('string');

          const parts = commitDate.split(' ');
          expect(parts.length).toBe(6);
        });
        it('should throw if Intl is unavailable', () => {
          const oldIntl = Intl;
          (global as any).Intl = undefined;

          expect(() => {
            faker.git.commitDate();
          }).toThrow(
            new FakerError(
              'This method requires an environment which supports Intl.NumberFormat and Intl.DateTimeFormat'
            )
          );

          (global as any).Intl = oldIntl;
        });
      });
```

Shinigami92:
@matthewmayer So this is not working? https://vitest.dev/guide/mocking.html#globals
If so, please report to Vitest

matthewmayer:
> @matthewmayer So this is not working? https://vitest.dev/guide/mocking.html#globals If so, please report to Vitest

not really sure how that would work. seems to be for mocking globals you DON'T have, not removing globals you do have?

Shinigami92:
get creative, try a bit around :slightly_smiling_face:

matthewmayer:
Can't get it to work. I can't seem to get code to run in the test before faker is imported. 

This is a way beyond my capabilities so if anyone else is available to add tests I'd appreciate it.

Shinigami92:
@matthewmayer It took me a fraction of a second to just do this in `git.spec.ts`

image

(_ignore the TS error, we are talking about raw JavaScript here_)

Then I logged just 

image

in `commitDate` function

Now you just need to move the check-logic to runtime, so you don't check `GIT_DATE_FORMAT_BASE == null` but `Intl.DateTimeFormat == null` instead of the `console.log` I did

I hope this helps to bring you forward

matthewmayer:
Ah ok moving the check logic to run time indeed makes it easy.

Glad I only wasted a fraction of a second of your time 😂

Shinigami92 (dismissed):
IMO it is okay for now to do it the dirty way and use `globalThis.Intl.DateTimeFormat = backup`
It could result in broken further test if the test does not run successfully, but common would this occur :shrug: 

---

Edit: dammit, somehow the ts script check failed
I will try to fix this later... sry for that

Shinigami92 (dismissed):

ejcheng (dismissed):

src/modules/git/index.ts
```diff
@@ -175,6 +181,11 @@ export class GitModule {
     } = {}
   ): string {
     const { refDate = this.faker.defaultRefDate() } = options;
+    if (!Intl || !Intl.DateTimeFormat || !Intl.NumberFormat) {
```
ST-DDT, line 189: Maybe add a comment why you check these instead of our used constants.

matthewmayer: to get the tests to work? :D

ST-DDT (approved):

src/modules/git/index.ts
```diff
@@ -175,6 +181,12 @@ export class GitModule {
     } = {}
   ): string {
     const { refDate = this.faker.defaultRefDate() } = options;
+    // We check if Intl support is missing rather than if GIT_DATE_FORMAT_BASE/GIT_TIMEZONE_FORMAT is null. This allows us to test the error case in environments that do have Intl support by temporarily removing Intl at runtime.
```
Shinigami92, line 188: How about this?

```suggestion
    // https://github.com/faker-js/faker/issues/2171
```

ST-DDT: The linked issue is totally unrelated to why the comment is there.
The comment is there to show why we check for `Intl`, rather then `GIT_DATE_FORMAT_BASE/GIT_TIMEZONE_FORMAT`

Shinigami92: Accept of the fact that react-native is for example a real world use case

Beside that... The error message below already says why this condition exist, so why an extra comment?

Just also a random source to underly my opinion:
[YouTube Video: Don't Write Comments](https://www.youtube.com/watch?v=Bf7vDBBOBUA&ab_channel=CodeAesthetic)

ST-DDT: Alternatively, we could use scoped imports that are loaded while `Intl` is absent.

````ts
const noIntlGitModule;

beforeAll(() => {
  const backup = Intl;
  Intl = null;
  const NoIntlGitModule = import('../src/modules/git?noIntl');
  noIntlGitModule = new NoIntlGitModule(faker);
  Intl = backup;
});

it("test", () => noIntlGitModule.commitEntry());
````

@Shinigami92 What do you think?

Shinigami92: looks okay to me (if this works) 🙂 

I just did not use `before/after`, because it would need to wrap the test in extra describe blocks and so on, and then I thought it overcomplexifies the test suite and the current solution just works
But if you want to do it the clean way, go on 👍

matthewmayer: I don't think we need a particularly beautiful solution for this. It's one relatively unimportant method. More important to fix the bug.

Shinigami92: Yeah, this is why I recommended to keep it simple and don't need to explain in a comment what the code is doing
The reason in the comment is also unnecessary as of if you would change the if statement, the tests would fail
And if you then change the tests, you doing it wrong!
So to me the comment is 100% irrelevant and does not provide any value

So we could e.g. shorten this to just mention that we "support" environments without Intl and remove the rest left and right
But even this can be indirectly found out via git blame 👀 
If you do not have git blame or looking at the source code with Notepad, then a link to the source issue is helpful if you want to find out more

matthewmayer: @ST-DDT what do you think? Just remove the comment again? It was not originally commented :D

I agree with the logic that, if someone came along and thought "hmm wouldnt it make more sense to check for GIT_DATE_FORMAT_BASE==null here instead" that would immediately cause the tests to fail, so they wouldn't get far down the rabbit hole

ejcheng (dismissed):

src/modules/git/index.ts
ST-DDT: My idea behind the comment is to save the future `us`es/someones from wasting time in frustration.
I watched the video and I agree with it. But the comment is NOT there to explain what it does, but why it does this not obvious thing (one of the few exceptions the video mentions for comments).

My idea behind the test code suggestion, is to remove this hard to understand code from the source code aka having better code (also like suggested in the video). I actually prefer this over the comment as it is "more obvious".

Here is my POC:

- [Branch](https://github.com/faker-js/faker/tree/pr/matthewmayer/2172-ST-DDT)
- [Diff to this PR](https://github.com/matthewmayer/faker/compare/fix/limit-intl-error...faker-js:faker:pr/matthewmayer/2172-ST-DDT)
- [Diff to next](https://github.com/faker-js/faker/compare/pr/matthewmayer/2172-ST-DDT)

ST-DDT: If you think this is superfluous, you can mark this as resolved.

matthewmayer: your solution is clever, but i dont really like the way that the throw() is not longer within the commitDate method that actually throws. I'd say lets just mark resolved for now, i think all solutions here are going to involve some level of confusing code.

Shinigami92: In the danger of that in e.g. a year someone stumbels over the code and argument with https://caniuse.com/?search=intl.%20NumberFormat that every environment "now" supports it,
for the sake of god (and that we can just ship and satisfy our customers) I will approve...

Shinigami92 (dismissed):

ST-DDT (changes requested):
Missing @throws jsdocs.

matthewmayer:
> Missing @throws jsdocs.

added @throws

Shinigami92 (approved):

ejcheng (approved):

ST-DDT (approved):

PR #1871, opened 2023-02-23
feat: return randomly one of animals different types

yasinpalizban:
A new feature  for animal module, The function called ainmal.random() which  select randomly  of  animal types then pick up randomly one of selected animal.

here's  code
`
faker.ainmal.random();

`

PR #2174, opened 2023-05-24
feat(word): add param to enable random word capitalization 

### LINKED ISSUE #2165: casing for random words

The deprecated `faker.random.words()` has an option for mixed case, all caps, all lower, etc.
The suggested replacements (`faker.lorem.words()`, `faker.word.words()`) don't appear to offer the same functionality, even with non-default arguments. When updating to use the suggested replacements, I imagine some people might like to be able to still ensure that capital letters are, at least possible.

### Suggested solution

Add arguments to these methods to allow the user to influence the case without having to add their own logic to capitalize the case in a random or non-random way.

### Alternative

Presumably a user could use a custom function or string methods to manipulate the case.

### Additional context

For example:
```
qua ability South Books
Frozen Soap Gorgeous up
Modern Potassium Cab Wisconsin
```

lorem:
```
molestiae ut doloremque eos
perferendis veritatis nihil inventore
ducimus provident nam vero
```

word.words:
```
culture around sunlamp reform
utterly and needily occasional
book upright whoa sane
```

ZweZeya:
… to address issue #2165


src/modules/word/index.ts
```diff
@@ -638,4 +638,33 @@ export class WordModule {
       .multiple(() => this.sample(), { count })
       .join(' ');
   }
+
+  capitalize(
```
matthewmayer: would be better to make an additional option to the `words` method than a completely seperate method

src/modules/word/index.ts
```diff
@@ -638,4 +638,33 @@ export class WordModule {
...
+              /**
+               * The maximum number of words to return.
+               */
+              max: number;
+            };
+      } = {}
+  ): string {
+    return this.words(options)
+      .split(' ')
+      .map(word => Math.random() > 0.5 ? word.charAt(0).toUpperCase() + word.slice(1) : word)
```
matthewmayer: use faker.datatype.boolean instead of a raw Math.random call. This ensures it uses the faker seed for reproducible results.

matthewmayer (changes requested):

src/modules/word/index.ts
ST-DDT: IMO it is unclear what this method does exactly. because it doesn't only capitalize but also generate random words.

src/modules/word/index.ts
ZweZeya: okay I will take a look into it, I will also integrate this into the words method as part of the options

src/modules/word/index.ts
```diff
@@ -596,6 +596,7 @@ export class WordModule {
    *
    * @param options The optional options object or the number of words to return.
    * @param options.count The number of words to return. Defaults to a random value between `1` and `3`.
+   * @param options.capitalize Whether the words are capatalized. Defaults to `none`
```
matthewmayer: Typo "capatalized" should be "capitalized"

src/modules/word/index.ts
```diff
@@ -626,16 +627,40 @@ export class WordModule {
                  */
                 max: number;
               };
+          /**
+           * Whether words are capatalized.
+           *
+           * @default none
+           */
+          capatilize?: 'random' | 'all' | 'none';
```
matthewmayer: typo should be capitalize

src/modules/word/index.ts
```diff
@@ -626,16 +627,40 @@ export class WordModule {
                  */
                 max: number;
               };
+          /**
+           * Whether words are capatalized.
```
matthewmayer: "capitalized". It would be good to explain what each option does, e.g. its not clear if all means "Foo Bar" or "FOO BAR"

matthewmayer (changes requested):
you should also add some tests in test/word.spec.ts to test the parameter works as expected

src/modules/word/index.ts
ZweZeya: sure, I will amend the typo and try to add some tests in test/word.spec.ts

ZweZeya:
image

Hi, I am running into this error when I run ```pnpm run preflight```.

matthewmayer:
> image
> 
> Hi, I am running into this error when I run `pnpm run preflight`.

this came up before, check you have v7 and v8 branches checked out as well, not just next branch https://github.com/faker-js/faker/pull/1871#issuecomment-1445431009

src/modules/word/index.ts
```diff
@@ -626,16 +627,44 @@ export class WordModule {
...
     if (typeof options === 'number') {
-      options = { count: options };
+      options = { count: options, capitalize: 'none' };
     }
 
     const { count = { min: 1, max: 3 } } = options;
 
-    return this.faker.helpers
+    let sentence = this.faker.helpers
       .multiple(() => this.sample(), { count })
```
matthewmayer: it would probably make more sense to move the logic for capitalizing inside the function here

`() => this.sample()`

That avoids having to join, split, and then join again

matthewmayer (changes requested):

src/modules/word/index.ts
ZweZeya: true, I'll give it a shot

ZweZeya:
> > image
> > Hi, I am running into this error when I run `pnpm run preflight`.
> 
> this came up before, check you have v7 and v8 branches checked out as well, not just next branch [#1871 (comment)](https://github.com/faker-js/faker/pull/1871#issuecomment-1445431009)

I haven't checked other branches, but I have checked past commits. It is the same error.

ST-DDT:
> I haven't checked other branches, but I have checked past commits. It is the same error.

You probably have to run `git fetch --all` once to "download" the v6/v7 branches.

src/modules/word/index.ts
```diff
@@ -626,16 +627,39 @@ export class WordModule {
...
+           * while 'random' will arbitrarily choose words to have their first letters capitalized
+           *
+           * @default none
+           */
+          capitalize?: 'random' | 'all' | 'none';
         } = {}
   ): string {
     if (typeof options === 'number') {
-      options = { count: options };
+      options = { count: options, capitalize: 'none' };
```
ST-DDT:
```suggestion
      options = { count: options };
```

src/modules/word/index.ts
```diff
@@ -626,16 +627,39 @@ export class WordModule {
...
     }
 
-    const { count = { min: 1, max: 3 } } = options;
+    const { count = { min: 1, max: 3 }, capitalize } = options;
```
ST-DDT:
```suggestion
    const { count = { min: 1, max: 3 }, capitalize = 'none' } = options;
```

src/modules/word/index.ts
```diff
@@ -596,6 +596,7 @@ export class WordModule {
    *
    * @param options The optional options object or the number of words to return.
    * @param options.count The number of words to return. Defaults to a random value between `1` and `3`.
+   * @param options.capitalize Capitalize the first letter of each word. Defaults to `none`
```
ST-DDT:
```suggestion
   * @param options.capitalize Capitalize the first letter of each word. Defaults to `'none'`.
```

We should explain which options exists (for the non TS users) and what each option does.

ZweZeya: sure, I will add a brief description in the JSDoc and change the code bits according to your suggested changes

src/modules/word/index.ts
```diff
@@ -626,16 +629,39 @@ export class WordModule {
...
+    const { count = { min: 1, max: 3 }, capitalize = 'none' } = options;
 
     return this.faker.helpers
-      .multiple(() => this.sample(), { count })
+      .multiple(
+        () => {
+          let word = this.sample();
+          if (
+            capitalize === 'all' ||
+            (capitalize === 'random' && this.faker.datatype.boolean)
```
matthewmayer: shouldnt this be `this.faker.datatype.boolean()`, otherwise youre just checking for the existence of the method, not actually calling it

src/modules/word/index.ts
```diff
@@ -626,16 +629,39 @@ export class WordModule {
...
+          ) {
+            word = word.charAt(0).toUpperCase() + word.slice(1);
```
matthewmayer: it might be better to use `.toLocaleUpperCase(this.faker.definitions.metadata.code)` to handle languages which have different capitalization rules like Turkish: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase

test/__snapshots__/word.spec.ts.snap
```diff
@@ -106,6 +110,8 @@ exports[`word > 42 > words > with options.count 1`] = `"eek loudly alibi abnorma
 
 exports[`word > 42 > words > with options.count range 1`] = `"nondisclosure stud across dreamily accurate circumvent boo news loftily through devoice about when boohoo graze busily to grouchy beneath"`;
 
+exports[`word > 42 > words > with options.count range and capitalize = random 1`] = `"Nondisclosure Stud Across Dreamily Accurate Circumvent Boo News Loftily Through Devoice About When Boohoo Graze Busily To Grouchy Beneath"`;
```
matthewmayer: this snapshot test seems wrong, probably due to the incorrect boolean check above.

matthewmayer (changes requested):

src/modules/word/index.ts
```diff
@@ -626,16 +629,39 @@ export class WordModule {
                  */
                 max: number;
               };
+          /**
+           * Capitalize the first letter of each word.
+           *
+           * The option `'all'` will result in every word to have its first letter capitalized,
+           *
+           * while `'random'` will arbitrarily choose words to have their first letters capitalized
+           *
```
matthewmayer: we should clarify what `none` actually does. For example in some locales, e.g. `de`, certain words are already capitalized (since German always capitalizes nouns)

> fakerDE.word.words()
`'stapfen Elektrizität yowza'`

So in that case would you expect "none" to make everything lowercase? Or just leave it alone.

matthewmayer (changes requested):

src/modules/word/index.ts
ZweZeya: I see..okay, I will add a description explaining that 'none' will leave it alone.

PR #2176, opened 2023-05-24
docs(number): update default max value

### LINKED ISSUE #2175: Documentation for number.int default max value is not correct
### Pre-Checks

### Describe the bug

We are migrating from Faker 7 to Faker 8 and noticed deprecation warning for `datatype.number` to `number.int`
When applying the change to `number.int`, the [number generated is much larger than `min + 99999` as specified in the doc](https://fakerjs.dev/api/number.html#int) which cause our db test to break

Seems like it relates to https://github.com/faker-js/faker/issues/1598 but the doc was not updated

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) x64 Apple M1 Pro
    Memory: 16.33 MB / 16.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 14.17.6 - ~/Library/Caches/fnm_multishells/1601_1684935268559/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 6.14.15 - ~/Library/Caches/fnm_multishells/1601_1684935268559/bin/npm
  Browsers:
    Brave Browser: 113.1.51.114
    Firefox Developer Edition: 113.0
    Safari: 16.4
  npmPackages:
    @faker-js/faker: ^8.0.1 => 8.0.1
```

- [X] CJS
- [ ] ESM

### Used Package Manager

yarn

dephiros:
Address https://github.com/faker-js/faker/issues/2175
- Fix the default max
- Add a note in the migration guide


Shinigami92:
Oh, I switched the target branch from v8 to next, because this is out "main"/default branch
next will get merged into v8 when the release is made

Shinigami92 (dismissed):

src/modules/number/index.ts
matthewmayer, line 1: Should review all of the examples for example this one is unrealistic because it still follows the old min+99999 rule

faker.number.int({ min: 1000000}) // 1031433

matthewmayer (commented):
.

src/modules/number/index.ts
dephiros: @matthewmayer, `1031433` should be a potential valid value in this case still right?
Do you mean using a larger value like: `2900970162509863` to signal to people that the value could be a lot larger?

matthewmayer: Yes

dephiros: @matthewmayer I pushed a commit to adjust the example value when no max is specified to look more realistic

ST-DDT (approved):

Shinigami92 (approved):

matthewmayer (approved):

PR #2177, opened 2023-05-24
test(docs): ensure defaults are consistent

ST-DDT:
Created after seeing https://github.com/faker-js/faker/issues/2175.
Unfortunately there is currently no way to automatically detect issues where the jsdocs on the signature level differ from those on the options object because typedoc erases the signature ones if the option's one is present. But I found a way to at least check that the options one is present. And that the not-option ones match the js impl default one.

(it also can only check the last api signature)


scripts/apidoc/parameterDefaults.ts
```diff
@@ -54,7 +54,7 @@ function cleanParameterDefault(value?: string): string | undefined {
   }
 
   // Strip type casts: "'foobar' as unknown as T" => "'foobar'"
-  return value.replace(/ as unknown as [A-Za-z<>]+/, '');
+  return value.replace(/( as unknown)? as [A-Za-z<>]+/, '');
```
ST-DDT: For the unique method's args parameter: https://fakerjs.dev/api/helpers.html#unique

src/modules/datatype/index.ts
```diff
@@ -28,7 +28,7 @@ export class DatatypeModule {
    * Returns a single random number between zero and the given max value or the given range with the specified precision.
    * The bounds are inclusive.
    *
-   * @param options Maximum value or options object.
+   * @param options Maximum value or options object. Defaults to `99999`.
```
ST-DDT, line 18: This is inconsistent with...

src/modules/datatype/index.ts
```diff
@@ -298,7 +298,7 @@ export class DatatypeModule {
    * If the probability is `>= 1.0`, it will always return `true`.
    * The probability is limited to two decimal places.
    *
-   * @param options The optional options object or the probability (`[0.00, 1.00]`) of returning `true`. Defaults to `0.5`.
+   * @param options The optional options object or the probability (`[0.00, 1.00]`) of returning `true`.
```
ST-DDT, line 288: ...this line, due the method implementation's default value `{}` vs `value`.

src/modules/random/index.ts
```diff
@@ -230,7 +230,7 @@ export class RandomModule {
   /**
    * Generating a string consisting of letters in the English alphabet.
    *
-   * @param options Either the number of characters or an options instance. Defaults to `{ count: 1, casing: 'mixed', bannedChars: [] }`.
+   * @param options Either the number of characters or an options instance.
```
ST-DDT, line 218: Removed for improved docs readability. The value will be ignored by the docs generation anyway.

src/modules/random/index.ts
```diff
@@ -291,7 +291,7 @@ export class RandomModule {
    * Generating a string consisting of alpha characters and digits.
    *
    * @param count The number of characters and digits to generate. Defaults to `1`.
-   * @param options The options to use. Defaults to `{ bannedChars: [] }`.
+   * @param options The options to use.
```
ST-DDT, line 279: Same here

src/modules/random/index.ts
xDivisionByZerox: This should really be:

```suggestion
   * @param options Either the number of characters or an options instance. Defaults to `{}`.
```

To be consistent with other places in our codebase. Examples:
- https://github.com/faker-js/faker/blob/ee12c32fced6cdf6fef2c5993af84825fd88606f/src/modules/airline/index.ts#L141
- https://github.com/faker-js/faker/blob/ee12c32fced6cdf6fef2c5993af84825fd88606f/src/modules/commerce/index.ts#L58
- https://github.com/faker-js/faker/blob/ee12c32fced6cdf6fef2c5993af84825fd88606f/src/modules/finance/index.ts#L410
- https://github.com/faker-js/faker/blob/ee12c32fced6cdf6fef2c5993af84825fd88606f/src/modules/internet/index.ts#L72

This includes all other places as well.

ST-DDT:
Should we use ``Defaults to `{}`.`` everywhere or nowhere? I implemented this for ~10/50 locations.

src/modules/random/index.ts
Shinigami92: > This should really be [...]

Yeah, it seems so :thinking: 

@ST-DDT do you want to automate this? Or is something hard on this?
I would give you an approval for this PR, but it would get removed when you update the PR.
So let me know for what you have decided.

ST-DDT:
**Team decision**

We will omit/remove the empty options default for now in the jsdocs.
We will ensure all options parameters have a JS default value such as `{}` to ensure proper api docs defaults. (e.g. the color module doesn't have them yet) (separate PR: https://github.com/faker-js/faker/issues/2432)

ST-DDT:
Ready to review.

scripts/apidoc/typedoc.ts
```diff
@@ -246,6 +246,49 @@ export function extractRawDefault(reflection?: CommentHolder): string {
...
 
+/**
+ * Extracts and optionally removes the default from the comment summary.
+ *
+ * @param comment The comment to extract the default from.
+ * @param eraseDefault Whether to erase the default text from the comment.
+ *
+ * @returns The extracted default value.
+ */
+export function extractSummaryDefault(
```
xDivisionByZerox, line 257: Why did you pull this out of the `*/signature.ts` file? Is this function not handling signatures?

src/modules/location/index.ts
```diff
@@ -1140,8 +1153,21 @@ export class LocationModule {
     options?:
       | [latitude: number, longitude: number]
       | {
+          /**
+           * The original coordinate to get a new coordinate close to.
+           */
           origin?: [latitude: number, longitude: number];
```
xDivisionByZerox, line 1154: Shouldn't this be `[faker.location.latitude(), faker.location.longitude()]` instead?

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
```diff
@@ -93,6 +96,54 @@ describe('verify JSDoc tags', () => {
...
+  function assertNestedParameterDefault(
+    name: string,
+    parameterType?: SomeType
+  ): void {
+    if (!parameterType) {
+      return;
+    }
+
+    switch (parameterType.type) {
+      case 'array':
```
xDivisionByZerox, line 105: Does a touple match the `'array'` case?

src/modules/location/index.ts
ST-DDT: This the type definition not the default definition.
The origin parameter doesnt have a default.
Or did I missunderstood you?

xDivisionByZerox: The origin does indeed habe a default value:

https://github.com/faker-js/faker/blob/5410239245b4a6fe8c1976f8aa33c970923f9f40/src/modules/location/index.ts#L1179-L1182

Thats why I was asking why the default was no provided in the summary like this:

```suggestion
          /**
          * The original coordinate to get a new coordinate close to.
          *
          * @default `[faker.location.latitude(), faker.location.longitude()]`
          */
          origin?: [latitude: number, longitude: number];
```

ST-DDT: > The origin does indeed habe a default value:

Close, but no. The option doesn't have a default. 
If the option is missing a different code path is chosen e.g. returning a random location instead of a location close to that origin location.

test/scripts/apidoc/verify-jsdoc-tags.spec.ts
ST-DDT: No, it does not. We currently don't have a tuple type with either a default or any special docs handling.

src/modules/location/index.ts
xDivisionByZerox: I'm stupid, I see. I thought the `return [...]` was assigning the value like `origin = [...]`. Sorry for the confusion.

scripts/apidoc/typedoc.ts
ST-DDT: I moved it here because all the other exported `extract*(typedocThing) -> *` methods are here as well.
Especially the `extractRawDefault` method that is used in conjuction with this one.
The signature.ts file only exports one functions so far, but I'm not really sure where to put this method (especially since it is kind of destructive).
I also considered re-writing the related code to keep the data immutable (separate PR).

xDivisionByZerox: Ok I understand. Not that it really mattered, but I wanted to understand your reasoning. Thank you.

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

PR #2181, opened 2023-05-28
test(docs): disable versions check for external contributors

ST-DDT:
Fixes https://github.com/faker-js/faker/pull/2174#issuecomment-1562585163

This disables the versions check for external contributors/those that do not use the faker main repository as their origin.


ST-DDT:
Result if origin != faker main repo:

grafik

Result if origin = faker main repo:

grafik

test/docs/versions.spec.ts
```diff
@@ -1,8 +1,25 @@
...
 import { oldVersions } from '../../docs/.vitepress/versions';
 
-describe('docs versions', () => {
+function isFakerOrigin(): boolean {
+  try {
+    const originUrl = execSync('git remote get-url origin')
+      .toString('utf8')
+      .trim();
+    return (
+      originUrl === 'git@github.com:faker-js/faker' ||
```
Shinigami92, line 12: Do you think using `originUrl.includes('faker-js/faker')` is sufficient enough?

Shinigami92 (commented):
Will the CI still check the docs version, even when the PR is on another fork?

ST-DDT:
> Will the CI still check the docs version, even when the PR is on another fork?

Good point. I have to check.

test/docs/versions.spec.ts
xDivisionByZerox: I don't think it is as you can technically name your fork however you like. So some name that would match your suggestion would simply be `faker-js/faker-fork`. Correct me if I'm wrong tho.

Shinigami92: Yeah, in GitLab you could have that

Shinigami92 (changes requested):
I could be wrong, but it seems that this PR disables the test when a contributor creates a PR in their own fork. But when we then merge that PR, the test would run on our `next` branch and could then fail.

So this doesn't look like it is the right solution, but sadly I also do not have another idea right now :confused:

Shinigami92:
Is the test in our CI always executed via a PR and therefore the origin will always be `faker-js/faker`?
Or does the CI origin run on the fork?

ST-DDT:
I created #2394 as an external contributor test.

- #2394

Shinigami92 (approved):
Yes, the test gets executed: https://github.com/faker-js/faker/actions/runs/6201950881/job/16839680435?pr=2394#step:8:110

ST-DDT:
From #2394 :

https://github.com/faker-js/faker/actions/runs/6201950881/job/16839679994?pr=2394#step:2:39

> git remote add origin https://github.com/faker-js/faker

https://github.com/faker-js/faker/actions/runs/6201950881/job/16839679994?pr=2394#step:8:108

> stdout | unknown test
> https://github.com/faker-js/faker
> 
> ✓ test/docs/versions.spec.ts  (1 test) 6ms

It uses the remote of the target repository as origin.

xDivisionByZerox (approved):

PR #2188, opened 2023-05-30
feat(number): support step on faker.number.int()

### LINKED ISSUE #2186: Allow passing precision to faker.number.int

In faker 7 I could specify
```
faker.datatype.number({ min: 30, max: 200, precision: 10 })
```
to get a number between 30 and 200, with an increment of 10, so: 140, 90, 180, ...

But with faker 8, the precision was removed from int.

```
faker.number.int({ min: 30, max: 200 })
```

### Suggested solution

Allow passing a precision to the int faker function

```
faker.number.int({ min: 30, max: 200, precision: 10 })
```

### Alternative

_No response_

### Additional context

_No response_

leon:
Fixes #2186

In faker 7 I could specify
```
faker.datatype.number({ min: 30, max: 200, precision: 10 }) // 140, 90, 180, ...
faker.datatype.number({ min: 5000, max: 20000, precision: 100 }) // 5200, 18500, 12900,...
```

But with faker 8, the precision was removed from int.

```
faker.number.int({ min: 30, max: 200 })
```

### Suggested solution

Allow passing a step to the int faker function

```
faker.number.int({ min: 30, max: 200, step: 10 })
```

### DIFF
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -67,13 +67,19 @@ export class NumberModule {
            * @default Number.MAX_SAFE_INTEGER
            */
           max?: number;
+          /**
+           * Increment value for generated number.
+           *
+           * @default 1
+           */
+          step?: number;
         } = {}
   ): number {
     if (typeof options === 'number') {
       options = { max: options };
     }
 
-    const { min = 0, max = Number.MAX_SAFE_INTEGER } = options;
+    const { min = 0, max = Number.MAX_SAFE_INTEGER, step = 1 } = options;
     const effectiveMin = Math.ceil(min);
     const effectiveMax = Math.floor(max);
 
@@ -91,11 +97,17 @@ export class NumberModule {
       throw new FakerError(`Max ${max} should be greater than min ${min}.`);
     }
 
+    if (step <= 0) {
+      throw new FakerError(`Step should be a positive number.`);
+    }
+
     const mersenne: Mersenne =
       // @ts-expect-error: access private member field
       this.faker._mersenne;
     const real = mersenne.next();
-    return Math.floor(real * (effectiveMax + 1 - effectiveMin) + effectiveMin);
+    const range = (effectiveMax - effectiveMin) / step;
+    const rand = Math.floor(real * (range + 1));
+    return effectiveMin + rand * step;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/modules/number/index.ts
```diff
@@ -91,11 +97,17 @@ export class NumberModule {
       throw new FakerError(`Max ${max} should be greater than min ${min}.`);
     }
 
+    if (step <= 0) {
+      throw new FakerError(`Step should be a positive number.`);
```
matthewmayer, line 101: Should we also throw on a noninteger step?

src/modules/number/index.ts
```diff
@@ -91,11 +97,17 @@ export class NumberModule {
...
+    }
+
     const mersenne: Mersenne =
       // @ts-expect-error: access private member field
       this.faker._mersenne;
     const real = mersenne.next();
-    return Math.floor(real * (effectiveMax + 1 - effectiveMin) + effectiveMin);
+    const range = (effectiveMax - effectiveMin) / step;
+    const rand = Math.floor(real * (range + 1));
+    return effectiveMin + rand * step;
```
matthewmayer, line 110: This does not work correctly.

For example `faker.number.int({min:0, max:5, step: 2})` sometimes returns 6.

```
range is (5 - 0)/2 = 2.5
rand is say floor(0.99 * 3.5) = 3
returns 0 + 3 * 2 = 6
```

matthewmayer (changes requested):

src/modules/number/index.ts
matthewmayer: Also per https://github.com/faker-js/faker/pull/1855#issuecomment-1443404570

`faker.number.int({min:5,  max: 100, step: 10})` should return {10, 20, ...} not {5, 15, 25...}

Shinigami92:
Marked this for `needs decision` so we have time to discuss in a team meeting how the best way is to solve this and if the current ideas are the "right" way of doing it
The next (full/relevant) team meeting wont be before July 2023

test/number.spec.ts
```diff
@@ -163,6 +163,22 @@ describe('number', () => {
           new FakerError(`No integer value between 2.1 and 2.9 found.`)
         );
       });
+
```
Shinigami92, line 166: line 9 should be extracted and separated together with a new `it` line

ST-DDT:
Closed in favor of #2586

PR #2190, opened 2023-05-31
fix(locale): limited ja first names without passing sex

KeisukeYamashita:
Hello, I have noticed that `ja` person first name only has `21` entries while `female_first_name` has more than `100` and `male_first_name` as well.

Even though hard work has done in each sex is defined, if nothing is specified, only a limited number of options will be returned. So I have merged the two arrays as `first_name` and now the `first_name()` function can return from almost `279` candidates (`258` increase 🚀 ).

Moreover, this avoids duplicate management of each list separately, making them easier to manage.

## Reference

This is exactly the same way as some other languages, for example: 

・`fr`

https://github.com/faker-js/faker/blob/27a8ff7113895c5e5ee52235db364e0d87fe8a0e/src/locales/fa/person/prefix.ts#L1-L5

・`ne`

https://github.com/faker-js/faker/blob/27a8ff7113895c5e5ee52235db364e0d87fe8a0e/src/locales/ne/person/first_name.ts#L4-L8


matthewmayer (approved):
LGTM. Having most locales using the mergeArray function will simplify things when we restructure the gendered definitions.

ejcheng (approved):

xDivisionByZerox:
@KeisukeYamashita Just to be sure and for future reference: Does the concept of gender-neutral first names not exist in Japanese or where the names in the `first_name.ts` file simply not gender-neutral?

ST-DDT (approved):

KeisukeYamashita:
@xDivisionByZerox Thank you for your comment!

In Japan, AFAK、there are some gender neutral names like アオイ、レイ、etc. It's can be used for male and female names.

Perhaps I was mistaken, but I thought the firstName function returned either male or female names (including some gender-neutral names), not only gender-neutral names.

I thought the situation was similar in other countries, so I used mergeArray as a reference.

PR #2191, opened 2023-05-31
feat(locale): add direction to ja

KeisukeYamashita:
Hello, I was trying to use Faker to generate dummy directions but I found the `ja` local is missing.
Therefore, I have added it like

https://github.com/faker-js/faker/blob/27a8ff7113895c5e5ee52235db364e0d87fe8a0e/src/locales/en/location/direction.ts#L1-L10

Thank you in advance.


ejcheng (approved):

xDivisionByZerox (approved):
Thank you for your contribution!

ST-DDT (approved):

PR #2193, opened 2023-05-31
feat(locale): support street and street address for ja

KeisukeYamashita:
Fixed (added) Japanese address output. 

Previously, English addresses were generated incorrectly, but the format has been changed to the correct address format.
Street address example:

```javascript
import { fakerJA as faker } from "@faker-js/faker";

// Before
faker.location.streetAddress() // 5668 杏Square. Incorrect
faker.location.streetAddress(true) // 5668 杏Square. Incorrect

// After
faker.location.streetAddress() // 8丁目7番4号. Correct
faker.location.streetAddress(true) // 9丁目2番3号 木村768号室. Correct
```

Symbols can now be used in street addresses because street names in Japan sometimes contain numbers, but this has no other destructive effect. Of course, not all country road formats include numbers, but I believe they are used in many countries. I supported it because the use of symbols is completely restricted now.

Thank you in advance.

## Disclaimer

I am not a government official or lawyer in charge of Japanese geography. However, in its current state, it is difficult to use it for address dummy data, so a PR was created.


KeisukeYamashita:
Sorry, the CI failed.
Let me check 🙏

ST-DDT:
Looks like you forgot to run `pnpm run preflight` 😉 .

src/locales/ja/location/street_pattern.ts
```diff
@@ -1,4 +1 @@
-export default [
-  '{{person.first_name}}{{location.street_suffix}}',
-  '{{person.last_name}}{{location.street_suffix}}',
-];
+export default ['#丁目#番#号'];
```
matthewmayer: while it is a little ugly, you could use a pattern like:
```
'{{number.int({"max": 10, "min": 1})}}丁目{{number.int({"max": 10, "min": 1})}}番{{number.int({"max": 10, "min": 1})}}号'
```

This can be used directly by `faker.helpers.fake`, so it's not necessary to change the street() method to use replaceSymbols.

matthewmayer (changes requested):

matthewmayer:
https://en.wikipedia.org/wiki/Japanese_addressing_system

src/locales/ja/location/street_pattern.ts
KeisukeYamashita: Thank you very much for your kind suggestion, I will update it tomorrow 🙏

KeisukeYamashita: @matthewmayer  Thank you very much for your advice, I have updated by this commit 👉 https://github.com/faker-js/faker/pull/2193/commits/18ecd9d04bed1ec326ee520845daba7c5f5c1648

Since I don't have to add `replaceSymbols` to street APIs anymore, I have reverted in this commit 👉 https://github.com/faker-js/faker/pull/2193/commits/f2ca178ef7625ae79d3e2bbe0b71da5a175818c6

KeisukeYamashita:
> Looks like you forgot to run pnpm run preflight 😉 .

Thank you very much for your comment, I have ran and push  this commit 👉 https://github.com/faker-js/faker/pull/2193/commits/a83f75ac4f942ae362568294c4eb9f69ecd62551

src/locales/ja/location/street_pattern.ts
```diff
@@ -1,4 +1,3 @@
 export default [
-  '{{person.first_name}}{{location.street_suffix}}',
-  '{{person.last_name}}{{location.street_suffix}}',
+  '{{number.int({"max": 10, "min": 1})}}丁目{{number.int({"max": 10, "min": 1})}}番{{number.int({"max": 10, "min": 1})}}号',
```
matthewmayer: Is 1...10 the most sensible range for each of these? What numbers do the 丁目, 番 and 号 numbers typically go up to?

KeisukeYamashita: It is determined by the local government and can increase or decrease depending on the location, but is almost always in the around single digits.

As far as I know, there are no clear rules; it is determined by the history of the location and its management laws.


A city located in a region called Hokkaido, there are more than 40丁目, the largest number in Japan, but it is quite special, and a single digit is a natural address.
This is same to "番地" and "号".

It is usually one digit and I have never seen a double digit street in my life (although it exists).
So, for example, if dummy data is generated and the address has something like 20丁目50番30号, Japanese people will feel a little uncomfortable, although it is not a mistake.

matthewmayer: maybe make the max 9 then?

KeisukeYamashita: True, I made an update here 👉 https://github.com/faker-js/faker/pull/2193/commits/93995ded03588e9fbd09282bfb2afdde01b34985
Thank you for your suggestion 🙏

matthewmayer (dismissed):

src/locales/ja/location/secondary_address.ts
```diff
@@ -0,0 +1 @@
+export default ['{{person.last_name}}###号室'];
```
ST-DDT, line 1: Here are still the `#` from the previous `replaceSymbol` implementation.

ST-DDT (dismissed):

src/locales/ja/location/secondary_address.ts
ST-DDT, line 1: This does not work:

````ts
import { fakerJA } from "@faker-js/faker";

console.log(fakerJA.location.secondaryAddress()); // '{{person.last_name}}522号室'
````

**Review by ST-DDT — CHANGES_REQUESTED** [review]

File: src/locales/ja/location/secondary_address.ts (hunk 3 above)
Comment (KeisukeYamashita) [reply src/locales/ja/location/secondary_address.ts:1]: Oh, let me fix and test it locally 🙏 Thank you for catching!

Comment (KeisukeYamashita) [reply src/locales/ja/location/secondary_address.ts:1]: @ST-DDT  Thank you very much for catching it 🙏 

As you mentioned, when the `secondaryAddress` method is called directly instead of via `street` (or any other methods that rely on secondary address), the Mustache syntaxes are not parsed.

I have updated to parse Mustache syntaxes on `secondaryAddress` method calls 👉 https://github.com/faker-js/faker/pull/2193/commits/3d5a846779c95ddd72c98b896ece1142a6c1eaca

```typescript
import {fakerJA as faker} from '../src'

console.log(faker.location.secondaryAddress()) // 佐藤167号室
```

Thank you in advance.

src/modules/location/index.ts
```diff
@@ -284,14 +284,16 @@ export class LocationModule {
...
+    const format = this.faker.helpers
       .arrayElement(this.faker.definitions.location.secondary_address)
       .replace(/#+/g, (m) =>
         this.faker.string.numeric({
           length: m.length,
           allowLeadingZeros: false,
         })
       );
+
+    return this.faker.helpers.fake(format);
```
ST-DDT: IMO you can directly pass the template to the fake function.

```suggestion
    return this.faker.helpers
      .fake(this.faker.definitions.location.secondary_address)
      .replace(/#+/g, (m) =>
        this.faker.string.numeric({
          length: m.length,
          allowLeadingZeros: false,
        })
      );
```

KeisukeYamashita: Thank you!

src/locales/ja/location/street_pattern.ts
xDivisionByZerox: @matthewmayer how is your solution different from the `#` character? I saw that you used `10` as max value but now it is back to 9 which means it should be the same as `#`. Am I having an error in my thought process?

matthewmayer: Because it doesn't require any changes to the street method. It's only a ja definitions change.

ST-DDT:
Sorry for the long wait. This somehow got lost in the PR backlog.
I think we have to discuss internally if the fake pattern usage here is what we want.

ST-DDT:
**Team Decision**

- We accept this change. Please rebase so that we can merge this.

KeisukeYamashita:
Thank you for your consideration!
I've fixed the conflict and updated the main branch. Thanks in advance.

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2194, opened 2023-06-02
fix(locale): remove continent from ja countries

KeisukeYamashita:
Hello, sorry for the bothering 🙏 

I noticed that there are many continents in the country list.

I would like to remove it because the dummy data for user profiles in the service, as it generates strange profiles, such as Antarctica as the country of residence.

Thank you in advance.


matthewmayer (approved):

ejcheng (approved):

ejcheng:
Translated the removed continents to English to check, lgtm 👍

ST-DDT (approved):

PR #2211, opened 2023-06-14
add browser and os as parameters

adrian-moloca:


matthewmayer:
we discussed before and there seemed to be consensus that the current userAgent implementation is excessively complex and hard to maintain - see https://github.com/faker-js/faker/pull/1761#issuecomment-1406431993 - i think this simplification should probably be prioritised over adding more parameters

matthewmayer:
btw it looks like you accidentally remove `/dist` from .gitignore which has caused a large number of files to be added to this PR

.gitignore
```diff
@@ -74,7 +74,6 @@ node_modules/
 versions.json
 
 # Dist
-/dist
```
ST-DDT, line 77: Please revert removing dist from the gitignore.

ST-DDT (changes requested):

ST-DDT:
**Team decision**

The PR looks broken and not updated in a long time.
If you are still interested please create a new feature request issue first.

PR #2222, opened 2023-06-23
refactor(git): don't use Intl

### LINKED ISSUE #2221: ReferenceError: Intl is not defined
### Pre-Checks

### Describe the bug

Hello, I'm trying to use Faker extension and when I run the test, I get the error message on the console:
````txt
ReferenceError: Intl is not defined
        at ./node_modules/@faker-js/faker/dist/esm/chunk-2JBE4HP3.mjs (webpack://***/./node_modules/@faker-js/faker/dist/esm/chunk-2JBE4HP3.mjs:4:25475(2732))
        at call (native)
        at __webpack_require__ (webpack://***/webpack/bootstrap:19:0(33))
        at ./node_modules/@faker-js/faker/dist/esm/chunk-BFO2MQ2Q.mjs (webpack://***/./node_modules/@faker-js/faker/dist/esm/chunk-BAJCEGMK.mjs:1:43764(24))
        at call (native)
        at __webpack_require__ (webpack://***/webpack/bootstrap:19:0(33))
        at ./node_modules/@faker-js/faker/dist/esm/index.mjs (webpack://***/./node_modules/@faker-js/faker/dist/esm/chunk-ZVFD5DJO.mjs:1:7399(291))
        at call (native)
        at __webpack_require__ (webpack://***/webpack/bootstrap:19:0(33))
        at ./src/actions/requests/requests.Employee.ts (webpack://***/./src/actions/index.ts:11:69(67))
        at call (native)
        at __webpack_require__ (webpack://***/webpack/bootstrap:19:0(33))
        at webpack://***/webpack/runtime/node%20module%20decorator:5:0(36)
        at file://***/dist/***.bundle.js:23809:3(477)
````

### Minimal reproduction code

_No response_

### Additional Context

package.json devDependencies:
````json
  "devDependencies": {
    "@babel/cli": "^7.21.0",
    "@babel/core": "^7.21.4",
    "@babel/node": "^7.20.7",
    "@babel/plugin-proposal-class-properties": "^7.18.6",
    "@babel/plugin-proposal-object-rest-spread": "^7.20.7",
    "@babel/preset-env": "^7.21.4",
    "@babel/preset-typescript": "^7.21.4",
    "@faker-js/faker": "^8.0.2",
    "@types/mocha": "^10.0.1",
    "@typescript-eslint/eslint-plugin": "^5.57.1",
    "babel-eslint": "^10.1.0",
    "babel-loader": "^9.1.2",
    "eslint": "^8.38.0",
    "eslint-config-prettier": "^8.8.0",
    "eslint-config-standard-with-typescript": "^34.0.1",
    "eslint-plugin-import": "^2.27.5",
    "eslint-plugin-n": "^15.7.0",
    "eslint-plugin-prettier": "^4.2.1",
    "eslint-plugin-promise": "^6.1.1",
    "prettier": "2.8.7",
    "typescript": "^5.0.4",
    "webpack": "^5.78.0",
    "webpack-cli": "^5.0.1"
  }
````

import: 
`import { faker } from '@faker-js/faker';`

### Environment Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 2.31 GB / 15.73 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (114.0.1823.51)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @faker-js/faker: ^8.0.2 => 8.0.2
```

- [ ] CJS
- [ ] ESM

### Used Package Manager

npm

matthewmayer:
fix #2221 ref #2171

Removes the use of Intl in faker.git.commitDate


src/modules/git/index.ts
```diff
@@ -185,28 +157,35 @@ export class GitModule {
...
-
-    // Timezone offset
-    dateParts.push(
-      GIT_TIMEZONE_FORMAT.format(
-        this.faker.number.int({ min: -11, max: 12 }) * 100
-      )
-    );
-
-    return dateParts.join(' ');
+    const days = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
```
ST-DDT, line 163: Maybe move them to the top of the file as a constant and please also add a comment that these are always the English names and not just some missed i18n part.

xDivisionByZerox, line 163: I know that the previous version was English only as well. So, just the general question: Should we look into localizing this output in the future?

xDivisionByZerox (dismissed):

src/modules/git/index.ts
matthewmayer: What does git do?

ST-DDT (commented):
Oh, it looks like I forgot to post these.

src/modules/git/index.ts
ST-DDT: See https://github.com/faker-js/faker/pull/2222#discussion_r1239729853

matthewmayer: Seems git doesn't localize the dates with the default date option, see

https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-log.html
```
--date=default is the default format, and is based on ctime(3) output. It shows a single line with three-letter day of the week, three-letter month, day-of-month, hour-minute-seconds in "HH:MM:SS" format, followed by 4-digit year, plus timezone information, unless the local time zone is used, e.g. Thu Jan 1 00:00:00 1970 +0000.
```

test/git.spec.ts
```diff
@@ -114,22 +114,6 @@ describe('git', () => {
 
           expect(commitEntry).not.contains('\r\n');
         });
-
-        it('should throw if Intl is unavailable', () => {
```
Shinigami92, line 118: Please add a "negative"-test case, where we check for `it should not throw or fail in an environment that does not support Intl, like react-native`
And then `mock/spy` or `delete` Intl inside that test

That way we ensure that no other contributor comes alongside and re-implement it
as well as "we officially support non-Intl environments"

Shinigami92 (changes requested):

test/git.spec.ts
matthewmayer: I feel like this might be better as a lint rule (check there are no usages of Intl) than a test?

src/modules/git/index.ts
matthewmayer: I will add a comment, but I think the arrays should probably stay inside this function? These are only used for this specific function to handle git's arcane default date format, we don't want to encourage their use elsewhere.

test/git.spec.ts
Shinigami92: Would be an acceptable option to me :+1: 

Another idea would be to do it not for this method, but in a global vitest scope, so the whole test-suite does not have access to `Intl` (maybe achievable via global setup)

src/modules/git/index.ts
matthewmayer: added comment

test/git.spec.ts
matthewmayer: added eslint rule

Shinigami92 (approved):

ST-DDT (approved):

PR #2227, opened 2023-06-27
feat(lorem): add character length parameter to lorem.text

### LINKED ISSUE #1546: Add character length parameter to `lorem.text`

Sometimes we just want a text that has a particular length instead of conforming to arbitrary length units (lines/paragraphs).

### Suggested solution

Add length parameter to the `lorem.text` method.

`faker.lorem.text({ length: number | { min: number, max: number }})`

`faker.lorem.text({ length: 150 })`

> Nostrum nostrum aperiam vero eveniet quasi. Neque libero consequuntur molestias iste nihil corrupti ducimus suscipit. Ratione ducimus occaecati offic.

### Alternative

`faker.lorem.paragraphs(<sufficientlyLargeNumber>).substring(0, 150)`

### Additional context

Originally requested in [Discord](https://discord.com/channels/929487054990110771/929487054990110777/1040008594550161528) by kryp71c.

ghost:
Adds a character length parameter to `faker.lorem.text` as described in #1546.

Changes made:

- ~~Added a new parameter `length?: number | { min: number, max: number }` such that `lorem.text` returns text that has a particular length if the parameter is passed~~
- Added a new optional options object `{length: number | { min: number, max: number }}` such that `lorem.text` returns text that has a particular length if the parameter is passed
- Added a new optional quick primitive argument such that `lorem.text` returns text that the length passed as argument
- Added random seeded tests for the new parameter to `lorem.spec.ts` and updated its snapshot

Closes #1546


test/__snapshots__/lorem.spec.ts.snap
```diff
@@ -97,9 +97,9 @@ exports[`lorem > 42 > slug > with length range 1`] = `"minus-a-qui-tempore-quo-p
 
 exports[`lorem > 42 > text > noArgs 1`] = `"Qui tempore quo perspiciatis perspiciatis quia in quia architecto ipsa. Pariatur veniam unde fugit facilis mollitia. Ipsa reiciendis nam. Earum dolorem alias qui asperiores qui natus aliquam iste. Consequatur eum sit dolore rem iste aliquid."`;
 
-exports[`lorem > 42 > text > with length 1`] = `"Qui tempore quo perspiciatis perspiciatis quia in quia architecto ipsa. Pariatur veniam unde fugit facilis mollitia. Ipsa reiciendis nam. Earum dolorem alias qui asperiores qui natus aliquam iste. Consequatur eum sit dolore rem iste aliquid."`;
+exports[`lorem > 42 > text > with length 1`] = `" Qui temp."`;
```
ST-DDT: IMO the text should not start with a whitespace.

src/modules/lorem/index.ts
```diff
@@ -318,6 +318,10 @@ export class LoremModule {
   /**
    * Generates a random text based on a random lorem method.
    *
+   * @param length The length (range) of text to generate.
```
matthewmayer: Not sure we should introduce new single arguments like this, we are trending towards putting everything in options objects. What if we need another parameter later?

test/lorem.spec.ts
```diff
@@ -325,6 +325,85 @@ describe('lorem', () => {
...
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+          expect(actual.length).toBeGreaterThanOrEqual(25);
+          expect(actual.length).toBeLessThanOrEqual(40);
+        });
+
+        it('should return null for a negative length', () => {
+          const actual = faker.lorem.text(-3);
+
+          expect(actual).toBeNull();
```
matthewmayer: I feel this should throw as is probably programmer error.

test/lorem.spec.ts
```diff
@@ -325,6 +325,85 @@ describe('lorem', () => {
...
+
+        it('should return an empty string for [0, 1]', () => {
+          const actual = faker.lorem.text({ min: 0, max: 1 });
+
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toBe('');
+        });
+
+        it('should return an empty string for [0, 32]', () => {
+          const actual = faker.lorem.text({ min: 0, max: 32 });
```
matthewmayer: This doesn't seem that intuitive. I would expect to get string's with random lengths between 0 and 32

test/lorem.spec.ts
```diff
@@ -325,6 +325,85 @@ describe('lorem', () => {
...
+          const actual = faker.lorem.text({ min: 77, max: 77 });
+
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toHaveLength(77);
+        });
+
+        it('should return null for [3, 1]', () => {
+          const actual = faker.lorem.text({ min: 3, max: 1 });
+
```
matthewmayer: This should also throw

matthewmayer (changes requested):

test/lorem.spec.ts
ST-DDT: IIRC we return empty strings for negative size strings as well.

test/lorem.spec.ts
ghost: Thank you. I fixed this.

test/lorem.spec.ts
ghost: Thank you. I fixed this.

src/modules/lorem/index.ts
ghost: I removed the single argument and added an option object.

test/__snapshots__/lorem.spec.ts.snap
ghost: Thank you. I fixed this.

test/lorem.spec.ts
ghost: I am not sure which behaviour is better. In faker.string.alpha an empty string is returned, if I understand the code correctly. But I do not find this intuitive. Currently the method throws an error in this case. Is this fine?

xDivisionByZerox: Regardless of the decision, this should be discussed in a separate issue.

test/lorem.spec.ts
xDivisionByZerox: This should return an empty string to have a consistent behavior across other methods.

ghost: Now, it returns an empty string.

src/modules/lorem/index.ts
xDivisionByZerox: While we should have an options object for future extendability, we often allow the user to provide a "quick primitive argument" for the option that is most likely to be used.

Examples:
```ts
// equivalent
faker.finance.accountNumber(5);
faker.finance.accountNumber({ length: 5 });

// equivalent
faker.location.zipCode('###');
faker.location.zipCode({ format: '###' }});

// equivalent
faker.number.int(1337);
faker.number.int({ max: 1337 }});

// equivalent
faker.string.alphanumeric(1337);
faker.string.alphanumeric({ length: 1337 }});

// equivalent
faker.word.adverb(42);
faker.word.adverb({ length: 42 }});
```

ghost: Thank you for the examples. I added the quick primitive argument again.

test/lorem.spec.ts
matthewmayer: I feel {min:3, max:1} is almost certainly due to programmer error (Ie they accidentally reversed it) so it's better to throw than return an empty string.

src/modules/lorem/index.ts
```diff
@@ -326,10 +332,32 @@ export class LoremModule {
    * // Quis ut dolor dolores facilis possimus tempore voluptates.
    * // Iure nam officia optio cumque.
    * // Dolor tempora iusto.'
+   * faker.lorem.text(14) // 'Doloribus aut.' (14 characters)
+   * faker.lorem.text({length: 14}) // 'Doloribus aut.' (14 characters)
+   * faker.lorem.text({min: 10, max: 15}) // 'autem non quis.' (15 characters)
```
matthewmayer: I think {length:x} or {length:{min:x, max:y}} would be clearer. Then you avoid edge cases like what should happen if you pass {length:1, max:2}

matthewmayer (changes requested):

test/lorem.spec.ts
xDivisionByZerox: I checked. Methods in the StringModule do indeed throw if `max` is smaller than `min`. Internally, they use `helpers.rangeToNumber`, which uses `number.int` under the hood. `number.int` covers the throwing behavior. That's my bad.

CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ This is a shorthand for running the following scripts in order:
 
 - `pnpm install` - installs npm packages defined in package.json
 - `pnpm run generate:locales` - generates locale files
-- `pnpm run generate:api-doc` - generates API documentation
+- `pnpm run generate:api-docs` - generates API documentation
```
xDivisionByZerox: While I love that you did spot this mistake, please don't mix concerns of PR. This PR has the concern to implement a feature, not fixing a documentation typo.

test/lorem.spec.ts
ghost: Oh. Now, it throws again :)

src/modules/lorem/index.ts
ghost: Good point. I changed the code.

test/lorem.spec.ts
```diff
@@ -325,6 +325,125 @@ describe('lorem', () => {
...
+        });
+
+        it('should return an empty string for [-13, 0]', () => {
+          const actual = faker.lorem.text({ length: { min: -13, max: 0 } });
+
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toBe('');
+        });
+
+        it('should return an empty string for [0, 1]', () => {
```
matthewmayer: The test is right but the description is wrong, it doesn't always return empty string

test/lorem.spec.ts
```diff
@@ -325,6 +325,125 @@ describe('lorem', () => {
...
+
+        it('should return a string of length 77 for [77, 77]', () => {
+          const actual = faker.lorem.text({ length: { min: 77, max: 77 } });
+
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toHaveLength(77);
+        });
+
+        it('should return an empty string for [3, 1]', () => {
```
matthewmayer: Again desc is wrong

src/modules/lorem/index.ts
```diff
@@ -340,7 +371,54 @@ export class LoremModule {
...
+      typeof options.length?.min !== 'undefined' ? options.length.min : 0;
+    const max =
+      typeof options.length?.max !== 'undefined' ? options.length.max : 2 * min;
+    if (min > max) {
+      throw new FakerError(
+        `Max ${options.length.max} should be greater than min ${options.length.min}.`
+      );
+    }
+
+    const effectiveMin = Math.max(min, 0);
```
matthewmayer: I think it would simplify the implementation to just throw for a negative min too

matthewmayer (dismissed):

test/lorem.spec.ts
ghost: I corrected the description.

src/modules/lorem/index.ts
ghost: You're right. I also think the implementation is simpler now.

test/lorem.spec.ts
ghost: I corrected the description.

src/modules/lorem/index.ts
```diff
@@ -326,10 +332,35 @@ export class LoremModule {
    * // Quis ut dolor dolores facilis possimus tempore voluptates.
    * // Iure nam officia optio cumque.
    * // Dolor tempora iusto.'
+   * faker.lorem.text(14) // 'Doloribus aut.' (14 characters)
+   * faker.lorem.text({length: 14}) // 'Doloribus aut.' (14 characters)
+   * faker.lorem.text({length: {min: 10, max: 15}}) // 'autem non quis.' (15 characters)
+   * faker.lorem.text({length: {min: 10, max: 12}}) // 'autem non.' (10 characters)
+   * faker.lorem.text({length: {min: 3, max: 10}}) // 'autem.' (6 characters)
```
xDivisionByZerox: I dont think we need the explicit character information at the end. We also dont need that many repetitive examples:

```suggestion
   * faker.lorem.text(14) // 'Doloribus aut.'
   * faker.lorem.text({length: 14}) // 'Doloribus aut.'
   * faker.lorem.text({length: {min: 10, max: 15}}) // 'autem non quis.'
```

CONTRIBUTING.md
ST-DDT: You can create a different PR for that, that way the PRs stay on one topic and are easier to review and merge.

matthewmayer (commented):
I think the description of `text()` in the overview could also be modified to reflect this can be used to produce text of a certain length

```The generic [`text()`](https://fakerjs.dev/api/lorem.html#text) method can be used to generate some text between one sentence and multiple paragraphs```

**Comment by matthewmayer** [comment]
The behavior of the function when the length parameter is supplied is quite different to the version without.

I'm wondering if therefore it would make sense to just make it a separate method, like:

```
* faker.lorem.characters(14) // 'Doloribus aut.' (14 characters)
* faker.lorem.characters({length: 14}) // 'Doloribus aut.' (14 characters)
* faker.lorem.characters({length: {min: 10, max: 15}}) // 'autem non quis.' (15 characters)
* faker.lorem.characters({length: {min: 10, max: 12}}) // 'autem non.' (10 characters)
```

File: CONTRIBUTING.md (hunk 7 above)
Comment (ghost) [reply CONTRIBUTING.md]: Oh, sorry. I did not pay attention to this. I created a [different PR](https://github.com/faker-js/faker/pull/2264) for that.

**Comment by ST-DDT** [comment]
> I'm wondering if therefore it would make sense to just make it a separate method, like:
> 
> ```
> * faker.lorem.characters(14) // 'Doloribus aut.' (14 characters)
> ```

IMO the result aren't characters, but some kind of text/words. Maybe rename the `length` parameter to `characters` instead?

**Comment by matthewmayer** [comment]
its not purely a naming issue, the behavior of text() is a little surprising, because depending on which other faker method is called internally sometimes you get sentences separated by newlines, or one big paragraph etc.

image

**Comment by ST-DDT** [comment]
Maybe we should deprecate `text` in favor of `sentences` and `paragraphs`. We will discuss this tomorrow, if we have the time.
(and add the characters option to the other methods)

**Comment by ST-DDT** [comment]
There seems to be merge conflicts. Could you please fix them?

**Comment by ghost** [comment]
Sorry for the late reply. I fixed the merge conflicts.

File: src/modules/lorem/index.ts [hunk 12]
```diff
@@ -318,10 +324,33 @@ export class LoremModule {
...
+          /**
+           * The length (range) of text to generate.
+           */
+          length?:
+            | number
+            | {
+                /**
+                 * The minimum length of text to generate.
+                 */
+                min?: number;
```
ST-DDT:
```suggestion
                min: number;
```

src/modules/lorem/index.ts
```diff
@@ -318,10 +324,33 @@ export class LoremModule {
...
+                /**
+                 * The maximum length of text to generate.
+                 */
+                max?: number;
```
ST-DDT:
```suggestion
                max: number;
```

src/modules/lorem/index.ts
```diff
@@ -332,7 +361,57 @@ export class LoremModule {
...
     const method = this.faker.helpers.arrayElement(methods);
 
-    return `${this[method]()}`;
+    if (typeof options === 'undefined') {
+      return `${this[method]()}`;
+    }
+
+    if (typeof options === 'number') {
+      return this.faker.lorem.text({ length: options });
+    }
```
ST-DDT:
```suggestion
    if (typeof options === 'number') {
      options = { length: options };
    }
```

src/modules/lorem/index.ts
```diff
@@ -318,10 +324,33 @@ export class LoremModule {
...
    * // Dolor tempora iusto.'
+   * faker.lorem.text(14) // 'Doloribus aut.'
+   * faker.lorem.text({length: 14}) // 'Doloribus aut.'
+   * faker.lorem.text({length: {min: 10, max: 15}}) // 'autem non quis.'
    *
    * @since 3.1.0
    */
-  text(): string {
+  text(
+    options?:
```
ST-DDT:
```suggestion
    options:
```

src/modules/lorem/index.ts
```diff
@@ -318,10 +324,33 @@ export class LoremModule {
...
+                /**
+                 * The minimum length of text to generate.
+                 */
+                min?: number;
+                /**
+                 * The maximum length of text to generate.
+                 */
+                max?: number;
+              };
+        }
```
ST-DDT:
```suggestion
        } = {}
```

src/modules/lorem/index.ts
```diff
@@ -332,7 +361,57 @@ export class LoremModule {
...
 
-    return `${this[method]()}`;
+    if (typeof options === 'undefined') {
+      return `${this[method]()}`;
+    }
+
+    if (typeof options === 'number') {
+      return this.faker.lorem.text({ length: options });
+    }
+
```
ST-DDT, line 364: You can use https://fakerjs.dev/api/helpers.html#rangetonumber to simplify this:

```suggestion

    const length = options.length == null ? null : faker.helpers.rangeToNumber(options.length);

    if (length == null) {
      return `${this[method]()}`;
    }

```

src/modules/lorem/index.ts
```diff
@@ -332,7 +361,57 @@ export class LoremModule {
 
     const method = this.faker.helpers.arrayElement(methods);
 
-    return `${this[method]()}`;
+    if (typeof options === 'undefined') {
+      return `${this[method]()}`;
+    }
```
ST-DDT:
```suggestion
```

Then you don't need this part and the behavior is more predictable.

e.g. `text()` behaves the same as `text({})`

ST-DDT (changes requested):

ghost:
Thank you. I am not sure if I implemented your feedback correctly, but I tried to do my best.

src/modules/lorem/index.ts
```diff
@@ -310,6 +311,11 @@ export class LoremModule {
   /**
    * Generates a random text based on a random lorem method.
    *
+   * @param options The options for the text to generate or options.length as quick primitive argument.
+   * @param options.length The length (range) of text to generate.
+   * @param options.length.min The minimum length of text to generate.
+   * @param options.length.max The maximum length of text to generate.
```
ST-DDT: We dont add `@param` blocks for these doubly nested options.

src/modules/lorem/index.ts
```diff
@@ -332,7 +361,49 @@ export class LoremModule {
 
     const method = this.faker.helpers.arrayElement(methods);
 
-    return `${this[method]()}`;
+    if (typeof options === 'number') {
+      options = { length: options };
+    }
+
```
ST-DDT, line 364: You can use https://fakerjs.dev/api/helpers.html#rangetonumber to simplify the code below:

```suggestion

    if (options.length == null) {
      return this[method]();
    }

    const length = this.faker.helpers.rangeToNumber(options.length);

```

src/modules/lorem/index.ts
```diff
@@ -332,7 +359,46 @@ export class LoremModule {
...
+        );
+      }
+
+      if (options.length.max != null && options.length.max < 0) {
+        throw new FakerError(
+          `Max ${options.length.max} should be a non-negative integer.`
+        );
+      }
+    }
+
```
ST-DDT: There is no need to check for these.
1. We don't check for the presence of required values
2. We ignore negative bounds and treat them as zero (for string lengths).

```suggestion
```

src/modules/lorem/index.ts
```diff
@@ -332,7 +359,46 @@ export class LoremModule {
...
+    const length = this.faker.helpers.rangeToNumber(options.length);
+
+    if (length === 0) {
```
ST-DDT:
```suggestion
    if (length <= 0) {
```

src/modules/lorem/index.ts
```diff
@@ -332,7 +359,46 @@ export class LoremModule {
...
+      return '';
+    }
+
+    let text = `${this[method]()}`;
+    while (text.length <= length) {
+      text = `${text} ${this[method]()}`;
+    }
```
ST-DDT: It is also possible to write this as a do-while-loop, but I'm not sure about that:

```suggestion
    let text = '';
    do {
      text = `${text}${this[method]()} `;
    } while (text.length < length);
```

test/modules/lorem.spec.ts
```diff
@@ -325,6 +325,117 @@ describe('lorem', () => {
...
+          expect(actual.length).toBeLessThanOrEqual(66);
+        });
+
+        it('should return text with a length in [0, 51]', () => {
+          const actual = faker.lorem.text({ length: { min: 0, max: 51 } });
+
+          expect(actual).toBeTypeOf('string');
+          expect(actual.length).toBeGreaterThanOrEqual(0);
+          expect(actual.length).toBeLessThanOrEqual(51);
+        });
```
ST-DDT: IMO there are a few redundant tests here.

ST-DDT (commented):
Also, should we stick with `sentence()` (if length is provided) to produce a more consistent text?
(We end with a dot, so it would make sense to have dots in between as well.)

src/modules/lorem/index.ts
ghost: What is the advantage of doing is this way?

ghost:
Yes, it would make sense to put dots in between as well.

src/modules/lorem/index.ts
ST-DDT: There are two tiny benefits:

1. The trailing ` ` adds one to the text.length and thus terminates the loop potentially early (this would also work in the normal while loop)
2. The method is invoked in only one location

With the tiny drawback that there is one additional operation with the empty string

src/modules/lorem/index.ts
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
...
+    const length = this.faker.helpers.rangeToNumber(options.length);
+
+    if (length <= 0) {
+      return '';
+    }
+
+    let text = '';
+    do {
+      text = `${text}${this[method]()} `;
+    } while (text.length <= length + 1);
```
ST-DDT:
```suggestion
    } while (text.length < length - 1);
```

If `text.length === length - 1` then the substring below doesn't remove anything and appends the '.' string => `text.length === length`. It would end in `foo .` but that would still be the case even if you append another word to it, because it would get removed via the substring call.

---

Should we deduct the `-1` from `length` early?

````ts
// deduct one from the length for the final dot
length--;
````

---

Alternatively see this comment: https://github.com/faker-js/faker/pull/2227#discussion_r1299768500

File: src/modules/lorem/index.ts [hunk 26]
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
...
+
+    const length = this.faker.helpers.rangeToNumber(options.length);
+
+    if (length <= 0) {
+      return '';
+    }
+
+    let text = '';
+    do {
+      text = `${text}${this[method]()} `;
```
ST-DDT, line 377: If we go for `sentence()` (not sure yet):

```suggestion
      text = `${text}${this.sentence()} `;
```

src/modules/lorem/index.ts
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
 
     const method = this.faker.helpers.arrayElement(methods);
```
ST-DDT, line 359: If we go for `sentence()` (not sure yet):

You can move this to the `if (options.length == null)` block.

src/modules/lorem/index.ts
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
...
+    if (length <= 0) {
+      return '';
+    }
+
+    let text = '';
+    do {
+      text = `${text}${this[method]()} `;
+    } while (text.length <= length + 1);
+
+    return `${text.trim().substring(0, length - 1)}.`;
```
ST-DDT: To prevent whitespace followed by a dot:

```suggestion
    return `${text.trim().substring(0, length - 1).replace(/\s$/, '.')}.`;
```

test/modules/lorem.spec.ts
```diff
@@ -325,6 +325,77 @@ describe('lorem', () => {
...
+          expect(actual.length).toBeLessThanOrEqual(32);
+        });
+
+        it('should return a string of length 77 for [77, 77]', () => {
+          const actual = faker.lorem.text({ length: { min: 77, max: 77 } });
+
+          expect(actual).toBeTruthy();
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toHaveLength(77);
+        });
```
ST-DDT: IMO these are already covered by `should return text with a length in range [25, 40]'`

src/modules/lorem/index.ts
ST-DDT: Alternatively, we could not deduct 1 (here and in the loop) from the length and use `.replace(/.$/, '.')` without appending the final dot.

ST-DDT:
@inkedtree We kind of are unsure what requirements we have for the method/the parameter. And would like to hear your requirements for the method.
This method generates vastly different types of text (some with newlines, some without).
So do we need the length parameter for the specific versions as well? Such as sentences, lines?
Or should we limit the text method to sentences?
Would it get confusing if we have `lines( { length: number})` aka does the length refer to each line or to the entire string?

ghost:
@ST-DDT

Sorry, for the late reply. I am not sure, if I understand all your questions correctly.

To be honest, I am not entirely sure what my requirements are after reading about all available methods in the `lorem` module. Originally, I wanted to have a method that generates sentences with a particular length. So, basically a paragraph, I guess. But now I am not sure, if it wouldn't be better to add the parameter to `paragraph(s)` for example in this case.

Sometimes when I am using faker I would like to be able to generate sentences, lines and so on with a specific length (range).

I think, it would get confusing if the `length` parameter does not refer to the length of the entire string, e.g. of lines.

src/modules/lorem/index.ts
ghost: I changed it similar to the way you described in https://github.com/faker-js/faker/pull/2227#discussion_r1299768500.

src/modules/lorem/index.ts
```diff
@@ -318,10 +321,33 @@ export class LoremModule {
...
+      | number
+      | {
+          /**
+           * The length (range) of text to generate.
+           */
+          length?:
+            | number
+            | {
+                /**
+                 * The minimum length of text to generate.
```
ejcheng, line 341:
```suggestion
                 * The minimum amount of characters to generate.
```

src/modules/lorem/index.ts
```diff
@@ -318,10 +321,33 @@ export class LoremModule {
...
+                 */
+                min: number;
+                /**
+                 * The maximum length of text to generate.
```
ejcheng, line 345:
```suggestion
                 * The maximum amount of characters to generate.
```

src/modules/lorem/index.ts
```diff
@@ -310,6 +310,9 @@ export class LoremModule {
   /**
    * Generates a random text based on a random lorem method.
    *
+   * @param options The options for the text to generate or options.length as quick primitive argument.
+   * @param options.length The length of text to generate as number or range.
```
ejcheng, line 314:
```suggestion
   * @param options.length The length of text (in characters) to generate as number or range.
```

ejcheng (changes requested):

ST-DDT:
**Team Decision**

- We accept this feature request. Please rebase it and work on the suggestions we point out.

src/modules/lorem/index.ts
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
...
-    return `${this[method]()}`;
+    if (typeof options === 'number') {
+      options = { length: options };
+    }
+
+    if (options.length == null) {
+      return `${this[method]()}`;
+    }
+
+    const length = this.faker.helpers.rangeToNumber(options.length);
```
ST-DDT, line 369: Please use our standard way of decstructuring options:
```suggestion
    const { length: lengthOrRange } = options;

    if (lengthOrRange == null) {
      return `${this[method]()}`;
    }

    const length = this.faker.helpers.rangeToNumber(lengthOrRange);
```

src/modules/lorem/index.ts
ST-DDT, line 377: Please only use `this.sentence()` here and maybe randomly (e.g. faker.datatype.boolean()) add a newline to simulate paragraphs/lines.

src/modules/lorem/index.ts
```diff
@@ -332,7 +358,26 @@ export class LoremModule {
...
+    if (length <= 0) {
+      return '';
+    }
+
+    let text = '';
+    do {
+      text = `${text}${this[method]()} `;
+    } while (text.length < length);
+
+    return `${text.substring(0, length).replace(/.$/, '.')}`;
```
ST-DDT, line 380: Please just return the value without dot at the end.

```suggestion
    return text.substring(0, length);
```

test/modules/lorem.spec.ts
```diff
@@ -325,6 +325,53 @@ describe('lorem', () => {
...
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toBe('');
+        });
+
+        it('should return a dot for length one', () => {
+          const actual = faker.lorem.text({ length: 1 });
+
+          expect(actual).toBeTypeOf('string');
+          expect(actual).toBe('.');
+        });
```
ST-DDT, line 374: This test is no longer needed,

ST-DDT (changes requested):

ST-DDT:
Closed as stale. We can reopen this when you want to continue with it.

bmulholland:
I just wrote a custom function for this and would love to have it baked into faker. It looks like the original submitter deleted their Github account. I'd be open to addressing the comments and submitting a fresh PR. Would that be accepted, @ST-DDT ?

xDivisionByZerox:
> I just wrote a custom function for this and would love to have it baked into faker. It looks like the original submitter deleted their Github account. I'd be open to addressing the comments and submitting a fresh PR. Would that be accepted, @ST-DDT ?

@bmulholland feel free to open another PR. We always love when the community gets involved. Make sure to read our contribution guide before hand and like the original issue.

PR #2229, opened 2023-06-29
refactor(phone): deprecate `format` argument for `phone.number`

### LINKED ISSUE #1541: Deprecate phone.number format parameter

The `format` can be used to overwrite the pattern that is used to generate the phone number.
Instead of passing it to phone.number the user should pass it directly to `helpers.replaceSymbolWithNumber` and build their own phone number method. Alternatively, they can add the format to our repository and thus provide it for everyone.

### Suggested solution

Deprecate the `format` parameter for removal.

### Alternative

- https://github.com/faker-js/faker/issues/1542

### Additional context

_No response_

xDivisionByZerox:
Closes #1541.


ST-DDT (approved):

ejcheng (approved):

sheremet-va:
Is there a reason why this deprecation doesn't mention the migration path? This change is not even mentioned in changelog.

matthewmayer:
See #2413 where we are discussing how to improve the deprecation message.

PR #2230, opened 2023-07-01
feat(locale): add Esperanto

RobinvanderVliet:
This pull request is ready to be merged!

Feedback and reviews are welcome and appreciated! This is my first merge request.


src/locales/eo/person/last_name.ts
```diff
@@ -0,0 +1,27 @@
+export default [
+  'Auld',
```
matthewmayer, line 7: Is there anything particularly Esperanto-ish about these surnames? Would it be better to let it fall back to the en surname list if there are no popular eo surnames?

RobinvanderVliet: I don't think it's a good idea to use an exclusively English fallback list, as the Esperanto community is very diverse and international.

This is a list of family names of famous deceased people who were important to Esperanto and the community in some way. It contains not only English names, but also names in Polish, German, Dutch, French and some other languages.

matthewmayer: That seems fine, perhaps you could add a comment about that in the file as the source is not so obvious.

ST-DDT (dismissed):

src/locales/eo/person/last_name.ts
```diff
@@ -0,0 +1,109 @@
+// It would not be a good idea to have an English-only fallback list,
+// because the Esperanto community is very diverse and international.
+// Therefore, this list contains the family names of famous deceased people
+// who were in some way important to Esperanto and its community.
+// Most Esperanto speakers will recognize at least some of them.
+// The list contains 100 family names from multiple languages.
```
Shinigami92: Thinking about to formulate this in a bit more "positive language" (not using "[...] would not be a good idea [...]"), I made a ChatGPT request for some ideas: https://chat.openai.com/share/873706bf-58f8-43d7-9005-553179f58a8b

RobinvanderVliet: Than you for your comment, I used this suggestion!

ST-DDT (dismissed):

src/locales/eo/person/last_name.ts
ST-DDT: IMO this is exageratingly positive/sugar coated, but if you like it, I dont mind.

Shinigami92: I also found the last suggestion from ChatGPT slightly too far :laughing: 

Maybe we should just use the version one chat-comment before :sweat_smile: 

> Embracing the wonderful diversity and international essence of the Esperanto community, it's recommended to have a versatile fallback list that goes beyond just English. This list celebrates the legacy of influential figures connected to Esperanto and its community who have left a lasting impact. Numerous Esperanto speakers will have the joy of recognizing many of these names. The compilation comprises 100 family names representing a variety of languages.

ST-DDT: Why dont we keep it simple and just say.

> Esperanto is a lingually(?) diverse community so we chose to use the names of important persons in that community instead of joining the names of all those languages.

Shinigami92: Because I didn't used :brain: but ChatGPT

RobinvanderVliet: I simplified the comment even more.

Shinigami92 (approved):

ST-DDT (approved):

ST-DDT:
For future PRs: please do one category of locale data (e.g. person) or ~< 20 files at a time unless necessary.
As that makes the PRs less time consuming to review in full.

RobinvanderVliet:
> For future PRs: please do one category of locale data (e.g. person) or ~< 20 files at a time unless necessary. As that makes the PRs less time consuming to review in full.

Thank you for the suggestion, I will remember it for my next pull request! I created some more files for the Esperanto locale, which I will submit in my next merge request, after this one gets merged.

PR #2232, opened 2023-07-01
feat(locale): add Dutch province abbreviations

RobinvanderVliet:
According to: https://en.wikipedia.org/wiki/ISO_3166-2:NL


xDivisionByZerox (dismissed):
While I already approve, it would be nice if you could already sort the values alphabetically (a-z). We plan on 
enforcing this via the locale generation script in the future. So if you could already do this we could reduce the git diff in the future :)

RobinvanderVliet:
> While I already approve, it would be nice if you could already sort the values alphabetically (a-z). We plan on enforcing this via the locale generation script in the future. So if you could already do this we could reduce the git diff in the future :)

I intentionally kept them in the same order as [state.ts](https://github.com/faker-js/faker/blob/next/src/locales/nl/location/state.ts), because I thought the two files were connected in some way.

If that is not needed, I can order them alphabetically.

ST-DDT:
> I intentionally kept them in the same order as [state.ts](https://github.com/faker-js/faker/blob/next/src/locales/nl/location/state.ts), because I thought the two files were connected in some way.
> 
> If that is not needed, I can order them alphabetically.

No, the files aren't connected with each other, so please sort this one alphabetically.
We keep connected data in the same file so they can be changed in one step and dont risk them diverging from each other. See also our airline.airline data.

RobinvanderVliet:
I sorted them in this pull request and also in #2233.

xDivisionByZerox (approved):

matthewmayer:
@RobinvanderVliet don't worry about merging next into your PR branches, this will be done automatically before the feature is squash merged. You only need to bother if there are conflicts.

ejcheng (approved):

matthewmayer (approved):

ST-DDT (approved):

Issue #2207, opened 2023-06-09
`County` method used with `uniqueArray` helper crashes faker.


jviall:
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md).
- [X] Read the [docs](https://fakerjs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/faker-js/faker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Faker issue and not related to a combination with another package.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/faker-js/faker/discussions) or join our [Discord Chat Server](https://chat.fakerjs.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.
- [ ] I am willing to provide a PR.

### Describe the bug

Using the `county` method with the `uniqueArray` completely crashes not just faker but my entire runtime. In my case, I discovered it in a React + Storybook project, where I switched from using `country` to using `county`, and my entire browser became unresponsive.

Attached is a stackblitz (based on the one linked in the docs), that does the same thing, comparing the `country` method to the `county` method using the `uniqueArray` helper, where generating some unique countries works fine, but counties do not.

### Minimal reproduction code

https://stackblitz.com/edit/faker-js-demo-rqnnk9?file=index.ts

### Additional Context

I personally use faker in a Vite + Storybook + React environment, using ES Modules targeting an ESNext build. However, it is reproduce-able (as shown in the stackblitz) in plain JS modules.

### Environment Info

```shell
System:
    OS: macOS 13.2.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 63.45 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/Library/Caches/fnm_multishells/22781_1678232715369/bin/node
    Yarn: 3.4.1 - /usr/local/bin/yarn
    npm: 9.5.1 - ~/Library/Caches/fnm_multishells/22781_1678232715369/bin/npm
  Browsers:
    Chrome: 114.0.5735.106
    Firefox: 112.0.2
    Safari: 16.3
  npmPackages:
[...]

PR #2238, opened 2023-07-05
feat(location): update en county list

matthewmayer:
re #2207 #1944

Updates the en county list to be a lomger list of US-esque county names instead of a 
truncated list of GB county names.


ejcheng (dismissed):

ST-DDT:
Shouldn't we mix other states counties in as well otherwise en == en_US.
Maybe lets discuss this in the next team meeting.

matthewmayer:
We don't do that for state()

ST-DDT:
There seems to be merge conflicts. Could you please fix them?

ST-DDT:
**Team decision**

We will use both the "British" and US county names for `en`.

matthewmayer:
Done. 

Note this is a good example of why you don't always want to sort definitions alphabetically re #2265

src/locales/en/location/county.ts
ST-DDT, line 1: Should we use [mergeArrays](https://github.com/faker-js/faker/blob/aa1bb131d874591e4d46671b1e82b23fc2a7ad74/src/internal/merge.ts#L10) for this?

matthewmayer: i was concerned that might introduce some unwanted dependencies between locales which might cause issues with tree-shaking in the future (but i dont understand how tree shaking works enough to know if that would be the case)

ST-DDT: Me neither (although I don't expect it to have much of an impact on ESM). 
I hope others can shed some light on this.

ST-DDT (approved):

Shinigami92 (approved):

src/locales/en/location/county.ts
xDivisionByZerox: But we should at least add a test which checks that `en` is a combination of `en_US` and `en_GB`.

ST-DDT: 🤔

ST-DDT: I don't think that's necessary.

PR #2239, opened 2023-07-05
fix(helpers): prevent uniqueArray from hanging

### LINKED ISSUE #2207: `County` method used with `uniqueArray` helper crashes faker.

matthewmayer:
fix #2207

Adds a max retry logic to uniqueArray.

I chose to use 1000*length rather than a fixed number of retries, as for example if you are 
asking for a length of 2000, then 1000 retries would be insufficient.

This seems sufficient for most sensible use cases.

Obviously uniqueArray has no way to "know" how many unique values the function can return, you could define a function which returns "heads", "tails", but occasionally returns "edge" every 1000000 runs, or if the system date is a Thursday, etc etc.


xDivisionByZerox (approved):
The implementation looks good. The failing pipeling seems to be unrelated to your changes. So I already approve 👍

matthewmayer:
> The implementation looks good. The failing pipeling seems to be unrelated to your changes. So I already approve 👍 

Yea that codecov job is quite flaky and seems to fail every so often at random.

ST-DDT:
Instead of reinventing the unique logic, shouldn't we reuse the unique method/logic?

matthewmayer:
> Instead of reinventing the unique logic, shouldn't we reuse the unique method/logic?

I thought that was deprecated?

Shinigami92 (approved):

ejcheng (approved):

ST-DDT:
Let's discuss this in the next team meeting.

ST-DDT:
**Team Decision**

We consider this a bandaid fix that solves the issue.
In the future we will reevaluate the entire unique feature again and then also use it consistently in all our unique methods.

ST-DDT (approved):

PR #2240, opened 2023-07-05
feat(commerce): Add method for generating ISBN-10 and ISBN-13

### LINKED ISSUE #2196: ISBN 10 and ISBN13

Add ISBN 10 and 13 generators to Faker. They look like this:

```
isbn10: "0140449337"
isbn13: "978-0140449334"
```

### Suggested solution

Add it to `commerce` or add a new root level generator for books of which this ISBN generator would be a member of.

### Alternative

_No response_

### Additional context

This is already implemented in faker libraries in other ecosystems: https://faker.readthedocs.io/en/master/providers/faker.providers.isbn.html

RobinvanderVliet:
This is my first pull request contributing some actual code to this project, so any feedback is welcome!

This merge request resolves issue #2196.


src/modules/commerce/index.ts
```diff
@@ -266,4 +338,82 @@ export class CommerceModule {
...
+        } = {}
+  ): string {
+    if (typeof options === 'number') {
+      options = { variant: options };
+    }
+
+    const { variant = 13, separator = '-' } = options;
+
+    const prefix = '978';
+    const group: string = this.faker.number.int({ min: 0, max: 1 }).toString();
```
ST-DDT: You might want to use `faker.helpers.objectKey(ISBN_LENGTH_RULES)` here instead.

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,82 @@ export class CommerceModule {
...
+    const data: string[] = [prefix, group, registrant, publication];
+    if (variant === 10) {
+      data.shift();
+    }
+
+    const isbn: string = data.join('');
+
+    let checksum = 0;
+    for (let i = 0; i < variant - 1; i++) {
+      const weight: number = variant === 10 ? i + 1 : i % 2 ? 3 : 1;
```
ST-DDT: IMO in cases like these it is obvious that weight is a number, you don't need the explicit type definition.

test/commerce.spec.ts
```diff
@@ -158,6 +169,55 @@ describe('commerce', () => {
           );
         });
       });
+
+      describe(`isbn()`, () => {
+        it('should return ISBN-13 with hyphen separators when not passing arguments', () => {
```
ST-DDT: Please add a test for isn10+13 each that validates the generated isbn with validatorjs's [`isISBN()`](https://github.com/validatorjs/validator.js/blob/master/src/lib/isISBN.js)

See here for example: https://github.com/faker-js/faker/blob/4a15bd0d4eea5e5f6a7d1e6be0c2bf9e59feb069/test/internet.spec.ts#L555

Please pass the variant as options to the check, so it checks for the exact variant only.

ST-DDT (commented):
Please note that the associated issue is marked as waiting for user interest, so this PR might not be merged immediately.

src/modules/commerce/index.ts
RobinvanderVliet: I removed the explicit type definition. I am used to always adding them.

test/commerce.spec.ts
RobinvanderVliet: I added the check to the unit tests.

src/modules/commerce/index.ts
RobinvanderVliet: Done!

src/modules/commerce/index.ts
```diff
@@ -1,6 +1,78 @@
 import type { Faker } from '../../faker';
 import { deprecated } from '../../internal/deprecated';
 
+// Source for official prefixes: https://www.isbn-international.org/range_file_generation
+const ISBN_LENGTH_RULES: {
+  [group: string]: Array<[rangeMaximum: number, length: number]>;
+} = {
```
ST-DDT: If we simplify the type here to:

```suggestion
const ISBN_LENGTH_RULES: Record<
  string,
  Array<[rangeMaximum: number, length: number]>
> = {
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+    if (typeof options === 'number') {
+      options = { variant: options };
+    }
+
+    const { variant = 13, separator = '-' } = options;
+
+    const prefix = '978';
+    const group: string = this.faker.helpers
+      .objectKey(ISBN_LENGTH_RULES)
+      .toString();
```
ST-DDT: Then we can simplify this to:

```suggestion
    const [group, groupRules] =
      this.faker.helpers.objectEntry(ISBN_LENGTH_RULES);
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+    const element: string = this.faker.string.numeric(8);
+
+    const registrantLength: number = ISBN_LENGTH_RULES[group].find(
+      ([rangeMaximum]): boolean =>
+        parseInt(element.slice(0, -1)) <= rangeMaximum
+    )[1];
```
ST-DDT: Then you can change `ISBN_LENGTH_RULES[group]` to `groupRules`. Also the `parseInt(element.slice(0, -1))` appears to be static, so IMO you can calculate that only once and reuse the value in the filter.

```suggestion
    const elementValue = parseInt(element.slice(0, -1));

    const registrantLength = groupRules.find(
      ([rangeMaximum]) => elementValue <= rangeMaximum
    )[1];
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+      options = { variant: options };
+    }
+
+    const { variant = 13, separator = '-' } = options;
+
+    const prefix = '978';
+    const group: string = this.faker.helpers
+      .objectKey(ISBN_LENGTH_RULES)
+      .toString();
+    const element: string = this.faker.string.numeric(8);
```
ST-DDT: IMO there is no need to explicitly define that this is a string here.

```suggestion
    const element = this.faker.string.numeric(8);
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+
+    const registrantLength: number = ISBN_LENGTH_RULES[group].find(
+      ([rangeMaximum]): boolean =>
+        parseInt(element.slice(0, -1)) <= rangeMaximum
+    )[1];
+
+    const registrant: string = element.slice(0, registrantLength);
+    const publication: string = element.slice(registrantLength);
```
ST-DDT: Same here

```suggestion
    const registrant = element.slice(0, registrantLength);
    const publication = element.slice(registrantLength);
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+
+    const data: string[] = [prefix, group, registrant, publication];
```
ST-DDT:
```suggestion
    const data = [prefix, group, registrant, publication];
```

src/modules/commerce/index.ts
```diff
@@ -266,4 +338,84 @@ export class CommerceModule {
...
+    if (variant === 10) {
+      data.shift();
+    }
+
+    const isbn: string = data.join('');
```
ST-DDT:
```suggestion
    const isbn = data.join('');
```

RobinvanderVliet: Thank you for your review! I've processed your comments. :+1:

src/modules/commerce/index.ts
```diff
@@ -258,4 +331,83 @@ export class CommerceModule {
...
+   *
+   * @example
+   * faker.commerce.isbn() // '978-0-692-82459-7'
+   * faker.commerce.isbn(10) // '1-155-36404-X'
+   * faker.commerce.isbn(13) // '978-1-60808-867-6'
+   * faker.commerce.isbn({ separator: ' ' }) // '978 0 452 81498 1'
+   * faker.commerce.isbn({ variant: 10, separator: ' ' }) // '0 940319 49 7'
+   * faker.commerce.isbn({ variant: 13, separator: ' ' }) // '978 1 6618 9122 0'
+   *
+   * @since 8.0.0
```
ST-DDT:
```suggestion
   * @since 8.1.0
```

ST-DDT (dismissed):

ST-DDT (dismissed):

xDivisionByZerox:
**Team decision**

We accept this feature request. 

---

@RobinvanderVliet Please resolve to merge conflicts so we can open it for review.

ST-DDT:
@RobinvanderVliet Could you please fix the merge conflicts?

RobinvanderVliet:
I resolved the merge conflicts!

ST-DDT (approved):

xDivisionByZerox (approved):
I’m considering the use of a strategy pattern to improve the maintainability of the changed code. However, I have decided to accept the current implementation.

ejcheng (approved):

PR #2254, opened 2023-07-15
chore: enable strictBindCallApply

Shinigami92:
I experiment and moved the repeating code into an internal function.
It seems to work and we can really pass `this` just down into the function.

However typing this function correctly exceeds my TypeScript knowledge 😖


Shinigami92:
I asked ChatGPT for some help



However I used GitHub Copilot to help me a bit to define the types

I don't really understand what's going on, or to be more precise: "why is this working" :thinking: :shrug:

Shinigami92:
@ST-DDT and @xDivisionByZerox 

please check if you are both
1. satisfied with the name of the function
2. satisfied with the order of the params, or we should swap them (`(that, moduleClass)`)
3. want to change the JSDocs and maybe bring the inline comment up to the JSDoc

Additionally I think I should add a unit test that explicitly checks that the function is doing for what it is intended
But before I do that, please confirm that I'm on the right track and should move on with this

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param moduleClass The class to bind the methods of.
+ * @param that The class instance to bind the methods to.
+ */
+export function bindToNamespace<TClass extends { new (...args: any[]): any }>(
+  moduleClass: TClass,
+  that: InstanceType<TClass>
+): void {
+  // Bind `this` so namespaced is working correctly
+  for (const name of Object.getOwnPropertyNames(moduleClass.prototype) as Array<
```
xDivisionByZerox: Im pretty sure that you could completely get rid of the class argument in this function by doing the following here:


```suggestion
  for (const name of Object.getOwnPropertyNames(Object.getPrototypeOf(that)) as Array<
```

I'd suggest this as it is quite confusing to me that you have to explicitly have to pass the class reference which you are currently instantiating.

[POC in TS Playground](https://tsplay.dev/WYlxQm) (run and check the console)

xDivisionByZerox:
> 1. satisfied with the name of the function

I'd probably go more for something like `bindThisScope` or `bindThisForFunctions`, since `namespace` is a different concept in JS.

> 2. satisfied with the order of the params, or we should swap them (`(that, moduleClass)`)

[See my review comment](https://github.com/faker-js/faker/pull/2254#discussion_r1264367059).

> 3. want to change the JSDocs and maybe bring the inline comment up to the JSDoc

I'll have an additional review for that most likely.

> Additionally I think I should add a unit test that explicitly checks that the function is doing for what it is intended
 
Big +1. Working with `this` scopes can get quite confusing for some people. So, a test should help verify the intentions of the function.

src/internal/bind-to-namespace.ts
ST-DDT: I tested the changes with the faker project and it worked well.

ST-DDT: Do we still need the explicit cast?

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Bind all methods of a class to the class instance.
+ *
+ * @param moduleClass The class to bind the methods of.
+ * @param that The class instance to bind the methods to.
+ */
+export function bindToNamespace<TClass extends { new (...args: any[]): any }>(
```
ST-DDT: We could reduce the number of `any`s here:

```suggestion
export function bindToNamespace<TClass extends { new (faker: Faker): any }>(
```

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Bind all methods of a class to the class instance.
```
ST-DDT:
```suggestion
 * Bind all functions of the given instance to itself so you can use them independently.
```

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
```
ST-DDT:
```suggestion
 *
 * @internal
 *
```

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param moduleClass The class to bind the methods of.
+ * @param that The class instance to bind the methods to.
```
ST-DDT:
```suggestion
 * @param that The class instance of which the methods are to be bound to itself.
```

Not sure how to word this better.

src/internal/bind-to-namespace.ts
ST-DDT: Maybe call this `bindToItself` or `bindFunctionsToItself`?

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
```
ST-DDT:
```suggestion
 *
 * @example
 * const someModule = new SomeModule(faker);
 * bindToNamespace(someModule); // Usually called inside the constructor
 * const someMethod = someModule.someMethod;
 * someMethod(); // Works
 */
```

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+export function bindToNamespace<TClass extends { new (...args: any[]): any }>(
+  moduleClass: TClass,
+  that: InstanceType<TClass>
+): void {
+  // Bind `this` so namespaced is working correctly
+  for (const name of Object.getOwnPropertyNames(moduleClass.prototype) as Array<
+    keyof TClass | 'constructor'
+  >) {
+    if (name === 'constructor' || typeof that[name] !== 'function') {
```
ST-DDT: I would flip this, because `'typeof constructor === 'function'` so there is no need to explicitly check for the name, if the type doesn't match anyway.

```suggestion
    if (typeof that[name] !== 'function' || name === 'constructor') {
```

src/internal/bind-to-namespace.ts
ST-DDT: Or even just:

```suggestion
export function bindToNamespace<TClass extends { new (): any }>(
```

src/internal/bind-to-namespace.ts
```diff
@@ -0,0 +1,21 @@
...
+      continue;
+    }
+
+    that[name] = that[name].bind(that);
```
ST-DDT: Maybe you can even go a step further

```suggestion
    if (typeof that[name] === 'function' && name !== 'constructor') {
      that[name] = that[name].bind(that);
    }
```

Shinigami92:
Great help :raised_hands: Thank you both :heart_eyes_cat:

Shinigami92:
LOL Today I learned:

```node
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/internal/bind-this-to-member-functions.spec.ts > internal > bind-this-to-member-functions > should bind this to member functions
AssertionError: expected error to have message: Cannot read properties of undefined (reading 'faker')

- Expected
+ Received

- Cannot read properties of undefined (reading 'faker')
+ Cannot read property 'faker' of undefined
```

NodeJS v14 has another error message :laughing:

src/internal/bind-this-to-member-functions.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param instance The class instance of which the methods are to be bound to itself.
+ *
+ * @example
+ * const someModule = new SomeModule(faker);
+ * bindThisToMemberFunctions(someModule); // Usually called inside the constructor passing `this`
+ * const someMethod = someModule.someMethod;
+ * someMethod(); // Works
+ */
+export function bindThisToMemberFunctions<TClass extends { new (): any }>(
```
ST-DDT, line 14: IMO this should be the other way round:

```suggestion
export function bindMemberFunctionsToThis<TClass extends { new (): any }>(
```

ST-DDT, line 14: Please fix/mark this warning as ignored.

Why is this marked as warning instead of error? I thought we set everything to `error`.

test/internal/bind-this-to-member-functions.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { describe, expect, it } from 'vitest';
+import type { Faker } from '../../src';
+import { faker } from '../../src';
+import { bindThisToMemberFunctions } from '../../src/internal/bind-this-to-member-functions';
+
+describe('internal', () => {
```
ST-DDT, line 6: IMO you can omit this wrapper.

src/internal/bind-this-to-member-functions.ts
Shinigami92: I think it's coming from a `recommended` set

test/internal/bind-this-to-member-functions.spec.ts
Shinigami92: I reflected the folder structure, this is normally the way to go / a semi-standard

src/internal/bind-this-to-member-functions.ts
xDivisionByZerox: But the `this` scope already has the member function :thinking:

Shinigami92: But this gets bound to the member function, not the function to this

ST-DDT: I'm not sure whether I understand you here.

---

Maybe `bindThisOnMemberFunctions`?

test/internal/bind-this-to-member-functions.spec.ts
ST-DDT: We don't do that here 😉 (in this project).

We don't do that here

src/internal/bind-this-to-member-functions.ts
ST-DDT: @import-brain Can you help us with the wording here?

ejcheng: > @import-brain Can you help us with the wording here?

Because we're binding `this` to the member functions, `bindThisToMemberFunctions` would be the best wording in this case.

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2260, opened 2023-07-17
chore(deps): update dependency prettier to v3

Shinigami92:
This is a test PR on top of #2249
It shows what changes and what needs to be fixed

PR #2261, opened 2023-07-17
chore: convert to esm

Shinigami92:
Convert the whole faker code base to ESM


scripts/bundle.ts
```diff
@@ -16,27 +15,26 @@ if (existsSync(localeDir)) {
...
-  //   './src/index.ts',
-  //   ...Object.keys(locales).map((locale) => `./src/locale/${locale}.ts`),
-  // ],
+  entryPoints: [
+    './src/index.ts',
+    ...Object.keys(allLocales).map((locale) => `./src/locale/${locale}.ts`),
+  ],
   outdir: './dist/cjs',
-  bundle: false, // Creates 390MiB bundle ...
+  bundle: true, // Creates 49MiB bundle ...
```
ST-DDT: So this change fixes #1791 ?
Any idea why?

Shinigami92: no no ^^ you misunderstand something there
this actually bumps the download size of faker to >= 50 mb (from currently around < 10 mb)

so this https://www.npmjs.com/package/@faker-js/faker


would become something greater then 50 mb in unpacked size

matthewmayer:
What are the benefits of this?

Shinigami92:
> What are the benefits of this?

support for future features, top-level await in jsdoc generation scripts, enhance speed of disk access, and more

It's a try for https://github.com/faker-js/faker/pull/2260#issuecomment-1638380195

Shinigami92:
After already sinking in around 5 hours of net time, I will give this a last chance today (when I find the time)

I read an article about placing package.json files into the dist/esm and dist/cjs folders that have separate configured `"type"` fields.
That way the cjs folder would read resolved .js files as cjs instead of esm

If that does not work, I will give up for now and we might even consider to release a `@faker-js/faker` together with a `@faker-js/faker-cjs` 🤷
This would even come with some advantages like we can see in the weekly download stats how necessary the cjs version even is and also reduce the unpacked size of each version into half/half

Shinigami92:
Putting a `package.json` with `{ "type": "commonjs" }` doesn't work, as not the package.json from the source, but the one next to `dist` is relevant

Shinigami92:
awaiting merge of #2391

Shinigami92:
There are errors :confused:  
I ran out of time right now, need to look for them later

pnpm-lock.yaml
ST-DDT, line 1: Changing dependencies seems unrelated to this PR.
Please revert.

ST-DDT:
Looks like you have to change how https://github.com/faker-js/faker/blob/next/.github/workflows/commentCodeGeneration.ts is converted to js.
(Or run it using tsx instead?)

ST-DDT:
I assume the issue is more in the calling of the file:
https://github.com/faker-js/faker/blob/db88a1518e8d0ccfe191bd48a1ae8641a9a25303/.github/workflows/pr.yml#L59

ST-DDT (approved):

ejcheng (approved):

PR #1848, opened 2023-02-16
feat(locale): Add locale ro_MD

botzill:
Add a new locale for ro_MD

PR #2265, opened 2023-07-19
infra: implement inactive locales content generation enhancement

xDivisionByZerox:
## Description 

This PR introduces a locale file normalization into the `generateLocales` script. This allows us to remove additional maintainace burdens by automatically checking (and fixing) for the following this in the same file:
- duplicated locale entries 
- limiting the locale file entries to 1000
- sorting the entries alphabetically

This PR also includes all to changes to all locale files which are affeected by the new rule. Honestly, I have no idea how you want to review this...

## How to review

Please follow the steps @ST-DDT suggested in https://github.com/faker-js/faker/pull/2265#issuecomment-1642849581:

> The easiest way to review these is by reviewing the script and then resetting the locales to next and run the script again.
Then there shouldn't be any diff to the committed state.
Finally, some manual tests with some of the files.
So IMO there isn't a need to review them all manually.


xDivisionByZerox:
The pipeline is failing to locales that were implemented poorly before. Look at:

https://github.com/faker-js/faker/blob/8530a3e85af15457bfcf73fc7cac6ff6fddb6ed3/src/locales/ar/date/weekday.ts#L1-L13

We simply casted the type here. What should be the expected behaviour here?

matthewmayer:
I'm not sure blanket sorting everything makes sense

You lose inherent ordering like days of the week. 

And in many languages a naive sort based on Unicode character values doesn't reflect how strings are sorted. 

Plus it causes a lot of churn meaning git blame is less useful.

ST-DDT:
IMO we should split these composite files into a folder and individual files.

xDivisionByZerox:
> IMO we should split these composite files into a folder and individual files.

That doesn't really answer the question I had. Should the generation script be able to handle invalid locales or no? Since the `DateEntryDefinition` requires the keys `wide` and `abbr` and we cast it, we currently have invalid data definitions for some locales.

ST-DDT:
How many invalid entries do we have?
If they are a select few, we should fix them instead of working around them in our code.
Even if we just do them in a non optimal way, it would probably still better than not having them at all. If we cannot find the data we can fall back to [].

matthewmayer:
> I'm not sure blanket sorting everything makes sense
> 
> You lose inherent ordering like days of the week.
> 
> And in many languages a naive sort based on Unicode character values doesn't reflect how strings are sorted.
> 
> Plus it causes a lot of churn meaning git blame is less useful.

Additionally we may be losing useful information in the definitions by resorting.

For example while implementing https://github.com/faker-js/faker/pull/1704 it turned out the first_name file actually had all the male names first, then all the female names. So it was easy to split them. If this script had already been run prior to that, there wouldn't have been any easy way to separate the female and male names.

The number of files being touched here makes it unrealistic to manually review each one and see if any useful information is being discarded.

xDivisionByZerox:
> Additionally we may be losing useful information in the definitions by resorting.

That might be. But the thing is that we already request our contributors to sort te locale entries. Some examples:

- https://github.com/faker-js/faker/pull/2232#pullrequestreview-1509378499
- https://github.com/faker-js/faker/pull/1778#pullrequestreview-1284346195
- https://github.com/faker-js/faker/pull/1848#discussion_r1108663704

We generally want this. Sure, we might lose some context but gain standardization in return.

> The number of files being touched here makes it unrealistic to manually review each one and see if any useful information is being discarded.

Agreed. But sadly this always happends if we do major changes to the locale structure. What do you suggest?

ST-DDT:
> Additionally we may be losing useful information in the definitions by resorting.

AFAIK there aren't any data in our locales that are supposed to contain any extra information by their order.
There is currently the location.direction data that use the index to distinguish between cardinal and sub-cardinal directions, but IMO that can be easily resolved by splitting the data.

> For example while implementing https://github.com/faker-js/faker/pull/1704 it turned out the first_name file actually had all the male names first, then all the female names.

Is there a file left that would still support this or all of them already split?
If they are already split, then we don't loose anything. (I'm not aware of any)
If not we can split them before the merge.

> The number of files being touched here makes it unrealistic to manually review each one and see if any useful information is being discarded.

The easiest way to review these is by reviewing the script and then resetting the locales to `next` and run the script again.
Then there shouldn't be any diff to the committed state.
Finally, some manual tests with some of the files.
So IMO there isn't a need to review them all manually.

xDivisionByZerox:
Also your comment @matthewmayer: https://github.com/faker-js/faker/pull/1823#pullrequestreview-1294840755

> - The entries should be sorted alphabetical order

I'm not saying that it is not allowed to change ones opinon, but you had this idea once as well^^

xDivisionByZerox:
> The easiest way to review these is by reviewing the script and then resetting the locales to next and run the script again.
Then there shouldn't be any diff to the committed state.

Honestly, thats super big brain. Never thought about this kind of review 👀

xDivisionByZerox:
> How many invalid entries do we have?
If they are a select few, we should fix them instead of working around them in our code.
Even if we just do them in a non optimal way, it would probably still better than not having them at all. If we cannot find the data we can fall back to [].

All that are currently fail in the CI. So 3 AFAICT.

ST-DDT:
> All that are currently fail in the CI. So 3 AFAICT.

IMO 3 are easily fixable manually.
(We should do that in a separate PR though)

matthewmayer:
I agree new entries should always be in alphabetical order unless they are something with a logical order like days of the week. I'm just concerned about subtly losing information if old entries are sorted alphabetically.

xDivisionByZerox:
Blocked by #2293

scripts/generateLocales.ts
```diff
@@ -255,14 +255,90 @@ function updateLocaleFile(filePath: string): void {
...
+      return (
+        [...new Set(localeData)]
+          // limit entries to 1k
+          .slice(0, 1000)
+          // sort entries alphabetically
+          .sort() as T
+      );
+    }
+
+    for (const key of Object.keys(localeData).sort()) {
```
ST-DDT: Is the sort here relevant? Because the keys exist already so them won't change their order.

xDivisionByZerox: You are right. I had a previous version where I copied all keys and values to a new object. That did actually sort the keys alphabetically. I did not think about this when refactoring this part.

xDivisionByZerox: The question is: Should object keys be sorted? I think theyshould, but what are your opinions.

ST-DDT: Alphabetically, same as the rest. 
On the other hand we could just use the file tree for objects that should be sorted and keep the rest as is.
As for the actual decision, I think we would need some examples where these data might reside in/what would change.

matthewmayer (changes requested):
im fine with limiting to 1000 entries and checking for duplicates but i dont really see the value of sorting existing files per https://github.com/faker-js/faker/pull/2265#issuecomment-1642832576

ST-DDT:
There are merge conflicts here.
Lets talk about how to proceed with this going forward in the next meeting.

ST-DDT (changes requested):
If I reset the `src` folder to next and reapply the changes then 13 files are different.

ST-DDT:
Wrong email?

grafik

xDivisionByZerox:
> Wrong email?

Yeah, sh*t. I did those changes yesterday evening on the wrong PC...

scripts/generateLocales.ts
ST-DDT: IMO we can remove the `sort` from here and "sort that out" later.

xDivisionByZerox: I can also implement a proper sort. I simply forgot that this suggestion was still open.

ST-DDT: I'm fine either way. Maybe sorting it in one go reduces the commit history chaos.

src/locales/da/location/direction.ts
ST-DDT, line 1: Currently this file is expected to be sorted in a specific order.

ST-DDT (changes requested):

src/locales/da/location/direction.ts
xDivisionByZerox: I created #2476 to remove this requirement.

matthewmayer:
I would like to repeat my earlier concern that a blanket sorting of all definition files is not particularly useful and there are genuine reasons for deviating from a naive sort, e.g:

* Some things have a natural order like North, East, South, West, or days or the week. It's easier to review the 12 months of a year in a foreign language if they follow the standard Jan-Dec order.
* Or perhaps the states in a country are numbered, and its easier to follow that order (e.g. if country CC speaks both languages xx_CC and yy_CC, then we can easily cross-reference that each entry in the states list matches up in both locales).
* In some languages like Norwegian letters it would be better to sort letters with diacritics like Å in the correct sort order for the locale
* In some languages like Chinese ordering by Unicode character doesn't reflect any kind of "alphabetical" order at all, it would be better to sort by stroke order or by alphabetical pinyin
* Some lists may be sorted by popularity e.g. most common 1000 surnames in a country, (if we want to reduce to 500 its better to drop the ones at the bottom)
* Some lists may come from multiple sources and we want to make that clearer in the source e.g. popular surnames from different ethnic groups in a country
* Some definitions contain numbers or roman numerals when it makes more sense to sort "9, 10, 11" not "10, 11, 9"

xDivisionByZerox:
> * Some things have a natural order like North, East, South, West, or days or the week. It's easier to review the 12 months of a year in a foreign language if they follow the standard Jan-Dec order.

I don't see the advantage of having cardinal directions sorted in a "natural" order. There is no usecase in which this information is relevant to us. I can see the argument for the month. Since those values are a fixed number it might be mor usefull to use a map instead of an array there. This would allow reviewers to make easy assumtions about localized values.

> * Or perhaps the states in a country are numbered, and its easier to follow that order (e.g. if country CC speaks both languages xx_CC and yy_CC, then we can easily cross-reference that each entry in the states list matches up in both locales).

Why can't you do that if both lists (xx_CC and yy_CC) are sorted in the same why? Although, I'm not quite sure what you mean by "states in a country are numbered". Can you elaborate on this or give an example?

> * In some languages like Norwegian letters it would be better to sort letters with diacritics like Å in the correct sort order for the locale
> * In some languages like Chinese ordering by Unicode character doesn't reflect any kind of "alphabetical" order at all, it would be better to sort by stroke order or by alphabetical pinyin

Well thats no problem. I defaulted to the normal `Array.prototype.sort()`, but I can simply provide a compare function to allow for locale compares:

```ts
const arr = ['B', 'A', 'Å'];
const sorted = arr.toSorted((a, b) => a.localeCompare(b));
console.log(sorted); // ['A', 'Å', 'B']
```

> * Some lists may be sorted by popularity e.g. most common 1000 surnames in a country, (if we want to reduce to 500 its better to drop the ones at the bottom)

That would mean we had to update the sorting of those lists every now an then. Thats a hard no for me. I can't see this argument being valid.

> * Some lists may come from multiple sources and we want to make that clearer in the source e.g. popular surnames from different ethnic groups in a country

Ok, then just split your data into multiple soures as well:

```ts
const source1 = [...];
const source2 = [...];
const source3 = [...];

export default [
  source1,
  source2,
  source3,
];
```

Where is the problem with that? 

> * Some definitions contain numbers or roman numerals when it makes more sense to sort "9, 10, 11" not "10, 11, 9"

Same as with the language aware letters. I can extend the previos compare function to be numeric aware:

```ts
const arr = ['B', 'A', 'Å', 'Å9', 'Å11']
const sorted = arr.toSorted((a,b) => a.localeCompare(b, undefined, { numeric: true }));
console.log(sorted); // ['A', 'Å', 'Å9', 'Å11', 'B']
```

Any problems with that?

---

Apologies for ignoring your previous message. I wasn't sure what you wanted to point our by your concernes. If you wanted to discuss the sorting style I don't understand why you didn't put a comment to the actual line that was doing the sort. This would have allowed for a discuss on which algorythm to use there. It would have made it more clear for me what you want to talk about exactly.

matthewmayer:
> I don't see the advantage of having cardinal directions sorted in a "natural" order. There is no usecase in which this information is relevant to us. I can see the argument for the month. Since those values are a fixed number it might be mor usefull to use a map instead of an array there. This would allow reviewers to make easy assumtions about localized values.

I'd much rather this PR be introduced one method at a time so we can review to make sure if sorting makes sense or if we are losing any natural order. This PR is too big to review.  





> 
> Why can't you do that if both lists (xx_CC and yy_CC) are sorted in the same why? Although, I'm not quite sure what you mean by "states in a country are numbered". Can you elaborate on this or give an example?

For example France departments have an numerical order 

https://en.wikipedia.org/wiki/Departments_of_France?wprov=sfti1

If country xx_CC and yy_CC are sorted seperately then the Nth item in each list no longer lines up with the corresponding translation. So if state 1 is called Aflorida in xx and ZFlorida in YY they are no longer both in index position 1.  





> 
> Ok, then just split your data into multiple soures as well:
> 

Again we need to review the existing data to see if this is needed anywhere which is hard to do with a massive PR.

ST-DDT:
**Team Decision**

We will split this into multiple smaller PRs:

1. (This PR) Add sort+unique script with every module ignored
2. Enable one module at a time to check whether there are elements in there that need sorting
3. Truncate to (shuffled) 1000k elements in all modules at once

**Considerations:**

A list of the top 1k elements is an atomic thing, instead of just replacing one element in it, you would replace the entire list with the new one.
For our usecase it is not important whether the top or the last element is chosen from that list, because it is just an entry.
Sorting makes the list somewhat easier to manage at least for latin like characters.
We can try sorting the data with the locale key and check if that might result in better sort results, but if it doesn't we will use simple sort.

matthewmayer (approved):

ST-DDT (approved):

PR #2278, opened 2023-07-26
fix(location): Pad en_US ZIP codes left to 5 characters if needed

### LINKED ISSUE #2277: Some en_US ZIP codes are only 4 characters when a state is specified
Reported at #2276

```js
const {fakerEN_US} = require("@faker-js/faker")
fakerEN_US.location.zipCodeByState("NH") //3442
fakerEN_US.location.zipCode({state:"NH"}) //3442
```

This should be left-zero-padded.

This is caused by the fact that the postcode_by_state definitions treat the ZIP codes as integers

matthewmayer:
fix #2277

Added a test and a fix. Not sure if this is the best long-term solution, but since only en_US uses this method currently, it seems a fairly low-impact solution.

Also fixed Puerto Rico range so its not always 00000


src/locales/en_US/location/postcode_by_state.ts
```diff
@@ -122,6 +125,7 @@ export default {
   NH: {
     min: 3031,
     max: 3897,
+    length: 5,
   },
   NJ: {
     min: 7001,
```
ahaberbosch: NJ is also a leading zero state according to this

src/locales/en_US/location/postcode_by_state.ts
```diff
@@ -158,6 +163,7 @@ export default {
   PR: {
     min: 0,
     max: 0,
+    length: 5,
   },
   RI: {
     min: 2801,
```
ahaberbosch: RI Also leading zero zip code

ST-DDT:
**Team decision**

This looks like a temporary workaround that is tailored for en-us only, but doesn't consider other zip code formats such as Poland's:
https://en.wikipedia.org/wiki/Postal_codes_in_Poland#Postal_codes_used_for_more_than_one_settlement
Maybe we should consider adding support for fake patterns instead.

ahaberbosch:
It would be nice if ranges in addition to patterns could be specified. Realistically, most states don't seem to utilize the full range.
For example, AZ has a gap between 85087 and 85117, and several other gaps.

ahaberbosch:
> **Team decision**
> 
> This looks like a temporary workaround that is tailored for en-us only, but doesn't consider other zip code formats such as Poland's: https://en.wikipedia.org/wiki/Postal_codes_in_Poland#Postal_codes_used_for_more_than_one_settlement Maybe we should consider adding support for fake patterns instead.

There's already support for patterns, but it seems to be exclusive to the usage of zip code by state. Could they be merged?

matthewmayer:
> **Team decision**
> 
> This looks like a temporary workaround that is tailored for en-us only, but doesn't consider other zip code formats such as Poland's: https://en.wikipedia.org/wiki/Postal_codes_in_Poland#Postal_codes_used_for_more_than_one_settlement Maybe we should consider adding support for fake patterns instead.

in many countries, postcodes/ZIP codes don't match up with state boundaries.

For example note in that Wikipedia article "Postal districts and their subdivisions are not related to the administrative division of Poland."

So I doubt there will be much usage of this outside of en_US.

ahaberbosch:
> > **Team decision**
> > This looks like a temporary workaround that is tailored for en-us only, but doesn't consider other zip code formats such as Poland's: https://en.wikipedia.org/wiki/Postal_codes_in_Poland#Postal_codes_used_for_more_than_one_settlement Maybe we should consider adding support for fake patterns instead.
> 
> in many countries, postcodes/ZIP codes don't match up with state boundaries.
> 
> For example note in that Wikipedia article "Postal districts and their subdivisions are not related to the administrative division of Poland."
> 
> So I doubt there will be much usage of this outside of en_US.

Actually, it looks like the proposed PR could be useful for Mexico, specifically the mexico City boroughs. The Mexican zip codes are modelled after US zip codes, with some mexico city zip codes prefixed with zeroes.

RobinvanderVliet:
> So I doubt there will be much usage of this outside of en_US.

I tried implementing it for the Netherlands and our provinces, but I ran into a problem as our postal codes also contain letters. Dutch postal codes are in the format `/[1-9][0-9]{3} [A-Z]{2}/`.

I found the numeric ranges for the provinces, and for the letters we could just use 2 random letters.

ahaberbosch (approved):

matthewmayer:
My feeling is while we should support this for legacy reasons for en_US we shouldn't put too much effort into trying to make this work in every locale- if this feature didn't currently exist I suspect it would be rejected as a feature request, since we generally avoid features that allow you to create consistent data between methods (eg city that matches the state).

matthewmayer:
I saw from the meeting notes you want to try to achieve this using patterns instead so it would work for more countries, but not sure how the current behavior could be replicated using string patterns as it includes numeric ranges.

ST-DDT:
> I saw from the meeting notes you want to try to achieve this using patterns instead so it would work for more countries, but not sure how the current behavior could be replicated using string patterns as it includes numeric ranges.

`0{{number.int(range)}}`

matthewmayer:
have changed this to use string patterns passed to `faker.helpers.fake`

Shinigami92:
> have changed this to use string patterns passed to `faker.helpers.fake`

Oh :astonished: that is very interesting
We should discuss this in next meeting, but looks promising to me :ok_hand:

src/definitions/location.ts
```diff
@@ -8,7 +8,7 @@ export type LocationDefinition = LocaleEntry<{
    * Postcodes patterns by state
    */
   postcode_by_state: {
-    [state: string]: { min: number; max: number; length?: number };
+    [state: string]: string;
```
ST-DDT, line 11: Should we use `string[]` for this or maybe `string | string[]`?

ST-DDT (dismissed):

src/definitions/location.ts
matthewmayer: I see no reason to do that until there's a locale that needs multiple patterns per state?

ahaberbosch: > I see no reason to do that until there's a locale that needs multiple patterns per state?

If I understand the usage, then see my example i mention above:
> It would be nice if ranges in addition to patterns could be specified. Realistically, most states don't seem to utilize the full range.
> For example, AZ has a gap between 85087 and 85117, and several other gaps.

If it were an array, then I imagine it would be an array of patterns?

matthewmayer: i think we could do that in a subsequent PR, keep this PR simple so its just a mechanical change of the old format to the new fixing bug #2277 and easy to review. Support for multiple patterns and new locales could be added in subsequent PRs.

ST-DDT:
There seems to be a merge conflict.

ST-DDT (approved):

src/definitions/location.ts
xDivisionByZerox: **Team decision**

While we see the potential of needing multiple string patterns per zipcode, we currently dont see the requirement on avioding the gaps. If you see the need for that feature please open another issue so we can discuss it individually.

xDivisionByZerox (approved):

ejcheng (approved):

PR #2280, opened 2023-07-29
fix(test): typedoc signature test issues

ST-DDT:
This PRs tries to fix the frequent issues with macOS and our signature tests.
Unfortunately, I don't have a way to test it.

I explicitly ignore the temp folder in the tsconfig to avoid issues with temp files that are detected by either concurrently running tests with typedoc or files being removed async although the should be removed sync and thus being still there when typedoc runs again on the project.


Shinigami92 (approved):

matthewmayer (approved):

scripts/apidoc/typedoc.ts
```diff
@@ -35,6 +35,7 @@ export function loadProject(
     entryPoints: ['src/index.ts'],
     pretty: true,
     cleanOutputDir: true,
+    tsconfig: 'tsconfig.build.json',
```
xDivisionByZerox, line 38: If we use the `.build` version of the configuration here, shouldn't the configuration file itself contain the exclusion and not the "base" one?

Shinigami92: No, it is okay that way
The build one extends the non build

xDivisionByZerox: I know that it does. But if you are doing OOP, you are not putting all helper functions in the parent class of only one of your children need them. Hope this comparison makes my thought more clear.

ST-DDT: This is the actual code generation part. There we exclude everything in test.

The main tsconfig is used for the tests, there the test sources are included and thus the examples needs to be excluded specifically.

Prior to the tsconfig simplification PR the typedoc tests had their own tsconfig.

Or with other words:

tsconfig.build.json -> Only for src
tsconfig.json -> For everything (including the typedoc tests and the folder to ignore)

ST-DDT: I can revert this change as it is not specifically related to the issue and more of an optimization (for performance and to allow running docs:dev if there are certain compile errors in the tests). Basically it is two different issues being fixed here but all related to our tsconfigs.

ejcheng (approved):

PR #2284, opened 2023-07-31
feat: support custom randomizer

### LINKED ISSUE #2195: Customize the PRNG used by faker

**Faker** currently offers a way to somehow customize the PRNG via the ability to define the seed (or seeds). But at some point, the seed is a limited aspect of a PRNG. For instance, on a mersenne twister (like the one used by **Faker**), the PRNG state is made of 600+ 32bits integers, while the seed can only represent one of them (actually the seed via an array somehow addresses part of the problem). In addition, mersenne twister might not be perfect in some cases: not crypto, not fast enough. So offering the ability to _twist_ it can be a good option.

My case is linked to the property based testing framework called [fast-check](https://fast-check.dev/) — _in a nutshell, it makes users able to test their code against randomized values while offering replay and shrink capabilities but does not aim to generate fake data (there are libs tailored for that need)_. In this framework, we [suggest users to pass by fake data libraries such as faker](https://fast-check.dev/docs/advanced/fake-data/#seed-based-eg-faker-jsfaker) if they want fake data (which is definitely a common need). But the current approach consisting into wrapping **Faker** by passing it a seed sounds not that perfect. Indeed, we have internally a PRNG (mersenne, xoroshiro and others) that are already seeded and offset-ed and passing it directly instead of the seed would be better.

**My question is: Is there a plan to offer the ability to customize the PRNG used by Faker?**

### Suggested solution

I think my suggestion is still a little bit drafty at the moment. But the idea was to had a `.prng` making users of **Faker** able to change the PRNG used by **Faker**. As **Faker** already offers a `.seed` and in order not to break this function or make it useless with the new one I would suggest one of the two options below:

- `.prng` receives an instance of PRNG, but if we call `.seed` we go back to the default PNRG of **Faker**
- `.prng` has to be called with an object fitting the [current schema](https://github.com/faker-js/faker/blob/e4a6bfe7bed44e5b9d6b3c386b776d293bedd956/src/internal/mersenne/mersenne.ts#L28)

From a caller point of view, changing the default random generator used by **Faker** would be done as follow:

```js
import prand from 'pure-rand';

let currentPrng = prand.xoroshiro128plus(0);
faker.prng({
  next: () => {
    return (currentPrng.unsafeNext() >>> 0) / 0x1_0000_0000.
  },
  seed: (value) => {
    const seedValue = typeof value === "number" ? value : value.reduce((a, b) => a ^ b, 0);
    currentPrng = prand.xoroshiro128plus(seedValue);
  },
});
```

_Example changing the PRNG by one coming from the library pure-rand_

### Alternative

_No response_

### Additional context

Thank you so much for this awesome library. Feel free to close the issue, if the feature is definitely not part of the plans 👍
If the feature makes sense, I can give an hand for it!

ST-DDT:
Adds the ability to Faker to use a custom prng.

- Required for #1499
- Implements #2195 

Intended use case for clone and derive:

````ts
  clone(): Faker {
    const instance = new Faker({
      locale: this.rawDefinitions,
      prng: this.prng.clone(),
    });
    instance.setDefaultRefDate(this._defaultRefDate);
    return instance;
  }

  derive(): Faker {
    const instance = new Faker({
      locale: this.rawDefinitions,
      prng: this.prng.derive(),
    });
    instance.setDefaultRefDate(this._defaultRefDate);
    return instance;
  }
````

(Regardless of whether we will actually have either of them)

Intended use case cross compatibility with other libraries other than faker:

````ts
const prng = newPRNG();
const customFaker = new Faker({ ..., prng });
const randExp = new new RandExp(/no{1,}/, { prng });

prng.setSeed(1337);
customFaker.number.int(); // 1st seed value
randExp.gen(); // 2nd seed value
customFaker.string.sample(); // 3rd seed value
````

Note: Currently faker's internal PRNG isn't exported.

### CODE REVIEW

File: src/internal/mersenne.ts [hunk 1]
```diff
@@ -323,3 +325,27 @@ export default class MersenneTwister19937 {
   }
   // These real versions are due to Isaku Wada, 2002/01/09
 }
+
+/**
+ * Generate seed based random numbers.
+ *
+ * @internal
+ */
+export function newMersennePRNG(): PRNG {
```
xDivisionByZerox: Might be me but I'd prever a name like `generateMersennePRNG` here. `newMersennePRNG` seems like you forgot the space to create a new class instance.

ST-DDT:
We have to decide whether we want to merge this in v8.x or v9.
If we merge it in v9 we can rename the option/field/type to `random`/`Random` without breaking things/causing confusion.

src/internal/mersenne.ts
ST-DDT: Changed

ST-DDT:
Blocked by #2361 as it may require changes to the Mersenne interface.

- #2361

Shinigami92:
This PR should not be set on-hold, but the other way around: #2361 / #2357 should be set on hold and related to this PR
See the last two days of discussion in #2357

src/faker.ts
```diff
@@ -205,6 +207,14 @@ export class Faker {
...
      */
     locale: LocaleDefinition | LocaleDefinition[];
+
+    /**
+     * The PRNG to use. Defaults to faker's Mersenne Twister based PRNG.
+     * Only overwrite this if you want to reuse the same PRNG in different libraries to ensure reproducible results across all of them.
+     *
+     * @default newMersennePRNG()
+     */
+    prng?: PRNG;
```
Shinigami92: Please rename this and all related fields to a non abbreviation like `seeder` or `generator`

Shinigami92 (changes requested):

ST-DDT:
> This PR should not be set on-hold, but the other way around: #2361 / #2357 should be set on hold and related to this PR See the last two days of discussion in #2357

As mentioned in the team meeting. The exposed API is dependent on the decisions we make for those issues.
I don't like merging a public API PR, when it is likely to require changes or at least a more detailed specification.

Shinigami92:
> > This PR should not be set on-hold, but the other way around: #2361 / #2357 should be set on hold and related to this PR See the last two days of discussion in #2357
> 
> As mentioned in the team meeting. The exposed API is dependent on the decisions we make for those issues. I don't like merging a public API PR, when it is likely to require changes or at least a more detailed specification.

Ah ok I see, you want to wait half a week until we all decide and write down that we agree with the `Math.random => [0.0,1.0)` contract 👌

src/faker.ts
ST-DDT: AFAICT `RandomGenerator` is the most upvoted one at this time.

- https://github.com/faker-js/faker/discussions/2308

Is there another poll where seeder/generator are ahead or a decision that I missed?

Shinigami92: The keyword was `like`, I just gave examples
It's more important for me to not use an abbreviation here

Shinigami92: @JessicaSachs, @danielroe and @Shinigami92 accepted `randomizer`

src/faker.ts
```diff
@@ -123,6 +124,8 @@ export class Faker extends SimpleFaker {
    *
    * @param options The options to use.
    * @param options.locale The locale data to use.
+   * @param options.randomizer The Randomizer to use. Defaults to faker's Mersenne Twister based pseudo random number generator.
+   * Only overwrite this if you want to reuse the same Randomizer in different libraries to ensure reproducible results across all of them.
```
Shinigami92: I feel like somehow this sentence should be changed
How about telling that it can be used to e.g. be compatible with external libs 🤔

Shinigami92 (dismissed):
IMO theoretically this is good to go

- Maybe we need some special docs in the usage guide
- Maybe randomizer should be prefixed with _ or not be a private anymore (it's marked as readonly)
- Maybe the marked doc sentence can be enhanced

but all of that is non-blocking to me

ST-DDT:
FFR: Here an example of a [pure-rand](https://github.com/dubzzz/pure-rand) `Randomizer`:

````ts
import { Randomizer } from '@faker-js/faker';
import { RandomGenerator, mersenne } from 'pure-rand';

function generatePureRandRandomizer(
  seed: number | number[] = Date.now() ^ (Math.random() * 0x100000000),
  factory: (seed: number) => RandomGenerator = mersenne
): Randomizer {
  const self = {
    next: () => (self.generator.unsafeNext() >>> 0) / 0x100000000,
    seed: (seed: number | number[]) => {
      self.generator = factory(typeof seed === 'number' ? seed : seed[0]);
    },
  } as Randomizer & { generator: RandomGenerator };
  self.seed(seed);
  return self;
}
````

File: docs/.vitepress/api-pages.ts [hunk 4]
```diff
@@ -4,6 +4,7 @@ export const apiPages = [
   { text: 'Overview', link: '/api/' },
   { text: 'Faker', link: '/api/faker.html' },
   { text: 'SimpleFaker', link: '/api/simpleFaker.html' },
+  { text: 'Randomizer', link: '/api/randomizer.html' },
```
ST-DDT: Not sure whether this should be up here or down below next to the `Utilities`.

src/faker.ts

ST-DDT (commented):
I will create a Usage/Custom Randomizer page soon

ST-DDT:
**Docs-Preview**

<details>
<summary>Click to expand</summary>

grafik
grafik
grafik

</details>

scripts/apidoc/fakerClass.ts
```diff
@@ -21,23 +21,33 @@ export async function processFakerClasses(
   return Promise.all(fakerClasses.map(processClass));
 }
 
+export async function processFakerRandomizer(
+  project: ProjectReflection
+): Promise<ModuleSummary> {
+  const randomizerClass = project
+    .getChildrenByKind(ReflectionKind.Interface)
+    .filter((clazz) => clazz.name === 'Randomizer')[0];
```
Shinigami92: Is using `find` here not more performant? Instead of traversing the full array and then pick index 0, it would traverse the array until the element is found or reached Nth element
The result should be the same here

src/randomizer.ts
```diff
@@ -0,0 +1,63 @@
...
+ *       self.state += self.step;
+ *       return (self.state %= 1);
+ *     },
+ *     seed: (seed) => {
+ *       const value = typeof seed === 'number' ? seed : seed[0];
+ *       const cleaned = value.toString().replace(/[^\d]+/g, '');
+ *       self.step = +('1.' + clean) / 2 + 0.26183095653171535;
+ *       self.state = 0.8683615301373573 * self.step + 0.24476001502354827;
+ *     },
+ *   } as Randomizer & { state: number; step: number };
```
Shinigami92:
```suggestion
 *   const self: Randomizer & { state: number; step: number } = {
 *     next: () => {
 *       self.state += self.step;
 *       return (self.state %= 1);
 *     },
 *     seed: (seed) => {
 *       const value = typeof seed === 'number' ? seed : seed[0];
 *       const cleaned = value.toString().replace(/[^\d]+/g, '');
 *       self.step = +('1.' + clean) / 2 + 0.26183095653171535;
 *       self.state = 0.8683615301373573 * self.step + 0.24476001502354827;
 *     },
 *   };
```

or 

```suggestion
 *   const self = {
 *     next: () => {
 *       self.state += self.step;
 *       return (self.state %= 1);
 *     },
 *     seed: (seed) => {
 *       const value = typeof seed === 'number' ? seed : seed[0];
 *       const cleaned = value.toString().replace(/[^\d]+/g, '');
 *       self.step = +('1.' + clean) / 2 + 0.26183095653171535;
 *       self.state = 0.8683615301373573 * self.step + 0.24476001502354827;
 *     },
 *   } satisfies Randomizer & { state: number; step: number };
```

scripts/apidoc/fakerClass.ts
ST-DDT: Is there a lint rule for that? /s

src/randomizer.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Interface for a pseudo-random number generator.
+ * This interface can be used to implement a custom randomizer using third party libraries.
+ *
+ * <p>Created instances are expected to be seeded on creation.</p>
+ *
+ * @example
```
matthewmayer: We should explain more what these example randomizers are for. I think there is a danger people will copy-paste this code without understanding that these are toy/broken examples. We can show examples of how to hook into multiple third-party libraries, and warn that in most cases you should stick to the built-in Randomizer unless you have specific needs.

matthewmayer (changes requested):

src/randomizer.ts
ST-DDT: Both result in type incompatibility errors.

src/randomizer.ts
ST-DDT: I'll adjust the related locations.

Should we explicitly mention one or two random libraries here?
Should we replace our current examples with one for e.g. pure-rand?

scripts/apidoc/fakerClass.ts
Shinigami92: We could try https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-find.md

`eslint-plugin-unicorn` is build by @fisker
He is working with me in the prettier and eslint ecosystem alot! Maybe we should consider using unicorn, I never used that plugin before, but it has official support by eslint-define-config

src/randomizer.ts
Shinigami92: You have green light from my side to mention pure-rand here ✅

scripts/apidoc/fakerClass.ts
fisker: > I never used that plugin before

Give it a try, there are lots of good stuff. 😄

src/randomizer.ts
Shinigami92: Okay, I have to check this in IDE
The current (TS-safetyness) solution does not feel right to me 😕

src/randomizer.ts
ST-DDT: Changed the example to use that.

src/randomizer.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Interface for a random number generator.
+ *
+ * **Note:** Normally there is no need to implement this interface directly,
+ * unless you want to achieve a specific goal with it.
+ *
+ * This interface enables you to use random generators from third party libraries such as [pure-rand](https://github.com/dubzzz/pure-rand).
+ * Created instances are expected to be seeded on creation.
```
matthewmayer: what does this sentence mean? expected to be seeded on creation by ... the Randomizer? the calling code that instantiates the Randomizer? Seeded with what value?

matthewmayer (changes requested):

src/randomizer.ts
```diff
@@ -0,0 +1,53 @@
...
+ * ): Randomizer {
+ *   const self = {
+ *     next: () => (self.generator.unsafeNext() >>> 0) / 0x100000000,
+ *     seed: (seed: number | number[]) => {
+ *       self.generator = factory(typeof seed === 'number' ? seed : seed[0]);
+ *     },
+ *   } as Randomizer & { generator: RandomGenerator };
+ *   self.seed(seed);
+ *   return self;
+ * }
```
matthewmayer, line 30: can we make this into a full example by then instantiating a Faker object with the newly created Randomizer?

matthewmayer (changes requested):

src/randomizer.ts
ST-DDT: How about:

```suggestion
 *
 * Instances are expected to be ready for use before being passed to any Faker constructor,
 * this includes being `seed()`ed with either a random or fixed value.
```

?

src/randomizer.ts
ST-DDT: Done

src/randomizer.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * **Note:** Normally there is no need to implement this interface directly,
+ * unless you want to achieve a specific goal with it.
+ *
+ * This interface enables you to use random generators from third party libraries such as [pure-rand](https://github.com/dubzzz/pure-rand).
+ * Created instances are expected to be seeded on creation.
+ *
+ * @example
+ * import { Randomizer, SimpleFaker } from '@faker-js/faker';
+ * import { RandomGenerator, mersenne } from 'pure-rand';
```
matthewmayer: maybe use a non-mersenne generator from pure-rand instead? e.g. xoroshiro128plus (replacing one Mersenne implementation with another doesn't seem so useful)

matthewmayer (dismissed):

src/randomizer.ts
ST-DDT: Changed

ST-DDT:
Note to self:

> I will create a Usage/Custom Randomizer page soon

Shinigami92 (dismissed):

scripts/apidoc/fakerClass.ts
Shinigami92: I'm working on it in #2417

- #2417

src/randomizer.ts
```diff
@@ -0,0 +1,57 @@
...
+ *   } as Randomizer & { generator: RandomGenerator };
+ *   self.seed(seed);
+ *   return self;
+ * }
+ *
+ * const simpleFaker = new SimpleFaker({ randomizer: generatePureRandRandomizer() });
+ */
+export interface Randomizer {
+  /**
+   * Generates a random float between `[0, 1)`.
```
xDivisionByZerox: We where discussing that some developers might not know what the "P" in "PRNG" stands for, so they might get confused if we named the faker constructor property like that. Might be a bit too far but should we consider that not every developer knows this mathmatic syntax or at least get thrown off by it? I just want to make sure that dev that use our library get the best possible DX. Tell me if you think that I'm making stuff harder than it might be.

src/randomizer.ts
```diff
@@ -0,0 +1,57 @@
...
+   * This method is called `next` so that it could be used as an [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol).
```
xDivisionByZerox: This statement as t is now is not true. If you would currently try to use an object, that satisfies the `Randomizer` interface, in a `for..of` loop you will get an Error: 
`Type 'Randomizer' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)`

Additionally, out `next()` signature is incompatible with the one of an `Iterator` as it is expected to return an object containing the keys `value` and `done`.

That being said, the original idea behind the "next" name was to align with the iterator protocol. But I#m not sure if we should mention this here for now.

src/randomizer.ts
```diff
@@ -0,0 +1,57 @@
+/**
```
xDivisionByZerox, line 1: Since this is a file that only exports an interface, should it be a `*.d.ts` file? Otherwise if would get compiled into a JS file and being included in our final bundle without actually doing anything.  Not sure how this usally works TBH.

src/randomizer.ts
matthewmayer: "between 0 (inclusive) and 1 (exclusive)."

"greater or equal to 0, and less than 1."

ST-DDT: Changed

src/randomizer.ts
ST-DDT: I removed that section.

src/randomizer.ts
ST-DDT: Well, our build tools take care of that anyway.
ESM doesn't ship it, CJS ships it as an empty module, the types are generated as expected.
I think there isn't a need to rename the file right now.
Maybe we put the default and legacy generator functions in here (or maybe even the default generator instance).
I'll update the respective PR once this one is merged.

xDivisionByZerox (dismissed):

matthewmayer:
Should we add an example of how you go from a prebuilt localised Faker instance (say fakerEN_GB) to a version with a custom randomizer? That may not be that intuitive as I think you'd have to drop the prebuilt instance and use manual fallback locales in order to access the Faker constructor.

ST-DDT:
Could you please write it in pseudocode? I'm not sure what you are asking for.

matthewmayer:
```js
// if you are currently using a prebuilt version of Faker 
import {fakerEN_GB as faker} from '@faker-js/faker';
console.log(faker.location.city())

// you will need to manually construct a new Faker instance with the correct fallback languages
// and then you can set a custom randomizer function
import { Faker, en, en_GB, base } from '@faker-js/faker';
const faker2 = new Faker({
  locale: [en_GB, en, base],
  randomizer: yourCustomRandomizerFunction
})
console.log(faker2.location.city())
```

ST-DDT:
I think adding the Faker constructor with some ommissions should be plenty. Maybe we should add a link to the constructor though.

````ts
Const faker2 = new Faker({
  locale: ...,
  randomizer: newRandomizer(),
});
````

And `@see Faker#constructor`

matthewmayer:
It's more guiding people that they cannot simply do something like 

fakerEN_GB.randomizer = ...

ST-DDT:
Here the updated screenshots for the documentation:

<details>
<summary>Click to expand</summary>

http://localhost:5173/api/randomizer.html

grafik
grafik
grafik

http://localhost:5173/guide/randomizer.html

grafik
grafik
grafik
grafik
grafik

</details>

matthewmayer (approved):

dubzzz:
From the updated documentation you sent, the API looks great. I'm looking forward to connect it 🥰

Thanks a lot for it 👍

xDivisionByZerox (approved):

dubzzz:
I want to thank you all for this awesome feature.

It unlocks a better integration of Faker within fast-check as detailed in this [article](https://fast-check.dev/blog/2024/07/18/integrating-faker-with-fast-check/).

If you read it and have some remarks, please let me know, I'll be happy to discuss and fix them.

PR #2290, opened 2023-08-05
chore(helpers): deprecate regexpStyleStringParse

### LINKED ISSUE #1985: deprecate regexpStyleStringParse
**Team decision**

- For now we don't change the existing method
- Instead we will add a new method `string.fromRegExp(string | RegExp): string`
- We will create a separate issue regarding deprecation of the old method

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/1569#issuecomment-1343328129_

ejcheng:
closes #1985


xDivisionByZerox (dismissed):

matthewmayer (changes requested):
Should we add something to the migration guide about this? Because helpers.fromRegExp is not a complete replacement for helpers.regexpStyleStringParse

For example fromRegExp can't handle numeric ranges, it only has character ranges?
```js
faker.helpers.regexpStyleStringParse("a[1-100]") // a53 etc
faker.helpers.fromRegExp("a[1-100]") //a1 or a0
```

docs/guide/upgrading.md
```diff
@@ -312,6 +312,17 @@ If you need an array of useful values, you are better off creating your own one
...
 
+### `faker.helpers.regexpStyleStringParse` deprecated in favor of `faker.helpers.fromRegExp`
+
+The `regexpStyleStringParse` method in `faker.helpers` has been deprecated. A possible replacement is `faker.helpers.fromRegExp`. However, please note that `faker.helpers.fromRegExp` is not a perfect replacement for `faker.helpers.regexpStyleStringParse`.
+
+For example, `fromRegExp` cannot handle numeric ranges:
+
+```js
+faker.helpers.regexpStyleStringParse("a[1-100]") // "a53", etc.
+faker.helpers.fromRegExp("a[1-100]") // "a1" or "a0"
```
matthewmayer: I think we need to explain instead how to get back the numeric range feature if you were previously relying on that behavior.

matthewmayer (changes requested):

src/modules/helpers/index.ts
```diff
@@ -234,7 +234,7 @@ export class HelpersModule {
   ): string {
     // default values required for calling method without arguments
 
-    string = this.regexpStyleStringParse(string); // replace [4-9] with a random number in range etc...
+    string = this.fromRegExp(string); // replace [4-9] with a random number in range etc...
```
ST-DDT: Have you checked whether any of our existing patterns uses the old logic?

src/modules/helpers/index.ts
```diff
@@ -234,7 +234,7 @@ export class HelpersModule {
   ): string {
```
ST-DDT, line 317: The description still contains ``range patterns `[4-9]` ``. IMO we should adjust that to ``regex patterns `[4-9]` ``.

ST-DDT (commented):
Is this a breaking change? Because `replaceCreditCardSymbols` behaves different for
`"[0-100]L"` inputs now.

matthewmayer:
replaceCreditCardSymbols should probably continue to use the old method until the next major version bump then.

ST-DDT:
> replaceCreditCardSymbols should probably continue to use the old method until the next major version bump then.

We don't call deprecated methods from non-deprecated ones. So we would have to delay this entire PR to v9 then.

matthewmayer:
Could soft deprecate it without formally marking it as deprecated. 

Add a comment to the description encouraging the use of the new method and referring to the migration guide. 

Then deprecate in v9 and remove in v10

ejcheng:
> Could soft deprecate it without formally marking it as deprecated. 
> 
> Add a comment to the description encouraging the use of the new method and referring to the migration guide. 
> 
> Then deprecate in v9 and remove in v10

This is a good compromise. Should we go ahead with this?

I can create an issue to "hard" deprecate this method targeted for v9, if you want.

ST-DDT:
**Team decision**

We will deprecate the method immediately and "inline" the mod in the remaining usages.
(Alternatively, using an internal helper method that isn't exported/documented)
The original method will be removed in v9.

The inlined method might be rendered obsolete by adjusting the patterns or just switching to fromRegexp in v9 or when implementing:

- https://github.com/faker-js/faker/issues/2303

src/modules/helpers/index.ts
```diff
@@ -259,8 +261,40 @@ export class HelpersModule {
...
+   * @example
+   * faker.helpers.regexpStyleStringParse() // ''
+   * faker.helpers.regexpStyleStringParse('#{5}') // '#####'
+   * faker.helpers.regexpStyleStringParse('#{2,9}') // '#######'
+   * faker.helpers.regexpStyleStringParse('[500-15000]') // '8375'
+   * faker.helpers.regexpStyleStringParse('#{3}test[1-5]') // '###test3'
+   *
+   * @since 5.0.0
+   */
+  protected regexpStyleStringParse_(string: string = ''): string {
```
ST-DDT: Maybe mark this as `@internal`?

src/modules/helpers/index.ts
matthewmayer: Maybe 4-9 is a regex pattern but say 1-100 is not. You can't do numeric ranges in regex.

test/support/seededRuns.ts
```diff
@@ -289,13 +289,18 @@ class TestGenerator<
...
       .map(([key]) => key)
       .sort();
     vi_it('should test all methods', () => {
-      expect(actual).toEqual(expected);
+      if (this.moduleName === 'helpers') {
+        actual.push('regexpStyleStringParse_');
+        expect(actual.sort()).toEqual(expected);
+      } else {
+        expect(actual).toEqual(expected);
+      }
```
ST-DDT: Is it possible to call `skip` instead?

ejcheng:
I'm pretty sure I'm using `skip` incorrectly, I can't get the tests to pass...

test/support/seededRuns.ts
```diff
@@ -289,15 +289,15 @@ class TestGenerator<
...
     const expected = Object.entries(this.module)
       .filter(([, value]) => typeof value === 'function')
       .map(([key]) => key)
       .sort();
     vi_it('should test all methods', () => {
       if (this.moduleName === 'helpers') {
-        actual.push('regexpStyleStringParse_');
-        expect(actual.sort()).toEqual(expected);
+        // @ts-expect-error: 'regexpStyleStringParse_' is not assignable to parameter of type 'MethodOf<TModule>'
+        this.skip('regexpStyleStringParse_');
```
ST-DDT: Close, I was referring to putting it to the actual spec file instead of here 😉

test/support/seededRuns.ts
```diff
@@ -295,6 +295,8 @@ class TestGenerator<
       .map(([key]) => key)
       .sort();
     vi_it('should test all methods', () => {
+      // @ts-expect-error: Argument of type 'string' is not assignable to parameter of type 'MethodOf<TModule>'.
```
ST-DDT: IMO this shouldn't tell what it hides but why.

```suggestion
      // @ts-expect-error: The method is protected and thus invisible in this scope
```

ST-DDT: Can you move this code/exclusion to the actual spec file please?

ST-DDT:
> I'm pretty sure I'm using `skip` incorrectly, I can't get the tests to pass...

I think the vitest `it` block overwrites `this`.

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -259,8 +261,42 @@ export class HelpersModule {
...
```
xDivisionByZerox: Why not use a plain function which gets a faker instance as argument? In that case we would not need the confusing test workaround.

matthewmayer: Maybe also name it something like legacyRegexpStringParse so it's clearer what it does and that it will eventually be removed

ST-DDT:
I think a plain function refers to a standalone function instead of a class function here.

src/modules/helpers/index.ts
```diff
@@ -216,7 +300,7 @@ export class HelpersModule {
   /**
    * Replaces the symbols and patterns in a credit card schema including Luhn checksum.
    *
-   * This method supports both range patterns `[4-9]` as well as the patterns used by `replaceSymbolWithNumber()`.
+   * This method supports both regex patterns `[4-9]` as well as the patterns used by `replaceSymbolWithNumber()`.
```
```suggestion
   * This method supports both range patterns `[4-9]` as well as the patterns used by `replaceSymbolWithNumber()`.
```

While this looks like regex it isnt yet.

ST-DDT (commented):
Otherwise this looks good to me. Thanks.

matthewmayer (changes requested):
Propose adding back a section into upgrading.md similar to the following:

### `faker.helpers.regexpStyleStringParse` deprecated in favor of `faker.helpers.fromRegExp`

The `regexpStyleStringParse` method in `faker.helpers` has been deprecated in Faker 8.1. A likely replacement is the more powerful `faker.helpers.fromRegExp`. 

```js
faker.helpers.regexpStyleStringParse("a{3,6}") // aaaaa
faker.helpers.fromRegExp("a{3,6}") // aaaaa
```

However, please note that `faker.helpers.fromRegExp` is not an exact replacement for `faker.helpers.regexpStyleStringParse` as `fromRegExp` cannot handle numeric ranges. This would now need to be handled seperately.

```js
faker.helpers.regexpStyleStringParse("a{3,6}[1-100]") // "aaaa53", etc.
faker.helpers.fromRegExp("a{3,6}") + faker.number.int({min:1, max:100})
```

ST-DDT (dismissed):

matthewmayer (dismissed):

src/modules/helpers/index.ts
```diff
@@ -72,6 +72,90 @@ function getRepetitionsBasedOnQuantifierParameters(
...
+  while (token != null) {
+    repetitions = parseInt(token[2]);
+    string =
+      string.slice(0, token.index) +
+      token[1].repeat(repetitions) +
+      string.slice(token.index + token[0].length);
+    token = REP_REG.exec(string);
+  }
+  // Deal with range `[min-max]` (only works with numbers for now)
+  //TODO: implement for letters e.g. [0-9a-zA-Z] etc.
```
ST-DDT: I think it is safe to remove the TODO here.

ST-DDT (dismissed):

matthewmayer (approved):

docs/guide/upgrading.md
```diff
@@ -312,6 +312,22 @@ If you need an array of useful values, you are better off creating your own one
 
 The `datetime` method previously found in `faker.datatype` has been deprecated, use `faker.date.between` or `faker.date.anytime` instead.
 
+### `faker.helpers.regexpStyleStringParse` deprecated in favor of `faker.helpers.fromRegExp`
+
+The `regexpStyleStringParse` method in `faker.helpers` has been deprecated in Faker 8.1. A likely replacement is the more powerful `faker.helpers.fromRegExp`.
+
+```js
+faker.helpers.regexpStyleStringParse('a{3,6}'); // aaaaa
+faker.helpers.fromRegExp('a{3,6}'); // aaaaa
```
Shinigami92, line 321: :rofl:

Shinigami92 (approved):

ST-DDT (approved):

PR #2298, opened 2023-08-06
chore: remove unused dependencies

ST-DDT:
Remove unused dependencies.
Not exactly sure about these ones nor that these are the only unused ones.

They might be used somewhere but `pnm run preflight` still runs without issues without them.


Shinigami92:
Did you try run `docs:dev` and such things?

ST-DDT:
> Did you try run `docs:dev` and such things?

preflight and docs:dev

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):
The react deps might be from vitepress v0? In the end, this shouldn't really matter anyway.

PR #2299, opened 2023-08-07
docs(number): remove wrong default value for float precision

mshima:

### DIFF
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -126,8 +126,6 @@ export class NumberModule {
           max?: number;
           /**
            * Precision of the generated number.
-           *
-           * @default 0.01
            */
           precision?: number;
         } = {}


src/modules/number/index.ts
```diff
@@ -126,8 +126,6 @@ export class NumberModule {
           max?: number;
           /**
            * Precision of the generated number.
-           *
-           * @default 0.01
```
xDivisionByZerox, line 130: This value gets set implicitly through the mersenne implementation. Default floating point digits are `16`. I'm not really sure what to do here TBH.

ST-DDT: The default is wrong. We don't have an explicit default. So removing it is fine IMO.

ST-DDT (approved):

xDivisionByZerox (approved):

src/modules/number/index.ts
mshima: If the default is implicit then the examples looks wrong:
```
   * @example
   * faker.number.float() // 0.5688541042618454
   * faker.number.float(3) // 2.367973240558058
   * faker.number.float({ min: -1000000 }) //-780678.849672846
   * faker.number.float({ max: 100 }) // 17.3687307164073
   * faker.number.float({ precision: 0.1 }) // 0.9
   * faker.number.float({ min: 10, max: 100, precision: 0.001 }) // 35.415
```

I haven't tested, since the examples and default value seems to conflict, I've created the PR after a quick look at source code.

mshima: After a quick test, examples looks ok.

Shinigami92 (approved):

PR #2318, opened 2023-08-14
feat(locale): improve persons and locations for zh_CN

CoolPlayLin:
# Works
- add two email addresses
- fix the name of China states
- add a lot of names, first name and last names


src/locales/zh_CN/location/state.ts
```diff
@@ -23,14 +23,14 @@ export default [
...
+  '西藏自治区',
   '浙江省',
   '江西省',
   '广东省',
   '福建省',
   '台湾省',
   '海南省',
-  '香港',
-  '澳门',
+  '香港特别行政区',
```
matthewmayer: I guess Hong Kong and Macao shouldn't be included as they have their own ISO codes and assessing schemes. There is already a en_HK locale, there could be zh_HK etc locales in future.

CoolPlayLin: @matthewmayer 
This locale is provided for developers of China mainland. In China mainland, it indeed calls this name. Is it necessary to be modified?

> I guess Hong Kong and Macao shouldn't be included as they have their own ISO codes and assessing schemes. There is already a en_HK locale, there could be zh_HK etc locales in future.

HK and Macao is also a part of China.😊

matthewmayer: For comparison en_US doesn't include Puerto Rico, en_AU doesn't include Christmas Island, etc. 

Dependencies with their own ISO code don't appear in the state.ts file for the main country locale.

CoolPlayLin: @matthewmayer 
But this is zh_CN. But in HK is zh_HK. They have difference in writing.
Example:
zh_CN：香港特别行政区
zh_HK：香港特別行政區

CoolPlayLin: @matthewmayer 
>For comparison en_US doesn't include Puerto Rico, en_AU doesn't include Christmas Island, etc.
>
>Dependencies with their own ISO code don't appear in the state.ts file for the main country locale.

IMO, this method is because they have same language. But Chinese character have Simplified and Traditional. Although their meaning is same, they have difference in writing.

matthewmayer: if necessary in future we could have zh_HK_hans and zh_HK_hant as both simplified and traditional are used there, like we have sr_RS_latin for Serbia, in Serbia, in Latin variant.

CoolPlayLin: I still don't think so. But I understand your perspective, but I believe there's a valid rationale for including Hong Kong and Macau in this file. While these regions do have their own ISO codes and localization systems, considering them as integral parts of China and sharing the same Chinese language, it could be effective to incorporate them into the main country locale file.

As you mentioned, even though Hong Kong and Macau use traditional and simplified characters respectively, the meanings remain the same; only the writing style differs. Creating separate locales for them would better cater to user needs, particularly those residing in these areas, providing software experiences that are more accurate and regionally relevant.

Considering the potential for similar cases in the future, we could adopt a structure akin to other language locales, such as using zh_HK and zh_MO to represent Hong Kong and Macau locales. This approach would not only maintain code cleanliness but also better manage language variations across different regions.

In essence, I believe that incorporating Hong Kong and Macau into this file is a more comprehensive and adaptable approach, better suited to meeting the diverse needs of users in different areas. Thank you for your understanding and thoughtful consideration.

CoolPlayLin: @matthewmayer 

I fully understand your comparison to other locales like en_US not including Puerto Rico and en_AU not including Christmas Island. It's a valid point that dependencies with their own ISO codes are typically managed separately to ensure a clear organization within the state.ts file for the main country locale.

However, the situation with the Chinese locales, intended for use within mainland China, does present a unique scenario. While Hong Kong and Macau indeed possess their own ISO codes and distinct linguistic characteristics, they still maintain a strong connection as integral parts of China. This linguistic connection stems from the shared Chinese language, albeit with variations in character writing.

Given this distinction, including zh_HK and zh_MO locales within the main country file could actually reinforce the principle of linguistic accuracy and respect the fact that these regions are part of the same nation. Unlike the examples you've mentioned, where the primary language largely remains consistent across the main locale and the localized region, the linguistic differences within the Chinese locales are more nuanced.

By accommodating these nuances, we would ensure that the software provides a more accurate experience for users in these regions. While maintaining the structure of the state.ts file is important, the inclusion of zh_HK and zh_MO could be seen as a nuanced approach that supports the diverse linguistic landscape within mainland China.

In essence, this adjustment could strike a balance between maintaining organization and providing a user-centric experience, which aligns with the primary purpose of the file. Thank you for considering the specific intricacies of this situation and how they relate to the localization needs of mainland China.

CoolPlayLin: These different points is maybe caused by cultural differences

matthewmayer: I will defer to the decision of the @faker-js/maintainers but I would suggest to avoid Faker getting into disputes about "what is a country" and "what is a state" we follow the status quo which is that country() returns a territory from https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes which includes various things which are not countries, and `state()` returns a subdivision from the subdivisions of that country unless that subdivision has its own ISO_3166 code

https://en.wikipedia.org/wiki/ISO_3166-2#Subdivisions_included_in_ISO_3166-1

So that excludes say Hong Kong from zh, Puerto Rico from en, Aland from fi, Aruba from nl.

CoolPlayLin: OK.

CoolPlayLin:
@matthewmayer

src/locales/zh_CN/location/state.ts
```diff
@@ -6,13 +6,13 @@ export default [
   '黑龙江省',
   '吉林省',
   '辽宁省',
-  '内蒙古',
+  '内蒙古自治区',
```
matthewmayer, line 9: Also not sure about these changes I understand these are the longer more formal names but Faker is typically used to fill things like a database of user accounts, drop-down list of states etc. in those use cases it seems like using the shorter name would be more common on Chinese sites. 

Eg on sina

https://sina.cn/index/location?vt=4&pos=108

image

matthewmayer (changes requested):

src/locales/zh_CN/location/state.ts
CoolPlayLin: Isn't the full name more complete and less ambiguous?

matthewmayer: I guess there's no great solution because there are really only two levels of abbreviations allowed in Faker (state and stateAbbr eg Mississippi and MS) while China has three (eg 广西壮族自治区 or 广西 or 桂)

src/locales/zh_CN/location/state.ts
```diff
@@ -6,13 +6,13 @@ export default [
...
   '河北省',
-  '新疆',
+  '新疆自治区',
```
matthewmayer: should this be 新疆维吾尔自治区 for consistency with the other autonomous regions?

matthewmayer (changes requested):

src/locales/zh_CN/person/last_name.ts
```diff
@@ -1,113 +1,1044 @@
...
+  '端木',
+  '颛孙',
+  '子车',
+  '司寇',
+  '亓官',
+  '鲜于',
+  '锺离',
+  '闾丘',
+  '公良',
+  '段干',
```
matthewmayer: We are trying to keep all locale definitions to 1000 entries max can you reduce this?

matthewmayer (changes requested):

src/locales/zh_CN/person/first_name.ts
matthewmayer, line 1: in most locales now we just have a male_first_name and female_first_name file and use a snippet like this to merge the two lists:

https://github.com/faker-js/faker/blob/next/src/locales/en_AU/person/first_name.ts

matthewmayer (changes requested):

src/locales/zh_CN/person/last_name.ts
matthewmayer: see https://github.com/faker-js/faker/issues/1802#issuecomment-1410953526

src/locales/zh_CN/location/state.ts
liulyxandy-codemao: > (eg 广西壮族自治区 or 广西 or 桂)

桂 is the abbreviation for 广西, and the full name of 广西 is 广西壮族自治区.
In daily life, Chinese people use 广西 a lot. But 广西壮族自治区 is more formal. We use it in many websites and components (e.g. http://www.gxzf.gov.cn/ or https://dwqs.github.io/react-area-linkage/ or http://datav.aliyun.com/portal/school/atlas/area_selector)

image

(↑ 广西壮族自治区 in the map)

PS: In fact, 广西 is NOT a province now. "广西省"(Guangxi province) was used more than sixty years ago.

src/locales/zh_CN/location/state.ts
liulyxandy-codemao: Yep, that should be 新疆维吾尔自治区

src/locales/zh_CN/location/state.ts
matthewmayer: @CoolPlayLin @liulyxandy-codemao i actually wonder if it would make more sense to have `stateAbbr()` return the short name (广西, 四川, 北京 etc) rather than the single-character abbreviations, which are informal and less frequently used for online use-cases?

src/locales/zh_CN/internet/free_email.ts
```diff
@@ -13,4 +13,6 @@ export default [
   'yahoo.cn',
   'yahoo.com.cn',
   'yeah.net',
+  'foxmail.com',
```
ST-DDT, line 16: This domain doesn't seem to load for me.

src/locales/zh_CN/person/first_name.ts
ST-DDT: Or is this list specifically gender neutral?

src/locales/zh_CN/internet/free_email.ts
CoolPlayLin: This email is indeed attached to qqmail. QQmail have feature of one account with multiple mailboxes. This domain is a part of mail could be added

CoolPlayLin:
@ST-DDT

src/locales/zh_CN/internet/free_email.ts
CoolPlayLin: This is the official setting page
image

CoolPlayLin:
image

src/locales/zh_CN/location/state.ts
liulyxandy-codemao: @matthewmayer in China, we seldom use the the single-character abbreviations. We only use the short name and the full name of the provinces and areas.

src/locales/zh_TW/location/default_country.ts
```diff
@@ -1 +1 @@
-export default ['Taiwan (R.O.C.)'];
+export default ['中國臺灣'];
```
matthewmayer: Whatever your opinion about the political status of Taiwan, mixing in changes to an unrelated locale is not likely to get your PR approved. Moreover the default_country file is not actually used by Faker (see #1926 ). Please revert.

matthewmayer (dismissed):

src/locales/zh_TW/location/default_country.ts
CoolPlayLin: This is according to ISO Codes from wikipedia

CoolPlayLin: image

matthewmayer: This is a PR for updating the zh_CN locale only. If you want to propose changes to the zh_TW locale, propose them in a seperate PR.

CoolPlayLin: @matthewmayer 
I remember You have said
> they have their own ISO codes and assessing schemes

It should follow the ISO code

CoolPlayLin: > This is a PR for updating the zh_CN locale only. If you want to propose changes to the zh_TW locale, propose them in a seperate PR.

OK

CoolPlayLin:
@matthewmayer

matthewmayer:
Thanks! Please avoid force-pushing branches where possible, it makes it easier for people to review the history of the PR when reviewing; and the final PR will be squashed before merging so keeping a "clean" set of commits is not important.

src/locales/zh_CN/location/state.ts
ejcheng: > I guess there's no great solution because there are really only two levels of abbreviations allowed in Faker (state and stateAbbr eg Mississippi and MS) while China has three (eg 广西壮族自治区 or 广西 or 桂)

@matthewmayer Sorry for unresolving this, but should I bring up how we want to structure `state` vs. `stateAbbr` for this locale in the next team meeting?

@liulyxandy-codemao is right about the abbreviations. For example, in this case, "内蒙古自治区" is the full official name for the Inner Mongolia Autonomous Region, but in practice, Chinese people don't use the formal name, we just say "内蒙古" (Inner Mongolia).

CoolPlayLin: @import-brain merge it?

matthewmayer: Yes would suggest discussing at next meeting.

ST-DDT:
@import-brain @matthewmayer Could you please summarize what needs to be discussed here exactly?
Otherwise it is very hard to find in the resolved review comments.

ejcheng:
> @import-brain @matthewmayer Could you please summarize what needs to be discussed here exactly? Otherwise it is very hard to find in the resolved review comments.

For province names in Chinese, there are multiple ways to write them--- 广西壮族自治区 (full formal name) or 广西 (commonly used name) or 桂 (one character abbreviation). I wanted to discuss how we should allocate these names between `state` and `stateAbbr` in the Chinese locale. The full formal name isn't commonly used, so I was thinking we could have the commonly used names in `state` and the one character abbreviations in `stateAbbr`, what do you think?

CoolPlayLin:
@import-brain The full name of province is really common in address. In address, it should more detailed. But I think we the 广西 is better than 桂. The abbr with two character is more known in China. This is because some of province even have more than one official abbr with one character. For instance, 云南省(Yunnan Province) could call abbr with one character as 云 or 滇. If we use such abbr, it would cause more issue and trouble need to resolve.
All in all, I recommand not use abbr with one character or come up with a way to resolve it.
CC@matthewmayer @ST-DDT

matthewmayer:
Right 

Option A is "formal name in state, common name in stateAbbr"

Option B is "common name in state, one character name in stateAbbr"

Option C would be modifying the methods to allow for three different possible values.

CoolPlayLin:
IMO, If it would use for generate address , it must use full name of province

CC @matthewmayer @ST-DDT @import-brain

ST-DDT:
**Team notice**

We are currently internally investigating the possibilities.

ST-DDT:
**Team decision**

Short term (now):

We will use the full name for state and the common name for state_abbr

Long term (v9):

We will change the state method to return a complex object: `{ fullName, commonName, abbreviation }` and let the user take one of those or whatever is there.

matthewmayer:
@CoolPlayLin please modify the state and state_abbr definitions so state returns the official name e.g.  '广西壮族自治区', and state_abbr returns the short name '广西'. The one-character abbrevations like 桂 will no longer be used.

matthewmayer (approved):

ejcheng (approved):

ST-DDT (approved):

PR #2322, opened 2023-08-17
docs: mention file limits for locale PRs

ST-DDT:
When adding or changing locales, some PRs get huge and thus hard to review.
I would like to add a rule that requests PRs to stay limited in size.


CONTRIBUTING.md
```diff
@@ -42,6 +42,11 @@ If adding new data definitions to Faker, you'll often need to find source data.
 - But if you are compiling a list of, for example, popular personal names or cities, don't copy directly from a single source (Wikipedia, 'most popular' articles, government data sites etc). A compilation of facts [can be copyrighted](https://en.wikipedia.org/wiki/Copyright_in_compilation).
 - It's best to refer to multiple sources and use your own judgement/knowledge to make a sample list of data.
 
+## Adding new locale or updating existing one
```
xDivisionByZerox, line 45: I feel like this could be combined with the previous secion `sourcing data for definitions`. What do you think?

ST-DDT: For me the section above is more about legal stuff instead of a how to. My original goal was to have something to link to. But I dont have strong feelings. Should I remove the header?

xDivisionByZerox: For now, it's fine, I think. I totally get your point of having a link available

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2323, opened 2023-08-18
refactor(internet): remove avatar_uri definitions

matthewmayer:
Removes the unused src/locales/en/internet/avatar_uri.ts file, approx 27KB of definitions

Arguably a breaking change but given that
- its only accessible via faker.helpers.fake
- it seems to reference filenames on a server which no longer exists (e.g. 
see 
https://github.com/coding-in-public/blog-post-filter/blob/main/posts.json 
)

Could probably be done in a minor rather than a major release?


ST-DDT:
Not sure about these ones.
IMO these look like filenames to me.
On the other hand we dont use them. 🤔

matthewmayer:
Could replace the whole file with a single entry with minimal file size like

`['this_file_is_deprecated.jpg']`

So in the unlikely event anyone is using this they get a warning. 

Then remove completely in 9.0

ST-DDT:
My concern is not about how to remove it, but whether the names in that list are useful by itself.

matthewmayer:
They don't seem very useful. faker.internet.username or faker.system.filename are more useful and properly documented

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2321, opened 2023-08-16
feat(locale): add da locale

stonor:

PR #2324, opened 2023-08-18
feat(locale): add minimal da locale, start with location

stonor:
As suggested in #2321, add new language in smaller chunks.


src/locales/da/location/city_name.ts
ST-DDT, line 1: Please sort these name lists alphabetically (except direction).

stonor: Done

matthewmayer (commented):
Should we consider making this locale da_DK as Danish is spoken in some other territories with their own ISO codes like Greenland or Faroe Islands?

ST-DDT:
> Should we consider making this locale da_DK as Danish is spoken in some other territories with their own ISO codes like Greenland or Faroe Islands?

No, because these other territories probably share a lot of the names/values with Denmark and we shouldn't duplicate them.
If there are da_DK specific things then we can still split or re-arrange them then.

ST-DDT (approved):

matthewmayer:
I'm not sure when we do lang vs lang_COUNTRY though. Eg ka_GE and sr_RS were added recently

ejcheng (approved):

matthewmayer (approved):

Shinigami92 (approved):

PR #2325, opened 2023-08-18
feat(location): Support ISO 3166-1 numeric country codes

### LINKED ISSUE #2302: Support ISO 3166-1 numeric as variant for the country code API

Currently only 'alpha-2' (two letter code) and 'alpha-3' (three letter code) are supported. In many cases, the [`numeric`](https://en.wikipedia.org/wiki/ISO_3166-1_numeric) representation (three digit code) is needed. 

### Suggested solution

Support `numeric` as a variant so the API can return the three-digit country code.

### Alternative

_No response_

### Additional context

_No response_

RobinvanderVliet:
This merge request resolves issue #2302.


src/modules/location/index.ts
```diff
@@ -320,37 +320,45 @@ export class LocationModule {
...
     }
 
     const { variant = 'alpha-2' } = options;
-    const key = variant === 'alpha-3' ? 'alpha3' : 'alpha2';
+    const key =
+      variant === 'alpha-3'
+        ? 'alpha3'
+        : variant === 'numeric'
+        ? 'numeric'
+        : 'alpha2';
```
suhaibmujahid: Nit:
```suggestion
    const key = variant.replace('-', '')
```

suhaibmujahid (approved):
Thank you, @RobinvanderVliet! LGTM.

src/modules/location/index.ts
matthewmayer: That doesn't do the same thing. The original defaults to alpha2 if an unknown variant is passed

suhaibmujahid: Yes, but is that behaviour intended? I guess it would be better to fail if the variant was misspelled. But yes, it could be  problematic if its value is `null` since it will not fallback to the default value in that case.

RobinvanderVliet: I think this behavior is intended, I see a similar pattern in other parts of the code.

suhaibmujahid: Thank you for the clarification, @RobinvanderVliet!

src/modules/location/index.ts
```diff
@@ -320,37 +320,45 @@ export class LocationModule {
    * Returns a random [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) country code.
    *
    * @param options The code to return or an options object. Defaults to `{}`.
-   * @param options.variant The variant to return. Can be either `'alpha-2'` (two-letter code)
-   * or `'alpha-3'` (three-letter code). Defaults to `'alpha-2'`.
+   * @param options.variant The variant to return. Can be either `'alpha-2'` (two-letter code),
+   * `'alpha-3'` (three-letter code) or `'numeric'` (numeric code). Defaults to `'alpha-2'`.
```
xDivisionByZerox: Should we use a list here, as it might be less difficult to read?

```suggestion
   * @param options.variant The variant to return. Can be one of:
   * - `'alpha-2'` (two-letter code)
   * - `'alpha-3'` (three-letter code)
   * - `'numeric'` (numeric code)
   * Defaults to `'alpha-2'`.
```

src/modules/location/index.ts
xDivisionByZerox: Might be me, but having an IIFE switch here would improve the readability:

 ```suggestion
    const key = (() => {
      switch(variant) {
        case 'numeric': 
          return 'numeric';
        case 'alpha-3': 
          return 'alpha3';
        case 'alpha-2': 
        default: 
          return 'alpha2';
      }
    })();
```

Absolute best would be to extract the IIFE into a function over the class (we are not able to use private functions) and use it. That way we could apply a nice name additionally.

Comment (ST-DDT) [reply src/modules/location/index.ts]: It is only used in a single place. Not sure whether we need an external method for that. (IIFE but not an external fn)

Comment (matthewmayer) [reply src/modules/location/index.ts]: The switch version is nicer. Nested ternaries are very hard to read.

Comment (Shinigami92) [reply src/modules/location/index.ts]: A more easier solution would be:

```suggestion
    const key = {
      'alpha-3': 'alpha3',
      'numeric': 'numeric',
    }[variant] ?? 'alpha2';
```

ST-DDT:
There seems to be some issues with the new implementation.

matthewmayer:
I don't think you can do that kind of object mapping in Typescript. Probably use the switch version.

Shinigami92:
> I don't think you can do that kind of object mapping in Typescript. Probably use the switch version.

No, just add the `alpha2` case, and maybe also a `as const`

matthewmayer:
i think the switch version is much nicer to read and makes it clear what the default is
```
const key = (() => {
     switch(variant) {
       case 'numeric': 
         return 'numeric';
       case 'alpha-3': 
         return 'alpha3';
       case 'alpha-2': 
       default: 
         return 'alpha2';
     }
   })();
```

xDivisionByZerox (dismissed):

src/modules/location/index.ts
```diff
@@ -320,37 +320,53 @@ export class LocationModule {
    * Returns a random [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) country code.
    *
    * @param options The code to return or an options object. Defaults to `{}`.
-   * @param options.variant The variant to return. Can be either `'alpha-2'` (two-letter code)
-   * or `'alpha-3'` (three-letter code). Defaults to `'alpha-2'`.
+   * @param options.variant The variant to return. Can be one of:
+   * - `'alpha-2'` (two-letter code)
+   * - `'alpha-3'` (three-letter code)
+   * - `'numeric'` (numeric code)
+   * Defaults to `'alpha-2'`.
```
```suggestion
   * @param options.variant The variant to return. Can be one of:
   *
   * - `'alpha-2'` (two-letter code)
   * - `'alpha-3'` (three-letter code)
   * - `'numeric'` (numeric code)
   *
   * Defaults to `'alpha-2'`.
```

Otherwise it will be displayed as:

grafik

matthewmayer (approved):

ST-DDT (approved):

PR #2339, opened 2023-08-24
feat(locales): add hacker abbreviations

jamesbuch:
Add a few more hacker abbreviations.


jamesbuch:
The abbreviations are included in PR #2340

PR #2348, opened 2023-08-26
chore: ignore dotfiles in generate locales

matthewmayer:
On MacOS, the system often creates .DS_Store files when navigating around

When `pnpm run generate:locales` is run this generates invalid locale index files e.g.
```
  26 | import vehicle from './vehicle';
  27 | import word from './word';
> 28 | import .DS_Store from './.DS_Store';
```

This modifies the generateLocales.ts script to ignore files that start with a .


scripts/generateLocales.ts
```diff
@@ -162,6 +162,7 @@ async function generateLocalesIndexFile(
   depth: number
 ): Promise<void> {
   let modules = readdirSync(path);
+  modules = modules.filter((file) => !file.startsWith('.'));
```
Shinigami92, line 165: How about only allow `*.ts`? :thinking:

matthewmayer: we could simplify lines 164-167, removing the helper functions to:
```
const modules = readdirSync(path)
    .filter((file) => file.endsWith('.ts') && file !== 'index.ts')
    .map((file) => file.replace('.ts', ''));
```

Shinigami92: Sounds really good to me :slightly_smiling_face:  
@ST-DDT should have the last word, because he is the core-author of these files

matthewmayer: haha i probably should have run the code first, its actually subfolders like "company" here, so actually none of them end in *.ts

Shinigami92: ah okay, I actually just trusted your suggested code, and because "file" was named "file" I expected that you tested it already

matthewmayer: This function is caused recursively, so it readdirSync may for example return:

```
[
  'animal',      'app',
  'cell_phone',  'color',
  'commerce',    'date',
  'finance',     'index.ts',
  'location',    'lorem',
  'metadata.ts', 'music',
  'person',      'team',
  'vehicle'
]
```

at the root level  or
```
[
  'female_first_name.ts',
  'female_last_name.ts',
  'female_last_name_pattern.ts',
  'female_middle_name.ts',
  'female_prefix.ts',
  'first_name.ts',
  'index.ts',
  'male_first_name.ts',
  'male_last_name.ts',
  'male_last_name_pattern.ts',
  'male_middle_name.ts',
  'male_prefix.ts',
  'name.ts',
  'prefix.ts',
  'suffix.ts',
  'title.ts'
]
```

at the second level. so it needs to allow for both *.ts files and directories

Shinigami92: This is not actually "src" code, but script
I suggest instead of using `readdirSync` we could use `glob`, which might already be used in the bundle script.
That way we could glob-search for `**/*.ts` :thinking:  
Maybe this makes much more easy, does it?

matthewmayer: i think that would require a much bigger rewrite of this script. For now i think the one-line original fix may be simpler.

Shinigami92: Oh my god, well I debugged what `let modules = readdirSync(path);` actually returns and it is far from what I original expected... That is not really well readable code :confused: 

So yeah, back to your suggested initial code... I approve...

Maybe I should discuss later on with @ST-DDT to fully rewrite this section and optimize it, because right now, you really need to run the code to actually understand what's going on

Shinigami92 (approved):

scripts/generateLocales.ts
matthewmayer: yeah readdirSync is not very intuituve. Its a bit better if you add {withFileTypes: true} because then it returns dirent objects instead of just plain filenames/directory names, and you can do things like filter on dirent.isDirectory()

But that's definitely for another day.

ST-DDT: Are you referring to readdirSync in this method or generally?
IMO the code is acceptable to read for what it does. 
But I wrote it so I'm probably biased.
The hard part not being reading the files but the import and code generation stuff.
IMO in this method particulary it is not important whether something is a file or folder.
Feel free to suggest a new implementation in a new PR so that we can decide on the best variant.

matthewmayer: Honestly i think its fine, its not the most beautiful code in the word but it does the job; i would only bother to rewrite it if we were significantly changing the way locales were structured.

ST-DDT (approved):

scripts/generateLocales.ts
Shinigami92: I'm also fine with it :ok_hand:  
It is just the concern fighting the symptoms and not the cause

PR #2356, opened 2023-08-29
chore(finance): remove defunct credit card issuers

### LINKED ISSUE #2344: Some credit card issuers are defunct
https://fakerjs.dev/api/finance.html#creditcardnumber

Various issuers found in the en locale are defunct:

https://en.wikipedia.org/wiki/Solo_(debit_card)
https://en.wikipedia.org/wiki/Switch_(debit_card)
https://en.wikipedia.org/wiki/Laser_(debit_card)
https://www.crunchbase.com/organization/instapayment-inc

https://github.com/faker-js/faker/blob/next/src/locales/en/finance/credit_card/index.ts

These should probably be removed, but this is a breaking change so defer to 9.0

matthewmayer:
fix #2344


test/modules/finance.spec.ts
```diff
@@ -472,16 +472,8 @@ describe('finance', () => {
           expect(diners_club).toSatisfy(luhnCheck);
           const jcb = faker.finance.creditCardNumber('jcb');
           expect(jcb).toSatisfy(luhnCheck);
-          const switchC = faker.finance.creditCardNumber('mastercard');
```
matthewmayer, line 475: note this test was incorrect anyway

ST-DDT (approved):
Are these data only in en or do other locales have them as well?

matthewmayer:
Only en

ejcheng (approved):

PR #2357, opened 2023-08-29
feat!: high precision random number generator

### LINKED ISSUE #2355: faker.datatype.number often gives duplicates
### Pre-Checks

### Describe the bug

* In unit tests running on our CI environment (and occasionally on our macs), the following code has been observed to throw a "duplicate error"---where it cannot find a unique integer
* `fakeId` is called 50--100 times in total on the offending test runs. 
* the rate of failure on CI is around 1 in 1000 runs
* haven't done the math but it is still very surprising because we are generating up the "max safe integer"... which is equal to 2^53 – 1... a very, very large number, so I don't expect 100 generations to render duplicates

```
export const fakeId = () =>
  faker.unique(() => faker.datatype.number({ max: Number.MAX_SAFE_INTEGER }), undefined, { maxTime: 100 }).toString();
```

## Analysis

* The mersenne implementation is very sus---why use this at all? For the most part, fakerjs seems to be a *test* helper, so IMO, this is over engineering. We can likely just use Math.random under the hood
* Additionally, the implementation of `unique` is also sus on CI environments. Namely, this block:
https://github.com/faker-js/faker/blob/next/src/modules/helpers/unique.ts#L114-L122
* Most likely what's happening is the following---in our CI environments, we have a known issue with slowness. The mersenne implementation will unnecessarily return duplicates. And when both happen, `now - startTime` will be greater than `maxTime`, resulting in the thrown error

Suggestion:

* A workaround we can do on our end is pass a very large `maxTime`.
* but `unique` should rely on `maxRetries` only
* faker should have an option to opt-out of using mersenne

### Minimal reproduction code

* see above

### Additional Context

_No response_

### Environment Info

```shell
OS: linux/macosx
Node: 16
faker: 7.3.0
```

- [ ] CJS
- [ ] ESM

### Used Package Manager

npm

### LINKED ISSUE #2681: Export the default randomizer

Injecting the same randomizer to different Faker instances seems to become a common use case, especially with multilingual datasets.

Hovewer, the default randomizer factory is not exported from the package and therefore cannot easily be used.

### Suggested solution

It would be really helpful if the method `generateMersenne32Randomizer` was exported, maybe using an alias.

The [documentation](https://fakerjs.dev/guide/randomizer.html) could be updated as follows:

```typescript
import { en, Faker, Randomizer, zh_TW, generateDefaultRandomizer } from '@faker-js/faker';

const randomizer: Randomizer = generateDefaultRandomizer()

const customFakerEN = new Faker({
  locale: en,
  randomizer,
});

const customFakerZH_TW = new Faker({
  locale: [zh_TW, en],
  randomizer,
});

randomizer.seed(5);
// customFakerEN.seed(5); // Redundant
// customFakerZH_TW.seed(5); // Redundant

const firstName = fakerZH_TW.person.firstName(); // 炫明
const alias = fakerEN.person.firstName(); // John (different from before, because it is now the second call)
```

ST-DDT:
Fixes #2355

- #2355

## To Fix

- [x] fix/reduce duplicates returned from `faker.number.int()` and `faker.number.float()`
- [x] fix faker.number.int() only returning even numbers (31bits out of 53)
- [x] fix faker.number.int(max) is incapable of returning max value for large maxes


matthewmayer:
Probably also worth checking if genrandRes53CC is significantly slower, or if there are any other downsides to using it.

ST-DDT:
It's twice as slow.

matthewmayer:
I think this should be probably considered a breaking change given it causes nearly every snapshot to change. That would also allow for beta testing of the change before 9.0 in case the slower algorithm causes performance problems.

ST-DDT:
The breaking of the snapshots is caused by `next()` now effectively taking 2 values from the seed.
We could reduce this if we dynamically request either 32/53 bits of precision depending on the values in use.
e.g. an array with length 15 doesn't need 53 bits of precision for the index.

ST-DDT:
I'm not that deep into math, but for some reason using two values from the seed increases the variance of the values.
Which increase likelihood of this test to fail: https://github.com/faker-js/faker/actions/runs/6025687970/job/16346941314?pr=2357

ST-DDT:
> I'm not that deep into math, but for some reason using two values from the seed increases the variance of the values. Which increase likelihood of this test to fail: [faker-js/faker/actions/runs/6025687970/job/16346941314?pr=2357](https://github.com/faker-js/faker/actions/runs/6025687970/job/16346941314?pr=2357)

Nevermind, I found the error, currently the first and the last element aren't evenly distributed for number.int()

ST-DDT:
**Team Decision**

We will change mersenne to 53 bit precision to reduce the duplicates and create a separate PR for the float issues.

test/internal/mersenne/twister.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      const twister = newTwister(42);
+      const actual = twister.genrandInt32();
+      expect(actual).toBe(0);
+    });
+
+    it('should be able to return 2^32-1', () => {
+      const twister = newTwister(1350757185267805.8);
+      const actual = twister.genrandInt32();
+      expect(actual).toBe(2 ** 32 - 1);
+    });
```
ST-DDT: This is my favorite test in this PR. 😉 
It only took 80 CPU hours to find that seed...

Currently smallest known return value `20` for seed `147377213`.

test/internal/mersenne/twister.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      const actual = twister.genrandRes53CO();
+      expect(actual).toBe(0);
+    });
+
+    it('should be able to return 1', () => {
+      const twister = newTwister();
+      twister.genrandInt53 = () => 2 ** 53 - 1;
+      const actual = twister.genrandRes53CO();
+      expect(actual).not.toBe(1);
+      expect(actual).toBeGreaterThan(0.9999999999);
```
ST-DDT: Should we check for this instead?

```suggestion
      expect(actual).toBe(0.9999999999999999);
```

ST-DDT:
Reduced to only cover the most minimal required changes.

test/modules/date.spec.ts
```diff
@@ -801,22 +801,22 @@ describe('date', () => {
       faker.setDefaultRefDate(() => new Date(Date.UTC(2020, 0, 1)));
       faker.seed(20200101);
       const date = faker.date.past();
-      expect(date).toEqual(new Date('2019-02-25T21:52:41.824Z'));
+      expect(date).toMatchInlineSnapshot('2019-02-25T21:52:41.819Z');
```
Shinigami92, line 834: I might have oversee something, but through these `toMatchInlineSnapshot` changes, we lost the check whether date is a string or actually a Date instance

ST-DDT: That's not what the test is about.

xDivisionByZerox: Then we can simply extend the test:
```suggestion
      expect(date).toBeInstanceOf(Date);
      expect(date).toMatchInlineSnapshot('2019-02-25T21:52:41.819Z');
```

test/modules/helpers.spec.ts
```diff
@@ -1255,6 +1247,24 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`)
     }
   );
 
+  describe('uniqueArray', () => {
+    it('works as expected when seeded', () => {
+      const input = ['a', 'a', 'a', 'a', 'a', 'f', 'g', 'h', 'i', 'j'];
+      const length = 5;
+      faker.seed(100);
+      const unique = faker.helpers.uniqueArray(input, length);
+      expect(unique).toMatchInlineSnapshot(`
```
xDivisionByZerox: This should now be `toStrictEqual` again, no?

ST-DDT: I'm not sure what you refer to.

test/modules/date.spec.ts
```diff
@@ -801,22 +801,23 @@ describe('date', () => {
       faker.setDefaultRefDate(() => new Date(Date.UTC(2020, 0, 1)));
       faker.seed(20200101);
       const date = faker.date.past();
-      expect(date).toEqual(new Date('2019-02-25T21:52:41.824Z'));
+      expect(date).toBeInstanceOf(Date);
+      expect(date).toMatchInlineSnapshot('2019-02-25T21:52:41.819Z');
```
xDivisionByZerox, line 834: Why do we use inline snapshots here if all other snapshots have their own file? Is there an advantage for one or the other?

test/modules/helpers.spec.ts
xDivisionByZerox: I was questioning why we need a snapshot test here since we set the seed manually. This means that the expected value should always stay as it is now. That's why I thought we should use `toStrictEqual` instead of a snapshot test here. Unless I'm missing something.

test/modules/date.spec.ts
ST-DDT: As you can see the value changes whenever we work on the mersenne definitions. But I can change it back if you would like.

I used inline snapshots here to make it a self contained test. The seeded tests (amd their snapshots) are a wholy different type of test IMO.

xDivisionByZerox: The reason why I don't want a snapshot is that it will automatically get updated with all other snapshots. That would lead to a reviewer being more likely to skip over this specific one if it will get updated for some unknown reason. That's at least what I think. Am I overthrowing thus?

test/modules/helpers.spec.ts
ST-DDT: Thats another reason why I used inline snapshots: to have it more obvious something changed because it is in the test file and not part of the huge snapshot file.

matthewmayer:
Could we still allow access to the original 32-bit Mersenne  generator via the new #2195 feature? Then if someone prefers the original implementation (e.g. if speed is more importrant than occasional duplicates) they still have that option.

ST-DDT:
> Could we still allow access to the original 32-bit Mersenne generator via the new #2195 feature? Then if someone prefers the original implementation (e.g. if speed is more importrant than occasional duplicates) they still have that option.

Could you please describe the usecase for/idea behind this proposal?
We already have a guess why you suggest that, but we would like to know whether we might have forgotten something.

If we go that route, then this would be blocked by #2284 which is blocked by #2361.

matthewmayer:
Speed. You said this implementation is twice as slow? So if I'm generating millions of fake names I don't want it to be slower than before.

dubzzz:
> twice as slow

Why not backing the library by an official prng generator tailored for performance like: [pure-rand](https://www.npmjs.com/package/pure-rand)? It would probably allow the faker team to rather focus on generators themselves while keeping the optimization of random number generators into other libraries (this one is used by Jest). **I'm the author of this library**

_You can also choose to copy the implem from the lib I just shared but it sounds less sustainable. The implem is much faster than the one in faker (from my benchs)_

Cc @ST-DDT

ST-DDT:
@dubzzz A disclaimer that you are a/the maintainer of that library would have been nice.

> You can also choose to copy the implem from the lib I just shared but it sounds less sustainable. 

Copying a well maintained lib doesn't sound like a good idea to me either.

> The implem is much faster than the one in faker (from my benchs)

Does your implementation provide the full 53 bits of precision possible with JS?

dubzzz:
> A disclaimer that you are a/the maintainer of that library would have been nice.

Sorry, I forgot to underline it. There are many others if you want.

> Does your implementation provide the full 53 bits of precision possible with JS?

Not really. The generator itself generates 32bits values in an uniform way. But I do have helpers to generate values in an uniform way on larger ranges (see `unsafeUniformIntDistribution` for 53bits and the ones with ArrayInt in the name for above).

ST-DDT:
> > A disclaimer that you are a/the maintainer of that library would have been nice.
> 
> Sorry, I forgot to underline it. There are many others if you want.

No problem. We will have to discuss the choice of library internally anyway (if we go that route).

matthewmayer:
While it looks like a very nice library it's also maintained by one person. Given the history of the project Faker deliberately has many maintainers (and currently no dependencies).

ST-DDT:
@dubzzz Do you have a test checking that the default implementation can return `0x00000000` and `0xffffffff`respectively?

e.g. https://github.com/faker-js/faker/blob/a73653436dfbaf4a005dfc6913ecf0668b134c1d/test/internal/mersenne/twister.spec.ts#L19-L23

EDIT: I'm currently searching for a seed that results in 0 for our implementation.

Min: 3 - Seed: 994350336
Max: * - Seed: 726175548

dubzzz:
I don't think so. But I have test checking that it outputs the exact same set of values as the implementation of Mersenne Twister of the C++ standard and other tests. I can check the min/max.

Shinigami92:
Speaking from my absolutely personal perspective:
I've looked through your profile and I like what I see :slightly_smiling_face: 
[pure-rand](https://github.com/dubzzz/pure-rand) also do not have sub-dependencies (which is a good thing :+1:)
It is purely naively written in TypeScript :sparkles: 
So the only "bad" thing is that you are a one-man army (https://github.com/dubzzz/pure-rand/graphs/contributors) and Faker and its history ... anyways ... :sweat_smile: 

But I have one question: @dubzzz why did you have such a spike suddenly in your downloads?

dubzzz:
> why did you have such a spike suddenly in your downloads?

@Shinigami92 The spike is due to Jest, they started to use pure-rand internally. At some point in time they have been looking for a library to generate random numbers. So I suggested to rely on pure-rand, as I contributed several times to the project and as they have been using my other library fast-check for quite some time to test some of their internals they went for pure-rand. Actually when I suggested the library I also checked the alternatives in both maintenance level and download rates and the one I found were a bit below.

> While it looks like a very nice library it's also maintained by one person. Given the history of the project Faker deliberately has many maintainers (and currently no dependencies).

> Speaking from my absolutely personal perspective: I've looked through your profile and I like what I see 🙂

@matthewmayer @Shinigami92 Honestly, it was mostly an idea: choice is definitely fully yours.

For the record, my only query was to have a way to connect an external pnrg into faker as described into https://github.com/faker-js/faker/issues/2195. The initial query is still opened whatever your choice as it may unlock some usages I can have of faker: I want to provide a plugin fast-check (my other lib) to faker, which will be able to use faker, but to do so I'll have to be able to customize the instance of prng being used by faker.

That's said, let me know if I can help.

> A disclaimer that you are a/the maintainer of that library would have been nice.

And once again, sorry @ST-DDT. By the way I updated my original message to make it clearer.

matthewmayer:
#2195 still seems like the best way to handle this. We could have a page in the docs explaining how to plug in pure-rand (or other popular libraries) to make it very easy to use without adding a dependency.

ST-DDT:
What is left to do here? AFAICT nothing.
Is this going to be v8 or v9?

matthewmayer:
Changing the default random generator should definitely be v9. Allowing opt in to a 53 bit mersenne generator using #2195 could be in v8.

Shinigami92:
> Changing the default random generator should definitely be v9. Allowing opt in to a 53 bit mersenne generator using #2195 could be in v8. 

I slightly agree, and I would also prio it like that.

In last team meet we decided that the `generator.next()` function should fulfill `Math.random()`'s `[0.0,1.0)`, and IMO this should be documented and added first before moving on and fixing this bug, due to I assume we then would need additional tests here in this PR.

ST-DDT:
> In last team meet we decided that the `generator.next()` function should fulfill `Math.random()`'s `[0.0,1.0)`

IIRC we didn't. At least the meeting notes dont contain a decision for that.

Shinigami92:
⚠️ something horrible came to my mind tonight ⚠️

I would like to kindly ask if we first could prioritize the feature to provide a custom `seeder` / `generator`, because Faker is not only used for tests, but also generating e.g. seeding data to a database instance with e.g. `faker.string.uuid()` while relying on a constant defined seed value!
So if we would just merge this PR and do not provide the functionality to switch manually back to v8.0 behavior, it would not be possible anymore at all to recreate that seeded data.

So either this PR lands in v9 as a communicated breaking feature or we first merge / prioritize the customizable `seeder` / `generator`

ST-DDT:
Sure, lets talk about the specifics in the next team meeting.

Please keep in mind that we currently do not provide a guarantee to generate the same values on updates.

Shinigami92:
> Sure, lets talk about the specifics in the next team meeting.
> 
> 
> 
> Please keep in mind that we currently do not provide a guarantee to generate the same values on updates.

While this is true, it mostly just affected localized stuff and not fundamentally changed the ground.
I for example do not rely on generating localized values in production, but use `faker.string.uuid` or `faker.helpers.arrayElements` with own defined data for populate production database.
The more I think about it, the more I think that my approach of doing so is insecure and I should try to do backup and handmade migrations instead of relying on a seed value 🤔

matthewmayer:
Agreed. Changing the mersenne implementation causes every faker method to return different values for a fixed seed so it should definitely be considered a breaking change.

matthewmayer:
> Please keep in mind that we currently do not provide a guarantee to generate the same values on updates.

True but I think there's a big difference between "we added a few new dog breeds so now faker.animal.dog returns different values" vs "every method now returns different values"

ST-DDT:
Ready for review.

Shinigami92 (dismissed):

docs/guide/upgrading_v9/2357.md
```diff
@@ -0,0 +1,34 @@
+# Use High Precision RNG by default
+
+In v9 we switch from a 32 bit random value to a 53 bit random value.
+We don't change the underlying algorithm much, but we now consume two seed values each step instead of one.
+This affects generated values in three ways:
+
+- In large lists or long numbers the values are spread more evenly.
+  This also reduces the number of duplicates it generates.
+  For `faker.number.int()` this reduces the duplicates from `1/10k` to less than `1/8kk`.
```
matthewmayer: kk is not a common abbreviation. Million?

docs/guide/upgrading_v9/2357.md
```diff
@@ -0,0 +1,34 @@
...
+- Some seeded runs now return slightly different values, due to the last digits of random numbers changing,
+  some methods now generate slightly different results.
+- Subsequent values change their values. Because the generator now takes two values each,
+  the values afterwards are different from before.
+  If then generator previously generated `1,2,3,4,5`, it now generates `1,3,5,7,9`.
```
matthewmayer: This seems a weird example. It doesn't look random.

matthewmayer (changes requested):

docs/guide/upgrading_v9/2357.md
```diff
@@ -0,0 +1,34 @@
...
+If you are using vitest, you can do that using `pnpm vitest run -u`.
+
+## Keeping the old behavior
+
+You can keep the old behavior, if you create your own `Faker` instance
+and pass a `Randomizer` instance from the `generateMersenne32Randomizer()` function to it.
+
+```ts
+const faker = new Faker({
+  randomizer: generateMersenne32Randomizer(),
```
matthewmayer, line 69: Should we show how to import generateMersenne3Randomizer()?

matthewmayer (changes requested):

docs/guide/upgrading_v9/2357.md
ST-DDT: grafik
or
grafik
or
grafik

With or without imports?

docs/guide/upgrading_v9/2357.md
ST-DDT: If I add them above should I add them below again?

ST-DDT:
grafik

test/modules/helpers.spec.ts
Shinigami92: I think @xDivisionByZerox is right here.  
The inline snapshot would indicate to me "yeah something has changes, take the new one"
Whereas here it should fail if something has changed so we can see in vitests report what failed

docs/guide/randomizer.md
```diff
@@ -12,6 +12,23 @@ There are two connected use cases we have considered where this might be needed:
...
+```ts
+import {
+  generateMersenne32Randomizer, // Default prior to v9
+  generateMersenne53Randomizer, // Default since v9
+} from '@faker-js/faker';
+
+const randomizer = generateMersenne53Randomizer();
+```
+
+The 32bit `Randomizer` is faster, but the 53bit `Randomizer` generates better random values (with significantly less duplicates).
```
matthewmayer: "less" should be "fewer"

docs/guide/upgrading_v9/2357.md
```diff
@@ -0,0 +1,70 @@
+# Use High Precision RNG by default
+
+In v9 we switch from a 32 bit random value to a 53 bit random value.
+We don't change the underlying algorithm much, but we now consume two seed values each step instead of one.
+This affects generated values in three ways:
+
+- In large lists or long numbers the values are spread more evenly.
+  This also reduces the number of duplicates it generates.
+  For `faker.number.int()` this reduces the duplicates from `1 / 10_000` to less than `1 / 8_000_000`.
+- Some seeded runs now return slightly different values, due to the last digits of random numbers changing,
```
matthewmayer: This seems a bit vague. Does any method now return different values for a given seed?

matthewmayer (changes requested):

docs/guide/randomizer.md
ST-DDT:
```suggestion
The 32bit `Randomizer` is faster, but the 53bit `Randomizer` generates better random values (with significantly fewer duplicates).
```

docs/guide/upgrading_v9/2357.md
ST-DDT: Yes, just have a look at all the changed snapshot files.

matthewmayer: Maybe change last sentence to something like

Many Faker methods will return a different result in v9 compared to v8 for the same seed.

ST-DDT: Changed and reworded

Shinigami92 (dismissed):

matthewmayer (dismissed):

xDivisionByZerox (dismissed):

ST-DDT:
3 approving reviews. Enabling auto merge.

Shinigami92 (approved):

PR #2369, opened 2023-09-01
feat: split SimpleFaker class from Faker

### LINKED ISSUE #1351: Split the Faker class into smaller units

Currently the main Faker class contains many features most users don't need in their application.

### Suggested solution

The main Faker class should be split into multiple units that build on top of each other:

- CoreFaker (only seed and rng logic)
- BaseFaker extends CoreFaker (all modules that don't depend on locales
- LocalizedFaker extends BaseFaker (locale support, but no locale modules yet)
- Faker extends LocalizedFaker (all default modules, same as now)

See also: https://github.com/faker-js/faker/discussions/805#discussioncomment-2690233

That way could users could just use `baseFaker.string.uuid()` and have a very minimal footprint after tree shaking.

The class names aren't final yet.
This needs to be documented properly.

### Alternative

_No response_

### Additional context

_No response_

Shinigami92:
references #1351 
- #1351


Shinigami92:
In `DateModule` we have a reference to `definitions`

https://github.com/faker-js/faker/pull/2369/files#diff-33d9552c7d1fb7f98cb50e15888b577d2609cd9b6f17a333b641752c5049629dR1001

I have multiple ideas how to solve this:
- Split a `BaseDateModule` from `DateModule` (theoretically we could do that with other modules as well)
- Add a check whether we are currently have `definitions` or not, if not the `throw Error` and indicate to use `Faker` instance instead
- open for alternative ideas

xDivisionByZerox:
> In `DateModule` we have a reference to `definitions`
> 
> [#2369 (files)](https://github.com/faker-js/faker/pull/2369/files#diff-33d9552c7d1fb7f98cb50e15888b577d2609cd9b6f17a333b641752c5049629dR1001)
> 
> I have multiple ideas how to solve this:
> 
> * Split a `BaseDateModule` from `DateModule` (theoretically we could do that with other modules as well)
> * Add a check whether we are currently have `definitions` or not, if not the `throw Error` and indicate to use `Faker` instance instead
> * open for alternative ideas

IMO we should simply exclude the date module from the `BaseFaker` class. Gernerally, if a module uses locale data, exclude it. If you want a `Date` you can still do `new Date(faker.number.int({ ... }))`. I would take the easy route here.

ST-DDT:
IMO `Date`s are very useful for everyday testing.

Shinigami92:
Originally I just put it there because datatype uses date.between
But datatype is deprecated :eyes: 

---

Well... datatype has boolean, that is needed for helpers... helpers has multiple and other nice to have stuff
it's like a waterfall, and all of that just because of month names in date module

Shinigami92:
Okay... I tested building a `BaseDateModule`, but this is bad, because it produces faker.date conflicts between the BaseFaker and Faker class...

src/internal/bind-this-to-member-functions.ts
```diff
@@ -15,11 +15,14 @@
...
+  do {
+    for (const name of Object.getOwnPropertyNames(p)) {
+      if (typeof instance[name] === 'function' && name !== 'constructor') {
+        instance[name] = instance[name].bind(instance);
+      }
     }
-  }
+
+    p = Object.getPrototypeOf(p);
+  } while (p !== Object.prototype);
```
Shinigami92, line 27: @xDivisionByZerox we would like your opinion about following suggestion brought by @ST-DDT 

He likes the assignment inside the while-loop, where I don't like assignments in conditions at all

```ts
  let p = instance;
  while ((p = Object.getPrototypeOf(p)) !== Object.prototype) {
    for (const name of Object.getOwnPropertyNames(p)) {
      if (typeof instance[name] === 'function' && name !== 'constructor') {
        instance[name] = instance[name].bind(instance);
      }
    }
  }
```

There is a eslint rule https://eslint.org/docs/latest/rules/no-cond-assign which is by default configured with `except-parens`, which allows the suggested code
But when set to `always`, it would result in an error (as I would like it)

#happyabouthavingathirdmaintainer :P

Shinigami92:
Potentially TODOs:
- add docs about `baseFaker`/`BaseFaker`
- add tests?!
- discuss if we want a `@faker-js/faker/base`
- export type `Base*Module`

ST-DDT:
We are not sure about what the actual questions here are.

**Team Decision**

> - add docs about baseFaker/BaseFaker

Yes

> - add tests?!

Only tests that concern creating the base faker instance.

> - discuss if we want a @faker-js/faker/base

Currently, we don't see a need for that, we should rather fix the tree-shaking.
Because if we add that now and remove it in v9 this will result in bad DX.

> - export type Base*Module

Yes, only as types.

> - Additionally

- Split BaseFaker+baseFaker into it's own file

scripts/apidoc/baseFakerClass.ts
```diff
@@ -0,0 +1,60 @@
+import type { DeclarationReflection, ProjectReflection } from 'typedoc';
```
Shinigami92: This file is nearly a 1 to 1 copy of the `scripts/apidoc/fakerClass.ts` but with `base` as prefix

src/base-faker.ts
```diff
@@ -0,0 +1,213 @@
...
+  get defaultRefDate(): () => Date {
+    return this._defaultRefDate;
+  }
+
+  /**
+   * Creates a new instance of BaseFaker.
+   *
+   * In nearly any case you should use the prebuilt `baseFaker` instances instead of the constructor.
+   */
+  constructor() {}
```
Shinigami92: This empty constructor just exists for VitePress docs

scripts/apidoc/baseFakerClass.ts
```diff
@@ -0,0 +1,60 @@
...
+  baseFakerClass: DeclarationReflection
+): Promise<ModuleSummary> {
+  console.log(`Processing BaseFaker class`);
+  const { comment, deprecated } = analyzeModule(baseFakerClass);
+  const methods: Method[] = [];
+
+  console.debug(`- constructor`);
+  methods.push(await processConstructor(baseFakerClass));
+
+  methods.push(...(await processModuleMethods(baseFakerClass, 'faker.')));
```
ST-DDT: Shouldn't this be `baseFaker`?

src/base-faker.ts
ST-DDT: Maybe put that comment into the code so it isn't lost?

scripts/apidoc/baseFakerClass.ts
ST-DDT: Why don't you change the original implementation and let it return a list of `*Faker`s?

src/base-faker.ts
```diff
@@ -0,0 +1,213 @@
...
+ * import { baseFaker } from '@faker-js/faker';
+ * // const { baseFaker } = require('@faker-js/faker');
+ *
+ * // baseFaker.seed(1234);
+ *
+ * baseFaker.number.int(10); // 4
+ * baseFaker.string.uuid(); // 'c50e1f5c-86e8-4aa9-888e-168e0a182519'
+ */
+export class BaseFaker {
+  protected _defaultRefDate: () => Date = () => new Date();
```
ST-DDT: Why does it have to be `protected`?

src/base-faker.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * Gets a new reference date used to generate relative dates.
+   */
+  get defaultRefDate(): () => Date {
```
ST-DDT: Why is the getter up here and the setter down below?

matthewmayer (changes requested):
It would be good to document what the difference is between fakerBASE and BaseFaker in `/guide/localization.html` since they sound confusingly similar.

Shinigami92:
> It would be good to document what the difference is between fakerBASE and BaseFaker in `/guide/localization.html` since they sound confusingly similar.

Thanks for the reminder 👍

scripts/apidoc/baseFakerClass.ts
Shinigami92: done

src/base-faker.ts
Shinigami92: done

scripts/apidoc/baseFakerClass.ts
Shinigami92: done

src/base-faker.ts
Shinigami92: This is a specific TS weird thingy. When we would do `private _defaultRefDate` in `BaseFaker` and then theoretically comes along and implements a `private _defaultRefDate` in `Faker`, the classes are not compatible with each other in terms of abstraction. However, as we only use TS class modifiers here and not real JS private (`#defaultRefDate`) the field is anyway "public" at runtime!
So I instead just used `protected` to indicate that it is an "internal" field that is not exposed to the API/method calls, but if someone wants to mess around with it while extending the class, feel free.

src/base-faker.ts
Shinigami92: Originally I just copied the fields and methods over from `Faker` in exactly the order they were.
I would argue that getters and setters are field aliases with extra-stuff and fields are above the constructor. But even that is already not consistent right now :shrug:  
Maybe you could find a lint rule to force a style? Until then, I would suggest this is not blocking.

ST-DDT: So you are making a distinction between the former being a property setter vs the other being a setter function?

IMO they are the same.

docs/guide/localization.md
```diff
@@ -66,6 +66,10 @@ In this example there are 5 locales. Each of these is checked in order, and the
 - `en` is a generic `en` (English) locale definition. This is our most complete locale, so we add it to fill some gaps. Depending on your needs, you might want or not want to have it as a fallback.
 - `base` is the base locale definition which contains definitions that can be used in every language (e.g. emojis).
 
+::: warning Attention
+Do not confuse the `base` locale / `fakerBASE` with the `baseFaker` instance, the later being a prebuilt `Faker` instance without any locale support.
```
ST-DDT: Not sure about the wording here.

scripts/apidoc/fakerClass.ts
```diff
@@ -4,36 +4,48 @@ import type { Method } from '../../docs/.vitepress/components/api-docs/method';
 import { writeApiDocsModule } from './apiDocsWriter';
 import { analyzeModule, processModuleMethods } from './moduleMethods';
 import { analyzeSignature } from './signature';
-import { selectApiSignature } from './typedoc';
+import { extractModuleFieldName, selectApiSignature } from './typedoc';
 import type { ModuleSummary } from './utils';
 
 export async function processFakerClass(
```
ST-DDT: Maybe rename this to `processFakerClasses` because it processes more than one?

src/base-faker.ts
```diff
@@ -0,0 +1,215 @@
+import type { Mersenne } from './internal/mersenne/mersenne';
+import mersenne from './internal/mersenne/mersenne';
+import { DatatypeModule } from './modules/datatype';
+import { BaseDateModule } from './modules/date';
+import { BaseHelpersModule } from './modules/helpers';
+import { NumberModule } from './modules/number';
+import { StringModule } from './modules/string';
+
+/**
+ * This is Faker's base class containing only modules that can be used to generate basic data.
```
ST-DDT: Maybe change this to something like:

```suggestion
 * This is Faker class that doesn't need any localized data to generate it's output. It should be used if only values are needed that can be generated on the spot.
```

src/base-faker.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /** @internal */
+  private readonly _mersenne: Mersenne = mersenne();
+
+  readonly datatype: DatatypeModule = new DatatypeModule(this);
+  readonly date: BaseDateModule = new BaseDateModule(this);
+  readonly helpers: BaseHelpersModule = new BaseHelpersModule(this);
+  readonly number: NumberModule = new NumberModule(this);
+  readonly string: StringModule = new StringModule(this);
```
ST-DDT: What about these properties? Should we move them up to the others?

ST-DDT (commented):
There are two things here that are troubling me:

- `fakerBASE` is very close to `baseFaker` but these two are very distinct things. We should probably change either of them.
- I would like to have an indicator somewhere in the docs, which methods are available in the `BaseFaker` and which do require an actual `Faker` (or vice versa)

docs/guide/localization.md
matthewmayer: i dont this this warning is very useful as-is. I think instead we should document somewhere how to actually use baseFaker.

ST-DDT:
We should have some documentation when and how to use the simpleFaker instead of a full Faker.

src/simple-faker.ts
```diff
@@ -0,0 +1,215 @@
...
+import { SimpleHelpersModule } from './modules/helpers';
+import { NumberModule } from './modules/number';
+import { StringModule } from './modules/string';
+
+/**
+ * This is Faker class that doesn't need any localized data to generate it's output. It should be used if only values are needed that can be generated on the spot.
+ *
+ * Please have a look at the individual modules and methods for more information and examples.
+ *
+ * @example
```
matthewmayer, line 19: `@example` at a module level don't show up in the generated docs. Suggest putting this in an "Overview" section like other modules

src/simple-faker.ts
```diff
@@ -0,0 +1,215 @@
...
+import { DatatypeModule } from './modules/datatype';
+import { SimpleDateModule } from './modules/date';
+import { SimpleHelpersModule } from './modules/helpers';
+import { NumberModule } from './modules/number';
+import { StringModule } from './modules/string';
+
+/**
+ * This is Faker class that doesn't need any localized data to generate it's output. It should be used if only values are needed that can be generated on the spot.
+ *
+ * Please have a look at the individual modules and methods for more information and examples.
```
matthewmayer:
```suggestion
* This is a simplified Faker class that doesn't need any localized data to generate its output. It only includes methods from the `datatype`, `date`, `helpers` and `string` modules.
```

src/simple-faker.ts
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * @example
+ * import { simpleFaker } from '@faker-js/faker';
+ * // const { simpleFaker } = require('@faker-js/faker');
+ *
+ * // simpleFaker.seed(1234);
+ *
+ * simpleFaker.number.int(10); // 4
+ * simpleFaker.string.uuid(); // 'c50e1f5c-86e8-4aa9-888e-168e0a182519'
```
matthewmayer, line 26: add another example which throws, e.g. `simpleFaker.person.fullName()`

matthewmayer (dismissed):

src/simple-faker.ts
Shinigami92: not related to this PR

src/simple-faker.ts
Shinigami92: it does not throw, it's not possible to call as TS already does not provide `simpleFaker.person`

ST-DDT (dismissed):

src/simple-faker.ts
matthewmayer: Screenshot 2023-09-14 at 18 43 14

i think it would be better to move any example code into the Overview section, otherwise there's no easy way to see how to use this new feature.

Shinigami92: You mean except of the usage guide?

Also it looks like this already does also not apply for https://fakerjs.dev/api/faker.html, so I definitely count this as not relevant for this PR and should be done in another PR
Potentially by someone who knows how to implement that via the api-gen scripts...

matthewmayer: i just find this page quite confusing at the moment because it mentions the "prebuilt simpleFaker instances" and yet this is not mentioned on the page at all (because its in the invisible @example)

image

src/modules/date/index.ts
```diff
@@ -23,23 +23,8 @@ function toDate(
...
- *
- * For more control, any of these methods can be customized with further options, or use [`between()`](https://fakerjs.dev/api/date.html#between) to generate a single date between two dates, or [`betweens()`](https://fakerjs.dev/api/date.html#betweens) for multiple dates.
- *
- * You can generate random localized month and weekday names using [`month()`](https://fakerjs.dev/api/date.html#month) and [`weekday()`](https://fakerjs.dev/api/date.html#weekday).
- *
- * These methods have additional concerns about reproducibility, see [Reproducible Results](https://fakerjs.dev/guide/usage.html#reproducible-results).
- */
-export class DateModule {
-  constructor(private readonly faker: Faker) {
+export class SimpleDateModule<TFaker extends SimpleFaker = SimpleFaker> {
```
xDivisionByZerox: Why is a generic argument required for this position? Since `Faker` extends `SimpleFaker` it should be possible to initialize this module with all arguments that derive from SimpleFaker. I think this is an over-complication, as you don't use the generic anywhere to enhance the return values of the API.

Shinigami92: Please show me how you would do it. I didn't get it to work probably without this.

xDivisionByZerox: Take a look here: https://github.com/faker-js/faker/commit/49c4f79377e4718b2ccc24b4cb455829df40a162

Shinigami92: That's interesting... I would assume that this would shadow the underlying member field 🤔 
@ST-DDT what is your opinion about this? Should we do @xDivisionByZerox way or "my" generic approach?

ST-DDT: We should check if this actually shadows the field. And if it does for production builds then go for the non generic one. But blth variants are almost identical for me.

Shinigami92: I tried 

```ts
import { faker, simpleFaker } from './src';

console.log(Object.getPrototypeOf(faker.date));
console.log(Object.getPrototypeOf(simpleFaker.date));
console.log(faker.date.month);
console.log(simpleFaker.date.month);
```

before and after, but no changes are there 🤔 
I don't know if there is a better check or how I could really check and access a field of an inheritance extension 🤷

Shinigami92: Very interesting, instead of shadowing, it looks like TS is smart and redefines just the type with that
It actually checks that the type is compatible and if it's not, then there is a compile time error

src/simple-faker.ts
```diff
@@ -0,0 +1,215 @@
...
+import { DatatypeModule } from './modules/datatype';
+import { SimpleDateModule } from './modules/date';
+import { SimpleHelpersModule } from './modules/helpers';
+import { NumberModule } from './modules/number';
+import { StringModule } from './modules/string';
+
+/**
+ * This is a simplified Faker class that doesn't need any localized data to generate its output.
+ *
+ * It only includes methods from the `datatype`, `date`, `helpers` and `string` modules.
```
ST-DDT: This list is missing the number module.

Maybe use an list like:

> - `datatype`
> - date (without `month` and `weekday`)
> - `helpers` (without fake)
> - number
> - string

or if the exclude list is to long/distracting just `(partially)`.

docs/guide/usage.md
```diff
@@ -127,6 +127,19 @@ or alternatively you can set a default reference date for all these methods:
...
+Also **helpers** like `arrayElement` or `multiple` are available.
+
+This is useful if you just want to generate e.g. `uuid`s for your production environment, but don't want to initiate/load a full Faker instance, which would include locale data.
+
+```ts
+import { simpleFaker } from '@faker-js/faker';
+
+const uuid = simpleFaker.string.uuid();
+```
+
```
ST-DDT, line 142: Maybe add a link to the api documentation to provide more information which modules are available exactly.

docs/guide/usage.md
```diff
@@ -127,6 +127,19 @@ or alternatively you can set a default reference date for all these methods:
 faker.setDefaultRefDate('2023-01-01T00:00:00.000Z');
 ```
 
+## Simple data generation
+
+Faker provides a `simpleFaker` that can be used to generate data that are not based on any locales like numbers and strings.  
+Also **helpers** like `arrayElement` or `multiple` are available.
+
+This is useful if you just want to generate e.g. `uuid`s for your production environment, but don't want to initiate/load a full Faker instance, which would include locale data.
```
ST-DDT: IMO we shouldn't recommend using it for production either.

```suggestion
This is useful if you just want to generate e.g. `uuid`s for your test environment, but don't want/need to initiate/load a full Faker instance, which would include at least 500KB of locale data.
```

Are 500KB correct?

Shinigami92: it's an estimation 👍

ST-DDT (approved):

src/simple-faker.ts
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * This is a simplified Faker class that doesn't need any localized data to generate its output.
+ *
+ * It only includes methods from:
+ * - `datatype`
+ * - `date` (without `month` and `weekday`)
+ * - `helpers` (without `fake`)
+ * - `number`
+ * - `string`
```
xDivisionByZerox, line 17: I would not put every module that is included here. Adds unnecessary maintainance burden if we delete or add modules.

Shinigami92: TBH we have this "burden" already with soooo many overview JSDocs like HelpersModule which could e.g. also include stuff then gets potentially deprecated

ST-DDT: I'm not sure about this manually maintained list either, but we need some kind of documentation which methods are available on baseFaker. Optimally on each method.

I thought about something like this (which might be overkill for this PR, not sure):

| | | |
| --- | --- | --- |
| ❌  BaseFaker | ✅ Faker | 🗺️ 15 Locales ``<hover><ul><li>🇩🇪 de_DE</li>...`` |

or

| | | |
| --- | --- | --- |
| ✅ BaseFaker | ✅ Faker | 🗺️  N/A |

Maybe between the params and the example block.

Shinigami92: How about we just start with the current PR and if there is really a need, we can enhance later anyway

xDivisionByZerox: Not a big fan. Just want to leave this here so I can tell you "I told you" if we need to change it later xD

xDivisionByZerox (approved):

PR #2381, opened 2023-09-09
feat(helpers): add support for complex intermediate return types in fake method

### LINKED ISSUE #1850: Add support for complex intermediate return types in fake method

Currently it is not possible to use methods, that return complex objects in the fake method, because it will only display `[object]`.

### Suggested solution

Add support for the following syntax:

````ts
faker.helpers.fake('{{airline.airport.code}}')
````

Which should be equivalent to:

faker.airline.airport().code

This should work for both method references and locale data references. 

### Alternative

_No response_

### Additional context

Ref: https://github.com/faker-js/faker/issues/1733#issuecomment-1433439926

### PR DESCRIPTION
**Pull request by ghost**
Fixes #1850 

- #1850 

I started to add support for complex intermediate return types in the fake method of the helpers module as described in #1850.

Currently, it is a work in progress. I would like to get feedback from you.

I wrote a few tests and will add more in the next days, for example for locale data references.

### DIFF
--- a/test/modules/helpers.spec.ts
+++ b/test/modules/helpers.spec.ts
@@ -1150,6 +1150,29 @@ describe('helpers', () => {
         it('should not trim whitespace', () => {
           expect(faker.helpers.fake('   ---   ')).toBe('   ---   ');
         });
+
+        it('should be able to handle methods that return a complex object', () => {
+          expect(
+            faker.definitions.airline.airport.map((airport) => airport.iataCode)
+          ).toContain(faker.helpers.fake('{{airline.airport.iataCode}}'));
+        });
+
+        it('should return undefined if a method returns a complex object but the property is undefined', () => {
+          expect(faker.helpers.fake('{{airline.airport.code}}')).toBe(
+            'undefined'
+          );
+        });
+
+        it('should be able to pass multiple dynamic templates with complex objects', () => {
+          expect(
+            faker.definitions.airline.airplane.map((airplane) => airplane.name)
+          ).toContain(
+            faker.helpers.fake([
+              '{{airline.airplane.name}}',
+              '{{airline.airplane.name}}',
+            ])
+          );
+        });
       });
 
       describe('rangeToNumber()', () => {
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -1190,6 +1190,7 @@ export class HelpersModule {
 
     // split the method into module and function
     const parts = method.split('.');
+    const property = parts.length === 3 ? parts.pop() : undefined;
 
     let currentModuleOrMethod: unknown = this.faker;
     let currentDefinitions: unknown = this.faker.rawDefinitions;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/modules/lorem/index.ts [hunk 1]
```diff
@@ -318,10 +321,33 @@ export class LoremModule {
...
+                 * The minimum length of text to generate.
+                 */
+                min: number;
+                /**
+                 * The maximum length of text to generate.
+                 */
+                max: number;
+              };
+        } = {}
+  ): string {
```
xDivisionByZerox: This entire feature should not be in this PR.

test/modules/helpers.spec.ts
```diff
@@ -1150,6 +1150,29 @@ describe('helpers', () => {
...
         });
+
+        it('should be able to handle methods that return a complex object', () => {
+          expect(
+            faker.definitions.airline.airport.map((airport) => airport.iataCode)
+          ).toContain(faker.helpers.fake('{{airline.airport.iataCode}}'));
+        });
+
+        it('should return undefined if a method returns a complex object but the property is undefined', () => {
+          expect(faker.helpers.fake('{{airline.airport.code}}')).toBe(
```
xDivisionByZerox, line 1161: Currently if a property is unknown on we receive an error: 

```md
Invalid module method or definition: person.abc
- faker.person.abc is not a function
- faker.definitions.person.abc is not an array
``` 

I would prefer if we could throw this error here as well and add a case for `faker.person is a function but does not return an object with the property abc`. I'll gladly take suggestions on the phrasing.

test/modules/helpers.spec.ts
```diff
@@ -1150,6 +1150,29 @@ describe('helpers', () => {
...
+            'undefined'
+          );
+        });
+
+        it('should be able to pass multiple dynamic templates with complex objects', () => {
```
xDivisionByZerox, line 1166: I'm not sure if this test actually adds any value since we already have a test for an array argument:

https://github.com/faker-js/faker/blob/604d52d9874db1763c185483ed182199f7a00f4f/test/modules/helpers.spec.ts#L1082-L1089

src/modules/lorem/index.ts
ghost: I removed it from this PR.

ST-DDT (changes requested):
Please also add a test that checks that if the fn didnt return a complex object and the property was requested that then an error is thrown.

Also I'm not sure about the current implementation.
E.g. Because it would break for `{{airline.airport().iataCode}}`
and it no longer supports it for 4 or deeper properties, but I'm not sure it has to.

matthewmayer:
Which of these do we expect to work

A {{airline.airport().iataCode}}

B {{airline.airport.iataCode}}

C {{airline.airport.iataCode()}}

ST-DDT:
> Which of these do we expect to work
> 
> A {{airline.airport().iataCode}}
> 
> B {{airline.airport.iataCode}}
> 
> C {{airline.airport.iataCode()}}

I would expect A and B to work.

ST-DDT:
Thanks for your contribution ❤️, we used it to create a new PR for this feature.

---

Superseded by #2550

- #2550

PR #2391, opened 2023-09-11
chore!: switch to tsup

Shinigami92:
This PR switches the build process from using `esbuild` nativly to `tsup` (which uses `esbuild` under the hood).
It simplifies the config and prepares for easier migration later to `esm` (see #2261)

- #2261

And even better! It magically reduces the `dist` folder content:

|               esbuild | ->  |                  tsup |
| --------------------: | :-: | --------------------: |
| 14.4 MiB (15,053,311) | ->  | 10.6 MiB (11,066,749) |
|           5,715 files | ->  |           3,072 files |
|     1,210 sub-folders | ->  |       605 sub-folders |


ST-DDT:
> And even better! It magically reduces the `dist` folder content:
> esbuild 	-> 	tsup
> 14.4 MiB (15,053,311) 	-> 	10.6 MiB (11,066,749)
> 5,715 files 	-> 	3,072 files
> 1,210 sub-folders 	-> 	605 sub-folders

Mind giving some details what/why files/data are now missing?

Shinigami92:
> Mind giving some details what/why files/data are now missing?

I was hoping you'd want to check out the PR and discover it for yourself, but okay, I can spoil it for you:

As you can already see in the `package.json`

https://github.com/faker-js/faker/blob/5445afec96024b2cf37a0e8b636cc13bfc2de4be/package.json#L31-L53

the `main`, `module` and `exports` fields have changed.
So the files are no longer in separate folders but side-by-side.
This already splits the sub-folders count into half.

The other changes comes from (🥁): tsup supports splitting for cjs 🎉

ST-DDT:
Well, not only does tsup support splitting, but also chunking (Hence the reduced number of files for cjs).

So for cjs users it is no longer possible to import directly from `.../dist/cjs/locales/af_ZA/cell_phone/formats.js`.
(Not that I would recommend doing so anyway)
Do we expect any of our users to use that kind of import paths?

Shinigami92:
> Well, not only does tsup support splitting, but also chunking (Hence the reduced number of files for cjs).

// chunking == splitting
 
> So for cjs users it is no longer possible to import directly from `.../dist/cjs/locales/af_ZA/cell_phone/formats.js`. (Not that I would recommend doing so anyway) Do we expect any of our users to use that kind of import paths?

No! Anything not defined in `exports` is **NOT** supported.

matthewmayer:
> > Well, not only does tsup support splitting, but also chunking (Hence the reduced number of files for cjs).
> 
> // chunking == splitting
> 
> > So for cjs users it is no longer possible to import directly from `.../dist/cjs/locales/af_ZA/cell_phone/formats.js`. (Not that I would recommend doing so anyway) Do we expect any of our users to use that kind of import paths?
> 
> No! Anything not defined in `exports` is **NOT** supported.

while its not supported, its still a breaking change, so i think it should land in 9.0

package.json
```diff
@@ -41,13 +41,13 @@
   "exports": {
     ".": {
       "types": "./dist/types/index.d.ts",
-      "require": "./dist/cjs/index.js",
-      "import": "./dist/esm/index.mjs"
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs"
```
prisis, line 45: Hey :)

shouldnt this look like this, without the types keyword?

```
					"require": {
                "types": "./dist/index.d.ts",
                "default": "./dist/index.js"
            },
            "import": {
                "types": "./dist/index.d.mts",
                "default": "./dist/index.mjs"
            }
```

Shinigami92: We do not have separate types for cjs or mjs so it reduces the size even further
That's why we bundle the types separately and not with tsup

prisis: You can use https://github.com/arethetypeswrong/arethetypeswrong.github.io to check the package.json

prisis: Ah, yeah you dont use tsup to bundle them, sorry

I saw this the last time, https://github.com/storybookjs/storybook/pull/19269#discussion_r991455547, so im not sure if we can have the same types for esm and cjs

ST-DDT: There seems to be some kind of differences: https://arethetypeswrong.github.io/?p=%40faker-js%2Ffaker%408.0.2

Shinigami92: Please compare this result: https://drive.google.com/file/d/1CTeTF2hjgeJTPMjl8fx1zwEASxnBxbe3/view?usp=sharing (uploaded a zip file with both results)

As you can see, these "problems" existed before and after and we never had real problems with that!
So yes, it says that there should be .d.cts files or whatever, but Faker is a special case, because we do not want to duplicate the types

ST-DDT (dismissed):

ejcheng (dismissed):

tsup.config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineConfig({
+  entry: [
+    'src/index.ts',
+    ...Object.keys(allLocales).map((locale) => `src/locale/${locale}.ts`),
+  ],
+  outDir: 'dist',
+  clean: true,
+  format: ['esm', 'cjs'],
+  target: ['es2019', 'node14.17'],
```
Shinigami92: when we want to merge this in v9 instead of v8.x, then this needs to be
```suggestion
  target: ['es2022', 'node18'],
```

ST-DDT:
@Shinigami92 Could you please fix the merge conflicts here?

ST-DDT:
I fixed the merge conflicts for you, but you still have to "update" the rest of the PR.

Shinigami92:
I checked if it is viable to ship `sourcemaps`, but hell no :laughing: 

```node
// With sourcemaps:
22.0 MiB (23,026,364)
580 files, 610 sub-folders

// Without sourcemaps:
10.0 MiB (10,437,487)
290 files, 610 sub-folders
```

12 MiB is NOT worth it :grimacing:

docs/guide/upgrading_v9/2391.md
```diff
@@ -0,0 +1,3 @@
+### Using tsup for the build process
+
+We only support exports defined via `package.json` but after the switch to `tsup`, there are now complete restructures in the dist folder resulting it minified and chunked files for cjs.
```
Shinigami92, line 3: @matthewmayer please have a look over this line and feel free to suggest improvements :slightly_smiling_face:

ST-DDT (dismissed):

ST-DDT (dismissed):

ST-DDT (approved):

Shinigami92:
@ST-DDT already approved a week ago and only a rebase were done since then
We count this a one-week approval by only one maintainer, but dont violate our merge rules

PR #2395, opened 2023-09-16
feat(locale): add airline database science commerce and vehicle for zh_CN

CoolPlayLin:
## Translation Work Status 📋
- [x] airline 
- [x] database
- [x] science
- [x] commerce
- [x] vehicle


src/locales/zh_CN/team/creature.ts
matthewmayer, line 1: Team file is currently unused so I wouldn't bother translating this one.

CoolPlayLin: What?

CoolPlayLin: @matthewmayer  OK, I won't translate this file and I am going to delete it directly

CoolPlayLin: done

src/locales/zh_CN/airline/airline.ts
ST-DDT, line 1: Why do some names contain latin characters and most of the others not?

src/locales/zh_CN/science/unit.ts
```diff
@@ -0,0 +1,118 @@
+export default Object.freeze([
+  {
+    name: '米',
```
ST-DDT, line 3: Google translate had issues translating this in file mode.
If I passed it as standalone, it translates this as rice and meter.
But if I put meter in, it returns 仪表.
I have no clue about Chinese and which one is beeter, so I posted this to just raise awareness about this.

ST-DDT (commented):
I try to review the remaining parts this PR later.
This might take a while.

src/locales/zh_CN/science/unit.ts
ejcheng: > Google translate had issues translating this in file mode. If I passed it as standalone, it translates this as rice and meter. But if I put meter in, it returns 仪表. I have no clue about Chinese and which one is beeter, so I posted this to just raise awareness about this.

The one in the PR currently (米) is correct, 仪表 means meter as in "measuring device" (like electric meter, gas meter, etc.)

src/locales/zh_CN/airline/airline.ts
CoolPlayLin: @ST-DDT I haven't found these Chinese names of these airlines. I don't think I could create a new word to name these airlines in Chinese and using latin character is maybe a way to avoid ambiguity.

matthewmayer: Could just remove the ones without real translations. The list of Airlines does not need to be the same each locale and I assume that any airline that has scheduled service to China and is more relevant to zh locale would have a translation.

CoolPlayLin: done

xDivisionByZerox (approved):
Looks good. Some values had not a literal translation, but I guess that those are still fine.

Next time, put separate modules into separate pull requests, please. Thank you :)

ST-DDT (approved):

Issue #2350, opened 2023-08-27
Verify all credit card issuer patterns are valid with validator.js


ejcheng:
We're currently not sure if all of our credit card issuer patterns are valid patterns. We should check them with validator.js.

HonzaMac:
Do you mean to check them in tests here, or in runtime as new dependency in faker?

ST-DDT:
In tests here.

xDivisionByZerox:
This issue only requires the implementation of the tests or fixing possible invalid patterns as well? If only the first one, this might be a `good first issue`.

ST-DDT:
It might also be fixing patterns. Or making sure the validatorjs' validation patterns for the credit cards are valid.

ST-DDT:
> Blocked by #2344

Merged/Fixed

matthewmayer:
I ran a large number of generated numbers through `validator.isCreditCard`. First the general check (no provider)
```
{
  mastercard: { valid: 91151, invalid: 0 },
  jcb: { valid: 90173, invalid: 0 },
  american_express: { valid: 90996, invalid: 0 },
  diners_club: { valid: 91178, invalid: 0 },
  discover: { valid: 30566, invalid: 61030 },
  visa: { valid: 90643, invalid: 0 },
  maestro: { valid: 16220, invalid: 74712 },
}
```

`discover` and `maestro` cards have problems, rest are OK.

Additionally when running `validator.isCreditCard(ccnumber, {provider: ...})` some `diners_club` and `mastercard` cards fail. `maestro` is not a supported provider by validator.

matthewmayer:
There are six patterns for `discover` in Faker.

These two succeed:
```
65##-####-####-###L
6011-####-####-###L
```

These four fail:
```
A) 6011-62##-####-####-###L
B) 64[4-9]#-62##-####-####-###L
C) 65##-62##-####-####-###L
D) 64[4-9]#-####-####-###L
```

For reference the validatorjs regex is `/^6(?:011|5[0-9][0-9])[0-9]{12,15}$/`

Referring to https://web.archive.org/web/20170822221741/https://www.discovernetwork.com/downloads/IPP_VAR_Compliance.pdf

it would seem to be that patterns A, B, C are incorrect on Faker's end. They generate 20 digit CC numbers which are invalid as the max length is 19. Pattern D is incorrect on validator's end, as Discover cards may start with 64 and be 16 digits.

matthewmayer:
[...]

PR #2396, opened 2023-09-16
chore(finance): update mastercard patterns and test

matthewmayer:
See  
https://github.com/faker-js/faker/issues/2350#issuecomment-1710316086

Updates the mastercard patterns to remove the weird 677189... pattern 
which is identified as a unionpay card, and instead add the much more 
common patterns starting 2221 – 2720 which were introduced in 2017 
https://www.cardfellow.com/blog/new-mastercard-bins/

Following this change, all mastercard card numbers pass validator's 
isCreditCard 
check, so switch to using that instead of a custom regex


ST-DDT:
Please also update this new file: https://github.com/faker-js/faker/blob/next/src/locales/zh_CN/finance/credit_card/mastercard.ts

src/locales/en/finance/credit_card/mastercard.ts
```diff
@@ -1 +1 @@
-export default ['5[1-5]##-####-####-###L', '6771-89##-####-###L'];
+export default ['5[1-5]##-####-####-###L', '2[221-720]-####-####-###L'];
```
ST-DDT, line 1: We have to keep these in mind because currently it is the only place that uses the `legacyRegexpStringParse`.

test/modules/finance.spec.ts
ST-DDT, line 1: Please also add a seeded test for mastercard, as that makes it easier to detect if we break the function/remove legacyRegexpStringParse in the future.

src/locales/en/finance/credit_card/mastercard.ts
matthewmayer: This does show off the value in having numeric ranges in the pattern!Notice how much more compact this representation is compared to the regex version in validator.

ST-DDT: How about a method for explicit string number range -> number replacement. e.g. only [min-max] support?
Or maybe keep it as an internal feature of the creditCard number method?
I'm really not sure what to do there going forward.

matthewmayer: I would just keep it as an internal method.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #2399, opened 2023-09-17
docs(contributing): clarify locale PR rules

ST-DDT:
The old rule only sets a hard rule without going into the reasons why.
I hope the new phrasing makes the intentions more clear.

e.g. the science, finance and airline modules are way harder to review then just words or commerce.
Even if they contain relatively few files.


Shinigami92 (dismissed):

CONTRIBUTING.md
```diff
@@ -45,7 +45,7 @@ If adding new data definitions to Faker, you'll often need to find source data.
 ## Adding new locale or updating existing one
 
 After adding new or updating existing locale data, you need to run `pnpm run generate:locales` to generate/update the related files.
-If you change more than 20 locale files, please consider splitting your PR into one per category (e.g. person, location).
+Please split your PRs into one per category (e.g. person, location) to simplify/speed up reviews.
```
xDivisionByZerox: I fear that mentioning the PR process might be 6 late for most people. I don't think that most people will structure their comits in a way to be able to split them into different branches and thus PRs. That why j tried to rephrase it so you have to start thinking as soon as you do changes, not when creating PR. I also did go in more detail on the explanation. 

```suggestion
Please only change files related to one module (e.g. person, location) whenever possible. This can simplify/speed up the review process. Additionally, it allows the maintainers to track PRs in a meaningful way by adding related labels.
```

xDivisionByZerox (dismissed):

Shinigami92 (approved):

CONTRIBUTING.md
ST-DDT: Nice. I tried multiple variants myself beforehand but I was never satisfied.
I was probably to close to the issue to see the broader horizon. 
Thanks for your suggestion! ❤️

ejcheng (approved):

matthewmayer (approved):

PR #2400, opened 2023-09-17
fix(locale): Dutch phone number

RobinvanderVliet:
A Dutch phone number in national format consists of 10 digits. For mobile numbers, the second digit is always six. To obtain the international format, substitute the leading zero with "+31".

This pull request guarantees adherence to this format for all phone numbers and also includes examples of the international format.


src/locales/nl/phone_number/formats.ts
```diff
@@ -1 +1,8 @@
-export default ['(####) ######', '##########', '06########', '06 #### ####'];
+export default [
+  '(0###) ######',
+  '06 #### ####',
+  '0#########',
+  '06########',
```
ST-DDT, line 5: AFAICT the first one contains the later. Is there a reason to have both?
I think it is OK if it is for distribution sake (more power for mobile numbers).

src/locales/nl/phone_number/formats.ts
```diff
@@ -1 +1,8 @@
...
+  '+31#########',
+  '+316########',
```
ST-DDT, line 7: Same here.

ST-DDT (approved):

src/locales/nl/phone_number/formats.ts
RobinvanderVliet: It's purely for distribution. It would be good to have a more or less equal distribution of mobile and non-mobile phone numbers.

matthewmayer (approved):

PR #2405, opened 2023-09-18
ci: check mergable by label

Shinigami92:
This ci blocks PRs by the labels on hold and do not merge yet, so it cannot accidentally be merged by not reading the labels

Code is slightly copied from https://stackoverflow.com/questions/71502652/prevent-merging-of-branch-conditionally-based-on-label


.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,30 @@
...
+        if: contains(github.event.pull_request.labels.*.name, 'do NOT merge yet')
+        run: |
+          echo "This PR is labeled as do NOT merge yet!"
+          exit 1
+
+  fail-by-on-hold-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if PR is labeled on hold
+        if: contains(github.event.pull_request.labels.*.name, 's: on hold')
```
Shinigami92: In VSCode this currently shows following error, and I don't know why and what is broken 🤔 

```node
Nested mappings are not allowed in compact mappings at line 27, column 13:

        if: contains(github.event.pull_request.labels.*.name, 's: on hold')
            ^
```

image

Shinigami92: fixed by applying https://github.com/actions/runner/issues/1019#issuecomment-1120492712

.github/workflows/check-mergable-by-label.yml
ST-DDT, line 1: I would merge them into one Job so that our job list does grow that much. And it shouldnt matter that much either way which label blocks it exactly.

Shinigami92: I thought about this as well 🤔 But on the other side it makes it more clear to a contributor why exactly the CI fails

If both are merged into on job, we cannot diff anymore and cannot show a specific message

Shinigami92:
"s: on hold" does not work correctly ❌

Shinigami92:
do not merge yet works correctly ✅

Shinigami92:
Need to work on this later, I have meetings 👋

.github/workflows/check-mergable-by-label.yml
ST-DDT: > I thought about this as well 🤔 But on the other side it makes it more clear to a contributor why exactly the CI fails

I think the label names are self explanatory especially in combination with "blocked by label" CI.

> If both are merged into on job, we cannot diff anymore and cannot show a specific message

The label being added or removed already has its own github notification, there is no need for a bot to add more (especially if it works for members only).

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,23 @@
...
+      - unlabeled
+
+permissions: {}
+
+jobs:
+  fail-by-blocking-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if PR is labeled by a blocking label
+        if: contains(github.event.pull_request.labels.*.name, 'do NOT merge yet') || endsWith(github.event.pull_request.labels.*.name, 'on hold')
```
Shinigami92: The `:` in `s: on hold` makes trouble, so I'm testing `endsWith` instead

Shinigami92:
Current solution is based on https://chat.openai.com/share/8cb16814-2855-4fbd-87e5-bde8ba349728

Now it seems to work

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,34 @@
...
+          for LABEL in "${BLOCKING_LABELS[@]}"; do
+            if [[ "$PR_LABELS" == *"$LABEL"* ]]; then
+              echo "This PR has a blocking label: $LABEL"
+              exit 1
+            fi
+          done
+
+          echo "No blocking labels found"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ST-DDT: Have you tried running this without the token?

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,34 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+      - labeled
+      - unlabeled
+
+permissions:
+  pull-requests: read
```
ST-DDT: I don't see where is permission is used. Is that something implicit?

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,34 @@
+name: Check mergable by label
```
ST-DDT:
```suggestion
name: PR-Labels
```

ST-DDT: or
```suggestion
name: PR Labels
```

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,34 @@
...
+      - unlabeled
+
+permissions:
+  pull-requests: read
+
+jobs:
+  fail-by-blocking-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if PR is labeled by a blocking label
```
ST-DDT:
```suggestion
      - name: Check for blocking labels
```

.github/workflows/check-mergable-by-label.yml
```diff
@@ -0,0 +1,34 @@
+name: Check mergable by label
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
```
ST-DDT, line 9: AFAICT these two should be unrelated.

https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request

Shinigami92: These are especially for rebasing all current PRs, so they get checked when this PR gets merged

ST-DDT (approved):

prisis (approved):

.github/workflows/check-mergable-by-label.yml
xDivisionByZerox: So should we remove those at a later point in time?

xDivisionByZerox (approved):

.github/workflows/check-mergable-by-label.yml
Shinigami92: I don't think the effort is worth it 😅
It just works with or without them

PR #2406, opened 2023-09-18
docs: show examples for modules

matthewmayer:
Shows @example blocks for modules (ie Faker and SimpleFaker) in the 
generated docs.

Screenshot 2023-09-18 at 19 22 35

### DIFF
--- a/scripts/apidoc/apiDocsWriter.ts
+++ b/scripts/apidoc/apiDocsWriter.ts
@@ -37,8 +37,8 @@ editLink: false
  * @param moduleName The name of the module to write the docs for.
  * @param lowerModuleName The lowercase name of the module.
  * @param comment The module comments.
- * @param deprecated The deprecation message.
  * @param examples The example code.
+ * @param deprecated The deprecation message.
  * @param methods The methods of the module.
  */
 export async function writeApiDocsModule(
--- a/scripts/apidoc/apiDocsWriter.ts
+++ b/scripts/apidoc/apiDocsWriter.ts
@@ -38,14 +38,14 @@ editLink: false
  * @param lowerModuleName The lowercase name of the module.
  * @param comment The module comments.
  * @param deprecated The deprecation message.
- * @param example The example code.
+ * @param examples The example code.
  * @param methods The methods of the module.
  */
 export async function writeApiDocsModule(
   moduleName: string,
   lowerModuleName: string,
   comment: string,
-  example: string | undefined,
+  examples: string | undefined,
   deprecated: string | undefined,
   methods: Method[]
 ): Promise<ModuleSummary> {
@@ -53,7 +53,7 @@ export async function writeApiDocsModule(
     moduleName,
     lowerModuleName,
     comment,
-    example,
+    examples,
     deprecated,
     methods
   );
@@ -86,7 +86,7 @@ export async function writeApiDocsModule(
  * @param moduleName The name of the module to write the docs for.
  * @param lowerModuleName The lowercase name of the module.
  * @param comment The module comments.
- * @param example The example code.
+ * @param examples The example code.
  * @param deprecated The deprecation message.
  * @param methods The methods of the module.
  */
@@ -94,7 +94,7 @@ async function writeApiDocsModulePage(
   moduleName: string,
   lowerModuleName: string,
   comment: string,
-  example: string | undefined,
+  examples: string | undefined,
   deprecated: string | undefined,
   methods: Method[]
 ): Promise<void> {
@@ -124,7 +124,7 @@ async function writeApiDocsModulePage(
 
   ${comment}
 
-  ${example == null ? '' : `<div class="examples">${example}</div>`}
+  ${examples == null ? '' : `<div class="examples">${examples}</div>`}
 
   :::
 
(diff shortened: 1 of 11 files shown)


Shinigami92:
(Currently on mobile) the PR title should just be docs: ...

test/scripts/apidoc/__snapshots__/module.spec.ts.snap
ST-DDT, line 1: Please add a test case that is not undefined.

scripts/apidoc/moduleMethods.ts
```diff
@@ -65,10 +68,19 @@ async function processModule(
 export function analyzeModule(module: DeclarationReflection): {
   comment: string;
   deprecated: string | undefined;
+  example: string | undefined;
 } {
+  const exampleTags = extractRawExamples(module);
+  let example;
+  if (exampleTags.length > 0) {
+    const code = '```';
+    example = mdToHtml(`${code}ts\n${exampleTags.join('\n').trim()}${code}\n`);
```
ST-DDT: I think we should expose something like `extractRawExamples(module).join('\n').trim()` as a method, because we only use it like that 4 times. e.g. `extractRawJoinedExamples()` (./typedoc)

Also I think we should add a `codeToHtml` helper method as well. (./markdown)

scripts/apidoc/moduleMethods.ts
```diff
@@ -65,10 +68,19 @@ async function processModule(
...
+  }
+
   return {
     comment: adjustUrls(extractDescription(module)),
     deprecated: extractDeprecated(module),
+    example,
```
ST-DDT: I think for methods we name this `examples` so we should keep it consistent.

scripts/apidoc/signature.ts
```diff
@@ -74,9 +74,9 @@ export async function analyzeSignature(
 
   let examples = `${accessor}${methodName}${signatureTypeParametersString}(${signatureParametersString}): ${signature.type?.toString()}\n`;
 
-  const exampleTags = extractRawExamples(signature);
-  if (exampleTags.length > 0) {
-    examples += `${exampleTags.join('\n').trim()}\n`;
+  const exampleTags = extractJoinedRawExamples(signature);
+  if (exampleTags) {
+    examples += exampleTags;
```
ST-DDT, line 77:
```suggestion
    examples += `${exampleTags}\n`;
```

Alternatively change the `codeToHtml` method to include that newline

scripts/apidoc/typedoc.ts
```diff
@@ -248,6 +248,18 @@ export function extractRawExamples(reflection?: CommentHolder): string[] {
   return extractRawCode('@example', reflection);
```
ST-DDT, line 248: The `extractRawExamples` function no longer has to be exported.

scripts/apidoc/moduleMethods.ts
```diff
@@ -70,12 +70,11 @@ export function analyzeModule(module: DeclarationReflection): {
...
-  const exampleTags = extractRawExamples(module);
-  let examples;
-  if (exampleTags.length > 0) {
-    const code = '```';
-    examples = mdToHtml(`${code}ts\n${exampleTags.join('\n').trim()}${code}\n`);
-  }
+  const examplesRaw = extractJoinedRawExamples(module);
+  const code = '```';
+  const examples = examplesRaw
+    ? mdToHtml(`${code}ts\n${examplesRaw}${code}\n`)
```
ST-DDT:
```suggestion
    ? mdToHtml(`${code}ts\n${examplesRaw}\n${code}`)
```

test/scripts/apidoc/module.example.ts
```diff
@@ -9,3 +9,11 @@ export class ModuleFakerJsLinkTest {}
...
  */
 export class ModuleNextFakerJsLinkTest {}
+
+/**
+ * This is a description for a module with a code example
+ *
+ * @example
+ * new ModuleFakerJsExampleTest()
+ */
+export class ModuleFakerJsExampleTest {}
```
ST-DDT: The `FakerJs` part refers to the target link, IMO there is no need to repeat that here.

```suggestion
export class ModuleExampleTest {}
```

scripts/apidoc/apiDocsWriter.ts
```diff
@@ -38,19 +38,22 @@ editLink: false
  * @param lowerModuleName The lowercase name of the module.
  * @param comment The module comments.
  * @param deprecated The deprecation message.
+ * @param examples The example code.
```
ST-DDT: These two tags are swapped compared to the method definition.

scripts/apidoc/signature.ts
matthewmayer: is this ok now?

Shinigami92 (approved):

xDivisionByZerox (approved):

scripts/apidoc/signature.ts
```diff
@@ -74,9 +72,9 @@ export async function analyzeSignature(
...
   }
```
ST-DDT, line 78: IMO this could be inlined like this: 

````ts
const examples = `${accessor}${methodName}${signatureTypeParametersString}(${signatureParametersString}): ${signature.type?.toString()}\n${extractJoinedRawExamples(signature) ?? ''}`;
````

but that isn't worth blocking this.

ST-DDT (approved):

PR #2408, opened 2023-09-19
infra: lint all existing jsdocs

ST-DDT:
This PR fixes all lint errors that aren't `Missing JSDoc comment` in scripts.

Note, that currently the jsdocs lint rules don't apply to the scripts at all.


matthewmayer:
Can you make this part of `preflight` so lint issues dont creep back in?

ST-DDT:
Theoretically yes. The problem is that I didnt fix the missing jsdoc lint issues. So we either had to fix them or disable the rule for scripts (and tests).
The non jsdocs lint checks are enabled for our scripts.

I'm somewhat torn on this and would like to hear more opinions on this.

Shinigami92 (dismissed):

Shinigami92:
> I'm somewhat torn on this and would like to hear more opinions on this.

JSDocs are checked in src because they are provided when consumers of Faker hover over the methods and stuff.
The scripts are internal and not published/exposed, so "I" don't care about JSDocs in scripts, they just help to remember what is for what when coming back a half year later and try to find stuff out. They are not public-docs.

matthewmayer:
Agree Jsdocs in the scripts are a nice to have but much much less important than in src. 

If they exist they should probably be linted correctly, but missing jsdocs in scripts is not important for self explanatory methods.

ST-DDT:
I enabled the jsdoc checks for our entire code base, but only require jsdocs for `src`.

.eslintrc.js
```diff
@@ -77,40 +78,39 @@ module.exports = defineConfig({
       { allowNumber: true, allowBoolean: true },
     ],
     '@typescript-eslint/unbound-method': 'off',
+
+    'jsdoc/require-jsdoc': 'off',
+    'jsdoc/no-types': 'error',
+    'jsdoc/require-returns': 'off',
+    'jsdoc/tag-lines': 'off',
+    'jsdoc/sort-tags': [
```
Shinigami92: I prefer to sort these in alphabetic order to minimize future conflicts

What is the current order?

ST-DDT: Copied from previous

Shinigami92: Should I do an extra PR after we merged this one? Or would you like to sort it directly within this PR?

ST-DDT: The other elements are sorted already AFAICT and it isnt that much efford to sort the jsdoc ones.

Is there an eslint rule for that?

Shinigami92: Comparing it with https://github.com/faker-js/faker/pull/2408/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L87-L92 doesn't even shows a copy-paste to me 🤔 
The `no-types` is in another position, and the `require-returns-type` is fully missing

Shinigami92: > The other elements are sorted already AFAICT and it isnt that much efford to sort the jsdoc ones.
> 
> Is there an eslint rule for that?

there can't be an eslint rule for that, because all rules are just in a plain JS object "config"
and because there are manually grouped by plugins, there is no rule that can enforce that

but I'm already working on something called `defineRules` 👀, but this is something for another day

ST-DDT: > Comparing it with [#2408 (files)](https://github.com/faker-js/faker/pull/2408/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L87-L92) doesn't even shows a copy-paste to me 🤔 The `no-types` is in another position, and the `require-returns-type` is fully missing

Sorry for the confusion. The `require-jsdoc` is new and the `require-returns-type` + x are part of the `recommended-typescript-error` set.

ST-DDT: Sorted

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):

PR #2415, opened 2023-09-23
feat(module): add new notations module for science

### LINKED ISSUE #2416: Add new notations module for science

Add notations module to science so that we could use symbol to express units in science

### Suggested solution

According to https://en.wikipedia.org/wiki/List_of_common_physics_notations

### Alternative

_No response_

### Additional context

_No response_

CoolPlayLin:
- Resolve #2416


matthewmayer:
You should make a matching issue for this PR so that we can gauge user interest for the new feature. New methods (as opposed to new translations for existing methods) generally need several 👍 before they are considered for inclusion.

CoolPlayLin:
Why it fails?

matthewmayer:
> Why it fails?

you need to add a JSDoc to the new method, with at least `@example` and `@since` tags

CoolPlayLin:
@matthewmayer What should I write `since`?

matthewmayer:
> @matthewmayer What should I write `since`?

doesn't really matter. any future version e.g. 8.2.0 or 9.0.0. If the issue gets upvoted enough to be approved the PR would be adjusted later to match the correct release version.

CoolPlayLin:
done

PR #2431, opened 2023-10-05
chore: randomizer cleanup

ST-DDT:
Cleanup PR for some missed changes.


src/internal/mersenne.ts
```diff
@@ -328,13 +328,14 @@ class MersenneTwister19937 {
 
 /**
  * Generates a MersenneTwister19937 randomizer with 32 bits of precision.
+ * This is the default randomizer used by faker.
```
ejcheng:
```suggestion
 * This is the default randomizer used by Faker.
```

ejcheng (dismissed):

matthewmayer (approved):

ejcheng (dismissed):

matthewmayer (dismissed):

test/mersenne.spec.ts
```diff
@@ -6,7 +6,7 @@ import { times } from './support/times';
 
 const NON_SEEDED_BASED_RUN = 25;
 
-describe('mersenne twister', () => {
+describe('generateMersenne32Randomizer()', () => {
```
xDivisionByZerox, line 9: If the tests are now named after the generation function I would expect a check for the correct return type (an object matching the `Randomizer` interface). We do similare things for our faker methods even with TypeScript type safety.

src/internal/mersenne.ts
```diff
@@ -328,13 +328,14 @@ class MersenneTwister19937 {
...
  * Generates a MersenneTwister19937 randomizer with 32 bits of precision.
+ * This is the default randomizer used by Faker.
  *
  * @internal
  */
 export function generateMersenne32Randomizer(): Randomizer {
   const twister = new MersenneTwister19937();
 
-  twister.initGenrand(Math.ceil(Math.random() * Number.MAX_SAFE_INTEGER));
+  twister.initGenrand(Date.now() ^ (Math.random() * 0x100000000));
```
xDivisionByZerox: I might not be smart enought to understand this but how are these implementations equal? Since this PR is a `chore` nothing should change at runtime.

xDivisionByZerox (changes requested):
Sorry to be nitpicky.

src/internal/mersenne.ts
ST-DDT: They are not equal per se. They are equal in the terms that it is a "random number".
The new code is following the recommended default seed we have in our documentation.
I can revert this if you want.

xDivisionByZerox: The new implementation can resolve to a negative integer. Thats at least the result I'm getting after spamming it in the browser console a bunch of times.

image

 Is this intentional? The old implementation was not able to do that AFAICT.

test/mersenne.spec.ts
ST-DDT: Added

src/internal/mersenne.ts
ST-DDT: The values itself don't really matter as long as they are random and 32 bit.
I reverted that for now and will change that later (v9) when I try to simplify the mersenne implementation to improve it's performance.

xDivisionByZerox (approved):

ejcheng (approved):

PR #2434, opened 2023-10-05
infra: enable no-useless-escape eslint rule

ST-DDT:
This lint rule is likely to be enabled/"covered" by eslint rule `unicorn/better-regex` after #2418 is merged.

- #2418

This PR is independent of that PR though.


xDivisionByZerox (approved):
This is acutally pretty cool. I always forget which characters to escape and default to simply ascaping all symbols. This lets me know while coding without having to look it up additionally. I like that.

matthewmayer (approved):

PR #2435, opened 2023-10-05
infra: enable strictNullChecks in tsconfig

### LINKED ISSUE #1834: Turn on strictNullChecks in tsconfig

### LINKED ISSUE #2503: Using specific locales and calling prefix or jobTitle throws error: Calling `faker.helpers.arrayElement()` without arguments is no longer supported.
### Pre-Checks

### Describe the bug

When using a specific locale and method `person.prefix` or `person.jobTitle` it throws an error: 

```
Calling `faker.helpers.arrayElement()` without arguments is no longer supported.
```

I suspect the issue is the locale data has empty/missing array which is passed into the helper method by prefix or jobTitle. 

#### Expected behaviour
I would expect this not to crash/throw an error. 

#### Effected locales
az
id_ID
ru
zh_CN
zh_TW

#### Solutions
1. All locales have an empty array set which is passed into arrayElement method
2. Allow arrayElement to be called with undefined

### Minimal reproduction code

https://codesandbox.io/s/sad-maxwell-m44x5y?file=/src/index.mjs

### Additional Context

There may be other locales and methods that are also effected by this too.

### Environment Info

```shell
System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M1
    Memory: 76.72 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.8.1 - ~/.nvm/versions/node/v20.8.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.8.1/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.8.1/bin/npm
    pnpm: 8.9.2 - ~/.nvm/versions/node/v20.8.1/bin/pnpm
  Browsers:
    Brave Browser: 118.1.59.122
    Safari: 17.0
  npmPackages:
    @faker-js/faker: ^8.2.0 => 8.2.0
```

- [X] CJS
- [X] ESM

### Used Package Manager

pnpm

ST-DDT:
Fixes #1834

- #1834


docs/.vitepress/versions.ts
```diff
@@ -23,8 +23,9 @@ function readOtherLatestReleaseTagNames(): string[] {
       }
       return acc;
     }, {});
+
```
ST-DDT, line 30:
```suggestion
```

src/modules/date/index.ts
```diff
@@ -387,14 +391,15 @@ export class SimpleDateModule {
         },
     legacyTo?: string | Date | number
   ): Date {
-    if (typeof options !== 'object' || options instanceof Date) {
+    if (options instanceof Date || typeof options !== 'object') {
```
ST-DDT, line 391: Flipping this somehow fixes an ts-error in line 405 🤷‍♂️ .

src/modules/person/index.ts
```diff
@@ -97,7 +108,7 @@ export class PersonModule {
       this.faker.rawDefinitions.person ?? {};
 
     return selectDefinition(this.faker, this.faker.helpers.arrayElement, sex, {
-      generic: first_name,
+      generic: first_name ?? undefined,
```
ST-DDT:
```suggestion
      generic: first_name,
```

cypress/e2e/api.cy.ts
```diff
@@ -34,7 +34,7 @@ describe('API Test', () => {
       cy.get('.api-group li').each(($el) => {
         const anchor = $el.find('a');
         const text = anchor.text();
-        const link = anchor.attr('href').split('#')[0];
+        const link = anchor.attr('href')!.split('#')[0];
```
xDivisionByZerox: I mena we wouldn't really activate this just to work around it again, would we?

```suggestion
        const link = anchor.attr('href')?.split('#')[0] ?? '';
```

src/modules/finance/index.ts
```diff
@@ -1163,7 +1163,7 @@ export class FinanceModule {
 
     const result = `${ibanFormat.country}${checksum}${s}`;
 
-    return formatted ? result.match(/.{1,4}/g).join(' ') : result;
+    return formatted ? result.replace(/(.{4})(?=.)/g, '$1 ') : result;
```
ST-DDT: Which is easier to read?

- `result.match(/.{1,4}/g)!.join(' ')`
- `result.replace(/(.{4})(?=.)/g, '$1 ')`
- `result.replace(/(.{4})/g, '$1 ').trimRight()`

The first seems to be the fastest (using unreliable online tests).

test/modules/finance_iban.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('finance_iban', () => {
 
           expect(iban).toSatisfy(validator.isIBAN);
 
-          const ibanFormatted = iban.match(/.{1,4}/g).join(' ');
+          const ibanFormatted = iban.match(/.{1,4}/g)!.join(' ');
```
ST-DDT: Should I adjust the code here as well?

src/locales/ar/date/weekday.ts
```diff
@@ -1,5 +1,13 @@
 export default {
-  abbr: null,
```
xDivisionByZerox: Didn't we explicitly decide to allow `null` for "not appicable" in out locale data? I think we should have a look into our `LocaleDefinition` type before "fixing" those values. 

In the case:
Should we add conert the month property from `DateEntryDefinition` to `LocaleEntry<DateEntryDefinition>`? Or even better, should `LocaleEntry` recursivly apply itself to types which extend `object`? That why we only have to apply `LocaleEntry` ones to the actual `LocaleDefinition` type itself.

tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "stripInternal": true,
 
     // We need to disable these for now, and need to tackle them in another PR
-    "strictNullChecks": false,
```
ST-DDT, line 12: Is removing this correct, or do we have to set this to `true` instead?

ST-DDT:
Ready for review.

tsconfig.json
xDivisionByZerox: We have the `strict` property aboth that would usally include `strictNullChecks`. So removing it here enables it again. Kinda similar to what we doing with the unicorn plugin.

docs/.vitepress/versions.ts
```diff
@@ -24,7 +24,7 @@ function readOtherLatestReleaseTagNames(): string[] {
       return acc;
     }, {});
   return Object.entries(latestReleaseTagNames)
-    .map(([major, tags]) => semver.maxSatisfying(tags, `^${major}`))
+    .map(([major, tags]) => semver.maxSatisfying(tags, `^${major}`)!)
```
xDivisionByZerox:
```suggestion
    .map(([major, tags]) => semver.maxSatisfying(tags, `^${major}`))
    .filter((version): version is string => typeof version === 'string')
```

scripts/apidoc/fakerClass.ts
```diff
@@ -62,7 +66,7 @@ async function processClass(
 function hasConstructor(clazz: DeclarationReflection): boolean {
   return clazz
     .getChildrenByKind(ReflectionKind.Constructor)
-    .some((constructor) => constructor.signatures.length > 0);
+    .some((constructor) => constructor.signatures!.length > 0);
```
xDivisionByZerox:
```suggestion
    .some((constructor) => (constructor.signatures?.length ?? 0) > 0);
```

src/modules/commerce/index.ts
```diff
@@ -386,7 +386,7 @@ export class CommerceModule {
 
     const registrantLength = groupRules.find(
       ([rangeMaximum]) => elementValue <= rangeMaximum
-    )[1];
+    )![1];
```
xDivisionByZerox:
```suggestion
    ) ?? [0, 0])[1];
```

Sadly can't add lines which have not been touched. So I cant set the open bracket in line 387.
Alternitivly we could check if the `find` resolves to `undefined` and throw an error in that case. That would get rid of the non-null assertion.

src/modules/date/index.ts
xDivisionByZerox: Because the last thing to resolve here is `typeof options !== 'object'` so if this fails `options` is definitly an object; if it doesn't fail ts can infer that `options` will definitly be an object after the block statement as well. This might sound stupid but the statement before TypeScript would have been able to tell that `options` was 100% a `Date`, but could say the same for `object`. 
Although, I cant tell you an example where that would happen. Propably some prototype shenanigan.

src/modules/date/index.ts
```diff
@@ -387,14 +391,15 @@ export class SimpleDateModule {
...
       deprecated({
         deprecated: 'faker.date.between(from, to)',
         proposed: 'faker.date.between({ from, to })',
         since: '8.0',
         until: '9.0',
       });
-      options = { from: options, to: legacyTo };
+      // We use options as fallback for legacyTo avoid TS errors for unintended usage.
+      options = { from: options, to: legacyTo ?? options };
```
xDivisionByZerox, line 399: Can we extend the comment to remove this workaround in v9?

```suggestion
      // We use options as fallback for legacyTo avoid TS errors for unintended usage.
      // TODO: Remove in v9
      options = { from: options, to: legacyTo ?? options };
```

src/modules/date/index.ts
```diff
@@ -1108,7 +1114,14 @@ export class DateModule extends SimpleDateModule {
...
 
-    return this.faker.helpers.arrayElement(source[type]);
+    const values = source[type];
+
+    if (values == null) {
+      // This should never happen due to the checks above, but TS doesn't know that
+      throw new FakerError(`No month values found for type '${type}'.`);
+    }
+
+    return this.faker.helpers.arrayElement(values);
```
xDivisionByZerox: Can't we simply this by letting `arrayElement` handle the "error tha is never possible"?

```suggestion
    return this.faker.helpers.arrayElement(source[type] ?? []);
```

src/modules/date/index.ts
```diff
@@ -1298,6 +1311,13 @@ export class DateModule extends SimpleDateModule {
...
 
-    return this.faker.helpers.arrayElement(source[type]);
+    const values = source[type];
+
+    if (values == null) {
+      // This should never happen due to the checks above, but TS doesn't know that
+      throw new FakerError(`No weekday values found for type '${type}'.`);
+    }
+
+    return this.faker.helpers.arrayElement(values);
```
xDivisionByZerox: Same here?

```suggestion
    return this.faker.helpers.arrayElement(source[type] ?? []);
```

src/modules/finance/index.ts
```diff
@@ -679,9 +679,9 @@ export class FinanceModule {
    */
   currencySymbol(): string {
     let symbol: string;
-    while (!symbol) {
+    do {
       symbol = this.currency().symbol;
-    }
+    } while (!symbol);
```
xDivisionByZerox: I had to look up while this is even getting looped. We should make the check more percise since the symbol itself always exists (as an empty string). Alternatively we could update the `Currency` definition to allow optional or `null` values.

```suggestion
    } while (symbol.length === 0);
```

test/faker.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('faker', () => {
   describe('rawDefinitions', () => {
     it('locale rawDefinition accessibility', () => {
       // Metadata
-      expect(faker.rawDefinitions.metadata.title).toBeDefined();
+      expect(faker.rawDefinitions.metadata!.title).toBeDefined();
```
xDivisionByZerox: I mean this literally checks for the value being defined. No need to assert non-null.

```suggestion
      expect(faker.rawDefinitions.metadata?.title).toBeDefined();
```

test/vitest-extensions.ts
```diff
@@ -11,6 +11,7 @@ expect.extend({
     return {
       pass: uniqueDuplication.length !== 0,
       message: () =>
+        // @ts-expect-error: `isNot` is incorrectly inferred as `function`.
```
xDivisionByZerox: Maybe default `isNot` to false in the destruction? But I don't really get why the `this` scope is infered incorrectly. A workaround is no longer required in TS  5.2.x, tho.

xDivisionByZerox (changes requested):

src/locales/ar/date/weekday.ts
ST-DDT: The null is only supported on the second aka entry level. Everything below needs extra handling. We cannot extend the proxying downward because we also return complex objects.

docs/.vitepress/versions.ts
ST-DDT: The list is only there if there is at least one entry. So it is always there.
Not sure about the filter code that doesnt filter and only serves TS types.

scripts/apidoc/fakerClass.ts
ST-DDT: I also considered this, but if there is a constructor without signature, I want to know that. Aka it shouldnt be possible unless typedoc changes their internal API.

src/modules/commerce/index.ts
ST-DDT: This can never happen or only if the code itself is wrong. Why write code that would hide the bug and return potentially bad values?

src/modules/date/index.ts
ST-DDT: The whole deprecation block will be gone in v9.

src/modules/date/index.ts
ST-DDT: Sure, but then there wouldnt be a hint which data are actually missing.

test/faker.spec.ts
ST-DDT: AFAICT the ! is not an assertion, its a not null definition expression.
Or am I wrong?

cypress/e2e/api.cy.ts
ST-DDT:
```suggestion
        const link = anchor.attr('href')?.split('#')[0] ?? 'MISSING';
```

test/vitest-extensions.ts
ST-DDT: TS claims that isNot is a function and always there thus the expression resolves to true always. But it is not.

test/faker.spec.ts
xDivisionByZerox: The name of the operator is [Non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). It does only assert in a TS context tho.

The problem is that `faker.rawDefinitions.metadata` might be `null` or `undefined`. This will propably never happen, but TS doesn't know that. In general you should not use the the NNA operator. See [@typescript-eslint/no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion/)

test/vitest-extensions.ts
xDivisionByZerox: Thats why I proposed to default the the property in the destruction aboth: `const { isNot = false } = this;`. This would let TypeScript infer the type as `function | false` which ill not always resolve to true.

test/faker.spec.ts
ST-DDT: In the end it doesn't hurt to change this. Done.

test/vitest-extensions.ts
ST-DDT: Thanks for the explanation. Changed.

docs/.vitepress/versions.ts
ST-DDT: I found a way to avoid this code.

test/modules/finance_iban.spec.ts
xDivisionByZerox: Yes. We could consider removing this lines as they only check that we really implemented what we implemented. That would happen in another PR tho.

src/locales/ar/date/weekday.ts
xDivisionByZerox: We should extend the `abbr` property with a null parameter and include a comment about the condition which is required for it to possible be null. If those values ARE null the functions relying on the data should throw an error. Similar to the locale proxy.

src/modules/commerce/index.ts
xDivisionByZerox: We will reconsider the implementation in a different PR.

src/modules/date/index.ts
xDivisionByZerox: Will be considered by solving https://github.com/faker-js/faker/pull/2435#discussion_r1349312186.

src/modules/date/index.ts
xDivisionByZerox: Will be considered by solving https://github.com/faker-js/faker/pull/2435#discussion_r1349312186.

src/modules/finance/index.ts
ST-DDT: Fixed

src/locales/ar/date/weekday.ts
ST-DDT: Fixed

src/modules/date/index.ts
ST-DDT: Fixed

src/modules/date/index.ts
ST-DDT: Fixed.

ST-DDT:
Ready for review again. 😉

src/locale-proxy.ts
```diff
@@ -54,6 +54,34 @@ export function createLocaleProxy(locale: LocaleDefinition): LocaleProxy {
...
+ * Checks that the value is not null or undefined and throws an error if it is.
+ *
+ * @param value The value to check.
+ * @param path The path to the locale data.
+ */
+export function assertLocaleData<T>(
+  value: T,
+  ...path: string[]
+): asserts value is NonNullable<T> {
+  if (value == null) {
```
xDivisionByZerox: Is this outer `if` really required?

ST-DDT: No, it is not. I just considered it as the fast path, but I'm not sure whether it has any performance implications.

xDivisionByZerox: I'd say it is obsolete here TBH.

ST-DDT: Changed.

src/modules/helpers/index.ts
```diff
@@ -422,6 +422,7 @@ export class SimpleHelpersModule {
       const quantifierMin: string = token[2];
       const quantifierMax: string = token[3];
       const quantifierSymbol: string = token[4];
+      const tokenIndex: number = token.index!;
```
xDivisionByZerox: I thought a little bit about this. Can we not default this to `0`? Should we perhaps exit the `while` loop if it doesn't exist? Thats the two solutions I can think off. I dont really like using the NNAO here.

ST-DDT: We could switch from `string.match(regex)` to `regexp.exec(string)`.
That would fix that issue as well.

I don't think defaulting to 0 would do us any good. Maybe -1 or NaN so it explodes at the next usage instead doing something random.

ST-DDT: The index is always present, when the regex does not use the global flag.

xDivisionByZerox: > We could switch from string.match(regex) to regexp.exec(string).

How would you retrieve the index in that case? I'm a bit lost.

> I don't think defaulting to 0 would do us any good. Maybe -1 or NaN so it explodes at the next usage instead doing something random.

Well that works for me as well. ~~I would propably prefer `-1` tho.~~ We should use `NaN` as negative numbers behave differently on `String.prototype.slice`.

> The index is always present, when the regex does not use the global flag.

Where did you find this information? I was searching for this when I wrote my previous comment but didn't find anything. :(

ST-DDT: > > We could switch from string.match(regex) to regexp.exec(string).
> 
> How would you retrieve the index in that case? I'm a bit lost.

In that case the index property is always there.

ST-DDT: > > The index is always present, when the regex does not use the global flag.
> 
> Where did you find this information? I was searching for this when I wrote my previous comment but didn't find anything. :(

Try and error + some stackoverflow.

ST-DDT: FFR: https://typescript-eslint.io/rules/prefer-regexp-exec/

ST-DDT:
See #2466 for a regex related NNA `!` relief.

- #2466 

(Technically not blocking, but might be worth merging first)

ST-DDT:
Ready to review again. 😉

src/modules/helpers/index.ts
```diff
@@ -1121,7 +1121,7 @@ export class SimpleHelpersModule {
     ) => RecordKey,
   >(
     method: TMethod,
-    args: Parameters<TMethod> = [] as Parameters<TMethod>,
+    args: Parameters<TMethod> = [] as unknown as Parameters<TMethod>,
```
xDivisionByZerox, line 1123: We could probably "fix" this casting by some generic making but as the function is deprecated anyway I'd not waste time optimizing this.

xDivisionByZerox (dismissed):

src/modules/helpers/index.ts
ST-DDT: Having a default here is basically a lie because some parameters may be required... But that is something for a different PR if at all.

xDivisionByZerox (dismissed):

src/modules/date/index.ts
```diff
@@ -1286,6 +1287,32 @@ export class DateModule extends SimpleDateModule {
...
   }
 }
+
+/**
+ *
+ * @param options asd
+ * @param options.deprecated as
+ * @param options.replacement asd
+ */
+export function demo(
```
xDivisionByZerox: This is propably not supposed to be here?

src/modules/date/index.ts
```diff
@@ -1286,6 +1287,32 @@ export class DateModule extends SimpleDateModule {
...
+  } = {}
+): boolean {
+  const {
+    // eslint-disable-next-line deprecation/deprecation
+    deprecated = false,
+    // eslint-disable-next-line deprecation/deprecation
+    replacement = deprecated,
+  } = options;
+  return replacement;
+}
```
ST-DDT:
```suggestion

```

src/modules/date/index.ts
ST-DDT: Yeah, I used that code to test some cases for the deprecation warnings that for some reason appeared as part of this PR.

- https://github.com/gund/eslint-plugin-deprecation/issues/74
- https://github.com/gund/eslint-plugin-deprecation/issues/75

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

ST-DDT:
**Team Decision**

This is largely unchanged since creation and has been approved for more than a week (if you ignore the minor merge conflicts).

Issue #2439, opened 2023-10-07
Check eslint-plugin-unicorn rules to enable/configure/disable


ST-DDT:
Followup on #2418 

- #2418 

Currently all the following rules are disabled because they cause issues in the code:

## Undecided

| Rule | Count | Status/PR |
| --- | ---: | --- |
| [`unicorn/prevent-abbreviations`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prevent-abbreviations.md) | 232 | 🤔 - #3320 |

Last updated: 2023-10-07

## Rejected

| Rule | Count | Status/PR |
| --- | ---: | --- |
| [`unicorn/import-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/import-style.md) | 8 | 🔴 - #2901 - doesn't do anything for us |
| [`unicorn/no-nested-ternary`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-nested-ternary.md) | N/A | 🔴 - incompatible with prettier |
| [`unicorn/no-null`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-null.md) | N/A | 🔴 - incompatible with TypeScript | 
| [`unicorn/no-object-as-default-parameter`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-object-as-default-parameter.md) | 6 | 🔴 - [Unicorn-Issue](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2199) |
| [`unicorn/no-zero-fractions`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-zero-fractions.md) | 17 | 🔴 - #2453 |
| [`unicorn/numeric-separators-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/numeric-separators-style.md) | 131 | 🔴 - #2815 |
| [`unicorn/number-literal-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/number-literal-case.md) | N/A | 🔴 - incompatible with prettier |
[...]

PR #2440, opened 2023-10-07
infra(unicorn): no-hex-escape

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-hex-escape`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-hex-escape.md) lint rule.


xDivisionByZerox (dismissed):
> Enforces a convention ]...[ for consistency and clarity.

Good think unicode encoding is so much more clear for me as a human than hex encoding xD

ejcheng (dismissed):

matthewmayer (changes requested):
I don't find either very readable. Could we define a constant for a non breaking space like

const nbsp = "\ua00"

ST-DDT:
> I don't find either very readable. Could we define a constant for a non breaking space like
> 
> const nbsp = "\ua00"

Done

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2443, opened 2023-10-07
infra(unicorn): prefer-object-from-entries

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/prefer-object-from-entries`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-object-from-entries.md) lint rule.


xDivisionByZerox (approved):
Makes sense to get rid of the "unreadable" `reduce` functions, I guess. No strong feelings about this.

ejcheng (approved):

matthewmayer (approved):

PR #2444, opened 2023-10-07
docs: check and improve handling of duplicate tags

ST-DDT:
~~Ref: #2439~~

- ~~#2439~~

---

Preparation work for enableing the [`unicorn/require-array-join-separator`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/require-array-join-separator.md) lint rule.


matthewmayer:
This is a behavior change no? The default for join is ,

ST-DDT:
Technically yes. Practically not, because we only ever have one.

matthewmayer:
If we did have more than one should that throw an error?

ST-DDT:
> If we did have more than one should that throw an error?

Changed. It looks like there was one method with two `@throws` tags, that I have missed in my earlier checks.

matthewmayer:
This PR should probably be renamed now then

ST-DDT:
> This PR should probably be renamed now then

The title is already very long so appending it might be hard.
Do you have a suggestion?

matthewmayer:
Split the changes about allowing one one of each tag to a seperate PR?

ST-DDT:
IMO this only does a single thing, what should I split here?

I renamed it to reflect its doc impact

xDivisionByZerox:
Preview of the changes on multiple `@throw` tags:

image

---

I don't know if I'm the biggest fan as you can't tell right now that these are 2 distinct throw situations. But it is better than not having them at all.

fisker:
This end up with nothing related to `require-array-join-separator` at all... LOL

 `require-array-join-separator` mean to make code easier for beginners and for consistency.

xDivisionByZerox:
> This end up with nothing related to `require-array-join-separator` at all... LOL
> 
>  `require-array-join-separator` mean to make code easier for beginners and for consistency.

True. Enabling the rule in this PR is kind of missleading.

.eslintrc.js
```diff
@@ -62,7 +62,6 @@ module.exports = defineConfig({
     'unicorn/prefer-ternary': 'off',
     'unicorn/prefer-top-level-await': 'off',
     'unicorn/prevent-abbreviations': 'off',
```
ST-DDT:
```suggestion
    'unicorn/prevent-abbreviations': 'off',
    'unicorn/require-array-join-separator': 'off',
```

scripts/apidoc/typedoc.ts
```diff
@@ -382,5 +414,5 @@ export function extractThrows(reflection?: CommentHolder): string | undefined {
  * @returns The contents of the `@since` tag.
  */
 export function extractSince(reflection: CommentHolder): string {
```
xDivisionByZerox, line 416: How can this return `string` if `extractSingleTagContent` returns `string | undefined`? 🤔

ST-DDT: Because https://github.com/faker-js/faker/pull/2435 is not yet merged.

- #2435

xDivisionByZerox (approved):

PR #2451, opened 2023-10-09
infra(unicorn): prefer-array-some

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/prefer-array-some`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-some.md) lint rule.


test/modules/system.spec.ts
```diff
@@ -405,13 +405,11 @@ describe('system', () => {
         );
 
         it('should be able to return non-standard cron expressions', () => {
-          const validResults = [...'0123456789'.split(''), '*', '@'];
+          const validResults = new Set('0123456789*@');
           expect(
             faker.system.cron({ includeNonStandard: true })[0],
```
matthewmayer: This whole test seems awkwardly written. If we're only testing the first char maybe refactor to use startsWith?

matthewmayer (changes requested):

test/modules/system.spec.ts
ST-DDT: I'm not sure how to make it less awkward.

Maybe like this:

````ts
const validPrefixes = [...'0123456789*@'];
const actual = faker.system.cron({ includeNonStandard: true });
expect(
    validPrefixes.some((prefix) => actual.startsWith(prefix))
).toBe(true);
````

?

ST-DDT: Alternatively:

````ts
const validResults = new Set('0123456789*@');
const actual = faker.system.cron({ includeNonStandard: true })[0];

expect(validResults).toContain(actual);
````

ST-DDT: Done

xDivisionByZerox (dismissed):

ejcheng (dismissed):

matthewmayer (dismissed):

ST-DDT:
Ready for review

xDivisionByZerox (approved):

PR #2453, opened 2023-10-09
infra(unicorn): no-zero-fractions

ST-DDT:
Ref: #2439

- #2439

---

Permanently disabled the [`unicorn/no-zero-fractions`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-zero-fractions.md) lint rule.


matthewmayer:
I don't personally like this rule even if 1.0 and 1 mean the same thing in JS 1.0 hints that you are dealing with floating point numbers.

ST-DDT (commented):
**Team Decision**

No, we don't want this because the zero fractions point to decimal values, even if it doesn't have a runtime impact.

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2457, opened 2023-10-09
feat(number): adding support for exclude in number.int()

olafurw:
I needed a way to generate a set of numbers that didn't already include values I already had in my database. This can be done in userland but can be bug prone in regards to the set being exhausted. So I thought it would be nice as a feature and I decided to give making a PR a try.

It does add a bit of work to the function that can be done only if the length of the array is > 0, but I wasn't sure on your stance on that kind of change, since I felt it split the function in two sections.

### DIFF
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -28,8 +28,10 @@ export class NumberModule {
    * @param options Maximum value or options object. Defaults to `{}`.
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `Number.MAX_SAFE_INTEGER`.
+   * @param options.exclude An array of numbers to exclude from the generated range. Defaults to `[]`.
    *
    * @throws When options define `max < min`.
+   * @throws When exclude array covers all possible values in the range.
    *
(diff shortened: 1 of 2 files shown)


ST-DDT:
Thanks for your feature proposal.


---

Have you tried using faker.helpers.unique instead?
https://fakerjs.dev/api/helpers.html#unique
(Note: It will likely get "feature trimmed" a bit in v9)

olafurw:
I saw it and I noticed the deprecated sign so I didn't think about it more since I assumed you were going to remove it.

matthewmayer:
In theory you could add an exclude to almost any faker method. I don't think int() deserves special treatment.

ST-DDT:
I also considered adding a helper method such as:

`generateMatching(factory: () => T, condition: (t: T) => boolean, { onError: 'continue' | 'fail', maxRetries: ...}): T`

olafurw:
> In theory you could add an exclude to almost any faker method. I don't think int() deserves special treatment.

Yup, I wanted feedback on this one before going forwards (or not if it would not be accepted)

ST-DDT:
**Team Decision**

- We don't want this. This is basically the `unique` use-case with slight differences.
- We might add other methods that "generate repeatedly until condition" given user interest

PR #2459, opened 2023-10-09
infra(unicorn): no-instanceof-array

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-instanceof-array`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-instanceof-array.md) lint rule.


src/modules/system/index.ts
```diff
@@ -179,22 +179,15 @@ export class SystemModule {
...
-      }
-    });
+    if (typeof mimeType === 'string') {
+      return this.faker.helpers.arrayElement(mimeTypes[mimeType].extensions);
+    }
 
+    const extensionSet = new Set(
+      Object.values(mimeTypes).flatMap(({ extensions }) => extensions)
+    );
     const extensions = Array.from(extensionSet);
```
ST-DDT: This could potentially be altered to `faker.helpers.objectValue(mimeTypes).extensions`.
(But that would change the generated values)

xDivisionByZerox: We can do that in a different PR. Good catch tho.

xDivisionByZerox (approved):

PR #2460, opened 2023-10-10
feat(internet): Add RFC 5737 testing IPv4 support

thomas-phillips-nz:
Adds support for generating random testing-safe IPv4 addresses according to [RFC 5737](https://www.rfc-editor.org/rfc/rfc5737)

Sometimes you need to generate an IP address but don't want the system being tested to try and resolve or send traffic to the IP

I guess this could be added as an option to the existing ipv4() function?


xDivisionByZerox:
> [...] Sometimes you need to generate an IP address but don't want the system being tested to try and resolve or send traffic to the IP [...]

Please don't do that with ANY data generated by Faker (emails, phone numbers, etc). This should include IP addresses as well.
See our README:

https://github.com/faker-js/faker/blob/350dfbf2c57f023a78567da60bf4792ebd1a41f7/README.md?plain=1#L46-L48

thomas-phillips-nz:
Totally fine for this to be closed if you think it's unnecessary with that guidance in the readme

ST-DDT:
**Team Proposal**

How about the following api?

````ts
ipv4(options: {ipRange: LiteralUnion<'TEST-NET-1' | 'TEST-NET-2' | 'TEST-NET-3'>}) {
  let { ipRange } = options;
  if (ipRange != null) {
    if (ipRange === 'TEST-NET-1')
      ipRange  = '192.0.2.0/24'
    ...

    const [ipBase, mask] = ipRange.split('/');
    const maskInt = +mask;
    const maskBit = 2 ** mask -1;
    const ipBaseInt = ipToInt32(ipBase) & ^maskBit;
    const ipSuffix = faker.number.int(maskBit);
    return toIpv4(ipBaseInt + ipSuffix)
  } else {
    // current impl
  }
}
````

ST-DDT:
@thomas-phillips-nz What do you think of our proposal?

ST-DDT:
Superseded by https://github.com/faker-js/faker/pull/2992

- https://github.com/faker-js/faker/pull/2992

Thanks for suggesting this feature.

PR #2461, opened 2023-10-10
infra(unicorn): no-array-for-each

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-array-for-each`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md) lint rule.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -56,7 +56,6 @@ module.exports = defineConfig({
     'unicorn/filename-case': 'off',
     'unicorn/import-style': 'off',
     'unicorn/no-array-callback-reference': 'off',
-    'unicorn/no-array-for-each': 'off',
     'unicorn/no-array-reduce': 'off',
     'unicorn/no-await-expression-member': 'off',
     'unicorn/no-for-loop': 'off',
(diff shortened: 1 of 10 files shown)


ST-DDT:
I'm not sure about this one.

matthewmayer (approved):

xDivisionByZerox (approved):
Personally, I prefer `for..of` loops over `Array.prototype.forEach`. It make the code more consistent as you can use `for..of` for all iterables and not only arrays. Additionally, I've read some time ago that they were supposed to be more performant, but according to [this](https://stackoverflow.com/a/50844413/14392422) StackOverflow comment this is no longer the case. Last argument I can bring for the `for..of` loop is that you dont leave the `this` scope. This resolves potential issues with hoisting or `this`-binding.

xDivisionByZerox (approved):
Personally, I prefer `for..of` loops over `Array.prototype.forEach`. It make the code more consistent as you can use `for..of` for all iterables and not only arrays. Additionally, I've read some time ago that they were supposed to be more performant, but according to [this](https://stackoverflow.com/a/50844413/14392422) StackOverflow comment this is no longer the case. Last argument I can bring for the `for..of` loop is that you dont leave the `this` scope. This resolves potential issues with hoisting or `this`-binding.

PR #2462, opened 2023-10-10
infra(unicorn): consistent-destructuring

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/consistent-destructuring`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-destructuring.md) lint rule.


ST-DDT:
The current unique method implementation is bad... we should un-recursive it and cleanup the options in v9.

matthewmayer (approved):

xDivisionByZerox (approved):

PR #2469, opened 2023-10-10
infra(unicorn): escape-case

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/escape-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/escape-case.md) lint rule.


ST-DDT:
I'm not sure about this rule, because our numbers use lowercase hex characters.

xDivisionByZerox:
> [...] our numbers use lowercase hex characters.

Maybe we should discuss if that is the expected behavior. There is propably a ISO or RFC that standardizes representation of those values.

xDivisionByZerox (approved):
I don't see a problem with the rule as it only verifies static escaped characters. I see that there would be a mismatch between the values we use in our code base and the values we provide for users, but I would not consider this a different topic of discussion.

Issue #2361, opened 2023-08-31
number.float is unable to return 1/max exactly


ST-DDT:
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md).
- [X] Read the [docs](https://fakerjs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/faker-js/faker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Faker issue and not related to a combination with another package.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/faker-js/faker/discussions) or join our [Discord Chat Server](https://chat.fakerjs.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.
- [X] I am willing to provide a PR.

### Describe the bug

faker.number.float() is unable to return 1.

### Minimal reproduction code

````ts
it('should be able to return 1', () => {
  // @ts-expect-error: access private member field
  const mersenne = faker._mersenne;
  mersenne.next = () => 0.9999999999999999; // based on the definition of mersenne.next()
  const actual = faker.number.float();
  expect(actual).toBe(1); // 💥 fails
});
````

### Additional Context

Originally part of:

- https://github.com/faker-js/faker/pull/2357

### Environment Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 Intel(R) Core(TM) i7-7820X CPU @ 3.60GHz
    Memory: 47.42 GB / 63.69 GB
  Binaries:
    Node: 18.17.1 - C:\Program Files\nodejs\node.EXE
    npm: 9.7.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.5.1 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Spartan (44.22621.2134.0), Chromium (116.0.1938.62)
    Internet Explorer: 11.0.22621.1
```

### Which module system do you use?

- [ ] CJS
- [X] ESM

### Used Package Manager

pnpm

dubzzz:
It seems to be the usual behaviour for a random number generator generating values between 0 (included) and 1 (not included). For instance, `Math.random` does exactly the same.

[...]

PR #2470, opened 2023-10-11
test(mersenne): add tests for value ranges

ST-DDT:
This PR adds tests asserting that our `MersenneTwister19937` implementation can generate all expected values.

Partially covers #2361 with tests

- #2361


test/faker.spec.ts
```diff
@@ -113,14 +113,14 @@ describe('faker', () => {
       faker.seed(1);
 
       const actual = faker.animal.cat();
-      expect(actual).toBe('Korat');
+      expect(actual).toMatchInlineSnapshot('"Korat"');
```
ST-DDT: The inline snapshots are "required" because the value is mersenne and seed specific and the inline snapshots are easier to update.

test/mersenne.spec.ts
ST-DDT, line 1: Renamed to `test/internal/mersenne.spec.ts` and extended with Twister specific tests.

test/internal/mersenne.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      const actual = twister.genrandReal2();
+      expect(actual).toBe(0);
+    });
+
+    it('should be able to return almost 1', () => {
+      const twister = newTwister();
+      twister.genrandInt32 = () => 2 ** 32 - 1;
+      const actual = twister.genrandReal2();
+      expect(actual).toBe(TWISTER_32CO_MAX_VALUE);
+    });
```
xDivisionByZerox: I don't understand what these tests are supposed to do.

test/internal/mersenne.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    beforeAll(() => {
+      randomizer.seed(seed);
+    });
+
+    describe('next', () => {
+      it('should return random number from interval [0, 1)', () => {
+        const actual = randomizer.next();
+
+        expect(actual).toBeGreaterThanOrEqual(0);
+        expect(actual).toBeLessThanOrEqual(MERSENNE_MAX_VALUE);
```
xDivisionByZerox, line 131: Can we not simply do `toBeLessThan(1)`? Looking at the description that would be more clear for me than having a "random" float here.

test/modules/date.spec.ts
```diff
@@ -801,22 +801,26 @@ describe('date', () => {
       faker.setDefaultRefDate(() => new Date(Date.UTC(2020, 0, 1)));
       faker.seed(20200101);
       const date = faker.date.past();
-      expect(date).toEqual(new Date('2019-02-25T21:52:41.824Z'));
+      expect(date).toBeInstanceOf(Date);
+      expect(date).toMatchInlineSnapshot('2019-02-25T21:52:41.824Z');
```
xDivisionByZerox: Why did you change these tests? They seem unrelated to the actual change.

test/modules/helpers.spec.ts
```diff
@@ -1253,6 +1245,24 @@ Try adjusting maxTime or maxRetries parameters for faker.helpers.unique().`)
     }
   );
 
+  describe('uniqueArray()', () => {
```
xDivisionByZerox: Is this change required for the test addition?

test/internal/mersenne.spec.ts
ST-DDT: They test that we can generate values really close to `1` instead of just `0.9`.

xDivisionByZerox: But you mock the implementation. So you test that the mock returns something what you just said it should return. 🤔

ST-DDT: We test for that value 16 lines above.

ST-DDT: I added a comment to the mock why it is there.

test/internal/mersenne.spec.ts
ST-DDT: Added the secondary verification as well.

test/modules/date.spec.ts
ST-DDT: Reverted, stashed for a later PR.

test/modules/helpers.spec.ts
ST-DDT: Reverted and stashed for a later PR.

test/faker.spec.ts
ST-DDT: Reverted and stashed for a later PR.

xDivisionByZerox (approved):

ejcheng (approved):

PR #2476, opened 2023-10-13
refactor(locale)!: remove location data index-value-binding

xDivisionByZerox:
Removes the requirement for direction locale data lists to be sorted in a specific order.


ST-DDT:
Looks good to me.
Not sure whether we should push this in v8 or v9.

ST-DDT (dismissed):

src/locales/fa/location/direction.ts
```diff
@@ -1,10 +1,6 @@
...
-  'جنوب',
-  'غرب',
-  'شمال شرق',
-  'شمال غرب',
-  'جنوب شرق',
-  'جنوب غرب',
-];
+export default {
+  cardinal: ['شمال', 'شرق', 'جنوب', 'غرب'],
+  cardinal_abbr: [],
```
matthewmayer: Note this and similar changes are a behavior change, previously
fakerFA.location.direction({abbreviated:true}) would fall back to English and return say NE

Now it throws
[Error]: Cannot get value from empty dataset.

I think thats OK but it should probably be considered a breaking change for this reason

matthewmayer (changes requested):

src/locales/fa/location/direction.ts
matthewmayer: Alternatively if there is no data for cardinal_abbr etc then the abbreviated option should be ignored and the full name returned.

ST-DDT: I think we should go for breaking then.

xDivisionByZerox: Good catch. I left them like this instead of making the subset undefined  or even null because I think that this data has been forgotten to be implemented. Those values are part of the legacy package. I was not able to figure out if abbreviated content is not applicable in those locales. By throwing people with knowledge about these locales can tell us the expected behavior:
- abbreviated is same as wide context
- abbreviated context does not exist in those locales
- implement the correct abbreviated data

matthewmayer: agree, though it would be better if it threw the helpful "The locale data for.... are missing in this locale" error rather than the mysterious "Cannot get value from empty dataset" error

xDivisionByZerox: In that case, I could wrap the current type in the `LocaleEntry` type to allow `undefined` (and `null`) values. `undefined` should throw the error you suggest. I would be fine with that.

matthewmayer: Note since N, E, S, W are included in CLDR (for use in geographical coordinates) its fairly easy to look up what the equivalent of "N", "E", "S", "W" is in most languages e.g.:

https://www.unicode.org/cldr/charts/42/by_type/units.coordinates.html#2e784791aa57e39c

matthewmayer: In examples like fa or ja where there is no abbreviated version, what should we do? In that case it seems more sensible to repeat the full version in the abbreviated field.

matthewmayer: Since data has now been added for the missing locales I would no longer consider this breaking.

ST-DDT (approved):

ST-DDT:
Since we change the layout of our locale data, I think we should merge this in v9.0.

matthewmayer (dismissed):

ST-DDT:
This probably needs a check whether any new locales/locale data have been added in the meantime.

ST-DDT:
Does this need a breaking change notice?

src/definitions/location.ts
```diff
@@ -65,7 +65,31 @@ export type LocationDefinition = LocaleEntry<{
...
+
+    /**
+     * The abbreviated names of ordinal compass directions.
+     * Ordinal directions are combinations of cardinal directions.
+     */
+    ordinal_abbr: string[];
+  };
 
   /**
    * The abbreviated names of the compass directions.
```
ST-DDT: The old `direction_abbr: string[]` needs to be removed.

ST-DDT (changes requested):
I checked that no new files have been added in the meantime.

Shinigami92 (changes requested):
`docs/guide/upgrading_v9/2476.md` file required

xDivisionByZerox:
Looking at #2505, would it make sense to "flatten" this definitions as well? So instead of `direction` object with nested properties, we'd have 4 files/4properties on the location definition object.

ST-DDT:
> Looking at #2505, would it make sense to "flatten" this definitions as well? So instead of `direction` object with nested properties, we'd have 4 files/4properties on the location definition object.

We could do that, but these files are so small, that it would be harder to review the files if split up.
IMO we keep them in a single file.

src/definitions/location.ts
xDivisionByZerox: Done

xDivisionByZerox:
I put "needs decision" as I'd like to discuss [my last comment](https://github.com/faker-js/faker/pull/2476#issuecomment-1987025765) in the next team meeting.

ST-DDT (approved):

Shinigami92 (approved):

Shinigami92 (approved):

ST-DDT:
**Team Decision**

We will not flatten the definitions in the same way as #2505 .

Reason:
- Being a nested object ensures that all values are always present, as the type requires it. On the other hand entries in categories are optional by force and thus it might lead to cases where `abbr` is set but the full value is missing.

PR #2478, opened 2023-10-14
chore(release): 8.2.0

xDivisionByZerox:


ST-DDT:
Are there no changes to the pnpm-lock file?

xDivisionByZerox:
I repeated the release steps and got infact changes in the lock file this time. They must have been disappeared due to an error in the `clear` script I got during the first run.

Problem is that these changes introduce the same problems as in #2184.

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
```
ST-DDT: You can fix that by manually reverting these.

ST-DDT (approved):

pnpm-lock.yaml
xDivisionByZerox: I see. I thought they would be put  in again when the CI runs.

prisis (approved):

PR #2479, opened 2023-10-15
infra(unicorn): no-array-reduce

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-array-reduce`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md) lint rule.


test/all_functional.spec.ts
```diff
@@ -10,12 +10,33 @@ const IGNORED_MODULES = new Set([
...
+
+function getMethodNamesOf(module: keyof Faker): string[] {
+  return Object.keys(fakerEN[module]).filter(isMethodOf(module));
 }
 
-function isMethodOf(mod: string) {
-  return (meth: string) => typeof fakerEN[mod][meth] === 'function';
+function isMethodOf(module: string): (method: string) => boolean {
+  return (method: string) => typeof fakerEN[module][method] === 'function';
 }
```
xDivisionByZerox: I'd expect an object (the module object itself) as a parameter in the `isMethodOf` function. Currently, these two functions are tightly coupled. If you change the main object (fakerEN) in one function you have to change it in the other one too.

.eslintrc.js
```diff
@@ -36,6 +36,7 @@ module.exports = defineConfig({
 
     'deprecation/deprecation': 'error',
 
+    'unicorn/no-array-reduce': 'off', // not easy to avoid for new contributors
```
xDivisionByZerox: I don't agree with you here. Every `.reduce()` function can be represented as can be represented by a `for` loop. On the one problem we have in our code base you even wrote a comment on how this could be refactored. Using descriptive functions instead of a `.reduce()` should be preferred in my opinion.

src/modules/word/filterWordListByLength.ts
```diff
@@ -13,6 +13,7 @@ const STRATEGIES = {
     wordList: ReadonlyArray<string>,
     length: { min: number; max: number }
   ): string[] => {
+    // This could be replaced with Map.groupBy once it is available
```
xDivisionByZerox: Why not write a helper function for now?

```ts
function groupBy<TVal, TKey>(
  items: Iterable<TVal>,
  keyEvaluator: (value: TVal) => TKey
): Map<TKey, TVal[]> {
  const map = new Map<TKey, TVal[]>();
  for (const item of items) {
    const key = keyEvaluator(item);
    map.set(key, [...(map.get(key) ?? []), item]);
  }

  return map;
}
```

Or if you prefer object maps:
```ts
function groupBy<TVal, TKey extends string | number | symbol>(
  items: Iterable<TVal>,
  keyEvaluator: (value: TVal) => TKey
): Record<TKey, TVal[]> {
  const map = {} as Record<TKey, TVal[]>;
  for (const item of items) {
    const key = keyEvaluator(item);
    map[key] = [...(map[key] ?? []), item];
  }

  return map;
}
```

xDivisionByZerox (changes requested):

test/all_functional.spec.ts
ST-DDT: Like this?

.eslintrc.js
ST-DDT: Changed.

src/modules/word/filterWordListByLength.ts
ST-DDT: Done.

test/all_functional.spec.ts
xDivisionByZerox: Yes, thank you. This makes the function more pure. 👍

test/all_functional.spec.ts
```diff
@@ -10,12 +10,36 @@ const IGNORED_MODULES = new Set([
...
+  return !IGNORED_MODULES.has(moduleName);
+}
+
+function getMethodNamesOf(module: Faker[keyof Faker]): string[] {
+  return Object.keys(module).filter(isMethodOf(module));
 }
 
-function isMethodOf(mod: string) {
-  return (meth: string) => typeof fakerEN[mod][meth] === 'function';
+function isMethodOf(module: Faker[keyof Faker]): (method: string) => boolean {
```
xDivisionByZerox: Nitpick: The "module" technically could be any object, which allows the function to be used in a more generic context.

```suggestion
function isMethodOf(module: object): (method: string) => boolean {
```

xDivisionByZerox (dismissed):
Remaining suggestions are optional. Thats why I'll already approve.

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

PR #2485, opened 2023-10-17
refactor: simplify module creation

ST-DDT:
Simplifies the modules by reusing the constructors from a shared module base.

This is a v8.x only feature, that will likely be obsolete in v9.0.

What do you think?


src/modules/module-base.ts
```diff
@@ -0,0 +1,25 @@
+import type { Faker } from '../faker';
+import { bindThisToMemberFunctions } from '../internal/bind-this-to-member-functions';
+import type { SimpleFaker } from '../simple-faker';
+
+/**
+ * Base class for all modules that use a `SimpleFaker` instance.
+ *
+ * @internal
+ */
+export abstract class SimpleModuleBase {
```
xDivisionByZerox: Should this file not be located in `src/internal/` if it is marked as internal?

xDivisionByZerox (changes requested):
Can we add extend the test factory (or write a separate one) to include automatic test for these `this` bindings on all methods? Currently I'm not 100% sure that I can do. 

```ts
const method = faker[moduleKey][methodKey]; 
method();
```

Additionally, while being a nice little addition I have a hard time seeing the advantage of introducing it now.

ST-DDT:
> Can we add extend the test factory (or write a separate one) to include automatic test for these `this` bindings on all methods? Currently I'm not 100% sure that I can do.
> 
> ```ts
> const method = faker[moduleKey][methodKey]; 
> method();
> ```

We already have that:

https://github.com/faker-js/faker/blob/a1936938574f574ecc137d9f718a48a2256e49c9/test/support/seededRuns.ts#L146-L147

src/modules/module-base.ts
ST-DDT: Done

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2487, opened 2023-10-20
infra(unicorn): better-regex

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/better-regex`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/better-regex.md) lint rule.


test/modules/internet.spec.ts
```diff
@@ -297,7 +297,7 @@ describe('internet', () => {
           const [prefix, suffix] = email.split('@');
 
           expect(prefix).toMatch(
-            /^Mike((\d{1,2})|([.!#$%&'*+-/=?^_`{|}~]Smith\d{1,2})|([.!#$%&'*+-/=?^_`{|}~]Smith))/
+            /^Mike((\d{1,2})|([!#$%&'*+-/=?^_`{|}~]Smith\d{1,2})|([!#$%&'*+-/=?^_`{|}~]Smith))/
```
ST-DDT, line 300: For those of you that wonder like me where the `.` went:

`+-/` **===** `+` - `/` **===** `+`, `-`, `.`, `/`

test/modules/color.spec.ts
```diff
@@ -98,21 +98,21 @@ describe('color', () => {
       describe(`rgb({ format: 'css' })`, () => {
         it('should return a random rgb color in css format', () => {
           const color = faker.color.rgb({ format: 'css' });
-          expect(color).match(/^(rgb\([0-9]{1,3}, [0-9]{1,3}, [0-9]{1,3}\))$/);
+          expect(color).match(/^rgb\((?:\d{1,3}, ){2}\d{1,3}\)$/);
```
matthewmayer, line 101: this seems harder to read than the original

test/modules/color.spec.ts
```diff
@@ -98,21 +98,21 @@ describe('color', () => {
...
         });
       });
 
       describe(`rgb({ format: 'binary' })`, () => {
         it('should return a random rgb color in binary format', () => {
           const color = faker.color.rgb({ format: 'binary' });
-          expect(color).match(/^([01]{8} [01]{8} [01]{8})$/);
+          expect(color).match(/^(?:[01]{8} ){2}[01]{8}$/);
```
matthewmayer, line 108: this seems harder to read than the original

test/modules/datatype.spec.ts
```diff
@@ -400,33 +400,33 @@ describe('datatype', () => {
         it('generates a valid UUID', () => {
           const UUID = faker.datatype.uuid();
           const RFC4122 =
-            /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+            /^[\da-f]{8}-[\da-f]{4}-[1-5][\da-f]{3}-[89ab][\da-f]{3}-[\da-f]{12}$/;
```
matthewmayer: i find [0-9a-f] easier to read and parse than [\da-f]

test/modules/location.spec.ts
```diff
@@ -184,7 +184,7 @@ describe('location', () => {
         it('returns random zipCode - user specified format', () => {
           let zipCode = faker.location.zipCode({ format: '?#? #?#' });
 
-          expect(zipCode).toMatch(/^[A-Za-z]\d[A-Za-z]\s\d[A-Za-z]\d$/);
+          expect(zipCode).toMatch(/^[a-z]\d[a-z]\s\d[a-z]\d$/i);
```
matthewmayer, line 187: this seems harder to read. if you're focusing on the regex part you might miss the i flag at first.

matthewmayer (commented):
In general these are improvements but sometimes it seems to be harder to read.

test/modules/internet.spec.ts
matthewmayer: that is SUPER confusing. A human reading this would never think that

test/modules/color.spec.ts
xDivisionByZerox: That means that the regex in general is too complex for that use case IMO. Splitting this one statement into multiple expect statements would be welcome:

```ts
const color = faker.color.rgb({ format: 'css' });
expect(color.startsWith('rgb')).toBe(true);
expect(color).toContain('(');
expect(color.endsWith(')')).toBe(true);

const values = color.substring(4, color.length - 1).split(',');
expect(values).toHaveLength(3);
for(const value of values) {
  const numeric = Number.parseInt(value, 10);
  expect(numeric).not.toBe(NaN);
  expect(numeric).toBeGreaterThanOrEqual(0);
  expect(numeric).toBeLessThanOrEqual(255);
}
```

test/modules/datatype.spec.ts
xDivisionByZerox: This should only do a `toSatisfy` check against `Validator.isUUID`.

test/modules/location.spec.ts
xDivisionByZerox: For me, this is easier to read. Would also argue that "if you focus on the regex part" you will read the flags since they are part of the regex. Could also write it via the RegExp constructor to make it more clear that is part of regex itself: `new RegExp(/.../, 'i')`.

ST-DDT: I agree with @xDivisionByZerox here. Having patterns that state both upper and lowercase raises the expectation/assumption that there is something case sensitive in there. I don't like the `new RegExp(/.../, 'i')` way though.

test/modules/datatype.spec.ts
ST-DDT: Changed

test/modules/internet.spec.ts
xDivisionByZerox: TBH, this regex pattern in general is super confusing. I just looked at it and noped out immediately.

ST-DDT:
**Team Decision**

- We will wait until after deprecation removals
- We will continue with the rule in smaller chunks.
- If regex patterns are too complex we should consider adding a comment what it does instead of chopping it in smaller pieces.

PR #2488, opened 2023-10-21
infra(security): switch to use workflow in target branch

CoolPlayLin:
# Why?

some contribution would modify the workflow file, when the event included `pull_request`, it would run its modified file directly. This would cause some safe issue because hacker can use it to get token via Github's vulnerability or to another bad things by hackers. And `pull_request` CANNOT get the secret token due to [Github's policy](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow). it will make following code fail directly
```yml
      - name: Comment
        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1
        with:
          script: |
            const script = require('${{ github.workspace }}/.github/workflows/commentCodeGeneration.js')
            await script(github, context, ${{ steps.generate.outcome == 'success' && steps.diff.outcome == 'success' }})
```

Use `pull_request_target` is a BETTER way to resolve it. It would let action run workflow file in target branch instead of modified workflow file. Because it CAN get the secret token, it can help the above code run succeed.

### DIFF
--- a/.github/workflows/check-mergable-by-label.yml
+++ b/.github/workflows/check-mergable-by-label.yml
@@ -1,7 +1,7 @@
 name: PR Labels
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - reopened
--- a/.github/workflows/ci-old.yml
+++ b/.github/workflows/ci-old.yml
@@ -1,0 +1,233 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - next
+  pull_request:
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  unit-test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node_version: [14, 16, 18, 20]
+        include:
+          - os: macos-latest
+            node_version: 16
+          - os: macos-latest
+            node_version: 18
+          - os: macos-latest
+            node_version: 20
+          - os: windows-latest
+            node_version: 16
+          - os: windows-latest
+            node_version: 18
+          - os: windows-latest
+            node_version: 20
+      fail-fast: false
+    timeout-minutes: 10
+
+    name: 'Build & Unit Test: node-${{ matrix.node_version }}, ${{ matrix.os }}'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        with:
+          # Required for docs/versions tests
+          fetch-depth: 0
+
+      - name: Install pnpm (node 14, pnpm 7)
+        if: matrix.node_version == 14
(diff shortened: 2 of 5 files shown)


ST-DDT:
Well, the CI looks like it doesnt run at all. Any idea why?

CoolPlayLin:
this is the feature of `pull_request_target` -- it needs to merge into target branch first to make it work normally
@ST-DDT

CoolPlayLin:
These old file should delete after merging @ST-DDT

ST-DDT:
From the [GitHub-Docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)

> **Warning:** ... Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. ...

According to that statement we should not use pull_request_target for our builds?! :thinking:

CoolPlayLin:
>untrusted code from the pull request with this event

Well, it will run workflow in target branch instead of contributor's workflow file. I think next branch do NOT have untrusted code if all contribution were reviewed carefully

@ST-DDT

CoolPlayLin:
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

ST-DDT:
> https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

> TL;DR: Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

ST-DDT (changes requested):
I like the idea of fixing the unable to comment issue that we have (because it external PRs only have read permissions), but granting write permissions this way doesnt sound like a good idea to me. (Even if the workflow file is from next, the PR itself is not and thus builds & executes (the tests using) untrusted code)

ST-DDT:
If you consider `pull_request_target` to be safer/better than `pull_request`, why do your own repositories use the later?

damienwebdev:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

I'm considering this malicious activity.

From the GitHub docs:

> Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork.

matthewmayer:
https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests

This seems a fairly good write up of the differences between pull_request and pull_request_target and what to be aware of when using the latter.

xDivisionByZerox:
> [...] it needs to merge into target branch first to make it work normally

Isn't that exactly the opposite of what you want from your CI?

> I'm considering this malicious activity.

I agree with @damienwebdev here. The PR description is phrased in a way to make it look like the changes made improve the security of our repository. In reality the opposite is happening and open Faker for attacks described in the article shared by @matthewmayer.

PR #2489, opened 2023-10-21
docs: how to use with Playwright

jfgreffier:
Playwright is a popular end-to-end testing solution. The tests are somewhat similar to Jest, but not exactly so here is an example of Faker integration into a Playwright test.


Shinigami92:
@debs-obrien Does this meet your requirements? Or should we improve something, such as a way of style like Playwright should best be used?

docs/guide/frameworks.md
```diff
@@ -98,3 +98,46 @@ describe('Testing the application', () => {
...
+    await page.getByLabel('email').fill(email);
+    await page.getByLabel('username').fill(username);
+    await page.getByLabel('password', { exact: true }).fill(password);
+    await page.getByLabel('confirm password').fill(password);
+    await page.getByRole('button', { name: 'Register' }).click();
+
+    // Now, we try to login with these credentials.
+    await page.goto('https://www.example.com/login');
+    await page.getByLabel('email').fill(email);
+    await page.getByLabel('username').fill(username);
```
xDivisionByZerox: While not related to Playwrite itself it's off-putting to me that you are required to supply the username as well as the email to sign in. Usually, only one of both is required.

docs/guide/frameworks.md
```diff
@@ -98,3 +98,46 @@ describe('Testing the application', () => {
...
+Integration with [Playwright](https://playwright.dev/) is also easy:
+
+```ts
+import { faker } from '@faker-js/faker/locale/en';
+import { test, expect } from '@playwright/test';
+
+// This will re-seed our faker instance after each test.
+test.afterEach(() => {
+  faker.seed();
+});
```
xDivisionByZerox: We don't do that in the cypress example. So I'm not sure if we should add it there as well or remove it here.
Additionally, while you write that this "re-seeds the faker instance", the seed is still random each time. So I don't see the advantage of doing it here.

xDivisionByZerox:
[There is a formatting error.](https://github.com/faker-js/faker/actions/runs/6596632739/job/17927857451?pr=2489#step:7:9) Please run the formatter locally via `pnpm run format`. 

---

Thank you for your contribution :heart:

docs/guide/frameworks.md
ST-DDT: The vitest one has a similar block, I think we should move them to the describe block to ensure this is run only after relevant tests, because seeding is "expensive". (Potentially in a separate PR).

docs/guide/frameworks.md
jfgreffier: Copy-paste error. My bad!

docs/guide/frameworks.md
jfgreffier: I removed the seed, we could add it in a dedicated seed example for clarity. In another PR that is

ST-DDT:
FFR: Please avoid force pushing as it makes incremental reviews harder.

docs/guide/frameworks.md
```diff
@@ -98,3 +98,39 @@ describe('Testing the application', () => {
...
+    await page.getByLabel('username').fill(username);
+    await page.getByLabel('password', { exact: true }).fill(password);
+    await page.getByLabel('confirm password').fill(password);
+    await page.getByRole('button', { name: 'Register' }).click();
+
+    // Now, we try to login with these credentials.
+    await page.goto('https://www.example.com/login');
+    await page.getByLabel('email').fill(email);
+    await page.getByLabel('password').fill(password);
+    await page.getByRole('button', { name: 'Register' }).click();
```
xDivisionByZerox: This name should propably be something like "Sign In" or "Login" as you already did register an account in the previous step.

```suggestion    
    await page.getByRole('button', { name: 'Login' }).click();
```

docs/guide/frameworks.md
```diff
@@ -98,3 +98,39 @@ describe('Testing the application', () => {
...
+
+    // We should have logged in successfully to the dashboard page.
+    await expect(page).toHaveURL(/.*dashboard/);
```
xDivisionByZerox, line 133: Nitpick: Since you navigate in the previous steps to an absolute URL I think it would be valid to match a full URL here as well.
```suggestion
    await expect(page).toHaveURL('https://www.example.com/dashboard');
```

xDivisionByZerox (changes requested):

docs/guide/frameworks.md
jfgreffier: Personnaly I would leave matching, similarly to the Cypress example using `cy.url().should('include', '/dashboard')`

xDivisionByZerox: Valid reasoning from your side. I'll accept your version.

xDivisionByZerox (approved):

ST-DDT (approved):

ST-DDT:

PR #2490, opened 2023-10-21
infra(unicorn): no-for-loop

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-for-loop`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-for-loop.md) lint rule.


xDivisionByZerox (dismissed):

test/docs/versions.spec.ts
```diff
@@ -38,8 +38,7 @@ describe.runIf(isFakerOrigin())('docs versions', () => {
         semver.valid(version)
       );
       const latestMajorRelease = semver.major(releaseVersions[0].version);
-      for (let i = 0; i < releaseVersions.length; i++) {
-        const { version, link } = releaseVersions[i];
+      for (const [i, { version, link }] of releaseVersions.entries()) {
```
matthewmayer: this seems like a perfectly sensible use of a for-loop.

matthewmayer (changes requested):

test/docs/versions.spec.ts
xDivisionByZerox: The issue with the conventional "for" loop lies in its use of an index accessor, which can become rather cumbersome to handle, especially in TypeScript's strict mode. In such cases, you must consistently check whether the accessed member is not `undefined`. To circumvent this, it's often preferable to utilize the `for...of` loop, as it guarantees that all members within the iterator chain are defined (unless the generic iterator type includes `undefined` itself). Furthermore, in the typical "for" loop, the index parameter, "i" is declared with "let" whereas in the `for...of` loop, it is a constant. This alteration prevents modifications to the iterator index within the loop, reducing the likelihood of a developer introducing a potential issue.

matthewmayer: counterpoint, every programmer knows what `for (let i = 0; i < releaseVersions.length; i++) {` does but ` for (const [i, { version, link }] of releaseVersions.entries()) {` requires a double-take.

xDivisionByZerox: I'd like to introduce you to some of my former co-workers if you start with the "every programmer knows" point without defining the exact domain for your statement (:

---

Counterpoint, every programmer who wants to contribute to an open-source project, that highly embraces the functionalities of modern JavaScript features, should know about modern array functions. Sorry, but I don't want to teach people basics like `Array.prototype.entries()`.

I would be open to discussing the destruction part in the loops definition section.

```ts
for (const entry of array.entries()) {
  const [index, value] = entry;
  // or (although I think we have a "prefer-destruction" rule)
  const index = entry[0];
  const value = entry[1];
  // entry is strongly typed as [number, T] tuple, so index accessing is fine with strict mode
}
```

ST-DDT: > counterpoint, every programmer knows what `for (let i = 0; i < releaseVersions.length; i++) {` does but ` for (const [i, { version, link }] of releaseVersions.entries()) {` requires a double-take.

Would renaming i -> index help with the clarity of the statement?

> I'd like to introduce you to some of my former co-workers if you start with the "every programmer knows" point without defining the exact domain for your statement (:

I think the argument is that if you know what a for-i loop is then you also know a for-of loop and vice versa (or at least can derive that).
That leaves the deconstruction. I know that is a little bit harder and I would be perfectly fine to split this into two lines:

````ts
for (const [index, value] of array.entries()) {
  const { version, link } = value;
````

Especially with the rename from i->index this should be fairly easy to read/understand.

We use deconstruction quite a lot in our code, so I think I would consider simple versions thereof a requirement to contribute to the project.

Finally, this case is in a test that most people won't ever look into.

---

There is no need to decide on this now, we can also ask for input from the other maintainers/members.

---

Some statistics (from next):

- for-of loops: 71
  - with deconstructor: 6
- for-i loops: 35
  - using single index access: 5
  - using multi index access: 3
- deconstructors: 120

Comment (matthewmayer) [reply test/docs/versions.spec.ts]: the deconstruction on a seperate line definitely makes it easier to read, would support that

Comment (ST-DDT) [reply test/docs/versions.spec.ts]: @matthewmayer Does that resolve your issues?

**Review by matthewmayer — APPROVED** [review]

File: src/modules/finance/iban.ts [hunk 2]
```diff
@@ -1398,8 +1398,8 @@ const iban: Iban = {
   ],
   mod97: (digitStr) => {
     let m = 0;
-    for (let i = 0; i < digitStr.length; i++) {
-      m = (m * 10 + +digitStr[i]) % 97;
+    for (const element of digitStr) {
+      m = (m * 10 + +element) % 97;
```
matthewmayer, line 1402: curious why the double + is needed here?

xDivisionByZerox (approved):

PR #2494, opened 2023-10-22
fix(finance): maskedNumber has incorrect defaults

ST-DDT:
Fix the incorrect defaults introduced in #2055

- #2055

### See also:

- [Merge-Commit](https://github.com/faker-js/faker/commit/2098b4d20e9604c95ab24f2f89452a11ae5f49d7#diff-ed919e84dad7aa63abddfc6222d0c69df214a632cd29cbce6866b9a0f968d618L346-L350)
- [PR #2055 diff](https://github.com/faker-js/faker/pull/2055/files#diff-ed919e84dad7aa63abddfc6222d0c69df214a632cd29cbce6866b9a0f968d618L346-L350)

---

### Alternative solution:

- Change the documented defaults

### Additional suggestion:

Deprecate the method entirely (because it is probably more useful if done like this by the user: 

````ts
someMethod().replace(/(?<=.{4})\w(?=.{2})/g,'*')
````

which converts `DE241358431563434532` to `DE24**************32`

We should probably document that somewhere in our docs too, although I assume they should already have something similar in their code somewhere.

### CODE REVIEW

**Review by xDivisionByZerox — DISMISSED** [review]

**Comment by matthewmayer** [comment]
Given that the defaults were accidentally changed during a method rename and we didn't catch that, suggests there should be a test for that? e.g. check that maskedNumber() with no params includes (... )

**Comment by ST-DDT** [comment]
> Given that the defaults were accidentally changed during a method rename and we didn't catch that, suggests there should be a test for that? e.g. check that maskedNumber() with no params includes (... )

Done

**Review by matthewmayer — DISMISSED** [review]

File: test/modules/finance.spec.ts [hunk 1]
```diff
@@ -217,6 +217,12 @@ describe('finance', () => {
       });
 
       describe('maskedNumber()', () => {
+        it('should return contain parenthesis, ellipsis and have a length by default', () => {
```
xDivisionByZerox: Nitpick:

```suggestion
        it('should return contain parenthesis, ellipsis and have a length of 4 by default', () => {
```

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2505, opened 2023-10-27
refactor(person)!: flatten jobs definitions

matthewmayer:
Currently the definitions for jobArea, jobDescriptor and jobType methods 
are nested inside a title.ts file. This causes undesired errors (ref #2503 ) when some 
of the job definitions are supplied but not others (e.g. in fr, ur, ar).

This removes the special cases, and just makes job_area, job_descriptor 
and job_title simple string arrays like other definitions.

The bulk locale changes were done with a simple script followed by linting.

This changes definition files but we generally don't consider those breaking changes when only accessed via faker.helpers.fake? Could be deferred until 9.0 if considered breaking.

```js
const { allLocales } = require('@faker-js/faker');
const fs = require('fs');
const keys = Object.keys(allLocales);
for (let key of keys) {
  if (!['base'].includes(key)) {
    const localFaker = allLocales[key];
    if (localFaker.person.title) {
      const title = localFaker.person.title;
      if (title.level) {
        write('job_area', title.level, key);
      }
      if (title.descriptor) {
        write('job_descriptor', title.descriptor, key);
      }
      if (title.job) {
        write('job_type', title.job, key);
      }
      try {
        fs.rmSync("src/locales/" + key + "/person/title.ts")
      } catch (e) {}
    }
  }
}

function write(method, definitions, locale) {
  console.log(method, locale, definitions.length);
  fs.writeFileSync(
    'src/locales/' + locale + '/person/' + method + '.ts',
    'export default ' + JSON.stringify(definitions, null, 2)
  );
}

```


src/index.ts
```diff
@@ -48,9 +48,6 @@ export type {
   /** @deprecated Use PersonDefinitions instead */
   PersonDefinition as NameDefinitions,
   PersonDefinition,
-  /** @deprecated Use PersonTitleDefinitions instead */
-  PersonTitleDefinition as NameTitleDefinitions,
-  PersonTitleDefinition,
```
xDivisionByZerox, line 55: We usually don't mark locale changes as deprecated. But this PR would change the current export API, which I would consider breaking.

src/locales/af_ZA/person/index.ts
```diff
@@ -5,13 +5,19 @@
...
 import last_name from './last_name';
 import last_name_pattern from './last_name_pattern';
 import male_first_name from './male_first_name';
 
 const person: PersonDefinition = {
   female_first_name,
   first_name,
+  job_area,
+  job_descriptor,
+  job_type,
```
xDivisionByZerox: If these descriptions have been added, shouldn't there be a `title` property that gets removed? I guess these ate the locale data from the en locale (currently on the phone, so verifying is hard for me).

xDivisionByZerox (dismissed):
Keep in mind that the prebuild faker instances contain more locale data than the actual locale. You should itterate over the local objects instead in your generation script. This should also reduce the diff significantly (due to current false positives).

src/index.ts
ST-DDT: I consider moving of the locale data a breaking change.
If we just do that in v9.0 there is no problem with removing the exports.

matthewmayer:
> Keep in mind that the prebuild faker instances contain more locale data than the actual locale. You should itterate over the local objects instead in your generation script. This should also reduce the diff significantly (due to current false positives).

My bad I thought rawDefinitions was for that. Will fix for the locales that were previously inheriting.

src/index.ts
matthewmayer: marked as breaking.

src/locales/af_ZA/person/index.ts
matthewmayer: resolved, now only affected locales are updated

matthewmayer:
I noticed that the `sk, cs_CZ, he, pl` data are just untranslated copies of `en`. Would it make sense to remove them as part of this PR?

ST-DDT:
> I noticed that the `sk, cs_CZ, he, pl` data are just untranslated copies of `en`. Would it make sense to remove them as part of this PR?

If the files are just plain copies, we can remove them now and the users wouldnt get any breakages. IMO We dont have to wait for v9 with that.

ST-DDT:
Please fix the "merge conflicts" (although I think it will most likely merge locally without issues).

matthewmayer:
if this won't be merged until v9 is it worth bothering to fix merge conflicts until we are ready to start on v9?

ST-DDT:
No, there is no need to fix them now. I just left a comment to go along with the needs rebase label for improved visibility.

matthewmayer:
OK :D 90% of my notifications for this project are about merge commits, so i wont bother adding to the volume for now!

ST-DDT:
> OK :D 90% of my notifications for this project are about merge commits, so i wont bother adding to the volume for now!

I wished there was a way to turn them off (or get them merged 😉)...

docs/guide/upgrading_v9/2505.md
```diff
@@ -0,0 +1,3 @@
+### Jobs definitions reorganized
+
+The locale definitions used by `faker.person.jobTitle()`, `faker.person.jobDescriptor()`, `faker.person.jobArea()` and `faker.person.jobType()` have been reorganized and are no longer nested under `definitions.person.title`. If you are using the public methods, no changes are required. You only need to change your code if you are accessing the raw definitions e.g. in `faker.helpers.fake()`.
```
xDivisionByZerox, line 3: Can we mention the replacement patterns here? 

| Before | After |
| - | - |
|`person.title.descriptor` | `person.job_descriptor` |
|`person.title.level` | `person.job_area` |
|`person.title.job` | `person.job_type` |

docs/guide/upgrading_v9/2505.md
```diff
@@ -1,3 +1,10 @@
 ### Jobs definitions reorganized
 
 The locale definitions used by `faker.person.jobTitle()`, `faker.person.jobDescriptor()`, `faker.person.jobArea()` and `faker.person.jobType()` have been reorganized and are no longer nested under `definitions.person.title`. If you are using the public methods, no changes are required. You only need to change your code if you are accessing the raw definitions e.g. in `faker.helpers.fake()`.
+
+| Before | After |
+| - | - |
+|`person.title.descriptor` | `person.job_descriptor` |
+|`person.title.level` | `person.job_area` |
+|`person.title.job` | `person.job_type` |
```
xDivisionByZerox: Soory, I did not think you take my suggestion one to one^^
Since this is a small table, prettifying would not hurt, would it?

```suggestion
| Before                    | After                   |
| ------------------------- | ----------------------- |
| `person.title.descriptor` | `person.job_descriptor` |
| `person.title.level`      | `person.job_area`       |
| `person.title.job`        | `person.job_type`       |
```

xDivisionByZerox: TBH, I thought prettier would do that anyway. Seems like that is not the case.

ST-DDT (dismissed):

ST-DDT:
Ready for review

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):

ST-DDT:
3 approving reviews. Enabling auto-merge.

ST-DDT (approved):

PR #2506, opened 2023-10-27
fix(internet): username method to return value that always includes…

### LINKED ISSUE #1000: In email functionality : Second name is not appending in email most of the time
### Describe the bug

Trying to create 100 emails with first name, second name, and provider. but generating email doesn't have a second name most of the time.

### Code:
const firstName = "groot";
const secondName = "at";
const random = Math.floor(Math.random() *100000);
secondName = secondName+random;
console.log("First name: ",firstName,"Second name : ",secondName)
const email = faker.internet.email(firstName,secondName,provider).replace(/[^a-zA-Z0-9@.]/g, "");
console.log(email);

### Response: 
[0-0] First name: groot Second name: at29913
[0-0] groot56@gmail.com
[0-0] First name: groot Second name: at30314
[0-0] groot.at30314@gmail.com
[0-0] First name: groot Second name: at13277
[0-0] groot.at1327778@gmail.com
[0-0] First name: groot Second name: at60615
[0-0] grootat6061518@gmail.com
[0-0] First name: groot Second name: at84925
[0-0] groot5@gmail.com
[0-0] First name: groot Second name: at67487
[0-0] grootat6748741@gmail.com
[0-0] First name: groot Second name: at1124
[0-0] groot.at112430@gmail.com
[0-0] First name: groot Second name: at15594
[0-0] groot94@gmail.com
[0-0] First name: groot Second name: at77943
[0-0] groot.at7794310@gmail.com
[0-0] First name: groot Second name: at66877
[0-0] groot10@gmail.com
[0-0] First name: groot Second name: at73118
[0-0] groot41@gmail.com
[0-0] First name: groot Second name: at43517
[0-0] grootat43517@gmail.com
[0-0] First name: groot Second name: at66282
[0-0] groot62@gmail.com
[0-0] First name: groot Second name: at4783
[0-0] groot.at4783@gmail.com
[0-0] First name: groot Second name: at80369
[0-0] groot.at803690@gmail.com
[0-0] First name: groot Second name: at98724
[0-0] grootat9872427@gmail.com
[0-0] First name: groot Second name: at42108
[0-0] groot62@gmail.com
[0-0] First name: groot Second name: at23822
[0-0] grootat23822@gmail.com
[0-0] First name: groot Second name: at75730
[0-0] grootat757300@gmail.com
[0-0] First name: groot Second name: at17076
[0-0] grootat17076@gmail.com
[0-0] First name: groot Second name: at27913
[0-0] groot25@gmail.com
[0-0] First name: groot Second name: at6139
[0-0] groot.at613994@gmail.com
[0-0] First name: groot Second name: at29583
[0-0] groot92@gmail.com
[0-0] First name: groot Second name: at34788
[0-0] groot7@gmail.com
[0-0] First name: groot Second name: at53021
[0-0] grootat53021@gmail.com
[0-0] First name: groot Second name: at91160
[0-0] groot.at9116062@gmail.com
[0-0] First name: groot Second name: at74768
[0-0] groot85@gmail.com
[0-0] First name: groot Second name: at41251
[0-0] grootat41251@gmail.com

### Additional Info

in above response most of the time second name is not appending please check on that issue

amillwood:
Updated the username method to always return a value that includes the last name when provided.

Resolves #1000


ST-DDT:
Sorry for the delay in review, we will look at it shortly.

amillwood:
No worries. No rush, just a minor contribution - hope it helps

src/modules/internet/index.ts
```diff
@@ -616,7 +616,12 @@ export class InternetModule {
     } = options;
 
     let result: string;
-    switch (this.faker.number.int(2)) {
+    switch (
+      this.faker.number.int({
+        min: options.lastName || legacyLastName ? 1 : 0,
+        max: 2,
+      })
```
matthewmayer: this seems a bit awkward. Can we change case 0 to be case 2 instead? Then we can just do something like this to make it easier to read:

```js
const hasLastName = options.lastName || legacyLastName
const strategy = this.faker.number.int(hasLastName ? 1 : 2)
switch (strategy) {
```

matthewmayer (changes requested):

src/modules/internet/index.ts
amillwood: That is cleaner, great suggestion.  Updated!  (Also updated the JSDoc for the methods)

src/modules/internet/index.ts
```diff
@@ -478,7 +490,7 @@ export class InternetModule {
 
   /**
    * Generates a username using the given person's name as base.
-   * The resulting username may use neither, one or both of the names provided.
+   * The resulting username will use the names provided or choose a random name if one is not provided.
```
matthewmayer: This is not completely true. If the names are non ascii and can't be transliterated to ascii then they are not used in the resulting username.

src/modules/internet/index.ts
```diff
@@ -616,19 +634,21 @@ export class InternetModule {
...
-    switch (this.faker.number.int(2)) {
+    const hasLastName = options.lastName || legacyLastName;
+    const strategy = this.faker.number.int(hasLastName ? 1 : 2);
+    switch (strategy) {
       case 0:
-        result = `${firstName}${this.faker.number.int(99)}`;
+        result = `${firstName}${this.faker.helpers.arrayElement([
+          '.',
+          '_',
+        ])}${lastName}${this.faker.number.int(99)}`;
```
matthewmayer: Maybe create a variable like `separator` above the switch which is randomly . or - (shall we add empty string as a third option?) that can be reused in the cases?

matthewmayer (changes requested):

ST-DDT:
One problem I see with the username method itself is that it produces very vage usernames that you might choose for private accounts.
But I think a lot of professional users of faker need usernames that companies would use and thus roughly follows the firstName.lastName strategy.

src/modules/internet/index.ts
amillwood: Oh, good point.  Should I revert to the previous description?  Or what do you suggest?

src/modules/internet/index.ts
amillwood: Updated.  I did not add an empty string to the array, because the `allowSpecialCharacters` logic uses the separator to insert the special characters.  I wasn't sure how you all would want to proceed with that (i.e. update the test cases, or update the special character logic to insert the special character in a different way)

matthewmayer:
Companies don't always use FirstName.lastname anyway though

At companies I have worked at there have been patterns like

john.smith@
john@
jsmith@
smithj@

ST-DDT:
> Companies don't always use FirstName.lastname anyway though

My point is that companies follow a more standardized approach to usernames than private persons do.

matthewmayer:
I think using both FirstName and lastname when provided makes sense, but we could have more ways to combine them for greater variety eg

More separators including empty string
Surname or firstname first
Sometimes lowercase the whole string

test/modules/internet.spec.ts
```diff
@@ -252,6 +252,7 @@ describe('internet', () => {
           const [prefix, suffix] = email.split('@');
 
           expect(prefix).includes('Aiden');
+          expect(prefix).includes('Harann');
```
ST-DDT, line 255: Please include a similar test for the username method.

ST-DDT:
**Team Decision**

- We accept this as a fix for the problem at hand, we will rethink the username method in the future.
  - (More patterns,username modes e.g.  company, private, gamer...)

src/modules/internet/index.ts
amillwood: Reverted to previous description

test/modules/internet.spec.ts
amillwood: updated

src/modules/internet/index.ts
```diff
@@ -249,7 +255,7 @@ export class InternetModule extends ModuleBase {
 
     const {
       firstName = this.faker.person.firstName(),
-      lastName = legacyLastName ?? this.faker.person.lastName(),
+      lastName = legacyLastName,
```
ST-DDT: If lastName isn't generated here, I would also omit the firstName generation here to stay consistent.
What do the others think?

src/modules/internet/index.ts
```diff
@@ -455,7 +467,7 @@ export class InternetModule extends ModuleBase {
...
```
ST-DDT: Same here

src/modules/internet/index.ts
```diff
@@ -611,19 +629,20 @@ export class InternetModule extends ModuleBase {
     } = options;
 
     let result: string;
-    switch (this.faker.number.int(2)) {
+    const hasLastName = options.lastName || legacyLastName;
```
ST-DDT: IMO this should be part of the deconstructor of options.

````ts
lastName: hasLastName = legacyLastName
````

File: test/modules/internet.spec.ts [hunk 8]
```diff
@@ -252,6 +252,7 @@ describe('internet', () => {
...
           expect(prefix).toMatch(
             /^Aiden((\d{1,2})|([._]Harann\d{1,2})|([._](Harann)))/
```
ST-DDT: In line 400 you improved the regex, maybe do that here as well.

test/modules/internet.spec.ts
```diff
@@ -396,9 +397,7 @@ describe('internet', () => {
           expect(username).toBeTruthy();
           expect(username).toBeTypeOf('string');
           expect(username).includes('Aiden');
```
ST-DDT, line 399: Please also check for the lastName specifically as it is easier to understand the check than when it is contained in the regex.

```suggestion
          expect(username).includes('Aiden');
          expect(username).includes('Harann');
```

test/modules/internet.spec.ts
```diff
@@ -365,7 +366,7 @@ describe('internet', () => {
 
           expect(suffix).toMatch(/^example\.(com|net|org)$/);
           expect(faker.definitions.internet.example_email).toContain(suffix);
-          expect(prefix).toMatch(/^Mike([.!#$%&'*+-/=?^_`{|}~]Smith)?\d*/);
+          expect(prefix).toMatch(/^Mike[.!#$%&'*+-/=?^_`{|}~]Smith\d*/);
```
ST-DDT, line 369: AFAICT this could also use the simplified regex similar to line 351 or 400

src/modules/internet/index.ts
amillwood: That makes sense.  Updated to add first name within username method, if undefined.

src/modules/internet/index.ts
amillwood: Updated

src/modules/internet/index.ts
amillwood: updated

test/modules/internet.spec.ts
amillwood: Good catch, missed this one.  Updated

test/modules/internet.spec.ts
amillwood: added

test/modules/internet.spec.ts
amillwood: This one is for the special character test.  I think it is as simple as we can make it.  I just removed the optional part of the regex.  Because last name is provided in this test, it should now contain the special character + last name.

ST-DDT (dismissed):

xDivisionByZerox (dismissed):

ST-DDT:
@amillwood Could you please fix the merge conflict?

ST-DDT (approved):

xDivisionByZerox (approved):

matthewmayer (approved):

ST-DDT:

PR #2507, opened 2023-10-27
infra(unicorn): no-negated-condition

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-negated-condition`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-negated-condition.md) lint rule.


scripts/apidoc/diff.ts
```diff
@@ -8,13 +8,13 @@ import { nameDocsDiffIndexFile, pathDocsDiffIndexFile } from './utils';
  */
 async function loadRemote(url: string): Promise<DocsApiDiffIndex> {
   return fetch(url).then((res) => {
-    if (!res.ok) {
-      throw new Error(
-        `Failed to load remote diff index from ${url}: ${res.statusText}`
-      );
-    } else {
+    if (res.ok) {
```
matthewmayer: i would keep the current order, but just remove the else clause. Generally throwing early with error conditions and then finishing with the happy case seems more logical.

```js
if (!res.ok) {
   throw new Error(
      `Failed to load remote diff index from ${url}: ${res.statusText}`
    );      
}

return res.json() as Promise<DocsApiDiffIndex>;
```

matthewmayer (changes requested):

scripts/apidoc/diff.ts
ST-DDT: Done

matthewmayer (approved):

xDivisionByZerox (approved):

PR #2508, opened 2023-10-27
infra(unicorn): no-useless-switch-case

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/no-useless-switch-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-switch-case.md) lint rule.


xDivisionByZerox (dismissed):

ST-DDT:
**Team Decision**

- We want to enable https://typescript-eslint.io/rules/switch-exhaustiveness-check/
- After that is merged, we no longer need the default and can remove it.
  - Consideration, if a non typescript user inputs a bad value, currently it would go into the default case. We could throw an error but we don't do that for other bad values either, so non typescript users don't get special support for enums either.
- @ST-DDT: Will create an issue for `switch-exhaustiveness-check`s to require default cases for non union type switch cases

EDIT:

- https://github.com/typescript-eslint/typescript-eslint/pull/7880

ST-DDT:
Blocked by #2519 

- #2519

matthewmayer (dismissed):
Support merging in v9 so we can document the breaking change for non-TS users.

matthewmayer:
Per #2521 suggest adding the following to docs/guide/upgrading_v9/2508.md:


### Some methods now return undefined in Javascript when unknown enumeration values are passed

Some methods would previously fallback to a default value for an option when an unknown value was passed for a enum parameter. Now, these methods will return undefined instead. This only affects usage in Javascript, as in Typescript this usage would already throw a compile-time error.

For example:
```js
faker.color.rgb({format: "unexpectedvalue"}) 
// in Faker v8, is [ 110, 82, 190 ] like {format: "decimal"}
// in Faker v9, is undefined
```

This affects:
* The `format` property of `faker.color.rgb()` must be one of `'binary' | 'css' | 'decimal' | 'hex'` if provided
* The `format` property of `faker.color.cmyk()`, `faker.color.hsl()`, `faker.color.hwb()`, `faker.color.lab()`, `faker.color.lch()` must be one of `'binary' | 'css' | 'decimal'` if provided
* The `variant` property of `faker.location.countryCode()` must be one of `alpha-2`, `alpha-3`, `numeric` if provided
* The `casing` property of `faker.string.alpha()` and `faker.string.alphanumeric()` must be one of `'upper' | 'lower' | 'mixed'` if provided

ST-DDT:
> Per #2521 suggest adding the following to docs/guide/upgrading_v9/2508.md:

Added. Thanks for the suggestion.

ST-DDT:
Ready for review.

src/modules/internet/index.ts
```diff
@@ -640,7 +640,7 @@ export class InternetModule extends ModuleBase {
     } = options;
 
     let result: string;
-    const strategy = this.faker.number.int(hasLastName ? 1 : 2);
+    const strategy = this.faker.number.int(hasLastName ? 1 : 2) as 0 | 1 | 2;
```
Shinigami92: Should we rewrite this section, to make the code more readable?
(Because I initially questioned myself, why is 0 a possible value, cause there is only 1 or 2. And my brain skipped the part that 1 or 2 is "the max" value, and so 0 is included as lower bound.)

```suggestion
    const strategy = this.faker.number.int({ max: hasLastName ? 1 : 2 }) as 0 | 1 | 2;
```

src/modules/internet/index.ts
```diff
@@ -827,7 +826,7 @@ export class InternetModule extends ModuleBase {
     } = options;
 
     let result: string;
-    switch (this.faker.number.int(2)) {
+    switch (this.faker.number.int(2) as 0 | 1 | 2) {
```
Shinigami92: Same here

```suggestion
    switch (this.faker.number.int({ max: 2 }) as 0 | 1 | 2) {
```

Shinigami92 (commented):
theoretically approved, but please let someone like e.g. @xDivisionByZerox evaluate my suggestion :slightly_smiling_face:

src/modules/internet/index.ts
xDivisionByZerox: I'd rather suggest to extract this logic into it's own function. That way we don't have to assert `as 0 | 1 | 2` multiple times. I'm generally more in favor to write explicit code. That would be inconsistent with the rest of the codebase, tho.

Shinigami92: 😮 re-reading these lines now it clicked:
We should rewrite it to something like this:

```ts
const strategies = ['name-of-strat-1', 'name-of-strat-2']
if (hasLastname) strategies.push('name-of-strat-3')
const strategy = this.faker.helpers.arrayElement(strategies)
```

That would make it much more clear

ST-DDT:
Switched to strategy pattern. Ready to review.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #2509, opened 2023-10-27
infra(unicorn): prefer-code-point

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/prefer-code-point`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-code-point.md) lint rule.


src/modules/internet/index.ts
```diff
@@ -645,13 +645,13 @@ export class InternetModule {
...
 
-        if (char.charCodeAt(0) < 0x80) {
+        if (char.codePointAt(0) < 0x80) {
           // Keep ASCII characters
           return char;
         }
 
         // Final fallback return the Unicode char code value for Chinese, Japanese, Korean etc, base-36 encoded
-        return char.charCodeAt(0).toString(36);
+        return char.codePointAt(0).toString(36);
```
xDivisionByZerox: This should have been failing before, according to the desciption of the rule. Should we add a test verifying the correct behaviour for unicode names?

Example: `faker.internet.username({ firstName: '🦆' })`

xDivisionByZerox (dismissed):

src/modules/internet/index.ts
ST-DDT: Added the test.

test/modules/internet.spec.ts
```diff
@@ -421,6 +421,14 @@ describe('internet', () => {
           const username = faker.internet.userName('大羽', '陳');
           expect(username).includes('hlzp8d');
         });
+
+        it('should provide a fallback special unicode characters', () => {
+          const username = faker.internet.userName({
+            firstName: '🐼',
+            lastName: '❤️',
+          });
+          expect(username).includes('2qt8');
```
xDivisionByZerox, line 431: Sorry to bother you again.
While this test is good, I don't think we will know what '2qt8' is supposed to mean in 2 weeks from now.
Can we give that value a descriptive name and/or even generate it from the hex representation of the emoji? The hex value is easier to look up and verify IMO.

 ```suggestion
          const pandaHexValue = 0x1F43C;
          // we represent unicode in base 36
          const expectPandaString = pandaHexValue.toString(36);
          expect(username).includes('2qt8');
```

ST-DDT: We don't do that for the chinese version either and IMO it doesn't matter either as it is some weird unicode -> ascii mapping anyway.

xDivisionByZerox (approved):

PR #2511, opened 2023-10-27
infra(unicorn): prefer-top-level-await

ST-DDT:
Ref: #2439

- #2439

---

Temporarily disables the [`unicorn/prefer-top-level-await`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-top-level-await.md) lint rule.


xDivisionByZerox:
Can we not actually enable this rule and put the PR "on hold" until we switched? Don't see the advantage of providing the comment in the sourcecode right now, as it is more likely to be forgotten.

ST-DDT:
I would like to finish the unicorn list now and fix the "other/later" list of things that need to be in v9 in v9.
Please note that the switch to ESM main may or may not be in v9 at the earliest and I don't want to carry open PRs for extended periods of time.

xDivisionByZerox (approved):
> I would like to finish the unicorn list now and fix the "other/later" list of things that need to be in v9 in v9.
> Please note that the switch to ESM main may or may not be in v9 at the earliest and I don't want to carry open PRs for extended periods of time.

I see.

ejcheng (approved):

PR #2512, opened 2023-10-27
infra(unicorn): prefer-negative-index

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/prefer-negative-index`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-negative-index.md) lint rule.


xDivisionByZerox:
Personally, I find negative indexes quite unintuitive. Usually, a negative index is used to purposefully present an out of bound index. Look at `Array.prototype.indexOf`: an index of `-1` is returned in case the element was not found. Following the negative index logic, the element is actually positioned at `array.length - 1`, which is not the case.

The rule itself sadly doesn't have any description on WHY you should use it. I'm welcome for anybody providing reasoning on this.

ST-DDT:
**Team Decision**

- We accept this.
- Although we might want a code rewrite for the method that is currently affected.
  - e.g. change splice remove + index access -> index access adjusted by removed elements.

xDivisionByZerox (approved):

ejcheng (approved):

PR #2513, opened 2023-10-28
docs(algolia): use one search-index per branch

### LINKED ISSUE #2267: Searching unique on website suggests page that does not exist
When searching for "unique" on https://fakerjs.dev/, a suggestion is provided that seems to suggest there's a documentation page called "Unique". When I click on this suggestion, it takes me to the 404 page.

https://github.com/faker-js/faker/assets/48000208/02e70cb3-55f1-49df-b981-5e08dd61492d

ST-DDT:
Tries to fix: https://github.com/faker-js/faker/issues/2267

- https://github.com/faker-js/faker/issues/2267

---

I couldn't find a way to actually test the changes.


Shinigami92:
This looks straight forward to me 🤔 
Should we test in _production_?

Shinigami92 (dismissed):

ST-DDT:
I kind of already did, but the index doesn't get populated, so we would loose the current search capabilities.

Shinigami92:
> I kind of already did, but the index doesn't get populated, so we would loose the current search capabilities.

That would not be a problem for me
The index is new, so the data is also reset
It will be build up again

ST-DDT:
> It will be build up again

No, it won't. At least not for the last 5 months...

xDivisionByZerox:
> > It will be build up again
> 
> No, it won't. At least not for the last 5 months...

We would also need to backport this PR to v6 to prevent potential conflicts with the newly created indexes. There souldn't be any, but better be safe I guess.

Shinigami92:
> > It will be build up again
> 
> No, it won't. At least not for the last 5 months...

we could talk about what you tested and how maybe later in a voice-call or team-meeting
I'm a bit confused, or at least don't understand what an index is then

ST-DDT:
grafik

> I'm a bit confused, or at least don't understand what an index is then

There is nothing to be confused about.
(It normally gets updated during CI build)
The index not getting updated, is a bug or misconfiguration.
Maybe merging #2301 might fix it or it might be entirely unrelated to that.

Shinigami92:
do we maybe need to **create** an index there? with that button?
and then it gets filled and connected with the one in the code?

the one fakerjs index that is there already has 6.13K records, so as far as I understand the index dynamically builds up, based on what people are searching on the docs

ST-DDT:
Yes, you might have to create a new index, but even then it stays empty after the website build/time.
I can show it to you tomorrow, if you want.

ST-DDT:
The crawler is hosted and run on https://crawler.algolia.com/.
There appears to be some kind of failsafe, that prevents major changes to the index without human intervention.
For some reason this triggered 5 months ago and thus the index hasn't been updated since. (Fixed now)

We checked the configuration and noticed that we have to add the other subdomains to the whitelist.
This is currently in progress, once we did that, we can merge this and have individual indexes.
But we have to check periodically whether they still work. 😿 

Thanks @Shinigami92 for the help debugging this.

ST-DDT:
FFR: The index-names have to start with fakerjs 🤷‍♂️ 

grafik

docs/.vitepress/versions.ts
```diff
@@ -74,3 +74,7 @@ export const oldVersions = [
     link: `https://v${semver.major(version)}.fakerjs.dev/`,
   })),
 ].filter(({ link }) => link !== hiddenLink);
+
+export const algoliaIndex = isReleaseBranch
+  ? `faker-v${semver.major(version)}`
+  : 'faker-next';
```
ST-DDT:
```suggestion
  ? `fakerjs-v${semver.major(version)}`
  : 'fakerjs-next';
```

Shinigami92 (approved):

xDivisionByZerox (approved):

ST-DDT:
This now works as expected:

- Open https://docs-algolia-index-per-branch--vigilant-wescoff-04e480.netlify.app/
- Search "For generating a complete company name."
- Find exact results
- Open https://next.fakerjs.dev/
- Search "For generating a complete company name."
- Don't find exact results

Ref:

- Added after the last v8 release https://github.com/faker-js/faker/blob/a87602b3c45b1f7903460adfd6268c65c8669682/src/modules/company/index.ts#L22

matthewmayer (approved):

PR #2517, opened 2023-10-30
docs(phone): improve deprecation and see message

### LINKED ISSUE #2413: Improve deprecation message for faker.phone.number(format)
> It would be nice if at least a few options were available before the deprecation took effect. We have an simple use case like `faker.phone.number('##########')`, which I think would be hard to replicate until this is complete?

the direct replacement for `faker.phone.number('##########')` is `faker.helpers.replaceSymbolWithNumber('##########')` - maybe the deprecation message should be modified to make that clearer?

_Originally posted by @matthewmayer in https://github.com/faker-js/faker/issues/1542#issuecomment-1730810322_

ST-DDT:
Fixes https://github.com/faker-js/faker/issues/2413

- https://github.com/faker-js/faker/issues/2413

### DIFF
--- a/src/modules/phone/index.ts
+++ b/src/modules/phone/index.ts
@@ -62,7 +62,7 @@ export class PhoneModule {
       deprecated({
         deprecated: 'faker.phone.number(format)',
         proposed:
-          'faker.phone.number() or faker.string.numeric() or faker.helpers.fromRegExp()',
+          'faker.phone.number(), faker.string.numeric() or faker.helpers.fromRegExp()',
         since: '8.1',
         until: '9.0',
       });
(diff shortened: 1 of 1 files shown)


src/modules/phone/index.ts
```diff
@@ -52,7 +61,8 @@ export class PhoneModule {
     if (format != null) {
       deprecated({
         deprecated: 'faker.phone.number(format)',
-        proposed: 'faker.phone.number()',
+        proposed:
+          'faker.phone.number() or faker.string.numeric() or faker.helpers.fromRegExp()',
```
xDivisionByZerox:
```suggestion
          'faker.phone.number(), faker.string.numeric() or faker.helpers.fromRegExp()',
```

xDivisionByZerox (approved):

ejcheng (approved):

PR #2520, opened 2023-10-31
fix(locale): improve Swedish phone numbers format

Gyran:
https://en.wikipedia.org/wiki/Telephone_numbers_in_Sweden

Numbers starting with 070, 072, 073, 076 or 079 are reserved for mobile numbers in Sweden.


matthewmayer:
Note that faker.phone.number() does not specifically promise to return mobile numbers. Can also be landlines. If you have a use-case for mobile numbers only it might be good to get your input on https://github.com/faker-js/faker/issues/1542

Gyran:
That's true, I could add formats for landline numbers as well. But the current formats does not look like any Swedish numbers. So I thought that this is at least make the SV faker version return Swedish phone number (but not all kinds).
For my use case I only want mobile numbers and I think that is true for most Swedish use cases. But it would be nice as you say to be able to specify that.

Will write in the other issue as well, but in the meantime, is this change something you would be interested in or how can I change it to follow the guidelines? Add these formats to the cell formats instead?

matthewmayer:
If it's an improvement from the current data we should accept this PR anyway. 

At the moment the cell_phone definitions are not actively used by Faker. 

You can have a read through the discussion on the other issue and add any comments on how you would typically use this data to help guide us on what would be most useful in future.

matthewmayer (dismissed):

ST-DDT:
> I could add formats for landline numbers as well.

I think it would be best to include them for now so that we have them already when we split/improve the method

ST-DDT:
And please also include ones with the international prefix for variation.

src/locales/sv/phone_number/formats.ts
```diff
@@ -1 +1,25 @@
...
+  '023#####',
+  '+4623#####',
+  '08#######',
+  '+468#######',
+  '011#######',
+  '+4611#######',
+  '026######',
+  '+4626######',
+  '031######',
+  '+4631######',
```
ST-DDT: If I read the [wiki](https://en.wikipedia.org/wiki/Telephone_numbers_in_Sweden) article correctly, the area code can be almost anything. Are there some hidden rules that effectively limit the landline phone numbers to these 5?

Gyran: Non, I just picked 5 randomly because I didn't want to manually create one for each area code.
I can probably do a loop here instead and add it for all valid area codes

ST-DDT: Would it be sufficient to use

````
  '01#######',
  '+461#######',
  '02#######',
  '+462#######',
...
````

Because the third digit can be practically anything (With some hopefully irrelevant exceptions).

Gyran: Added in that way, and I guess most of them are valid. But some might be reserved for special purposes. For example 020 is a free of charge phone number and is not available as for a person.
But I think for the purpose of this library it's ok them they are all "structurally" correct and probably valid.

ST-DDT (approved):

matthewmayer (approved):

PR #2523, opened 2023-11-02
refactor(person): remove duplicate title definitions

matthewmayer:
Extracted from #2505

https://github.com/faker-js/faker/pull/2505#issuecomment-1783190723

If the job_title definitions etc are just exact copies of en, remove them and allow the standard fallback to en.


ST-DDT (approved):

src/locales/cs_CZ/person/title.ts
```diff
@@ -1,91 +0,0 @@
...
-    'Interactions',
-    'Factors',
-    'Usability',
-    'Metrics',
-  ],
-  job: [
-    'Supervisor',
-    'Associate',
-    'Executive',
-    'Liason',
```
xDivisionByZerox, line 68: This is actually `'Liaison'` (with an additional i) in the en locale.

---

Doesn't really matter. But leaving this here just in case someone shows up in the future.

xDivisionByZerox (approved):

PR #2526, opened 2023-11-04
refactor(locale): remove fr_CH data which is identical to fr

matthewmayer:
Removed identical animal, color, commerce, date, finance, music, vehicle and word definitions which can use fallback from fr instead

Debatable if this should be considered breaking. Would only affect behavior if you were using a custom Faker instance with the fr_CH data but no fr fallback


ST-DDT:
This PR can easily be reviewed with the following script executed on next:

````sh
diff -rs src/locales/fr_CH src/locales/fr | egrep '^Files .+ and .+ are identical$' | awk -F '(Files | and | are identical)' '{print $2}'
````

Although according to the script the following files are identical as well:

- src/locales/fr_CH/location/building_number.ts
- src/locales/fr_CH/location/city_pattern.ts
- src/locales/fr_CH/location/country.ts
- src/locales/fr_CH/location/direction.ts
- src/locales/fr_CH/location/secondary_address.ts
- src/locales/fr_CH/location/street_address.ts
- src/locales/fr_CH/location/street_pattern.ts
- src/locales/fr_CH/lorem/index.ts
- src/locales/fr_CH/lorem/words.ts
- src/locales/fr_CH/person/female_prefix.ts
- src/locales/fr_CH/person/first_name.ts
- src/locales/fr_CH/person/index.ts
- src/locales/fr_CH/person/last_name_pattern.ts
- src/locales/fr_CH/person/male_prefix.ts
- src/locales/fr_CH/person/name.ts
- src/locales/fr_CH/person/prefix.ts
- src/locales/fr_CH/person/sex.ts
- src/locales/fr_CH/person/title.ts
- src/locales/fr_CH/phone_number/index.ts

to remove them append ` | xargs rm`

matthewmayer:
I decided to only touch modules where the entire module was identical.

xDivisionByZerox:
A little reminder that the following contribution guideline should be respected by members and maintainers as well:

> ## Adding new locale or updating existing one
>
> [...]
> Please only change files related to one module (e.g. person, location) whenever possible. This can simplify/speed up the review process. Additionally, it allows the maintainers to track PRs in a meaningful way by adding related labels.

matthewmayer:
I can split this if you want, but as this is only removals not additions i feel there's a lot less to review (sourcing, translations, etc)

xDivisionByZerox:
I totally understand that reasoning. Nevertheless, just imagine someone outside the core team providing a PR in that way because they "feel" like it. We would point to our contribution rules and asked them to split the PR. If not we would not need those rules in the first place.

---

All that being said, I think we can let that slip this time. You don't need split this PR - at least from my point of view. 
As I said aboth: Just "A little reminder" 😄

xDivisionByZerox:
This PR reduces the bundle size of the package by the following amounts:

|                      | Before (358572d9e76f4cd22bfcb09c092a1eaf3a31f005) | Current (8bfc804ff0527f3ecdbd0368c55c83bd8596456b) | Reduction Peercent |
| -------------------- | ------------------------------------------------- | -------------------------------------------------- | ------------------ |
| <b>Package Size</b>  | 2.4 MB                                            | 2.4 MB                                             | N/A                |
| <b>Unpacked Size</b> | 10.3 MB                                           | 10.1 MB                                            | 1,94%              |
 ---

I tryed this out of curiosity, but might that be interesting information for us on PRs that have the `c: locale` label? (sorry for the off topic question)

ST-DDT:
> A little reminder that the following contribution guideline should be respected by members and maintainers as well:
> 
> > ## Adding new locale or updating existing one
> > [...]
> > Please only change files related to one module (e.g. person, location) whenever possible. This can simplify/speed up the review process. Additionally, it allows the maintainers to track PRs in a meaningful way by adding related labels.

IMO that rule only applies to additions or changes that pertains the functional values within.
Removals or auto-transformed stuff doesn't have to be split into different PRs.
The rule is there to help us review new/external things. I assume you put this comment here due to https://github.com/faker-js/faker/pull/2265#issuecomment-1762898401
If you consider the split for those scripts too much (unnecessary) work, we can discuss that again.

xDivisionByZerox (approved):

ST-DDT:
> I decided to only touch modules where the entire module was identical.

What about the other identical files. Do you do them as a separate PR?

matthewmayer:
> > I decided to only touch modules where the entire module was identical.
> 
> What about the other identical files. Do you do them as a separate PR?

My feeling is they can be left for now. Realistically all fr_XX locales will use the same words for animals, dates, etc but will more likely use different patterns for names, locations etc. So even if two files are identical in fr_XX/location/YYY.ts currently, that's probably temporary - it probably makes it easier for future contributors if there's already a file they can directly edit rather than have to figure out the overrides.

ST-DDT (approved):

PR #2531, opened 2023-11-08
feat(person): add job titles for fr

matthewmayer:
This uses the new job_title_pattern to add job titles with a logical order for French e.g.

```
Coordinateur en chef de la tactique
Executif central des facteurs
Stagiaire corporate des marchés
```

Due to the need to translate “of” in French to de, de la, des, du, etc depending on the number and gender of the noun, this is incorporated into the level definitions.

The fr_BE and fe_CH definitions are deleted so the standard fr ones can fall through.


matthewmayer:
@makhtar-sarr if you're available to help review this (i'm not a native French speaker) that would be appreciated!

src/locales/fr/person/title.ts
```diff
@@ -19,4 +19,62 @@ export default {
     'Agent',
     'Stagiaire',
   ],
+  descriptor: [
+    'avant',
```
makhtar-sarr: Hi, according to `job_title_pattern.ts` we can say in French "Superviseur avant de configuration", whereas in good French we should say "Superviseur de pré-configuration". So the word "avant" is replaced by the prefix "pré". If possible, I suggest you remove the word "avant".

src/locales/fr/person/title.ts
```diff
@@ -19,4 +19,62 @@ export default {
...
+    'central',
+    'client',
```
makhtar-sarr, line 24: In French, the descriptor can be placed at the end, e.g. "Gestionnaire de la configuration client".

src/locales/fr/person/title.ts
```diff
@@ -19,4 +19,62 @@ export default {
     'Agent',
     'Stagiaire',
   ],
+  descriptor: [
```
makhtar-sarr, line 22: To sum up, most of the words in the description should be qualifying adjectives to go with the pattern set up.

matthewmayer: So would it make sense to switch the pattern to `{{person.jobType}} {{person.jobArea}} {{person.jobDescriptor}}` in general instead?

makhtar-sarr: Yes, but if you want to keep this pattern `{{person.jobType}} {{person.jobDescriptor}} {{person.jobArea}}` then it would be better to have only adjectives in the descriptor.

matthewmayer: I tried applying some suggestions, reducing the descriptors to only adjectives and putting them last. Let me know if this looks OK. (I appreciate this is not going to be perfect French, but I think that may be impossible without Faker knowing more about French grammar).

```
'Architecte des interactions humain',
  'Consultant de la tactique national',
  "Executif de l'utilisabilité mondial",
  'Developpeur de la responsabilité futur',
  "Developpeur de l'optimisation principal",
  'Specialiste du marketing central',
  "Developpeur de l'utilisabilité futur",
  'Technicien des marchés direct',
  "Consultant de l'assurance central",
  'Architecte de groupe interne',
  'Consultant de division principal',
  'Specialiste des facteurs client',
  'Ingenieur de la création central',
  "Agent de l'optimisation principal",
  "Consultant de l'identité direct",
  'Coordinateur de la marque principal',
  "Directeur de l'optimisation national",
  "Coordinateur de l'identité principal",
  'Designer des comptes futur',
  'Producteur de la qualité futur',
  'Administrateur des solutions régional',
  "Analyste de l'assurance international",
  'Administrateur de la création international',
  "Specialiste de l'identité régional",
  'Analyste des comptes direct',
  'Developpeur de la marque humain',
```

makhtar-sarr: Ok, on the whole it's good, but if you could also take into account the gender of words, to avoid certain spelling mistakes that would be better, e.g. `'Coordinateur de la marque principale'` instead of `'Coordinateur de la marque principal'`. In French, gender counts for a lot.

matthewmayer: Agreed but we would have to split the lists by gender of the nouns. Which is possible but I think probably best to wait until after https://github.com/faker-js/faker/pull/2505 is merged (v9) when these definitions will be restructured as I don't want to complicate that.

ST-DDT (approved):

ST-DDT:
@matthewmayer Is this ready for merge?

matthewmayer:
Yes

PR #2530, opened 2023-11-07
docs: restructure apidocs page tree

XinYuanHaoFanPinChen:
Fixes: #2527

- #2527

PR #2532, opened 2023-11-08
refactor: improve groupBy

ST-DDT:
Adds a value mapping function to `groupBy`.

Intended for #2530


xDivisionByZerox (commented):
I feel like this missed the point of "grouping" by a key. This function would then break the single responsibility principle.

Note that the following polifills (which this function was inspired by) do not include a value mapper as well:
- [Object.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy)
- [Map.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy)
- [Array.prototype.groupToMap](https://tc39.es/proposal-array-grouping/#sec-map.groupby)

ST-DDT:
In the Java world they noticed quite early ([Java 8](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#groupingBy-java.util.function.Function-java.util.function.Supplier-java.util.stream.Collector-)), that there is a need to give the user the ability to also process/transform the value.
Common usecases include wrapper shedding, which is the reason why I added the parameter in the first place.

````ts
groupBy(
  [
    [1, "someValue"],
    [2, "blub"],
    [1, "anotherone"]
  ], 
  (e) => e[0],
  (e) => e[1]
);
````

=> 

````ts
{
  1: ["someValue", "anotherone"],
  2: ["blub"]
}
````

I can explain this in more detail in the next team-meeting if you would like to.

xDivisionByZerox:
Please add the behaviour to the indended place as #2530 is already merged. I'm fine with keeping the additional utility.

xDivisionByZerox (dismissed):

xDivisionByZerox:
"Off-topic" question: Should we have tests for our internal function?

xDivisionByZerox (approved):

PR #2533, opened 2023-11-08
feat(locale): and location to fr_SN

makhtar-sarr:
In this PR I add the most popular cities in Senegal as well as the zip code format.

https://en.wikipedia.org/wiki/List_of_cities_in_Senegal


ST-DDT (dismissed):

matthewmayer (commented):
How about location.state() does it make sense to use the regions for that?

https://en.wikipedia.org/wiki/Subdivisions_of_Senegal?wprov=sfti1#

makhtar-sarr:
Yes, that's a good idea, but should I remove them from the list of cities or not?

matthewmayer:
In other locales where state names happen to duplicate city names, we normally leave them in.

ST-DDT (approved):

matthewmayer (approved):
Looks good to me. Location module now seems well covered, could consider adding another PR with improved street names (otherwise will inherit from fr which might not be ideal).

ST-DDT:
Thanks for your contribution ❤️ .

PR #2535, opened 2023-11-08
fix(date): ensures correct range for birthdate

### LINKED ISSUE #2534: Birthdate with default values return age less than 18
### Pre-Checks

### Describe the bug

I was running into flaky tests in my application, and narrowed it down to faker's `date.birthdate` method.

According to the docs:

>The options to use to generate the birthdate. If no options are set, an age between 18 and 80 (inclusive) is generated.

But there are times that an age less than 18 years old is generated. This does not happen every time, but I was able to get a specific seed and refDate to reproduce the issue.

### Minimal reproduction code

https://stackblitz.com/edit/vitest-dev-vitest-md9oj6?file=test%2Fsuite.test.ts

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 13.0
    CPU: (8) arm64 Apple M2
    Memory: 65.42 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.20.2 - ~/.nvm/versions/node/v16.20.2/bin/node
    npm: 8.19.4 - ~/.nvm/versions/node/v16.20.2/bin/npm
    pnpm: 8.6.1 - ~/Library/pnpm/pnpm
    bun: 1.0.0 - ~/.bun/bin/bun
  Browsers:
    Chrome: 119.0.6045.123
    Edge: 119.0.2151.46
  npmPackages:
    @faker-js/faker: ^8.2.0 => 8.2.0
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

mumumilk:
fixes #2534


src/modules/date/index.ts
```diff
@@ -886,10 +888,13 @@ export class SimpleDateModule extends SimpleModuleBase {
     } else {
       // Avoid generating dates the first and last date of the year
       // to avoid running into other years depending on the timezone.
-      min = new Date(Date.UTC(0, 0, 2)).setUTCFullYear(
+      const minDay = refMonth === 0 && refDay === 1 ? 2 : refDay;
+      const maxDay = refMonth === 11 && refDay === 31 ? 30 : refDay;
+
+      min = new Date(Date.UTC(0, refMonth, minDay)).setUTCFullYear(
```
ST-DDT: These interfere with the usage when specifying min and max.
Because then it can only return the current date.

ST-DDT (changes requested):

ST-DDT:
Please add these tests:

https://github.com/faker-js/faker/blob/022fa540106e25818936c6ebfe8d10fe9ee15b30/test/modules/date.spec.ts#L576-L603

src/modules/date/index.ts
mumumilk: good catch, but as I was working on a fix, it occurred to me that the behavior is in fact correct the way it is currently.  

Maybe we should only update the docs to mention that `mode: age` will generate more accurate values (because it already uses the `refDate`) or change the default mode to age

ST-DDT: One way to ensure the age is 18+ is by subtracting 19 instead of 18 for the year mode.

ST-DDT: > change the default mode to age

That would be a breaking change. (We could do that, but not in a simple fix for v8.x)

mumumilk: >One way to ensure the age is 18+ is by subtracting 19 instead of 18 for the year mode.

much simpler than the way I was heading to, thanks

ST-DDT (approved):

xDivisionByZerox (approved):

ejcheng (approved):

PR #2536, opened 2023-11-10
feat(locale): Add Street to location fr_SN

makhtar-sarr:
This PR adds the most famous streets in the Dakar and Thiès region. As it's an exhaustive list, I've only included a few, although others may be added in future PRs.

https://senegal-streets.openalfa.com/dakar
https://senegal-streets.openalfa.com/thies


ST-DDT:
Blocked by #2533 

- #2533

ST-DDT:
This is no longer blocked, but there appears to be a merge conflict.

makhtar-sarr:
Sorry, I think I've solved it.

matthewmayer (approved):

src/locales/fr_SN/location/secondary_address.ts
```diff
@@ -0,0 +1 @@
+export default ['Apt. ###', '# étage'];
```
ST-DDT, line 1: This is the same as `fr`. Should we still add it?

src/locales/fr_SN/location/street_address.ts
ST-DDT, line 1: This is the same as fr. Should we still add it?

src/locales/fr_SN/location/street_pattern.ts
ST-DDT, line 1: Same here

src/locales/fr_SN/location/secondary_address.ts
matthewmayer: I think in the case where we think it is likely that in the future definitions could be different it's fine to leave duplicate files

For example, I'm sure the way that apartments are numbered in Senegal, Belgium, France and Canada are slightly different from each other. So even if they are identical for now, we can have separate definitions to make it easy for future localizers to update. 

If it's something that clearly will not be different like colours or north south east west for all French-speaking countries, we can rely on the inherited definitions

ST-DDT (approved):

src/locales/fr_SN/location/secondary_address.ts
makhtar-sarr: Yes, I agree with @matthewmayer, since the numbering of apartments can vary from one country to another, I think we can leave it as it is.

PR #2542, opened 2023-11-13
Create nif.js

vinioliveiras:

### DIFF
--- a/src/modules/finance/nif.js
+++ b/src/modules/finance/nif.js
@@ -1,0 +1,36 @@
+//credit : https://github.com/asmarques/pt-id  
+
+function pad(value, length) {
+  var result = value + '';
+
+  while (result.length < length) {
+    result = '0' + result;
+  }
+
+  return result;
+}
+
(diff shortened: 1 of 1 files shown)


src/modules/finance/nif.js
```diff
@@ -0,0 +1,36 @@
...
+
+  for (var i = 0; i < len; i++) {
+    sum += value[i] * (len + 1 - i);
+  }
+
+  var mod = sum % 11;
+  return '' + ((mod === 0 || mod === 1) ? 0 : 11 - mod);
+}
+function generate(prefix) {
+  var value = pad(Math.floor(Math.random() * 99999999), 8);
```
ST-DDT, line 25: This does not produce reproducible values.

ST-DDT (changes requested):
This looks totally different implementation wise then any of our other code.

Also this needs more explanation what it is actually for.

ST-DDT:
Duplicate of #2543

PR #2543, opened 2023-11-13
feat: add a script to generate Portuguese NIFs

vinioliveiras:
Adds support for generating random NIFs, NIF is a Fiscal Number in Portugal

Sometimes you need to generate random NIFs numbers to test some features

I guess this could be added as an option to the faker project, thanks!

### DIFF
--- a/src/modules/finance/nif.ts
+++ b/src/modules/finance/nif.ts
@@ -1,6 +1,6 @@
 //credit : https://github.com/asmarques/pt-id  
 
-function pad(value, length) {
+function pad(value: string|number, length: number) {
     var result = value + '';
   
     while (result.length < length) {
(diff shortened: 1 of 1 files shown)


src/modules/finance/nif.ts
```diff
@@ -0,0 +1,37 @@
...
+  
+    for (var i = 0; i < len; i++) {
+      sum += value[i] * (len + 1 - i);
+    }
+  
+    var mod = sum % 11;
+    return '' + ((mod === 0 || mod === 1) ? 0 : 11 - mod);
+  }
+  function nif(prefix) {
+    var value = pad(Math.floor(Math.random() * 99999999), 8);
```
ST-DDT, line 25: This does not produce reproducible values.

ST-DDT (changes requested):
This looks totally different implementation wise then any of our other code.

Also this needs more explanation what it is actually for.

PR #2545, opened 2023-11-13
feat(locale): Add phone number to fr_SN

makhtar-sarr:
In this PR, I've added the different phone number formats that exist in Senegal. I've also avoided using certain formats that I don't yet know in Senegal.

https://en.wikipedia.org/wiki/Telephone_numbers_in_Senegal


src/locales/fr_SN/phone_number/formats.ts
ST-DDT, line 1: Could you please sort these or explain why you sorted them like this?

makhtar-sarr: Yes, I've sorted it that way based on the number of users. e.g. `'77 ### ## ##'` is the most used format and so on.

ST-DDT: Sorry, this PR was overlooked.

Currently our data don't convey any meaning by their ordering, so please sort them alphabetically.

Also I couldn't find all of those prefixes in here https://en.wikipedia.org/wiki/Telephone_numbers_in_Senegal
Could you please provide an alternative source for this list?
https://en.wikipedia.org/wiki/Telephone_numbers_in_Senegal

Note: We will soon begin development for v9 and would like to resolve all open PRs for that.

ST-DDT (changes requested):
Currently our data don't convey any meaning by their ordering, so please sort them alphabetically.

Also I couldn't find all of those prefixes in here [en.wikipedia.org/wiki/Telephone_numbers_in_Senegal](https://en.wikipedia.org/wiki/Telephone_numbers_in_Senegal)
Could you please provide an alternative source for this list?
[en.wikipedia.org/wiki/Telephone_numbers_in_Senegal](https://en.wikipedia.org/wiki/Telephone_numbers_in_Senegal)

ST-DDT (changes requested):
Please update this PR to match our new phone number format options:

- https://next.fakerjs.dev/api/phone.html#number
- https://github.com/faker-js/faker/pull/2578

ST-DDT:
This PR seems to have gotten stale.
We can reopen this once this gets updated.

PR #2550, opened 2023-11-15
feat(helpers): add support for complex intermediate types

### LINKED ISSUE #1850: Add support for complex intermediate return types in fake method

ST-DDT:
Fixes #1850
Supersedes  #2381

- #1850
- #2381

This PR adds a new intermediate method `faker.helpers.eval` that replaces the entire fake pattern placeholder resolution that was previously located inside the `fake` method.

I extracted the resolution part in its own method because it would otherwise get too long/complex to read.

`faker.helpers.fake("{{person.firstName}})` -> `faker.helpers.eval("person.firstName") -> faker.person.firstName()`

The eval function is now also cable to resolve expressions that contain more than the previous two parts.

This is needed because of methods such as `faker.airline.airline()` that return complex objects.


src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (entrypoint[key as keyof typeof entrypoint] == null) {
+        try {
+          entrypoint = entrypoint();
+        } catch {
+          return undefined;
+        }
+      }
+
+      return entrypoint?.[key as keyof typeof entrypoint];
+    }
```
ST-DDT: I'm not sure about this implementation, because it prefers properties on the function over properties on its results:

`'airline.airline.name'` -> `'bound airline'` = function.name
`'airline.airline().name'` -> `'Lufthansa'`

If we let it prefer prefer properties on the results, then it will be impossible to access properties on the function that have the same name as the properties in the result.

`'airline.airline.name'` -> `'Lufthansa'`
`'airline.airline().name'` -> `'Lufthansa'`

test/modules/helpers.spec.ts
```diff
@@ -1026,34 +1130,23 @@ describe('helpers', () => {
...
-        it('does not allow invalid definitions data', () => {
-          expect(() => faker.helpers.fake('{{finance.credit_card}}')).toThrow(
-            new FakerError(`Invalid module method or definition: finance.credit_card
-- faker.finance.credit_card is not a function
-- faker.definitions.finance.credit_card is not an array`)
-          );
+        it('does allow complex data', () => {
+          const actual = faker.helpers.fake('{{science.unit}}');
+          expect(
+            faker.rawDefinitions.science?.unit?.map((value) =>
```
ST-DDT:
```suggestion
            faker.definitions.science.unit.map((value) =>
```

matthewmayer:
Does eval need to be public? It seems it could be an internal helper method. I don't think library users would normally have any need to use it?

ST-DDT:
> Does eval need to be public?

No, it doesnt need to be public. But it covers some usecases, that fake cannot because it is bound to be a string, whereas eval can return anything.

The eval function is useful for tools like:

- https://blog.dennisokeeffe.com/blog/2021-12-10-generating-fake-data-from-json-schema
- https://github.com/fstn/howto-typescript-data-generator

That wish to generate any data based on a schema definition, a jsdoc annotation, or a `@annotation`.

I'm fine with exposing it in a separate PR.

matthewmayer:
I think there might be logic in making it private first just because then we don't need to worry about any breaking changes, or publicly documenting it. Once we are clear it is stable and useful for third parties it could be exposed.

ST-DDT:
I agree with the publicly documenting part, but the expectations (except for the entrypoint parameters) are already defined by `fake` because it the actual implementation fake uses to resolve its stuff. IMO whether this is private or public is a small detail that can be adjusted anytime (before merge). What about the rest of the implementation?

matthewmayer:
Well if it's public that cannot be changed to private at "any time" only on a semver major release 😀

xDivisionByZerox:
I have to agree with @matthewmayer here. While there might be a use case for this method, I think it will be a really niche one. Even if not, we don't lose anything by making it internal for now. While, as already said, we might have some maintainable burdens when making it public first.

ST-DDT:
Sorry, there might have been some miscommunication here.
My issue was never whether the function was public or not, I only wanted some feedback on the actual implementation (and whether I should scrap the entire concept or whether it is worth continuing on).

I hid away the method now, so it does not cause any further confusion during reviews.

test/modules/helpers.spec.ts
```diff
@@ -1026,34 +1026,21 @@ describe('helpers', () => {
 
         it('does not allow invalid module name', () => {
           expect(() => faker.helpers.fake('{{foo.bar}}')).toThrow(
-            new FakerError(`Invalid module method or definition: foo.bar
-- faker.foo.bar is not a function
-- faker.definitions.foo.bar is not an array`)
-          );
-        });
-
-        it('does not allow missing method name', () => {
```
xDivisionByZerox, line 1036: Why did this test go?

test/modules/helpers.spec.ts
```diff
@@ -1149,6 +1136,13 @@ describe('helpers', () => {
         it('should not trim whitespace', () => {
           expect(faker.helpers.fake('   ---   ')).toBe('   ---   ');
         });
+
+        it('should support resolving a value in a complex object', () => {
+          const actual = faker.helpers.fake('{{airline.airline.iataCode}}');
```
xDivisionByZerox: I don't know if this is overengineered but should we first check whether `('{{airline.airline}}'` is of type `'object'`? I think that would make more clear what "complex" means in the test description. The same on the "does allow complex data" test.

test/modules/helpers-eval.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(actual).toBe(faker.string);
+  });
+
+  it('supports simple method calls', () => {
+    const actual = fakeEval('string.numeric', faker);
+    expect(actual).toBeTypeOf('string');
+    expect(actual).toMatch(/^\d$/);
+  });
+
+  it('supports method calls without arguments', () => {
```
xDivisionByZerox, line 32: Wouldn't it be more useful to spy on the methods here and check if the spy has been called?
This also applies to the later test cases.

src/modules/helpers/eval.ts
xDivisionByZerox: I'm not sure if it should access function properties at all, to be honest.

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (current.length === 0) {
+    throw new FakerError(`Expression '${expression}' cannot be resolved.`);
+  }
+
+  const value = current[0];
+  return typeof value === 'function' ? value() : value;
+}
+
+/**
+ * Evaluates a function call and returns the remaining string and the mapped results.
```
xDivisionByZerox: It doesn't return the remaining string, does it? It should be the current cursor position if I understood the code correctly. I think "remaining" is kinda confusing as a name in the return tuple.

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+function evalProcessFunction(
+  input: string,
+  entrypoints: ReadonlyArray<unknown>
+): [remaining: number, mapped: unknown[]] {
+  const [index, params] = findParams(input);
+  const nextChar = input[index + 1];
+  switch (nextChar) {
+    case '.':
+    case '(':
```
xDivisionByZerox, line 121: So this allows for functions that return functions to be called? Is that the reason for "(" to be a valid next character?

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * @param input The input string to parse.
+ */
+function findParams(input: string): [remaining: number, params: unknown[]] {
+  let index = input.indexOf(')');
+  while (index !== -1) {
+    const params = input.substring(1, index);
+    try {
+      // assuming that the params are valid JSON
+      return [index, JSON.parse(`[${params}]`) as unknown[]];
```
xDivisionByZerox, line 155: So for my option argument to work I have to use proper JSON format? Something like `faker.helpers.fake('{{internet.email({ firstName: "Foo", lastName: "Bar" })}}')` (no quoted property names) will throw an error. This might be a bad DX.
Unless I skipped over a normalization part outside of this function.

test/modules/helpers.spec.ts
ST-DDT: The PR supports varying depths, that includes a depth of 1.

src/modules/helpers/eval.ts
ST-DDT: Do you have a suggestion on how to fix that?

src/modules/helpers/eval.ts
ST-DDT: Yes. Similar to plain JS `factory()()`.

src/modules/helpers/eval.ts
ST-DDT: Oh, you are right. That is some leftover from a previous iteration.

src/modules/helpers/eval.ts
xDivisionByZerox: Alright, just wanted to be sure I understand this part correctly.

--- 

Although, now I'm thinking: Were would this be actually used in our library?

test/modules/helpers.spec.ts
xDivisionByZerox: I think I saw a test in "eval" which checks for exaclty this behaviour (resolving modules). Would you say that this really desired?

src/modules/helpers/eval.ts
ST-DDT: Removed for now.

src/modules/helpers/eval.ts
xDivisionByZerox: I'm not saying we have to support this, I felt that the acceptance criteria (what should work, what shouln't) for this feature are not that well defined. Maybe we should discuss this in the next team meeting?

src/modules/helpers/eval.ts
ST-DDT: Renamed

src/modules/helpers/eval.ts
ST-DDT: I plan to improve the regex support to return a factory instead of parsing the regex over and over again.

test/modules/helpers-eval.spec.ts
ST-DDT: Done

test/modules/helpers.spec.ts
ST-DDT: This is not really possible because fake always returns a string.

src/modules/helpers/eval.ts
ST-DDT: We can sure do that.
Please keep in mind, that this hasn't changed from before.
https://github.com/faker-js/faker/blob/d53c7ac708887ebad85ce65be026c6406295bbf6/src/modules/helpers/index.ts#L1468

(Well the fallback behavior has changed, but I think about reverting it or maybe iron it out.)

test/modules/helpers.spec.ts
xDivisionByZerox: You are right. I forgot this. Technically we could parse it and check, but I don't think we gain any information at that point.

src/modules/helpers/eval.ts
xDivisionByZerox: > Please keep in mind, that this hasn't changed from before.

Good point. In that case we can leave it as is. Maybe we can take a look at the DX side in another PR.

xDivisionByZerox (dismissed):

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,214 @@
...
+    // Remove garbage and resolve array values
+    current = current
+      .filter((value) => value != null)
+      .map((value): unknown =>
+        Array.isArray(value) ? faker.helpers.arrayElement(value) : value
+      );
+  } while (remaining.length > 0 && current.length > 0);
+
+  if (current.length === 0) {
+    throw new FakerError(`Expression '${expression}' cannot be resolved.`);
```
ST-DDT:
```suggestion
    throw new FakerError(`Cannot resolve expression '${expression}'`);
```

xDivisionByZerox (dismissed):

test/modules/__snapshots__/person.spec.ts.snap
```diff
@@ -50,7 +50,7 @@ exports[`person > 42 > suffix > with sex 1`] = `"III"`;
 
 exports[`person > 42 > zodiacSign 1`] = `"Gemini"`;
 
-exports[`person > 1211 > bio 1`] = `"infrastructure supporter, photographer 🙆‍♀️"`;
+exports[`person > 1211 > bio 1`] = `"teletype lover, dreamer 👄"`;
```
matthewmayer, line 53: Do you know why this specific snapshot changes? Does it point to anything deeper?

matthewmayer (changes requested):

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,214 @@
...
+ * It is also possible to use multiple parameters (comma separated).
+ *
+ * ```js
+ * const pin = fakeEval('string.numeric(4, {"allowLeadingZeros": true})', faker);
+ * ```
+ *
+ * This method can resolve expressions with varying depths (dot separated parts).
+ *
+ * ```ts
+ * const airlineModule = fakeEval('airline', faker); // AirlineModule
```
matthewmayer, line 42: `faker.helpers.fake("{{person}}")`

gives a weird error

```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'B'
    |     property 'datatype' -> object with constructor 'l'
    --- property 'faker' closes the circle
    at JSON.stringify (<anonymous>)
    at Y.fake (/Users/matt/Documents/git/faker/dist/cjs/modules/helpers/index.js:1:8351)
```

matthewmayer (changes requested):

test/modules/helpers-eval.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    expect(actual).toBe('This-Works');
+  });
+
+  it('should be able to return empty strings', () => {
+    const actual = fakeEval('string.alphanumeric(0)', faker);
+    expect(actual).toBeTypeOf('string');
+    expect(actual).toBe('');
+  });
+
+  it('supports returning complex objects', () => {
```
matthewmayer, line 96: `faker.helpers.fake("{{airline.airline.}}")` appears to work. Should we disallow a single trailing . like this?

matthewmayer (changes requested):

src/modules/helpers/eval.ts
```diff
@@ -0,0 +1,214 @@
...
+          // try again with the next index
+        }
+      }
+    }
+
+    index = input.indexOf(')', index + 1);
+  }
+
+  throw new FakerError(
+    `Function parameters cannot be parsed as JSON or simple string: '${input}'`
```
matthewmayer: can we add a test for this? There are some behavior changes here, for example in old version

`faker.helpers.fake("{{internet.email('')}}")`  works, but now it throws

matthewmayer (changes requested):

test/modules/__snapshots__/person.spec.ts.snap
ST-DDT: Yes, now the values are resolved in parallel, so if there is a matching method  and locale data, it will resolve (arrayElement) them both before selecting the first remaining.
This is required now because the paths can have arrays of complex objects and thus a simple lookahead for a property is no longer possible.

src/modules/helpers/eval.ts
ST-DDT: Fixed

test/modules/helpers-eval.spec.ts
ST-DDT: Added error test for this case

matthewmayer (dismissed):
`faker.helpers.fake("{{person.first_name()}}")` used to work but now throws. Maybe add a test to explicitly check for this?

In general due to subtle behavior changes like this i feel this should be considered breaking. I can imagine there are example in the wild of `faker.helpers.fake("{{person.first_name()}}")` instead of the correct `faker.helpers.fake("{{person.firstName()}}")` which will break.

xDivisionByZerox:
To not introduce breaking changes, could we ignore parenthesis for now?

ST-DDT:
Done

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2551, opened 2023-11-17
chore(deps): update eslint-define-config

Shinigami92:
Upgrade `eslint-define-config` to v2.0.0 with new community driven type support system

Currently there is not _vitest_ rules support, I can try add that to [vitest](https://www.npmjs.com/package/eslint-plugin-vitest) directly


package.json
```diff
@@ -89,6 +89,11 @@
   "devDependencies": {
     "@actions/github": "~6.0.0",
     "@algolia/client-search": "~4.19.1",
+    "@eslint-types/deprecation": "2.0.0-1",
+    "@eslint-types/jsdoc": "~46.9.0",
+    "@eslint-types/prettier": "5.0.1-1",
+    "@eslint-types/typescript-eslint": "~6.11.0",
+    "@eslint-types/unicorn": "~49.0.0",
```
xDivisionByZerox, line 96: So now I don't only need eslint itself plus the plugins I'm using, but also an extra package per plug-in for types only? I'm not sure how to feel about that. 

Additionally, the common practice so far was to configure the plug-ins as less as possible and only use presets. I'm questioning the general purpose of the "defineConfig" function if you go the way I described:

1. Simple configurations that only use presets (no custom rule config) can be made secure by referencing the JSON schema for the eslint configuration
2. Even if you have some rules that require custom configuration and you configure one rule incorrectly, you will get an error next time you run eslint. While the autocomplete is helpful when configuring one of those niche rules, you most likely are searching for the documentation of the rule anyway. 

I'm not saying the library itself is not useful. I'm only questioning the value of it in the Faker project and workflow.

Shinigami92: Yes, it is basically the same as with `@types/*` packages
Plugins now have the possibility to support eslint-define-config on their own and the maintenance burden is shifted away from the eslint-define-config project
For now only 6 plugins are supported, every other plugin should hopefully define their own support in their own packages

So e.g. hopefully we can do `/// <reference types="eslint-plugin-vitest/config-support" />` or similar in the future

ST-DDT: Are these types actively maintained right now?

ST-DDT:
> Currently there is not _vitest_ rules support, I can try add that to [vitest](https://www.npmjs.com/package/eslint-plugin-vitest) directly

What does that imply for our repository?

package.json
Shinigami92: For these `@eslint-types/*` nothing really changed, they are maintained the exactly same way as before
Except that eslint-define-config v1 is not maintained anymore

Shinigami92:
> > Currently there is not _vitest_ rules support, I can try add that to [vitest](https://www.npmjs.com/package/eslint-plugin-vitest) directly
> 
> What does that imply for our repository?

Nothing special, just that right there is no type-check and auto-completion for vitest rules until they ship support

ST-DDT (approved):

PR #2560, opened 2023-11-25
docs: add missing throw descriptions in JSDocs

xDivisionByZerox:
Add missing and updated `@throw` descriptions in the JSDocs for the following functions:

| Updated               | Added                 |
| --------------------- | --------------------- |
| faker.number.int      | faker.number.int      |
| faker.number.float    | faker.number.float    |
| faker.number.binary   | faker.number.binary   |
| faker.number.octal    | faker.number.octal    |
| faker.number.hex      | faker.number.hex      |
| faker.number.bigInt   |                       |
| faker.datatype.number | faker.datatype.number |
|                       | faker.datatype.float  |

Additionally, to make sure that those statements are actually true, I adjusted or extended the test for those functions as well.


ST-DDT:
This is somewhat blocked by #2444

- #2444

ST-DDT (dismissed):

xDivisionByZerox:
> This is somewhat blocked by https://github.com/faker-js/faker/pull/2444

Should we add https://github.com/faker-js/faker/labels/s:%20on%20hold?

ST-DDT:
The jsdocs get improved by this regardless, so I don't consider it worth hard blocking.
(It only looks slightly ugly in the api-docs)

src/modules/number/index.ts
```diff
@@ -24,7 +24,8 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `Number.MAX_SAFE_INTEGER`.
    *
-   * @throws When options define `max < min`.
+   * @throws When `min` is greater than `max`.
+   * @throws When `min` and `max` do not provide a sufficient range to generate integers in between them.
```
```suggestion
   * @throws When `min` and `max` do not form a sufficient range to generate integers from.
```

src/modules/number/index.ts
```diff
@@ -168,7 +172,8 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `1`.
    *
-   * @throws When options define `max < min`.
+   * @throws When `min` is greater than `max`.
+   * @throws When `min` and `max` do not provide a sufficient range to generate integers in between them.
```
```suggestion
   * @throws When `min` and `max` do not form a sufficient range to generate integers from.
```

src/modules/number/index.ts
```diff
@@ -217,7 +222,8 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `7`.
    *
-   * @throws When options define `max < min`.
+   * @throws When `min` is greater than `max`.
+   * @throws When `min` and `max` do not provide a sufficient range to generate integers in between them.
```
```suggestion
   * @throws When `min` and `max` do not form a sufficient range to generate integers from.
```

src/modules/number/index.ts
```diff
@@ -266,7 +272,8 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.min Lower bound for generated number. Defaults to `0`.
    * @param options.max Upper bound for generated number. Defaults to `15`.
    *
-   * @throws When options define `max < min`.
+   * @throws When `min` is greater than `max`.
+   * @throws When `min` and `max` do not provide a sufficient range to generate integers in between them.
```
```suggestion
   * @throws When `min` and `max` do not form a sufficient range to generate integers from.
```

src/modules/number/index.ts
```diff
@@ -313,7 +320,7 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.min Lower bound for generated bigint. Defaults to `0n`.
    * @param options.max Upper bound for generated bigint. Defaults to `min + 999999999999999n`.
    *
-   * @throws When options define `max < min`.
+   * @throws When `min` is greater to `max`.
```
```suggestion
   * @throws When `min` is greater than `max`.
```

src/modules/number/index.ts
matthewmayer: How about just "when there are no integers between min and max"

ST-DDT:
This is no longer soft blocked by #2444.

- #2444

src/modules/number/index.ts
xDivisionByZerox:
```suggestion
   * @throws When there are no integers between `min` and `max`.
```

matthewmayer (approved):

ST-DDT (approved):

ejcheng (approved):

PR #2562, opened 2023-11-28
infra(tsconfig): noImplicitAny

### LINKED ISSUE #1837: Turn on noImplicitAny in tsconfig

ST-DDT:
Fixes #1837

- #1837


ST-DDT:
I'm really unhappy with typescripts behavior here and I hope this gets better in a future version.

scripts/generate-locales.ts
```diff
@@ -23,10 +23,8 @@ import {
 } from 'node:fs';
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import type { Options } from 'prettier';
-import { format } from 'prettier';
-import options from '../.prettierrc.js';
 import type { LocaleDefinition, MetadataDefinition } from '../src/definitions';
+import { formatMarkdown, formatTypescript } from './apidoc/format.js';
```
xDivisionByZerox: Is the `*.js` correct here?

xDivisionByZerox:
Are you talking about `Object.keys` not returning the type `(keyof T)[]`? If thats the case, there is [a PR comment in the typescript repository](https://github.com/Microsoft/TypeScript/pull/12253#issuecomment-263132208) that explains why it is this way.

We could make an internal helper function that is "correctly" typed:

```ts
function keys<T extends {}>(obj: T): (keyof T)[] {
  return Object.keys(obj) as (keyof T)[];
}
```

That way, we only have to cast in one place.

scripts/apidoc/format.ts
```diff
@@ -1,5 +1,6 @@
 import type { Options } from 'prettier';
 import { format } from 'prettier';
+// @ts-expect-error: No types available.
 import prettierConfig from '../../.prettierrc.js';
```
xDivisionByZerox: Would we get type information if this was a JSON file?

scripts/generate-locales.ts
ST-DDT: Fixed.

scripts/apidoc/format.ts
ST-DDT: I worked around the problem.

.prettierrc.d.ts
```diff
@@ -0,0 +1,4 @@
+import type { Options } from 'prettier';
+
+declare const options: Options;
+export = options;
```
ST-DDT, line 4: Or is that better:

```suggestion
declare module './.prettierrc' {
  const options: Options;
  export default options;
}
```

xDivisionByZerox: But now we have an additional file to make the config work. Not sure about that.

ST-DDT: I just tested JSON as well, but there seems to be issues with string vs union types.

grafik

Should I revert to the `ts-expect-error` then?

xDivisionByZerox: If JSON does not work out of the box I'm fine with the custom definition file, for now. We might want to look into this at a later point again. Or maybe, prttier will allow TS configuration files, in the future.

xDivisionByZerox (approved):

PR #2563, opened 2023-11-30
feat(helpers)!: stricter checking for function signature passed to `multiple`

pomali:
faker.helpers.multiple() supports using mapFn like in Array.from(), but type signature of function didn't match, PR fixes this and adds test for this case 

(but since it changes declared interface I chose `feat` instead of `fix` as semantic type)


xDivisionByZerox:
I'm not sure if this fits the use case of `faker.helpers.multiple` :thinking:

If you require the parameters of the mal function, you are most likely doing some static value generation. 

I need some time to think about this in more detail.

pomali:
Yes, usecase here is generating two sets of objects with matching ids, so I can "join" them later.
Eg.

```
interface School { locationId: number }
interface Location { id: number }
```

I could create an function with no arguments that simulates this behavior, but it seemed busywork since underlying code already supports it.

I am using it in place without shared state (in mock service worker in two different endpoints)

src/modules/helpers/index.ts
```diff
@@ -1246,11 +1246,12 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    * @example
    * faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
    * faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple((_el, i) => `Person ${i + 1}` ) // [ 'Person 1', 'Person 2', 'Person 3' ]
    *
    * @since 8.0.0
    */
   multiple<TResult>(
-    method: () => TResult,
+    method: (() => TResult) | ((v: unknown, k: number) => TResult),
```
ST-DDT: Since the first value is always undefined, I would omit it from the signature.
Also please name the k parameter index or something more expressive.

ST-DDT (changes requested):
The index parameter might be useful when generating a list of entries that should have sequential ids.

However I would like to block this until v9.0 development commences, because then the method might get a new parameter (ref fakerCore) anyway.

src/modules/helpers/index.ts
ST-DDT: ~~If you make the parameter optional then you dont need two signatures.~~

rtm-ctrlz: There is no need for two signatures even without making parameters optional, because parameters will be passed any way, no matter how callback or types for callback parameter was defined.

And by the way, it is not an error for a function not to have parameters in case where they will passed, take event handlers for example.

Shinigami92: So @ST-DDT and I discussed about this and thinking about to do something like `return Array.from({ length: count }, (_, index) => method(faker, index));`.

We will discuss this with the rest of the team in one of the next meetings.

ST-DDT: **Team Decision**

We will add this and will start with this:

```suggestion
    method: (v: unknown, index: number) => TResult,
```

After/in the process of https://github.com/faker-js/faker/issues/2667 we will then replace the `v` with a proper value.

ST-DDT:
Sorry for the delay in feedback.
Could you please update this PR with the latest suggestions?

pomali:
> Sorry for the delay in feedback. Could you please update this PR with the latest suggestions?

No problem, sorry for mine delay. Updated.

ST-DDT:
Could you please fix the broken tests and examples?
The tests were already broken due to the wrong typing of the method before, but now the method actually requires proper types and thus the bad usage actually shows up.

`firstName` -> `() -> firstName()`

ST-DDT:
Please also enable `allow edits by maintainers` on this PR so we can update it ourself to the latest commits on next.

test/modules/helpers.spec.ts
```diff
@@ -173,11 +173,14 @@ describe('helpers', () => {
...
-        .it('with method and count range', faker.number.int, {
+      t.it('with only method', () => faker.number.int())
+        .it('with method and count', () => faker.number.int(), {
+          count: 5,
+        })
+        .it('with method and count range', () => faker.number.int(), {
           count: { min: 1, max: 10 },
-        });
+        })
+        .it('with method using index', (_el, i) => i * 3);
```
ST-DDT:
```suggestion
        .it('with method using index', (_, i) => i * 3);
```

test/modules/helpers.spec.ts
```diff
@@ -1144,30 +1147,46 @@ describe('helpers', () => {
...
+            }
+          );
           expect(result).toBeTypeOf('object');
           expect(Array.isArray(result)).toBe(true);
           expect(result.length).toBeGreaterThanOrEqual(1);
           expect(result.length).toBeLessThanOrEqual(10);
         });
+
+        it('should generate values using index of created value', () => {
+          const result = faker.helpers.multiple((_el, i) => i * 2, {
```
ST-DDT:
```suggestion
          const result = faker.helpers.multiple((_, i) => i * 2, {
```

ST-DDT (dismissed):

ST-DDT:
Do we consider this breaking?

Because previously you could incorrectly pass any function reference directly to the method even, if it would be called with garbage values, so it is kind of a bugfix, but one that prevents you from doing bad things.

Shinigami92 (dismissed):

Shinigami92:
> Do we consider this breaking?

I think it's better to mark it as breaking as not doing it and then it was breaking

ST-DDT (dismissed):

rtm-ctrlz (approved):
GitHub still asking me for review :)

In my opinion this change should be accepted and marked as breaking change.

In most cases this change should not lead to breaks, but it worth a notice.

docs/guide/upgrading_v9/2563.md
```diff
@@ -0,0 +1,27 @@
...
+
+We now expose the parameters from the `method` parameter to our users.
+That allows for usecases where the index is part of the generated data e.g. as id.
+
+```ts
+faker.helpers.multiple((_, index) => ({ id: index, ...}), ...); // [{id: 0, ...}, ...]
+```
+
+While the actual implementation of the method hasn't changed, it's signature has.
+Previously it was possible to incorrectly pass any function reference to `faker.helpers.multiple`,
```
matthewmayer: I think we can write this in a more friendly way. Instead of saying previously it was possible to "incorrectly" pass any function reference to faker.helpers.multiple - this is exactly how it is documented to work at the moment at https://fakerjs.dev/api/helpers#multiple 

Maybe a better example is to show a method that has a second parameter.

So for example code like:

`faker.helpers.multiple(faker.date.past, {count:2})`
//[ 1969-06-28T19:24:59.269Z, 1969-06-10T02:42:19.045Z ]
works differently to
`faker.helpers.multiple(()=>faker.date.past(), {count:2})`
// [ 2023-09-05T12:16:22.537Z, 2023-05-01T15:15:18.872Z ]

Because the 2nd parameter (refDate) is unexpectedly passed the indexes 0 and 1, ie relative to datetime 0, around the 1970 UNIX timestamp epoch.



[
  1969-03-07T18:23:19.963Z, 1969-06-12T21:22:55.144Zh
]

src/modules/helpers/index.ts
```diff
@@ -1135,11 +1135,12 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    * @example
    * faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
    * faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
```
matthewmayer: change this to use `() => faker.person.firstName()` ?

src/modules/helpers/index.ts
```diff
@@ -1135,11 +1135,12 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    * @example
```
matthewmayer, line 1136: we could also change 
`@param method The method used to generate the values.` to indicate that this will be called with (_, index)

matthewmayer (changes requested):

docs/guide/upgrading_v9/2563.md
matthewmayer: Suggested rewrite:

### Stricter checking for function signature passed to `faker.helpers.multiple` method

The `faker.helpers.multiple` method takes a function reference as its first parameter. Previously you may have written code like this to generate multiple values.
```ts
faker.helpers.multiple(faker.date.past, {count:2})
```

However this code has a bug - `faker.helpers.multiple` passes the loop index as the second parameter to the method, which in this case would set the `refDate` of the `faker.date.past()` call to 0, making all dates before 1970.

Instead you should generally use a lambda function like
```ts
faker.helpers.multiple(()=>faker.date.past(), {count:2})
```
to get the desired behavior. In v9.0, we use stricter type-checking in Typescript to detect when a function is called which is not compatible with `(v: unknown, index: Number)` which can cause compile-time errors in places where previously there were potential runtime errors.

**Bad**

```ts
faker.helpers.multiple(faker.person.firstName, ...); // ❗
// In Typescript, this is now a compile time error
// Argument of type '(sex?: "female" | "male" | undefined) => string'
// is not assignable to parameter of type '(v: unknown, index: number) => unknown'.
```

**Good**

```ts
faker.helpers.multiple(() => faker.person.firstName(), ...); // ✔
```

The new types also allow for easier use-cases where the index is part of the generated data e.g. as id.

```ts
faker.helpers.multiple((_, index) => ({ id: index, ...}), ...); // [{id: 0, ...}, ...]
```

matthewmayer (changes requested):

docs/guide/upgrading_v9/2563.md
ST-DDT: Thanks for these suggestions, they really help a lot.

src/modules/helpers/index.ts
ST-DDT: Good catch

src/modules/helpers/index.ts
ST-DDT: Edited

ST-DDT (dismissed):

src/modules/helpers/index.ts
```diff
@@ -1129,17 +1129,19 @@ export class SimpleHelpersModule extends SimpleModuleBase {
...
+   * The method will be called with `(_, index)`, to allow using the index in the generated value e.g. as id.
    * @param options The optional options object.
    * @param options.count The number or range of elements to generate. Defaults to `3`.
    *
    * @example
-   * faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
-   * faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple(() => faker.person.firstName()) // [ 'Aniya', 'Norval', 'Dallin' ]
+   * faker.helpers.multiple(() => faker.person.firstName(), { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple((_, i) => `Person ${i + 1}` ) // [ 'Person 1', 'Person 2', 'Person 3' ]
```
matthewmayer: This example seems a bit artificial as there's not any fake part to the function. Maybe like `${faker.color.human()} ${i+1}`

```suggestion
   * faker.helpers.multiple((_, i) => `${faker.color.human()}-${i + 1}` ) // [ 'orange-1', 'orchid-2', 'sky blue-3' ]
```

matthewmayer (changes requested):

src/modules/helpers/index.ts
```diff
@@ -1129,17 +1129,19 @@ export class SimpleHelpersModule extends SimpleModuleBase {
...
+   * The method will be called with `(_, index)`, to allow using the index in the generated value e.g. as id.
    * @param options The optional options object.
    * @param options.count The number or range of elements to generate. Defaults to `3`.
    *
    * @example
-   * faker.helpers.multiple(faker.person.firstName) // [ 'Aniya', 'Norval', 'Dallin' ]
-   * faker.helpers.multiple(faker.person.firstName, { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple(() => faker.person.firstName()) // [ 'Aniya', 'Norval', 'Dallin' ]
+   * faker.helpers.multiple(() => faker.person.firstName(), { count: 3 }) // [ 'Santos', 'Lavinia', 'Lavinia' ]
+   * faker.helpers.multiple((_, i) => `${faker.color.human()}-${i + 1}` ) // [ 'orange-1', 'orchid-2', 'sky blue-3' ]
```
```suggestion
   * faker.helpers.multiple((_, i) => `${faker.color.human()}-${i + 1}`) // [ 'orange-1', 'orchid-2', 'sky blue-3' ]
```

matthewmayer (approved):

ST-DDT (approved):

Shinigami92 (approved):
Really good migration guide :+1:

PR #2564, opened 2023-11-30
refactor(number): deprecate precision in favor of multipleOf in float

### LINKED ISSUE #1596: Refactor `number.float` parameter `precision` to `fractionDigits`

xDivisionByZerox:
## Description

This PR deprecated the property `precision` in `faker.number.float`. Additionally, adds the property `multipleOf`. 

## Additional information

Fixes #1596.
Extracted from #1855.


src/modules/number/index.ts
```diff
@@ -84,22 +85,26 @@ export class NumberModule extends SimpleModuleBase {
   }
 
   /**
-   * Returns a single random floating-point number for a given precision or range and precision.
-   * The lower bound is inclusive, the upper bound is exclusive, unless precision is passed.
+   * Returns a single random floating-point number for a given multipleOf or range and multipleOf.
```
ST-DDT: This parts sounds off to me.

Maybe use just:

```suggestion
   * Returns a single random floating-point number.
```

src/modules/number/index.ts
```diff
@@ -84,22 +85,26 @@ export class NumberModule extends SimpleModuleBase {
...
+   * The lower bound is inclusive, the upper bound is exclusive, unless multipleOf is passed.
```
```suggestion
   * The lower bound is inclusive, the upper bound is exclusive, unless `multipleOf` is passed.
```

src/modules/number/index.ts
```diff
@@ -84,22 +85,26 @@ export class NumberModule extends SimpleModuleBase {
...
    *
    * @param options Upper bound or options object.
    * @param options.min Lower bound for generated number. Defaults to `0.0`.
    * @param options.max Upper bound for generated number. Defaults to `1.0`.
    * @param options.precision Precision of the generated number, for example `0.01` will round to 2 decimal points.
    * If precision is passed, the upper bound is inclusive.
```
ST-DDT, line 96: Maybe mention that it is deprecated here as well?
```suggestion
   * @param options.precision Precision of the generated number, for example `0.01` will round to 2 decimal points.
   * If precision is passed, the upper bound is inclusive.
   * This option is deprecated, please use `multipleOf` instead.
```

?

src/modules/number/index.ts
```diff
@@ -84,22 +85,26 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @param options.multipleOf The generated number will be a multiple of this property.
+   * This property can be used to achieve specific decimal points.
```
```suggestion
   * This property can be used to limit the result to a specific number of decimal digits.
```

(I have never used/heart of `decimal points`, but that might just be me not being familiar with it)

ST-DDT (commented):
It looks like some methods still use the precision parameter.

src/modules/number/index.ts
xDivisionByZerox: I searched for similar occurencens (Faker constructor) and the JSDocs do not mention the deprecation there as well 🤔

test/modules/number.spec.ts
```diff
@@ -258,6 +279,22 @@ describe('number', () => {
         expect(results).toEqual([0, 0.4, 0.8, 1.2, 1.6]);
       });
 
+      it('provides numbers with a given multipleOf of 0.4 steps', () => {
+        const results = [
+          ...new Set(
+            Array.from({ length: 50 }, () =>
```
matthewmayer: Maybe do 100. 50 will fail every ~70000 runs.

matthewmayer (dismissed):

test/modules/number.spec.ts
```diff
@@ -269,15 +306,38 @@ describe('number', () => {
...
+            multipleOf: 0.01,
+          });
+          expect(actual).toBe(Number(actual.toFixed(2)));
+        }
+      });
+
       it('throws an error for precision 0', () => {
         expect(() => faker.number.float({ precision: 0 })).toThrow(
-          new FakerError('Precision should be greater than 0.')
+          new FakerError('Multiple of should be greater than 0.')
```
xDivisionByZerox: Is this correct? Since we give "precision" and receive an error talking about "multipleOF".

src/modules/number/index.ts
```diff
@@ -141,12 +163,12 @@ export class NumberModule extends SimpleModuleBase {
       throw new FakerError(`Max ${max} should be greater than min ${min}.`);
     }
 
-    if (precision !== undefined) {
-      if (precision <= 0) {
-        throw new FakerError(`Precision should be greater than 0.`);
+    if (multipleOf !== undefined) {
+      if (multipleOf <= 0) {
+        throw new FakerError(`Multiple of should be greater than 0.`);
```
xDivisionByZerox: Should this directly recall the name of the property? So, rephrase "Multiple of" to "multipleOf"?

src/modules/helpers/index.ts
```diff
@@ -978,7 +978,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
     const random = this.faker.number.float({
       min: 0,
       max: total,
-      precision: 1e-9,
+      multipleOf: 1e-9,
```
ST-DDT: I think this is a max precision float, so we could potentially remove the option from here.

src/modules/number/index.ts
ST-DDT: I would refer to the property, because otherwise you still have broken English.

test/modules/number.spec.ts
ST-DDT: I think it's fine. Alternatively you could do `multipleOf/precision should be greater than 0.` (with a todo comment for cleanup)

src/modules/helpers/index.ts
xDivisionByZerox: Is max precision not 18? 🤔 

```ts
console.log(faker.helpers.multiple(() => faker.number.float().toString().split('.')[1].length, { count: 10 }));
```

Gives the following:

```json
[17,16,16,17,18,17,17,16,15,16]
```

ST-DDT: I think this should/can be a max precision float. 
(It was probably just forgotten during the "default 2 precision -> max precision float" refactor)

xDivisionByZerox: Ok, I misunderstood you the first time. Done.

ST-DDT (dismissed):

ST-DDT:
@xDivisionByZerox There appears to be merge conflicts.

ST-DDT (approved):

matthewmayer (approved):

PR #2574, opened 2023-12-11
docs: clarify that location.state refers to ISO 3166-2

### LINKED ISSUE #2360: Clarify that location state refers to ISO 3166-2

Currently the definition of location.state is not very precise.

### Suggested solution

Mention that it should refer to [ISO_3166-2](https://en.wikipedia.org/wiki/ISO_3166-2).

### Alternative

_No response_

### Additional context

_No response_

ejcheng:
closes #2360

This PR makes the docs of `location.state()` slightly more precise by explaining that the method generally returns the ISO 3166-2 subdivisions for a country.


src/modules/location/index.ts
```diff
@@ -372,6 +372,7 @@ export class LocationModule extends ModuleBase {
 
   /**
    * Returns a random localized state, or other equivalent first-level administrative entity for the locale's country such as a province or region.
+   * Generally, these are the ISO 3166-2 subdivisions for a country. However, not all locales have a country attached to them (e.g. en, ar, es).
```
matthewmayer:
```suggestion
* Generally, these are the ISO 3166-2 subdivisions for a country. 
* If a locale doesn't have a country, the method may return ISO 3166-2 subdivisions from one or more countries that uses that language, for example the `ar` locale includes subdivisions from Arabic-speaking countries such as Tunisia, Algeria, Syria, Lebanon, etc. 
* For historical compatibility reasons, the default `en` locale only includes states in the United States (identical to `en_US`), however you can use other English locales such as `en_IN`, `en_GB`, `en_AU` if needed.
```

xDivisionByZerox (approved):

ST-DDT:
Does this also fixes https://github.com/faker-js/faker/issues/1944

- https://github.com/faker-js/faker/issues/1944

src/modules/location/index.ts
```diff
@@ -372,6 +372,9 @@ export class LocationModule extends ModuleBase {
 
   /**
    * Returns a random localized state, or other equivalent first-level administrative entity for the locale's country such as a province or region.
+   * Generally, these are the ISO 3166-2 subdivisions for a country.
+   * If a locale doesn't correspond to one specific country, the method may return ISO 3166-2 subdivisions from one or more countries that uses that language. For example, the `ar` locale includes subdivisions from Arabic-speaking countries, such as Tunisia, Algeria, Syria, Lebanon, etc.
+   * For historical compatibility reasons, the default `en` locale only includes states in the United States (identical to `en_US`). However, you can use other English locales, such as `en_IN`, `en_GB`, and `en_AU`, if needed.
```
ST-DDT, line 377: Should we change that in v9.0?

ST-DDT (approved):
Should we put these description parts also on the locale definitions?

PR #2578, opened 2023-12-13
feat(phone)!: add new style parameter

### LINKED ISSUE #1542: Add phone.number generation options

matthewmayer:
fix #1542

Script for doing the grunt work of renaming the formats.ts and generating the initial raw and national files

```ts
const {
    allLocales
} = require("@faker-js/faker")
const lpn = require("libphonenumber-js")
const fs = require("fs")
const fsx = require("fs-extra")
//guess the country for locales with no country attached
const fallbacks = {
    "az": "AZ",
    "da": "DK",
    "de": "DE",
    "dv": "MV",
    "el": "GR",
    "en": "US",
    "es": "ES",
    "fa": "IR",
    "fr": "FR",
    "he": "IL",
    "hr": "HR",
    "hu": "HU",
    "hy": "AM",
    "it": "IT",
    "ja": "JP",
    "ko": "KR",
    "lv": "LV",
    "mk": "MK",
    "ne": "NP",
    "nl": "NL",
    "pl": "PL",
    "ro": "RO",
    "ru": "RU",
    "sk": "SK",
    "sv": "SE",
    "th": "TH",
    "tr": "TR",
    "uk": "UA",
    "vi": "VN"
}

for (let locale of Object.keys(allLocales)) {
    //loop through all locales that have an existing formats.ts
    const filename = "../faker/src/locales/" + locale + "/phone_number/formats.ts"
    if (fs.existsSync(filename)) {
        //extract formats array from TS
        const data = fs.readFileSync(filename, "utf8")
        let arrayString = data.replace('export default ', '').replace(';', '');
        let formats = eval(arrayString)
        //figure out the country code
        let country = null
        if (locale.includes("_")) {
            country = locale.split("_")[1]
        } else if (fallbacks[locale]) {
            country = fallbacks[locale]
        }
        if (country) {
            let countryCode = lpn.getCountryCallingCode(country).toString()
            let nationals = []
            let raws = []
            for (let format of formats) {
                if (!format.includes("x")) { //skip extensions
                    //replace each # in the format string with placeholder
                    //we need to make sure its not one of the ones in the format or country code
                    //so we can losslessly restore it later
                    //so start at 2, and increment until we find a free one
                    let hash = 2
                    while (format.includes(hash) || countryCode.includes(hash)) {
                        hash++
                    }
                    //do the same for !
                    let pling = hash + 1
                    while (format.includes(pling) || countryCode.includes(pling)) {
                        pling++
                    }
                    //replace the # and ! with the placeholders
                    let phone = format.replaceAll('#', hash).replaceAll('!', pling)
                    //later we'll need to restore them
                    const restorePlaceholders = (string) => {
                        return string.replaceAll(hash, "#").replaceAll(pling, "!")
                    }
                    try {
                        //parse the phone number
                        const phoneNumber = lpn.parsePhoneNumber(phone, country)
                        //format national
                        let national = restorePlaceholders(phoneNumber.formatNational())
                        nationals.push(national)
                        //format raw
                        const raw = restorePlaceholders(phoneNumber.getURI().replace("tel:", ""))
                        raws.push(raw)
                    } catch (e) {
                        //
                    }
                }
                //move the file from formats.ts to human.ts
                try {
                    fsx.moveSync(filename, filename.replace("formats", "human"))
                } catch (e) {

                }
                //write out the new files, avoiding dupes
                fs.writeFileSync(filename.replace("formats", "national"), "export default " + JSON.stringify([...new Set(nationals)]) + ";")
                fs.writeFileSync(filename.replace("formats", "raw"), "export default " + JSON.stringify([...new Set(raws)]) + ";")

            }
        }
    }
}
```


src/definitions/phone_number.ts
```diff
@@ -12,5 +12,7 @@ export type PhoneNumberDefinition = LocaleEntry<{
    *
    * @see faker.helpers.replaceSymbolWithNumber(format): For more information about how the patterns are used.
    */
-  formats: string[];
+  human: string[];
+  national: string[];
+  raw: string[];
```
ST-DDT: I'm thinking about nesting these.

Because if you have a different format in one part it makes sense to have a different format in the others as well.
 
This should prevent the same issues as we have/had with firstnames.

Maybe name the section format again and we can refactor it in a different pr to split it into landline and mobile phone.

````ts
format: { human, national, international }
````

**Comment by matthewmayer** [comment]
We'll need rebase after https://github.com/faker-js/faker/pull/2712 lands

**Review by ST-DDT — COMMENTED** [review]
I forgot this had a pending comment...

File: src/definitions/phone_number.ts (hunk 1 above)
Comment (matthewmayer) [reply src/definitions/phone_number.ts]: Sorry don't really follow. What does nesting achieve?

Comment (ST-DDT) [reply src/definitions/phone_number.ts]: If you replace the entry, that automatically replaces all the children.

**Flat**

````ts
const localeA = {
  phone: {
    international: ['A'],
    human: ['A'],
  }
};

const localeB = {
  phone: {
    international: ['B'],
  }
};

mergeLocales(localeA , localeB); 
=>
 {
  phone: {
    international: ['B'],
    human: ['A'],
  }
}
````

Which might not be a problem, by the time B is added, but if the data change in A it may not longer be compatible with B.

**Nested**

````ts
const localeA = {
  phone: {
    format: {
      international: ['A'],
      human: ['A'],
    }
  }
};

const localeB = {
  phone: {
    format: {
      international: ['B'],
    }
  }
};

mergeLocales(localeA , localeB); 
=>
 {
  phone: {
    format: {
      international: ['B'],
    }
  }
}
````

On the entry level the data might be randomly missing. On the nested level, it is ensured that all variants are present.
And since a single method uses them, it makes it a lot easier to implement since you don't have to check for missing data for the exact usecase the user is currently using.

Comment (matthewmayer) [reply src/definitions/phone_number.ts]: format or formats

Comment (ST-DDT) [reply src/definitions/phone_number.ts]: > format or formats

`format` we always use singular

Comment (matthewmayer) [reply src/definitions/phone_number.ts]: Also note it's currently phone_number not phone

So definitions would be at say locales/foo/phone _number/format/raw.ts

?

File: src/locales/ar/phone_number/formats.ts
Comment (ST-DDT) [inline src/locales/ar/phone_number/formats.ts:1]: What happened to this file/locale?

File: src/modules/phone/index.ts [hunk 2]
```diff
@@ -20,10 +20,39 @@ export class PhoneModule extends ModuleBase {
...
+    /**
+     * Style of the generated phone number:
+     * - `'human'`: (default) A human-input phone number, e.g. 555-770-7727 or 555.770.7727 x1234
+     * - `'national'`: A phone number in a standardized national format, e.g. (555) 123-4567.
+     * - `'raw'`: A phone number in the raw format, e.g. +15551234567
+     *
+     * @default 'human'
+     */
+    style: 'human' | 'national' | 'raw';
+  }): string {
```
ST-DDT: Please merge these call signatures into one.

And use the usual option default mechanics.

src/modules/phone/index.ts
```diff
@@ -20,10 +20,39 @@ export class PhoneModule extends ModuleBase {
...
+    const { style } = options ?? { style: 'human' };
+    const formats = this.faker.definitions.phone_number.format;
+
+    const definitions = formats[style] || formats.human;
```
ST-DDT: AFAICT `formats[style]` should always be present.

src/modules/phone/index.ts
```diff
@@ -20,10 +20,39 @@ export class PhoneModule extends ModuleBase {
...
+   *
+   * @since 7.3.0
+   */
+  number(options?: { style: 'human' | 'national' | 'raw' }): string;
+  number(options?: {
+    /**
+     * Style of the generated phone number:
+     * - `'human'`: (default) A human-input phone number, e.g. 555-770-7727 or 555.770.7727 x1234
+     * - `'national'`: A phone number in a standardized national format, e.g. (555) 123-4567.
+     * - `'raw'`: A phone number in the raw format, e.g. +15551234567
```
ST-DDT: Should we call this `international`? We should also mention that it is a standardized format.

src/modules/phone/index.ts
```diff
@@ -20,10 +20,39 @@ export class PhoneModule extends ModuleBase {
...
+   * @example
+   * faker.phone.number() // '961-770-7727'
+   *
+   * @since 7.3.0
+   */
+  number(options?: { style: 'human' | 'national' | 'raw' }): string;
+  number(options?: {
+    /**
+     * Style of the generated phone number:
+     * - `'human'`: (default) A human-input phone number, e.g. 555-770-7727 or 555.770.7727 x1234
```
ST-DDT:
```suggestion
     * - `'human'`: (default) A human-input phone number, e.g. `555-770-7727` or `555.770.7727 x1234`
```

Same for the others.

src/modules/phone/index.ts
```diff
@@ -20,10 +20,39 @@ export class PhoneModule extends ModuleBase {
...
+   * Generates a random phone number.
+   *
+   * @param options Options object
+   * @param options.style Style of the phone number. Defaults to 'human'
+   *
+   * @see faker.string.numeric(): For generating a random string of numbers.
+   * @see faker.helpers.fromRegExp(): For generating a phone number matching a regular expression.
+   *
+   * @example
+   * faker.phone.number() // '961-770-7727'
```
ST-DDT, line 22: Please add an example for each style.

ST-DDT (dismissed):

src/locales/ar/phone_number/formats.ts
matthewmayer: It was a dupe of the en locale using US country code etc. I think better to not include and let it fall back to en. Real ar data could be added later.

src/modules/phone/index.ts
matthewmayer: Not in JS

src/modules/phone/index.ts
matthewmayer: I prefer raw it's easier to tell apart from national.

matthewmayer: The format appears to be E.123, although I'm not sure how well known that is.

src/modules/phone/index.ts
ST-DDT: If the user's don't input a valid style name, then they are on their own.

See also: https://github.com/faker-js/faker/blob/next/docs/guide/upgrading_v9/2508.md

src/modules/phone/index.ts
ST-DDT: The problem is, `raw` and `E123` doesn't mean anything for most people.
For me `raw` is before sanitization/standarization, more like human input.

src/modules/phone/index.ts
matthewmayer: returning `undefined` in this case would seem more consistent?

src/modules/phone/index.ts
matthewmayer: sorry what is the "usual option default mechanics."?

src/modules/phone/index.ts
matthewmayer: if we remove the fallback we get a cyrptic error
```
Uncaught:
u [Error]: Calling `faker.helpers.arrayElement()` without arguments is no longer supported.
    at j.arrayElement (/Users/matt/Documents/git/faker/dist/chunk-2L32Z2V2.cjs:4:16849)
    at se.number (/Users/matt/Documents/git/faker/dist/chunk-2L32Z2V2.cjs:8:364)
```

so i think its better to throw a more helpful error

docs/guide/upgrading_v9/2578.md
```diff
@@ -0,0 +1,15 @@
...
+
+`faker.phone.number()` generates a phone number for the current locale. However, previously there was little control over the generated number, which might or might not include country codes, extensions, white space and punctuation.
+
+If you wanted more control over the number, it was previously necessary to pass an explicit `format` parameter. This has now been removed. Instead, you can consider one of two options:
+
+1. The new `style` parameter has convenient options for common use cases. There are three possible values.
+
+- - `'human'`: (default, existing behavior) A human-input phone number, e.g. `555-770-7727` or `555.770.7727 x1234`
+- - `'national'`: A phone number in a standardized national format, e.g. `(555) 123-4567`.
+- - `'raw'`: A phone number in a raw international forma with country code, e.g. `+15551234567`
```
ejcheng:
```suggestion
- - `'raw'`: A phone number in a raw international format with country code, e.g. `+15551234567`
```

src/modules/phone/index.ts
ST-DDT:
````ts
phoneNumber(options: {style?: ...} = {}) {
  const { style = 'human' } = options;
}
````

File: src/modules/phone/index.ts [hunk 8]
```diff
@@ -12,18 +12,40 @@ export class PhoneModule extends ModuleBase {
   /**
    * Generates a random phone number.
    *
+   * @param options Options object
+   * @param options.style Style of the phone number. Defaults to 'human'
```
ST-DDT:
```suggestion
   * @param options.style Style of the phone number. Defaults to `'human'`.
```

src/modules/phone/index.ts
ST-DDT: How about `full` or `normalized` or something in that direction?

src/definitions/phone_number.ts
```diff
@@ -12,5 +12,9 @@ export type PhoneNumberDefinition = LocaleEntry<{
    *
    * @see faker.helpers.replaceSymbolWithNumber(format): For more information about how the patterns are used.
    */
-  formats: string[];
+  format: {
+    human: string[];
+    national: string[];
+    raw: string[];
```
ST-DDT: Could you please add a little documentation for each of these, so that their purpose is clear when defining the data?

matthewmayer: added

src/modules/phone/index.ts
matthewmayer: renamed to international per maintainers on Discord.

src/modules/phone/index.ts
```diff
@@ -12,18 +12,42 @@ export class PhoneModule extends ModuleBase {
...
+       * - `'national'`: A phone number in a standardized national format, e.g. `(555) 123-4567`.
+       * - `'international'`: A phone number in the E.123 international format, e.g. `+15551234567`
+       *
+       * @default 'human'
+       */
+      style?: 'human' | 'national' | 'international';
+    } = {}
+  ): string {
+    const { style = 'human' } = options;
+    const formats = this.faker.definitions.phone_number.format;
```
ST-DDT, line 43: Formats is only used once.
Maybe inline it.

ST-DDT (approved):

ST-DDT:
The description states it is still WIP. Is that outdated or what is left to do?

matthewmayer:
Nope sorry not WIP any more. Ready for final review.

romain-sen:
I am sorry but I am lost, using parameters with faker.phone.number() with parameters is deprecated. So why this PR seems to add parameters ? Is the deprecation canceled or something ?

ST-DDT:
The new parameters are entirely different and incompatible with the previous ones.

romain-sen:
In which version will it be available ?

ST-DDT:
> In which version will it be available ?

v9.0.0 (+preview releases)

We plan to release an alpha soon. Maybe next week?

PR #2581, opened 2023-12-17
fix(number): improve float generation for precisions of form 10^-n

### LINKED ISSUE #331: datatype.number does not always generate numbers with expected precision

### LINKED ISSUE #2567: Generating a float with precision of 0.00001 gives larger precision number
### Pre-Checks

### Describe the bug

When trying to generate a float using **faker.number.float({ precision: 0.00001 })**, the generated float generates a number with more than 5 decimal place values after the decimal point. Generated float values where precision is set to 0.1 to 0.0001, and 0.000001 and greater are consistently working. 

### Minimal reproduction code

https://runkit.com/embed/g86r0jyau1rk

### Additional Context

fake

Using faker-js to write playwright tests; example shows a generated value logged out where it was set to **faker.number.float({ max: 99999.99999, precision: 0.00001 })**.

### Environment Info

```shell
System:
    OS: Linux 5.10 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-1280P
    Memory: 6.67 GB / 15.46 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 20.9.0 - /usr/local/bin/node
    Yarn: 3.6.3 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
  npmPackages:
    @faker-js/faker: ^8.0.2 => 8.0.2
```

- [X] CJS
- [ ] ESM

### Used Package Manager

npm

matthewmayer:
fix #2567

Changes the way that the 1/precision is calculated in faker.number.float() using Math.log10 and Math.pow

This ensures that if precision is 0.00001, you get exactly 100000 for example, not 99999.99999999999


matthewmayer:
Note this obviously conflicts with #2564, probably easiest to merge that first since it touches more code

src/modules/number/index.ts
```diff
@@ -150,7 +150,9 @@ export class NumberModule extends SimpleModuleBase {
         throw new FakerError(`Precision should be greater than 0.`);
       }
 
-      const factor = 1 / precision;
+      const factor = Number.isInteger(Math.log10(precision))
+        ? Math.pow(10, -Math.log10(precision)) //mathematically this is the same as 1/precision, however it helpfully gives integer values for all precisions of the form 10^-n
```
ST-DDT: Maybe cache the log10 result?

test/modules/number.spec.ts
```diff
@@ -270,6 +286,17 @@ describe('number', () => {
...
       });
 
+      it('provides numbers with an exact precision of 0.00001', () => {
+        for (let i = 0; i < 100; i++) {
+          const actual = faker.number.float({
+            min: 0.5,
+            max: 0.99,
+            precision: 0.00001,
+          });
+          expect(actual).toBe(Number(actual.toFixed(5)));
```
ST-DDT: Maybe do this test for all values `10**x where 10 > x > -10 and x is int` instead of just testing some magic value.

ST-DDT:
And thanks for tackling this ❤️

test/modules/number.spec.ts
matthewmayer: added it for all negative x

test/modules/number.spec.ts
```diff
@@ -275,25 +275,16 @@ describe('number', () => {
...
-      it('provides numbers with an exact precision of 0.00001', () => {
-        for (let i = 0; i < 100; i++) {
-          const actual = faker.number.float({
-            min: 0.5,
-            max: 0.99,
-            precision: 0.00001,
-          });
-          expect(actual).toBe(Number(actual.toFixed(5)));
+      it('provides numbers with an exact precision of 10^-x for x=1 to 18', () => {
+        for (let factor = 1; factor <= 18; factor++) {
```
ST-DDT: Please use it.each for the outer loop and also check factors with negative and zero values.
(They might need a different check, so we could potentially put them in a separate it block.)

Lastly I'm not sure about calling this factor IMO this should be called exponent (and have its value negated) or digits (to match toFixed).

matthewmayer: im not sure checking precisions like 1, 10, 100 is very useful, after all if you are doing

faker.number.float({min:0, max:1000, precision:100}) would you not be better served by using faker.number.int() ?

ST-DDT: Faker.number.int doesnt have a precision parameter.
I'll think about it some more.

matthewmayer: `faker.number.float({min:0, max:1000, precision:100})` is just `100 * faker.number.int({min:0, max:10})`

ST-DDT: So is `faker.number.float({min:0, max:1, precision:0.1})` just `0.1 * faker.number.int({min:0, max:10})`

matthewmayer: Basically yes. The problem is that when you multiply by numbers like 0.00001 in JS you get these weird floating point errors which we are trying to avoid.

matthewmayer: We could add multipleOf for int too for symmetry

ST-DDT: I also thought about that and we should do that, although in a different PR.

src/modules/number/index.ts
```diff
@@ -150,7 +150,10 @@ export class NumberModule extends SimpleModuleBase {
         throw new FakerError(`Precision should be greater than 0.`);
       }
 
-      const factor = 1 / precision;
+      const logPrecision = Math.log10(precision);
+      const factor = Number.isInteger(logPrecision)
+        ? Math.pow(10, -logPrecision) //mathematically this is the same as 1/precision, however it helpfully gives integer values for all precisions of the form 10^-n
```
ST-DDT: Please move the comment to the front, maybe even another line up as it refers to the entire block. 

```suggestion
      // Workaround to get integer values for the reverse of all precisions of the form 10^-n
      const factor = Number.isInteger(logPrecision)
        ? (10 ** -logPrecision)
```

Also I prefer `**` over Math.pow (We should probably enable [prefer-exponentiation-operator](https://eslint.org/docs/latest/rules/prefer-exponentiation-operator) too)
Not sure whether the parenthesis are needed here.

ST-DDT (commented):
Sorry, I didn't have enough free time to look at the precision behavior yet.

test/modules/number.spec.ts
```diff
@@ -259,17 +259,37 @@ describe('number', () => {
...
+                precision: 0.2,
+              })
+            )
+          ),
+        ].sort();
+
+        expect(results).toEqual([0, 0.2, 0.4]);
       });
 
+      const exponents = Array.from({ length: 18 }, (_, i) => i + 1);
```
ST-DDT: Please use our `times(x)` util for that.

https://github.com/faker-js/faker/blob/8996a7430b9bdc3a86cc2d972c4f743a301cc5c6/test/support/times.ts#L8-L10

test/modules/number.spec.ts
```diff
@@ -259,17 +259,37 @@ describe('number', () => {
...
+      it.each(exponents)(
+        `provides numbers with an exact precision of 10^-%d`,
+        (exponent) => {
+          for (let i = 0; i < 100; i++) {
+            const actual = faker.number.float({
+              min: 0.5,
+              max: 0.99,
+              precision: Math.pow(10, -exponent),
```
ST-DDT: Please also use `**` here.

src/modules/number/index.ts
ST-DDT: I created #2585 to enable that rule.

- #2585

ST-DDT (approved):

ST-DDT:
Thanks for addressing all my comments 👍

PR #2601, opened 2024-01-14
refactor(internet): deprecated redundant avatar method

### LINKED ISSUE #2599: Deprecate or replace internet.avatar  in favor of image.avatar
We already have [`faker.image.avatar`](https://fakerjs.dev/api/image.html#avatar), so IMO there is no need to have a separate implementation in the internet module.
So we should either deprecate the `faker.internet.avatar` method for removal, or make it an alias of `faker.image.avatar`.
It definitely should jsdoc link to the other method.

`faker.internet.avatar`:

https://github.com/faker-js/faker/blob/1328985f36b9e6670aff7704cb1914fd7a4514d4/src/modules/internet/index.ts#L41-L54

`faker.image.avatar(Legacy)`:

https://github.com/faker-js/faker/blob/1328985f36b9e6670aff7704cb1914fd7a4514d4/src/modules/image/index.ts#L85-L99

Found while investigating: https://github.com/faker-js/faker/issues/2598

- https://github.com/faker-js/faker/issues/2598

ST-DDT:
Fixes #2599

- #2599


src/modules/internet/index.ts
```diff
@@ -41,16 +41,26 @@ export class InternetModule extends ModuleBase {
...
-    return `https://cloudflare-ipfs.com/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/${this.faker.number.int(
-      1249
-    )}.jpg`;
+    deprecated({
+      deprecated: 'faker.internet.avatar()',
+      proposed: 'faker.image.avatarLegacy() or faker.image.avatar()',
+      since: '8.4',
+      until: '9.0',
+    });
+    // TODO @ST-DDT 2024-01-14: Replace with `faker.image.avatar()` in v9
```
```suggestion
    // TODO @ST-DDT 2024-01-14: Remove or replace with `faker.image.avatar()` in v9
```

matthewmayer (approved):

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox (approved):
If we deprecated this method now, we can't decide on replacing the implementation in the v9, as you comment suggests.

PR #2606, opened 2024-01-18
chore(release): 8.4.0

xDivisionByZerox:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [8.4.0](https://github.com/faker-js/faker/compare/v8.3.1...v8.4.0) (2024-01-24)
+
+
+### Features
+
+* **helpers:** add support for complex intermediate types ([#2550](https://github.com/faker-js/faker/issues/2550)) ([24482a3](https://github.com/faker-js/faker/commit/24482a30042eec5b553b30d60985e89fd69a8660))
+* **number:** add parameter `fractionDigits` in float ([#1855](https://github.com/faker-js/faker/issues/1855)) ([41d8778](https://github.com/faker-js/faker/commit/41d87789c7ff353acfd0f5ca88a99c0d1fd3b500))
+* **person:** add job titles for fr ([#2531](https://github.com/faker-js/faker/issues/2531)) ([ba28ab6](https://github.com/faker-js/faker/commit/ba28ab6e59009cceb5b691ddd212e09df22b586e))
+
+
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


ST-DDT:
Should we wait for https://github.com/faker-js/faker/pull/2607 ?
Should we also regenerate the pnpm-lock file as usual?

matthewmayer:
if you think you'll likely do a 8.4.1 no need to wait. If you plan for 8.4.0 to be final pre-9.0.0 release (barring any critical bugfixes) it would be nice to wait.

Shinigami92:
IMO it is fine to wait, so that the docs are directly fine and don't need an extra release

Shinigami92:
https://github.com/faker-js/faker/blob/next/src/modules/finance/index.ts#L379-L612 does not have a `@deprecation` :warning: 
as well as https://github.com/faker-js/faker/blob/next/src/modules/finance/index.ts#L990-L1167
and https://github.com/faker-js/faker/blob/next/src/modules/location/index.ts#L430-L735

In location.longitude the defaults are wrongly documented, (90 instead of 180)

xDivisionByZerox:
> [`next`/src/modules/finance/index.ts#L379-L612](https://github.com/faker-js/faker/blob/next/src/modules/finance/index.ts?rgh-link-date=2024-01-19T18%3A55%3A05Z#L379-L612) does not have a `@deprecation` ⚠️ as well as [`next`/src/modules/finance/index.ts#L990-L1167](https://github.com/faker-js/faker/blob/next/src/modules/finance/index.ts?rgh-link-date=2024-01-19T18%3A55%3A05Z#L990-L1167) and [`next`/src/modules/location/index.ts#L430-L735](https://github.com/faker-js/faker/blob/next/src/modules/location/index.ts?rgh-link-date=2024-01-19T18%3A55%3A05Z#L430-L735)
> 
> In location.longitude the defaults are wrongly documented, (90 instead of 180)

This is what I would put under "if we find anything during the next week", as we discussed in yesterdays meeting. I'd like to release v8.4.0, even if not perfect. Since you already found some unpollished code, we need to release a 8.5.0 as well afterwards.

Opinions on this?

ST-DDT (dismissed):
No pnpm-lock update?

xDivisionByZerox:
> No pnpm-lock update?

I never really understood why we do this in the first place.

The lock file ensures that all dependencies work as expected, if installed under the exact conditions described in the file. Why would I ignore that and update it ON RELEASE to potentially introduce a dependency tree that might be broken?   
Like it partially happened in [the release PR for 8.2.0](https://github.com/faker-js/faker/pull/2478#issuecomment-1763125277). While not breaking the dependency tree, the rebuild introduced parts in the file that resolve in the project not being able to build. You are required to **manually** remove these part. That in itself feels really shady to me.

The lock file should only be updated by renovate or when dependencies are updated manually IMO.

ejcheng (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #2607, opened 2024-01-19
docs: improve docs for faker.number.float

matthewmayer:
cleanup of docs after #1855 and #2564

Make it clearer that precision is a deprecated version of multipleOf, and that you should only pass one of {precision, multipleOf, fractionDigits}. Fix an example. Move information about inclusive/exclusive to the min/max parameters.

### DIFF
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -120,7 +120,7 @@ export class NumberModule extends SimpleModuleBase {
       | number
       | {
           /**
-           * Lower bound for generated number, inclusive. Defaults to `0.0`.
+           * Lower bound for generated number, inclusive.
            *
            * @default 0.0
            */
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -125,7 +125,7 @@ export class NumberModule extends SimpleModuleBase {
            */
           min?: number;
           /**
-           * exclusive, unless `multipleOf`, `precision` or `fractionDigits` are passed. Defaults to `1.0`.
+           * Upper bound for generated number, exclusive, unless `multipleOf`, `precision` or `fractionDigits` are passed. Defaults to `1.0`.
            *
            * @default 1.0
            */
--- a/src/modules/number/index.ts
+++ b/src/modules/number/index.ts
@@ -86,21 +86,14 @@ export class NumberModule extends SimpleModuleBase {
   }
 
   /**
-   * Returns a single random floating-point number.
-   * The lower bound is inclusive, the upper bound is exclusive, unless `multipleOf` is passed.
+   * Returns a single random floating-point number, by default between `0.0` and `1.0`. To change the range, pass a `min` and `max` value. To limit the number of decimal places, pass a `multipleOf` or `fractionDigits` parameter.
    *
    * @param options Upper bound or options object.
-   * @param options.min Lower bound for generated number. Defaults to `0.0`.
-   * @param options.max Upper bound for generated number. Defaults to `1.0`.
-   * @param options.precision Precision of the generated number, for example `0.01` will round to 2 decimal points.
-   * If precision is passed, the upper bound is inclusive.
-   * @param options.multipleOf The generated number will be a multiple of this property.
-   * This property can be used to limit the result to a specific number of decimal digits.
-   * For example `0.01` will round to 2 decimal points.
-   * If `multipleOf` is passed, the upper bound is inclusive.
-   * This option is incompatible with the `fractionDigits` option.
-   * @param options.fractionDigits The maximum number of digits to appear after the decimal point.
-   * This option is incompatible with the `multipleOf` option.
+   * @param options.min Lower bound for generated number, inclusive. Defaults to `0.0`.
+   * @param options.max Upper bound for generated number, exclusive, unless `multipleOf`, `precision` or `fractionDigits` are passed. Defaults to `1.0`.
+   * @param options.precision Deprecated alias for `multipleOf`. Only one of `multipleOf`, `precision` or `fractionDigits` should be passed.
+   * @param options.multipleOf The generated number will be a multiple of this parameter. Only one of `multipleOf`, `precision` or `fractionDigits` should be passed.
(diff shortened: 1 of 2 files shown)


src/modules/number/index.ts
```diff
@@ -125,30 +120,29 @@ export class NumberModule extends SimpleModuleBase {
       | number
       | {
           /**
-           * Lower bound for generated number.
+           * Lower bound for generated number, inclusive. Defaults to `0.0`.
```
ST-DDT: The default is already below.

```suggestion
           * Lower bound for generated number, inclusive.
```

src/modules/number/index.ts
```diff
@@ -125,30 +120,29 @@ export class NumberModule extends SimpleModuleBase {
...
            *
            * @default 0.0
            */
           min?: number;
           /**
-           * Upper bound for generated number.
+           * Upper bound for generated number, exclusive, unless `multipleOf`, `precision` or `fractionDigits` are passed. Defaults to `1.0`.
```
```suggestion
           * Upper bound for generated number, exclusive, unless `multipleOf`, `precision` or `fractionDigits` are passed.
```

src/modules/number/index.ts
```diff
@@ -111,12 +104,14 @@ export class NumberModule extends SimpleModuleBase {
    * @example
    * faker.number.float() // 0.5688541042618454
    * faker.number.float(3) // 2.367973240558058
-   * faker.number.float({ min: -1000000 }) //-780678.849672846
+   * faker.number.float({ min: -1000000, max: 0 }) //-780678.849672846
```
ST-DDT: Maybe merge this and the following example into one, as we have quite a lot examples here.

matthewmayer: it could perhaps just be removed. i thought the original example was a bit confusing because `faker.number.float({ min: -1000000 })` actually gives you between -1000000 and 1, which is probably not what you want.

ST-DDT (approved):

Shinigami92 (approved):

xDivisionByZerox:
General question: Do we want to write JSDocs in one line? In our Markdown files we generally break after certain keypoints (end of sentence, punctuation, reaching character line limit). I was having trouble reading these long description lines on my phone. Might not be that hard to read on a bigger screen, but I generally prefer breaking the line, at least after the end of a sentence. I'm I the only one? What's your take on this?

Shinigami92:
> General question: Do we want to write JSDocs in one line? In our Markdown files we generally break after certain keypoints (end of sentence, punctuation, reaching character line limit). I was having trouble reading these long description lines on my phone. Might not be that hard to read on a bigger screen, but I generally prefer breaking the line, at least after the end of a sentence. I'm I the only one? What's your take on this?

Thanks for raising this, now I feel confirmed as I encountered this as well but did not say anything

matthewmayer:
Ideally we could have a lint or prettier rule to do this automatically as I'm usually on a desktop with word wrap enabled when editing, so it's not something I typically think about.

Shinigami92:
> Ideally we could have a lint or prettier rule to do this automatically as I'm usually on a desktop with word wrap enabled when editing, so it's not something I typically think about.

Either eslint-plugin-jsdoc has something for this already and needs to be configured, or as I suggest: do not use the IDE Editor Panel but hover over the function and investigate how it is rendered in the IDE tooltip.

matthewmayer:
> > Ideally we could have a lint or prettier rule to do this automatically as I'm usually on a desktop with word wrap enabled when editing, so it's not something I typically think about.
> 
> Either eslint-plugin-jsdoc has something for this already and needs to be configured, or as I suggest: do not use the IDE Editor Panel but hover over the function and investigate how it is rendered in the IDE tooltip.

single newlines in Markdown have no effect on how the text is rendered. So this is only important when viewing the raw markdown.

Shinigami92:
> > > Ideally we could have a lint or prettier rule to do this automatically as I'm usually on a desktop with word wrap enabled when editing, so it's not something I typically think about.
> > 
> > 
> > Either eslint-plugin-jsdoc has something for this already and needs to be configured, or as I suggest: do not use the IDE Editor Panel but hover over the function and investigate how it is rendered in the IDE tooltip.
> 
> single newlines in Markdown have no effect on how the text is rendered. So this is only important when viewing the raw markdown.

No, what we mean is the preference of:

```
Returns a single random floating-point number, by default between `0.0` and `1.0`.

To change the range, pass a `min` and `max` value. To limit the number of decimal places, pass a `multipleOf` or `fractionDigits` parameter.
```

instead of

```
Returns a single random floating-point number, by default between `0.0` and `1.0`. To change the range, pass a `min` and `max` value. To limit the number of decimal places, pass a `multipleOf` or `fractionDigits` parameter.
```

Please recognize the blank line between the sentences, because this results in a real line break in the rendered tooltip.

matthewmayer:
> > > > Ideally we could have a lint or prettier rule to do this automatically as I'm usually on a desktop with word wrap enabled when editing, so it's not something I typically think about.
> > > 
> > > 
> > > Either eslint-plugin-jsdoc has something for this already and needs to be configured, or as I suggest: do not use the IDE Editor Panel but hover over the function and investigate how it is rendered in the IDE tooltip.
> > 
> > 
> > single newlines in Markdown have no effect on how the text is rendered. So this is only important when viewing the raw markdown.
> 
> No, what we mean is the preference of:
> 
> ```
> Returns a single random floating-point number, by default between `0.0` and `1.0`.
> 
> To change the range, pass a `min` and `max` value. To limit the number of decimal places, pass a `multipleOf` or `fractionDigits` parameter.
> ```
> 
> instead of
> 
> ```
> Returns a single random floating-point number, by default between `0.0` and `1.0`. To change the range, pass a `min` and `max` value. To limit the number of decimal places, pass a `multipleOf` or `fractionDigits` parameter.
> ```
> 
> Please recognize the blank line between the sentences, because this results in a real line break in the rendered tooltip.

its not terribly consistent at the moment. Most descriptions of methods are very terse. Where there are multiple sentences, sometimes linebreaks are used and sometimes they are not

https://fakerjs.dev/api/internet.html#username
https://fakerjs.dev/api/finance.html#ethereumaddress

Shinigami92:
You can decide, I approved already :shrug:

xDivisionByZerox (approved):
I'll approve as well. We should have some standards/rules to follow regarding comment format. Since we don't have that, I'll not block this PR with my question.

PR #2620, opened 2024-01-24
feat(music): add album and artist methods

jeremyhofer:
This PR introduces two new methods to the `music` module. `albumName` and `artist`. 

Data was sourced from various resources from a number of resources online, including:
- The dataset from when the `songName` method was added in https://github.com/faker-js/faker/pull/996: https://manghammath.com/80s%20Hits/The%20Top%201000%20Songs%20of%20AllTime.pdf
- Rolling Stone 500 Greatest Albums: https://www.rollingstone.com/music/music-lists/best-albums-of-all-time-1062063/arcade-fire-%ef%bb%bffuneral-1062733/
- Chart Masters compiled data for Artists and Albums: https://chartmasters.org/

The `@since` version on the methods was set to `8.5.0`, as it looks like the `8.4.0` tag has been created.


matthewmayer (changes requested):
1. We normally limit to 1000 entries per file
2. Suggest splitting into two PRs, one for adding data for existing methods (since that can be done in a patch release) and one for proposing new methods
3. Some of the genres look a bit awkward like "Holiday: Other". Maybe remove ones like that.

jeremyhofer:
@matthewmayer thanks for the feedback!

I pulled the genre and song name additions out of this PR. Will raise a PR to add and clean up the additional genres. Given the song name set is near 1000 entries I'm discarding all my changes there.

I updated the added `album_name` and `artist` sets down to 1000 entries each.

src/modules/music/index.ts
```diff
@@ -5,9 +5,45 @@ import { ModuleBase } from '../../internal/module-base';
  *
  * ### Overview
  *
- * Generate a random music genre with [`genre()`](https://fakerjs.dev/api/music.html#genre) or song name with [`songName()`](https://fakerjs.dev/api/music.html#songname). Both may be localized.
+ * Generate random music content.
+ *
+ * For a random album name, use [`albumName()`](https://fakerjs.dev/api/music.html#albumname).
+ *
+ * For a ranom artist, use [`artist()`](https://fakerjs.dev/api/music.html#artist).
```
matthewmayer: typo, random

matthewmayer (dismissed):

matthewmayer:
Non ascii names are OK as long as that's how they are normally written in English

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
+export default [
+  '"Awaken, My Love!"',
+  "(What's The Story) Morning Glory?",
+  '- Tragedy +',
+  '.mp3',
```
ST-DDT: This might cause confusion and is IMO not worth the confusion.

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  '13 Reasons Why (Season 3)',
+  '2000',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  '21st Century Breakdown',
+  '30 De Febrero',
+  '42',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  '432 Hz Deep Healing',
+  '5-Star',
+  '5.0',
+  '528 Hz Meditation Music',
+  '54+1',
+  '8 Mile',
+  '808s & Heartbreak',
+  '9 To 5 And Odd Jobs',
+  '9',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  '99.9%',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  '?',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  'Non Stop Erotic Cabaret',
+  'Non-Fiction',
+  'Northsbest',
+  'Nostalgia',
+  'Nostalgia, Ultra',
+  'Notes On A Conditional Form',
+  'Now Or Never',
+  'O Embaixador (Ao Vivo)',
+  'O My Heart',
+  'O',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  'Wilder Mind',
+  'Wildfire',
+  'Willy And The Poor Boys',
+  'Wings / You Never Walk Alone',
+  'Wish You Were Here',
+  'Wish',
+  'Without Warning',
+  'Wonder',
+  'X&Y',
+  'X',
```
ST-DDT: Same here

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
...
+  'XOXO',
+  'Y Que Quede Claro',
+  'YBN: The Mixtape',
+  'Yo Creo',
+  'You Will Regret',
+  'You',
+  'Youngblood',
+  'Younger Now',
+  'Youth',
+  'xx',
```
ST-DDT: I think we should use case insensitive sort

src/locales/en/music/album_name.ts
```diff
@@ -0,0 +1,947 @@
+export default [
+  '"Awaken, My Love!"',
+  "(What's The Story) Morning Glory?",
```
ST-DDT: Does this refer to the thing I think it does?

src/locales/en/music/artist.ts
```diff
@@ -0,0 +1,988 @@
...
+  'Chicago',
+  'Chris Brown',
+  'Chris Isaak',
+  'Chris Young',
+  'Christina Aguilera',
+  'Christina Perri',
+  'Christopher Cross',
+  'Chuck Berry',
+  'Ciara',
+  'Cigarettes After Sex',
```
ST-DDT, line 169: Should we remove this in case of value transformation by the user? (aka `value.split(' ')[2]`)
I think we have enough other values to omit this one.

src/locales/en/music/album_name.ts
matthewmayer: It's a double entendre for sure, but it's also a pretty famous album. I don't think it's offensive.

matthewmayer:
FWIW although we would normally wait for 10 upvotes to approve a change like this this does seem a fairly sensible extension of the current music module so I'd be fine with approving this for next minor or major release.

src/locales/en/music/artist.ts
jeremyhofer: I'm not sure about the rationale of removing due to a value transformation. However, I agree there are more than enough examples and have no issue removing this one.

jeremyhofer:
Updated to remove albums that could be confusing - the ones that were called out, as well as others I felt could similarly cause some confusion. Updated to case insensitive sort the data as well.

Thanks for considering these methods! Would it be possible to add the ask to propose and vote on new modules and methods to the docs and CONTRIBUTING.md? I was unaware this was the general expectation when I got started in the repo :smiley:

matthewmayer:
> Would it be possible to add the ask to propose and vote on new modules and methods to the docs and CONTRIBUTING.md? I was unaware this was the general expectation when I got started in the repo 😃

Its not really a hard requirement, its more to help triage the issues that are added to the project. So if someone adds an issue saying "it would be cool if we had artist names" we will normally wait until >10 upvotes to see if other people would find it useful. If someone sources suitable data, makes a PR, adds tests etc, then the threshold for inclusion is lower!

So I think being too strict about it and documenting in CONTRIBUTING might discourage people from actually contributing code, which we wouldnt want.

src/locales/en/music/artist.ts
xDivisionByZerox: > Should we remove this in case of value transformation by the user? (aka value.split(' ')[2])

I would not consider this as custom value transformation doesn't make any sense in this context.

> I think we have enough other values to omit this one.

Fair enought. Although, with over 5 Million sold albums world wide they are far more well known than the german rapper CRO (line 182) for example. He has not even 2 Million sold albums.  
Excluding them only because they have an undired word in their name feels kinda wrong to me. While I agree on the fact that we should have somewhat of an profanity filter, I'd not agree on it being applied to names. In the worst case scenario this could potential hurt the overall diversity of values generated, as giving yourself an "outragous" name (as an artist) might a percise decision you make to guid attention to specific topics.

src/modules/music/index.ts
```diff
@@ -5,9 +5,45 @@ import { ModuleBase } from '../../internal/module-base';
...
+ * All data types may be localized.
  */
 export class MusicModule extends ModuleBase {
+  /**
+   * Returns a random album name.
+   *
+   * @example
+   * faker.music.albumName() // '1989'
+   *
+   * @since 8.5.0
```
xDivisionByZerox: We will not be releasing a version 8.5 as we already work on bringing the current codebase to v9.0. So these methods will (most likely) land in v9.0 as well.

```suggestion
   * @since 9.0.0
```

src/modules/music/index.ts
```diff
@@ -5,9 +5,45 @@ import { ModuleBase } from '../../internal/module-base';
...
+    );
+  }
+
+  /**
+   * Returns a random artist name.
+   *
+   * @example
+   * faker.music.artist() // 'The Beatles'
+   *
+   * @since 8.5.0
```
xDivisionByZerox: Same here.

```suggestion
   * @since 9.0.0
```

xDivisionByZerox (changes requested):
Some small updates to the documentation. Otherwise this looks good to me.

ST-DDT:
We talked about this PR in today's team meeting.
There we discussed whether the methods (and definitions) should use the `Name` suffix.
In our oppinion `album` and `albumName` convey the same information and thus we should only use `album()`.

The definitions should follow the method's name unless there is a need for disambiaguation such as with `_pattern`.
(Since we don't plan to add patterns for these methods we should use the method name for the definitions)

We will rename `songName` -> `song` in a separate PR.

@matthewmayer What do you think?

matthewmayer:
Sure. Shorter is better like city instead of cityName

ST-DDT:
@jeremyhofer Could you please apply the requested changes and fix the merge conflicts?

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):

src/modules/music/index.ts
```diff
@@ -5,9 +5,45 @@ import { ModuleBase } from '../../internal/module-base';
  *
  * ### Overview
  *
- * Generate a random music genre with [`genre()`](https://fakerjs.dev/api/music.html#genre) or song name with [`songName()`](https://fakerjs.dev/api/music.html#songname). Both may be localized.
+ * Generate random music content.
+ *
+ * For a random album name, use [`albumName()`](https://fakerjs.dev/api/music.html#album).
```
ST-DDT:
```suggestion
 * For a random album name, use [`album()`](https://fakerjs.dev/api/music.html#album).
```

src/locales/en/music/album_name.ts
ST-DDT, line 1: The file should probably be renamed to just `album`.
same for song name.

src/definitions/music.ts
```diff
@@ -4,6 +4,16 @@ import type { LocaleEntry } from './definitions';
  * The possible definitions related to music.
  */
 export type MusicDefinition = LocaleEntry<{
+  /**
+   * The names of some albums.
+   */
+  album_name: string[];
```
ST-DDT: This should be renamed to just album.
Same for song name.

src/modules/music/index.ts
```diff
@@ -5,9 +5,45 @@ import { ModuleBase } from '../../internal/module-base';
...
+ *
+ * For a random artist, use [`artist()`](https://fakerjs.dev/api/music.html#artist).
+ *
+ * For a random genre, use [`genre()`](https://fakerjs.dev/api/music.html#genre).
+ *
+ * For a random song name, [`songName()`](https://fakerjs.dev/api/music.html#song).
```
ST-DDT:
```suggestion
 * For a random song name, [`song()`](https://fakerjs.dev/api/music.html#song).
```

src/modules/music/index.ts
```diff
@@ -24,11 +60,11 @@ export class MusicModule extends ModuleBase {
    * Returns a random song name.
    *
    * @example
-   * faker.music.songName() // 'White Christmas'
+   * faker.music.song() // 'White Christmas'
    *
    * @since 7.1.0
    */
-  songName(): string {
+  song(): string {
```
ST-DDT: This is a breaking change

ST-DDT (changes requested):

src/modules/music/index.ts
xDivisionByZerox: Yeah, I'll revert this. I overred the "in a separate PR" thing in [the original change request comment](https://github.com/faker-js/faker/pull/2620#issuecomment-2050181213).

ST-DDT (dismissed):

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

ST-DDT (approved):

matthewmayer (approved):

Shinigami92 (approved):

PR #2621, opened 2024-01-25
feat(music): add additional genres

jeremyhofer:
This PR is split off from #2620 where I originally added additional genres along with new `album` and `artistName` methods.

Genres were compiled primarily from these resources:
- https://www.musicgenreslist.com/
- https://en.wikipedia.org/wiki/List_of_music_genres_and_styles
- https://koop.org/shows/genres-definitions/
- https://www.musicgateway.com/blog/spotify/what-are-the-different-genres-of-music

Based on feedback in #2620 from the original set I added I tried to clean up some of the odd/awkward genres adopted from the sources, e.g. just specifying "Holiday" or "Classical" instead of listing some of the sub-types from the resources. Happy to clean up the list further if desired!


src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,313 @@
 export default [
-  'Rock',
-  'Metal',
-  'Pop',
+  "Jungle/Drum’n'bass",
```
matthewmayer: Split these? I am reliably informed they are not the same genre :D

https://www.youtube.com/watch?v=0thYqrBSWK4

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,313 @@
...
+  'South Africa',
+  'South America',
+  'Southern Gospel',
+  'Southern Rock',
+  'Space Rock',
+  'Spanish',
+  'Spoken Word',
+  'Stage And Screen',
+  'Standards',
+  'Standup Comedy',
```
matthewmayer: not really a music genre?

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,313 @@
...
+  'Songwriter',
+  'Soul',
+  'Soundtrack',
+  'South Africa',
+  'South America',
+  'Southern Gospel',
+  'Southern Rock',
+  'Space Rock',
+  'Spanish',
+  'Spoken Word',
```
matthewmayer: not really a music genre?

matthewmayer (changes requested):

src/locales/en/music/genre.ts
matthewmayer: also insert at correct alphabetical position (regardless of the quoting style)

jeremyhofer: Split and properly sorted. Thanks for the input here, a lot of references I was finding were always linking these two names tightly together. Was torn on which route to go with it.

src/locales/en/music/genre.ts
jeremyhofer: Removed!

src/locales/en/music/genre.ts
jeremyhofer: Removed!

jeremyhofer:
In addition to the call outs about `Standup Comedy` and `Spoken Word`, I cleaned up other questionable genres that were in the list, trying to keep it pure to music only. This included removing items related to activities like `Study` or `Exercise`, since really any music can be used for that.

matthewmayer (dismissed):

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,293 @@
...
+  'Chicano',
+  'Children’s Music',
+  'Chill',
+  'Choral',
+  'Christian',
+  'Christmas',
+  'Classical',
+  'Club',
+  'College Rock',
+  'Comedy',
```
ST-DDT: This doesn't sound like music to me.

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,293 @@
...
+  'Crunk',
+  'Dance',
+  'Dancehall',
+  'Death Metal',
+  'Deep House',
+  'Delta Blues',
+  'Detroit Techno',
+  'Dirty South',
+  'Disco',
+  'Disney',
```
ST-DDT, line 66: Not sure about this one.

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,293 @@
...
+  'Dixieland',
+  'Doo-wop',
+  'Downtempo',
+  'Dream Pop',
+  'Drill',
+  'Drinking Songs',
+  'Drone',
+  "Drum’n'bass",
```
ST-DDT: Is the usage of two different apostrophes correct here `’` `'`?

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,293 @@
...
+  'Dub',
+  'Dubstep',
+  'EDM',
+  'Early Music',
+  'East Coast Rap',
+  'Easter',
+  'Easy Listening',
```
ST-DDT, line 80: Does that count as genre?

ST-DDT (commented):
I'm not sure whether we should include genres like `<any country> X`.
What do the others think?

src/locales/en/music/genre.ts
jeremyhofer: Agree, will drop this one.

src/locales/en/music/genre.ts
jeremyhofer: This one was fairly common in the genre lists I pulled up. I don't have any strong opinions on it. I'd be fine removing it.

src/locales/en/music/genre.ts
jeremyhofer: This is generally how it was styled in the resources I was looking through. Simply `Drum and Bass` is how it is listed on wikipedia. It also has some shorter forms listed like `D'n'B`. I am fine changing this up some.

https://en.wikipedia.org/wiki/Drum_and_bass

src/locales/en/music/genre.ts
jeremyhofer: This was another common genre listed on the resources I reviewed. I agree, it's a bit odd to me. I'd be fine removing it.

jeremyhofer:
@ST-DDT I had similar thoughts on the country names. There were some examples where e.g. `France` and `French` were both included from the sources I compiled from. In that case I opted for `French`. Could do similar for other countries listed.

src/locales/en/music/genre.ts
xDivisionByZerox: If you are going to change it I'm in favor of the apostrophe-free style `Drum and Bass`.

ST-DDT: Choose either:

```suggestion
  "Drum'n'bass",
```

```suggestion
  "Drum’n’bass",
```

```suggestion
  "Drum n bass",
```

matthewmayer:
I tend to agree that France or French is not really a genre. 

If a French is performing rock, it's Rock. If it's a classical piece, it's Classical, rather than just lumping it under French. 

So maybe let's remove all country/language genres.

src/locales/en/music/genre.ts
```diff
@@ -1,20 +1,271 @@
...
+  'Mainstream Jazz',
+  'Malay',
+  'Mandopop',
+  'March',
+  'Mariachi',
+  'Medieval',
+  'Meditation',
+  'Metal',
+  'Metalcore',
+  'Middle East',
```
ST-DDT: Does this count as country as well or would you consider this separate?

ST-DDT (dismissed):

src/locales/en/music/genre.ts
jeremyhofer: Good catch. I think this falls in the country/region bucket as well. Will update to remove.

ST-DDT (approved):

xDivisionByZerox:
Sorry for my late feedback. The list you provided is very comprehensive. Would it be better to organize the genres into "main" and "sub" categories? I can see people requesting the possibility of only getting "common" genres. For features like this we would need to introduce a breaking change. Is that something we should consider?

matthewmayer:
> Sorry for my late feedback. The list you provided is very comprehensive. Would it be better to organize the genres into "main" and "sub" categories? I can see people requesting the possibility of only getting "common" genres. For features like this we would need to introduce a breaking change. Is that something we should consider?

i'd say we dont really do that anywhere else. You can't request "common" first names or "common" dog breeds or "common" countries. There's no standard list of what a musical genre is.

xDivisionByZerox:
> i'd say we dont really do that anywhere else. You can't request "common" first names or "common" dog breeds or "common" countries. There's no standard list of what a musical genre is.

We have:

- [commonFileExt](https://fakerjs.dev/api/system.html#commonfileext)
- [commonFileName](https://fakerjs.dev/api/system.html#commonfilename)
- [commonFileType](https://fakerjs.dev/api/system.html#commonfiletype)

I see more reason to differenciate betrween common and sub music genres as some genres are literally described as "x is a subgenre of y". For example: 

- [Acid house (also simply known as just "acid") is a subgenre of house music [...]](https://en.wikipedia.org/wiki/Acid_house)
- [Death metal is an extreme subgenre of heavy metal music [...]](https://en.wikipedia.org/wiki/Death_metal)

---

I'm just trying to think ahead to avoid potential breaking chages. If all of you think this is unnecessary I'm fine with that as well.

Shinigami92 (approved):

xDivisionByZerox (approved):

ST-DDT:
**Team Decision**

- We will add this
- If we need a selection of common genres, then we can open a new issue

ejcheng (approved):

Issue #2432, opened 2023-10-05
The default value isn't always shown in the jsdocs


ST-DDT:
Ref: https://github.com/faker-js/faker/pull/2177#issuecomment-1745397481

The api docs don't show the default values for some methods (especially in the color module):

![grafik](https://github.com/faker-js/faker/assets/1579362/c97f38d6-a21c-4936-94d0-b2d91ef97308)

This is caused by the default being applied late in code (L376) and not in the signature (L370):

https://github.com/faker-js/faker/blob/fd6557e4b3716ebe41aa38a58b4ae112e6baf0d2/src/modules/color/index.ts#L365-L376

ST-DDT:
Blocked by #2177

- #2177

ST-DDT:
I'll have a look at this soon.

PR #2628, opened 2024-01-28
docs: rewrite api-docs generation using ts-morph

ST-DDT:
This PR replaces the typedoc based api-doc generation with one that uses the typescript AST.

This PR is work in progress, but if you have suggestions please let me know.


ST-DDT:
I found some broken JSDocs (see diff in this PR), that should be fixed in a separate PR before starting v9.0.
If anyone is interested in creating PRs for that please let  me know.
Otherwise I will do them when I have time.

ST-DDT:
Looks like I have plenty of time during my travel now, so I'll tackle the jsdocs now.

ST-DDT:
I completed reading ~~all~~/most data required to build the apidocs.
The following things are still missing though.

- the full signature in the examples (without jsdocs)
- the md -> html conversion prior to writing the page data
- option parameters

scripts/apidoc2/diff-index.ts
```diff
@@ -0,0 +1,79 @@
...
+  const diffIndex: ApiDiffHashs = Object.fromEntries(
+    pages.map((page) => [page.title, pageDiffHashs(page)])
+  );
+  writeFileSync(pathDocsDiffIndexFile, JSON.stringify(diffIndex));
+}
+
+function pageDiffHashs(page: RawApiDocsPage): ApiPageDiffHashs {
+  return {
+    pageHash: diffHash({
+      name: page.title,
```
ST-DDT: Todo rename to title

ST-DDT:
How should we display the signature of the method:

1. grafik

(no pre and suffixes)

or

2. grafik

(function prefix and semicolon suffix)

or

3. grafik

(faker.module prefix and no suffix)

or &lt;insert suggestion here&gt;?

---

And should we display the signature along with the examples or should we put them in a different code block?

Shinigami92:
> How should we display the signature of the method:
> 
> 
> 
> 1. grafik
> 
> 
> 
> (no pre and suffixes)
> 
> 
> 
> or
> 
> 
> 
> 2. grafik
> 
> 
> 
> (function prefix and semicolon suffix)
> 
> 
> 
> or
> 
> 
> 
> 3. grafik
> 
> 
> 
> (faker.module prefix and no suffix)
> 
> 
> 
> or &lt;insert suggestion here&gt;?
> 
> 
> 
> ---
> 
> 
> 
> And should we display the signature along with the examples or should we put them in a different code block?

2. because of correct syntax highlighting 

I'm also a fan of semicolon

matthewmayer:
I would put them in seperate code blocks. Perhaps they should also have subheaders?

## Parameters

## Examples
<code></code>

## Signature
<code></code>

## Source

Shinigami92:
I cant see the images somehow, but yeah. I like the idea of separating the examples from the signature

ST-DDT:
> I would put them in seperate code blocks. Perhaps they should also have subheaders?

Good idea. I will do all layout changes in a different PR (along the signature tabs), I'll prepare it in the data structures though. This PR is already large enough as is.

ST-DDT:
> I cant see the images somehow

There arent any images in matthewmayer's mockup to see in the first place.

matthewmayer:
> > I cant see the images somehow
> 
> There arent any images in matthewmayer's mockup to see in the first place.

ah those are just empty `<code>` blocks in my comment, not broken images :)

matthewmayer:
I must admit to knowing very little about how this works under the hood. What's the main advantage of this change? Faster? Less dependencies? More flexible on the output format?

ST-DDT:
> What's the main advantage of this change? Faster? Less dependencies? More flexible on the output format?

- Better access to the actual source values (specifically options jsdocs)
- provides access to the "raw" values instead of partially markdown-ified values (specifically the examples)
- It can be used for our v9 code refactoring for the standalone functions (the tooling not this code/PR).

ST-DDT:
The generate api docs are mostly the same, currently there is only a diff in the following files:

- [x] airline
- [x] date
- [x] finance
- [x] helpers
- [x] internet
- [x] lorem
- [x] random
- [x] string
- [x] system
- [x] word

Most of them should be very minor such as `{\n  max: number,\n  min: number\n}` vs `{ min: number; max: number; }`

ToCheck:

- [x] `@see` tags
- [x] are the methods sorted correctly e.g. helpers.replaceSymbols

ST-DDT:
- [x] Check LiteralUnion<'a' | 'b'> -> 'a' | 'b' | ? https://github.com/faker-js/faker/pull/1953#issuecomment-1666888020

ST-DDT:
TODOs:
- [x] Figure out what to do with that line: https://github.com/faker-js/faker/blob/db88a1518e8d0ccfe191bd48a1ae8641a9a25303/src/faker.ts#L59 because it actually throws in the tests now.
- [x] Fix SignatureTests tests

ST-DDT:
This is almost done. It is currently waiting for #2668 to be merged + a final review by myself + a short review guide for other reviewers.

- #2668

scripts/apidoc/output/page.ts
```diff
@@ -111,7 +111,8 @@ const defaultCommentRegex = /\s+Defaults to `([^`]+)`\..*/;
 
 function toMethodData(method: RawApiDocsMethod): ApiDocsMethod {
   const { name, signatures, sourcePath } = method;
-  const signatureData = signatures[signatures.length - 1];
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const signatureData = signatures.at(-1)!;
```
ST-DDT: 😿

ST-DDT:
Ready to review

---

Guide to review this PR:

- Have a look at the generated output (or try your own code example with `pnpm run docs:dev:api-docs`)
- Backup `docs/api` from `next` and `diff` it to the newly generated ones
- Diff `test/scripts/apidoc/__snapshots__/module.spec.ts.snap` vs `test/scripts/apidoc/__snapshots__/class.spec.ts.snap`
- Diff `test/scripts/apidoc/__snapshots__/signature.spec.ts.snap` vs `test/scripts/apidoc/__snapshots__/method.spec.ts.snap`
- Have a look at the code itself

Please mark all "debug" code you think should not be in the final merge.

scripts/apidoc/project.ts
```diff
@@ -0,0 +1,30 @@
+import type { ProjectOptions, SourceFile } from 'ts-morph';
+import { Project } from 'ts-morph';
+
+try {
+  await import('../../src'); // Make watch mode work on the source files
+} catch {
+  // Ignore
+}
```
ST-DDT: Remove or keep?

package.json
```diff
@@ -66,12 +66,13 @@
     "build": "run-s build:clean build:code build:types",
     "generate": "run-s generate:locales generate:api-docs",
     "generate:api-docs": "tsx ./scripts/apidoc.ts",
+    "generate:api-docs:watch": "tsx watch ./scripts/apidoc.ts",
```
ST-DDT: Remove or keep?

docs/.vitepress/components/api-docs/method.ts
```diff
@@ -1,19 +1,19 @@
-export interface Method {
+export interface ApiDocsMethod {
```
Shinigami92, line 1: Should we name this `ApiDocMethod` and `ApiDocMethodParameter`? Or is there something specific to the plural form?

scripts/apidoc/output/constants.ts
```diff
@@ -0,0 +1 @@
+export const scriptCommand = 'pnpm run generate:api-docs';
```
Shinigami92: I prefer to name such constants with `CONSTANT_CASE`

```suggestion
export const SCRIPT_COMMAND = 'pnpm run generate:api-docs';
```

scripts/apidoc/output/page.ts
```diff
@@ -0,0 +1,171 @@
...
+import type { ApiDocsMethod } from '../../../docs/.vitepress/components/api-docs/method';
+import type { RawApiDocsPage } from '../processing/class';
+import type { RawApiDocsMethod } from '../processing/method';
+import { formatMarkdown } from '../utils/format';
+import { adjustUrls, codeToHtml, mdToHtml } from '../utils/markdown';
+import { pathApiDocsDir } from '../utils/paths';
+import { required } from '../utils/value-checks';
+import { scriptCommand } from './constants';
+
+// Moved here because this must not be formatted by prettier
```
Shinigami92: Moved from what? Maybe this is a debug-comment and can be removed before merging? :thinking:

scripts/apidoc/processing/class.ts
```diff
@@ -0,0 +1,231 @@
...
+  DOC_INTERFACE_NAMES,
+  DOC_MODULE_FILTER,
+  DOC_UTILITY_NAMES,
+  shouldProcessType,
+} from './select';
+
+/**
+ * Represents a raw page in the API docs.
+ */
+export interface RawApiDocsPage {
```
Shinigami92: Here is the same, isn't it more a `RawApiDocPage`?

scripts/apidoc/processing/class.ts
```diff
@@ -0,0 +1,231 @@
...
+  result.methods.unshift(...processClassConstructors(clazz));
+  return result;
+}
+
+// Modules
+
+export function processModuleClasses(project: Project): RawApiDocsPage[] {
+  return processModules(
+    Object.values(getAllClasses(project, (v) => DOC_MODULE_FILTER(v))).sort(
+      (a, b) => a.getNameOrThrow().localeCompare(b.getNameOrThrow())
```
Shinigami92: This feels unsafe, why should it throw inside a sort?
Can we for example pre-process the names and then sort afterwards?

scripts/apidoc/processing/class.ts
```diff
@@ -0,0 +1,231 @@
...
+  return modules
+    .filter((m) => shouldProcessType(getModuleName(m)))
+    .map((m) => {
+      try {
+        return processModule(m, 'Modules');
+      } catch (error: unknown) {
+        throw newProcessingError({
+          type: 'module',
+          name: getModuleName(m),
+          source: m,
```
Shinigami92:
```suggestion
    .filter((module) => shouldProcessType(getModuleName(module)))
    .map((module) => {
      try {
        return processModule(module, 'Modules');
      } catch (error: unknown) {
        throw newProcessingError({
          type: 'module',
          name: getModuleName(module),
          source: module,
```

scripts/apidoc/processing/class.ts
```diff
@@ -0,0 +1,231 @@
...
+  );
+}
+
+export function processProjectClasses(project: Project): RawApiDocsPage[] {
+  return processClasses(valuesForKeys(getAllClasses(project), DOC_CLASS_NAMES));
+}
+
+function processClasses(classes: ClassDeclaration[]): RawApiDocsPage[] {
+  return classes
+    .filter((m) => shouldProcessType(m.getNameOrThrow()))
```
Shinigami92: Can we name some of these more specifically?
While reading through the code in GitHub Web UI, I struggle sometimes a bit if it is a module or a method.
I first need to jump with my eyes to e.g. the function that is used and then assume.

scripts/apidoc/utils/paths.ts
```diff
@@ -0,0 +1,24 @@
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const thisDir = dirname(fileURLToPath(import.meta.url));
```
Shinigami92: Should these all be CONSTANT_NAMES?

test/scripts/apidoc/method.example.ts
```diff
@@ -282,7 +346,12 @@ export class SignatureTest {
    *
    * @see test.apidoc.methodWithExample()
    *
+   * @since 1.0.0
+   *
    * @deprecated do something else
+   *
+   *
+   *
```
Shinigami92: Why these three lines?

test/scripts/apidoc/method.example.ts
```diff
@@ -74,6 +74,8 @@ export type AB = 'a' | 'b';
 export class SignatureTest {
   /**
    * Test with no parameters.
+   *
+   * @since 1.0.0
```
Shinigami92: Should all these changes in this file be in this PR? They feel like you added them here because they were forgotten to be added previously :thinking:

Shinigami92 (commented):
This was just a first look me reading through the GitHub Web UI checking more or less opinionated things and code readability without having an IDE.  
I will do a runtime check / checkout of code base in a later state.

scripts/apidoc/output/page.ts
ST-DDT: No that part cannot be formatted by prettier (without destroying) and thus is extracted as a separate variable.
IIRC it existed like that in the old code as well.

docs/.vitepress/components/api-docs/method.ts
ST-DDT: I tried to do it somewhat consistently the generate command is `pnpm run generate:api-docs` as well, but we could change that all to singular.

scripts/apidoc/processing/class.ts
ST-DDT: Not without calling `.getNameOrThrow()` in the process.

I could also use `required(a.getName(), 'name')`, but that wouldn't be any better.
This should never throw, because we don't use anonymous classes.

scripts/apidoc/processing/class.ts
ST-DDT: so `m` -> `class`?

test/scripts/apidoc/method.example.ts
ST-DDT:
```suggestion
```

test/scripts/apidoc/method.example.ts
ST-DDT: No, the old code doesn't require them, the new one does.
Previously the code just returned `MISSING`, but the new code throws an error.

scripts/apidoc/processing/class.ts
Shinigami92: This is a special case where `cls` or `clazz` might be needed

scripts/apidoc/output/page.ts
ST-DDT:
```suggestion
// Extracted to a constant because the contents must not be formatted by prettier
```

scripts/apidoc/diff.ts
```diff
@@ -1,20 +1,23 @@
-import type { DocsApiDiffIndex } from './utils';
-import { nameDocsDiffIndexFile, pathDocsDiffIndexFile } from './utils';
+import type { ApiDiffHashs } from './output/diff-index';
```
Shinigami92: ultra picky 😅
but isn't it `ApiDiffHashes`?

ST-DDT: Renamed

docs/.vitepress/components/api-docs/method.ts
ST-DDT: I moved this part to the meeting notes since I would like to have a functional review first.

Shinigami92 (commented):
- I did a visual `docs:dev` check, and that looks good to me
- The `docs:build` command takes 15 seconds compared to 18 seconds on next, so we saved 3 seconds somehow ^^
- Not sure if we need the `watch` commands, I did not use them
- Awaiting decision of naming
- Later we could enhance the dev logs by using e.g. https://github.com/poppinss/cliui
  With that it is possible to create progressive logs like in this example: https://github.com/poppinss/cliui#tasks

WikiRik:
Would this fix https://github.com/faker-js/faker/issues/2432 as well by any chance?

ST-DDT:
> Would this fix #2432 as well by any chance?

AFAICT: No, it doesn't because it is caused by a different set of issues.

Mainly:

````ts
function foobar(options?: {...}) {
  const { ... } = options ?? {};
}
// vs
function foobar(options: {...} = {}) {
  const { ... } = options;
}
````

For which method did you encounter that problem?

**Comment by ST-DDT** [comment]
> * Later we could enhance the dev logs by using e.g. https://github.com/poppinss/cliui
>       With that it is possible to create progressive logs like in this example: https://github.com/poppinss/cliui#tasks

Please keep in mind that we explicitly disabled that feature for other things such as vitest.

**Comment by Shinigami92** [comment]
> > * Later we could enhance the dev logs by using e.g. [poppinss/cliui](https://github.com/poppinss/cliui)
> >   With that it is possible to create progressive logs like in this example: [poppinss/cliui#tasks](https://github.com/poppinss/cliui#tasks)
> 
> Please keep in mind that we explicitly disabled that feature for other things such as vitest.

I'm not sure if this is related with each other 🤔 Also it would be just a pure DX faker internal thing.

File: package.json (hunk 4 above)
Comment (xDivisionByZerox) [reply package.json]: Remove. Never used it. Others don't seem to care that much.

Comment (ST-DDT) [reply package.json]: Done

File: scripts/apidoc/processing/select.ts [hunk 16]
```diff
@@ -0,0 +1,5 @@
+export const DOC_CLASS_NAMES = ['Faker', 'SimpleFaker'];
+export const DOC_INTERFACE_NAMES = ['Randomizer'];
+export const DOC_UTILITY_NAMES = ['mergeLocales'];
+export const DOC_MODULE_FILTER = (module: string): boolean =>
+  module.endsWith('Module') && !module.startsWith('Simple');
```
xDivisionByZerox: All of these are only used in `processing/class.ts`. What was your reasoning to put them into a separate file?

scripts/apidoc/project.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Project } from 'ts-morph';
+
+export function getProject(options: Partial<ProjectOptions> = {}): Project {
+  return new Project({
+    ...options,
+    tsConfigFilePath: options.tsConfigFilePath ?? 'tsconfig.build.json',
+  });
+}
+
+export function getAll<T>(
```
xDivisionByZerox: This function looks and feels more like a util function. Should it be moved to `./utils/get-all.ts` insted?

scripts/apidoc/processing/method.ts
```diff
@@ -0,0 +1,199 @@
...
+  name: string;
+  /**
+   * The signatures of the method.
+   */
+  signatures: RawApiDocsSignature[];
+  /**
+   * The relative path to the source file followed by the line number and column.
+   *
+   * @example
+   * `src/faker.ts:123:45`
```
xDivisionByZerox: Instead of processing all these values and then sticking everything into one property, why not keep `filePath`, `lineNumber` and `colomn` as separate properties. Maybe even in a separate `Source` object :thinking:
All the places that use this property, destructure or override it in some way that negates the previous processing.

scripts/apidoc/output/page.ts
```diff
@@ -0,0 +1,171 @@
...
+    description: mdToHtml(description),
+    parameters: parameters.map((param) => ({
+      ...param,
+      type: param.type.text,
+      default: param.default ?? extractSummaryDefault(param.description),
+      description: mdToHtml(param.description.replace(defaultCommentRegex, '')),
+    })),
+    since,
+    sourcePath: sourcePath.replaceAll(/:(\d+):\d+/g, '#L$1'),
+    throws: throws.length === 0 ? undefined : mdToHtml(throws.join('\n'), true),
```
xDivisionByZerox: Maybe we want to display a list of possible errors instead of a text? If we want to change this in the future, this seams kinda complicated, tho.

scripts/apidoc/processing/class.ts
xDivisionByZerox: We will stick with `RawApiDocsPage`. Lables, scripts and commit topics commonly use `docs` over `doc`.

Most important is that all the occurances should be consistently named.

scripts/apidoc/output/page.ts
ST-DDT: This just restores the current behavior, later we will change way things are displayed and then we will tackle that as well.

scripts/apidoc/processing/method.ts
ST-DDT: Good idea.

test/scripts/apidoc/__snapshots__/signature.spec.ts.snap
xDivisionByZerox, line 1: This file should not be located here, but under `/test/scripts/apidocs/__snapshots__` (apidocs with "**s**"). Or am I missing something?

test/scripts/apidocs/__snapshots__/method.spec.ts.snap
```diff
@@ -0,0 +1,1671 @@
...
+  ): T;",
+      "since": "1.0.0",
+      "throws": [],
+    },
+  ],
+  "source": {
+    "column": -1,
+    "filePath": "test/scripts/apidocs/method.example.ts",
+    "line": -1,
+  },
```
xDivisionByZerox, line 109: All sources have incorrect `column` and `line` properties. Is this expected in the snapshot tests?

scripts/apidocs/processing/type.ts
```diff
@@ -0,0 +1,228 @@
...
+  return {
+    type: 'union',
+    types,
+    text: types
+      .map((t) => t.text)
+      .map((t) =>
+        // T | U -> (T | U)
+        // But NOT Array<T | U> -> (Array<T | U>)
+        // () => T -> (() => T)
+        (t.includes('|') && !/^[A-Z]+<[^>]+>$/i.test(t)) || t.includes('(')
```
xDivisionByZerox: Instead of writing the comment (which took some brain power to compute whats goint on for me) we could extract the boolean statement into a variable with a fitting name. Maybe the RegExp itself as well.

Maybe:
- Regex => isGenricArrayTypeRegEx
- Statement => isDirectUnion

scripts/apidocs/utils/value-checks.ts
xDivisionByZerox, line 1: Should all of these functions return assertions instead of values? Currently, they are use in that way.

xDivisionByZerox (commented):
I'm still a bit torn on the "process" naming over a simple "get" prefix. Maybe even better, using a factory pattern.

Nevertheless, I think this PR includes a lot of work. We should propably go with the current implementation for now. If we feel discomfort with maintaining the documentation in the future, we can refactor then.

Thats why these will propably be my last comment on this PR, before approving.

test/scripts/apidocs/__snapshots__/method.spec.ts.snap
ST-DDT: I explicitly blanked these values to avoid scrambling all test snapshots due to changing line numbers if a new test case has been added.

scripts/apidocs/utils/value-checks.ts
ST-DDT: Then I would still have to actually get the values afterwards.

I don't like TS assertions because it doesn't allow me to return things.
(But that might just me bein a Java developer)

> Currently, they are use in that way.

IIRC they are almost every time used inline instead of as a standalone assertion.
So I'm not sure what you are hinting at.

xDivisionByZerox: > IRC they are almost every time used inline instead of as a standalone assertion.

You are totally right. Must have mixed something up inbetween the files. Sorry about that.

xDivisionByZerox: > IRC they are almost every time used inline instead of as a standalone assertion.

You are totally right. Must have mixed something up inbetween the files. Sorry about that.

test/scripts/apidocs/__snapshots__/method.spec.ts.snap
xDivisionByZerox: Ok, I see it now:

https://github.com/faker-js/faker/blob/222599367a36dfe8f945f38c642f84dd47a384c2/test/scripts/apidocs/method.spec.ts#L21-L25

That reasoning perfectly makes sense for me.

test/scripts/apidoc/__snapshots__/signature.spec.ts.snap
ST-DDT: I forgot to delete these files.

Removed.

scripts/apidocs/processing/type.ts
ST-DDT: Simplified check and applied.

ST-DDT:
> Maybe even better, using a factory pattern.

Not sure what you are asking for here.

---

Ready for review.

Shinigami92 (approved):
I would like to unblock this from merge
We can still improve in further iterations :slightly_smiling_face:

xDivisionByZerox (approved):

ST-DDT:
I merge this slightly early to avoid merge conflicts after the dependency updates.

PR #2630, opened 2024-01-28
docs: fix missing or invalid jsdocs

ST-DDT:
During #2628 I found some issues with the existing jsdocs. This PR fixes them.
We should probably include them in a v8.x bugfix release to have a clean end to v8.

### DIFF
--- a/src/faker.ts
+++ b/src/faker.ts
@@ -175,7 +175,7 @@ export class Faker extends SimpleFaker {
    * @param options.locale The name of the main locale to use.
    * @param options.localeFallback The name of the fallback locale to use.
    *
-   * @example Why are you looking at examples here? It's deprecated!
+   * @example new Faker({ locales });
    *
    * @since 6.0.0
    *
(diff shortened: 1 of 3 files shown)


src/faker.ts
```diff
@@ -173,6 +175,10 @@ export class Faker extends SimpleFaker {
    * @param options.locale The name of the main locale to use.
    * @param options.localeFallback The name of the fallback locale to use.
    *
+   * @example Why are you looking at examples here? It's deprecated!
```
Shinigami92: I wont like to support passive aggressive phrases in our docs :thinking:

Shinigami92 (changes requested):

src/faker.ts
ST-DDT: Then please provide a working example.

Shinigami92: the absolut minimum would be:
```suggestion
   * @example new Faker({ locales });
```

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #2631, opened 2024-01-28
chore(deps): update pnpm-lock

ST-DDT:
This PR hopefully fixes the irrelevant warning during push:
https://github.com/faker-js/faker/security/dependabot/16

````diff
- vite: 5.0.11(@types/node@20.11.6)
+ vite: 5.0.12(@types/node@20.11.6)
````

For the future we should let renovate do this, but I just want the warning gone.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -148,9 +148,9 @@ packages:
     resolution: {integrity: sha512-alScpSVnYmjNEXboZjarjukQEzgCRmjMv6Xj47fsdnqGS73bjJNDpiiXmp8jr0UZLdUB6d9jW63IcmddUP+l0g==}
     dependencies:
       '@actions/http-client': 2.2.0
-      '@octokit/core': 5.0.2
-      '@octokit/plugin-paginate-rest': 9.1.5(@octokit/core@5.0.2)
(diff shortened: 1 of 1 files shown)


xDivisionByZerox (approved):
> For the future we should let renovate do this, but I just want the warning gone.

Isn't u/dependabot the bot that usually creates security auto fix PR's?

ST-DDT:
grafik

I'm not sure why exactly but this is either because it is not a direct dependency, or because we have at least two major versions of that dependency in the lock file.

Shinigami92:
> grafik
> 
> 
> 
> I'm not sure why exactly but this is either because it is not a direct dependency, or because we have at least two major versions of that dependency in the lock file.

That could be right 🤔
We definitely have one from Vitest and one from VitePress

Normally in that case de-duplication takes place if possible.
I cannot look right now into the change because of mobile screen

---

Edit:

We do have v5 and v4 in our dependency-tree. I'm not totally sure what you are referring to now. Maybe lets talk later on Discord to show and clarify what the problems are.

ST-DDT:
> I'm not totally sure what you are referring to now. Maybe lets talk later on Discord to show and clarify what the problems are.

The issue is that dependabot does not create a fix PR itself and I had to do that myself.

There is no problem with this PR itself.

Shinigami92 (approved):

PR #2608, opened 2024-01-19
refactor!: remove v8 deprecations

Shinigami92:
Remove all v8 deprecations

_we might need to split this PR into several smaller PRs, until this PR itself is small enough_

PR #2634, opened 2024-01-29
refactor(image)!: remove v8 deprecated providers

Shinigami92:
extracted from https://github.com/faker-js/faker/pull/2608


docs/guide/upgrading_v9/2634.md
```diff
@@ -0,0 +1,3 @@
+### Remove deprecated image providers
+
+Removed deprecated image providers from `faker.image`.
```
xDivisionByZerox: Sorry if we discuss this somewhere else before. But if I see correctly we did not merge anything for v9 so far.

Should we include a list/table of API calls that are no longer supported? That why a user that is migrating can simply do a global search in their codebase when they update. I know that in an optimal world, test or build processes would fail for said user ("cannot read x of y") but better be safe I guess.

Shinigami92: You can already find such a table prepared in https://github.com/faker-js/faker/pull/2608#issuecomment-1901742274

However, this PR now only removes the image providers and is just a piece of many deprecation removals

The `docs/guide/upgrading_v9/*.md` are only there for us, so before merging/releasing v9 we can rework the whole migration guide. There are just temp files.

xDivisionByZerox: > You can already find such a table prepared in https://github.com/faker-js/faker/pull/2608#issuecomment-1901742274

Thank you for pointing that out. As  far as I can see, the image modules are not in this table (at least not the full API).

> The docs/guide/upgrading_v9/*.md are only there for us, so before merging/releasing v9 we can rework the whole migration guide. There are just temp files.

I know. The point is that a single PR creates a template for the migration that is as complete as it can be. So why not include the part of the table that is actually gets removed here? If it is not included here, there is a chance that we might miss that. If you do not split the migration templates, why bother making this PR in the first place? Where does the table from the comment end up in, if all depreations are spread into different PRs?

Shinigami92: So first, removing all deprecations in one PR would violate our contribution guidelines because to many files are affected at the same time.

Secondly there are right now failing CI checks in the other big PR #2608 and it is ultra hard to debug out of the full clutter of that many changes to find out what the issues are right now.
So cutting these down into smaller chunks helps to already remove some deprecations until we found what is the missing and failing rest.

ST-DDT: I think @xDivisionByZerox (and I) is asking to have the table with alternative image methods here in this document.

Shinigami92: I fully misunderstood you intend. I discussed with @ST-DDT and now I got you. Added.

docs/guide/upgrading_v9/2634.md
```diff
@@ -0,0 +1,21 @@
...
+| faker.placeholder.imageUrl          | faker.image.urlPlaceholder  |
+| faker.placeholder.randomUrl         | faker.image.urlPlaceholder  |
+| faker.unsplash.image                | faker.image.url             |
+| faker.unsplash.imageUrl             | faker.image.url             |
+| faker.unsplash.food                 | faker.image.url             |
+| faker.unsplash.people               | faker.image.url             |
+| faker.unsplash.nature               | faker.image.url             |
+| faker.unsplash.technology           | faker.image.url             |
+| faker.unsplash.objects              | faker.image.url             |
+| faker.unsplash.buildings            | faker.image.url             |
```
ST-DDT: Maybe refer to a method that supports categories?
Such as `faker.image.urlLoremFlickr({ category: 'food' })`?

Shinigami92: That does not exists anymore, I think because the related providers are not available anymore at all :grimacing:

ST-DDT (dismissed):

docs/guide/upgrading_v9/2634.md
matthewmayer: faker.image.urlLoremFlickr({ category: 'food' }) does still exist. The remaining two providers are LoremFlickr (which does support categories) and Picsum (which does not).

docs/guide/upgrading_v9/2634.md
```diff
@@ -0,0 +1,21 @@
+### Remove deprecated image providers
+
+Removed deprecated image providers from `faker.image`.
+
+| old                                 | replacement                 |
+| ----------------------------------- | --------------------------- |
+| faker.lorempicsum.image             | faker.image.urlPicsumPhotos |
```
matthewmayer: im confused by this "old" column, for example there was never a method called faker.lorempicsum.image() as far as i know?

matthewmayer (changes requested):

docs/guide/upgrading_v9/2634.md
matthewmayer: ah i think it should be `faker.image.lorempicsum.image()` ?

Shinigami92: Well... I think our deprecations are wrong then :grimacing: 
https://github.com/faker-js/faker/blob/bc3ebb793666c0d7e58cf428ba12465a7b290cbe/src/modules/image/providers/lorempicsum.ts#L33-L38
I just copy pasted out of the source code

Did we not see that or do we need to release a v8.4.2 now?

matthewmayer: I think almost no one was probably using these methods as they weren't publicly documented. They would only use them indirectly via the faker.image.foo methods?

faker.image.lorempicsum.image()

ST-DDT (dismissed):

docs/guide/upgrading_v9/2634.md
```diff
@@ -0,0 +1,21 @@
+### Remove deprecated image providers
+
+Removed deprecated image providers from `faker.image`.
```
matthewmayer: Maybe we should explain that these were already returning broken image urls?

matthewmayer (changes requested):

docs/guide/upgrading_v9/2634.md
Shinigami92:
```suggestion
Removed deprecated image providers from `faker.image`. They already returned broken image URLs anyways.
```

ST-DDT (approved):

matthewmayer (approved):

PR #2635, opened 2024-01-30
docs: clarify that the result of faker.date.betweens() is sorted

matthewmayer:
Inspired by https://github.com/faker-js/faker/discussions/2633 - wasn't obvious from the docs that the result was sorted.


src/modules/date/index.ts
```diff
@@ -409,7 +409,7 @@ export class SimpleDateModule extends SimpleModuleBase {
   }
 
   /**
-   * Generates random dates between the given boundaries.
+   * Generates random dates between the given boundaries. The dates will be returned in an array sorted from oldest to newest.
```
ST-DDT: I'm a little bit unsure about newest here because for me newest does not include future ones, but that might be just my imagination.
Would ascendeing order be clearer?

matthewmayer: "In chronological order"?

ST-DDT:
```suggestion
   * Generates random dates between the given boundaries. The dates will be returned in an array sorted in chronological order.
```

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

PR #2653, opened 2024-02-09
infra(unicorn): prefer-string-replace-all

Shinigami92:
Ref: #2439

- #2439

---

Permanently enables the [`unicorn/prefer-string-replace-all`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md) lint rule.


docs/guide/upgrading_v9/2653.md
```diff
@@ -0,0 +1,6 @@
+### Use replaceAll over replace in code
+
+`replaceAll` is a feature that is only supported by Node >=v15, so this is a breaking change for users that are using Node < v15.  
+However, we dropped Node < v18 anyways.
+
+https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll
```
ST-DDT: For me this is redundant, we already have 

https://github.com/faker-js/faker/blob/bc3ebb793666c0d7e58cf428ba12465a7b290cbe/docs/guide/upgrading_v9/2121.md?plain=1#L1-L3

for that. IMO we don't have to explain to the users that the code breaks because of `??=` or `String.replaceAll`.
We want to use new code features, that is the reason we drop support for old versions.

Shinigami92: Yeah, the upgrading guidelines will get merged before releasing `v9.0.0-alpha.0` anyways.
So this is more like an explanation why there was a `!` / breaking change marker for that PR.

Later we can then e.g. merge that together and not only write "drop v16" but also "why -> code examples / which code features are used now"

ST-DDT: The problem with the breaking `!` marker, is that it causes the PR to be listed in the changelog as breaking changes, and IMO it adds unnecessary noise to that section.

Shinigami92: Ah :thinking: you have a point there

Okay, I will remove the ! from PR title.

Not sure yet if I should keep or remove the md file, because that still might help to explain the "why" later when we merge the migration files.
I would like to leave this decision to @matthewmayer

matthewmayer: i agree. updating to use Node-18 specific features is not breaking now that Node 14/16 has been dropped, and doesn't require an explanation in the upgrading guide.

docs/guide/upgrading_v9/2121.md
```diff
@@ -1,3 +1,10 @@
 ### Node 14 and 16 No Longer Supported
 
 Support for Node.js versions 14 and 16 has been discontinued as these versions have reached their [end-of-life](https://github.com/nodejs/Release). Faker.js 9.0 requires a minimum of Node.js version 18.
+
+#### Examples of code-upgrades that are now used:
```
ST-DDT: I think this is unneccessary as outlined here https://github.com/faker-js/faker/pull/2654#discussion_r1485033229

ST-DDT (changes requested):

ST-DDT (dismissed):

ejcheng (dismissed):

ST-DDT (approved):

PR #2654, opened 2024-02-09
infra(unicorn): prefer-at

Shinigami92:
Ref: #2439

- #2439

---

Permanently enables the [`unicorn/prefer-at`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-at.md) lint rule.


src/modules/helpers/index.ts
```diff
@@ -989,7 +989,8 @@ export class SimpleHelpersModule extends SimpleModuleBase {
     }
 
     // In case of rounding errors, return the last element
-    return array[array.length - 1].value;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return array.at(-1)!.value;
```
Shinigami92, line 993: Adding a test for this is not related to the PR :slightly_smiling_face:

ST-DDT (dismissed):

ST-DDT:
IMO this isn't breaking for the same reasons as outlined there: https://github.com/faker-js/faker/pull/2653#discussion_r1484660651

scripts/apidoc/typedoc.ts
```diff
@@ -313,7 +314,9 @@ export function extractSummaryDefault(
 
   if (eraseDefault) {
     summary.splice(-2, 2);
-    const lastSummaryPart = summary[summary.length - 1];
+    const lastSummaryPart =
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
matthewmayer: this ends up being much more verbose than the original code due to the need to disable the no-non-null-assertion. Is this temporary or is there a better way around this?

docs/guide/upgrading_v9/2654.md
matthewmayer, line 1: i dont think its necessary to include this if its not end-user facing.

matthewmayer (dismissed):

docs/guide/upgrading_v9/2654.md
Shinigami92: What do you mean by not including if not end-user facing? Most if not nearly all chores and infra changes are not end-user facing :thinking: 
This enhances the code base for us, the devs. Right now we do not have `noUncheckedIndexedAccess` turned on, and so we miss many potential hidden null-safety bugs. This change does not only improve DX (developer experience) through shorter code + inline docs

image

but also enforces the need of check for `undefined` if there was no last value (e.g. when the array was empty)
_Sadly_ TypeScript is not able to infer the code flow to check that `.at(-1)` is coming after a `array.length > 0` check, but how even could it, cause the input parameter for `at` is just a number.

scripts/apidoc/typedoc.ts
Shinigami92: this is mostly because I specifically decided to put that absolutely near to the affected code part instead of just above the line

so it could also be

```ts
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
const lastSummaryPart = summary.at(-1)!;
```

docs/guide/upgrading_v9/2121.md
```diff
@@ -1,3 +1,10 @@
 ### Node 14 and 16 No Longer Supported
 
 Support for Node.js versions 14 and 16 has been discontinued as these versions have reached their [end-of-life](https://github.com/nodejs/Release). Faker.js 9.0 requires a minimum of Node.js version 18.
+
+#### Examples of code-upgrades that are now used:
```
ST-DDT: I think you can drop this section entirely.
If you think it is important to point out why we dropped support, just mention "we would like to use new features that these versions provide such as `||=` and the others are EOL anyway." One such example is sufficient.

ST-DDT (changes requested):

docs/guide/upgrading_v9/2121.md
Shinigami92: Lets discuss internally further on or wait on next team meeting

docs/guide/upgrading_v9/2654.md
matthewmayer: It's a good change. It just doesn't need to be in the migration guide. Because the migration guide is for people who are using Faker. If you are upgrading from v8 to v9 you just need to ensure you are using Node 18+. The fact that Faker is internally now using certain features of Node 18 should not matter to use as a consumer of the library.

docs/guide/upgrading_v9/2121.md
Shinigami92: We decided to only reference other migration files, but not change them directly

ST-DDT (dismissed):

docs/guide/upgrading_v9/2654.md
```diff
@@ -0,0 +1,7 @@
+### Examples of code-upgrades that are now used
+
+_This upgrade is an extension to_ [#2121](./2121.md)
+
+Used Node >= v16.6 feature [`at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).
+
+We could hint users to use polyfill or transpile-down steps.
```
ST-DDT, line 7: Note from my future me: We will have at most a very minimum description of "how to transpile down for node < x". I won't help them shoot themselves in the foot and complain to us about it.

And I won't keep track of any node <18 feature usage myself unless specifically requested by our community.

ejcheng (dismissed):

docs/guide/upgrading_v9/2654.md
matthewmayer: If you are stuck using node 14/16 you should just stay on faker@8 rather than try to transpile things.

ST-DDT (approved):

PR #2656, opened 2024-02-09
infra(eslint): logical-assignment-operators

Shinigami92:
Permanently enables the [`logical-assignment-operators`](https://eslint.org/docs/latest/rules/logical-assignment-operators) lint rule.

This will be in draft until we merged the deprecation PRs, because this will reduce the diff to only affect 1 line in test.

### DIFF
--- a/test/modules/finance.spec.ts
+++ b/test/modules/finance.spec.ts
@@ -168,7 +168,7 @@ describe('finance', () => {
         it('should set a specified length', () => {
           let expected = faker.number.int(20);
 
-          expected = expected || 4;
+          expected ||= 4;
 
           const mask = faker.finance.maskedNumber({
             length: expected,
--- a/test/modules/finance.spec.ts
+++ b/test/modules/finance.spec.ts
@@ -166,9 +166,7 @@ describe('finance', () => {
         });
 
         it('should set a specified length', () => {
-          let expected = faker.number.int(20);
-
-          expected ||= 4;
+          const expected = faker.number.int(20) || 4;
 
           const mask = faker.finance.maskedNumber({
             length: expected,
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -33,6 +33,7 @@ module.exports = defineConfig({
   },
   rules: {
     eqeqeq: ['error', 'always', { null: 'ignore' }],
+    'logical-assignment-operators': 'error',
     'no-else-return': 'error',
     'no-restricted-globals': ['error', 'Intl'],
     'prefer-exponentiation-operator': 'error',


test/modules/finance.spec.ts
```diff
@@ -205,7 +205,7 @@ describe('finance', () => {
         it('should set a specified length', () => {
           let expected = faker.number.int(20);
 
-          expected = expected || 4;
+          expected ||= 4;
```
```suggestion
          const expected = faker.number.int(20) || 4;
```

test/modules/finance.spec.ts
```diff
@@ -239,7 +239,7 @@ describe('finance', () => {
...
```
```suggestion
          const expected = faker.number.int(20) || 4;
```

xDivisionByZerox (dismissed):

ST-DDT (approved):

test/modules/finance.spec.ts
```diff
@@ -166,9 +166,7 @@ describe('finance', () => {
         });
 
         it('should set a specified length', () => {
-          let expected = faker.number.int(20);
-
-          expected = expected || 4;
+          const expected = faker.number.int(20) || 4;
```
xDivisionByZerox, line 169: Quick question. Instead of using a dynamic value here, wouldn't it be better to use a static one. Now, the test for maskNumber relies on the implementation of int, which would allow it to potentially randomly fail. If one would want to test multiple values, an array or it.each would work for that. 

If one does not want that, would it not be possible to set min and max explicitly here. That way, the logic assignment is completely eliminated. Furthermore, seeing the || makes my brain think about catching null and undefined values for a second. By getting rid of it, I would argue that the test is easier to read.

This is not the scope for this issue, but I would say that writing tests are one of my main weaknesses. That's why I wanted to get feedback if the premise of my thought process is fundamental, correct, or flawed.

ST-DDT: The comment sais it all

> the length of mask picks 4 if the random number generator picks 0

IMO 1-20 provides better coverage and is easier to understand than the || and the comment.

We already changed some code during these improvements, so I think we can do that here as well.

Shinigami92: Well... the whole test-intention slipped out of scope 😬 

Previously it was testing something like when 0 is passed, 4 is intended to be the fallback.
However, somewhere here https://github.com/faker-js/faker/blob/6dee178558b87b73bba1395c11d2ffe3d156dad1/src/modules/finance/index.ts#L323 or here https://github.com/faker-js/faker/blob/6dee178558b87b73bba1395c11d2ffe3d156dad1/src/modules/string/index.ts#L608-L609 the behavior changed and the test did not covered correctly at all.

@ST-DDT already wrote with me on Discord that we might want to throw an error and rewrite the behavior.

ST-DDT: Simplifing the existing test isnt out of scope as it doesnt change what the test does and you are already editing the line anyway for the exact reason this PR exists in the first place.
So lets get this out of the way first and then we can have a look at maskedNumber and potentially new tests later.

ejcheng (approved):

test/modules/finance.spec.ts
xDivisionByZerox: This conversation got resolved. I'm still not sure if I understand what should be dont here, right now. Change the test in the PR or merge and change later?

Shinigami92: Merge and change later is currently the by far most easiest option

xDivisionByZerox (approved):

test/modules/finance.spec.ts
ST-DDT:
```suggestion
          const expected = faker.number.int({ min: 1, max: 20 });
```

Issue #2335, opened 2023-08-22
Can add `sideEffects: false` to `package.json`


haydenull:
Is it possible to add sideEffects: false in the package.json file?

This can help rollup with tree shaking.

For example, I want to use fakerjs in development mode but remove it in production mode.

```ts
import { faker } from '@faker-js/faker'

if (import.meta.env.DEV) console.log(faker)
```

When sideEffects is not set to false, even if the condition of if statement is false, faker will still be bundled together.

ST-DDT:
What are the requirements for our code to use that safely?

haydenull:
In web development, intercept API requests and use faker return random data.

However, in production mode, interception is not necessary.

ST-DDT:
Sorry, thats not what I meant.
What requirements has sideEffect: false for the Faker code? (Why isnt it true by default if it is that good.)

haydenull:
Because removing code during tree shaking is a risky behavior, if a package not specify sideEffects as false, bundlers like rollup will assume that the code in this package has side effects.

In order to ensure the proper functioning of the code, rollup will retain them during tree shaking.

Wether to set `sideEffects: false` depends on whether the code of faker has any side effects. No modifications to the source code are required.

However, I am not familiar with the code of faker, so I cannot make a judgment.

ST-DDT:
> Wether to set `sideEffects: false` depends on whether the code of faker has any side effects.

What is considered a side effect? A Faker internally shared global? Or does side effect require interaction with a third party library like registering a plugin in prettier?

haydenull:
I think for faker, the main question is wether it modifies window or global objects. Or wether it introduced other packages with side effects.

`sideEffects: false` instructs the bundler to remove this package if its exported content is not being used.

I think if my code doesn't use any methods exported by faker, removing faker from build output would not break the program. Therefore, it can be considered as having no side effects.

matthewmayer:
even if harmless, if this affects the behavior of the bundler it should probably be considered a semver-major change.

PR #2661, opened 2024-02-11
refactor(helpers)!: remove v8 deprecated unique

Shinigami92:
extracted from #2608

- #2608

referencing #1785

- #1785

should be merged before starting #2335

- #2335


ST-DDT:
As mentioned earlier. I don't think we should remove unique from faker as it is a very useful feature when generating test data.
I would like to check with the rest of the team and to a certain extend the community, whether we should remove it or replace it.

EDIT: Reason: #2667 in combination with #2664 will likely invalidate the circumstances that the initial decision was made upon.

Please vote this comment with 
- 👍 for removal of unique and 
- 👎 for retaining unique, although with a different implementation and interface.

matthewmayer:
What about the possibility of taking an external replacement like https://www.npmjs.com/package/enforce-unique and moving it within the faker-js org, but as a separate package/repo?

Shinigami92:
> As mentioned earlier. I don't think we should remove unique from faker as it is a very useful feature when generating test data.
> 
> I would like to check with the rest of the team and to a certain extend the community, whether we should remove it or replace it.
> 
> 
> 
> Please vote this comment with 
> 
> - 👍 for removal of unique and 
> 
> - 👎 for retaining unique, although with a different implementation and interface.

I need to say that I 100% dislike to bring up a long discussion we already had and even written down in our internal notes. As Matthew already wrote: there is another package that does the same if not better.
And for all the reasons: please lock back in history notes. Period.

matthewmayer:
Let's just make sure we have very good documentation explaining how to migrate from using the existing method to a third-party package.

matthewmayer:
i went ahead and added some documentation and suggested alternatives direct on this branch. Feedback welcome.

Shinigami92:
Thank you @matthewmayer 
I rebased it, fixed a conflict and fixed some minor typos

LGTM 👍

docs/guide/unique.md
```diff
@@ -0,0 +1,120 @@
...
+
+Some methods and locales use a much larger data sets than others. For example `faker.animal.type` has only 13 possible animals to choose from. By contrast `faker.person.fullName()` pulls from a list of hundreds of first names, surnames and prefixes/suffixes, so it can generate hundreds of thousands of unique names. Even then, the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_Paradox) means that duplicate values will quickly be generated.
+
+Sometimes you want to generate unique values, for example you may wish to have unique values in a database email column.  
+There are a few possible strategies for this:
+
+1. Use `faker.helpers.uniqueArray()` if you want to generate all the values at one time. For example
+
+```ts
+faker.helpers.uniqueArray(faker.internet.email, 1000); // will generate 1000 unique email addresses
```
ST-DDT, line 21: IMO it is inconsequetial to remove unique but keep uniqueArray.

docs/guide/usage.md
```diff
@@ -233,45 +233,6 @@ Here, we could also pass in the `sex` value as argument, but in our use-case the
...
-```
-
-By wrapping Faker's `email` function with the [`unique`](../api/helpers.md#unique) helper function, we ensure that the return value of `email` is always unique.
-
-::: warning
-The `faker.helpers.unique` is targeted to be removed from Faker in the future.  
-Please have a look at the issue [#1785](https://github.com/faker-js/faker/issues/1785).  
-We will update these docs once a replacement is available.
-:::
+Unlike the `_id` property that uses an `uuid` implementation, which is unique by design, the `email` may contain duplicates. We have a dedicated guide page on generating [unique values](unique).
```
ST-DDT: Uuid is not unique by design but by chance.

ST-DDT:
Please dont rebase unless necessary as it makes reviews harder.

docs/guide/unique.md
Shinigami92: I dont understand?! uniqueArray uses shuffle which uses int which is based on the seed

unique however does not use seed at all and is a plain utility function not bound to faker's seed

ST-DDT: https://github.com/faker-js/faker/blob/7c7f78daa0fef802af916c2f192cc36bee6c88d2/src/modules/helpers/index.ts#L760

Shinigami92: https://github.com/faker-js/faker/blob/7c7f78daa0fef802af916c2f192cc36bee6c88d2/src/modules/helpers/index.ts#L751

matthewmayer: Difference is that uniqueArray is just unique for the lifetime of that method call. 

Whereas unique was unique for the lifetime of that faker instance.

docs/guide/usage.md
matthewmayer: Maybe just remove the comment about UUIDs completely? It's not very important for this section of the docs.

matthewmayer:
This will also need a migration guide snippet. Hopefully it can be quite short now and just link to the unique guide page.

docs/guide/unique.md
ST-DDT: @Shinigami92 If my post wasnt clear enough: There are two branches in that method. One for arrays and one for functions.

Yes, and I would like to have a method that I can pass an array of known ones and it generates a single new one.

Shinigami92: Did you know when you pass `faker.datatype.boolean({ probability: 0 })` it always return `false` and even dont consume a faker seed.
However, unique does not use something from faker internally at all, in any kind of branch.

But TBH, I'm tired of starting a discussion of something we already discussed to its exhaustiveness half a year ago.

ST-DDT: > However, unique does not use something from faker internally at all, in any kind of branch.

I'm aware.



> But TBH, I'm tired of starting a discussion of something we already discussed to its exhaustiveness half a year ago.

**Important**: My question is not about `unique`, but `uniqueArray(() => fn())`.

Is `uniqueArray(()=>fn())` a faker-ish method for you?

<details><summary>If yes</summary>
Hypothetical: Is a method using uniqueArray for generating random values first and returning a function to return/pop an elememt from that list faker-ish for you?

````ts
const list = uniqueArray(source);
return () => if list.empty throw else yield list.shift();
````

Not that I think about adding/replacing it like that.

</details>

<details>
<summary>If no,</summary>
 we should remove that as well for consistency.
</details>

Comment (Shinigami92) [reply docs/guide/unique.md:21]: I'm not sure right now, maybe discuss in a meeting 🤔 
I never used `uniqueArray` and I also was not aware that it was already there since the faker-takeover 👀 

But I would definitely see this in another PR for now, cause this PR currently aims for removing the one method that bleeds side-effects and so holds other PR back.

**Comment by Shinigami92** [comment]
> This will also need a migration guide snippet. Hopefully it can be quite short now and just link to the unique guide page.

added

File: docs/guide/usage.md (hunk 2 above)
Comment (ST-DDT) [reply docs/guide/usage.md]:
```suggestion
Unlike the `_id` property that uses an `uuid` implementation, which has a low chance of duplicates, the `email` function is more likely to produce duplicates, especially if the call arguments are similar. We have a dedicated guide page on generating [unique values](unique).
```

docs/guide/unique.md
```diff
@@ -0,0 +1,120 @@
...
+    email,
+  };
+}
+```
+
+## Migrating from faker.helpers.unique
+
+Prior to v9, Faker provided a [`faker.helpers.unique()`](https://v8.fakerjs.dev/api/helpers.html#unique) method which had a global store to keep track of duplicates. This was removed in v9.
+
+In general you should switch to using one of the strategies above. For example, migrating to `enforce-unique`:
```
ST-DDT: IMO this section better suited in the migration/upgrading guide.

docs/guide/unique.md
```diff
@@ -0,0 +1,120 @@
...
+
+Basic example:
+
+```js
+//OLD
```
ST-DDT: I think most of our code uses a space after the comment slashes e.g. `// OLD`.
Should we change these for consistency or is that just my personal preference?

docs/guide/unique.md
```diff
@@ -0,0 +1,120 @@
...
+import { UniqueEnforcer } from 'enforce-unique';
+
+const enforcer = new UniqueEnforcer();
+const city = enforcer.enforce(faker.location.city, {
+  maxRetries: 100,
+  maxTime: 1000,
+});
+```
+
+Note `enforce-unique` does not support the `exclude` or `store` options. If you were previously using these, you may wish to build your own unique logic instead.
```
ST-DDT:
```suggestion
Note: `enforce-unique` does not support the `exclude` or `store` options. If you were previously using these, you may wish to build your own unique logic instead.
```

Or maybe use an actual `::: tip Note` block?

docs/guide/unique.md
Shinigami92: I observed these comments written almost and only by @matthewmayer  
After a while I decided to not care anymore
Either @matthewmayer should use `// ...` or stay at `//...` :shrug:

docs/guide/unique.md
Shinigami92: I'm for the `::: tip Note` variant, cause that is mostly the one suggested for inside VitePress

docs/guide/unique.md
Shinigami92: @matthewmayer could you please take this part over?

matthewmayer: Sure

docs/guide/unique.md
```diff
@@ -0,0 +1,58 @@
...
+
+```ts
+faker.seed(55);
+faker.animal.type(); //'cat'
+faker.animal.type(); //'bird'
+faker.animal.type(); //'horse'
+faker.animal.type(); //'horse'
+```
+
+Some methods and locales use a much larger data sets than others. For example `faker.animal.type` has only 13 possible animals to choose from. By contrast `faker.person.fullName()` pulls from a list of hundreds of first names, surnames and prefixes/suffixes, so it can generate hundreds of thousands of unique names. Even then, the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_Paradox) means that duplicate values will quickly be generated.
```
ST-DDT: > much larger data sets than others. For example `faker.animal.type` has only 13

Optional: IMO it is strange to talk about larger sets and then talk about having only 13. Maybe flip the comparison around?

matthewmayer: Or change "larger" to "smaller"?

matthewmayer (dismissed):

MansurAliKoroglu:
I published a new version of `enforce-unique`. Now it supports exclude.

docs/guide/unique.md
```diff
@@ -0,0 +1,58 @@
...
+
+```ts
+faker.seed(55);
+faker.animal.type(); //'cat'
+faker.animal.type(); //'bird'
+faker.animal.type(); //'horse'
+faker.animal.type(); //'horse'
+```
+
+Some methods and locales use a much smaller data sets than others. For example `faker.animal.type` has only 13 possible animals to choose from. By contrast `faker.person.fullName()` pulls from a list of hundreds of first names, surnames and prefixes/suffixes, so it can generate hundreds of thousands of unique names. Even then, the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_Paradox) means that duplicate values will quickly be generated.
```
ejcheng:
```suggestion
Some methods and locales use much smaller data sets than others. For example, `faker.animal.type` has only 13 possible animals to choose from. In contrast, `faker.person.fullName()` pulls from a list of hundreds of first names, surnames, and prefixes/suffixes, so it can generate hundreds of thousands of unique names. Even then, the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_Paradox) means that duplicate values will quickly be generated.
```

docs/guide/unique.md
```diff
@@ -0,0 +1,58 @@
...
+
+Sometimes you want to generate unique values, for example you may wish to have unique values in a database email column.
```
ejcheng:
```suggestion
Sometimes, you want to generate unique values. For example, you may wish to have unique values in a database email column.  
```

docs/guide/unique.md
```diff
@@ -0,0 +1,58 @@
...
+faker.animal.type(); //'horse'
+faker.animal.type(); //'horse'
+```
+
+Some methods and locales use a much smaller data sets than others. For example `faker.animal.type` has only 13 possible animals to choose from. By contrast `faker.person.fullName()` pulls from a list of hundreds of first names, surnames and prefixes/suffixes, so it can generate hundreds of thousands of unique names. Even then, the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_Paradox) means that duplicate values will quickly be generated.
+
+Sometimes you want to generate unique values, for example you may wish to have unique values in a database email column.  
+There are a few possible strategies for this:
+
+1. Use `faker.helpers.uniqueArray()` if you want to generate all the values at one time. For example
```
ejcheng:
```suggestion
1. Use `faker.helpers.uniqueArray()` if you want to generate all the values at one time. For example:
```

ejcheng (approved):

matthewmayer (approved):

ST-DDT (approved):
Approved as per old team decision so it's no longer hard blocked, but I still think it's wrong.

Shinigami92:
> Approved as per old team decision so it's no longer hard blocked, but I still think it's wrong.

I owe you a favor 👍

PR #2676, opened 2024-02-21
infra: don't remove pnpm-lock on clean

ST-DDT:
IIRC this was mentioned in some release related discussions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "dist"
   ],
   "scripts": {
-    "clean": "rimraf coverage .eslintcache dist docs/.vitepress/dist locale pnpm-lock.yaml node_modules",
+    "clean": "rimraf coverage .eslintcache dist docs/.vitepress/dist locale node_modules",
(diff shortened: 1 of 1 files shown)


xDivisionByZerox (approved):

Shinigami92:
This would result in non-updated deeps sub-dev-dependencies.
We should verify that renovate does a deep upgrade for the lock file.
However this does not affect the released content at all and is just a thing for contributor using the faker repository while developing, cause the lock file is not shipped when publishing to the npm-registry.

ejcheng (approved):

ST-DDT:
> This would result in non-updated deeps sub-dev-dependencies. We should verify that renovate does a deep upgrade for the lock file. However this does not affect the released content at all and is just a thing for contributor using the faker repository while developing, cause the lock file is not shipped when publishing to the npm-registry.

I created https://github.com/faker-js/faker/pull/2677 to address that.

PR #2678, opened 2024-02-22
refactor(random)!: remove v8 deprecated random module

Shinigami92:
extracted from #2608


docs/guide/upgrading_v9/2678.md
```diff
@@ -0,0 +1,3 @@
+### Remove deprecated random module
+
+Removed deprecated random module
```
ST-DDT, line 3: IMO this should have a list of alternatives same as the other removed methods.

Shinigami92: I'm not sure if we really need this.
1. all consumers are already notified with runtime warnings a long time now
2. there are not specific methods in the module that are deprecated but the whole module
3. IMO we will have anyway a big section in the migration guide later and there we can do a table (if really needed) in one step

ST-DDT: If https://github.com/faker-js/faker/pull/2653 needs an upgrading guide, then this one needs one even more so.
Also it is marked with `!`, so it is "important" for users.

ST-DDT (approved):

matthewmayer (approved):

PR #2685, opened 2024-02-25
feat(helpers)!: use const generics where possible

### LINKED ISSUE #2679: Support `const` modifier on type parameters

As a developer using faker I want to benefit from `const` modifier on type parameters so that I don't need to explicitly add `as const`.

### Suggested solution

Add `const` modifier on type parameters.

### Alternative

_No response_

### Additional context

This feature was introduced by TypeScript v5.0.0 and would break TypeScript support < v5

ST-DDT:
Implements #2679

- #2679

This feature requires TS5+ and thus is a breaking change.


Shinigami92:
Should we add a https://vitest.dev/api/expect-typeof.html for this specific case?

ST-DDT:
> Should we add a [vitest.dev/api/expect-typeof.html](https://vitest.dev/api/expect-typeof.html) for this specific case?

I haven't used that feature yet, and I do not know whether/how to proceed regarding testing of this type only feature.

ST-DDT:
Would you consider this `!` breaking or not in the light of #1953?

- #1953

This is a user facing change, but it technically isn't anymore breaking than using the new node features after updating to them.

docs/guide/upgrading_v9/2685.md
```diff
@@ -4,4 +4,13 @@ _This upgrade is an extension to_ [#1953](./1953.md)
 
 The helpers module now uses TS5 features, thus Faker requires TS5 for it to be used.
```
matthewmayer: This seems slightly misleading, it makes it sound like you have to use TS to use Faker. But obviously you can use it with Javascript.

Perhaps 
> "The helpers module now uses TS5 features, so if you are using Faker with Typescript, you must use TS5."

matthewmayer (changes requested):

Shinigami92 (approved):

matthewmayer (approved):

ejcheng (approved):

PR #2686, opened 2024-02-25
feat(locale): add uzbek locale (uz_UZ_latin)

Mirazyzz:
Added initial locale values for person in Uzbek language.


ST-DDT (commented):
IMO we should rename `uz_UZ` to `uz`.

ST-DDT:
Also please run `pnpm run preflight`

Mirazyzz:
> IMO we should rename `uz_UZ` to `uz`.

Uzbek language consist of latin and cyrillic alphabets. I thought we could include support for both in the future.
https://en.wikipedia.org/wiki/Uzbek_alphabet

matthewmayer:
> IMO we should rename `uz_UZ` to `uz`.

For new locales i generally prefer lang_COUNTRY - as soon as you start adding addresses and phone numbers, they are typically specific for a certain country.

ST-DDT:
> > IMO we should rename `uz_UZ` to `uz`.
> 
> Uzbek language consist of latin and cyrillic alphabets. I thought we could include support for both in the future. https://en.wikipedia.org/wiki/Uzbek_alphabet

Ok, if there are two variants of the locale then we should make that visible in the name as well similar to https://github.com/faker-js/faker/tree/next/src/locales/sr_RS_latin

Or is there a different common disambiguator?

Mirazyzz:
> > > IMO we should rename `uz_UZ` to `uz`.
> > 
> > 
> > Uzbek language consist of latin and cyrillic alphabets. I thought we could include support for both in the future. https://en.wikipedia.org/wiki/Uzbek_alphabet
> 
> Ok, if there are two variants of the locale then we should make that visible in the name as well similar to https://github.com/faker-js/faker/tree/next/src/locales/sr_RS_latin
> 
> Or is there a different common disambiguator?

so would it be better to rename it to **uz_latin**? I don't think there is a common disambiguator that can be used like "uz_UZ" and "uz_CY" or something like that so we could probably use **uz_latin** and **uz_cyrillic**.

ST-DDT:
I meant more like `uz_UZ_latin`

Language_country_variant

matthewmayer:
uz_UZ_latin 

Seems the most correct. Uzbek, as spoken in Uzbekistan, with Latin variant. 

That would allow for say in future 

uz_UZ_cyrillic
uz_AF_latin
uz_RU_cyrillic

src/locales/uz_UZ/metadata.ts
```diff
@@ -0,0 +1,12 @@
+import type { PreBuiltMetadataDefinition } from '../../definitions/metadata';
+
+const metadata: PreBuiltMetadataDefinition = {
+  title: 'Uzbek (Latin)',
+  code: 'uz_UZ',
```
matthewmayer: Should also include `country: 'UZ'` here

src/locales/uz_UZ/person/bio_parts.ts
```diff
@@ -0,0 +1,37 @@
...
+  'mualim',
+  'muhandis',
+  'musiqachi',
+  'o‘ila odobi',
+  'o‘qituvchi',
+  'o‘yinchim',
+  'odam',
+  'qo‘shiqchi',
+  'rivojlanuvchi',
+  "san'atkor",
```
matthewmayer: Is there a difference between the two types of apostrophe? If not use the straight ' one everywhere.

matthewmayer (changes requested):

Mirazyzz:
> uz_UZ_latin
> 
> Seems the most correct. Uzbek, as spoken in Uzbekistan, with Latin variant.
> 
> That would allow for say in future
> 
> uz_UZ_cyrillic uz_AF_latin uz_RU_cyrillic

Just for clarity, the Uzbek language does not have dialectic variations such as American English, British English, or Australian English. The differences in Uzbek primarily lie within the alphabet utilized. Historically, we used the Cyrillic alphabet but transitioned to Latin over the past three decades, which is currently recognized as the official alphabet. Nevertheless, there remains a portion of the population, predominantly individuals over the age of 50-60, who only read in Cyrillic. The important point to note is that the grammar, words, and their meanings remain consistent regardless of the alphabet used.

Given this context, I think that it might be more reasonable to use codes "uz_latin" and "uz_cyrillic" rather than associating them with specific countries, since the language does not vary from one country to another. However, I'm open to different solutions just wanted to give more context.

src/locales/uz_UZ/person/bio_parts.ts
Mirazyzz: Yes, there is a difference between them one serves as a part of a letter while the other alters sound and has potential to appear in various positions within a word. However, I'm a little hesitant of using both of them even though it would be more grammatically correct, but the future contributions might not adhere to these nuances, resulting in the indiscriminate use of **'** as a substitute for both. Let me know what you think about this, please.

ST-DDT: If it is grammatically more correct (and not overly context sensitive), I think having both apostrophe variants is fine.
If future contributors, change that then we aren't any worse off than using only a single variant right now.
(And will likely spark a discussion, about the correct one again)

src/locales/uz_UZ/metadata.ts
Mirazyzz: https://github.com/faker-js/faker/pull/2686#issuecomment-1964635244

If we'll be using **uz_latin**, **uz_cyrillic** then it's not going to be **PreBuiltMetadataDefinitionForCountry** as far as I understand.

matthewmayer:
Adding the country code, in addition to the alphabet variant is more about indicating that say `faker.location.city()` would return cities in Uzbekistan.

src/locales/uz_UZ/metadata.ts
Mirazyzz: Done

Mirazyzz:
@matthewmayer, @ST-DDT Is there any additional changes I have to make in order be able to merge?

ST-DDT:
There are two things required.
- The team's decision on the locale id.
- The review of the actual words (if you know somebody that is familiar with the language and help review the names and approve the PR, that would help us a lot). Otherwise it might take a few more days before we have the time to check the names/words.

Mirazyzz:
> There are two things required.
> 
> * The team's decision on the locale id.
> * The review of the actual words (if you know somebody that is familiar with the language and help review the names and approve the PR, that would help us a lot). Otherwise it might take a few more days before we have the time to check the names/words.

Thanks for your response. Is a person to review the names/words should be from the Faker team or it could be any other contributor?

ST-DDT:
> Is a person to review the names/words should be from the Faker team or it could be any other contributor?

The Faker team has to review the files anyway. But any volunteers review helps.

ST-DDT:
**Team Decision**

Please use `uz_UZ_latin` to follow our [localization guide](https://fakerjs.dev/guide/localization.html#locale-codes).

We will extend the localization guide to include the representation `<ISO-639-1>[_<ISO-3166-1>[_<variant>]]` to better describe the intend behind our textual guideline.

We might change it in a future major version and then adjust all other locales as well.

Mirazyzz:
> **Team Decision**
> 
> Please use `uz_UZ_latin` to follow our [localization guide](https://fakerjs.dev/guide/localization.html#locale-codes).
> 
> We will extend the localization guide to include the representation `<ISO-639-1>[_<ISO-3166-1>[_<variant>]]` to better describe the intend behind our textual guideline.
> 
> We might change it in a future major version and then adjust all other locales as well.

Done.

src/locales/uz_UZ_latin/person/male_first_name.ts
```diff
@@ -0,0 +1,229 @@
...
+  'Fazliddin',
+  'Fahriddin',
+  'Fayzbek',
+  'Firdavs',
+  'Fitrat',
+  'Fotih',
+  "G'afur",
+  "G'affor",
+  "G'iyos",
+  "G'ulom",
```
ST-DDT, line 69: This seems to also refer to "slave". 🤔 
Is that a bad translation from Google?

src/locales/uz_UZ_latin/metadata.ts
```diff
@@ -2,7 +2,7 @@ import type { PreBuiltMetadataDefinitionForCountry } from '../../definitions/met
 
 const metadata: PreBuiltMetadataDefinitionForCountry = {
   title: 'Uzbek (Latin)',
```
matthewmayer: For consistency with sr_RS_latin the title should be "Uzbek (Uzbekistan, Latin)

matthewmayer (changes requested):

src/locales/uz_UZ_latin/person/male_first_name.ts
Mirazyzz: https://uz.wikipedia.org/wiki/G%CA%BBulom_Zafariy
https://en.wikipedia.org/wiki/G%CA%BBafur_G%CA%BBulom

These are all proper names. Probably bad translation from Google indeed.

src/locales/uz_UZ_latin/metadata.ts
Mirazyzz: Done

Mirazyzz:
@ST-DDT what can be the reason of a failing build? Looking at the log I don't think it's related to my changes, could you assist me, please?

ST-DDT (dismissed):

Shinigami92 (dismissed):

src/locales/uz_UZ_latin/metadata.ts
```diff
@@ -0,0 +1,13 @@
+import type { PreBuiltMetadataDefinitionForCountry } from '../../definitions/metadata';
+
+const metadata: PreBuiltMetadataDefinitionForCountry = {
+  title: 'Uzbek (Uzbekistan, Latin)',
+  code: 'uz_UZ_latin',
+  country: 'UZ',
+  language: 'uz',
+  endonym: "O'zbekcha",
```
Shinigami92, line 9: dont we need variant here like in https://github.com/faker-js/faker/blob/c52ec8ac431debace995a4b8cf51e586b64a6483/src/locales/sr_RS_latin/metadata.ts#L8C3-L8C19 ?

Mirazyzz: What is the usage of the **variant**?

ST-DDT: It is part of the locale's metadata that we expose. And yes we should add it.

Shinigami92 (approved):

matthewmayer (approved):

ST-DDT (approved):

PR #2700, opened 2024-03-01
feat(helpers): arrayElements returns a ts tuple when possible

### LINKED ISSUE #2675: helpers.arrayElements should return a tuple instead of an array
### Pre-Checks

### Describe the bug

`faker.helpers.arrayElements([1, 2, 3], 2)` returns a type of `number[]`, when it should be `[number, number]`

`faker.helpers.arrayElements([1, 2, 3], {min: 2, max: 10})` returns a type of `number[]`, when it should be `[number, number, ...number[]]`

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 21.50 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 8.14.1 - ~/.nvm/versions/node/v20.11.0/bin/pnpm
  Browsers:
    Chrome: 121.0.6167.184
    Safari: 17.2.1
  npmPackages:
    @faker-js/faker: ^8.4.1 => 8.4.1
```

- [ ] CJS
- [X] ESM

### Used Package Manager

pnpm

cloud-walker:
Follow up experiment after #2675 discussion

### DIFF
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -1034,6 +1034,10 @@ export class SimpleHelpersModule extends SimpleModuleBase {
       );
     }
 
+    if (array.length === 0) {
+      return [] as never;
+    }
+
     const numElements = this.rangeToNumber(
       count ?? { min: 1, max: array.length }
     );
--- a/src/modules/helpers/index.ts
+++ b/src/modules/helpers/index.ts
@@ -2,6 +2,7 @@ import type { Faker, SimpleFaker } from '../..';
 import { FakerError } from '../../errors/faker-error';
 import { deprecated } from '../../internal/deprecated';
 import { SimpleModuleBase } from '../../internal/module-base';
+import type { LooseTupleOf, TupleOf } from '../../utils/types';
 import { fakeEval } from './eval';
 import { luhnCheckValue } from './luhn-check';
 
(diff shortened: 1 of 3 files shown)


ST-DDT:
Is this ready for review? There are quite a few unexpected `never`s.

cloud-walker:
No is not ready I guess, but I wanted to publish it anyway to let you see the progress 🤔 

For the `as never` parts, I fear that they may be necessary, as TypeScript may not be smart enough to understand the dynamic nature of the return types. But I'll do some tries to avoid them

cloud-walker:
Made it a draft for now

cloud-walker:
if you wanna leave your impressions in the meanwhile I'll try to address them

ST-DDT (commented):
Please add a test, that uses a large count e.g. 100k. These kind of recursive types tend to error our on "deep" recursions.

ST-DDT:
**Team Decision**

We consider this a non goal because

- It can be easily achieved using casts `as [string, string]` which has significantly better performance than recursive types
- Or self crafted methods `return [value1(), value2()] as const`
- You might also want to generate tuples with mixed types `[number, string]` which we would consider dynamic complex objects and thus are not in the scope of faker

PR #2702, opened 2024-03-02
test(finance): fix tests for amount

ST-DDT:
This PR fixes a rare bug in the finance tests, that cause them to fail when amount returns `0`.

Previously the tests checked for `0 < amount < 1001`
Now the check looks like this: `0 <= amount <= 1000`

I also removed the dedicated `it returns a string` test in favor of more checks in the other tests.
And I removed some description texts, because they state the exact same as the check itself.


matthewmayer (approved):

ejcheng (approved):

Shinigami92:
@xDivisionByZerox I would like if you could take a look on this and approve accordingly
Because I'm not fully sure if 0 was an intended result or not

xDivisionByZerox (approved):
> @xDivisionByZerox I would like if you could take a look on this and approve accordingly
Because I'm not fully sure if 0 was an intended result or not

Since the implementation did not change, this looks valid to me. 

--- 

Additionally, I don't see why we should disallow 0 as return value. WHat would happen if the user provides 0 as min AND max value to the function?

PR #2704, opened 2024-03-03
refactor(date)!: remove v8 deprecated date methods

Shinigami92:


test/all-functional.spec.ts
```diff
@@ -49,6 +49,10 @@ const BROKEN_LOCALE_METHODS = {
     suffixes: ['az'],
     companySuffix: ['az'],
   },
+  date: {
+    between: '*',
+    betweens: '*',
+  },
```
ST-DDT, line 55: Why do these fail now? Depending on the reason, this might be worth noting in the migration guide.

Shinigami92: Something with they do not accept no-params anymore.
Would you like to investigate this a bit? I'm already confused what BROKEN_LOCALE_METHODS has to do with functions that does not need locales at all :thinking:   
https://github.com/faker-js/faker/pull/2704/files#diff-b13a3b00991d4ed1d9424f2262bf69d6d2cb6f304c13cc923ae3e207e8187faeR46-R47

matthewmayer: Previously if you called faker.helpers.between() or faker.helpers.betweens() with no parameters you'd get one or multiple of the current date (not especially helpful!)

Now it throws instead.

matthewmayer: This section of the tests is basically a sanity check that if you call every method in every locale with no parameters, it doesn't throw (with certain defined exceptions).

Shinigami92: > This section of the tests is basically a sanity check that if you call every method in every locale with no parameters, it doesn't throw (with certain defined exceptions).

So I call that a "never expected" anyway, and therefore also does not need to be mentioned in the migration guide

matthewmayer: Its not a particularly nice error:

```
> faker.date.betweens()   
Uncaught TypeError: Cannot destructure property 'from' of 'e' as it is undefined.
    at K.betweens (/Users/matt/Documents/git/faker/dist/chunk-EJUOGH2O.cjs:4:7483)
```

Perhaps this should work the same as 

```
> faker.date.betweens({})   
[
  2024-03-03T11:36:52.163Z,
  2024-03-03T11:36:52.163Z,
  2024-03-03T11:36:52.163Z
]
```

Shinigami92: This is an error, that can only be thrown in JavaScript when not respecting the TypeScript function contract / signature.

The later however returns 3 equal dates, which is more to be an issue and feels like you did something wrong and just can overlook it.

matthewmayer: i would suggest either:
* We treat missing options parameter the same as `{}` (note the docs do describe it as "The **optional** options object."
* We throw a more useful error, such as "You must pass at least a `from` or `to` parameter"

Shinigami92: As I do not have the time and will to discuss this right now, we will put it onto the agenda for next meeting

ST-DDT: IMO the docs are wrong anyway, because from and to are required.

As for throwing the error, I would just drop a comment, that 
"if you are a JS user, you are now getting an error.
TS users are safe"

matthewmayer: Are from and to expected to be required though? Or are they supposed to default to now/the default ref date?

ST-DDT: > Are from and to expected to be required though? Or are they supposed to default to now/the default ref date?

Yes, because if you default to now/refDate for now, then you will always get now. In that case anytime would be a better fit.
If you put only one in, then you get these weird edge cases, where `from`/`to` might be on the wrong side of now/refDate and you get an error. Here `past` or `future` would be a better fit.
(We might want to explicitly link to those methods for those usecases though)

We might want to actually remove the refDate fallbacks (that are mostly unreachable now):

https://github.com/faker-js/faker/blob/6360d94f030c7fec15f93b794ffb4de7963d6f19/src/modules/date/index.ts#L195-L196

e.g. `toDate(from, () => throw new Error("Invalid from Date"))`

(Or alter the toDate method to `undefined => fallback, return if good, otherwise throw`)

matthewmayer: Past/future only allow for N years in the future though right? If I want "some date in the future, before fixed date X"

Then should we still allow {to:X} without from?

ST-DDT: > Then should we still allow {to:X} without from?

What should `from` default to? Epoch? refDate?

matthewmayer: Given there were some syntaxes that allowed omittting from and to previously (e.g. `console.log(faker.date.between("2020-01-01"))` was allowed, even in Typescript), i think the cleanest solution is to 

1. Always require `from` and `to`
2. Throw an error if `from` and `to` are not provided, for the benefit of Javascript users or people upgrading from legacy syntax
3. Document that if you want to use "now" as a from or to, you should explicitly pass this, ie

`faker.date.between({from: "2000-01-01", to: Date.now()})`

matthewmayer: (Perhaps this should be done as a separate PR after this lands though).

ST-DDT: I created a separate issue for that discussion, so that this PR is no longer blocked.

- https://github.com/faker-js/faker/issues/2711

ST-DDT (approved):

matthewmayer (approved):

PR #2719, opened 2024-03-05
refactor(date)!: stricter error handling of between

### LINKED ISSUE #2711: Re-evaluate signature of faker.date.between(s)
Extracted from #2704 to avoid blocking the cleanup progress.

- https://github.com/faker-js/faker/pull/2704#discussion_r1510284472
- https://github.com/faker-js/faker/pull/2704#discussion_r1510546241

The following suggestions were made so far:

- Keep from and to as required options.
- Throw an error if from and to are missing for JS users.
- Remove hidden fallback to refDate
- Change `toDate` to fail, if the given values are passed but aren't valid Date-like values (This affects all date methods).
- Extend documentation regarding dates after/before a value. (Links to past, future, anytime)

matthewmayer:
fix #2711

Note i didn't change `toDate` function, as the new version of `between` doesn't need it any more. That could be done in a seperate PR if required.

Some errors are only possible to trigger in JS (e.g. omitting params) but invalid string dates can be passed in TS too.


src/modules/date/index.ts
```diff
@@ -192,10 +192,14 @@ export class SimpleDateModule extends SimpleModuleBase {
...
-    const dateOffset = this.faker.number.int(toMs - fromMs);
+    const fromMs = new Date(from).getTime();
+    const toMs = new Date(to).getTime();
+
+    if (Number.isNaN(from) || Number.isNaN(to)) {
+      throw new FakerError('Both from and to are required.');
+    }
 
+    const dateOffset = this.faker.number.int(toMs - fromMs);
     return new Date(fromMs + dateOffset);
```
ST-DDT: IMO we could merge these lines into one/~or simplify it~:
```suggestion
    return new Date(this.faker.number.int({ from: fromMs, to: toMs }));
```

src/modules/date/index.ts
```diff
@@ -175,10 +175,14 @@ export class SimpleDateModule extends SimpleModuleBase {
...
    * @param options.to The late date boundary.
    *
+   * @throws If `from` or `to` are not provided.
+   * @throws If `from` is after `to`.
+   *
    * @example
    * faker.date.between({ from: '2020-01-01T00:00:00.000Z', to: '2030-01-01T00:00:00.000Z' }) // '2026-05-16T02:22:53.002Z'
    *
    * @since 8.0.0
+   *
```
ST-DDT:
```suggestion
```

src/modules/date/index.ts
```diff
@@ -190,13 +194,23 @@ export class SimpleDateModule extends SimpleModuleBase {
      */
     to: string | Date | number;
   }): Date {
+    if (!options) {
+      throw new FakerError('Both from and to must be valid dates.');
+    }
```
ST-DDT: Not sure about this one. Maybe temporary for v9 (with todo to remove in v10)

src/modules/date/index.ts
```diff
@@ -190,13 +194,23 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+
     const { from, to } = options;
+    const fromMs = new Date(from).getTime();
+    const toMs = new Date(to).getTime();
+
+    if (Number.isNaN(fromMs) || Number.isNaN(toMs)) {
+      throw new FakerError('Both from and to must be valid dates.');
```
ST-DDT: Maybe quote the variables?
```suggestion
      throw new FakerError('Both `from` and `to` must be valid dates.');
```

src/modules/date/index.ts
```diff
@@ -190,13 +194,23 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+    }
 
-    const fromMs = toDate(from, this.faker.defaultRefDate).getTime();
-    const toMs = toDate(to, this.faker.defaultRefDate).getTime();
-    const dateOffset = this.faker.number.int(toMs - fromMs);
+    if (fromMs > toMs) {
+      throw new FakerError('From date must be before to date.');
```
ST-DDT: Same here

```suggestion
      throw new FakerError('`from` date must be before `to` date.');
```

src/modules/date/index.ts
```diff
@@ -252,8 +269,11 @@ export class SimpleDateModule extends SimpleModuleBase {
           max: number;
         };
   }): Date[] {
-    const { from, to, count = 3 } = options;
+    if (!options) {
+      throw new FakerError('Both from and to must be valid dates.');
+    }
```
ST-DDT: Not sure about this one. Maybe temporary for v9 (with todo to remove in v10)

src/modules/date/index.ts
matthewmayer: This a courtesy for JS developers. Given that nearly every faker method works with no parameters I think it's worth having a nice error for the one that doesn't to point people in the right direction.

ST-DDT: The error should at least mention the actual options parameter that is actually missing.

matthewmayer: updated error message

src/modules/date/index.ts
matthewmayer: updated error message

ST-DDT: We created a separate issue to discuss type-less JavaScript support:

- https://github.com/faker-js/faker/issues/2725

ST-DDT:
Blocked by https://github.com/faker-js/faker/issues/2725

- https://github.com/faker-js/faker/issues/2725

ST-DDT (dismissed):

src/modules/date/index.ts
```diff
@@ -207,6 +222,9 @@ export class SimpleDateModule extends SimpleModuleBase {
    * @param options.to The late date boundary.
```
ST-DDT, line 203: Please also adjust the jsdocs:

> The optional options object.

src/modules/date/index.ts
```diff
@@ -175,6 +175,9 @@ export class SimpleDateModule extends SimpleModuleBase {
    * @param options.from The early date boundary.
```
ST-DDT, line 159: Here as well.

ST-DDT (dismissed):

ST-DDT:
I'm in contact with the codecov team regarding the increased error rate.

matthewmayer:
I guess this should revert back to using the new improved "toDate" now that has been merged?

ST-DDT:
Yes, please.

src/modules/date/index.ts
```diff
@@ -174,22 +177,34 @@ export class SimpleDateModule extends SimpleModuleBase {
      */
     to: string | Date | number;
   }): Date {
+    if (!options || (!options.from && !options.to)) {
```
ST-DDT: Maybe:

```suggestion
    if (!options || !options.from || !options.to) {
```

src/modules/date/index.ts
```diff
@@ -235,8 +250,13 @@ export class SimpleDateModule extends SimpleModuleBase {
           max: number;
         };
   }): Date[] {
-    const { from, to, count = 3 } = options;
+    if (!options || !(options.from && options.to)) {
```
ST-DDT: Maybe:

```suggestion
    if (!options || !options.from || !options.to) {
```

src/modules/date/index.ts
ST-DDT: Alternatively we could sneakily default `from` and `to` to `NaN`/`'not-provided'` and let the `toDate` error handling handle this.

src/modules/date/index.ts
```diff
@@ -174,22 +177,34 @@ export class SimpleDateModule extends SimpleModuleBase {
      */
     to: string | Date | number;
   }): Date {
+    if (!options || !options.from || !options.to) {
```
ST-DDT: I just noticed that this might cause issues with UTC epoch/`0`.

So maybe we should use a nullish check instead?

```suggestion
    if (!options || options.from == null || options.to == null) {
```
Or
```suggestion
    if (options == null || options.from == null || options.to == null) {
```

Sorry for the extra work.

matthewmayer: did this, and added an extra test for date 0

ST-DDT (dismissed):

src/modules/date/index.ts
```diff
@@ -174,22 +177,34 @@ export class SimpleDateModule extends SimpleModuleBase {
      */
     to: string | Date | number;
   }): Date {
+    if (options == null || options.from == null || options.to == null) {
+      throw new FakerError(
```
Shinigami92, line 182: Should this be marked with a TODO to be removed in v10?
I assume we don't need to check something that is covered by Typescript, at least not in the long run.

src/modules/date/index.ts
```diff
@@ -235,8 +250,13 @@ export class SimpleDateModule extends SimpleModuleBase {
           max: number;
         };
   }): Date[] {
-    const { from, to, count = 3 } = options;
+    if (options == null || options.from == null || options.to == null) {
+      throw new FakerError(
```
Shinigami92, line 256: Same here

src/modules/date/index.ts
matthewmayer: what was the conclusion of https://github.com/faker-js/faker/issues/2725 ? should this kind of usage be flagged with a TODO?

ST-DDT: TLDR: Only check for stuff not handled by TS, except if it is a breaking change and then only for 1 major version.

ST-DDT: Please use our usual todo syntax:

https://github.com/faker-js/faker/blob/7f11bf2d37fc5e8e5cc6bf7a90ea6789ef3bb1d4/scripts/apidoc/typedoc.ts#L146

ST-DDT (approved):

Shinigami92 (approved):

PR #2720, opened 2024-03-05
refactor(finance)!: rewrite creditCardNumber to work with issuer enum

### LINKED ISSUE #181: Credit Card number parameter
Screen Shot 2022-01-16 at 2 35 07 PM

In the part of the Finance section of the API docs shown above, I think it would improve the docs if we listed all possible inputs of the 'provider' parameter, such as 'Visa', 'MasterCard', etc. What would happen if someone used gibberish as a parameter there, would an error be thrown?

ST-DDT:
Fixes #181 (our oldest open issue)

- #181

(I removed the jsdocs from the implementation signature, as it isn't visible to our users anyway.)


src/modules/finance/index.ts
```diff
@@ -746,7 +746,16 @@ export class FinanceModule extends ModuleBase {
   /**
    * Generates a random credit card number.
    *
+   * If you need to generate a credit card number for a specific issuer, you can pass the issuer name as a string.
+   * The issuer name is case-insensitive.
+   * A list of supported issuers can be found in the `faker.definitions.finance.credit_card` object.
+   * You can access that value at runtime directly or from the sources on GitHub (`src/locales/<name>/finance/credit_card`).
```
matthewmayer: would it be easier to just directly list them out here? there aren't many

matthewmayer (changes requested):

src/modules/finance/index.ts
ST-DDT: Changed

Shinigami92 (dismissed):

src/modules/finance/index.ts
xDivisionByZerox: > would it be easier to just directly list them out here? there aren't many

Not a fan of this. This means we have multiple sources of truth. Additionally, if they aren't that many to list them out, why not create a separate string union entirely?

ST-DDT: I also considered that, but that would be breaking the implementation (which I feel tempted to do anyway).

I thought about using strictly

````ts
creditCardNumber(options: { issuer: CreditcardIssuer }} // without the current wildcard format support
````

Should we discuss this in the next team meeting as well?

Comment (matthewmayer) [reply src/modules/finance/index.ts]: i think the improvement to developer experience outweighs the fact that there are multiple sources of truth. New card issuers would only appear extremely rarely (who is going to launch a visa/mastercard competitor in 2024?)

Comment (xDivisionByZerox) [reply src/modules/finance/index.ts]: > i think the improvement to developer experience outweighs the fact that there are multiple sources of truth.

Despite my previous comment, I'm with you on this one 👍 

 > New card issuers would only appear extremely rarely (who is going to launch a visa/mastercard competitor in 2024?)

This makes me want to create a `CreditCardIssuer` type even more TBH.  
To not keep blocking this PR by this discussion I'd say that the additional type can be discussed in a separate issue.

**Comment by ST-DDT** [comment]
**Team Decision**

* We will use a literal CreditCardIssuer enum.
* To check, whether the credit card patterns contain localized parts.
  * If not, remove localized patterns and only use localized data to identify which issuer is preferred in a country

````ts
creditCardNumber(options: { issuer?: CreditCardIssuer } = {}) {
  const { issuer = (weighted)arrayElement(defintions.finance.preferred_credit_card_issuer) } = options;

  switch (issuer) {
   case issuer1: ...;
   case issuer2: ...;
}
````

File: src/modules/finance/index.ts (hunk 1 above)
Comment (matthewmayer) [reply src/modules/finance/index.ts]: as currently issuer names are allowed to be any case, perhaps we should start logging a warning if the issuer name is not lowercase? That will make migration easier when this becomes an enum.

Comment (ST-DDT) [reply src/modules/finance/index.ts]: I might just throw an error, because custom format patterns will be dropped as well.

Comment (matthewmayer) [reply src/modules/finance/index.ts]: Could throw a warning in v9 and make it an enum in v10

**Comment by ST-DDT** [comment]
Which variant would you prefer?

````ts
// 1A) Using our "options" syntax?
creditCardNumber(options: { issuer?: CreditCardIssuerType } = {})
creditCardNumber({ issuer: CreditCardIssuer.Visa })
````

or

````ts
// 1B) Like person.firstName
creditCardNumber(issuer: CreditCardIssuerType = this.creditCardIssuer())
creditCardNumber(CreditCardIssuer.Visa)
````

**Comment by ST-DDT** [comment]
Should we allow passing array's of allowed issuers?
And how so?

````ts
// 2A) Forbid
... issuer: CreditCardIssuerType ...
````
or
````ts
// 2B) Allow
... issuer: CreditCardIssuerType | CreditCardIssuerType[] ...
````
or
````ts
// 2C) Require
... issuer: CreditCardIssuerType[] ...
````

File: test/modules/finance.spec.ts [hunk 2]
```diff
@@ -349,7 +355,9 @@ describe('finance', () => {
         });
 
         it('should return a valid credit card number', () => {
-          expect(faker.finance.creditCardNumber('')).toSatisfy(luhnCheck);
+          expect(
```
ST-DDT, line 357:
```suggestion
          // Deprecated
          expect(
```

src/modules/finance/index.ts
```diff
@@ -529,111 +546,148 @@ export class FinanceModule extends ModuleBase {
...
+
+      case CreditCardIssuer.Visa: {
+        format = this.faker.helpers.arrayElement([
+          '4###########L',
+          '4###-####-####-###L',
+        ]);
+        break;
+      }
+
+      default: {
```
ST-DDT, line 679: This should throw an error (due to overly exhaustive switch): https://github.com/typescript-eslint/typescript-eslint/issues/8645

- https://github.com/typescript-eslint/typescript-eslint/issues/8645

matthewmayer:
This should be marked breaking no？ it changes the behavior when a unknown or non lower case issuer is passed.

src/modules/finance/index.ts
```diff
@@ -529,111 +546,148 @@ export class FinanceModule extends ModuleBase {
...
-      options = { issuer: options };
+    if (typeof issuer !== 'string') {
+      issuer = issuer.issuer ?? this.creditCardIssuer();
+      deprecated({
+        deprecated: 'faker.finance.creditCardNumber({ issuer })',
+        proposed: 'faker.finance.creditCardNumber(issuer)',
+        since: '9.0',
+        until: '10.0',
+      });
     }
```
xDivisionByZerox, line 576: Do we really want to deprecate this? "quick options" are a part of fakers signatures in a lot of places. 🤔

test/modules/finance.spec.ts
```diff
@@ -76,8 +77,13 @@ describe('finance', () => {
...
         .it('with issuer', 'visa')
-        .it('with issuer option visa', { issuer: 'visa' })
-        .it('with issuer option mastercard', { issuer: 'mastercard' });
+        // Deprecated
+        .it('with issuer option visa', {
+          issuer: 'visa',
+        } as unknown as CreditCardIssuerType)
+        .it('with issuer option mastercard', {
+          issuer: 'mastercard',
+        } as unknown as CreditCardIssuerType);
```
xDivisionByZerox, line 86: Why are these deprecated? `'visa' | 'mastercard'` should be included in `CreditCardIssuerType`. What am I missing?

src/locales/el/finance/credit_card/maestro.ts
```diff
@@ -1 +0,0 @@
-export default ['/50#{9,16}L/', '/5[6-8]#{9,16}L/', '/56##{9,16}L/'];
```
xDivisionByZerox: These patterns where actually different from the other en `maestro` ones 🤔

src/modules/finance/index.ts
```diff
@@ -529,111 +546,148 @@ export class FinanceModule extends ModuleBase {
...
-   * faker.finance.creditCardNumber({ issuer: 'visa' }) // '4882664999007'
-   * faker.finance.creditCardNumber({ issuer: '63[7-9]#-####-####-###L' }) // '6375-3265-4676-6646'
-   * faker.finance.creditCardNumber('visa') // '1226423499765'
+   * faker.finance.creditCardNumber('visa') // '4882664999007'
    *
-   * @since 5.0.0
+   * @since 9.0.0
    */
+  creditCardNumber(issuer?: CreditCardIssuerType): string;
   creditCardNumber(
```
xDivisionByZerox, line 561: Is the JSDoc not required here?

src/modules/finance/index.ts
```diff
@@ -529,111 +546,148 @@ export class FinanceModule extends ModuleBase {
...
+      });
     }
 
-    const { issuer = '' } = options;
+    // TODO @ST-DDT 2024-03-09: Remove the fallback in v10
+    issuer ||= this.creditCardIssuer();
+    const actualIssuer = issuer.toLowerCase() as CreditCardIssuerType;
+    if (actualIssuer !== issuer) {
+      deprecated({
+        deprecated: 'faker.finance.creditCardNumber(iSsUeR)',
```
Shinigami92, line 583: This looks absolutely weird to me :eyes:

src/modules/finance/index.ts
```diff
@@ -22,6 +23,22 @@ export interface Currency {
   symbol: string;
 }
 
+/**
+ * The supported credit card issuers.
+ */
+export enum CreditCardIssuer {
+  AmericanExpress = 'american_express',
+  DinersClub = 'diners_club',
```
Shinigami92, line 31: Do we have any conventions yet? Otherwise I would be in favor of using `-` instead of `_` for the string values.

src/modules/finance/index.ts
```diff
@@ -529,111 +546,148 @@ export class FinanceModule extends ModuleBase {
...
-      // The user chose an optional scheme
-      format = issuer;
-    } else {
-      // Choose a random issuer
-      // Credit cards are in an object structure
-      const formats = this.faker.helpers.objectValue(localeFormat); // There could be multiple formats
-      format = this.faker.helpers.arrayElement(formats);
+
+    switch (actualIssuer) {
+      case CreditCardIssuer.AmericanExpress: {
```
Shinigami92, line 593: This wont work when the string value of the enum contains a separator like `_` or `-`, but the `actuallIssuer` is just lowercased in [line 580](https://github.com/faker-js/faker/pull/2720/files#diff-ed919e84dad7aa63abddfc6222d0c69df214a632cd29cbce6866b9a0f968d618R580).

src/modules/finance/index.ts
ST-DDT: Any idea how to show case insesitive is deprecated?

src/modules/finance/index.ts
ST-DDT: > Do we have any conventions yet?

No. I just used what we use/have already.

src/modules/finance/index.ts
ST-DDT: Why wouldnt it? The test seem to be working fine.

Shinigami92: Ah I see, its because `issuer` is now `CreditCardIssuerType` strictly typed instead of a `string`. But why do we need the `toLowerCase` then in the first place?

ST-DDT: Backwards compatibility.

Shinigami92: Is this a discussion about #2725 again? Otherwise the type def is wrong, because arbitrary strings  are not possible anymore.

src/modules/finance/index.ts
ST-DDT: That is the implementation signature. It isn't visible to users.

src/modules/finance/index.ts
ST-DDT: > Is this a discussion about #2725 again?

In the sense, that we follow our usual deprecation steps: Yes

That's why the following comment is there:

> // TODO @ST-DDT 2024-03-09: Remove the fallback in v10

matthewmayer: Wouldn't it be better to do this in two stages? In 9.0 log a runtime warning if an unknown or mixed case issuer is passed. Then in 10.0 switch to the enum.

ST-DDT: 🤔 Not sure about that.
JS users already have the backwards compatibility code.
TS users would have more time to migrate at the cost of runtime warnings instead of compile time errors.

src/modules/finance/index.ts
matthewmayer: what about 
```
 deprecated: `faker.finance.creditCardNumber('${issuer}')`,
 proposed: `faker.finance.creditCardNumber('${actualIssuer}')`,
```

xDivisionByZerox: Or maybe:

```suggestion
        deprecated: 'faker.finance.creditCardNumber(issuerWithRandomCase)',
```

src/modules/finance/index.ts
xDivisionByZerox: Since this change is breaking anyway, we might want to standardize these cases. I'm not that much in favor of any case, but would like to have a "rule" for these types.

PR #2721, opened 2024-03-05
infra(unicorn): switch-case-braces

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/switch-case-braces`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/switch-case-braces.md) lint rule.


ST-DDT:
I'm totally unbiased regarding this rule. I just enabled it because it is the default.
I also noticed, that we do actually have both variants (with and without braces) in our code/sometimes even in the same method.

Shinigami92 (approved):
I started some month ago to prefer this rule (default) as well

matthewmayer (approved):

xDivisionByZerox (approved):

PR #2722, opened 2024-03-05
infra(unicorn): no-array-callback-reference

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/no-array-callback-reference`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-callback-reference.md) lint rule.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -38,6 +38,7 @@ module.exports = defineConfig({
     'prefer-exponentiation-operator': 'error',
     'prefer-template': 'error',
 
+    'unicorn/no-array-callback-reference': 'off', // reduces readability
     'unicorn/no-nested-ternary': 'off', // incompatible with prettier
     'unicorn/no-null': 'off', // incompatible with TypeScript
     'unicorn/no-zero-fractions': 'off', // deactivated to raise awareness of floating operations
@@ -49,7 +50,6 @@ module.exports = defineConfig({
     'unicorn/better-regex': 'off',
     'unicorn/consistent-function-scoping': 'off',
     'unicorn/import-style': 'off',
-    'unicorn/no-array-callback-reference': 'off',
     'unicorn/no-await-expression-member': 'off',
     'unicorn/no-object-as-default-parameter': 'off',
     'unicorn/numeric-separators-style': 'off',


ST-DDT:
This should hopefully prevent issues such as #2563 in the future.

- #2563

Shinigami92 (dismissed):

.eslintrc.cjs
```diff
@@ -49,7 +49,6 @@ module.exports = defineConfig({
     'unicorn/better-regex': 'off',
     'unicorn/consistent-function-scoping': 'off',
     'unicorn/import-style': 'off',
-    'unicorn/no-array-callback-reference': 'off',
```
xDivisionByZerox, line 52: On a more general note: This rule seems to actively contribute towards prohibiting the pattern of writing custom type guards. Would that be our responsibility to figure out or do you think it would be clever to move the discussion to the unicorn repository by opening an issue there?

test/all-functional.spec.ts
```diff
@@ -14,6 +14,7 @@ const IGNORED_MODULES = new Set([
 function getMethodNamesByModules(faker: Faker): { [module: string]: string[] } {
   return Object.fromEntries(
     Object.keys(faker)
+      // eslint-disable-next-line unicorn/no-array-callback-reference -- we need this for type narrowing
       .filter(isTestableModule)
```
xDivisionByZerox: Does moving the type guard to the inline return type not work with this rule? I don't see why this should be an exception if the rule doesn't allow any.

```suggestion
      .filter((v): v is keyof Faker => isTestableModule(v))
```

.eslintrc.cjs
ST-DDT: I'm not sure. Maybe we could argue with unicorn to add an option to allow methods that take exactly one argument?

test/all-functional.spec.ts
```diff
@@ -14,7 +14,7 @@ const IGNORED_MODULES = new Set([
 function getMethodNamesByModules(faker: Faker): { [module: string]: string[] } {
   return Object.fromEntries(
     Object.keys(faker)
-      .filter(isTestableModule)
+      .filter((v): v is keyof Faker => isTestableModule(v))
```
matthewmayer: i dont really like this rule as it makes the code harder to read. The intent of original can be understood at a glance, even if the newer version is theoretically more type-safe.

Shinigami92 (dismissed):

ST-DDT:
I agree with the comments from @xDivisionByZerox and @matthewmayer here.
Enabling this rule harms code readability/adds unnecessary boilerplate to it.
I also considered adding an option to it, allowing for single argument parameters, but methods can be re-assigned without optional parameters without actually stripping them, so this would break the intention of the rule:

````
const fn: (arg1): T = (arg1, arg2?): T => { ...}
// fn still takes the second argument, although sneakily
return array.map(fn);
// would not show an error because the method type only takes a single argument
````

For this reason, this PR disables the rule permanently.

xDivisionByZerox (approved):

matthewmayer (approved):

ejcheng (approved):

Issue #1574, opened 2022-11-20
Rename company.suffix definition to company.legal_entity_type


ST-DDT:
### Clear and concise description of the problem

The company suffixes currently refer to the company/business entity type (not sure about this term) and not a real suffix in the usual sense.

### Suggested solution

We should rename the `suffix` to `type` (both the definitions and method).

### Additional context

Discovered here: https://github.com/faker-js/faker/pull/1573

ST-DDT:
Blocked by https://github.com/faker-js/faker/pull/1575

xDivisionByZerox:
This is not blocking for `v8`. If we had found this earlier and deprecated in `v7.6` that would be something else. But in the current state, we can deprecated (and add the desired function) at any point during `v8.x`.

xDivisionByZerox:
Regarding term: The german locale has a [legal_form](https://github.com/faker-js/faker/blob/1cadfb97856da17047fcb5dcb504c3418eeed145/src/locales/de/company/legal_form.ts) file. This could be a good name.

matthewmayer:
The method is already deprecated. So this is just about how to name the definition file? Seems that doesn't really matter (and could even be locale dependant, for example if a locale typically has prefixes instead of suffixes for companies)

xDivisionByZerox:
Probably also for later feature implementation. 
`company.suffix()` was simply not clear enough what its intent was. It could be implemented in a later version with a more descriptive name.
But for now only the definition file name.

These are my thoughts on this.

matthewmayer:
FWIW Wikipedia calls these "Legal entity types" https://en.wikipedia.org/wiki/List_of_legal_entity_types_by_country

xDivisionByZerox:
Discussion: What should the locale file be exactly renamed to? This decision should br made in a team meeting and should consider all already provided names.

ST-DDT:
**Team Decision**

- We will rename `suffix` to `legal_entity_type`.

ST-DDT:
This only affects the locale data/definitions -> v9.0 only.

RacketyWater7:
I intend to take this issue as my first; but need a little heads up.

ST-DDT:
> I intend to take this issue as my first; but need a little heads up.

Welcome to the project. What kind of heads up do you need?

ST-DDT:
@RacketyWater7 We would like to include this in v9.0.
Do you have time to tackle this in this month?

cakuki:
[...]

PR #2726, opened 2024-03-08
refactor(company)!: remove v8 deprecated company methods

Shinigami92:


ST-DDT (dismissed):

docs/guide/upgrading_v9/2726.md
```diff
@@ -0,0 +1,12 @@
+### Remove deprecated company methods
+
+Removed deprecated company methods
+
+| old                           | replacement                   |
+| ----------------------------- | ----------------------------- |
+| `faker.company.suffixes`      | `faker.company.name`          |
+| `faker.company.companySuffix` | `faker.company.name`          |
```
matthewmayer: these are not really direct replacements. Maybe write "Part of `faker.company.name` instead?

matthewmayer (changes requested):

ST-DDT (approved):

matthewmayer (approved):

docs/guide/upgrading_v9/2726.md
xDivisionByZerox, line 1: This file is missing the access to the definitions property: `faker.definitions.company.suffixes` or `faker.rawDefinitions.company.suffixes`.

src/definitions/company.ts
```diff
@@ -38,11 +38,4 @@ export type CompanyDefinition = LocaleEntry<{
    * Catchphrase nouns that can be displayed to an end user.
    */
   noun: string[];
-
-  /**
-   * Company/Business entity types.
-   *
-   * @deprecated Use `faker.company.name` instead.
-   */
-  suffix: string[];
```
xDivisionByZerox, line 47: According locale files should be removed.

xDivisionByZerox (changes requested):

src/definitions/company.ts
ST-DDT: They are still used in the company name patterns and will be renamed to [`legal_entity_type`](https://github.com/faker-js/faker/issues/1574) (eventually).

xDivisionByZerox: Simply renaming will not work, as a lot of these files contain the value `null` (not applicable). Without having thoroughly checked this, I can't believe that this will be the actual value ending up in the file for all these locales.

---

But you are correct, removing the files is not required for this PR.

docs/guide/upgrading_v9/2726.md
matthewmayer: i dont think we should encourage accessing the definitions directly, that is prone to breaking, and that definition is likely to get renamed soon anyway https://github.com/faker-js/faker/issues/1574

xDivisionByZerox: Then please explain to me why the following PRs were/are considered breaking and (should) have listed the changes of definition files in their migration guide?

- #2476
- #2505
- #2033

matthewmayer: In an ideal world, consumers would only be using public methods. Using raw definitions is a bit hacky and prone to breaking. So if we remove a public method for say "company suffix", if someone is migrating code that uses this, it would be probably better for them to just hardcode it as say `faker.helpers.arrayElement(['Inc', 'and Sons', 'LLC', 'Group'])` than introduce hacky code which relies on the raw definitions.

By contrast if you had code affected by #2476, #2505, #2033 then you were *already* using the hacky raw definitions, so we're not making your code worse by helping you migrate to the new locations of the definitions.

That said I'd be happy to omit some of these "definition only" migration guide snippets or just make them extremely brief.

ST-DDT: I think the main difference here is that the data weren't changed/moved.
We dropped support for having a method to access the data, but the data remain unchanged.
Maybe we should rename the method/data directly instead of removing the method and potentially adding it back with #1574 ? IMO we should not suggest the user to access the raw definitions here. 🤔 

- #1574

matthewmayer: > I think the main difference here is that the data weren't changed/moved.
> 
> We dropped support for having a method to access the data, but the data remain unchanged.
> 
> Maybe we should rename the method/data directly instead of removing the method and potentially adding it back with #1574 ? IMO we should not suggest the user to access the raw definitions here. 🤔 
> 
> 
> 
> - #1574 

If we think there's a need for company.legalEntityType() we could expose that yes. I'm not really sure why you'd want to use it in its own though.

xDivisionByZerox: Alright, sorry for my confusion. I thought that by removing the property from the definitions type, our build script would omit these files. That's my bad. This comment can be resolved.

xDivisionByZerox (approved):

PR #2727, opened 2024-03-09
refactor(finance)!: remove v8 deprecated finance methods

Shinigami92:

### DIFF
--- a/test/modules/finance.spec.ts
+++ b/test/modules/finance.spec.ts
@@ -193,17 +193,6 @@ describe('finance', () => {
           expect(+amount).toBeLessThanOrEqual(1000);
         });
 
-        // TODO @Shinigami92 2024-03-09: is this whole test wrong?
-        // duplicate of line 232
-        it('should use the default decimal location when not passing arguments', () => {
-          let amount = faker.finance.amount();
-
-          amount = faker.finance.amount({ min: 100, max: 100, dec: 1 });
-
-          expect(amount).toBeTruthy();
-          expect(amount).toBe('100.0');
-        });
-
         //TODO: add support for more currency and decimal options
         it('should not include a currency symbol by default', () => {
           const amount = faker.finance.amount();
--- a/test/modules/finance.spec.ts
+++ b/test/modules/finance.spec.ts
@@ -238,7 +238,6 @@ describe('finance', () => {
               min: number,
               max: number,
               dec: 0,
-              symbol: undefined,
               autoFormat,
             });
 
@@ -257,7 +256,6 @@ describe('finance', () => {
             min: number,
             max: number,
             dec: decimalPlaces,
-            symbol: undefined,
             autoFormat: true,
           });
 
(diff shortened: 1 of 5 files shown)


test/modules/finance.spec.ts
```diff
@@ -264,10 +193,12 @@ describe('finance', () => {
...
+        // TODO @Shinigami92 2024-03-09: is this whole test wrong?
+        // duplicate of line 232
         it('should use the default decimal location when not passing arguments', () => {
           let amount = faker.finance.amount();
 
-          amount = faker.finance.amount(100, 100, 1);
+          amount = faker.finance.amount({ min: 100, max: 100, dec: 1 });
 
           expect(amount).toBeTruthy();
           expect(amount).toBe('100.0');
```
Shinigami92: This test looks like a duplicate of https://github.com/faker-js/faker/pull/2727/files#diff-0d257ccb61186c82bc52c3f738b1fe2e1815eda321f5598010abdcee3b865afeR229 to me.

@ST-DDT is there something wrong here? Should I remove one? Should I rewrite one?

test/modules/finance.spec.ts
```diff
@@ -314,13 +245,13 @@ describe('finance', () => {
...
-              number,
-              0,
-              undefined,
-              autoFormat
-            );
+            const amount = faker.finance.amount({
+              min: number,
+              max: number,
+              dec: 0,
+              symbol: undefined,
```
Shinigami92: `symbol` might be omitted?

test/modules/finance.spec.ts
```diff
@@ -333,13 +264,13 @@ describe('finance', () => {
...
-            number,
-            decimalPlaces,
-            undefined,
-            true
-          );
+          const amount = faker.finance.amount({
+            min: number,
+            max: number,
+            dec: decimalPlaces,
+            symbol: undefined,
```
Shinigami92: `symbol` might be omitted?

test/modules/finance.spec.ts
ST-DDT: This test is broken.

It calls amount twice and discards the old result.

test/modules/finance.spec.ts
ST-DDT: Yes, previously it was required because of how the parameters are ordered.

test/modules/finance.spec.ts
ST-DDT: I think the original intention behind the test is to test for default `dec=1`? Maybe you can add that to the above test with default parameters?

Shinigami92: something like this? https://github.com/faker-js/faker/pull/2727/commits/ee59bcd66362697d576ad52b64265aa5b6398337

ST-DDT (approved):

test/modules/finance.spec.ts
ST-DDT: I thought of extending the existing test, but this works as well.

matthewmayer (approved):

xDivisionByZerox (approved):

PR #2747, opened 2024-03-13
docs: improve contributing guide

ST-DDT:
Created after reading our contribution guide again for https://github.com/faker-js/faker/discussions/2746

- https://github.com/faker-js/faker/discussions/2746


CONTRIBUTING.md
```diff
@@ -1,8 +1,14 @@
 A lot of effort has been put into `Faker` to create a useful and handy library.
 There are still a lot of things to be done, so all contributions are welcome!
-If you want to make `Faker` a better, please read the following contribution guide.
+If you want to make `Faker` a better place, please read the following contribution guide.
 
-# Important
+## Before you start
+
+- If you would like to propose a new feature, create an issue first.
+- Bug fixes or locale changes don't need an issue.
```
xDivisionByZerox: While I agree on locale changes not requiring an issue necessarily, I don't think the same should go for bugs. Having an issue with reproduction steps (as written in the next point) is super valuable.  
Additionally, there should be a verification whether said bug is actually a bug and not the code working as expected. This can prevent situations were someone might say "it does(n't) work on my machine".  
Lastly, an issue can better track acceptance criteria on the expected behaviour of the code as it doesn't (always) have to bother with actual implementation details.

ST-DDT: https://github.com/faker-js/faker/pull/2747/files#r1524020421

> While I agree on locale changes not requiring an issue necessarily, I don't think the same should go for bugs. Having an issue with reproduction steps (as written in the next point) is super valuable.
> Additionally, there should be a verification whether said bug is actually a bug and not the code working as expected. This can prevent situations were someone might say "it does(n't) work on my machine".
> Lastly, an issue can better track acceptance criteria on the expected behaviour of the code as it doesn't (always) have to bother with actual implementation details.

Like this?

```suggestion
- If you would like to fix a bug or propose a new feature, create an issue first.
- Locale changes don't need an issue.
```

CONTRIBUTING.md
```diff
@@ -1,8 +1,14 @@
 A lot of effort has been put into `Faker` to create a useful and handy library.
 There are still a lot of things to be done, so all contributions are welcome!
-If you want to make `Faker` a better, please read the following contribution guide.
+If you want to make `Faker` a better place, please read the following contribution guide.
 
-# Important
+## Before you start
+
+- If you would like to propose a new feature, create an issue first.
```
matthewmayer: maybe worth mentioning here that we usually wait until an issue for a new feature has some upvotes before starting development work on it?

matthewmayer (changes requested):

CONTRIBUTING.md
matthewmayer: Maybe we could rewrite it slightly more like:

* It's generally helpful to create an issue first:
  - if you are proposing a new feature, this allows other users to "upvote" the issue. Once an issue has enough upvotes (usually 10+) it will be reviewed for development.
  - if you notice a bug, this allows you to provide steps to reproduce, and allows other users to confirm this is actually a bug
  - it's not required to create an issue in all cases. For example for fixing a typo in documentation, or adding some new data for a locale, you could immediately create a pull request without an issue.

CONTRIBUTING.md
```diff
@@ -1,8 +1,18 @@
 A lot of effort has been put into `Faker` to create a useful and handy library.
 There are still a lot of things to be done, so all contributions are welcome!
-If you want to make `Faker` a better, please read the following contribution guide.
+If you want to make `Faker` a better place, please read the following contribution guide.
 
-# Important
+## Before you start
+
+It's generally helpful to create an issue first:
```
matthewmayer: Could "create an issue" be hyperlinked to the create issue page on GitHub?

matthewmayer (changes requested):

matthewmayer (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #2751, opened 2024-03-15
refactor(word): reduce definitions to 1000 in all locales

matthewmayer:
## Rationale

Reduced to 1000 entries max in the `words` module definitions in all locales.

I selected this module because it had the biggest impact on bundle size (e.g. the `hu` noun file alone is approx 144KB, 10K rows.) while still being relatively easy to review.

This is a stepping stone to enabling the `normalizeLocaleFile` for all files (previously I noted that I didn't think it was a good idea to try to enable that for all modules at once, as it introduces massive diffs which are impossible to review, and we might accidentally lose useful information, so currently it's only enabled for new modules like `food` but disabled for existing modules).

Even though we dont generally consider editing definitions as semver-major, we may as well make some bundle size improvements in v9 when data is changing anyway as a result of the 53 bit randomizer #2357

## Methodology

For any definitions in word module with >1000 entries, I shuffle the array, select 1000 at random and then sort (the sorting is why you see some green in the diffs and not just red)

In total this removes about 530KB from the full bundle.


ST-DDT:
Please also enable the script cleanup for the word module here:

https://github.com/faker-js/faker/blob/776a6258de07108548fb5918bdb8b572a278075a/scripts/generate-locales.ts#L329

matthewmayer:
That would cause all other locales to be re-sorted so I will do that in a subsequent PR to avoid making the diff too big.

ST-DDT (approved):

Shinigami92 (approved):

PR #2754, opened 2024-03-15
refactor!: remove v8 deprecated type aliases

Shinigami92:


ST-DDT (approved):

docs/guide/upgrading_v9/2754.md
```diff
@@ -0,0 +1,31 @@
+### Remove deprecated type aliases
+
+Removed deprecated type aliases
+
+| old                              | replacement                     |
```
matthewmayer, line 5: is the table overkill? How likely is it that consumers of the library would actually be using these definition types directly? Maybe we could just say that a number of types were renamed, such as `CompanyDefinitions` to `CompanyDefinition` and the old aliases have now been removed, without having to list every one?

Shinigami92: I suggest to add this to all the other replacement tables that were created in the migration folder and dont do a specific thing just because these are types

xDivisionByZerox: In my opinion it's perfectly fine to leave this in as is. It's compressed information for a specific kind of audience. Those are TypeScript developers that use our exported types. I don't see why this should not matter, as it breaks your code in the same way, as if you would forget to give the now required params for `date.betweens`. We go as far has having runtime checks there, but TypeScript dev don't even get a compressed migration information?

xDivisionByZerox (approved):

docs/guide/upgrading_v9/2754.md
matthewmayer: Ok the answer to "How likely is it that consumers of the library would actually be using these definition types directly" is clearly "likely"

So it's fine.

matthewmayer (approved):

PR #2755, opened 2024-03-15
infra(netlify): diff to next

ST-DDT:
Currently, if you merge next into a PR, a build preview is created, even if the PR itself does not change anything docs related.

The new check uses next as a comparison base and uses a common ancestor comparison.
E.g. any changes on next in the meantime are ignored.

`git diff next...some-branch` vs `git diff next some-branch`

Unfortunately, this will refresh the preview for all commits in a PR that changes docs, not just those that affect the docs.

So more build-minutes spend on PRs that would change the docs. And no more build-minutes spend on PRs that don't change the docs.


Shinigami92 (approved):

ST-DDT:
Alternatively, we could check for a `/^docs/` (+feat+refactor?) commit name.

Or maybe even a specific (empty) `enable netlify` commit.

I would have to check that though if it does strange things with update to next though.

ejcheng (approved):

xDivisionByZerox (approved):

xDivisionByZerox:
> Alternatively, we could check for a `/^docs/` (+feat+refactor?) commit name.
> 
> Or maybe even a specific (empty) `enable netlify` commit.
> 
> I would have to check that though if it does strange things with update to next though.

KISS. If we need to change the process again, we can have a look if it's required.

PR #2756, opened 2024-03-15
refactor(date)!: birthdate improvements

ST-DDT:
I found this branch lying around.

- Slightly simplifies the methods implementation/bringing it closer to our standard.
- Tries to improve the error message if min > max
- Adds some additional tests

I'm unsure about the contents of the error message:

- `Max age 25 (599616000000) should be greater than or equal to min age 31 (757382400000).`
- `Max year 1990 (662515200000) should be greater than or equal to min year 2000 (946771200000).`

The value in the parenthesis is the epoch timestamp.
Should we remove it?

- `Max age 25 should be greater than or equal to min age 31.`
- `Max year 1990 should be greater than or equal to min year 2000.`

What if min or max is undefined?

- `Max age 25 should be greater than or equal to min age undefined.`
- `Max year undefined should be greater than or equal to min year 2000.`

We could omit/remove the check if the additional checks from #2719 is merged.

- #2719


test/modules/date.spec.ts
```diff
@@ -477,65 +477,86 @@ describe('date', () => {
...
+        it('returns a random birthdate in one year', () => {
           const min = 1990;
-          const max = 2000;
+          const max = 1990;
 
           const birthdate = faker.date.birthdate({ min, max, mode: 'year' });
 
           // birthdate is a date object
           expect(birthdate).toBeInstanceOf(Date);
+          expect(birthdate.toISOString()).not.toMatch(/T00:00:00.000Z/);
```
Shinigami92: I dont understand this `expect` 🤔

Shinigami92 (commented):
If we want these epoch numbers in error message, we might need to add info about that to the message itself

test/modules/date.spec.ts
ST-DDT: A previous version of the method returned always midnight.
This isnt a new test it just got moved a bit.

matthewmayer:
Min and max have defaults (18 and 80) so you should be allowed to omit one or the other? 
Under what circumstances would `Max age 25 should be greater than or equal to min age undefined.` appear?

ST-DDT:
> Min and max have defaults (15 and 80) so you should be allowed to omit one or the other?

As long as you dont violate the other default. Yes.

matthewmayer:
So I would expect
`faker.date.birthdate({max:25})` to work (18-25)
`faker.date.birthdate({max:15})` to throw `Max age 15 should be greater than or equal to min age 18.`

test/modules/date.spec.ts
Shinigami92: But it is a possible value? (even when really uncommon)

ST-DDT:
> So I would expect `faker.date.birthdate({max:25})` to work (18-25) `faker.date.birthdate({max:15})` to throw `Max age 15 should be greater than or equal to min age 18.`

And that would work, if it wasnt for the year mode.
Should I throw two different errors based on mode?

test/modules/date.spec.ts
ST-DDT: Yes, it theoretically is possible.

Shinigami92: Then I suggest to rewrite or add a test, so that specific case is covered.
E.g. like we did with some other methods in a for-loop or (something like that) and then `expect` that it also generates not only 0-values.

Right now the test indicates that **never** a 0-values will be generated.

matthewmayer:
I generally think the more specific the error the better.

matthewmayer:
Looking at this again in the current docs, it's really weird that the default mode is year and yet the default for min and max are ages.

ST-DDT:
> Looking at this again in the current docs, it's really weird that the default mode is year and yet the default for min and max are ages.

Should we change that as well?

test/modules/date.spec.ts
ST-DDT: I removed the test as the intention of the test cannot easily be derived just from looking at it.

matthewmayer:
I feel age mode is much more useful. Year mode isn't really much different to between()

matthewmayer:
Perhaps to help upgrades we should log a warning if you don't specify mode and min/max are say >1000 (because then you surely meant years not ages).

ST-DDT:
> Perhaps to help upgrades we should log a warning if you don't specify mode and min/max are say >1000 (because then you surely meant years not ages).

I added a check.

But I'm considering making min/max/mode all required if present.

````ts
birthdate(
    options: {
      /**
       * The date to use as reference point for the newly generated date.
       *
       * @default faker.defaultRefDate()
       */
    } | {
      /**
       * The minimum age or year to generate a birthdate.
       */
      min: number;
      /**
       * The maximum age or year to generate a birthdate.
       */
      max: number;
      /**
       * The mode to generate the birthdate. Supported modes are `'age'` and `'year'` .
       *
       * There are two modes available `'age'` and `'year'`:
       * - `'age'`: The min and max options define the age of the person (e.g. `18` - `42`).
       * - `'year'`: The min and max options define the range the birthdate may be in (e.g. `1900` - `2000`).
       */
      mode: 'age' | 'year';
      /**
       * The date to use as reference point for the newly generated date.
       *
       * @default faker.defaultRefDate()
       */
    } = {}
  ): Date
````

**Comment by matthewmayer** [comment]
I think that's better.

**Comment by Shinigami92** [comment]
I assume that discriminated options based on mode would be overkill to just separate the docs?

```ts
... | {
      /**
       * The minimum age to generate a birthdate.
       */
      min: number;
      /**
       * The maximum age to generate a birthdate.
       */
      max: number;
      mode: 'age';
    } | {
      /**
       * The minimum year to generate a birthdate.
       */
      min: number;
      /**
       * The maximum year to generate a birthdate.
       */
      max: number;
      mode: 'year';
...
```

---

Does it make a difference for the tooltip and/or Vitepress docs if the `mode` is above the `min` and `max`?

ST-DDT:
This might cause issues with the apidocs as the param name is no longer unique and has a different description.

With the potential new apidocs generation, we could go for different signatures though as a future improvement.

src/modules/date/index.ts
```diff
@@ -431,16 +430,19 @@ export class SimpleDateModule extends SimpleModuleBase {
...
     const refDate = toDate(options.refDate, this.faker.defaultRefDate);
     const refYear = refDate.getUTCFullYear();
+    const optionsSet = [min, max, originalMode].filter((x) => x != null).length;
+    if (optionsSet % 3 !== 0) {
+      deprecated({
+        deprecated:
+          "faker.date.birthdate({ min: 18, max: 80 }) or faker.date.birthdate({ mode: 'age' })",
+        proposed: "faker.date.birthdate({ min: 18, max: 80, mode: 'age' })",
+        since: '9.0',
+        until: '10.0',
```
matthewmayer: does this need an `until`? Seems this check would need to be permanent?

Could just have a custom error like `You must provide all of the min, max and mode parameters`.

matthewmayer (changes requested):

src/modules/date/index.ts
matthewmayer: i guess this is currently a warning not an error. I think an error might be better (even though its theoretically "more" breaking) given the currnet behavior is quite unintuitiuve.

Shinigami92:
I would like to discuss in a meeting if we want to change the default `mode` and if `age > 1000` might be allowed for an ancient Dungeons&Dragons Dragon 🐉 

Also I will re-evaluate this PR when it was rebased in the changes of #2757, so there is ensured no diff/conflict.

ST-DDT:
> to change the default `mode`

We removed the default.

> if `age > 1000` might be allowed for an ancient Dungeons&Dragons Dragon 🐉

It is allowed. We currently just require the user to fully state what they want instead of defaulting half the values to some value hoping to fix the input to something that can be generated and makes sense.

Shinigami92:
> > to change the default `mode`
> 
> 
> 
> We removed the default.
> 
> 
> 
> > if `age > 1000` might be allowed for an ancient Dungeons&Dragons Dragon 🐉
> 
> 
> 
> It is allowed. We currently just require the user to fully state what they want instead of defaulting half the values to some value hoping to fix the input to something that can be generated and makes sense.

Ah that's good, I was not up to date. But yeah for this case I also think it is better to force request a mode instead of falling back to a default and produce unexpected behavior.

Edit: Ah I think I mixed that up, because `'age'` is still the default while destructing the option.

src/modules/date/index.ts
ST-DDT: This is only for the migration period, after that JS users get/have default values at most and have to read the docs.

I might switch to an error for the migration period. Not sure about that yet.

xDivisionByZerox:
**Team Decision**:

We will implwment this method by using multiple designated signatures. Where each mode is one signature with its own default values for min and max.

---

Blocked until #2628 + follow up is merged, to verify that we can display multiple signatures in a satisfying way.

ST-DDT:
> We will implement this method by using multiple designated signatures. Where each mode is one signature with its own default values for min and max.

Done

> Blocked until #2628 + follow up is merged, to verify that we can display multiple signatures in a satisfying way.

I also added/kept the merged signature, so theoretically we could merge this independently of #2628

src/modules/date/index.ts
```diff
@@ -335,90 +335,201 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+      min: originalMin,
+      max: originalMax,
+      mode: originalMode,
+    } = options;
+
+    // TODO @ST-DDT 2024-03-17: Remove check in v10
+    const optionsSet = [originalMin, originalMax, originalMode].filter(
+      (x) => x != null
+    ).length;
+    if (optionsSet % 3 !== 0) {
```
Shinigami92, line 515: I think this can be simplified using `[].some` or similar

(Send from phone)

ST-DDT: Unfortunately not. It is basically: `!(all || none)` != `some`

Shinigami92: 🤯

Shinigami92:
Potentially approved, currently don't have a bigger screen

src/modules/date/index.ts
```diff
@@ -335,90 +335,201 @@ export class SimpleDateModule extends SimpleModuleBase {
...
-      /**
-       * The mode to generate the birthdate. Supported modes are `'age'` and `'year'` .
-       *
-       * There are two modes available `'age'` and `'year'`:
-       * - `'age'`: The min and max options define the age of the person (e.g. `18` - `42`).
-       * - `'year'`: The min and max options define the range the birthdate may be in (e.g. `1900` - `2000`).
-       *
-       * @default 'year'
-       */
       mode?: 'age' | 'year';
```
Shinigami92: To align with the other signatures, I suggest to move this over the min,max
(inline suggestion not possible 🙁)

Shinigami92 (dismissed):

ejcheng (dismissed):

matthewmayer:
Can we see the updated docs in a netlify preview? The defaults still seem a bit confusing.

ST-DDT:
> Can we see the updated docs in a netlify preview?

Updated

> The defaults still seem a bit confusing.

What do you refer to?

matthewmayer:
Looking at https://deploy-preview-2756.fakerjs.dev/api/date.html#birthdate

- "Returns a random birthdate in the given range of years." doesn't seem ideal. I think we should explain both possible modes here?

- it's not obvious what the default are if you don't pass options object. 

- max says "The maximum year to generate a birthdate in." Shouldn't that be "year or age"?

- maybe reorder the parameters mode,min,max,refDate

ST-DDT:
> * "Returns a random birthdate in the given range of years." doesn't seem ideal. I think we should explain both possible modes here?
> 
> * it's not obvious what the default are if you don't pass options object.
> 
> * max says "The maximum year to generate a birthdate in." Shouldn't that be "year or age"?

I think I might have forgotten to adjust the description after creating the merged signature.



> * maybe reorder the parameters mode,min,max,refDate

They are sorted alphabetically (per type in the union), I will not change it in the old generation, if we replace it soon anyway.

matthewmayer:
Why does the code for mode year need to use the refDate? It seems that should only be needed for age?

ST-DDT:
> Why does the code for mode year need to use the refDate? It seems that should only be needed for age?

Fixed. Ready for review.

test/modules/date.spec.ts
```diff
@@ -122,10 +118,6 @@ describe('date', () => {
           mode: 'age',
           refDate,
         })
-        .it('with year mode and refDate', {
-          mode: 'year',
-          refDate,
-        })
         .it('with year and refDate', {
```
Shinigami92: Is the test wrong here? 🧐🤨

ST-DDT: Yes, in multiple ways. Fixed.

src/modules/date/index.ts
```diff
@@ -357,90 +357,196 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+   * Returns a random birthdate. By default, the birthdate is generated for an adult between 18 and 80 years old.
+   * But you can customize the `'age'` range or the `'year'` range to generate a more specific birthdate.
    *
-   * @param options The options to use to generate the birthdate. If no options are set, an age between 18 and 80 (inclusive) is generated.
-   * @param options.min The minimum age or year to generate a birthdate.
-   * @param options.max The maximum age or year to generate a birthdate.
-   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `now`.
-   * @param options.mode The mode to generate the birthdate. Supported modes are `'age'` and `'year'` .
+   * @param options The options to use to generate the birthdate.
+   * @param options.refDate The date to use as reference point for the newly generated date. Defaults to `faker.defaultRefDate()`.
```
matthewmayer, line 364: should we mention this is only used in 'age' mode?

src/modules/date/index.ts
```diff
@@ -357,90 +357,196 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+
+        return this.between({ from, to });
+      }
+
+      case 'year': {
+        const { min = refYear - 80, max = refYear - 19 } = options;
+        // Avoid generating dates on the first and last date of the year
+        // to avoid running into other years depending on the timezone.
+        const from = new Date(Date.UTC(0, 0, 2)).setUTCFullYear(min);
+        const to = new Date(Date.UTC(0, 11, 30)).setUTCFullYear(max);
```
matthewmayer: can't we just do:
```js
const { min , max } = options;
const from = new Date(Date.UTC(0, 0, 2)).setUTCFullYear(min);
const to = new Date(Date.UTC(0, 11, 30)).setUTCFullYear(max);
```

matthewmayer (changes requested):

src/modules/date/index.ts
ST-DDT: IMO that is not neccesary. The year specific signature does not have a refDate option in the signature.
And this signature actually uses the value.

src/modules/date/index.ts
ST-DDT: You mean without the defaults?
Theoretically yes, but it is currently required for backwards compatibility for JS users.

matthewmayer: what code in JS would still need the old defaults now that we're throwing an error unless we get 0 or 3 of {min, max, mode} passed?

src/modules/date/index.ts
matthewmayer: it would be nice for that to appear on the web docs though.

i just noticed refDate now appears twice in the generated docs:

https://deploy-preview-2756.fakerjs.dev/api/date.html#birthdate

image

src/modules/date/index.ts
ST-DDT: Good point, but due to the signature, it shows min and max as potentially absent => require either a default or a non null assertion.

I moved the defaults to the top, so it is consistent with the "defaults" as stated in the jsdocs.
While I think this isn't perfect, I consider this good enough.

src/modules/date/index.ts
ST-DDT: I added it to the merged signature.

> i just noticed refDate now appears twice in the generated docs:

Yeah, I know. Not going to fix this in the old docs generation though.

matthewmayer (approved):

Shinigami92 (approved):
- Not sure if we should first merge #2628

ST-DDT:
> Not sure if we should first merge #2628 

IMO that PR doesnt change much regarding this PR. A followup for both displaying/handling multiple signature might though.

I'll likely merge the other PR first, because it is bigger. Unless it is delayed a lot.

PR #2757, opened 2024-03-15
refactor(date)!: fail on invalid dates

ST-DDT:
If a given date is invalid, it is treated as an error instead of falling back to `refDateSource`


docs/guide/upgrading_v9/2757.md
```diff
@@ -0,0 +1,4 @@
+### Fail on invalid dates
+
+Previously, if invalid dates where passed to the date module, it would fallback to `faker.refDateSource`.
```
matthewmayer: `where` -> `were`

We should specifically list out which methods/parameters this affects

matthewmayer (changes requested):

docs/guide/upgrading_v9/2757.md
ST-DDT:
```suggestion
Previously, if invalid dates were passed to the date module, it would fallback to `faker.refDateSource`.
```

ST-DDT: > We should specifically list out which methods/parameters this affects

This affects all methods in the date module that take `Date`-ish values.

test/modules/date.spec.ts
```diff
@@ -144,6 +144,22 @@ describe('date', () => {
   describe.each(times(NON_SEEDED_BASED_RUN).map(() => faker.seed()))(
     'random seeded tests for seed %i',
     () => {
+      describe('toDate()', () => {
+        describe.each(['anytime', 'past', 'future', 'recent', 'soon'] as const)(
```
xDivisionByZerox: Is `'birthdate'` missing in here?

```suggestion
        describe.each(['anytime', 'past', 'future', 'recent', 'soon', 'birthdate'] as const)(
```

docs/guide/upgrading_v9/2757.md
matthewmayer: i think we could be clearer on what an "invalid" date means then.
e.g.
faker.date.soon({refDate:'2000'}) - '2000' is certainly not a valid `Date`, but its valid because we parse strings into `Date`s.

matthewmayer: How about something like this:

---

Various methods in the `faker.date` module allow you to pass a `Date`-ish value: that is, either a Javascript Date, or a timestamp number or string that can be converted to a `Date` via the `new Date()` constructor.

Previously if you passed something which could not be parsed to a `Date`, it would fall back to the current reference date. Now, this throws an error raising awareness of that bad value.

This affects the `refDate` parameter of  the `anytime()`, `birthdate()`, `past()`, `future()`, `recent()` and `soon()`, methods as well as the `from` and `to` parameters of `between()` and `betweens()`.

test/modules/date.spec.ts
ST-DDT: Changed

docs/guide/upgrading_v9/2757.md
ST-DDT: Thanks for your suggestion. I changed the text accordingly.

src/modules/date/index.ts
```diff
@@ -1,30 +1,23 @@
-import type { Faker } from '../..';
+import { type Faker } from '../..';
```
Shinigami92: The `@typescript-eslint/consistent-type-imports` with option `separate-type-imports` is not triggering here?!
https://typescript-eslint.io/rules/consistent-type-imports/

src/modules/date/index.ts
```diff
@@ -426,8 +397,8 @@ export class SimpleDateModule extends SimpleModuleBase {
     let min: number;
     let max: number;
```
Shinigami92, line 400: Should we rename these to `epochMin/Max` and then destruct the `min/max` options like in every other place?
Also then we can remove the comment in [line 396](https://github.com/faker-js/faker/pull/2757/files#diff-33d9552c7d1fb7f98cb50e15888b577d2609cd9b6f17a333b641752c5049629dR396)

src/modules/date/index.ts
ST-DDT: I dont see a CI error so no.
Not sure why it changed in the first place.

src/modules/date/index.ts
ST-DDT: We have a separate PR for birthdate:

- #2756

src/modules/date/index.ts
ST-DDT:
```suggestion
import type { Faker } from '../..';
```

matthewmayer (approved):

test/modules/date.spec.ts
```diff
@@ -564,22 +584,36 @@ describe('date', () => {
...
       faker.seed(20200101);
       const date2 = faker.date.past();
-      expect(date2).toMatchInlineSnapshot('2019-02-25T21:52:41.819Z');
+      expect(date2).toMatchInlineSnapshot(`2019-11-06T02:07:17.181Z`);
+    });
+
+    it('should not use the refDateSource when refDate is provided (with function)', () => {
+      faker.setDefaultRefDate(() => {
+        throw new Error('Should not be called');
+      });
```
Shinigami92: We can use vitest native spy here and check that it is not called

It would look something like this:

```ts
const mySpy = vi.spyOn(faker, defaultRefDate)

expect(mySpy).not.toHaveBeenCalled()
```

or similar

Shinigami92 (commented):
I wondered why the snapshots are different, but this is due to not calling `date.setTime(future/past)` anymore. This is okay to me, as it is still on the same Day, but just a different time.

So there is only one thing we can improve in the vitest test, and then I'm okay with approving this

ST-DDT:
> I wondered why the snapshots are different, but this is due to not calling `date.setTime(future/past)` anymore. This is okay to me, as it is still on the same Day, but just a different time.

Both date and time are different for all methods in the past. Because we now generate the range directly instead of subtracting it. So the result is basically flipped.

src/modules/date/index.ts
```diff
@@ -5,26 +5,19 @@ import { SimpleModuleBase } from '../../internal/module-base';
 import { assertLocaleData } from '../../locale-proxy';
 
 /**
- * Converts date passed as a string, number or Date to a Date object.
+ * Converts date passed as a string, number or Date to a valid Date object.
  * If nothing or a non-parsable value is passed, then it will take the value from the given fallback.
```
matthewmayer: line 9 is no longer accurate.

matthewmayer (changes requested):

Shinigami92 (approved):

matthewmayer (approved):

xDivisionByZerox (approved):

PR #2773, opened 2024-03-26
fix(person): add dedicated first names in ko locale

qqww08:
Fixed an issue where Korean and English were mixed. 
And we added the latest trend names for 2022 and 2023.

```tsx
fakerKO.person.fullName(). // 위 Blake
```


ST-DDT:
> Fixed an issue where Korean and English were mixed.

I didnt see any english names in the list.
Or do you refer to the ko translations of English names?
Or do you refer to the bug where firstName(sexType) return English names? #1677 

> And we added the latest trend names for 2022 and 2023.

This list seems to be significantly shorter than the previous one. Can we bump it to maybe a 100 each?

xDivisionByZerox (dismissed):

qqww08:
@ST-DDT 

There is no problem with firstName and lastName.

Try calling `fakerKO.person.fullName()`

I will update the list again with more names.

Shinigami92 (dismissed):

matthewmayer (commented):
Would it be better to change src/locales/ko/person/name.ts to remove the space between the family name and given name ie
```ts
{ value: '{{person.lastName}} {{person.firstName}}', weight: 1 },
```
to
```ts
{ value: '{{person.lastName}}{{person.firstName}}', weight: 1 },
```

per https://en.wikipedia.org/wiki/Korean_name "When a Korean name is written in Hangul, there is no space between the surname and the given name."

qqww08:
@ST-DDT 
> This list seems to be significantly shorter than the previous one. Can we bump it to maybe a 100 each?

We have added the ranking of the most used names in Korea from 1st to 300th. 
https://efamily.scourt.go.kr/index.jsp

I referred to Korean government statistics.

> Would it be better to change src/locales/ko/person/name.ts to remove the space between the family name and given name ie
per https://en.wikipedia.org/wiki/Korean_name "When a Korean name is written in Hangul, there is no space between the surname and the given name."


That's right, in Korea there are no spaces in the name. This was reflected immediately.

Shinigami92 (approved):

matthewmayer (approved):

ST-DDT (approved):

PR #2775, opened 2024-03-27
feat(locale): add location for uz_UZ_latin

Mirazyzz:


src/locales/uz_UZ_latin/location/street_pattern.ts
```diff
@@ -0,0 +1,4 @@
+export default [
+  '{{location.street_suffix}} {{location.street_name}}',
```
ST-DDT: This looks strange to me.
Why does it start with the suffix?
Also IMO we should name the street_name file street_name_part or similar.

ST-DDT (commented):
Please sort the files alphabetically.

src/locales/uz_UZ_latin/location/street_pattern.ts
Mirazyzz: I missed that one, suffix should be after the street name.

Mirazyzz:
> Please sort the files alphabetically.

Do you want the street names to be sorted alphabetically? They're pretty large and looking at other locales I thought it's not necessary.

Shinigami92:
I suggest to add the locale to the PR title

Shinigami92 (dismissed):
Count me already as approved but fix the change request from @ST-DDT pls

matthewmayer:
> > Please sort the files alphabetically.
> 
> 
> 
> Do you want the street names to be sorted alphabetically? They're pretty large and looking at other locales I thought it's not necessary.

All *new* locale data should be sorted alphabetically unless there's a good reason (eg months of the year can be kept as Jan-Feb-mar...)

ST-DDT (changes requested):
Please make the requested changes.

Mirazyzz:
> Please make the requested changes.

Ready! sorry for the delay.

src/locales/uz_UZ_latin/location/street_pattern.ts
```diff
@@ -0,0 +1 @@
+export default ['{{location.street_name}} {{location.street_suffix}}'];
```
matthewmayer: This should be changed to location.street_name_part now right?

matthewmayer (changes requested):

src/locales/uz_UZ_latin/location/street_pattern.ts
Mirazyzz: exactly, my bad.

ST-DDT (dismissed):

Shinigami92 (dismissed):

src/locales/uz_UZ_latin/location/street_address.ts
```diff
@@ -0,0 +1,4 @@
+export default {
+  normal: '{{location.street}}, {{location.buildingNumber}}',
+  full: '{{location.street}}, {{location.buildingNumber}} {{location.secondaryAddress}}',
```
matthewmayer: Is this correct?
If a secondary address (e.g. room/apartment number is provided) this immediately follows the building number with no comma etc?

without secondary address:
```
fakerUZ_UZ_latin.location.streetAddress()
> Munif Mahanov qishlog'i, 905
> Sobit Miranov xiyoboni, 35a
```

with secondary address:
```
fakerUZ_UZ_latin.location.streetAddress(true)
> Munif Mahanov qishlog'i, 905 972 uy
> Sobit Miranov xiyoboni, 35a 817 xonadon
```

matthewmayer (changes requested):

src/locales/uz_UZ_latin/location/street_address.ts
Mirazyzz: Yes, you're right I think it's better to separate them with comma. By the way, it's also better to put "-" for secondary address between house number and title (uy, xonadon). In Uzbek "-" is used for ordinal numbers. Do you think I should change it too?

matthewmayer: Yes

matthewmayer (approved):

Shinigami92 (approved):

ST-DDT (approved):

PR #2786, opened 2024-04-05
feat(animal): add more simple animals to animal.type()

### LINKED ISSUE #1618: Method to get any random animal

Hi! I just want a method that returns an animal, of any type.  Sometimes you just want  any kind of animal rather than one of a specific species.

### Suggested solution

maybe something like

```typescript
faker.animal.animal()
// or
faker.animal.random()
```

### Alternative

_No response_

### Additional context

_No response_

matthewmayer:
fix #1618

This adds more common animals and uses more common names for existing ones e.g. cetecean->whale, crocodilia->crocodile


Shinigami92 (approved):
❤️

ST-DDT (approved):

ejcheng (approved):

src/modules/animal/index.ts
```diff
@@ -194,7 +194,7 @@ export class AnimalModule extends ModuleBase {
    * Returns a random animal type.
```
Shinigami92, line 194: Should we add a warning here, that this is not compatible with `faker.animal[faker.animal.type()]()`?
Because we have similar use-cases with `person.sexType` and `color.cssSupportedFunction`

PR #2788, opened 2024-04-07
docs(upgrading): add link to upgrading snippets

ST-DDT:
Currently the docs don't contain a hint how to upgrade.

https://next.fakerjs.dev/guide/upgrading.html

This PR adds a link to the individual snippets until we merge them into one big file.

https://deploy-preview-2788.fakerjs.dev/guide/upgrading.html


matthewmayer:
I'd rather concatenate all the snippets into the files. The alpha is useful for looking for mistakes in the docs as well as mistakes in the code, and that's easier with a concatenated file.

Shinigami92 (approved):

ejcheng (approved):

xDivisionByZerox:
> I'd rather concatenate all the snippets into the files. The alpha is useful for looking for mistakes in the docs as well as mistakes in the code, and that's easier with a concatenated file.

Not sure about that. The reason we opted for these separate migration files was the idea to write a comprehensive migration files that makes sense in it's structure (order of section, grouping of related changes, etc). We would assentially do all this work twice by already concatinating all files.

That being said, I'm not sure about simply puting a link to the snippet directory - the UX feels really bad.

matthewmayer:
We could do a rough version just by deciding on a sensible order and noting down the order of files for future reference

`cat xxxx.md yyyy.md zzzz.md > ../upgrading.md`

ST-DDT:
IMO we could just merge the migration guide files together, since we already have most breaking stuff already. So before any release me merge the snippets into the main file.
AFAICT we already completed most of the migration categories, so any additional migration task would end up in a new section anyway and not likely affect the existing ones.
Or that being said, we could stop the separate migration files at this point.

PR #2789, opened 2024-04-07
refactor(locale)!: remove unused global locale faker instance

ST-DDT:
Removes the unused `global` faker instance that is only a copy of the `base` one.


xDivisionByZerox (approved):

matthewmayer (approved):

Shinigami92 (commented):
I'm a bit confused here 🤔 
Why is it okay here to remove it directly without any runtime deprecation warning?
Also, is this just a missed left-over and we forgot to remove it once we build the PR?

xDivisionByZerox:
> Why is it okay here to remove it directly without any runtime deprecation warning?

Because it was never meant to be in a release. This is an artifact from the introduction of the base locale, AFAICT.

Shinigami92 (approved):
With the responsibility on the other team-members, I approve

PR #2790, opened 2024-04-07
fix: tree-shaking

### LINKED ISSUE #1791: Investigate why importing anything from '@faker-js/faker' directly causes all of faker to be bundled.

### LINKED ISSUE #2335: Can add `sideEffects: false` to `package.json`

xDivisionByZerox:
# Summary

This PR addresses tree-shaking issues by optimizing the handling of locale imports and exports. The changes ensure that only the necessary locale modules are included in the final bundle, enhancing performance and reducing unnecessary file sizes.

This is a major requirement for the release of v9.0.

# Details

The previous implementation was causing unnecessary modules to be included during tree-shaking due to the additional export of the implicit `allLocales` variable. 

https://github.com/faker-js/faker/blob/03ac47198313126c3cc8cb3fb9ef184360bb79eb/src/index.ts#L39

This was refactored into an explicit export under `./locales/index.ts`. The `generate-locales` script has been updated to accommodate this change. 

Additionally, the package.json file has been updated to specify `"sideEffects": false`, which informs bundlers that all files within this package are side-effect-free and can be safely tree-shaken.

_To be perfectly honest, even after investigating this issue for multiple hours, I'm still not perfectly sure WHY this exact change fixes the issue._

**Please note, that only the combination of these changes will take any action. Applying one or the other change separately does not resolve the issue of everything being included in a final project bundle.**

# Testing

To verify these changes, follow these steps:

1. Check out this pull request locally.
1. Build the faker library using `pnpm run build`.
1. Clone our [playground repository](https://github.com/faker-js/playground).
1. Follow the instructions in the README.md file of the playground repository.
1. Override the faker dependency in the playground's `package.json` (replace "overrides-for-dev" with "overrides-for-dev").
1. Build/compile the playground projects `webpack-cjs` and `webpack-esm`.
1. Ensure that the console prints a bundle size number of around 400-480 KB when linking binaries from this pull request branch.
1. Compare this bundle size with the approximately 2.2 MB bundle size when linking binaries from the next branch.

These steps will confirm that the tree-shaking optimizations have reduced the bundle size appropriately and that the library functions as expected with the updated locale handling.

# Additional

I consider this a breaking change as of https://github.com/faker-js/faker/issues/2335#issuecomment-1687709538.

Fixes #1791, #2335.

- #1791 
- #2335


ST-DDT (dismissed):
Awesome, works as expected ❤️

grafik
grafik

Shinigami92 (dismissed):
This is so weird ...

Theoretically you changed nothing cause `export * from './locale';` in index does the same export for `allLocales` as before... but it just works magically 🤷 

I really don't understand what's going on here

Also I really don't like this `sideEffects` key, that is undocumented non-standard webpack introduced, but ecosystem accepted. Why is it not e.g. that it needs to be set to `sideEffects: true` explicitly if you have side effects and the default should be false...

However, great work and don't take my ecosystem-rant personally ❤️

Shinigami92:
This is **not** even a breaking change, is it? 👀 

> # Additional
>
> I consider this a breaking change as of https://github.com/faker-js/faker/issues/2335#issuecomment-1687709538.
> 
> Fixes #1791, #2335.
> 
> - #1791 
> - #2335

Ah..., well,
sideEffect is now "okay-ish" as we don't have side effects anymore as historically with the unique method, but this got successfully resolved
Also the investigation is now also not a breaking change anymore ^^'

IMO we can remove the label, but you already also did not add a `!` to the PR title, so I'm okay with that

xDivisionByZerox:
> ... but you already also did not add a ! to the PR title, so I'm okay with that

Haven't done any actual OSS coding in a while + my work does not use conventional commits (so no ! for breaking changes). So I honestly forgot that this is required ^^

---

I'd still think we should consider this breaking. Just in case. I'll add a migration guide file.

xDivisionByZerox:
> ... but you already also did not add a ! to the PR title, so I'm okay with that

Haven't done any actual OSS coding in a while + my work does not use conventional commits (so no ! for breaking changes). So I honestly forgot that this is required ^^

matthewmayer:
While it's not breaking, it does mean you no longer need to use the workaround to import one locale right? So we could explain that in the migration guide.

matthewmayer:
Can we also remove the "Individual localized packages" section of the Usage docs if this is no longer needed?

docs/guide/upgrading_v9/2790.md
```diff
@@ -0,0 +1,7 @@
+### Fix Tree Shaking
+
+Prior to this version, users had to resort to workarounds by importing specific faker instances from dedicated paths to overcome tree shaking issues.
+With the implementation of this fix, such workarounds should no longer be necessary.
```
matthewmayer, line 9: Can we give an example of the old code which is no longer necessary?

xDivisionByZerox: Agreed. I've done that.

xDivisionByZerox:
> Can we also remove the "Individual localized packages" section of the Usage docs if this is no longer needed?

Wasn't sure about that, but simpl removed it now. 

Please check the guide section on the docs preview. I feel like the sections "Custom locales and fallbacks" and "Available lcoales" should be switched now.

Shinigami92 (dismissed):

docs/guide/upgrading_v9/2790.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+With the implementation of this fix, such workarounds should no longer be necessary.
+That means that you should be able to import different localized faker instances from the root of your package.
+
+```ts
+import { fakerDE, fakerES, fakerFR } from '@faker-js/faker';
+```
+
+The dedicated import paths will still stay for now, to allow a graduate migration for our users.
```
matthewmayer: Graduate -> gradual

xDivisionByZerox:
```suggestion
The dedicated import paths will still stay for now, to allow a gradual migration for our users.
```

ST-DDT (approved):

matthewmayer (approved):

ejcheng (approved):

ST-DDT:
Thanks for fixing this. ❤️

SalahAdDin:
Testing it soon, thanks!

Shinigami92:
> Testing it soon, thanks!

I cant speak for the whole team, but maybe we will do an alpha release today in next 10h :eyes: :rocket:

ST-DDT:
FFR: https://github.com/faker-js/faker/issues/2807

- https://github.com/faker-js/faker/issues/2807

PR #2791, opened 2024-04-07
refactor(locale): normalize animal data

xDivisionByZerox:
# Description

Normalize the entries in the `animal` module definitions in all locales.

I choose this module as only one file (`src/locales/fr/animal/bird.ts`) that exceeded the entry limit. Th rest of the changes come from sorting and deduping the datasets.


ST-DDT (dismissed):
The test snapshots need to be updated, but otherwise this looks good to me.

matthewmayer (commented):
Is there anything we can do to speed up the normalisation code? It seems to add several seconds to every preflight run even when only running for a couple of modules.

ST-DDT (dismissed):

Shinigami92 (dismissed):

xDivisionByZerox:
> Is there anything we can do to speed up the normalisation code?

Guess we can parallelize the processing of all methods in a module as well as the module processing in general. This would not fix the data normalization in general, tho. Any ideas on that?

ST-DDT (approved):

Shinigami92 (approved):

PR #2803, opened 2024-04-09
refactor(locale): remove discontinued credit card issuer "maestro"

xDivisionByZerox:
# Description

Removes the discontinued credit card issue "maestro" as described in https://github.com/faker-js/faker/issues/2350#issuecomment-1722230447.


xDivisionByZerox:
Please tell me if this is considered breaking. Afterwards I can set the correct labels/title/milestone (and add a migration snippet).

ST-DDT:
> Please tell me if this is considered breaking. Afterwards I can set the correct labels/title/milestone (and add a migration snippet).

I would consider it breaking because a "well known value" will stop working even if the type does not cover it.

---

See also https://github.com/faker-js/faker/pull/2720

- https://github.com/faker-js/faker/pull/2720

ST-DDT (approved):

ST-DDT:
Are the other credit card issuers valid?

xDivisionByZerox:
> Are the other credit card issuers valid?

This is **not** supposed to be a fix for #2350, as removing a discontinued issue is different from fixing invalid patterns, based on an issue, IMO.

What I want to say is: I did not check the other pattern as I would do so in a different PR (which is not breaking as it is a fix)

matthewmayer:
> > Please tell me if this is considered breaking. Afterwards I can set the correct labels/title/milestone (and add a migration snippet).
> 
> 
> 
> I would consider it breaking because a "well known value" will stop working even if the type does not cover it.
> 
> 
> 
> ---
> 
> 
> 
> See also https://github.com/faker-js/faker/pull/2720
> 
> 
> 
> - https://github.com/faker-js/faker/pull/2720

I think it's worth a quick mention in the migration guide to note that;
- creditCardNumber with no options will never return maestro card numbers any more
- creditCardNumber({issuer:"maestro"}) will fall back to returning a number from a random issuer.

ST-DDT:
> * creditCardNumber({issuer:"maestro"}) will fall back to returning a number from a random issuer.

And any other method that would have an option removed/renamed would work as well.
Only in this case we dont even get a ts error.
IMO `MaestroCreditCard` != random `CreditCard`

Just for clarification: I dont require a migration guide, if #2720 is merged after this before any release.

matthewmayer:
for reference we didn't consider #2344 breaking

matthewmayer (approved):

PR #2811, opened 2024-04-13
docs: minor migration guide updates

matthewmayer:
- standardize title case for headings
- clearer headings
- grammar simplification
- present tense
- add link to enforce-unique
- enforce-unique does now include exclude


docs/guide/upgrading.md
```diff
@@ -18,7 +18,7 @@ v9 has not yet been released. This page contains a work-in-progress list of brea
 
 ## General Breaking Changes
 
-### Node 14 and 16 No Longer Supported
+### Node 14 and 16 Not Supported
```
ST-DDT: Maybe

```suggestion
### Require Node v18+
```

or

```suggestion
### Upgrade to Node v18+
```

ST-DDT (dismissed):

docs/guide/upgrading.md
matthewmayer: Changed to `### Requires Node 18+`

ST-DDT (dismissed):

docs/guide/upgrading.md
```diff
@@ -44,26 +44,25 @@ If you are unable to upgrade to TS5, you have to keep using Faker v8.
 
 ### Fix Tree Shaking
 
-Prior to this version, users had to resort to workarounds by importing specific faker instances from dedicated paths to overcome tree shaking issues.
+Prior to this version, users had to resort to a workaround by importing specific faker instances from dedicated paths to overcome an issue which caused all locales to be bundled even if only one was used.
```
ejcheng:
```suggestion
Prior to this version, there was an issue where all locales would be bundled even if only one was used. Users had to resort to a workaround by importing specific faker instances from dedicated paths.
```

docs/guide/upgrading.md
```diff
@@ -44,26 +44,25 @@ If you are unable to upgrade to TS5, you have to keep using Faker v8.
...
 
 ```ts
 import { faker } from '@faker-js/faker/locale/de';
 ```
 
-With the implementation of this fix, such workarounds should no longer be necessary.
-That means that you should be able to import different localized faker instances from the root of your package.
+With the implementation of this fix, this workaround should no longer be necessary. You should be able to import different localized faker instances from the root of your package, and the bundle will only include the specific locales.
```
ejcheng:
```suggestion
With this fix, the workaround should no longer be necessary. You will be able to import different localized faker instances from the root of your package with the bundle only including those specific locales.
```

docs/guide/upgrading.md
```diff
@@ -44,26 +44,25 @@ If you are unable to upgrade to TS5, you have to keep using Faker v8.
...
 
 ```ts
 import { fakerDE, fakerES, fakerFR } from '@faker-js/faker';
 ```
 
-The dedicated import paths will still stay for now, to allow a gradual migration for our users.
+The dedicated import paths are kept in v9, to allow a gradual migration for our users.
 
-While the implementation of this change does not constitute as breaking according to semantic versioning guidelines, it does impact the behavior of users bundlers.
+While the implementation of this change does not constitute as breaking according to semantic versioning guidelines, it does impact the behavior of users' bundlers.
```
ejcheng:
```suggestion
While this is not a breaking change according to semantic versioning guidelines, it does impact the behavior of users' bundlers.
```

docs/guide/upgrading.md
```diff
@@ -134,19 +133,19 @@ const faker = new Faker({
 });
 ```
 
-### Using `tsup` for the build process
+### Using `tsup` for the Build Process
 
-We only support exports defined via `package.json` but after the switch to `tsup`, there are now complete restructures in the dist folder resulting it minified and chunked files for cjs.
+After the switch to `tsup`, the dist folder now contains minified and chunked files for CJS. However, as we support exports defined via `package.json` this should not affect your code.
```
ejcheng:
```suggestion
After the switch to `tsup`, the dist folder now contains minified and chunked files for CJS. However, as we support exports defined via `package.json`, this should not affect your code.
```

docs/guide/upgrading.md
```diff
@@ -626,9 +625,9 @@ This affects:
 - The `variant` property of `faker.location.countryCode()` must be one of `alpha-2`, `alpha-3`, `numeric` if provided
 - The `casing` property of `faker.string.alpha()` and `faker.string.alphanumeric()` must be one of `'upper' | 'lower' | 'mixed'` if provided
 
-### faker.phone.number `style` replaces explicit `format`
+### Phone Number `style` Replaces Explicit `format`
 
-`faker.phone.number()` generates a phone number for the current locale. However, previously there was little control over the generated number, which might or might not include country codes, extensions, white space and punctuation.
+`faker.phone.number()` generates a phone number for the current locale. Previously there was little control over the generated number, which might or might not include country codes, extensions, white space and punctuation.
```
ejcheng:
```suggestion
`faker.phone.number()` generates a phone number for the current locale. Previously, there was little control over the generated number, which may or may not have included country codes, extensions, white space, and punctuation.
```

docs/guide/upgrading.md
```diff
@@ -562,11 +561,11 @@ We plan to rethink this method some more in the future: [#2579](https://github.c
 
 `faker.image.dataUri()` now returns an image url with a random width and height by default, additionally the type of the image is now random. To obtain the previous behavior, pass `{width: 640, height: 480, type: 'svg-uri'}`.
 
-### Require `from` and `to` in `faker.date.between()` and `betweens()`
+### Require `from` and `to` in `faker.date.between` and `betweens`
 
-Previously, in `faker.date.between()` and `faker.date.betweens()` if the `from` or `to` parameter was omitted (in Javascript) or an invalid date (in Javascript or Typescript), they would default to the current date or reference date. Now, both boundaries must now be given explictly. If you still need the old behavior, you can pass `Date.now()` or the reference date for `from` or `to`.
+Previously, in `faker.date.between()` and `faker.date.betweens()` if the `from` or `to` parameter was omitted (in Javascript) or an invalid date (in Javascript or Typescript), they would default to the current date or reference date. Now, both boundaries must now be given explicitly. If you still need the old behavior, you can pass `Date.now()` or the reference date for `from` or `to`.
```
ejcheng:
```suggestion
Previously, in `faker.date.between()` and `faker.date.betweens()` if the `from` or `to` parameter was omitted (in Javascript) or an invalid date (in Javascript or Typescript), they would default to the current date or reference date. Now, both boundaries must be given explicitly. If you still need the old behavior, you can pass `Date.now()` or the reference date for `from` or `to`.
```

docs/guide/upgrading.md
```diff
@@ -18,7 +18,7 @@ v9 has not yet been released. This page contains a work-in-progress list of brea
 
 ## General Breaking Changes
 
-### Node 14 and 16 No Longer Supported
+### Requires Node 18+
 
 Support for Node.js versions 14 and 16 has been discontinued as these versions have reached their [end-of-life](https://github.com/nodejs/Release). Faker.js 9.0 requires a minimum of Node.js version 18.
```
Shinigami92: Should we constantly stick to the format `vX`?

```suggestion
Support for Node.js v14 and v16 has been discontinued as these versions have reached their [end-of-life](https://github.com/nodejs/Release). Faker.js v9 requires a minimum of Node.js v18.
```

docs/guide/upgrading.md
```diff
@@ -44,26 +44,25 @@ If you are unable to upgrade to TS5, you have to keep using Faker v8.
```
Shinigami92: The line above (43) could be in a note/hint box

docs/guide/upgrading.md
```diff
@@ -134,19 +133,19 @@ const faker = new Faker({
 });
 ```
 
-### Using `tsup` for the build process
+### Using `tsup` for the Build Process
 
-We only support exports defined via `package.json` but after the switch to `tsup`, there are now complete restructures in the dist folder resulting it minified and chunked files for cjs.
+After the switch to `tsup`, the dist folder now contains minified and chunked files for CJS. However, as we support exports defined via `package.json`, this should not affect your code.
```
```suggestion
After the switch to [tsup](https://tsup.egoist.dev), the `dist` folder now contains minified and chunked files for CJS. However, as we officially support only `exports` defined via `package.json`, this should not affect your code.
```

Shinigami92 (commented):
I need to check further from line 140+ later, but got to catch a meeting now

docs/guide/upgrading.md
```diff
@@ -116,7 +115,7 @@ If you have seeded tests, you have to update most test snapshots or similar comp
 
 If you are using vitest, you can do that using `pnpm vitest run -u`.
```
Shinigami92: We could tryout bullet-points for these adoptions (line 113-117, removing the blank lines in between)

```suggestion
If you are using [vitest](https://vitest.dev), you can do that using `pnpm vitest run -u`.
```

docs/guide/upgrading.md
```diff
@@ -294,7 +293,7 @@ Prior to v9, Faker provided a [`faker.helpers.unique()`](https://v8.fakerjs.dev/
 
 Please see the [unique values guide](/guide/unique) for alternatives.
 
-For example, many simple use cases can use [`faker.helpers.uniqueArray`](https://v8.fakerjs.dev/api/helpers.html#uniqueArray). Or you can migrate to a third party package such as `enforce-unique`:
+For example, many simple use cases can use [`faker.helpers.uniqueArray`](https://v8.fakerjs.dev/api/helpers.html#uniqueArray). Or you can migrate to a third party package such as [`enforce-unique`](https://www.npmjs.com/package/enforce-unique):
```
```suggestion
For example, many simple use cases can use [`faker.helpers.uniqueArray`](https://v8.fakerjs.dev/api/helpers.html#uniqueArray). Or you can migrate to a third party package such as [`enforce-unique`](https://www.npmjs.com/package/enforce-unique) where we granted [MansurAliKoroglu](https://github.com/MansurAliKoroglu) **officially** to take over the maintenance for this utility functionality:
```

docs/guide/upgrading.md
```diff
@@ -351,7 +350,7 @@ const city = enforcer.enforce(faker.location.city, {
 ```
 
 ::: tip Note
-`enforce-unique` does not support the `exclude` or `store` options. If you were previously using these, you may wish to build your own unique logic instead.
+`enforce-unique` does not support the `store` option. If you were previously using this, you may wish to build your own unique logic instead.
```
```suggestion
`enforce-unique` does not support the `store` option. If you were previously using this for really specific custom behavior, you might need to request an issue at https://github.com/MansurAliKoroglu/enforce-unique/issues/new. If you just need to reset the store, you can simply call the `reset()` method on the `UniqueEnforcer` instance.
```

docs/guide/upgrading.md
```diff
@@ -476,7 +475,7 @@ Removed deprecated random module
 
 ### Locale Aliases
 
-Removed deprecated locale aliases
+Removed deprecated locale aliases `cz`, `en_IND`, `ge` and `global`.
```
```suggestion
Renamed deprecated locale aliases `cz`, `en_IND`, `ge` and removed `global`.
```

docs/guide/upgrading.md
matthewmayer: I'm not really sure the sentence is needed at all. Seems fairly self-evident. (We don't say "if you are unable to upgrade to Node v18, you have to keep using Faker v8" etc etc)

docs/guide/upgrading.md
matthewmayer: i don't think we specifically have to say its "official". (if it was official, we'd move it within the faker-js org and support it). And we dont want to disourage anyone else from making a "competing" package with different opinions on how to handle uniqueness.

docs/guide/upgrading.md
matthewmayer: i feel this is maybe too much detail. Realistically if you want some custom behavior, you should probably just build it, not open an issue in a third party package and hope and wait.

docs/guide/upgrading.md
matthewmayer: i have removed it

docs/guide/upgrading.md
matthewmayer: done

docs/guide/upgrading.md
Shinigami92: I just would like to tell folks that this is the package that was born out of faker's implementation. There were people already that cried about that it is now a random unknown package with an "untrusted" maintainer... And I would like to have some wordings here so people stop thinking like this.

docs/guide/upgrading.md
Shinigami92: > [...] not open an issue in a third party package and hope and wait

I personally dislike anti-open-source like "craft your own thing, hide it from the outside-world and dont contribute back".

Also I would like to formulate the current sentence into a more positive one, not saying like "if you need that old functionality, you are screwed"

docs/guide/upgrading.md
matthewmayer: how about just 

Or you can migrate to a *recommended* third party package such as...

docs/guide/upgrading.md
matthewmayer: true, but "you might need to request an issue at " doesn't encourage contributing back either, it just redirects the complaints :)

How about something in between:

```md
`enforce-unique` does not support the `store` option. If you were previously using this, you may wish to build your own logic instead, and please consider [contributing](https://github.com/MansurAliKoroglu/enforce-unique]/issues/new) to the enforce-unique package.
```

docs/guide/upgrading.md
Shinigami92: @ST-DDT @xDivisionByZerox what do you think?

docs/guide/upgrading.md
Shinigami92: it's better, I'm just unsure if telling "store is not supported" is a lie, as there is a store and it is also resetable

docs/guide/upgrading.md
xDivisionByZerox: > i don't think we specifically have to say its "official". (if it was official, we'd move it within the faker-js org and support it).

I have to agree with this statement. 

> There were people already that cried about that it is now a random unknown package with an "untrusted" maintainer

Well, that was one of the problems we were aware of when we discussed the removal of this function from Faker. TBH, these people are kind of in the right.
Sure the new package "was born out of faker's implementation", but are you tracking the development of it? 
Did you ever get to know the new maintainer of this package, other than the "10" messages from the issue where they wanted to help? 
I am not saying that MansurAliKoroglu is a bad person. The problem is that I can't say the opposite with 100% certainty. As a maintainer of a widely used package, I kinda want to have this certainty. Especially, after the xz attack.

docs/guide/upgrading.md
xDivisionByZerox: > it's better, I'm just unsure if telling "store is not supported" is a lie, as there is a store and it is also resetable

After taking a look at the documentation of the package, I have to agree. While the statement "`enforce-unique` does not support the `store` option" is correct in itself (the is no argument called "store"), writing it this way might give users the wrong idea.

docs/guide/upgrading.md
Shinigami92: TBH saying person XY could be bad, especially in faker history context XD, can be always said :shrug: 
Even I could be a bad person and someday just deploy a bitcoin miner inside Faker, but I wont, cause this would ruin my IT carriere

MansurAliKoroglu: I am not a maintainer of faker but I would like to comment.

> i don't think we specifically have to say its "official". (if it was official, we'd move it within the faker-js org and support it).

I agree with this. `enforce-unique` is not an official faker-js org package. I am also OK with you moving the package into faker-js org and actually officially support it. Additionally, I understand @Shinigami92 's emphasis. For that, @matthewmayer 's suggestion looks fine to me. *"Or you can migrate to a recommended third party package such as..."*

> There were people already that cried about that it is now a random unknown package with an "untrusted" maintainer...

> but are you tracking the development of it?

About this however, I agree with @Shinigami92 You can never be sure someone could be a bad person or change their mind and put some harmful code in any open source repository in any time.

Currently the code in `enforce-unique` is not harmful. It's tiny enough to review the code and for sure faster than implementing a custom solution. To prevent harmful code in future versions, for an npm package it's easy to use a fixed version and npm will take care the rest. A fixed version's code never changes.

Yet, for other people's perspective, I don't know any way to ensure an open source package will **never** be harmful. :shrug:

matthewmayer:
i have updated the info about `enforce-unique` taking into account everyone's comments.

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

Shinigami92 (approved):

ST-DDT:
@RD-CL Thanks for bringing this to our attention.
Please contact us at `ert@fakerjs.dev` with details so that we can look into that.
GitHub issues is the wrong place to bring this up, so please refrain from further comments here.

PR #2812, opened 2024-04-13
infra(unicorn): no-await-expression-member

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/no-await-expression-member`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-await-expression-member.md) lint rule.


test/faker.spec.ts
```diff
@@ -18,7 +18,8 @@ describe('faker', () => {
       .filter((key) => typeof console[key] === 'function')
       .map((methodName) => vi.spyOn(console, methodName));
 
-    (await import('..')).default;
+    const file = await import('../src');
+    expect(Object.keys(file)).toContain('Faker');
```
Shinigami92: For what is this check?

ST-DDT: > should not log anything on startup (import)

Shinigami92: I don't understand

ST-DDT: Anyone using `import { ... } from '@faker-js/faker';` should not be spammed with any logs (e.g. because of deprecations).
There should only be logs after calling a (deprecated) method/accessing a (deprecated) property.

Shinigami92: I did not mean line https://github.com/faker-js/faker/pull/2812/files#diff-bd97acc8a72719c8b174e098f25710cf89ceed17f7bf691cdef26d905b93b081R27
I mean line https://github.com/faker-js/faker/pull/2812/files#diff-bd97acc8a72719c8b174e098f25710cf89ceed17f7bf691cdef26d905b93b081R22

ST-DDT: This checks that we actually import a valid file/the main index file.

Shinigami92: 1. I would have expected that it would fail in that case
2. Maybe we should add a comment, cause right now it was not clear

ST-DDT: Could you please make a suggestion?

ST-DDT: FFR:

- @Shinigami92 added the code https://github.com/faker-js/faker/blame/db88a1518e8d0ccfe191bd48a1ae8641a9a25303/test/faker.spec.ts#L24
- as requested here https://github.com/faker-js/faker/pull/857#discussion_r854572060 by me

Shinigami92: I'm on mobile, I currently do not have the time and screen to make an active suggestion 

Please also let me explain that import .. is fundamentally different to ../src
The one looks into the package.json and do a main/exports field resolution, the other runs just the source code without any transpilation steps

ST-DDT: I wasnt aware of the difference between the two import targets.

Shinigami92 (dismissed):

Shinigami92 (approved):

ejcheng (approved):

PR #2815, opened 2024-04-13
infra(unicorn): numeric-separators-style

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/numeric-separators-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/numeric-separators-style.md) lint rule.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -45,6 +45,7 @@ module.exports = defineConfig({
     'unicorn/no-null': 'off', // incompatible with TypeScript
     'unicorn/no-zero-fractions': 'off', // deactivated to raise awareness of floating operations
     'unicorn/number-literal-case': 'off', // incompatible with prettier
+    'unicorn/numeric-separators-style': 'off', // "magic numbers" may carry specific meaning
     'unicorn/prefer-ternary': 'off', // ternaries aren't always better
 
     // TODO @Shinigami92 2023-09-23: The following rules currently conflict with our code.
@@ -52,7 +53,6 @@ module.exports = defineConfig({
     'unicorn/better-regex': 'off',
     'unicorn/consistent-function-scoping': 'off',
     'unicorn/no-object-as-default-parameter': 'off',
-    'unicorn/numeric-separators-style': 'off',
     'unicorn/prefer-export-from': 'off',
     'unicorn/prefer-string-slice': 'off',
     'unicorn/prevent-abbreviations': 'off',


ST-DDT:
Not sure about this one. Sometimes it looks useful, at other times the numbers aren't actual numbers...

xDivisionByZerox (dismissed):

Shinigami92 (dismissed):
Sill thinking that it looks weird for digits <= 5 🤷

Shinigami92 (dismissed):

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
```diff
@@ -171,7 +171,7 @@ ${examples}`;
 
               // This only checks whether the whole method is deprecated or not
               // It does not check whether the method is deprecated for a specific set of arguments
-              it('verify @deprecated tag', { timeout: 30000 }, async () => {
+              it('verify @deprecated tag', { timeout: 30_000 }, async () => {
```
matthewmayer: Some cases like this are an improvement

test/modules/location.spec.ts
```diff
@@ -182,9 +182,9 @@ describe('location', () => {
         });
 
         it.each([
-          ['IL', 60001, 62999],
-          ['GA', 30001, 31999],
-          ['WA', 98001, 99403],
+          ['IL', 60_001, 62_999],
```
matthewmayer: But some like this are worse

matthewmayer (commented):
I'm not convinced about the need for this rule. I think we could leave it up to authors to use separators when appropriate.

test/modules/location.spec.ts
ST-DDT: Curious: Could you please give some insights why you consider this worse than the case above and the original?
Is it because the resulting value doesn't have it/isn't really a number?

matthewmayer: Yes

matthewmayer: It's not Beverly Hills 90_210

xDivisionByZerox:
I've evaluated the proposed rule again and it has become apparent for me that its implementation could generally enhance the uniformity of "magic number" definitions within a codebase. However, our repository's unique context, where numeric values often carry specific meanings, may be compromised by the adoption of numeric separator syntax. This concern was previously expressed by @matthewmayer ([see here](https://github.com/faker-js/faker/pull/2815#discussion_r1580565195)) as well.

While it is possible to configure the rule to enforce separators only where they currently exist, thereby standardizing their usage (e.g., in groups of three), this approach contradicts the rationale for introducing the rule in the first place.

Given these considerations, I propose that we opt not to include this rule in our ruleset.

ST-DDT:
**Team Decision**

- We will disable this rule permanently

Shinigami92 (approved):

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2824, opened 2024-04-15
feat(locale): add lorem for uz_UZ_latin 

Mirazyzz:


Mirazyzz:
What could be the root cause for the failing unit test: _test/scripts/apidocs/verify-jsdoc-tags.spec.ts > verify JSDoc tags > phone > number > 0 > verify @example tag
     → Test timed out in 5000ms._? Can someone help me @ST-DDT @matthewmayer ?

xDivisionByZerox:
We are currently analyzing the problem with the CI pipeline (see #2782). For now, any team member has to restart failed jobs. Sorry for the inconvenience :(

It's definitely not on your side, tho.

Mirazyzz:
@ST-DDT can I ask you to perform code review?

ST-DDT:
I'm currently busy, I'll review it when I have time.

matthewmayer:
The English (well really fake Latin) words in lorem definitions are lowercase. Maybe do the same?

ST-DDT:
> The English (well really fake Latin) words in lorem definitions are lowercase. Maybe do the same?

It looks like Uz also has uppercase words, so IMO lorem can also contain them.

ST-DDT (dismissed):
The words itself look fine, but IMO the purpose of `lorem` is creating a text that is not supposed to be read. So en and de use Latin for that. Why did you use actual words?
(Or is it just google translate that assigns a translation to them even though they don't have any?)

Shinigami92 (dismissed):

Mirazyzz:
> The words itself look fine, but IMO the purpose of `lorem` is creating a text that is not supposed to be read. So en and de use Latin for that. Why did you use actual words? (Or is it just google translate that assigns a translation to them even though they don't have any?)

If the intent is to use actual lorem I can change it though I was looking to Russian locale as an example and there were proper words, that's why I used the same approach. I thought these words could be used to generate some comments, messages etc. However, if you think it's better to keep it consistent I will do so :)

matthewmayer:
My feeling is that most languages that use the Latin alphabet should probably just stick to the fake-Latin lorem so they don't need a locale-specific version of lorem.  If you want "real" words you should use the word module.

Mirazyzz:
> My feeling is that most languages that use the Latin alphabet should probably just stick to the fake-Latin lorem so they don't need a locale-specific version of lorem. If you want "real" words you should use the word module.

Sounds reasonable. Should I copy the words from Latin locale then into Uzbek? Or should I take it from somewhere and export it from Uzbek locale?

matthewmayer:
You probably don't have to do anything. All locales currently fall back to en and en already has the Latin lorem definitions.

ST-DDT:
**Team Decision**

- Languages that use latin characters should use the latin lorem text (e.g. copy it from en).
- We will create a separate PR that will document this behavior of the lorem feature in the definitions and module.
- We will create further PRs to add/fix the other locales according to their charset.

Mirazyzz:
> **Team Decision**
> 
> * Languages that use latin characters should use the latin lorem text (e.g. copy it from en).
> * We will create a separate PR that will document this behavior of the lorem feature in the definitions and module.
> * We will create further PRs to add/fix the other locales according to their charset.

Sorry for the late fix. I copied words should I update the docs and create new PR for that?

xDivisionByZerox:
> Sorry for the late fix. I copied words should I update the docs and create new PR for that?

All good, no worries. The only thing you need to do is the bullet point one as you did in your latest commit f681ef6932639d7d6f965f47324a8cba4ce3c191. 

The other bullet points are for us, the faker team, to remember certain decisions made.

ST-DDT (approved):

matthewmayer:
If it is identical to `en` could we not just have words.ts be like this to avoid duplication:

```ts
import words from '../../en/lorem/words';
export default words;
```

ST-DDT:
I also considered this, but we could also do that in the mentioned cleanup PR.
For me both works.

ejcheng (approved):

Shinigami92 (approved):

xDivisionByZerox:
> If it is identical to `en` could we not just have words.ts be like this to avoid duplication:
> 
> ```ts
> import words from '../../en/lorem/words';
> export default words;
> ```

I'd suggest moving this to a separate PR as well. This way we can take the time to investigate if this is somehow influencing tree shaking for specific localized faker instances. I'd prefer to leave this kind of work to the faker team.

xDivisionByZerox (approved):

ST-DDT:
I created a separate issue for the team todos:

- https://github.com/faker-js/faker/issues/2884

Thank you @Mirazyzz  for your contributions ❤️

Mirazyzz:
> I created a separate issue for the team todos:
> 
> * [Improve specification of lorem module and definitions #2884](https://github.com/faker-js/faker/issues/2884)
> 
> Thank you @Mirazyzz for your contributions ❤️

@ST-DDT Thank you for constantly providing constructive feedback, it's a pleasure to contribute to your community and project. I myself enjoy using faker library and to planning to contribute more in the future!

PR #2835, opened 2024-04-19
infra: suppress pnpm error

Shinigami92:
With pnpm v9 the packageManager field in package.json get checked strictly now
This config suppress the error and converts it just into a warning like

`[ WARN ] This project is configured to use v9.0.1 of pnpm. Your current pnpm is v9.0.4`

This only affects contributors that uses pnpm natively instead of corepack

This is needed so contributors can still run `preflight` without having the exact pnpm bugfix version installed (like with pnpm v8 previously)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-manager-strict=false


matthewmayer (approved):

xDivisionByZerox (approved):

ST-DDT (approved):
I also tested it with corepack. Corepack still downloads the appropriate version of pnpm and uses the specified one over any previously installed version, even if newer.

ST-DDT:
Please keep an eye open, to remove this property once it is no longer needed e.g. because of compatible version range checks.

Issue #2763, opened 2024-03-21
[v9 MAINTENANCE] - Please Read


xDivisionByZerox:
# Information

v9 will fundamentally changes the way our internal codebase is structured. The biggest changes will be the way modules classes will be composed.
**These changes doesn't affect the way you work with faker.**

If you plan on contributing a new module we suggest to wait with it. Otherwise you will most likely have additional work with following our refactorings (merge conflicts).

**This only affects module files**. 
You can still contribute locale data.

---

- See #2667

xDivisionByZerox:
**Community Update on Version 9 Release**

Dear Community Members,

We hope this message finds you well. As you know, our team has been diligently working on the upcoming release of version 9, which promises significant improvements and new features. Our initial plan was to delay feature requests for version 9 due to extensive internal restructuring of the codebase. This restructuring aimed to facilitate future enhancements and ensure the longevity and robustness of our platform.

However, we encountered some delays, and we understand that this has affected our contributors, especially those waiting for new module features. We want to acknowledge the patience of our community members during this period.

The good news is that our primary goal of implementing tree-shaking is finally complete. This feature is a substantial step forward in optimizing our platform’s performance and efficiency. With this milestone achieved, we have decided to wrap up version 8 and move forward with the [release of version 9](https://github.com/faker-js/faker/milestone/12).

We have made the difficult decision to release version 9 without the second planned feature, “standalone modules and functions.” We believe this will allow us to deliver a stable and improved version to you sooner, rather than later. We are committed to ensuring that the transition to version 9 is as smooth as possible for everyone involved.

We appreciate your understanding and continued support. Our team is excited to bring you the improvements in version 9, and we look forward to your feedback.

Thank you for being an integral part of our journey.

Happy faking!

Issue #2667, opened 2024-02-14
Proposal: Design for standalone module functions


ST-DDT:
### Clear and concise description of the problem

Currently the faker methods are organized in modules and these are joined to the whole Faker class.
Most methods in the modules reference a function in a different module, by invoking it via the embedded Faker reference.
This makes it near impossible for building tools to drop unused parts of code package.

Changing the implementation to use standalone methods, that only call other standalone methods, mostly solves this issue.
Old code/syntax (e.g. faker.person.firstName) must remain valid.

#### 1. Extract the functions from the module classes into their own files (the same applies to internal helper methods)

````txt
- modules
  - airline
    - airline.ts
    - airport.ts
    - ...
  - helpers
    - arrayElement.ts
    - arrayElements.ts
    - multiple.ts
    - ...
  - ..
````

#### 2. Change the signature of the methods, so that it replaces it the reference to the main `Faker` instance e.g. by passing it as argument.

````ts
function firstName(fakerCore: FakerCore, ...remainingArgs) {
  return arrayElement(fakerCore, fakerCore.definitions.person.firstName);
}
````

fakerCore ≈> Faker without any modules (only definitions, randomizer, config)

#### 3. Call the standalone functions from the modules.

````ts
export function newPersonModule(fakerCore: FakerCore): PersonModule {
  return ({
    firstName: (..args) => firstName(fakerCore, ...args);
    ...
  });
}
````

#### 4. Optionally, "process" the function to make it easier to use with the same/a fakerCore.

````ts
export function fakerize<T extends (fakerCore: FakerCore, ...args: any[]) => any>( fn: T ): FakerizedMethod<T> {
  const fakerized = fn as FakerizedMethod<T>;
  fakerized.withCore = (fakerCore: FakerCore) => (...args: FakerizedParameters<T>) => fn(fakerCore, ...args);
  // Potentially more methods
  return fakerized;
}

// We need to check whether this negatively impacts readability of the type if not inlined
type FakerizedMethod<T extends () => any> = T & Readonly<{
    withCore: ( fakerCore: FakerCore ) => (...args: FakerizedParameters<T>) => ReturnType<T>;
  }>;

type FakerizedParameters<T extends (...args: any[]) => any> =
  Parameters<T> extends [FakerCore, ...infer U] ? U : never;
````

Usage:

````ts
// in firstName.ts
export const firstName = fakerize(firstName);

// by users or maybe pre-built per locale
[...]

PR #2836, opened 2024-04-19
feat(word): add method for generating letter

### LINKED ISSUE #558: Random greek alphabet letter

Shinigami92:
closes #558

- #558


src/modules/word/index.ts
```diff
@@ -625,4 +636,18 @@ export class WordModule extends ModuleBase {
       .multiple(() => this.sample(), { count })
       .join(' ');
   }
+
+  /**
+   * Returns a random letter.
+   *
+   * It can also return diacritic letters in some locales, like german `ä`.
```
matthewmayer, line 643: If we are going to do it this way, I think we should specifically call out the ability to generate a Greek letter using fakerEL since Greek letters are often used in scientific contexts and it's not obvious that this is possible from the current docs.

src/modules/word/index.ts
```diff
@@ -2,6 +2,17 @@ import { FakerError } from '../../errors/faker-error';
 import { ModuleBase } from '../../internal/module-base';
 import { filterWordListByLength } from './filter-word-list-by-length';
 
+export interface Letter {
+  /**
+   * The uppercase letter.
```
matthewmayer, line 7: Not all alphabets have an uppercase and lowercase variant for letters for example, Thai, Arabic, Georgian etc. What would you expect to be returned for those ? Would you just return the same letter repeated for lowercase and uppercase?

ST-DDT:
The referenced issue has barely any upvotes or comments. 🤔 
So I consider it low prio.
Also there is the current feature freezes to prepare for the v9 changes.

- #2763 
- #2667

src/modules/word/index.ts
Shinigami92: Yeah, I thought about this 😕 
Either we could discuss together a bit about this on Discord or in a team meeting

Maybe fallback and just fill upper and lower with same value might be the easiest solution

matthewmayer:
The other problem with putting Greek letters in fakerEL is I cannot easily use it in a helpers.fake pattern

Eg if I want 
"Incredible Soft Gloves are located in Section A, subsection δ" I can't do

"{{commerce.productName}} are located in Section {{foo.romanLetter}} subsection {{foo.greekLetter}}"

Shinigami92:
> The other problem with putting Greek letters in fakerEL is I cannot easily use it in a helpers.fake pattern
> 
> Eg if I want "Incredible Soft Gloves are located in Section A, subsection δ" I can't do
> 
> "{{commerce.productName}} are located in Section {{foo.romanLetter}} subsection {{foo.greekLetter}}"

Can you elaborate what this has to be done with this PR? Why does it not count for `chemicalElement` or `airline`?
I think this might then be more a problem of `fake` and is out of scope for this PR and feature implementation.

Or did I get your intend wrong? You want to use greek (mixed locale) context inside fake? This is new to me 🤔 And if so, you could still generate all this just without using `fake` but just wrap in a function.

```js
function x() {
  return `${faker.commerce.productName()} are located in Section ${faker.foo.romanLetter()} subsection ${fakerEL.word.letter()}`;
                                                                   ~~~~~~~~~~~~~~~~~~~~~~~ this does not even exists
}
```

matthewmayer:
Greek/roman letters are "international"

Like even if you are writing English you might use single Greek letters for some things.  Even if you are writing Chinese you might use single Roman letters for some things. 

You can of course use string interpolation. But fake patterns can be used by other faker definitions so it's nice if they can access useful data.

Shinigami92:
As I don't feel healthy (again) anymore, I will close this for now

PR #2837, opened 2024-04-19
refactor(locale): activate data normalization for color

xDivisionByZerox:
# Description

Normalize the entries in the `color` module definitions in all locales.

I choose this module because  only one file per locale. This should make reviews more trivial.


src/locales/es_MX/color/human.ts
```diff
@@ -1,32 +1,32 @@
 export default [
-  'rojo',
-  'verde',
-  'azul',
-  'amarillo',
-  'morado',
+  'Lima',
```
matthewmayer, line 2: Seems random that some colors are capitalised here?

ST-DDT: IMO we should fix it in a separate PR/issue though.

Shinigami92 (approved):

ST-DDT (approved):

src/locales/es_MX/color/human.ts
ST-DDT: Created a separate issue for that:

- https://github.com/faker-js/faker/issues/2841

PR #2838, opened 2024-04-20
feat: introduce FakerCore

ST-DDT:
First part of the standalone module function feature #2667 

- #2667 

## Overview

1. Introduce FakerCore #2838
2. De-proxy locale access #3189
3. De-class config function #3748 (seed, setDefaultRefDate, ...)
4. De-module functions #3748 (airline, ...)

## Description

Create the fakerCore container object.


Shinigami92 (commented):
I know the PR is still in draft but I would like to provide some early-feedback

1. I think the `refDate` should still be named something like `default` or `fallback` or `base`, so the intent is more clear and there might be no variable naming conflict when using an option's refDate + the ...RefDate
2. I suggest to not do something like `super({ fakerCore: fakerCore ?? { locale, randomizer, config: {} } })` but only pass the locale and randomizer so the config get's set like every other method call to a default internally. Or is there something special about that I currently overlook? 👀

ST-DDT:
1) Done
2) Changed slightly to improve readability.

ST-DDT:
Does anybody know why this test is failing with that particular error message?

>  ❯ test/simple-faker.spec.ts > simpleFaker > should not log anything on startup
     → Unexpected token ','

It seems to happen only with vitest/CJS.

Shinigami92:
> Does anybody know why this test is failing with that particular error message?
> 
> > ❯ test/simple-faker.spec.ts > simpleFaker > should not log anything on startup
> > → Unexpected token ','
> 
> It seems to happen only with vitest/CJS.

If I set `minify: false` in `tsup.config.ts` everything runs fine. Please report a bug-issue to https://github.com/egoist/tsup

Or it is related to `https://github.com/egoist/tsup/issues/1091`

ST-DDT:
The error is caused by this part:

````js
_class.prototype.__init23.call(this);,this.definitions=_chunkWAQHUEHTcjs.d.call(void 0, this.rawDefinitions)
                                     ^
````

Which is equivalent to this source code:

https://github.com/faker-js/faker/blob/e4da2f5478e3df74ed98d4249ace561dddc9f53d/src/faker.ts#L195-L197

**Comment by ST-DDT** [comment]
I created an issue in tsup with a minimal reproduction:

- https://github.com/egoist/tsup/issues/1124

And it has reeeaaally specific conditions, check the issue for more details.

File: src/faker.ts [hunk 1]
```diff
@@ -85,6 +86,11 @@ export class Faker extends SimpleFaker {
   readonly vehicle: VehicleModule = new VehicleModule(this);
   readonly word: WordModule = new WordModule(this);
 
+  get rawDefinitions(): LocaleDefinition {
+    // TODO @ST-DDT 2024-05-14: Should we deprecate this?
+    return this.fakerCore.locale;
+  }
```
ST-DDT, line 87: Should we deprecate this field?

xDivisionByZerox: Can you give reasoning on why we should  be doing that?

ST-DDT: Because you could access them via fakerCore directly.

Later: I thought about potentially removing faker.definitions in favor of an explicit method call as well.
`faker.definitions.module.entry` => `requireData(fakerCore, 'module', 'entry')`

EDIT: `requireData()` is required for the standalone module functions rewrite because there is no longer a `faker.definitions` to call to fail on missing data.

Shinigami92 (dismissed):
LGTM

ejcheng (dismissed):

matthewmayer (changes requested):
see my comment at https://github.com/faker-js/faker/issues/2667#issuecomment-2067690118 - would it make sense to delay merging this?

matthewmayer (dismissed):
changed my feedback to a comment to avoid a hard block, but i think there is still some logic in deferring this and merging together with the actual changes.

ST-DDT:
This PR contains the first part of the standalone module function PR:

- #3748

It uses a slightly different implementation then the one used in that PR but both are generally compatible.
I can modify either PR depending on which variant your prefer.

SMF PR:

- `Faker` -> `fakerToCore()` -> `FakerCore`
  https://github.com/faker-js/faker/blob/8fa7e270fdc130fa34e644dc282364e52f038baf/src/internal/faker-to-core.ts#L12-L22

This/FakerCore PR:

- `faker.fakerCore` permanent field in class
  https://github.com/faker-js/faker/blob/6dada60781938d5e76f87062abf0be9d1f9cf7b1/src/simple-faker.ts#L123-L125

test/faker.spec.ts
```diff
@@ -64,11 +63,8 @@ describe('faker', () => {
...
     describe('locale', () => {
       it('should throw error if no locales passed', () => {
-        expect(() => new Faker({ locale: [] })).toThrowError(
-          new FakerError(
-            'The locale option must contain at least one locale definition.'
-          )
-        );
+        expect(() => new Faker({ locale: [] })).not.toThrowError();
+        // TODO @ST-DDT 2026-03-22: Should we keep throwing an error here?
       });
```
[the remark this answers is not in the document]
Shinigami92: I think it might be a good UX to throw an error here, or direct users to use simpleFaker instead

Shinigami92 (commented):
I already gave an LGTM here: https://github.com/faker-js/faker/pull/2838#pullrequestreview-2088771780
left a comment here and I still think this PR is in a good state to get merged 🔜™️
is there a reason why it is still in draft? have you anything planned to change in this PR?

test/faker.spec.ts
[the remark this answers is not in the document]
ST-DDT: Please note that there is no difference constructor param wise between simple and full faker anymore.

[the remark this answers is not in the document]
Shinigami92: but maybe we could bring it back as a UX
lets wait for the third opinion -> @xDivisionByZerox

ST-DDT:
> is there a reason why it is still in draft? have you anything planned to change in this PR?

Only discussions regarding names, behaviour, API and PR contents.

src/faker.ts
```diff
@@ -120,51 +122,10 @@ export class Faker extends SimpleFaker {
...
-
-    if (Array.isArray(locale)) {
-      if (locale.length === 0) {
-        throw new FakerError(
-          'The locale option must contain at least one locale definition.'
-        );
-      }
-
-      locale = mergeLocales(locale);
-    }
```
Shinigami92, line 164: **note (non-blocking):** we will rethink to throw an UX error in v11

test/faker.spec.ts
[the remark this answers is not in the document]
Shinigami92: -> https://github.com/faker-js/faker/pull/2838#discussion_r3004963536

xDivisionByZerox:
**Team Decision**

We take this regardless of whether we also take the standalone modules PR (#3748)

---

@ST-DDT Please rebase/merge this branch against origin/next.

Shinigami92 (dismissed):

src/core.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { locale, randomizer, config, seed } = options;
+
+  if (randomizer != null && seed != null) {
+    randomizer.seed(seed);
+  }
+
+  return {
+    locale: Array.isArray(locale) ? mergeLocales(locale) : (locale ?? {}),
+    randomizer: randomizer ?? generateMersenne53Randomizer(seed),
+    config: config ?? {},
```
xDivisionByZerox: The default value from these properties can be set when restructuring them from `options` in the first line of the function. 

That way the randomizer is consistently seeded in line 89 as well, since the seed does not need to be provided immediately when generating the mersenne randomizer.

src/config.ts
xDivisionByZerox, line 1: Why did you decide to wrap the defaultRefDate into an additional "config" object? Isn't everything inside the core already kind of a "config"?

src/simple-faker.ts
```diff
@@ -29,13 +29,18 @@ import { generateMersenne53Randomizer } from './utils/mersenne';
...
+   * @internal
+   */
+  readonly fakerCore: FakerCore;
 
   /**
    * Gets a new reference date used to generate relative dates.
    */
   get defaultRefDate(): () => Date {
-    return this._defaultRefDate;
+    return this.fakerCore.config.defaultRefDate ?? (() => new Date());
```
xDivisionByZerox: As [copilot mentioned in its comment](https://github.com/faker-js/faker/pull/2838/changes#r2972016602) this will create new anonymous function instances over and over again. I feel like this should be defaulted during the creation of the core, so that it contains a "required" config instance and we do not have to handle nullish values.
This kind of plays into [my question why the config wrapper exist at all](https://github.com/faker-js/faker/pull/2838/changes#r3006435416).

xDivisionByZerox (dismissed):

src/faker.ts
xDivisionByZerox: Lets leave it in it's current state for now. I think we should still remove the comment.

src/config.ts
ST-DDT: To allow sharing the config instance with other core instances (similar to other randomizer/locale data)

ST-DDT: It is also typed specifically to allow other plugins/users to enhance it for their own methods.
(Although that would work for FakerCore as well)

src/simple-faker.ts
ST-DDT: I moved the default to a const.
Since the config is exposed as public, mutable, and optional (and we previously explicitly allowed setting it to null to reset it to default), I think we should keep the default on access to avoid potential errors when using them.
(Otherwise we wouldn't need the method in the first place)

src/core.ts
ST-DDT: It has slightly different default semantics, but I'll change it anyway.

src/core.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @returns The newly created FakerCore instance.
+ *
+ * @example
+ * import { createFakerCore, en } from '@faker-js/faker';
+ *
+ * createFakerCore() // no locale data, default randomizer and empty config
+ * createFakerCore({ locale: en }) // custom locale data, default randomizer and empty config
+ *
+ * @since 10.4.0
```
Shinigami92: **polish:** needs a bump, @xDivisionByZerox can you please let @ST-DDT know if he needs to do other cleanups? I won't like to get my approval constantly dismissed

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #2840, opened 2024-04-21
docs(usage): type definitions aren't strictly necessary

fzn0x:
type definitions aren't strictly necessary here


ST-DDT:
> type definitions aren't strictly necessary here

They might not be necessary there, but we have them there to reduce the cognitive load from the reader to guess what the method/field is supposed to do/return.

See also:

- https://typescript-eslint.io/rules/explicit-function-return-type/

What is your idea/intention with removing those (aka Why)?

fzn0x:
> > type definitions aren't strictly necessary here
> 
> They might not be necessary there, but we have them there to reduce the cognitive load from the reader to guess what the method/field is supposed to do/return.
> 
> See also:
> 
> * https://typescript-eslint.io/rules/explicit-function-return-type/
> 
> What is your idea/intention with removing those (aka Why)?

Typings still included without specifying `Users`.
Screenshot 2024-04-22 003654

fzn0x:
Improve the example for the Users type definition, or remove it altogether to allow users to explore the usage independently without encountering errors after choosing ESM or CJS.

ST-DDT (approved):

ST-DDT:
So the issue here is that the example on the readme uses a type that isnt specified anywhere on that page.

ejcheng (approved):

matthewmayer:
Maybe remove "export" too? If it's just a standalone example.

ST-DDT:
> Maybe remove "export" too? If it's just a standalone example.

I'm undecided on that. Maybe a little bit on the keep it part because I hope this pushes the users to reuse their faker methods and it slightly catches more attention.

matthewmayer (approved):

PR #2825, opened 2024-04-15
docs: use new X icon instead of twitter icon in social links

matthewmayer:
old

image

new

image

PR #2806, opened 2024-04-11
docs: add donate links

matthewmayer:
Add donation links to OC as discussed on Discord, two options

- In header

Screenshot 2024-04-11 at 20 17 58
Screenshot 2024-04-11 at 20 18 31

- In footer

Screenshot 2024-04-11 at 20 18 38

(maybe don't need both, see which you prefer and comment on wording)

PR #2802, opened 2024-04-09
docs: update roadmap for v9

xDivisionByZerox:
# Description

I noticed that the roadmap on our website is pretty outdated. I moved some stuff around and added a section for v9.

Feel free to add topics if you think they should be included in the roadmap.

PR #2847, opened 2024-04-23
docs: backport docs changes to v8 branch

matthewmayer:
Backports to the following changes to v8 branch so that the main fakerjs.dev website can be updated without needing to wait for v9 release.

- docs: add donate links (#2806)
- docs: use new X icon instead of twitter icon in social links (#2825)
- docs: update roadmap for v9 (#2802)
- docs: improve contributing guide (#2747)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,18 @@
 A lot of effort has been put into `Faker` to create a useful and handy library.
 There are still a lot of things to be done, so all contributions are welcome!
-If you want to make `Faker` a better, please read the following contribution guide.
+If you want to make `Faker` a better place, please read the following contribution guide.
 
-# Important
+## Before you start
+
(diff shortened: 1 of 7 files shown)


Shinigami92:
If we choose this strategy of backporting, we will never `--ff-only` again

matthewmayer:
> If we choose this strategy of backporting, we will never `--ff-only` again

Is it likely that we will need to merge into v8 in future? Most likely seems to be
- Minor website updates like this
- A super-critical bugfix requiring a 8.4.2 release

Is there an alternative to cherry-picking that keeps the history more clear?

ST-DDT:
At this point I dont expect any ff-only merges to v8 anyway.

ST-DDT:
@matthewmayer What is the status of this PR?

matthewmayer:
As there's not long to v9 now can probably just close it.

PR #2851, opened 2024-04-25
feat: add option to exclude emoji

### LINKED ISSUE #2623: Add option to exclude emoji

Please add an option to exclude emojis in `faker.person.bio()`. Would like to avoid having to write a script to remove/replace emojis when they are generated.

### Suggested solution

Perhaps one of 2 methods (or both):

1. Pass options to `bio()`
```
faker.person.bio({ emojis: false })
```
2. New method
```
faker.person.bioWithoutEmoji()
```

### Alternative

_No response_

### Additional context

_No response_

JDP818:
https://github.com/faker-js/faker/issues/2623#issue-2099738144

### DIFF
--- a/src/modules/person/index.ts
+++ b/src/modules/person/index.ts
@@ -9,6 +9,10 @@ export enum Sex {
 
 export type SexType = `${Sex}`;
 
+export type BioType =
+  | 'default'
+  | 'text';
+
 /**
  * Select a definition based on given sex.
  *
@@ -317,10 +321,18 @@ export class PersonModule extends ModuleBase {
    *
    * @since 8.0.0
    */
-  bio(): string {
-    const { bio_pattern } = this.faker.definitions.person;
-
-    return this.faker.helpers.fake(bio_pattern);
+  bio(
+    options: {
+      types?: ReadonlyArray<BioType>;
+    } = {}
+  ): string {
+    const {
+      types = Object.keys(this.faker.definitions.person) as BioType[],
+    } = options;
+    const bioType = this.faker.helpers.arrayElement(types);
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.person.bio_pattern[bioType]
+    );
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/modules/person/index.ts
```diff
@@ -312,15 +314,34 @@ export class PersonModule extends ModuleBase {
...
-
-    return this.faker.helpers.fake(bio_pattern);
+  bio(
+    options: {
+      /**
+       * A list of the bio pattern types that should be used.
+       *
+       * @default Object.keys(faker.definitions.person.bio_pattern)
+       */
+      types?: ReadonlyArray<BioType>;
```
xDivisionByZerox, line 333: Method options usally not provide extra or even replacement locale data. We usually allow users to control the method outcome through flags. In this case I could see such an option being named "includeEmojis" or "allowEmojis". We have similar options for:

- [`internet.email`](https://github.com/faker-js/faker/blob/1f7568ab1a7010c7f9c713f4ac4c2d5e0cfb350b/src/modules/internet/index.ts#L83)
- [`string.numeric`](https://github.com/faker-js/faker/blob/1f7568ab1a7010c7f9c713f4ac4c2d5e0cfb350b/src/modules/string/index.ts#L592)
- [`system.cron`](https://github.com/faker-js/faker/blob/1f7568ab1a7010c7f9c713f4ac4c2d5e0cfb350b/src/modules/system/index.ts#L321)

src/modules/person/index.ts
```diff
@@ -312,15 +314,34 @@ export class PersonModule extends ModuleBase {
...
+    } = {}
+  ): string {
+    const {
+      types = Object.keys(
+        this.faker.definitions.person.bio_pattern
+      ) as BioType[],
+    } = options;
+    const bioType = this.faker.helpers.arrayElement(types);
```
xDivisionByZerox, line 341: Couldn't we simply get rid of `'{{internet.emoji}}'` here?

ST-DDT: Maybe lets check the requirements and intentions behind including emojis in the next team meeting.

ST-DDT:
Closing the PR because we are waiting on more user interest helping us shape the method to a broader user base.

e.g. having a separate `style` parameter for `social-media` and `professional` vs a boolean toggle vs not including emojis at all

PR #2858, opened 2024-04-28
Images for wiki

pointspotdot:


xDivisionByZerox:
Before you put more effort into whatever you are doing here, may you add a description of what you are actually doing? Maybe an issue would have also been appropriately before opening a PR. I'd encourage you read the ["Before you start" section from our Contribution Guidelines](https://github.com/faker-js/faker/blob/1f7568ab1a7010c7f9c713f4ac4c2d5e0cfb350b/src/modules/system/index.ts#L321).

For all those fance images containing flow charts you might want to consider reading up on [Creating diagrams in GitHub](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) or using Markdown files containing [Mermaid syntax](https://mermaid.js.org/intro/#flowchart) directly.

For example using your "1203-currentflow.png" image:

```mermaid
graph LR;
    Start-->Step1[Call a name generation function];
    Step1-->Step2[Random name];
    Step2-->End;
```

matthewmayer:
Also note that
- Each pull request should only attempt to fix a single issue (it looks like you are trying to write code to fix 2011, 2623 and 1203 all in one pull request)
- You don't need to include screenshots of the changed code, since Github will automatically show that in the pull request
- If you need any other supporting images for your pull request, you can attach those on Github, rather than adding them to the repository

ST-DDT:
Temporily closing this until more information is provided.

What is the intended purpose of this PR?

PR #2864, opened 2024-04-29
infra: enable verbatimModuleSyntax

Shinigami92:
Enable `verbatimModuleSyntax` so `type` imports are used automatically when importing a type in VSCode via ctrl+space

Read also more about that here: https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax
We do not have any conflicts as we are using https://typescript-eslint.io/rules/consistent-type-imports/ anyway.


ST-DDT (approved):

xDivisionByZerox (approved):
I have no strong opinions about this since we already use the ESLint rule. Guess an extra layer of "security" won't hurt.

PR #2866, opened 2024-04-29
refactor(locale): normalize hacker data

xDivisionByZerox:
# Description

Normalize the entries in the `hacker` module definitions in all locales.

I choose this module as locale data are still limited. Additionally, no data set exceedes the 1k entry limit, so all locale file changes are based on entry sorting.


ST-DDT (approved):

src/locales/el/hacker/noun.ts
```diff
@@ -1,22 +1,22 @@
 export default [
-  'οδηγός',
-  'πρωτόκολο',
-  'ταχύτητα',
+  'feed',
```
Shinigami92, line 2: we should have a second look after this PR, why there are "english" words inside `el` where all other words are using `el` letters

Shinigami92 (approved):

src/locales/el/hacker/noun.ts
xDivisionByZerox: Same in `ru`. 👍

xDivisionByZerox: They seem to be legacy values from the original project: https://github.com/faker-js/faker/commit/61d02960a9132fa2c4d4d7259883b2e944233bb9

PR #2868, opened 2024-05-01
refactor(locale): normalize commerce data

xDivisionByZerox:
# Description

Normalize the entries in the `commerce` module definitions in all locales.

I choose this module as locale data are somewhat limited. Furthermore, none of the datasets surpass the 1k entry limit, meaning that all modifications to locale files involve sorting entries only.


ST-DDT (approved):

src/locales/az/commerce/product_name.ts
```diff
@@ -2,24 +2,24 @@ export default {
...
+    'Parlaq',
     'İnanılmaz',
-    'Fantastik',
+    'İntellektual',
     'Əlverişli',
-    'Parlaq',
-    'Mükəmməl',
   ],
-  material: ['Polad', 'Ağac', 'Beton', 'Plastik', 'Pambıq', 'Qranit', 'Rezin'],
+  material: ['Ağac', 'Beton', 'Pambıq', 'Plastik', 'Polad', 'Qranit', 'Rezin'],
```
matthewmayer, line 14: Should these have newlines ?

matthewmayer (changes requested):

matthewmayer (approved):

src/locales/az/commerce/product_name.ts
xDivisionByZerox: Prettier is responsible for formatting, including when a line break for a new is made. So I don't think I understand what you are referring to. Could you explain further?

Shinigami92: > Should these have newlines ?

Not in this case, cause < 80 cols

Shinigami92 (approved):

PR #2882, opened 2024-05-06
infra(deps): switch to commit-and-tag-version

ST-DDT:
Replaces the `standard-version` dependency with the maintained fork `commit-and-tag-version`.

- https://github.com/absolute-version/commit-and-tag-version
- [standard-version](https://github.com/conventional-changelog/standard-version) lists this as official fork/replacement

I did some tests locally and release seems to behave the exact same as `standard-version`, but you might want to try it out yourself.

I also explicitly disabled the tag generation, as we generate the tags via GH releases anyway.


Shinigami92 (dismissed):
nice

xDivisionByZerox:
We already had a config file for standard version: https://github.com/faker-js/faker/blob/043f68b0691f1cc1c11f2257b9c9e2cfa5047c72/.versionrc.json
If I understand correctly, we can keep using it. The new config should propably go into the config file instead of the package.json.

ST-DDT:
> We already had a config file for standard version: [`043f68b`/.versionrc.json](https://github.com/faker-js/faker/blob/043f68b0691f1cc1c11f2257b9c9e2cfa5047c72/.versionrc.json) If I understand correctly, we can keep using it. The new config should propably go into the config file instead of the package.json.

It doesn't show up in the config spec, but seems to work anyway.
https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/master/versions/2.2.0/README.md

xDivisionByZerox:
> It doesn't show up in the config spec, but seems to work anyway.

It's not obvious but as they state [in their README](https://github.com/absolute-version/commit-and-tag-version/blob/fabd8cd0525db0fc2f4c7d6aea1d39a764f3dfcc/README.md?plain=1#L186):

> Any of the command line parameters accepted by `commit-and-tag-version` can instead be provided via configuration.

In combination with the point aboth:

> You can configure `commit-and-tag-version` either by:
> 1. Placing a `commit-and-tag-version` stanza in your `package.json`
> 2. Creating a `.versionrc`, `.versionrc.json` or `.versionrc.js`.

xDivisionByZerox (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #2885, opened 2024-05-08
docs(lorem): define allowed words

### LINKED ISSUE #2884: Improve specification of lorem module and definitions
Languages that use latin characters should use the latin lorem text (e.g. copy it from en).
We will create a separate PR that will document this behavior of the lorem feature in the definitions and module.
After the change of definitions we should create another PR that actually changes the locale data to comply with the new spec and additional care should be taken regarding tree shakeability if the locale data are imported from other locales.

Especially the later part should be done by the team to ensure there are no side effects.

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/2824#issuecomment-2077678007_

xDivisionByZerox:
# Description

Document the expected words for a locale in the lorem module.

# Related to

This is the first part of #2884.

- #2884


ST-DDT (commented):
I thought about using normal words if they dont use Latin. 🤔

xDivisionByZerox:
> I thought about using normal words if they dont use Latin. 🤔

Oh wow, then I completly misunderstood what we discussed in the team meeting when we made the decision. Discuss again in tomorrows meeting?

matthewmayer:
We should survey what current non-Latin locales actually do and document that (eg are they nonsense words or real words or transliterations of Latin lorem?)

matthewmayer:
25 locales currently have a lorem/words.ts file. For each i generated 5 sample words

## Latin

| code | script | words(5) | notes |
| --- | --- | --- | --- |
| cs_CZ| Latn | quasi neque quasi delectus minima |standard Latin lorem |
| de| Latn | excepturi inventore nihil eveniet velit |standard Latin lorem |
| en| Latn | crur capillus denique veritas audacia | standard Latin lorem |
| fr| Latn | aliquid vitae accusamus suscipit est |standard Latin lorem |
| fr_CH| Latn | iure ratione dicta voluptas illo |standard Latin lorem |
| nl| Latn | veritatis quibusdam maxime magnam possimus |standard Latin lorem |
| pl| Latn | accusamus eaque deleniti quam distinctio |standard Latin lorem |
| pt_BR| Latn | vitae aliquid temporibus laudantium nam |standard Latin lorem |
| sk| Latn | quidem possimus corrupti odio voluptate |standard Latin lorem |
| tr| Latn | optio natus quis aspernatur molestias |standard Latin lorem |
| uz_UZ_latin| Latn | tutamen ullam magni auctor delectatio |standard Latin lorem |
| en_BORK| Latn | thees lebureeuoos gesh ooccoor injuy | real English words with spelling modifications
| lv| Latn | māxīmē vulnēro xīphīās soļ āēgrotātīo | standard Latin lorem with extra diacritics
| es_MX| Latn | Fichero Incorpóreo Basurear Engarbarse Gendarme | random real and nonsense Spanish words
| vi| Latn | yêu bè vàng ngọt độc | random real Vietnamese words

## Non-latin
(im not really enough of an expert in non-Latin languages to tell if these are nonsense or real words)

| code | script | words(5) | notes |
| --- | --- | --- | --- |
| ar| Arab | الذات اصرخ ليونة أتذكر فشيأ |
| dv| Thaa | އިންގިލާބެއް އެންމެ ތަރައްގީ މުޅިން ގެއްލުންނުވާ |
| el| Grek | nihil similique laudantium aliquid qui | standard Latin lorem (perhaps should be changed) |
| fa| Arab | تمام به پایان بلکه ستون داشت |
| he| Hebr | דולור תוק לפתיעם רוגצה קלאצי | Transliteration mimicking lorem ipsum in Hebrew
| hy| Armn | աշխարհում բոլորն մեկ հասած իրենց |
| ja| Jpan | 色々 独裁 錠 めいがら たて |
| ko| Kore | 형에 확정될 자유를 정한다. 범하고 |
| ru| Cyrl | направлений модели внедрения профессионального играет | Random real words
| ur| Arab | چاسدسد چسد ساسدبھ اسدفگبطاسدفد اسداسدھدسبابگ ابنسد |

src/definitions/lorem.ts
```diff
@@ -2,6 +2,9 @@ import type { LocaleEntry } from './definitions';
 
 /**
  * The possible definitions related to lorem texts.
+ *
+ * The words in this module are determined by the ISO 15924 script of the locale.
+ * If a locale uses the Latin script, it will utilize Latin lorem words, while a locale using the Cyrillic script will use Cyrillic lorem words, and so forth.
```
matthewmayer, line 7:
```suggestion
* If a locale uses the Latin script, it should generally utilize Latin "lorem ipsum" words, while a locale using another script should use real or nonsense words to give the same effect as Latin lorem text.
```

matthewmayer (commented):
Suggested rewrite. In some ways the concept of "lorem ipsum" is unique to Latin-based languages, because you can write words using Ancient Latin, a dead language, that are "foreign" and yet immediately obvious as "words" in most Latin-based languages.

So I'm not sure there's a definite "best" way to handle lorem-ipsum in non-Latin languages. In some languages it may make sense to transliterate Latin words into the script like "l-o-r-e-m", in other cases you may just want to us random words or characters.

ST-DDT:
**Team Proposal**

- The lorem module will always return the Latin lorem words/sentences.
- The lorem word definitions will be inlined into the module and removed from the locale data
- We will add replacement methods to the word module, that generate sentences/paragraphs with words of that locale

matthewmayer:
Would it make more sense to move the lorem definitions to the base locale?

Shinigami92:
> **Team Proposal**
> 
> * The lorem module will always return the Latin lorem words/sentences.
> * The lorem word definitions will be inlined into the module and removed from the locale data
> * We will add replacement methods to the word module, that generate sentences/paragraphs with words of that locale

I was not available at last team meeting, and I'm missing any reasons 👀 
Why will lorem always return latin?

ST-DDT:
> Why will lorem always return latin?

Because all locales that use Latin characters, use Latin anyway. And those that dont, use normal words and are sometimes even incompatible with the Latin sentence structure.

ST-DDT:
**Team Task**

Everybody should make their suggestions for the lorem module defimnitions and expectations.
We will discuss the proposals/expectations in the next team meeting.

Shinigami92:
> **Team Task**
> 
> Everybody should make their suggestions for the lorem module defimnitions and expectations. We will discuss the proposals/expectations in the next team meeting.

My expectation:

If we do not at least change the behavior of lorem, there should be a e.g. `word/lorem.blindtext` that generates randomized but localized placeholder texts ([wiki:de:Blindtext](https://de.wikipedia.org/wiki/Blindtext), [wiki:en:Filler_text](https://en.wikipedia.org/wiki/Filler_text))
_name and module is up for discussion_

ST-DDT:
@Shinigami92 Could you please elaborate on what you would consider the defining difference between the word module and the lorem module?

Shinigami92:
> @Shinigami92 Could you please elaborate on what you would consider the defining difference between the word module and the lorem module?

Personally I would say word is for generating words like nouns, verbs and so on, while lorem is more like for placeholder texts for example to test responsive table cells in a frontend.

However I did not read our docs yet and both modules were from times long before I came into the project and so I won’t like to be made responsible for any historical decisions.
Instead I would like more to find a good way in the future and not look into the past.
So I don’t care if it is called lorem, word or anything else but the functionality is provided.

ST-DDT:
> I won’t like to be made responsible for any historical decisions.

That was not my intention. I'm sorry.

---

I spend quite some time thinking about this. The following represents my personal opinion.
I'll split the answer to multiple comments in order to make it easier to react to them using emojis.
I hope the reactions (and other answers) help us (all) determine where we are one the same page and where we are not.

ST-DDT:
For me, lorem (module) is a specific type of blind text that you are not supposed to read/be able to understand.

ST-DDT:
If we define lorem to be latin, then those locales that currently return locale specific words would be impacted.
If we define lorem to be locale specific, then the locales that currently use latin would loose their original intent of creating un-understandable text.
If we not define lorem, then the locales that use latin are unable to generate (pseudo-)understandable blind texts unless we add new methods for that.

ST-DDT:
If we add the explicit concept of (pseudo-)understandable blind text, then it is likely that it will require the same or at least similar methods to the current lorem module, so that you are able to generate a string matching your length requirements.
These would cause conflicts with the existing methods and would need some form of disambiguation.
Either by prefixing them with `blind` e.g. blindText, blindSentence, or moving them to a different new module.

matthewmayer:
I think the real question is does "lorem" mean general blind text or specifically the Latin lotem ipsum text?

Does the idea of "Chinese lorem" or "Thai lorem" or "Hebrew lorem" make sense?

ST-DDT:
> Does the idea of "Chinese lorem" or "Thai lorem" or "Hebrew lorem" make sense?

Important question.
For me, Chinese lorem is just a blind text.
If it doesn't use latin (like) characters including translations it is a blind text but not lorem.
If it is a "phonetic translation", then it is just gibberish using the locales characters. not better than `faker.string.fromCharacters`.
The closest non-latin character based equivalent I can think of is the japanese DoReMi spelling training thing(?), because I assume their brain turns off as soon as it recognizes the "intro". Which kinds of leads me to "we should probably start our lorem text with lorem ipsum", but that kind of is a different feature request altogether.

@matthewmayer Could you please share your opinion on any of these questions? Ultimately we need answers/shared opinions to form any kind of consensus.

matthewmayer:
I don't really know. I think we need to try and involve some native speakers of non-Latin languages.

ST-DDT:
**Team Proposal**

- We want to make the lorem module to consist only of lorem ipsum (Latin)
- We want to make a new text (or similar) module that generates blind texts in the current locale
  - We are not sure yet whether these will be hard-coded pseudo realistic sentences or wild combinations of words that may or may not follow the normal syntax of the language (e.g. `firstName drives through city` vs `noun verb adjective` vs `word word word`)
  - The module should roughly mimic the methods in the lorem module
  - The lorem module and the blind text module should have links (at each method) between each other to help with discovery
- The actual redefinition on the lorem module is v10/not now (after we have the blind-text module)

ST-DDT:
Should we complete this discussion or is it obsolete now?

Shinigami92:
> Should we complete this discussion or is it obsolete now?

The PR now only adds some JSDoc, which might be very useful also for AI tools like Claude when adding future functionality to these files. So yeah, I think we should merge this, but there is a potential improvement from @matthewmayer we could also think about to apply.
So either apply this suggestion or reject it (best with a reason), then I can approve 👍

matthewmayer:
I think ok to merge this although I refer to my previous comment and think we might want to revisit this if we are able to get input from more people who use non-Latin languages.  

https://github.com/faker-js/faker/pull/2885#issuecomment-2143366047

xDivisionByZerox:
**Team Decision**

The lorem module will stay as plain lorem ipsum words as is.
We see the following advantages:

- Lorem is implicitly defined as Lorem Ipsum => redefining it as blind text might bring confusion to users
- we can removing the locale data and defining them as a static set in a later major version
  - tree shaking of those data will be possible when lorem is not used
  - we gain smaller overall locale data size

ST-DDT:
IMO we should delay this to v11 and do the hard change then directly.

PR #2886, opened 2024-05-08
refactor(locale): normalize science data

xDivisionByZerox:
# Description

Normalize the entries in the `science` module definitions in all locales. All locale files are untouched as every locale entry file is currently an export of a readonly array of objects. We ignore readonly arrays (from `Object.freez`) during normalization.


ST-DDT (approved):

ejcheng (approved):

xDivisionByZerox:
Does freezing the locale entries here even make sense? `Object.freeze` only prevents the array instance from beeing mutated. Since all entries in the list are objects, wouldn't they profit more from being frozen instead?

Shinigami92:
> Does freezing the locale entries here even make sense? `Object.freeze` only prevents the array instance from beeing mutated. Since all entries in the list are objects, wouldn't they profit more from being frozen instead?

I re-evaluated it and I'm a bit shocked as I thought that `Object.freeze` would be deep 🤔 but it's not

PR #2891, opened 2024-05-11
infra(deps): remove unused glob

ST-DDT:
It looks like we no longer use the `glob` dependency.

I searched for it and haven't found any usages of it, but please check as well.


Shinigami92:
> It looks like we no longer use the `glob` dependency.
> 
> I searched for it and haven't found any usages of it, but please check as well.

it was part of the build script
however, glob is part of node v22 anyways, so when we need it, we can use it from there

Shinigami92 (approved):

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2901, opened 2024-05-13
infra(unicorn): import-style

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/import-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/import-style.md) lint rule.

This rule defines how a specific module has to be imported aka `import * as foo from 'bar'` vs `import { oof } from 'bar'`.
Unfortunately, it has the opinion that you should import the entirety of `node:path`, because "they have similar functions, all likely to be utilized", which I disagree with. Since this doesn't apply to any non configured package, and the defaults are subjective (and wrong IMO), I think we should disable this.


xDivisionByZerox (approved):

.eslintrc.cjs
```diff
@@ -39,6 +39,7 @@ module.exports = defineConfig({
     'prefer-exponentiation-operator': 'error',
     'prefer-template': 'error',
 
+    'unicorn/import-style': 'off', // subjective & doesn't do anything for us
```
Shinigami92, line 42:
```suggestion
    'unicorn/import-style': [
      'error',
      { styles: { 'node:path': { named: true } } },
    ],
```

Shinigami92 (dismissed):

.eslintrc.cjs
ST-DDT: I tried that - see first commit - and I disagree.
The only thing we are using from that preset is `node:path` and there we don't use it and the other imports are totally unaffected.
We can talk about that in more detail, once I can talk again, If you would like.

Or let me ask the other way round:
- What is the benefit of enabling it that way?

Shinigami92: For example we use it here: https://github.com/faker-js/faker/blob/b27e1a8d4d0402553f6c5b0170bc371adc8f73f3/scripts/apidocs/utils/paths.ts#L1

with that rule enabled, it prefers `import { dirname, resolve } from 'node:path';` over `import path from 'node:path';`

with that rule set to off, anykind of syntax could be used

ST-DDT: Btw for that to be the case you have to set default to false.

But what about literally any other import in the entire project?

Shinigami92: If you would like, you can find a config for that
I would like to always prefer "named"-imports
I (personally) don't like the "default" of `unicorn/import-style`

ST-DDT: > If you would like, you can find a config for that

Would you like to give it a try?

> I would like to always prefer "named"-imports

Me too, but then there is the locales: https://github.com/faker-js/faker/blob/next/src/locales/cs_CZ/lorem/index.ts#L6

> I (personally) don't like the "default" of `unicorn/import-style`

You can disable them completely with: `extendDefaultStyles`

Shinigami92: I currently have set me a restriction to not actively contribute code and work on code in Open Source this week / until I'm back from USA, so I do a semi-open-source vacation.

Shinigami92: @fisker is `unicorn/import-style` not simply configurable to just and only use `named` for every import possible? Or is that rule not intended for that usecase?

fisker: > @fisker is `unicorn/import-style` not simply configurable to just and only use `named` for every import possible? 

No, not possible.


> Or is that rule not intended for that usecase?

You can send a feature request.

fisker: > What is the benefit of enabling it that way?

It can be useful to avoid different style on the same module in one repo. You can choose any style that works and you like, personally I use named/namespace for `node:path` too.

ST-DDT:
Since there is nothing we can do about it for now, I'll go ahead and merge this.
We can reconfigure this later once that feature is added.

EDIT: I don't know if you are going to create a PR, but if you do, you probably know more about how to configure it correctly anyway, so you can also create a new PR for the correct configuration.

PR #2902, opened 2024-05-15
refactor(locale): normalize date data

xDivisionByZerox:
# Description

Normalize the entries in the `date` module definitions in all locales.

I choose this module as no data set exceedes the 1k entry limit, so all locale file changes are based on entry sorting. The git diff deletion surplus comes from deleted empty lines in between object properties in the locale entries.


ST-DDT (approved):

matthewmayer (dismissed):
I think it's easier to review months, days etc if they are in their natural Jan-Feb-mar order than alphabetical.

ST-DDT:
**Team Decision**

- We will not make an exception for the date module.
- The date module contains 12+7 entries. It should be manageable to check them for completeness.
- We will add a test by using `Intl` for each locale.

Shinigami92 (approved):

ejcheng (approved):

PR #2909, opened 2024-05-18
infra(typescript-eslint): no-confusing-void-expression

ST-DDT:
Configures the [`@typescript-eslint/no-confusing-void-expression`](https://typescript-eslint.io/rules/no-confusing-void-expression/) rule.

IMO void arrow functions are easier to read than them being wrapped in brackets.

---

This is an attempt at removing some TODOs from the source.


test/modules/helpers.spec.ts
```diff
@@ -799,7 +799,8 @@ describe('helpers', () => {
         });
 
         it('should never return the callback result when probability is 0', () => {
```
xDivisionByZerox, line 801: Maybe off topic - sorry for that.
Would it be adventagious to provide an additional signature where typescript can more accuratly infer the return type based on the `options.probability` value? Something like this:

```ts
maybe<T>(fn: () => T, options: { probability: 0 }): undefined;
maybe<T>(fn: () => T, options: { probability: 1 }): T;
maybe<T>(fn: () => T, options: { probability: number }): T | undefined;
```

test/modules/helpers.spec.ts
```diff
@@ -799,7 +799,8 @@ describe('helpers', () => {
...
-          const actual = faker.helpers.maybe(() => expect.fail(), {
+          const method = expect.fail as () => string;
```
xDivisionByZerox: I have reservations about the use of type assertions Type assertions can inadvertently lead to bugs by causing a discrepancy between the expected types at runtime and during development. The current assertion seems misleading in the particularly context where `expect.fail` is actually typed to return `never` rather than `string`. 

Furthermore, considering that this project is open source, it is imperative to avoid employing workarounds that may obscure the true intent of the code. Instead of resorting to tricks for TypeScript/Linter compliance, I propose suggest `unknown` as the return type. This approach would not only alleviate the linter's confusion but also remain technically accurate without compromising the code's integrity.

```suggestion
          const method: () => unknown = expect.fail;
```

I'm sorry if this seems super nitpicky, but it's a concern important to me.

test/modules/helpers.spec.ts
ST-DDT: While I think that there is a usecase for probability: 0/1, I think it will never be explicitly typed as such.
Because if it always absent/present, then there is no need for the maybe.
And if it is just a number, that happens to be 0/1, then the additional signatures don't help.

test/modules/helpers.spec.ts
ST-DDT: Good catch. I like feedback like this as my coding style/habits are heavily influenced by Java.

Fixed

xDivisionByZerox (approved):

PR #2910, opened 2024-05-18
refactor(locale)!: use snake case for all locale data

ST-DDT:
Standardize [file naming convention](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md) for the renaming locale files.

| old                                         | replacement                                  |
| ------------------------------------------- | -------------------------------------------- |
| `faker.definitions.science.chemicalElement` | `faker.definitions.science.chemical_element` |
| `faker.definitions.system.directoryPaths`   | `faker.definitions.system.directory_paths`   |
| `faker.definitions.system.mimeTypes`        | `faker.definitions.system.mime_types`        |

---

This is an attempt an fixing some TODOs in our sources.


src/definitions/system.ts
```diff
@@ -7,12 +7,12 @@ export type SystemDefinition = LocaleEntry<{
   /**
    * Returns some common file paths.
    */
-  directoryPaths: string[];
+  directory_paths: string[];
```
matthewmayer, line 10: Should we make it singular too? Most other definitions are singular.

src/definitions/system.ts
```diff
@@ -7,12 +7,12 @@ export type SystemDefinition = LocaleEntry<{
...
 
   /**
    * The mime type definitions with some additional information.
    */
-  mimeTypes: { [mimeType: string]: SystemMimeTypeEntryDefinition };
+  mime_types: { [mimeType: string]: SystemMimeTypeEntryDefinition };
```
matthewmayer, line 15: Same

src/definitions/system.ts
ST-DDT: I created a separate issue for that (as this is the "lint-fix" part of it and there are more entries that needs to be singularized)

- #2911

xDivisionByZerox:
Why do we couple file anme and property key anyway? The propblem I have right now is that we have locale entries which might be lists of objects which intern have property keys as well. These are not considered in the PR. For example, look at [`en/airline/airline`](https://github.com/faker-js/faker/blob/7dc8a18f0b906f959670627240a6e5600bf47b75/src/locales/en/airline/airline.ts#L2) which is types as a list of [`Airline`](https://github.com/faker-js/faker/blob/7dc8a18f0b906f959670627240a6e5600bf47b75/src/modules/airline/index.ts#L17). It has the property `iataCode`, which is camelCase. So what I can do is `faker.helpers.fake('airline.airline.iataCode')` which contradictes the statement from the migration guide:

> All our locale data now use the `snake_case` naming scheme.

ST-DDT:
> Why do we couple file name and property key anyway?

Because the index files that merge them are generated.

We could transform the property keys in between, but it would be a massive breaking change. Sure we could do that, but we should talk about it first.

ST-DDT:
If you are just concerned about the wording, I can change it to refer to locale categories and entries specifically.

xDivisionByZerox:
> If you are just concerned about the wording, I can change it to refer to locale categories and entries specifically.

Thats propably it for me. Alternativly we could actually make all properties snake_case. The definition properties which are not "correctly" cased right now (ignoring the ones from this PR) are:

- `airline.airline.iataCode`
- `airline.airport.iataCode`
- `internet.http_status_code.clientError`
- `internet.http_status_code.serverError`
- `science.chemical_element.atomicNumber`

xDivisionByZerox (approved):

matthewmayer (approved):

PR #2913, opened 2024-05-18
chore(helpers): deprecate invocations on non-functions for fakeEval

ST-DDT:
This PR deprecates the current behavior of fake in preparion to be more consistent with standard JS behavior.

```ts
faker.helpers.fake('{{person.first_name()}}'); // 'Jenny'
```

Would now print a deprecation warning.

In v10 would fail with an exception:

```txt
FakerError: Cannot resolve expression 'person.first_name()'
```

---

This PR is part of an attempt to solve some of the TODOs in our source code.


docs/guide/upgrading.md
```diff
@@ -635,3 +635,19 @@ If you wanted more control over the number, it was previously necessary to pass
 The styles are locale-aware, so for example if you use pt_PT, phone numbers suitable for Portugal would be generated.
 
 2. If none of the `style`s match your needs, you can use `faker.string.numeric()` or `faker.helpers.fromRegExp()` to create a custom pattern.
+
+### Fake now Fails for Invoking Functions on Non-Functions
+
+Previously, function calls on non-functions were silently ignored.
+
+```ts
+faker.helpers.fake('{{person.first_name()}}''); // 'Jenny'
```
```suggestion
faker.helpers.fake('{{person.first_name()}}'); // 'Jenny'
```

matthewmayer (changes requested):
I don't really like introducing this change suddenly as it's likely to only be found at runtime. 

If we don't want to support this syntax any more I think we could log a warning in v9 and error in v10 instead.

ST-DDT:
**Team Decision**

- We forgot about the deprecation -> deprecate now and "remove" in v10

ST-DDT:
Fixed - ready for review

src/modules/helpers/eval.ts
```diff
@@ -133,9 +134,21 @@ function evalProcessFunction(
...
+      // TODO @ST-DDT 2023-12-11: Replace in v10
       // typeof entrypoint === 'function' ? entrypoint(...params) : undefined
-      typeof entrypoint === 'function' ? entrypoint(...params) : entrypoint
+      {
+        if (typeof entrypoint === 'function') {
+          return entrypoint(...params);
+        }
+
+        deprecated({
+          deprecated: 'Calling functions on non-functions',
```
xDivisionByZerox: Maybe:

```suggestion
          deprecated: 'Invoking expressions which are not functions',
```

Should we log the expression which got falsely invoked? That would help clearify the warning for users. Since this block is in a "map" function, it might be triggered multiple times without additional value.

Shinigami92:
Having a chore! is invalid 🤔 either it did not change runtime behavior or it is a breaking change 👀

src/modules/helpers/eval.ts
matthewmayer: I agree that the current error is probably going to be pretty confusing to end users

If I previously had code like 
```
faker.helpers.fake('{{person.first_name()}} {{person.lastName()}}'); 
```
I will probably need guidance as to what is wrong.

matthewmayer:
I think it's no longer breaking so ! Should be removed but it should get a brief write up at the bottom of the migration guide.

src/modules/helpers/eval.ts
ST-DDT: Should be fixed now.

ST-DDT:
Example log:

````txt
[@faker-js/faker]: Invoking expressions which are not functions is deprecated since v9.0 and will be removed in v10.0.
Please remove the parenthesis or replace the expression with an actual function.
person.first_name()
                 ^
````

**Review by xDivisionByZerox — DISMISSED** [review]

**Review by Shinigami92 — DISMISSED** [review]

File: src/modules/helpers/eval.ts [hunk 3]
```diff
@@ -133,9 +135,22 @@ function evalProcessFunction(
...
       // typeof entrypoint === 'function' ? entrypoint(...params) : undefined
-      typeof entrypoint === 'function' ? entrypoint(...params) : entrypoint
+      {
+        if (typeof entrypoint === 'function') {
+          return entrypoint(...params);
+        }
+
+        console.warn(
+          `[@faker-js/faker]: Invoking expressions which are not functions is deprecated since v9.0 and will be removed in v10.0.
+Please remove the parenthesis or replace the expression with an actual function.
```
matthewmayer:
```suggestion
Please remove the parentheses or replace the expression with an actual function.
```
Plural

xDivisionByZerox (approved):

Shinigami92 (approved):

matthewmayer (approved):

PR #2914, opened 2024-05-18
docs: delete CHANGELOG_old.md

ST-DDT:
Deletes the `CHANGELOG_old.md` which contains the changelog prior to v6.

IMO at this point this file is hardly relevant anymore and can be deleted.


ST-DDT:
I also considered trimming the "current" CHANGELOG.md (in a separate PR), to v8+ or v9+ with a link at the end to previous major versions.
What do you think?

xDivisionByZerox:
> I also considered trimming the "current" CHANGELOG.md (in a separate PR), to v8+ or v9+ with a link at the end to previous major versions. What do you think?

I'd prefer not to do this. I find myself looking into changelog files of other libraries. It's quite frustrating when you want to do a fuzzy search for when a specific thing was changed in the changelog file, but you only have the latest version in there. 
Sure, I could check the git history, but then I'd need to clone the repository first, which I often don't want.

Also, what would we gain by trimming the changelogs content to the current version? First thing that came to mind for me was better load time when viewing the file online. That should be neglectably low, right? 

Edit:
I don't think that my argument applies for the old changelog files, as these changes have been made by a separate repository before. So even if you'd want to find a specific changelog entry, we wouldn't be the one responsible for it.

xDivisionByZerox (approved):

ejcheng (approved):

matthewmayer:
Why not leave a link to the final revision of changelog_old at the bottom of changelog together with an explanatory note. 

If I want to understand the history of a project i look at its changelog, so that would better clarify that the project didn't start at v6.

ST-DDT:
There is a short text linking to the old one:

> For changes prior to version `v6`, see [the old changelog](https://github.com/faker-js/faker/blob/v6/CHANGELOG_old.md).

Do you want to us to explain the reason why it is cut there?
I'm not sure how to formulate that. Maybe like this?

> These changes were made before the current maintainer team took over, for more information read https://fakerjs.dev/about/announcements/2022-01-14.html#i-heard-something-happened-what-s-the-tldr

matthewmayer:
Sorry, my bad, you already made this change. This is sufficient

image

matthewmayer (approved):

PR #2919, opened 2024-05-20
docs: browser console usage guide

ST-DDT:
Include the Browser Console usage in our guide docs.

Ref:

- https://github.com/faker-js/faker/discussions/974#discussioncomment-9501192


docs/guide/usage.md
```diff
@@ -56,7 +67,7 @@ const randomEmail = faker.internet.email(); // Tomasa_Ferry14@hotmail.com
 ```
 
 ::: info Note
-It is highly recommended to use version tags when importing libraries in Deno, e.g: `import { faker } from "https://esm.sh/@faker-js/faker@v8.4.0"`.
+It is highly recommended to use version tags when importing libraries in Deno, e.g: `import { faker } from "https://esm.sh/@faker-js/faker@v8.4.1"`.
```
ST-DDT, line 77: Or should we update this to `v9.0.0-alpha.0`?

```suggestion
It is highly recommended to use version tags when importing libraries in Deno, e.g: `import { faker } from "https://esm.sh/@faker-js/faker@v9.0.0-alpha.0"`.
```

xDivisionByZerox: I would only make reference to stabile versions in our docs.

docs/guide/usage.md
```diff
@@ -46,6 +46,17 @@ For more information about changing and customizing the locales, please refer to
 Using the browser is great for experimenting 👍. However, due to all of the strings Faker uses to generate fake data, **Faker is a large package**. It's `> 5 MiB` minified. **Please avoid deploying the full Faker in your web app.**
 :::
 
+## Browser Console
```
matthewmayer: Don't think this header is needed. It can all come under the Browser header with the warning?

ejcheng: Maybe change it to just "Console"?

matthewmayer: Or make it a level 3 heading.

ST-DDT: IMO the `Browser` heading isn't optimal (That's why I put it in a different heading in the first place). 

Should we rename it to `Simple Website`?

matthewmayer: I think we should be cautious about making it seem like this is a sensible way to make a website. Doing it in the console is fine for a quick experiment but loading a 5MB JS file client-side is not a wise idea. So I think it needs to be linked to the warning above.

ST-DDT: I switched the two sections and made the html script part actually usable.

Shinigami92 (approved):

ejcheng (approved):

PR #2922, opened 2024-05-25
feat(bitcoinAddress): Multiple bitcoin address types and testnet

madoke:
The `faker.finance.bitcoinAddress()` only generates `legacy` and `segwit` addresses at random choice, and only for `mainnet`. In some cases the user might need to generate different types of addresses and use different networks. All of them are slightly different in terms of prefix, characters allowed and length.

The different types of bitcoin addresses and their prefixes are explained in the following articles:
- https://cointelegraph.com/learn/types-of-bitcoin-addresses
- https://en.bitcoin.it/wiki/List_of_address_prefixes

This PR introduces an options parameter that will allow the user to specify different types of address and also different networks. Examples:

```
const defaultBehavior = faker.finance.bitcoinAddress();
// 1TeZEFLmGPLEQrSRdAcnZLoWwYeiHwmRog  (legacy mainnet)

const bech32Address = faker.finance.bitcoinAddress({ type: 'bech32'})
// bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4 (bech32 mainnet)

const taprootTestnetAddress = faker.finance.bitcoinAddress({ type: 'taproot', network: 'testnet' })
// tb1phy9d80yes2vx72pa0vzyltvq3gr3vprp6j0jczgkxzcqk6h4vuyse0uspv (taproot testnet)
```

**Note:** The current `faker.finance.bitcoinAddress` function generates both `legacy` and `segwit` addresses, while in the current implementation with the type parameter, the default behavior is to generate `legacy` addresses only. Happy make some changes here to preserve the original behavior


src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
...
+     */
+    type?: BitcoinAddressType;
+    /**
+     * The bitcoin network (mainnet or testnet). Defaults to `mainnet`.
+     *
+     * @default 0
+     */
+    network?: BitcoinNetwork;
+  }): string {
+    const { type = 'legacy', network = 'mainnet' } = options || {};
```
ST-DDT: We usually default to `{}` in the parameters.

```suggestion
  } = {}): string {
    const { type = 'legacy', network = 'mainnet' } = options;
```

ST-DDT: I'm thinking of using random defaults here (or at least for `type`).

src/modules/finance/index.ts
```diff
@@ -22,6 +22,10 @@ export interface Currency {
   symbol: string;
 }
 
+export type BitcoinAddressType = 'legacy' | 'segwit' | 'bech32' | 'taproot';
+
+export type BitcoinNetwork = 'mainnet' | 'testnet';
```
ST-DDT: Note to self: Should we use/add an enum here?

src/definitions/finance.ts
```diff
@@ -26,4 +31,17 @@ export type FinanceDefinition = LocaleEntry<{
    * Types of transactions (e.g. `deposit`).
    */
   transaction_type: string[];
+
+  /**
+   * Specifications for generating different types of bitcoin addresses (e.g. `bech32`).
+   */
+  bitcoin_address_specs: Record<
```
ST-DDT: Note to self: Should we inline these into the module?

src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
...
+    /**
+     * The bitcoin address type (legacy, sewgit, bech32 or taproot). Defaults to `legacy`.
+     *
+     * @default 'legacy'
+     */
+    type?: BitcoinAddressType;
+    /**
+     * The bitcoin network (mainnet or testnet). Defaults to `mainnet`.
+     *
+     * @default 0
```
```suggestion
     * @default 'mainnet'
```

src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
...
-      exclude: '0OIl',
+  bitcoinAddress(options?: {
+    /**
+     * The bitcoin address type (legacy, sewgit, bech32 or taproot). Defaults to `legacy`.
+     *
+     * @default 'legacy'
+     */
+    type?: BitcoinAddressType;
+    /**
+     * The bitcoin network (mainnet or testnet). Defaults to `mainnet`.
```
```suggestion
     * The bitcoin network (`'mainnet'` or `'testnet'`).
```

src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
...
-
-    let address = this.faker.helpers.arrayElement(['1', '3']);
-
-    address += this.faker.string.alphanumeric({
-      length: addressLength,
-      casing: 'mixed',
-      exclude: '0OIl',
+  bitcoinAddress(options?: {
+    /**
+     * The bitcoin address type (legacy, sewgit, bech32 or taproot). Defaults to `legacy`.
```
```suggestion
     * The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`).
```

src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
   /**
    * Generates a random Bitcoin address.
    *
+   * @param options An options object.
+   * @param options.type The bitcoin address type (legacy, sewgit, bech32 or taproot). Defaults to `legacy`
```
```suggestion
   * @param options.type The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`). Defaults to `'legacy'`.
```

src/modules/finance/index.ts
```diff
@@ -486,23 +490,44 @@ export class FinanceModule extends ModuleBase {
...
+   * @param options.network The bitcoin network (mainnet or testnet). Defaults to `mainnet`.
```
```suggestion
   * @param options.network The bitcoin network (`'mainnet'` or `'testnet'`). Defaults to `'mainnet'`.
```

src/modules/finance/index.ts
madoke: Addressed here: https://github.com/faker-js/faker/pull/2922/commits/57729ca8043f41b777f4a1cfe89be38ad5174c67

Question: should we do the same for the `network` param ?

src/modules/finance/index.ts
madoke: We can on both. I followed the same pattern used with `EmojiType` and `HttpStatusCode` from the `internet` module, but happy to make the change

src/definitions/finance.ts
madoke: Like a static constant in the `modules/finance/index.ts` file ? They don't change with the locale, so I don't see any problem with that

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,103 @@ describe('finance', () => {
       });
 
       describe('bitcoinAddress()', () => {
+        // Utility for validating Bitcoin addresses (validator lib doesn't support taproot and testnets yet)
+        const bech32 = /^(bc1|tb1|bc1p|tb1p)[a-z0-9]{39,58}$/;
+        const base58 = /^(1|2|3|m)[A-HJ-NP-Za-km-z1-9]{25,39}$/;
+        const isBtcAddress = (address: string) =>
+          bech32.test(address) || base58.test(address);
```
ST-DDT: Could you please open a PR in validatorJS in tandem to add support for these as well?
I think the current regex based validation is enough for our purposes for now, but adding missing features in our dependencies is also important.
If you link the PR here, we can upvote and review/approve it for you as well.

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,103 @@ describe('finance', () => {
...
-           *  Therefore we take one from the total length of the address not including the first character ([13])
-           */
 
           expect(bitcoinAddress).toBeTruthy();
           expect(bitcoinAddress).toBeTypeOf('string');
-          expect(bitcoinAddress).toSatisfy(isValidBtcAddress);
-        });
+          expect(bitcoinAddress).toSatisfy(isBtcAddress);
+        });
+        it.each`
```
```suggestion
        });

        it.each`
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,103 @@ describe('finance', () => {
...
+            expectedPrefix: string;
+          }) => {
+            const bitcoinAddress = faker.finance.bitcoinAddress({
+              type,
+              network,
+            });
+
+            expect(bitcoinAddress).toBeTruthy();
+            expect(bitcoinAddress).toBeTypeOf('string');
+            expect(bitcoinAddress.startsWith(expectedPrefix)).toBe(true);
```
ST-DDT: This might produce better error messages, if we move the check into the satisfy check:

```suggestion
            expect(bitcoinAddress).toSatisfy((v) -> v.startsWith(expectedPrefix));
```

But that is optional IMO.

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,103 @@ describe('finance', () => {
...
-        });
+          expect(bitcoinAddress).toSatisfy(isBtcAddress);
+        });
+        it.each`
+          type         | expectedPrefix
+          ${'legacy'}  | ${'1'}
+          ${'segwit'}  | ${'3'}
+          ${'bech32'}  | ${'bc1'}
+          ${'taproot'} | ${'bc1p'}
+        `(
```
ST-DDT: IIRC for our other tests we usually just use tuple arrays.
You don't have to change that yet, I would like to hear the opinion from the other maintainers first.

src/modules/finance/index.ts
ST-DDT: Yeah we are currently slightly inconsistent with that in our API.
But currently I'm not sure which variant is better from a DX perspective.

src/modules/finance/index.ts
ST-DDT: How likely is it that you want the testnet / or a mixed list?

EDIT: This is a test lib, so maybe returning a test address by default is a good thing? 🤔 
I have to think about that. (Especially if validation libraries don't accept/know the testnet)

test/modules/finance.spec.ts
madoke: Sounds good, I'll try to do it later today

src/modules/finance/index.ts
madoke: From a developer's perspective I'd rather have an enum. The IDE auto completes, so I don't need to type them all the time

test/modules/finance.spec.ts
madoke: Happy to change, but imho this makes the test cases easier to read

src/definitions/finance.ts
madoke: I've changed it to be inline in the module on a `bitcoin.ts` file with all the definitions: https://github.com/faker-js/faker/pull/2922/commits/d6663e887f5210f9271cf965b9b16427b1322d6d. Let me know what you think !

src/modules/finance/index.ts
madoke: I've changed it to enum. The function call will occupy more characters, but on the other hand you have auto complete from the IDE which makes it a better developer experience in my opinion. Example:
```
faker.finance.bitcoinAddress({ type: BitcoinAddressType.Bech32, network: BitcoinNetwork.Testnet })
```
versus
```
faker.finance.bitcoinAddress({ type: 'bech32', network: 'testnet' })
```

Let me know what you think !

test/modules/finance.spec.ts
madoke: Sent them a PR as well: https://github.com/validatorjs/validator.js/pull/2406

test/modules/finance.spec.ts
xDivisionByZerox: I think this comes down to personal preference and the way one was introduced to writing this style of tests. I'd prefer keep it consistent for now.

src/modules/finance/bitcoin.ts
```diff
@@ -0,0 +1,46 @@
...
+  Bech32,
+  Taproot,
+}
+
+export enum BitcoinNetwork {
+  Mainnet,
+  Testnet,
+}
+
+export const BitcoinAddressSpecs = {
```
xDivisionByZerox: Could we explicitly define the type here? That way we can get rid of the `as Casing` type assertions. Additionally the value interface of each property is described more detailed.

```suggestion
type BitcoinAddressOptions = {
  prefix: Record<BitcoinNetwork, string>;
  length: { min: number; max: number };
  casing: Casing;
  exclude: string;
};

export const BitcoinAddressSpecs: Record<BitcoinAddressType, BitcoinAddressOptions> = {
```

src/modules/finance/index.ts
```diff
@@ -486,23 +491,56 @@ export class FinanceModule extends ModuleBase {
   /**
    * Generates a random Bitcoin address.
    *
+   * @param options An options object.
```
xDivisionByZerox:
```suggestion
   * @param options An optional options object.
```

src/modules/finance/index.ts
```diff
@@ -486,23 +491,56 @@ export class FinanceModule extends ModuleBase {
...
    * @example
-   * faker.finance.bitcoinAddress() // '3ySdvCkTLVy7gKD4j6JfSaf5d'
+   * faker.finance.bitcoinAddress() // '1TeZEFLmGPLEQrSRdAcnZLoWwYeiHwmRog'
+   *
+   * enum BitcoinAddressType { Legacy, Segwit, Bech32, Taproot }
+   * faker.finance.bitcoinAddress({ type: BitcoinAddressType.Bech32 }) // 'bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4'
+   *
+   * enum BitcoinAddressType { Legacy, Segwit, Bech32, Taproot }
+   * enum BitcoinNetwork { Mainnet, Testnet }
+   * faker.finance.bitcoinAddress({ type: BitcoinAddressType.Bech32, network: BitcoinNetwork.Testnet }) // 'tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx'
```
xDivisionByZerox: I don't think that it's required to include the enums in the `@example` section. Maybe add `@see` to point to them? Simply lsiting examples including the values should eb sufficent.

```suggestion
   * faker.finance.bitcoinAddress() // '1TeZEFLmGPLEQrSRdAcnZLoWwYeiHwmRog'
   * faker.finance.bitcoinAddress({ type: BitcoinAddressType.Bech32 }) // 'bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4'
   * faker.finance.bitcoinAddress({ type: BitcoinAddressType.Bech32, network: BitcoinNetwork.Testnet }) // 'tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx'
```

src/modules/finance/index.ts
```diff
@@ -486,23 +491,56 @@ export class FinanceModule extends ModuleBase {
   /**
    * Generates a random Bitcoin address.
    *
+   * @param options An options object.
+   * @param options.type The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`). Defaults to `'legacy'`.
```
xDivisionByZerox:
```suggestion
   * @param options.type The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`). Defaults to a random address type.
```

src/modules/finance/index.ts
```diff
@@ -486,23 +491,56 @@ export class FinanceModule extends ModuleBase {
...
+      type = this.faker.helpers.enumValue(BitcoinAddressType),
+      network = BitcoinNetwork.Mainnet,
+    } = options || {};
+    const addressSpec = BitcoinAddressSpecs[type];
+    const addressPrefix = addressSpec.prefix[network];
+    const addressLength = this.faker.number.int(addressSpec.length);
+
+    const keys = Object.keys(BitcoinAddressSpecs);
+
+    console.log(keys);
```
xDivisionByZerox: This was propably forgotten from debugging.

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,108 @@ describe('finance', () => {
...
+            const bitcoinAddress = faker.finance.bitcoinAddress({
+              type,
+            });
+
+            expect(bitcoinAddress).toBeTruthy();
+            expect(bitcoinAddress).toBeTypeOf('string');
+            expect(bitcoinAddress).toSatisfy(isBtcAddress);
+            expect(bitcoinAddress).toSatisfy((v: string) =>
+              v.startsWith(expectedPrefix)
+            );
```
xDivisionByZerox: I don't think that testing for the prefix is the right way to go here, as it is determined by the "network" property. These tests validate the "type" property. Maybe checking for invalid characters (`bitcoinSpec.[type].exclude`) or the casing (`bitcoinSpec.[type].casing`) would be more fitting as these are the only property determined by the parameters you are controlling in this specific test case. The [later test case](https://github.com/madoke/faker/blob/35faabb7a27ead186eaeffe0b8e69fe24052d427/test/modules/finance.spec.ts#L395-L428) gains more significance this way as well.

xDivisionByZerox (commented):
Great work so far! ❤️

src/modules/finance/index.ts
madoke: `npm run preflight` fails without them:
```
 FAIL  test/scripts/apidocs/verify-jsdoc-tags.spec.ts > verify JSDoc tags > finance > bitcoinAddress > 0 > verify @deprecated tag
ReferenceError: BitcoinAddressType is not defined
 ❯ test/scripts/apidocs/temp/finance/bitcoinAddress_0.ts:4:60
 ❯ test/scripts/apidocs/verify-jsdoc-tags.spec.ts:185:17
    183| 
    184|                 // Run the examples
    185|                 await import(`${path}?scope=deprecated`);
       |                 ^
    186| 
    187|                 // Verify that deprecated methods log a warning
```

The examples for `enumValue` under `src/modules/helpers/index.ts` use a similar approach. Is there any way around it ?

src/modules/finance/index.ts
madoke: Fixed: https://github.com/faker-js/faker/pull/2922/commits/69a4758bef8e40445a5bf7464ec35a9d61a14392

test/modules/finance.spec.ts
madoke: Changed to tuples: https://github.com/faker-js/faker/pull/2922/commits/4de5b7d0b662ccd0fe736c8cc4162b877ca3f965

test/modules/finance.spec.ts
madoke: We could test each address type with a dedicated regex. Check the changes in https://github.com/faker-js/faker/pull/2922/commits/d87cb1c24ee085e38ff0c90b7a8d9783799e57fa and let me know what you think. Happy to propagate them to the other tests as well

src/modules/finance/index.ts
ST-DDT: You have to export the enums in the main index.ts.
If that doesnt help, I have to check our import detection code.

test/modules/finance.spec.ts
```diff
@@ -338,47 +343,31 @@ describe('finance', () => {
...
               type,
             });
 
             expect(bitcoinAddress).toBeTruthy();
             expect(bitcoinAddress).toBeTypeOf('string');
             expect(bitcoinAddress).toSatisfy(isBtcAddress);
-            expect(bitcoinAddress).toSatisfy((v: string) =>
-              v.startsWith(expectedPrefix)
-            );
+            expect(bitcoinAddress).toSatisfy((v: string) => regex.test(v));
```
xDivisionByZerox: I suggest using `toMatch` when working with RegExp as the potential error message gives more detailed information than `toSatisfy` in this case.

```suggestion
            expect(bitcoinAddress).toMatch(regex);
```

test/modules/finance.spec.ts
xDivisionByZerox: Not exactly what I had in mind with my suggestion, but it looks very clean actually. The rexex is more straightforward to read than expected, as well. 👍

src/modules/finance/index.ts
madoke: Exporting in main `index.ts` didn't work for the raw enums.However I've noticed that other enums are following a different pattern of exporting an additional `<enum>Type` with the enum values as a string, like `Sex` and `SexType` for example. I've changed `BitcoinAddressType` and `BitcoinNetwork` to follow this pattern, and removed the imports from the `@example` tag.

test/modules/finance.spec.ts
madoke: All tests changed to use regexes and `toMatch` where possible [84c5b0e](https://github.com/faker-js/faker/pull/2922/commits/84c5b0ea714f79459ed14efd40bef89a327cab1a)

src/modules/finance/index.ts
madoke: https://github.com/faker-js/faker/pull/2922/commits/4f6f03b481537b8f47f0e2c7ba322944c981d3dc

src/modules/finance/index.ts
```diff
@@ -486,23 +493,50 @@ export class FinanceModule extends ModuleBase {
   /**
    * Generates a random Bitcoin address.
    *
+   * @param options An optional options object.
+   * @param options.type The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`). Defaults to a random address type.
+   * @param options.network The bitcoin network (`'mainnet'` or `'testnet'`). Defaults to `'Mainnet'`.
```
ejcheng:
```suggestion
   * @param options.network The bitcoin network (`'mainnet'` or `'testnet'`). Defaults to `'mainnet'`.
```

src/modules/finance/index.ts
```diff
@@ -486,23 +493,50 @@ export class FinanceModule extends ModuleBase {
...
    * Generates a random Bitcoin address.
    *
+   * @param options An optional options object.
+   * @param options.type The bitcoin address type (`'legacy'`, `'sewgit'`, `'bech32'` or `'taproot'`). Defaults to a random address type.
+   * @param options.network The bitcoin network (`'mainnet'` or `'testnet'`). Defaults to `'mainnet'`.
+   *
    * @example
-   * faker.finance.bitcoinAddress() // '3ySdvCkTLVy7gKD4j6JfSaf5d'
+   * faker.finance.bitcoinAddress() // '1TeZEFLmGPLEQrSRdAcnZLoWwYeiHwmRog'
+   *
```
```suggestion
```

src/modules/finance/index.ts
```diff
@@ -486,23 +493,50 @@ export class FinanceModule extends ModuleBase {
...
+   * faker.finance.bitcoinAddress({ type: 'bech32' }) // 'bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4'
+   *
```
```suggestion
```

src/modules/finance/index.ts
```diff
@@ -486,23 +493,50 @@ export class FinanceModule extends ModuleBase {
...
+       *
+       * @default 'mainnet'
+       */
+      network?: BitcoinNetworkType;
+    } = {}
+  ): string {
+    const {
+      type = this.faker.helpers.enumValue(BitcoinAddressFamily),
+      network = BitcoinNetwork.Mainnet,
+    } = options || {};
```
```suggestion
    } = options;
```

ST-DDT (dismissed):

ST-DDT:
Looks good to me. Thanks for your contribution ❤️

ST-DDT (dismissed):

ejcheng (dismissed):

src/modules/finance/index.ts
```diff
@@ -1,5 +1,12 @@
 import { FakerError } from '../../errors/faker-error';
 import { ModuleBase } from '../../internal/module-base';
+import {
+  BitcoinAddressFamily,
+  BitcoinAddressSpecs,
+  BitcoinNetwork,
+  type BitcoinAddressFamilyType,
+  type BitcoinNetworkType,
+} from './bitcoin';
```
Shinigami92:
```suggestion
import type {
  BitcoinAddressFamilyType,
  BitcoinNetworkType,
} from './bitcoin';
import {
  BitcoinAddressFamily,
  BitcoinAddressSpecs,
  BitcoinNetwork,
} from './bitcoin';
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,100 @@ describe('finance', () => {
...
+        });
+
+        it.each([
+          [BitcoinAddressFamily.Legacy, m_legacy],
+          [BitcoinAddressFamily.Segwit, m_segwit],
+          [BitcoinAddressFamily.Bech32, m_bech32],
+          [BitcoinAddressFamily.Taproot, m_taproot],
+        ])(
+          'should handle the type = $type argument',
+          (type: BitcoinAddressFamily, regex: RegExp) => {
```
Shinigami92: are these needed or can we use `as const` here?

```suggestion
        ] as const)(
          'should handle the type = $type argument',
          (type, regex) => {
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,100 @@ describe('finance', () => {
...
+            expect(bitcoinAddress).toMatch(regex);
+          }
+        );
+
+        it.each([
+          [BitcoinNetwork.Mainnet, [m_legacy, m_segwit, m_bech32, m_taproot]],
+          [BitcoinNetwork.Testnet, [t_legacy, t_segwit, t_bech32, t_taproot]],
+        ])(
+          'should handle the network = $network argument',
+          (network: BitcoinNetwork, regexes: RegExp[]) => {
```
Shinigami92: same here

```suggestion
        ] as const)(
          'should handle the network = $network argument',
          (network, regexes) => {
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,100 @@ describe('finance', () => {
...
+            const bitcoinAddress = faker.finance.bitcoinAddress({
+              network,
+            });
+
+            expect(bitcoinAddress).toBeTruthy();
+            expect(bitcoinAddress).toBeTypeOf('string');
+            expect(bitcoinAddress).toSatisfy(isBtcAddress);
+            expect(bitcoinAddress).toSatisfy((v: string) =>
+              regexes.some((r: RegExp) => r.test(v))
```
Shinigami92: cant these get inferred?

```suggestion
            expect(bitcoinAddress).toSatisfy((v) =>
              regexes.some((r) => r.test(v))
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,100 @@ describe('finance', () => {
...
+          [BitcoinAddressFamily.Bech32, BitcoinNetwork.Testnet, t_bech32],
+          [BitcoinAddressFamily.Taproot, BitcoinNetwork.Mainnet, m_taproot],
+          [BitcoinAddressFamily.Taproot, BitcoinNetwork.Testnet, t_taproot],
+        ])(
+          'should handle the type = $type and network = $network arguments',
+          (
+            type: BitcoinAddressFamily,
+            network: BitcoinNetwork,
+            regex: RegExp
+          ) => {
```
Shinigami92:
```suggestion
        ] as const)(
          'should handle the type = $type and network = $network arguments',
          (type, network, regex) => {
```

test/modules/finance.spec.ts
```diff
@@ -313,17 +324,96 @@ describe('finance', () => {
...
+          'should handle the network = $network argument',
+          (network, regexes) => {
+            const bitcoinAddress = faker.finance.bitcoinAddress({
+              network,
+            });
+
+            expect(bitcoinAddress).toBeTruthy();
+            expect(bitcoinAddress).toBeTypeOf('string');
+            expect(bitcoinAddress).toSatisfy(isBtcAddress);
+            expect(bitcoinAddress).toSatisfy((v) =>
```
Shinigami92: maybe `toSatisfy` supports generics?
I currently do not have vscode open

test/modules/finance.spec.ts
madoke: this one failed with a type checking error:
```
test/modules/finance.spec.ts:389:42 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.

389               regexes.some((r) => r.test(v))
```
Changing it to `expect(bitcoinAddress).toSatisfy<string>((v) =>` fixed it

test/modules/finance.spec.ts
madoke: Yes indeed. It failed ts-check after I merged. Fix just pushed

test/modules/finance.spec.ts
madoke: https://github.com/faker-js/faker/pull/2922/commits/777c8a6b64fe558d9af2e32a00cdec9319574162

src/modules/finance/index.ts
madoke: Had to fix a code formatting issue on this one: https://github.com/faker-js/faker/pull/2922/commits/2d638d3695fff0a6930949719587d9484a68b586

test/modules/finance.spec.ts
madoke: I'm curious about this change. It looks like asserting with `as const` and explicitly declaring the types produce the same effect of preventing entries with different types in the examples array. Is there any benefit on one vs the other or is it just readability ?

madoke:
Thanks for submitting suggestions folks, makes addressing comments a lot easier !

test/modules/finance.spec.ts
xDivisionByZerox: > Is there any benefit on one vs the other or is it just readability ?

The explicit declaration of parameter types serves as the definitive source of truth, ensuring that functions receive the intended arguments. This approach emphasizes the developer's intent by stating, "This function **must** accept these specific argument types." Any discrepancies will be flagged as errors, such as "type string | number is not assignable to type string."

On the other hand, employing dynamic type inference through the use of "as const" makes the data as the primary source of truth. This conveys the message, "This data is established; functions should be capable of processing it accordingly." Errors will then be highlighted within the function's implementation, for instance, "Method 'toLowerCase' does not exist on type number," (you previously expected "string" but got "string | number") indicating an unanticipated data type.

Hope that helps. 

---

With all these information, I'd actually opt to not use "as const" in this case. While it is a great tool, it's not the right use case at this instance.

src/modules/finance/bitcoin.ts
```diff
@@ -0,0 +1,72 @@
+import type { Casing } from '../string';
+
+/**
+ * The bitcoin address families
```
xDivisionByZerox: @ST-DDT are our documentation test only verifying method desciption punctuation?

```suggestion
 * The bitcoin address families.
```

src/modules/finance/bitcoin.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export enum BitcoinAddressFamily {
+  Legacy = 'legacy',
+  Segwit = 'segwit',
+  Bech32 = 'bech32',
+  Taproot = 'taproot',
+}
+
+/**
+ * The bitcoin address families
```
xDivisionByZerox:
```suggestion
 * The bitcoin address families.
```

src/modules/finance/bitcoin.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export type BitcoinAddressFamilyType = `${BitcoinAddressFamily}`;
+
+/**
+ * The different bitcoin networks
```
xDivisionByZerox:
```suggestion
 * The different bitcoin networks.
```

src/modules/finance/bitcoin.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export enum BitcoinNetwork {
+  Mainnet = 'mainnet',
+  Testnet = 'testnet',
+}
+
+/**
+ * The different bitcoin networks
```
xDivisionByZerox:
```suggestion
 * The different bitcoin networks.
```

src/index.ts
```diff
@@ -55,6 +55,10 @@ export type { DatabaseModule } from './modules/database';
 export type { DatatypeModule } from './modules/datatype';
 export type { DateModule, SimpleDateModule } from './modules/date';
 export type { Currency, FinanceModule } from './modules/finance';
+export type {
+  BitcoinAddressFamilyType,
+  BitcoinNetworkType,
+} from './modules/finance/bitcoin';
```
xDivisionByZerox, line 65: In addtion to the types we usually make the enums publically available as well. Or did I miss a comment which states that we don't want to do this?

```suggestion
export type {
  BitcoinAddressFamilyType,
  BitcoinNetworkType,
} from './modules/finance/bitcoin';
export {
  BitcoinAddressFamily,
  BitcoinNetwork,
} from './modules/finance/bitcoin';
```

xDivisionByZerox (commented):
Last questions and suggestions from my side.

src/modules/finance/bitcoin.ts
ST-DDT: Currently only modules, their methods, fakers and utilities get checked.
I couldn't find a test that checks for `.` at the end. Maybe that is a lint rule instead?

<details>
<summary>Click for the exact code</summary>

Processed units

https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/scripts/apidocs/generate.ts#L25-L32
https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/scripts/apidocs/processing/class.ts#L68-L72
https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/scripts/apidocs/processing/class.ts#L153-L157
https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/scripts/apidocs/processing/class.ts#L185-L197
https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/scripts/apidocs/processing/class.ts#L97-L107

Description test

https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/test/scripts/apidocs/verify-jsdoc-tags.spec.ts#L60-L74

</details>

test/modules/finance.spec.ts
madoke: I see the difference in the errors now, `as const` fails when we try to use the value in `faker.finance.bitcoinAddress`, not in the function declaration. It still does the job but agree that it makes sense to revert. @Shinigami92  are you ok with that ?

ST-DDT:
I'm somewhat confused by the lint error. IMO the current code looks correct to me. Maybe the two blocks need to be swapped?
e.g. enum first, then the type

Like this one:

https://github.com/faker-js/faker/blob/e6a2a9f75d0f8afde398c62c1f1fd36cd44dedb7/src/index.ts#L39-L40

src/index.ts
```diff
@@ -55,6 +55,14 @@ export type { DatabaseModule } from './modules/database';
...
+export {
+  BitcoinAddressFamily,
+  BitcoinNetwork,
+} from './modules/finance/bitcoin';
```
ST-DDT: Maybe like this?

```suggestion
export {
  BitcoinAddressFamily,
  BitcoinNetwork,
} from './modules/finance/bitcoin';
export type {
  BitcoinAddressFamilyType,
  BitcoinNetworkType,
} from './modules/finance/bitcoin';
```

test/modules/finance.spec.ts
Shinigami92: I might be trained a bit different, but for me: using types for parameters in a callback is mostly an anti-pattern and an indicator for wrongly typed library functions.
They can overshadow specific types and even **worse**, if they are miss-typed, the error is showing up at the _slightly_ wrong location ⚠️ 

good:
 

bad:

xDivisionByZerox: You use completely different types for your example. Of course you'll receive errors in that case. Widen is a string union (literally enum) to a simple string is bound to fail...
I sadly have to disagree with your statement based on my previo8s provided statement.

xDivisionByZerox: > using types for parameters in a callback is mostly an anti-pattern

Where did you get that from?

Shinigami92: > You use completely different types for your example. Of course you'll receive errors in that case. Widen is a string union (literally enum) to a simple string is bound to fail... I sadly have to disagree with your statement based on my previo8s provided statement.

exactly! that proves my point, and like I described: the error does not show up there or at the input where the values are provided but somewhere random downstream

> > using types for parameters in a callback is mostly an anti-pattern
> 
> Where did you get that from?

from my experience
reason described already above

xDivisionByZerox: I'd like to add further clarity to our discussion by expanding on my previous comment.

> You use completely different types for your example. Of course you'll receive errors in that case. Widen is a string union (literally enum) to a simple string is bound to fail...

In my earlier message, I mentioned that such a mismatch would result in failure, but I didn't elaborate on why. The example encounters errors at "wrong locations" because it disregards the specified types for function parameters. As I've previously noted: Using static parameter type definitions guarantees that functions are supplied with the intended arguments. But in that case the function needs to be cabable to handly all the required parameters as well. This concept may be easier to illustrate by defining a function first and then use it's reference as a value.

Let's revisit your "bad" example:

```ts
function verifyTestCase(type: string, regex: Regexp | null) {
  const bitcoinAddress = faker.finance.bitcoinAddress({
    type, // fails
  });

  expect(bitcoinAddress).toBeTruthy();
  expect(bitcoinAddress).toBeTypeOf('string');
  expect(bitcoinAddress).toSatisfy(isBtcAddress);
  expect(bitcoinAddress).toMatch(regex); // fails
}

it.each(
  [BitcoinAddressFamily.Legacy, m_legacy],
  [BitcoinAddressFamily.Segwit, m_segwit],
  [BitcoinAddressFamily.Bech32, m_bech32],
  [BitcoinAddressFamily.Taproot, m_taproot],
)('will work but fail in function', verifyTestCase);
```

> the error does not show up there or at the input where the values are provided but somewhere random downstream

Yes, this example will fail for the same reasons as you’ve identified in your comment. Hoewver, it should demonstarte that these errors are justified, as the function is defined to potentiall receive nullish "regex" and a "type" argument that is of type "string". This emphasises the idea I stated, that defining static prameter types focuses on the developer's intend with the function's use case. 

I hope that this makes the intend of my previous comment more clear.

---

Let me clearify that I'm not saying that using dynamic type inference is wrong! I want to hint that both, dynamic and static type definitions, have their own use case which you, as the developer, should distinguish between. That being said, calling static type definitions an "anti-pattern", although `as const` was introduced to a much later in the TS ecosystem, is a bit of an overstatement I'd say. Furthermore, providing no evidence towards that thesis is quite unfair I'd say.

Shinigami92: you are right, my "anti-pattern"-call was only related to inline-anonymous-callbacks

like for example I would **never** write `someArray.filter((element: string) => doSomething(element))`

xDivisionByZerox: A good example for useful static parameter types can be found in [TypeScript's Beginner Guide](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#explicit-types), where they encounter an upstream error after defining a function's parameter types. I have to admit that it's a really simple example, tho.

xDivisionByZerox: > you are right, my "anti-pattern"-call was only related to inline-anonymous-callbacks

Good call. In that case I somewhat missunderstood you. Apologies for that.

src/index.ts
madoke: Somehow I missed it. When I ran `pnpm run preflight` locally it actually fixed the file but it passed, so I thought everything was ok and pushed without comitting the updated file

ST-DDT: No problem. Thanks for your timely fix.

test/modules/finance.spec.ts
ST-DDT: I checked our existing code and we use the `as const` a few times in our tests (although not overly often).

I also found one instance, where we use `(options, count: number)`. It is kind of hard to specifically search for these.

IMO the usage of `as const` makes sense in these cases, because the test signature isn't really the source of truth here.
You have the data, that you want to actually use in your tests, and they don't actually change their type just because you add types to the parameters. They only change how TS represents their type using the `as const` keyword.

I also found an example that requires the `as const` and does not work with parameter types:

- with parameter types

grafik
grafik
grafik

- with as const

grafik

I couldn't think of a way that would require the other way round for our tests, but that might just be observation bias.

A few times in our code we don't need either of `as const` or parameter types (and might even remove them).

I consider parameter types useful if they clarify the intend of the method/parameters or are otherwise intentionally placed to achieve a certain goal, however with the static data right above the method I think in this case the `as const` is sufficient and shorter.

Should we discuss this further in the next team meeting?

ST-DDT (approved):

test/modules/finance.spec.ts
Shinigami92: > Should we discuss this further in the next team meeting?

As an optional bullet point if time is left over, sure 👍 

However I might ask Vitest-Team if they could think about generic-const 👀 like we did (`xy<const T>(...`)

Shinigami92 (approved):

xDivisionByZerox (approved):
Thank you for your review. It has been a pleasure to work with you.

PR #2928, opened 2024-05-27
refactor(locale): unfreeze arrays

### LINKED ISSUE #2890: Check whether complex objects should be frozen
Does freezing the locale entries here even make sense? `Object.freeze` only prevents the array instance from beeing mutated. Since all entries in the list are objects, wouldn't they profit more from being frozen instead?

_Originally posted by @xDivisionByZerox in https://github.com/faker-js/faker/issues/2886#issuecomment-2102362153_

Shinigami92:
closes #2890


ST-DDT (approved):

xDivisionByZerox:
Multiple questions from my side:
1. Although the title's humor is appreciated, could we consider revising it to more accurately reflect the content of the change? This would enhance the final result of the autogenerated changelog.
2. I'd consider this change a `refactor` over a `chore` given the potential side-effects during runtime. Would you agree?
3. While these changes are not breaking, I believe it would be prudent to include a note in the migration guide for v9, informing developers that all our locale data will be mutable in future versions. Providing developers with advance notice may prevent frustration and avoidable inquiries. What do you think?

Shinigami92:
> 1. Although the title's humor is appreciated, could we consider revising it to more accurately reflect the content of the change? This would enhance the final result of the autogenerated changelog.

done

> 2. I'd consider this change a `refactor` over a `chore` given the potential side-effects during runtime. Would you agree?

done

> 3. While these changes are not breaking, I believe it would be prudent to include a note in the migration guide for v9, informing developers that all our locale data will be mutable in future versions. Providing developers with advance notice may prevent frustration and avoidable inquiries. What do you think?

not sure, maybe we want to discuss? I think unfreeze wont affect the end-user in any way
If it would be the opposite (freezing objects) then it would have a potential negative effect if user did modify the objects

xDivisionByZerox (approved):

PR #2936, opened 2024-06-04
feat(internet): add jwt method

### LINKED ISSUE #1282: add `jwt` to `internet` module

As a developer, I use faker to "seed" my documentation examples. I was currently writing a section about the usage of the [JWT](https://jwt.io/introduction) token in my system. For visualization purposes, I wanted to provide an example JWT so developers can see what to expect. Sadly faker doesn't provide such a feature (at least not in the internet module where I would expect it).

### Suggested solution

Add a basic `jwt` function to the `internet` module that returns a string that is a JWT. So it should be "decryptable" but has a random signature.

I don't think it's necessary to input a payload object since that would get too close to a real implementation (?).

### Alternative

Currently, I use a hardcoded JWT (with incorrect signature and expired) for my documentation.

### Additional context

_No response_

---

> [!NOTE]  
> This issue has been marked with "good first issue". If you'd like to get started contributing to this project, this is the issue for you! The issue can be solved by following the acceptance criteria layed out in the [team decision comment](https://github.com/faker-js/faker/issues/1282#issuecomment-2077818380)

eLoyyyyy:
Closes #1282


src/modules/internet/index.ts
```diff
@@ -804,4 +806,111 @@ export class InternetModule extends ModuleBase {
...
+       */
+      jti?: string;
+    };
+    /**
+     * The date to use as reference point for the newly generated date.
+     *
+     * @default faker.defaultRefDate()
+     */
+    refDate?: string | Date | number;
+  }): string {
```
ST-DDT: We usually default the options object to an empty object in the method parameters.

```suggestion
  } = {}): string {
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,111 @@ export class InternetModule extends ModuleBase {
...
+    const refDate = options?.refDate ?? this.faker.defaultRefDate();
```
ST-DDT: We usually deconstruct the options object like this:

https://github.com/faker-js/faker/blob/badaa6d60593fe41a1afa0e9a8bef6ae5bb8a352/src/modules/date/index.ts#L95

I think you can apply the same to header options.

src/modules/internet/index.ts
```diff
@@ -804,4 +806,111 @@ export class InternetModule extends ModuleBase {
...
+
+    const header = {
```
ST-DDT: > We will have header and payload options that replace the default values

If the header option is present, it will replace the header that is generated by default.

````ts
const { header = { alg: this.jwtAlgorithm(), typ: 'JWT' }} = options.
````

File: src/modules/internet/index.ts [hunk 4]
```diff
@@ -804,4 +806,111 @@ export class InternetModule extends ModuleBase {
...
+      alg:
+        options && options.header && options.header.alg
+          ? options.header.alg
+          : this.faker.helpers.arrayElement(
+              this.faker.definitions.internet.jwt.alg
+            ),
```
ST-DDT: This should probably be it's own `jwtAlgorithm` method.

src/modules/internet/index.ts
```diff
@@ -804,4 +806,111 @@ export class InternetModule extends ModuleBase {
...
+      typ: 'JWT',
+    };
+
+    const iat = this.faker.date.recent({ refDate });
+    const exp = this.faker.date.soon({ refDate: iat });
+    const nbf = this.faker.date.anytime({ refDate });
+
+    const payload = {
```
ST-DDT: Same here as with the header.

ST-DDT (commented):
Thanks for contributing.

Could you please also comment in the linked issue so we can assign it to you so it easier to see that the issue is being worked on?

src/modules/internet/index.ts
```diff
@@ -23,6 +23,8 @@ export type HTTPStatusCodeType =
 
 export type HTTPProtocolType = 'http' | 'https';
 
+export type JwtAlgorithmType = 'alg';
```
ST-DDT: I'm not sure this is needed here.

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
       this.faker.definitions.internet.emoji[emojiType]
     );
   }
+
+  /**
+   * Generates a random JWT Algorithm
```
```suggestion
   * Generates a random JWT (JSON Web Token) Algorithm.
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   *
+   * @example
+   * faker.internet.jwtAlgorithm() // 'HS256'
+   * faker.internet.jwtAlgorithm() // 'RS512'
+   *
+   * @since 8.0.0
```
```suggestion
   * @since 9.0.0
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   */
+  jwtAlgorithm(): string {
```
ST-DDT, line 1035: Maybe add a `@see` link to jwt and vice versa.

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.internet.jwt.alg
+    );
+  }
+
+  /**
+   * Generates a random JWT (JSON Web Token).
+   *
+   * @param options Options object.
```
```suggestion
   * @param options The optional options object.
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   * @param options.header Header part of the token.
```
```suggestion
   * @param options.header The Header to use for the token. Defaults to a random object with the following fields: ...
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   * @param options.header.alg Algorithm header parameter.
+   * @param options.payload Payload part of the token.
```
```suggestion
   * @param options.payload The Payload to use for the token. Defaults to a random object with the following fields: ...
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   *
+   * @see https://datatracker.ietf.org/doc/html/rfc7519
+   *
+   * @example
+   * faker.internet.jwt()
+   * faker.internet.jwt({ header: { alg: 'HS256' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience', jti: 'JWT ID' }})
```
ST-DDT: IMO One example for `payload` is sufficient.

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   * @param options.payload.sub Subject claim.
+   * @param options.payload.aud Audience claim.
+   * @param options.payload.jti JWT ID claim.
+   * @param options.refDate The date to use as reference point for the newly generated date.
+   *
+   * @see https://datatracker.ietf.org/doc/html/rfc7519
+   *
+   * @example
+   * faker.internet.jwt()
+   * faker.internet.jwt({ header: { alg: 'HS256' }})
```
ST-DDT: maybe also add an example with a refDate

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+       * Header part of the token
+       */
+      header?: {
+        /**
+         * Algorithms to use.
+         *
+         * @default faker.internet.jwtAlgorithm()
+         */
+        alg?: string;
+      };
```
ST-DDT: Maybe this should accept any object?

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+         * @default faker.string.uuid()
+         */
+        aud?: string;
+        /**
+         * JWT ID claim
+         *
+         * @default faker.string.uuid()
+         */
+        jti?: string;
+      };
```
ST-DDT: Maybe this should accept any object?

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+    } = {}
+  ): string {
+    const { refDate = this.faker.defaultRefDate() } = options;
+
+    const iatDefault = this.faker.date.recent({ refDate });
+
+    const {
+      header = {
+        alg: this.jwtAlgorithm(),
+        type: 'JWT',
```
```suggestion
        typ: 'JWT',
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience', jti: 'JWT ID' }})
+   *
+   * @since 8.0.0
+   */
+  jwt(
+    options: {
+      /**
+       * Header part of the token
```
```suggestion
       * The Header to use for the token. If present, it will replace any default values.
       *
       * @default { alg: faker.internet.jwtAlgorithm(), typ: 'JWT' }
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+      header?: {
+        /**
+         * Algorithms to use.
+         *
+         * @default faker.internet.jwtAlgorithm()
+         */
+        alg?: string;
+      };
+      /**
+       * Payload part of the token.
```
```suggestion
       * The Header to use for the token. If present, it will replace any default values.
       *
       * @default { iat, ... }
```

src/modules/internet/index.ts
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+        ),
+        nbf: Math.round(this.faker.date.anytime({ refDate }).valueOf() / 1000),
+        iss: this.faker.company.name(),
+        sub: this.faker.string.uuid(),
+        aud: this.faker.string.uuid(),
+        jti: this.faker.string.uuid(),
+      },
+    } = options;
+
+    const encodedHeader = Buffer.from(JSON.stringify(header)).toString(
```
ST-DDT: The `Buffer` API seems to be missing in browsers.
Or is it just my test setup?

test/modules/internet.spec.ts
```diff
@@ -133,6 +136,15 @@ describe('internet', () => {
           protocol: 'http',
         });
     });
+
+    t.describe('jwt', (t) => {
+      t.it('noArgs', { refDate })
+        .it('with header.alg', { header: { alg: 'ES256' }, refDate })
```
ST-DDT:
```suggestion
        .it('with custom header', { header: { alg: 'ES256' }, refDate })
```

test/modules/internet.spec.ts
```diff
@@ -133,6 +136,15 @@ describe('internet', () => {
...
+        .it('with payload.iss', { payload: { iss: 'Acme' }, refDate })
+        .it('with payload.sub', { payload: { sub: 'subject' }, refDate })
+        .it('with payload.aud', { payload: { aud: 'audience' }, refDate })
+        .it('with payload.jti', { payload: { jti: 'jwt id' }, refDate });
```
ST-DDT: One is probably enough

```suggestion
        .it('with custom payload', { payload: { iss: 'Acme' }, refDate });
```

test/modules/internet.spec.ts
```diff
@@ -789,6 +801,29 @@ describe('internet', () => {
...
+          expect(jwt).toSatisfy(validator.isJWT);
+        });
+
+        it('should have 3 parts', () => {
+          const jwt = faker.internet.jwt();
+
+          const parts = jwt.split('.');
+
+          expect(parts.length).toBe(3);
+        });
```
ST-DDT:
```suggestion
          expect(jwt).toBeTypeOf('string');
          expect(jwt).toSatisfy(validator.isJWT);
        });
```

test/modules/internet.spec.ts
```diff
@@ -789,6 +801,29 @@ describe('internet', () => {
...
+      });
```
ST-DDT, line 1016: Please add a test with custom header and payload and ensure the values are actually present in the result.

````ts
const header = ...;
const payload = ...;

const actual = faker.internet.jwt({header, payload});

expect(actual).toBeTypeOf('string');
expect(actual).toSatisfy(validator.isJWT);

const parts = actual.split(3);

expect(parts[0]).toBe(Buffer.....)
expect(parts[1]).toBe(Buffer.....)
````

File: src/modules/internet/index.ts [hunk 25]
```diff
@@ -804,4 +806,127 @@ export class InternetModule extends ModuleBase {
...
+   *
+   * @example
+   * faker.internet.jwt()
+   * faker.internet.jwt({ header: { alg: 'HS256' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme', sub: 'Subject', aud: 'Audience', jti: 'JWT ID' }})
+   *
+   * @since 8.0.0
```
```suggestion
   * @since 9.0.0
```

ST-DDT (changes requested):

src/modules/internet/index.ts
eLoyyyyy: I don't think there's a `Buffer` API in browsers. Best I saw was `ArrayBuffer` https://caniuse.com/?search=ArrayBuffer

src/modules/internet/index.ts
eLoyyyyy: I agree. But I can't find any reference that I can follow that allows any object but documents the default.

ST-DDT: Have you tried `Partial<{}> & Record<string, unknown>`?
You might not need the partial since the keys are optional anyway.

eLoyyyyy: This is what I have so far:
```ts
type JwtDefaultPayload<T> = T extends undefined
  ? Record<'iss' | 'sub' | 'aud' | 'jti', unknown>
  : T
```

eLoyyyyy: > Have you tried `Partial<{}> & Record<string, unknown>`?

I'll try this. Thank you.

test/modules/internet.spec.ts
```diff
@@ -789,6 +798,36 @@ describe('internet', () => {
...
+          expect(jwt).toSatisfy(validator.isJWT);
+        });
+
+        it('should return the header and payload values from the token', () => {
+          const header = {
+            alg: faker.internet.jwtAlgorithm()
+          }
+
+          const payload = {
+            iss: faker.string.alphanumeric()
```
ST-DDT: Please add a key here that isnt part of the default key set. Same for header.

src/definitions/internet.ts
```diff
@@ -29,4 +29,6 @@ export type InternetDefinition = LocaleEntry<{
    * List of some HTTP status codes.
    */
   http_status_code: Record<HTTPStatusCodeType, number[]>;
+
+  jwt: Record<'alg', string[]>;
```
ST-DDT: Please add a minimalistic jsdocs description.

And maybe use an object instead of a record:

```suggestion
  jwt: {
    /** 
     * description
     */
    algorithm: string[]
  }
```

Not sure about whether we have to nest the jwt object here at all. I have to think about that some more.

src/modules/internet/index.ts
```diff
@@ -804,4 +804,106 @@ export class InternetModule extends ModuleBase {
...
+   * faker.internet.jwt({ payload: { iss: 'Acme' }})
+   * faker.internet.jwt({ refDate: '2020-01-01T00:00:00.000Z' })
+   *
+   * @since 9.0.0
+   */
+  jwt(
+    options: {
+      /**
+       * The Header to use for the token. If present, it will replace any default values.
+       * @default
```
```suggestion
       * The Header to use for the token. If present, it will replace any default values.
       *
       * @default
```

src/modules/internet/index.ts
```diff
@@ -804,4 +804,106 @@ export class InternetModule extends ModuleBase {
...
+       * {
+       *    alg = faker.internet.jwtAlgorithm();
+       *    typ = 'JWT'
+       * };
+       */
+      header?: Record<string, unknown>;
+      /**
+       * The Payload to use for the token. If present, it will replace any default values.
+       * @default
```
```suggestion
       * The Payload to use for the token. If present, it will replace any default values.
       *
       * @default
```

ST-DDT (commented):
Other than the following minor issues and the `Buffer` issue above, this looks good to me.

src/modules/internet/index.ts
ST-DDT: We could the following method instead of the `Buffer`, that should work for our usecase:

````ts
function toBase64Url(input: string): string {
    // Convert UTF-8 string to a Uint8Array
    const utf8Bytes = new TextEncoder().encode(input);
    // Convert Uint8Array to binary string
    const binaryString = Array.from(utf8Bytes, (byte) => String.fromCharCode(byte)).join('');
    // Base64 encode the binary string
    const base64 = btoa(binaryString);
    // Convert base64 to base64url
   return base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
}
````

Comment (ST-DDT) [reply src/modules/internet/index.ts]: @faker-js/maintainers Is there a way to fail the CI if APIs are used that aren't available in all environments?

Comment (xDivisionByZerox) [reply src/modules/internet/index.ts]: I can have a look into this. My first thought would be to set up a "mock" application in the browser that uses faker and will be tested with cypress (or other e2e frameworks). 
That would be a lot of overhead, tho. Preferably, typescript would fail when transpil8ng "to browser".

File: src/definitions/internet.ts (hunk 27 above)
Comment (eLoyyyyy) [reply src/definitions/internet.ts]: I agree. I just removed the nesting, for now.

File: src/modules/internet/index.ts (hunk 20 above)
Comment (ST-DDT) [reply src/modules/internet/index.ts]: Lets use the workaround `toBase64Url` for now.
I think we can add it as a new file to `src/internal` or `src/modules/internet`.

**Comment by Shinigami92** [comment]
is this up for review? or should I wait a bit?

File: src/modules/internet/util.ts
Comment (ST-DDT) [inline src/modules/internet/util.ts:1]: Please fix the lint issues in this file.

File: src/locales/base/internet/jwt_algorithm.ts
Comment (xDivisionByZerox) [inline src/locales/base/internet/jwt_algorithm.ts:1]: "none" would actually be a valid "alg" header parameters according to [RFC 7518 - Section 3.6](https://datatracker.ietf.org/doc/html/rfc7518#section-3.6). 
I'm not sure if we should include it. I just wanted to raise awareness. What do others think?

File: src/definitions/internet.ts [hunk 30]
```diff
@@ -29,4 +29,6 @@ export type InternetDefinition = LocaleEntry<{
    * List of some HTTP status codes.
    */
   http_status_code: Record<HTTPStatusCodeType, number[]>;
+
+  jwt_algorithm: string[];
```
xDivisionByZerox, line 33: This question is propably stupid, seeing aboth that HTTP status codes have their own entry:
Why do we put the algorithms in the locales instead of coupling them to the internet module?

While I understand that the current setup allows for custom locale overrides, I'm curious if this is a common scenario. I would rather think this being an issue in most cases. When I proposed the original prototype API I had in mind that this API should be minimal yet as closely aligned with the public standard, while maintaing randomness. This would not be guaranteed when users could override these values.

Thoughts?

src/modules/internet/util.ts
xDivisionByZerox, line 1: I'm leaning towards moving this file to `src/util/string.ts` (not sure about the file name). 

1. I don't see this file being only relevant for the internet module only (even if it is right now)
2. It may cause trouble when we split the module into separate function files in a later version (`internet/display-name.ts`, `internet/domain-name.ts`, etc).

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+       * {
+       *    iat = faker.date.recent();
+       *    exp = faker.date.soon();
+       *    nbf = faker.date.anytime();
+       *    iss = faker.company.name();
+       *    sub = faker.string.uuid();
+       *    aud = faker.string.uuid();
+       *    jti = faker.string.uuid();
+       * };
```
xDivisionByZerox: All your default value are designed to be copy'pasted if read from the doc/JSDocs preview. The current syntax would be incorrect. At least to create an object that could be applied as an argument. Same goes for the header `@default` value.

```suggestion
       * {
       *    iat: faker.date.recent(),
       *    exp: faker.date.soon(),
       *    nbf: faker.date.anytime(),
       *    iss: faker.company.name(),
       *    sub: faker.string.uuid(),
       *    aud: faker.string.uuid(),
       *    jti: faker.string.uuid(),
       * };
```

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+   * @param options.refDate The date to use as reference point for the newly generated date.
+   *
+   * @see https://datatracker.ietf.org/doc/html/rfc7519
+   * @see faker.internet.jwtAlgorithm(): For generating random JWT (JSON Web Token) Algorithm.
+   *
+   * @example
+   * faker.internet.jwt()
+   * faker.internet.jwt({ header: { alg: 'HS256' }})
+   * faker.internet.jwt({ payload: { iss: 'Acme' }})
+   * faker.internet.jwt({ refDate: '2020-01-01T00:00:00.000Z' })
```
xDivisionByZerox: Our examples usually contain an example output value. This way a developer can immediatly see what values to expect. Since one of our core focuses on Faker is good DX, I think we should provide these here as well.

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+   *
+   * @since 9.0.0
+   */
+  jwt(
+    options: {
+      /**
+       * The Header to use for the token. If present, it will replace any default values.
```
xDivisionByZerox:
```suggestion
       * The header to use for the token. If present, it will replace any default values.
```

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+       *
+       * @default
+       * {
+       *    alg = faker.internet.jwtAlgorithm();
+       *    typ = 'JWT'
+       * };
+       */
+      header?: Record<string, unknown>;
+      /**
+       * The Payload to use for the token. If present, it will replace any default values.
```
xDivisionByZerox:
```suggestion
       * The payload to use for the token. If present, it will replace any default values.
```

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+      /**
+       * The Header to use for the token. If present, it will replace any default values.
+       *
+       * @default
+       * {
+       *    alg = faker.internet.jwtAlgorithm();
+       *    typ = 'JWT'
+       * };
+       */
+      header?: Record<string, unknown>;
```
xDivisionByZerox, line 1073: We have a [`LiteralUnion`](https://github.com/faker-js/faker/blob/5f77d389a805c6527b489c699a0b517ca9b17739/src/utils/types.ts#L6-L8) that is like "any string but here are suggestion". Could we do the same for objects? That way the user **can** provide any values they want, but gets autocompletion for the most common fields? Same would apply for payload as well.

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+  ): string {
+    const { refDate = this.faker.defaultRefDate() } = options;
+
+    const iatDefault = this.faker.date.recent({ refDate });
+
+    const {
+      header = {
+        alg: this.jwtAlgorithm(),
+        typ: 'JWT',
+      },
```
xDivisionByZerox, line 1105: I'm thinking about a merge strategy here instead of overriding the entire value if provided. A user that provides this is very likely to only override the "alg" property and might likely forget the "type" property.

xDivisionByZerox (changes requested):
Good work so far 👍

src/locales/base/internet/jwt_algorithm.ts
ST-DDT: Since we don't do anything with the value and don't expect our users to do anything with it either, both would work for me.

src/definitions/internet.ts
ST-DDT: Currently we have most parts in the locales even if they are unlikely to change.
Maybe the user only supports a certain subset of algorithms, overwriting the locales would allow for reusing the default header.

If we inline this, we should probably inline other values as well.

src/modules/internet/util.ts
ST-DDT: IMO this is an internal function so it should be in the internal folder.
How about `src/internal/base64.ts`?

src/modules/internet/index.ts
ST-DDT: The question is whether the most common fields are useful for the user of the method or if an easier to read signature is more beneficial?
My point is, if the user is going to overwrite this value, they know which values they need and thus we don't have to suggest them ourselves.

src/modules/internet/index.ts
ST-DDT: Same as above, if they want specific fields, they will know which fields they want and only provide those.
If you think, that the placeholder values are important/useful to some, we could add a parameter for that:

- `default-header: 'replace' | 'merge' = random(...)`
- `default-payload: 'replace' | 'merge' = random(...)`

or maybe

- `includeHeaderDefaults: boolean = boolean()`
- `includePayloadDefaults: boolean = boolean()`

IMO there should be a way to not include the placeholder/default values.
What do you think?

src/modules/internet/index.ts
ST-DDT: Here an example of another method:

https://github.com/faker-js/faker/blob/5f77d389a805c6527b489c699a0b517ca9b17739/src/modules/lorem/index.ts#L229-L232

src/definitions/internet.ts
xDivisionByZerox: The difference would be that moving other values would result in breaking changes. These don't. You think being consistent is more valuable than having to introduce breaking changes down the road? If thats the case, feel free to resolve this conversation. 👍

src/modules/internet/util.ts
xDivisionByZerox: Interesting idea. By that definition (things from `src/util` are publically available), we would have to move `src/util/types.ts` into `src/internal/types.ts` as well, correct? (separate issue)

I'd be fine with your suggestion as it is better than mine. I'm still not super happy with the file name. Maybe we should clearify some architectur around the internal directory in a separate issue/team meeting (?).

src/definitions/internet.ts
ST-DDT: IMO this isn't about consistency. We haven't inlined the mimeTypes either (and wont do lorem) because the user might want to overwrite it with a set that better suits their needs. The same applies here.
So I don't see the necessity of a (breaking) change later either.

Should we talk about this some more in the next team meeting?

src/modules/internet/util.ts
ST-DDT: Added to the meeting notes.

src/modules/internet/index.ts
xDivisionByZerox: > IMO there should be a way to not include the placeholder/default values.

This argument makes sense to me. The current replace strategy is good enought for me.  
Having additional arguments that define the behaviour of the argument process feels incorrect to me.

xDivisionByZerox:
_Discussion point has been removed for being off topic_

---

@eLoyyyyy the test snapshots as well as some formatting needs to be updated. You can run `pnpm run preflight` to run CI locally before committing. That way you don't have to wait for results from CI. Alternativly, have a look at our [CONTRIBUTING Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#important) for more information on how to fix common problems.

ST-DDT:
> The formatting seems to be off in the documentation:
> 
> image

I dont see the problem. What would you expect?

eLoyyyyy:
> @eLoyyyyy the test snapshots as well as some formatting needs to be updated. You can run `pnpm run preflight` to run CI locally before committing. That way you don't have to wait for results from CI. Alternativly, have a look at our [CONTRIBUTING Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#important) for more information on how to fix common problems.

Updated the test snapshots. My bad for not running preflight before pushing code.

src/modules/internet/util.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @returns Base64 URL encoded string.
+ *
+ * @example const encodedHeader = toBase64Url(JSON.stringify(header));
+ */
+export function toBase64Url(input: string): string {
+  const utf8Bytes = new TextEncoder().encode(input);
+  const binaryString = Array.from(utf8Bytes, (byte) =>
+    String.fromCodePoint(byte)
+  ).join('');
```
xDivisionByZerox: A little nitpicky, but if you have time this line can be written without creating a new function scope.

```suggestion
  const binaryString = Array.from(utf8Bytes, String.fromCodePoint).join('');
```

src/modules/internet/util.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ *
```
xDivisionByZerox: I suggest to add a desciption to the function with two essential details:

1. Emphasize that this function is tailored for handling JSON Web Tokens (JWTs), as outlined in [RFC 7515 - Appendix C](https://datatracker.ietf.org/doc/html/rfc7515#appendix-C). Alternatively, consider designing the function as a pure "to base 64" converter, with character replacement handled within the JWT-specific implementation.
2. Specify that this function is platform-agnostic, suitable for both browser and Node.js environments.

xDivisionByZerox (commented):
Looks really good. I have questions/suggestions for one last time.

xDivisionByZerox:
> I dont see the problem. What would you expect?

The current documentation show this:

```ts
function jwt(
    options: {
      header?: Record<string, unknown>;
      payload?: Record<string, unknown>;
    } = {}
  ): string;
faker.internet.jwt() //...
```

I would expect standardized prettier formatting (look at the whitespaces):

```ts
function jwt(
  options: {
    header?: Record<string, unknown>;
    payload?: Record<string, unknown>;
  } = {}
): string;
faker.internet.jwt() //...
```

---

I can see this on other functions as well (example: [internet.mac()](https://next.fakerjs.dev/api/internet#mac), [number.bigInt()](https://next.fakerjs.dev/api/number.html#bigint)). Because of that I dismiss this topic from this PR and put it in our meeting notes.

src/modules/internet/util.ts
ST-DDT: Maybe add that this does basically the same as Buffer.from...toString(base64url)

src/modules/internet/util.ts
eLoyyyyy: Screenshot 2024-06-19 at 6 17 53 PM

I think it messes with the encoding. Since `String.fromCodePoint` accepts variable args, it passes both the value and the index from the `mapFn` in `Array.from`.

xDivisionByZerox: I see. Thank you for trying. I learned something new today :)

src/modules/internet/util.ts
ST-DDT: **Team Decision**

@eLoyyyyy Please move this file to `src/internal/base64.ts`

src/modules/internet/util.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * This is tailored for handling JSON Web Tokens (JWTs),
+ * as outlined in RFC 7515 - Appendix C.
```
ST-DDT: Since we label this as generic component this is no longer necessary.

```suggestion
```

ST-DDT (dismissed):
Thanks for your contribution. :heart:

src/modules/internet/index.ts
```diff
@@ -804,4 +805,104 @@ export class InternetModule extends ModuleBase {
...
+  jwt(
+    options: {
+      /**
+       * The header to use for the token. If present, it will replace any default values.
+       *
+       * @default
+       * {
+       *    alg: faker.internet.jwtAlgorithm(),
+       *    typ: 'JWT',
+       * };
```
xDivisionByZerox: Usually we don't have entire objects as defautl values. Thats why I have another question regarding the current visualization in the documentation:

image

While the default values in the displayed table are technically correct, they may appear unusual in the given context. Specifically, the dangling comma at the last property and the semicolon at the end stand out. No other value uses semicolons, which is why I find it somewhat unconventional. I#m not quite sure whether this is relevant, so what are your thoughts on this?

_The same issue applies to the payload property._

ST-DDT: Should we introduce jwtHeader and jwtPayload methods? I dont like that as it is a dynamic complex object. But we could use them if you want to add properties to our defaults instead of replacing them.

Alternatively we could just hide the specifics of the default e.g. `{ ... }`. And put the actual property names in the description or outright omit them. 🤔

xDivisionByZerox: > Should we introduce jwtHeader and jwtPayload methods?

I don't see how this is related to my question. 

> Alternatively we could just hide the specifics of the default e.g. { ... }.

I don't think thats necessary. I'm not bothered by the actual default value itself, but the formatting. Simply removing the dangling comma and semicolon would already solve this (unless the documentation generation uses different formatting). Like this:

```suggestion
       * {
       *    alg: faker.internet.jwtAlgorithm(),
       *    typ: 'JWT'
       * }
```

My question was whether anyone else is botheres by that problem, or if I'm beeing nitpicky.

ST-DDT: > > Should we introduce jwtHeader and jwtPayload methods?
> 
> I don't see how this is related to my question.

Instead of defining the whole object in there, we would simply refer to the method:

````ts
 * @default faker.internet.jwtHeader()
````

---

> My question was whether anyone else is bothered by that problem, or if I'm being nitpicky.

I like your suggested `@default` better than the representation used by this PR, even if prettier might disagree.

1) Should we handle this case on the script side instead? E.g. `defaultValue.replaceAll(/,(\n *)}/g, '$1}')`

The trailing `;` should be removed in the code as well (as we don't have that anywhere else either).

2) And should we also use `defaultValue.replaceAll(/: ([a-z0-9"'-])/g, ': $1')` (with nbsp)?

**This PR**
grafik

**With script optimizations**
grafik

---

I just noticed that the default value uses an indentation of 3 spaces instead of prettier's 2.

**For script change**
```suggestion
       * {
       *   alg: faker.internet.jwtAlgorithm(),
       *   typ: 'JWT',
       * }
```

**For source code change**
```suggestion
       * {
       *   alg: faker.internet.jwtAlgorithm(),
       *   typ: 'JWT'
       * }
```

ST-DDT:
@eLoyyyyy Could you please fix the merge conflicts?

src/internal/base64.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * This works the same as `Buffer.from(input).toString('base64')`
+ * This works the same as `Buffer.from(input).toString('base64url')`
```
ST-DDT: This lools like a merge error to me.

ST-DDT (dismissed):

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+  /**
+   * Generates a random JWT (JSON Web Token) Algorithm.
+   *
+   * @see faker.internet.jwt(): For generating random JWT (JSON Web Token).
+   *
+   * @example
+   * faker.internet.jwtAlgorithm() // 'HS256'
+   * faker.internet.jwtAlgorithm() // 'RS512'
+   *
+   * @since 9.0.0
```
```suggestion
   * @since 9.1.0
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+   *
+   * @example
+   * faker.internet.jwt()
+   * faker.internet.jwt({ header: { alg: 'HS256' }}) // 'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE3MTg2MTM3MTIsImV4cCI6MTcxODYzMzY3OSwibmJmIjoxNjk3MjYzNjMwLCJpc3MiOiJEb3lsZSBhbmQgU29ucyIsInN1YiI6IjYxYWRkYWFmLWY4MjktNDkzZS1iNTI1LTJjMGJkNjkzOTdjNyIsImF1ZCI6IjczNjcyMjVjLWIwMWMtNGE1My1hYzQyLTYwOWJkZmI1MzBiOCIsImp0aSI6IjU2Y2ZkZjAxLWRhMzMtNGUxNi04MzJiLTFlYTk3ZGY1MTQ2YSJ9.5iUgaCaFVPZ8d1QD0xMjoeJbmPVyUfKfoRQ6Njzm5MLp5F4UMh5REbPCrW70fAkr'
+   * faker.internet.jwt({ payload: { iss: 'Acme' }}) // 'eyJhbGciOiJFUzM4NCIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJBY21lIn0.syUt0GBukNac8Cn1AGKFq2SWAXWy1YIfl0uOYiwg6TZ3omAW0c7FGWY6bC7ZOFSt'
+   * faker.internet.jwt({ refDate: '2020-01-01T00:00:00.000Z' }) // 'eyJhbGciOiJFUzM4NCIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE1Nzc4MDY4NDUsImV4cCI6MTU3Nzg0NjI4MCwibmJmIjoxNTgxNTQyMDYwLCJpc3MiOiJLcmVpZ2VyLCBBbHRlbndlcnRoIGFuZCBQYXVjZWsiLCJzdWIiOiI5NzVjMjMyOS02MDlhLTRjYTYtYjBkZi05ZmY4MGZiNDUwN2QiLCJhdWQiOiI0ODQxZWYwNi01OWYwLTQzMWEtYmFmZi0xMjkxZmRhZDdhNjgiLCJqdGkiOiJmNDBjZTJiYi00ZWYyLTQ1MjMtOGIxMy1kN2Q4NTA5N2M2ZTUifQ.cuClEZQ0CyPIMVS5uxrMwWXz0wcqFFdt0oNne3PMryyly0jghkxVurss2TapMC3C'
+   *
+   * @since 9.0.0
```
```suggestion
   * @since 9.1.0
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+        ),
+        nbf: Math.round(this.faker.date.anytime({ refDate }).valueOf() / 1000),
+        iss: this.faker.company.name(),
+        sub: this.faker.string.uuid(),
+        aud: this.faker.string.uuid(),
+        jti: this.faker.string.uuid(),
+      },
+    } = options;
+
+    const encodedHeader = toBase64(JSON.stringify(header));
```
ST-DDT: You have to add a new variant to the base64.ts:

```suggestion
    const encodedHeader = toBase64Url(JSON.stringify(header));
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+    const encodedPayload = toBase64(JSON.stringify(payload));
```
```suggestion
    const encodedPayload = toBase64Url(JSON.stringify(payload));
```

src/internal/base64.ts
```diff
@@ -22,3 +22,28 @@ export const toBase64: (input: string) => string =
...
+ * @example const encodedHeader = toBase64Url(JSON.stringify(header));
+ */
+export const toBase64Url: (input: string) => string =
+  typeof Buffer === 'undefined'
+    ? (input) => {
+        const utf8Bytes = new TextEncoder().encode(input);
+        const binaryString = Array.from(utf8Bytes, (byte) =>
+          String.fromCodePoint(byte)
+        ).join('');
+        return btoa(binaryString);
```
ST-DDT: Please don't forget the `base64url` part for the non-Buffer path from your previous PR:

https://github.com/faker-js/faker/blob/d258b33f439ae99c490648343481170e210804f8/src/modules/internet/util.ts#L14-L19

ST-DDT (changes requested):

src/internal/base64.ts
```diff
@@ -22,3 +22,31 @@ export const toBase64: (input: string) => string =
...
+    ? (input) => {
+        const utf8Bytes = new TextEncoder().encode(input);
+        const binaryString = Array.from(utf8Bytes, (byte) =>
+          String.fromCodePoint(byte)
+        ).join('');
+        return btoa(binaryString)
+          .replaceAll('+', '-')
+          .replaceAll('/', '_')
+          .replaceAll(/=+$/g, '');
+      }
```
ST-DDT: IMO we can reduce the code duplication by using the existing method for the first part.

```suggestion
    ? (input) =>
        toBase64(input)
          .replaceAll('+', '-')
          .replaceAll('/', '_')
          .replaceAll(/=+$/g, '')
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+   */
+  jwtAlgorithm(): string {
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.internet.jwt_algorithm
+    );
+  }
+
+  /**
+   * Generates a random JWT (JSON Web Token).
+   *
```
ST-DDT, line 1043:
```suggestion
   *
   * Please note that this method generates a random signature instead of a valid one.
   *
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+  jwt(
+    options: {
+      /**
+       * The header to use for the token. If present, it will replace any default values.
+       *
+       * @default
+       * {
+       *    alg: faker.internet.jwtAlgorithm(),
+       *    typ: 'JWT',
+       * };
```
```suggestion
       * {
       *   alg: faker.internet.jwtAlgorithm(),
       *   typ: 'JWT'
       * }
```

src/modules/internet/index.ts
```diff
@@ -1019,4 +1020,104 @@ export class InternetModule extends ModuleBase {
...
+       * @default
+       * {
+       *    iat: faker.date.recent(),
+       *    exp: faker.date.soon(),
+       *    nbf: faker.date.anytime(),
+       *    iss: faker.company.name(),
+       *    sub: faker.string.uuid(),
+       *    aud: faker.string.uuid(),
+       *    jti: faker.string.uuid(),
+       * };
```
```suggestion
       * {
       *   iat: faker.date.recent(),
       *   exp: faker.date.soon(),
       *   nbf: faker.date.anytime(),
       *   iss: faker.company.name(),
       *   sub: faker.string.uuid(),
       *   aud: faker.string.uuid(),
       *   jti: faker.string.uuid()
       * }
```

ST-DDT (commented):
A few minor tweaks and then this is good to go.

src/modules/internet/index.ts
ST-DDT: I created https://github.com/faker-js/faker/pull/3190 to improve the visualization of the default value.

ST-DDT (approved):

Shinigami92 (approved):

ST-DDT:
@xDivisionByZerox Could you please review this PR (or dismiss your previous one)?

xDivisionByZerox (approved):

ST-DDT:
Thank you for you contribution. ❤️ 

This feature will be available in the upcoming v9.1 release of faker.

PR #2938, opened 2024-06-08
refactor(locale): group name entries by gender

### LINKED ISSUE #1677: Change file structure for gendered lists (names, prefixes etc)

xDivisionByZerox:
## Description

Groups locale name entries in the `PersonModule` by gender.

## Additional Info

Fixes #1677.

- #1677 

<details>
<summary> Script that was used to refactor locale files automatically </summary>

> ℹ️ Note
> The file being `.ts` is important, so you are able to directly `import()` other `.ts` files.

```ts
import { execSync } from 'node:child_process';
import { readdir, stat, unlink, writeFile } from 'node:fs/promises';
import { join, resolve } from 'node:path';

const localeRootDir = resolve('./src/locales');
const localeDirNames = await readdir(localeRootDir);

async function pathExist(path: string) {
  return stat(path)
    .then(() => true)
    .catch(() => false);
}

for (const localeDirName of localeDirNames) {
  const personPath = join(localeRootDir, localeDirName, 'person');
  if (!(await pathExist(personPath))) {
    continue;
  }

  const getEntryPath = (entryName: string) =>
    join(personPath, `${entryName}.ts`);

  const entryGroups = [
    ['prefix', 'female_prefix', 'male_prefix'],
    ['first_name', 'female_first_name', 'male_first_name'],
    ['middle_name', 'female_middle_name', 'male_middle_name'],
    ['last_name', 'female_last_name', 'male_last_name'],
    ['last_name_pattern', 'female_last_name_pattern', 'male_last_name_pattern'],
  ];
  for (const entryGroup of entryGroups) {
    const [mainEntry, femaleEntry, maleEntry] = entryGroup;
    const mainEntryPath = getEntryPath(mainEntry);
    const femaleEntryPath = getEntryPath(femaleEntry);
    const maleEntryPath = getEntryPath(maleEntry);

    const finalMainEntry: Record<string, unknown> = {};

    const keyPathMapList = [
      { key: 'generic', path: mainEntryPath },
      { key: 'female', path: femaleEntryPath },
      { key: 'male', path: maleEntryPath },
    ];
    for (const { key, path } of keyPathMapList) {
      if (!(await pathExist(path))) {
        continue;
      }

      const { default: defaultExportMember } = await import(`file://${path}`);
      if (!defaultExportMember) {
        continue;
      }

      finalMainEntry[key] = defaultExportMember;
    }

    // remove in second cycle in case files import each other
    for (const { path } of keyPathMapList) {
      await unlink(path).catch((): void => undefined);
    }

    if (Object.keys(finalMainEntry).length === 0) {
      continue;
    }

    await writeFile(
      mainEntryPath,
      `export default ${JSON.stringify(finalMainEntry)}`
    );
    execSync(`npx prettier --write ${mainEntryPath}`);
  }
}
```

</details>

> [!NOTE]
> The script did not account for non-applicable values. These have been manually justified in 4b86dcb60e5a7079dfea16b233861d259dd8206e. A list of affected files can be found [here](https://github.com/search?q=repo%3Afaker-js%2Ffaker+path%3Asrc%2Flocales%2F*%2Fperson%2F*prefix.ts+%22export+default+null%3B%22&type=code).


src/definitions/person.ts
```diff
@@ -1,27 +1,22 @@
 import type { LocaleEntry } from './definitions';
 
+export type NameEntry<T extends string | { value: string; weight: number }> = {
+  female?: T[];
+  generic?: T[];
+  male?: T[];
+};
```
ST-DDT: IMO we should check that if male is present, then female should be present as well.

````ts
| {
  male: T[];
  female: T[];
  generic?: T[];
}
| { generic: T[]; }
````

Comment (xDivisionByZerox) [reply src/definitions/person.ts]: How would this look with non-applicable data (with "null" definitions)? Just add `|  null` on everything or have specific unions as well?

Comment (ST-DDT) [reply src/definitions/person.ts]: IMO null does not apply to the individual elements.

Either you don't have a suffix at all in that locale or you have it.
If you have it, you have it gender specific or not.

Comment (xDivisionByZerox) [reply src/definitions/person.ts]: Thats a good argument! I'm not 100% sure if this really applies for all locales, but we can change it if we run into this problem.

File: src/definitions/person.ts [hunk 2]
```diff
@@ -1,27 +1,28 @@
 import type { LocaleEntry } from './definitions';
 
+export type NameEntry<T extends string | { value: string; weight: number }> =
```
ST-DDT: Should we name this `GenderedEntry`?

src/modules/person/index.ts
```diff
@@ -15,25 +15,19 @@ export type SexType = `${Sex}`;
...
+function selectDefinition<T extends string | { value: string; weight: number }>(
   faker: Faker,
   elementSelectorFn: (values: T[]) => string,
   sex: SexType | undefined,
-  {
-    generic,
-    female,
-    male,
-  }: { generic?: T[] | null; female?: T[] | null; male?: T[] | null },
+  nameEntry: NameEntry<T> | null | undefined,
```
ST-DDT: IMO name entry may never be null or undefined here.

src/modules/person/index.ts
```diff
@@ -100,18 +93,11 @@ export class PersonModule extends ModuleBase {
...
     return selectDefinition(
       this.faker,
       this.faker.helpers.arrayElement,
       sex,
-      {
-        generic: first_name,
-        female: female_first_name,
-        male: male_first_name,
-      },
+      this.faker.rawDefinitions.person?.first_name,
```
ST-DDT: This should now use the main definitions, since we no longer have to check whether any gendered variant is present.

```suggestion
      this.faker.definitions.person.first_name,
```

src/modules/person/index.ts
```diff
@@ -162,11 +131,7 @@ export class PersonModule extends ModuleBase {
       this.faker,
       this.faker.helpers.arrayElement,
       sex,
-      {
-        generic: last_name,
-        female: female_last_name,
-        male: male_last_name,
-      },
+      this.faker.rawDefinitions.person?.last_name,
```
ST-DDT: Same here

src/modules/person/index.ts
```diff
@@ -185,18 +150,11 @@ export class PersonModule extends ModuleBase {
...
     return selectDefinition(
       this.faker,
       this.faker.helpers.arrayElement,
       sex,
-      {
-        generic: middle_name,
-        female: female_middle_name,
-        male: male_middle_name,
-      },
+      this.faker.rawDefinitions.person?.middle_name,
```
ST-DDT: And here

src/modules/person/index.ts
```diff
@@ -336,18 +294,11 @@ export class PersonModule extends ModuleBase {
...
     return selectDefinition(
       this.faker,
       this.faker.helpers.arrayElement,
       sex,
-      {
-        generic: prefix,
-        female: female_prefix,
-        male: male_prefix,
-      },
+      this.faker.rawDefinitions.person?.prefix,
```
ST-DDT: And here

src/modules/person/index.ts
xDivisionByZerox: Since I forgot to use `faker.definitions` over `faker.rawDefinitions` in all the places, allowing `null | undefined` was necessary before. By applying all the suggested changes, this is now indeed the case. 

Done.

src/definitions/person.ts
xDivisionByZerox: I also considered having two separate typed: `*Entry` and `Weighted*Entry`. Those could be combined in a union. That way we don't have to repeat the awkward generic type in the person module files as well (in the `selectDefinition()` function).

I'm not sure about the name, but I'm not attached to the one I choose as well.

src/modules/person/index.ts
```diff
@@ -15,25 +15,19 @@ export type SexType = `${Sex}`;
...
   sex: SexType | undefined,
-  {
-    generic,
-    female,
-    male,
-  }: { generic?: T[] | null; female?: T[] | null; male?: T[] | null },
+  nameEntry: NameEntry<T>,
   type: string
 ): string {
+  const { generic, female, male } = nameEntry ?? {};
```
ST-DDT:
```suggestion
  const { generic, female, male } = nameEntry;
```

src/modules/person/index.ts
```diff
@@ -1,4 +1,4 @@
-import type { Faker } from '../..';
+import type { Faker, NameEntry } from '../..';
```
ST-DDT: Should we load the `NameEntry` directly from the definitions?
Or should we define the type here and import it into the definitions?

src/modules/person/index.ts
```diff
@@ -59,10 +53,9 @@ function selectDefinition<T>(
     } else {
       values = generic;
     }
-
-    assertLocaleData(values, `person.{${type}, female_${type}, male_${type}}`);
   }
 
+  assertLocaleData(values, `person.${type}`);
```
ST-DDT: This can no longer be absent. Because either generic or the gendered variant is required to be present.

src/modules/person/index.ts
```diff
@@ -15,25 +15,19 @@ export type SexType = `${Sex}`;
...
-    female,
-    male,
-  }: { generic?: T[] | null; female?: T[] | null; male?: T[] | null },
+  nameEntry: NameEntry<T>,
   type: string
 ): string {
+  const { generic, female, male } = nameEntry ?? {};
   let values: T[] | undefined | null;
 
   switch (sex) {
```
ST-DDT, line 27: Maybe change this switch case to:

````ts
  let values: T[];

  switch (sex) {
    case Sex.Female: {
      values = female ?? generic;
      break;
    }

    case Sex.Male: {
      values = male ?? generic;
      break;
    }

    default: {
      values = generic ?? faker.helpers.arrayElement([female, male]);
      break;
    }
  }
````

Then we can omit the check afterwards.

Maybe we can even change it to 

````ts
const values: T[] = (() => {
    switch (sex) {
      case Sex.Female: {
        return female ?? generic;
      }

      case Sex.Male: {
        return male ?? generic;
      }

      default: {
        return generic ?? faker.helpers.arrayElement([female, male]) ?? [];  // The final default is only for the types because TS fails to resolve it properly
      }
    }
  })();
````

File: src/modules/person/index.ts [hunk 12]
```diff
@@ -59,10 +53,9 @@ function selectDefinition<T>(
     } else {
       values = generic;
     }
-
-    assertLocaleData(values, `person.{${type}, female_${type}, male_${type}}`);
   }
 
+  assertLocaleData(values, `person.${type}`);
   return elementSelectorFn(values);
```
ST-DDT: With all these changes to this method it might be easier to limit it to only the gendered values selection part and just call the elementSelectorFn externally:

e.g.

````ts
function selectDefinitions<T extends string | { value: string; weight: number }>(
  faker: Faker,
  sex: SexType | undefined,
  nameEntry: NameEntry<T>
): T[] {
  const { generic, female, male } = nameEntry;

  switch (sex) {
    case Sex.Female: {
      return female ?? generic;
    }

    case Sex.Male: {
      return male ?? generic;
    }

    default: {
      return generic ?? faker.helpers.arrayElement([female, male]) ?? [];
    }
  }
}
````

And then call it like this:

````ts
firstName(sex?: SexType): string {
  return this.faker.helpers.arrayElement(selectDefinitions(
    this.faker,
    sex,
    this.faker.definitions.person.first_name,
  ));
}
````

File: src/definitions/person.ts (hunk 2 above)
Comment (ST-DDT) [reply src/definitions/person.ts]: Is there a reason why you limit T to `string | { value: string; weight: number }`?
Auto completion?
IMO we could omit the type specifics here (since you have to repeat them when you use them anyway.
If you remove the type definition here, you can also simplify the signature of `selectDefinition`.

IMO we can create alias types for `NameEntry<string>` e.g. `SimpleStringNameEntry` vs `WeightedStringNameEntry`..., that way we have the easy usage of `NameEntry<T>` for `selectDefinition` and easier writing of the definitions.
I wouldn't export the alias types though.

For some types (used by modules) the types are defined in the module and the definitions import them from there.
Should we do the same here as well?

File: src/modules/person/index.ts [hunk 13]
```diff
@@ -13,57 +12,37 @@ export type SexType = `${Sex}`;
...
-    if (female != null && male != null) {
-      values = faker.helpers.arrayElement([female, male]);
-    } else {
-      values = generic;
+      return (
+        generic ??
+        faker.helpers.arrayElement([female, male]) ??
+        // The last statement should never happen at run time since at this point in time
+        // the name entry satisfies at least (generic || (female && male)).
+        // Sadly, TS is not able to infer the type correctly.
```
Shinigami92: **polish (non-blocking):** Let's try to avoid negative connotated comments

```suggestion
        // TS is not able to infer the type correctly.
```

src/definitions/person.ts
Shinigami92: > Should we name this `GenderedEntry`?

Would it be `SexNameEntry`? 😅 Because we use the sex-type to distinguish between them.
However I don't really like the naming `SexNameEntry`

Shinigami92 (commented):
I did not look into the 606 locales changes... The rest looks already really good to me, just some open questions about naming and wording

src/definitions/person.ts
ST-DDT: > Would it be `SexNameEntry`?

I might have intentionally ignored that difference.

How about

- `SexVariedEntry`
- or `EntryVariedBySex`
- or `SexBasedEntry`
- or `SexSpecificEntry`?

Shinigami92: All of them sound somehow correct, should we discuss in next team meeting?

ST-DDT (dismissed):
Looks good to me, only the following things need to be discussed:

- Name of the helper type
- Location of the type definition 
- Exact generic shape of the type

ST-DDT:
Oh I'm sorry, looks like I caused a merged conflict 🙇‍♂️

xDivisionByZerox:
> Oh I'm sorry, looks like I caused a merged conflict 🙇‍♂️

All good, happens.I resolved them to the best of my abilities.

xDivisionByZerox:
# TODO

- [ ] Fix patterns where "company.legal_entity" reflects "company.prefix" instead of "company.suffix"

src/definitions/person.ts
ST-DDT: **Team Decision**

We will use the following:

```ts
// src/definitions/person.ts
export type PersonEntryDefinition<T> = {
  generic: T[];
  male: T[];
  female: T[];
}
type SimplePersonEntryDefinition = PersonEntryDefinition<string>;
type WeightedPersonEntryDefinition = PersonEntryDefinition<{value: string, weight: number}>;
````

```ts
// src/modules/person/index.ts
selectDef<T>(entry: PersonEntryDefinition<T>): T[];

firstName(sex?: SexType): string {
  const def: string[] = selectDefinition(
    this.faker,
    sex,
    this.faker.definitions.person.first_name
  );
  return this.faker.helpers.arrayElement(def);
}
```

src/modules/person/index.ts
```diff
@@ -13,57 +13,37 @@ export type SexType = `${Sex}`;
...
-  {
-    generic,
-    female,
-    male,
-  }: { generic?: T[] | null; female?: T[] | null; male?: T[] | null },
-  type: string
-): string {
-  let values: T[] | undefined | null;
-
+  nameEntry: PersonEntryDefinition<T>
```
ST-DDT:
```suggestion
  personEntry: PersonEntryDefinition<T>
```

src/modules/person/index.ts
```diff
@@ -13,57 +13,37 @@ export type SexType = `${Sex}`;
...
  *
  * @param faker Faker instance.
- * @param elementSelectorFn The method used to select the actual element.
  * @param sex Sex.
- * @param param2 Definitions.
- * @param param2.generic Non-sex definitions.
- * @param param2.female Female definitions.
- * @param param2.male Male definitions.
- * @param type Type of the definition.
+ * @param nameEntry Definitions.
```
ST-DDT:
```suggestion
 * @param personEntry The locale data to select from.
```

Shinigami92 (approved):

ST-DDT (approved):

PR #2960, opened 2024-06-20
test: remove usage of vitest.todo

ST-DDT:
This removes all usages of `vitest.todo`.

- The disabled test is obsolete and there is no point in enabling it again at this time.
- The seededTest-factories' todo method isn't in use and I don't see a reason why it should be used for any new features going forward.


Shinigami92 (approved):
not, so sure about this, but wont block

xDivisionByZerox (approved):

PR #2964, opened 2024-06-22
fix: move utils to the correct place

### LINKED ISSUE #2961: Move utils to the correct place
Based on our team meeting:

We will put internal utilities in `src/internal/*.ts`
We will put external utilities in `src/utils/*.ts`
External utilities will be exported from the root of our library.
External utilities are not required to use a faker seed.

TODOs:

- Move generateMersenne*Randomizers() to src/utils
- Move src/utils/types.ts to src/internal/

watercubz:
### Changes Made

- **Moved Functions**: Moved `generateMersenne*Randomizers()` functions to the `src/utils` directory.
- **File Moved**: Moved `src/utils/types.ts` to the `src/internal` directory.

### Additional Details

- Updated imports and references throughout the project to reflect the new locations.
- These changes ensure better organization of internal and external utilities as discussed in issue #2961.

### Screenshots

Attached screenshots showing the new directory structure and updated imports.
image
image
image
image

### Related Issue

- [Link to Issue #2961](https://github.com/faker-js/faker/issues/2961)


ST-DDT:
@watercubz Please also add a comment to the issue so that we can assign it to you for better visibility.

src/internal/mersenne.ts
ST-DDT, line 1: Please also move the

https://github.com/faker-js/faker/blob/05237e42ba25ef27af162b00b12115b74a99afe1/src/internal/mersenne.ts#L333

And the 53bit variant to src/utils/mersenne.ts

src/utils/randomizer.ts
ST-DDT, line 1: Not sure about this one. @ others: What do you think?

watercubz: Of course, if it doesn't convince you, let me change it and add the other requests you told me.

xDivisionByZerox: > Not sure about this one. @ others: What do you think?

From the issue:

> External utilities [(`src/utils/*.ts`)] are not required to use a faker seed.

I believe we need to clarify our definition of "utilities." Is a type/interface always considered a utility, or only if it is generic? Is the Randomizer even a utility in the first place or a core building block of our library?

Additionally, I'd like to discuss the possibility of creating a separate directory, `src/util/randomizer`. This directory could house the randomizer interface and specific randomizer adapters (e.g., mersenne, pure-rand, ...). If this question isn't directly related to the current topic, feel free to disregard it.

Shinigami92:
I think `src/internal/module-base.ts` should also moved to `src/utils/module-base.ts` if not even to `src/modules/module-base.ts` 🤔

src/utils/randomizer.ts
Shinigami92: IMO the interface for `Randomizer` should be exposed, because it is the return type of `generate*Randomizer`.  
So users can write `const randomizer: Randomizer = generateMersenne32Randomizer();`

ST-DDT:
**Team Decision**

- The randomizer will get their own folder because there will be multiple adapter implementations in the future. Since they are publically available they should NOT be located in the internal directory.
    - `src/randomizer.ts` -> `src/randomizer/types.ts`
    - `src/internal/mersenne.ts` {generateMersenne32Randomizer, generateMersenne53Randomizer} -> `src/randomizer/mersenne.ts`
- These types are not exported and thus should be located in `src/internal`
    - `src/utils/types.ts` -> `src/internal/types.ts`
    - `src/internal/module-base.ts` will not be moved

src/utils/mersenne.ts
Shinigami92, line 1: **issue (blocking):** MersenneTwister19937 is an internal class and should stay in internal

ST-DDT: Clarification: the mersenne class is internal, the generateXRandomizer utility functions are public.

watercubz: Well, then mersenne goes internal, and randomizer would go outside of utils and types in internal along with mersenne, would that be the correction to the problem?

ST-DDT: Does this answer your question:

- https://github.com/faker-js/faker/pull/2964#issuecomment-2195210180

watercubz: Of course I think I already understood the point, thank you very much, I will upload changes to those previously mentioned in a moment.

Shinigami92:
@watercubz is this ready for review again? You mentioned in https://github.com/faker-js/faker/pull/2964#discussion_r1660220037 that you want to upload something, but I cannot see further commits from there. Maybe you missed something?

If it is ready for review again, please click the "Ready for review" button so we get notified.

xDivisionByZerox:
Superseded by #3141.
Nevertheless, thank you for your contribution and work provided in this PR.

PR #2971, opened 2024-06-22
docs: add npm to social links

xDivisionByZerox:
# Description

Add a NPM social icon to the navigation of our documentation page.

image


ST-DDT (approved):

ST-DDT:
Not sure which social links we need.

Shinigami92 (approved):

xDivisionByZerox:
> Not sure which social links we need.

You think a user might get overwhelmed by the amount of icons if we would put in all applicable icon?

matthewmayer:
I think that is a concern especially for less well known ones like OpenCollective.

matthewmayer:
I think 6 is tolerable. I would try to avoid 7 or more.

PR #2975, opened 2024-06-23
refactor(locale): rename company affix files

### LINKED ISSUE #1574: Rename company.suffix definition to company.legal_entity_type

ST-DDT:
Fixes #1574

- #1574

This PR renames the company `prefix` and `suffix` files to `category`/`legal_entity_type` respectively.


ST-DDT:
I was unable to identify which part is the category or legal_entity_type in `lv` and `ru`.
Maybe they have the `legal_entity_type` sometimes as prefix and sometimes as suffix?

- https://github.com/faker-js/faker/tree/next/src/locales/lv/company
- https://github.com/faker-js/faker/tree/next/src/locales/ru/company

@serembon @salixzs: Are you able to clarify this?

I contacted you as creators of these commits:

- https://github.com/faker-js/faker/commit/cf64d76c757597a3d1efbdc54a269389ecaec8e7
- https://github.com/faker-js/faker/commit/81171c9f9f2e548f9a86eaae3683ead7e28dbd72

serembon:
@ST-DDT hello! [prefix](https://github.com/faker-js/faker/blob/next/src/locales/ru/company/prefix.ts) can only be  `legal_entity_type `

ST-DDT:
> @ST-DDT hello! [prefix](https://github.com/faker-js/faker/blob/next/src/locales/ru/company/prefix.ts?rgh-link-date=2024-06-23T21%3A52%3A17Z) can only be `legal_entity_type `

@serembon Thanks for your fast response. What would you call the suffix in that case?

serembon:
This is part of the company name. It is necessary to save this [name_patterns](https://github.com/faker-js/faker/blob/next/src/locales/ru/company/name_pattern.ts) for generating a name from parts

```
{{company.suffix}}{{company.suffix}}'
{{company.suffix}}{{company.suffix}}{{company.suffix}}
```

If you use only one part, the company name will not be similar to the real one.

ST-DDT:
So the suffix is actually the suffix in this case?

serembon:
> category

Right

serembon:
> So the suffix is actually the suffix in this case?

Right

ejcheng (approved):

Shinigami92 (approved):

xDivisionByZerox:
I just noticed that we missed the update of [`CompanyDefinition`](https://github.com/faker-js/faker/blob/49d7119025d252eecb043647b7d4465352938157/src/definitions/company.ts) in this PR.
@ST-DDT Was this intended, since "suffix" was never part of it in the first place?

ST-DDT:
> @ST-DDT Was this intended, since "suffix" was never part of it in the first place?

It was never part of it in the first place (It is only usd by fake patterns).

PR #2977, opened 2024-06-27
chore: sort dependencies

ST-DDT:
Not sure whether this is a chore or infra.


xDivisionByZerox (approved):
Dependency updates usually gets the topic `chore(deps)` from renovate. I think this should be a chore.

Shinigami92 (approved):
/braindump on

If it affects src, it normally is a chore for me. Otherwise it can be an infra.
However @xDivisionByZerox is also right 🤔

IMO they are not showing up in the changelog md anyway, and so it is not so relevant 

Also this is only changing order and not affecting src indirectly as a dep update could do...

/braindump off

PR #2981, opened 2024-06-30
infra(release): prepare release PR

ST-DDT:
This implements the first step of our automated release pipeline.

Tasks:

- Checkout Branch
- Install pnpm
- Setup node
- Install deps
- Run (pre-)release
- Switch and push branch
- Create draft PR

---

grafik

- Example-Beta-Release: 
  - https://github.com/ST-DDT/faker/actions/runs/9735099427/job/26863912275
  - https://github.com/ST-DDT/faker/pull/8
- Example-Stable-Release:
  - https://github.com/ST-DDT/faker/actions/runs/9735099757/job/26863912925
  - https://github.com/ST-DDT/faker/pull/9

---

### Automated Release Pipeline PRs:

1. https://github.com/faker-js/faker/pull/2981
2. https://github.com/faker-js/faker/pull/2988
3. https://github.com/faker-js/faker/pull/2990
4. https://github.com/faker-js/faker/pull/2991

PR #2988, opened 2024-07-01
infra(release): check release PR

ST-DDT:
This implements the second step of our automated release pipeline.

Tasks:

- Checkout faker
- Checkout playground
- Install pnpm
- Setup node
- Install deps - faker
- Build - faker
- Install deps - playground
- Build - playground
- Test - playground

---

Demo PRs:

- Release PR
  - https://github.com/ST-DDT/faker/pull/15
  - https://github.com/ST-DDT/faker/actions/runs/9742754375/job/26884700385?pr=15
- Non-Release PR
  - https://github.com/ST-DDT/faker/pull/14
  - https://github.com/ST-DDT/faker/actions/runs/9742749787/job/26884685270?pr=14

---

### Automated Release Pipeline PRs:

1. https://github.com/faker-js/faker/pull/2981
2. https://github.com/faker-js/faker/pull/2988
3. https://github.com/faker-js/faker/pull/2990
4. https://github.com/faker-js/faker/pull/2991


.github/workflows/check-release-pr.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Build - faker
+        run: |
+          cd faker
+          pnpm run build
+
+      - name: Install deps - playground
+        run: |
+          cd playground
+          sed -i 's/overrides-for-release/overrides/g' package.json
```
xDivisionByZerox: Is the global modifier really required here? 🤔

.github/workflows/check-release-pr.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  pull_request:
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  check_release_pr:
+    if: startsWith(github.head_ref, 'chore/release/')
```
xDivisionByZerox, line 14: This would mean that the action is triggered every time something happens to  a PR, but get blocked by this condition.
Could re not preamtivly block the scheduling of the action [based on  the branch name](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request)?

```yml
on:
  pull_request:
    branches:
      - 'chore/release/**'
```

ST-DDT: > To run a job based on the pull request's head branch name (as opposed to the pull request's base branch name), use the github.head_ref context in a conditional.

I'll try using a file based filter instead.

.github/workflows/check-release-pr.yml
ST-DDT: No, it is not, removed.

xDivisionByZerox (approved):

Shinigami92 (approved):
I would potentially have some questions, but I think I can approve it for now

e.g.:

- should we think about to always run this check? I mean for any PR 🤔 
- why should it only run on modified changelog.md

ST-DDT:
> * should we think about to always run this check? I mean for any PR 🤔

We could do that but it has a side effect, that all features that are used in the playground that might experience breaking changes need to be updated to the new feature before it is available as a dependency. This is also the case for the releasenPR but that those are even rarer.
In the end it doesnt matter.

ST-DDT:
> * why should it only run on modified changelog.md

There is no filter for release branches and that file is the best indicator for a release branch afaict.

PR #2989, opened 2024-07-01
chore: remove unused function mergeArrays

ST-DDT:
The `mergeArrays` function was needed as a workaround to create the generic files for first and last names.
Those have become obsolete now and thus we don't need this internal utility function anymore.

Ref:

- https://github.com/faker-js/faker/pull/2938


xDivisionByZerox (approved):
Not sure wether we might need it again in the future. I'll still approve, since we have the git history.

xDivisionByZerox:
For future refactor: Is there a TS config option to notify about dead code when resolving the main entry point? Should we look into that?

matthewmayer (approved):

Shinigami92:
> For future refactor: Is there a TS config option to notify about dead code when resolving the main entry point? Should we look into that?

I don't think there is.
It would need to traverse every .ts (tsx, dts, mts and their js and even Vue equivalents, and those I forgot)

e.g. a project could contain 1000 src files and there is a .vue file that imports a .ts file. You would need the whole dependency graph.

And even then, your project could have multiple entry points.
Feels like an impossible task 😅

Maybe tsup, esbuild or eslint could tackle such tasks. But bundlers won't affect the src code. And the scope for eslint is also questionable.

Shinigami92 (approved):

ST-DDT:
Code coverage ftw.

Shinigami92:
> Code coverage ftw.

#yes

however, when you have covered code with tests, it does not tell you if it is already used by the source 😉

PR #2990, opened 2024-07-01
infra(release): draft release

ST-DDT:
This implements the third step of our automated release pipeline.

Tasks:

- Checkout faker
- Extract release version
- Release Args - Prerelease
- Release Args - Release
- Create draft release

---

Demo: `chore(release): 9.0.0-beta.0`

- https://github.com/ST-DDT/faker/actions/runs/9752132257/job/26915082301

grafik

---

### Automated Release Pipeline PRs:

1. https://github.com/faker-js/faker/pull/2981
2. https://github.com/faker-js/faker/pull/2988
3. https://github.com/faker-js/faker/pull/2990
4. https://github.com/faker-js/faker/pull/2991

PR #2591, opened 2023-12-28
ci: publish workflow

Shinigami92:
This is a GitHub action workflow that let's maintainers of Faker run a npm publish process, so that the package `@faker-js/faker` will be published to npm registry.

It is respecting the version and dist-tag by reading the package.json.

The process will be:
1. Someone create a normal release PR
2. The release PR gets merged into next branch
3. Someone hits the publish workflow in the github actions menu `// <- this was done locally before`
4. After the release succeeded, someone will normally create the GitHub Release

It requires that one of the maintainers create a new access token on NPM for the `@faker-js/faker` package, or for the `@faker/*` orga-scope.
This then needs to be configured in GitHub as a secret with the name `NPM_AUTH_TOKEN`, so it can be picked up by the CI.

Note that `publish` will run the `prepublishOnly` script that is configured in the `package.json`

This workflow will NOT "craft" a release, nor will it create tags/releases on github.
It just and only automates the publishing to npm.

Hint: the workflow is tested for 3 months now in https://github.com/salsita/node-pg-migrate

PR #2991, opened 2024-07-02
infra(release): publish release

ST-DDT:
This implements the forth/last step of our automated release pipeline.

Tasks:

- Checkout faker
- Install pnpm
- Setup node version
- Install Deps
- Gather release info
- Prepare Readme
- Publish to npm registry
- Set dist-tag
- Push to release branch (e.g. v9)

---

Demo-Workflow run:

- https://github.com/ST-DDT/faker/actions/runs/9768753513/job/26966889916

Demo Readme:

> ## 📙 API Documentation
>
> - [Getting Started Guide](https://v9.fakerjs.dev/guide/)
> - [API Reference](https://v9.fakerjs.dev/api/)
> 
> ## 🚀 Features

---

### Automated Release Pipeline PRs:

1. https://github.com/faker-js/faker/pull/2981
2. https://github.com/faker-js/faker/pull/2988
3. https://github.com/faker-js/faker/pull/2990
4. https://github.com/faker-js/faker/pull/2991

Supersedes (as requested):

- #2591


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,65 @@
...
+    types: [published]
+
+permissions:
+  contents: write # to push the release branch
+
+jobs:
+  publish:
+    name: Publish Release
+    runs-on: ubuntu-latest
+    steps:
```
ST-DDT, line 15:
```suggestion

    steps:
```

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,66 @@
...
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Set latest/next dist-tag
+        run: |
+          pnpm dist-tag add @faker-js/faker@$RELEASE_VERSION $DIST_TAG
+
+      - name: Push to Release Branch
+        run: |
+          git push origin $GITHUB_REF_NAME:v$RELEASE_MAJOR
```
Shinigami92, line 66: **question (if-minor):** why is this needed?

ST-DDT: This pushes the `next` content to the `v9` branch.

Shinigami92: 1. what happens when we switch to v10 but there is not v10 branch yet? will it create automatically?
2. will this be done also for non-stables? e.g. v9.1.0-alpha.0

ST-DDT: > 1. what happens when we switch to v10 but there is not v10 branch yet? will it create automatically?

Yes

> 2. will this be done also for non-stables? e.g. v9.1.0-alpha.0

Yes, we didn't have that case before so I haven't thought about it.

Shinigami92 (approved):

.github/workflows/publish-release.yml
ST-DDT: Clarification: This would only be relevant for consideration if we do minor pre-releases.
(major pre-releases aren't affected, because there is no stable release that we could show instead.
We didn't have minor pre-releases before so I haven't implemented any checks for that.
If we ever do unstable minor releases, we can then decide whether the related branch should point to the stable release or the latest release (and either add the checks or do manual reverse pushes to that branch).

xDivisionByZerox (approved):

PR #3002, opened 2024-07-08
chore: small fixes and formatting

ST-DDT:
Extracted the formatting changes and small fixes from #2798

- #2798


docs/.vitepress/versions.ts
```diff
@@ -16,6 +16,7 @@ function readOtherLatestReleaseTagNames(): string[] {
     // Only consider tags for our deployed website versions
     .filter((tag) => semver.major(tag) >= 6)
     // Find the latest tag for each major version
+    // eslint-disable-next-line unicorn/no-array-reduce
```
xDivisionByZerox: We could rewrite this to only do one interation instead of the current 3 AND get rid of the reduce.

```ts
const tags = execSync('git tag -l').toString('utf8').split('\n');
const latestTagByMajor = {};
for (const tag of tags) {
  if (!semver.valid(tag) || semver.major(tag) < 6) {
    continue;
  }

  const majorVersion = semver.major(tag); // this is already done in the if block, could be refactored
  const latestTag = latestTagByMajor[majorVersion];
  if (latestTag == null || semver.lt(latestTag, tag)) {
    latestTagByMajor[majorVersion] = tag;
  }
}

return Object.values(latestReleaseTagNames).sort(semver.rcompare);
```

ST-DDT: Done

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #3020, opened 2024-07-15
fix(locale): correction of pt_PT location and person data

jeanpierrecarvalho:
Issue #3020: Correction of pt_PT Information

**Changes Made:**

Corrected country, region names an person info in Portuguese (pt_PT) in the data file.

**Details of Changes:**

- Updated the list of country names to use proper Portuguese translations.
- Ensured consistency in the use of special characters and accents according to the pt_PT locale.
- Verified and corrected abbreviations for various titles and street types in Portuguese.
- Update last names and prefix info in person
 
**Additional Details:**

The corrections were made to improve localization and ensure accurate representation of country names and titles in Portuguese.


jeanpierrecarvalho:
Hi @ST-DDT, just recreated this one. Can you have a look, please?

src/locales/pt_PT/person/last_name.ts
```diff
@@ -101,5 +101,25 @@ export default {
...
+    'Falcão',
+    'Amado',
+    'Fraga',
+    'Serra',
+    'Mariz',
+    'Mesquita',
+    'Meireles',
+    'Lourinho',
+    'Fidélis',
+    'Oliva',
```
ST-DDT: Please sort the file alphabetically.

ST-DDT (changes requested):

src/locales/pt_PT/person/last_name.ts
jeanpierrecarvalho: Hi @ST-DDT, sure! I will do it!

jeanpierrecarvalho: @ST-DDT done ✅

Shinigami92 (approved):

ST-DDT (approved):

ejcheng (approved):

src/locales/pt_PT/person/prefix.ts
```diff
@@ -1,5 +1,40 @@
...
+    'Sra.', // Senhora (Mrs.)
+    'Arq.', // Arquiteto (Architect, male)
+    'Arq.ª', // Arquiteta (Architect, female)
+    'Adv.', // Advogado (Lawyer, male)
+    'Adv.ª', // Advogada (Lawyer, female)
+    'Enf.', // Enfermeiro (Nurse, male)
+    'Enf.ª', // Enfermeira (Nurse, female)
+    'Téc.', // Técnico (Technician, male)
+    'Téc.ª', // Técnica (Technician, female)
+  ],
```
xDivisionByZerox, line 19: @faker-js/maintainers Wasn't `generic` supposed to be used for gender neutral prefixes? We changed the structure in #2938. 
I remember us discussing that generic should be used to represent gender neutral assosiations and can be omitted if they do not exist in a specific language language. They can exist as a standalone entry, tho.  
Thats what the [types](https://github.com/faker-js/faker/blob/a210619c3cd6d802f8a80e3db68c96b444d15a64/src/definitions/person.ts#L3-L13) reflect as well.

I'd say that this is not the case here. Am I wrong?

ST-DDT: `generic` is intended for cases that don't represent male or female specifically.
In this case it looks like it is just the union of male and female, so it can now be omitted, because it will fallback to the gendered version automatically if it cannot find generic ones.

https://github.com/faker-js/faker/blob/a210619c3cd6d802f8a80e3db68c96b444d15a64/src/definitions/person.ts#L5-L7
https://github.com/faker-js/faker/blob/a210619c3cd6d802f8a80e3db68c96b444d15a64/src/modules/person/index.ts#L21-L47

So you can remove the `generic` entry from these locale data.

ST-DDT: Since the error exists prior to this PR and this PR makes it easier to address in the future. I'm going to resolve this.
We can address this as part of #3058 later.

- #3058 

Is that okay for you @xDivisionByZerox ?

PR #3027, opened 2024-07-24
docs: polish migration guide

### LINKED ISSUE #3007: v9 migration guide polish
## Description

I've done some proof reading on the "finished" [migration guide for v9](https://v9.fakerjs.dev/guide/upgrading.html) and noticed some things that should propably be addressed before the final release.

- [Upgrade to TypeScript v5](https://v9.fakerjs.dev/guide/upgrading.html#upgrade-to-typescript-v5) and [Usage of TypeScript 5 Features](https://v9.fakerjs.dev/guide/upgrading.html#usage-of-typescript-5-features) directly contradict each other:
   "While generally not breaking for users, this might have small side effects on our types." vs "[...] so if you are using Faker with TypeScript, you must use TS5."
- [Fix Tree Shaking](https://v9.fakerjs.dev/guide/upgrading.html#fix-tree-shaking) should have a visual representation (build screenshots, raw data tables) of th impact of the change.
- [Use High Precision RNG by Default](https://v9.fakerjs.dev/guide/upgrading.html#use-high-precision-rng-by-default)
  - The variable names in the code example in make it harder to understand than necessary. Suggested replacements:
     | current names | suggested names |
     | ---------------- | ------------------- |
     | f32                   | oldFaker               |
     | r32                   | oldValue               |
     | f53                   | newFaker              |
     | r32                   | newValue              | 

     Alternativly, instead of having the comparison source code in the same code block, we could place two code blocks next to each other in a table (code highlighting doesn't seem to work on GH, I don't know about vitepress).
  - The adoption list includes an example for vitest specifically. Eigther include on for the most common test libraries or don't include on at all.
  - The section in general seems pretty long compared to the other sections in the guide. Maybe we find a way to get rid of some content and place it in a "In Details"-Blog-Post (or similar).
- The section [Using tsup for the Build Process](https://v9.fakerjs.dev/guide/upgrading.html#using-tsup-for-the-build-process) feels out of place. I don't get why the end user might be interested in this. If there is a reason, the section should be rephrased to make the intend more clear.
- [Removals of Deprecated Code](https://v9.fakerjs.dev/guide/upgrading.html#removals-of-deprecated-code) 
  - The subsections contain the table headers "old" and "replacement". I think "removed" and "replacement" would clearify that the first column is not being "the old way starting this version" but has actually been removed already.
  - The "Note" block in [Company Name Affix files reorganized](https://v9.fakerjs.dev/guide/upgrading.html#company-name-affix-files-reorganized) should be properly formatted
  - The syntax for function references in the tables is inconsistent. Sometimes parentheses are includes, somethimes they aren't (not talking about signature changes, example: [Finance Module](https://v9.fakerjs.dev/guide/upgrading.html#finance-module))
  - The [Helpers Module](https://v9.fakerjs.dev/guide/upgrading.html#helpers-module) section is very long and detailed. It should propably contain a "Read More" link instead of going into several implementation details.
    - The section [faker.helpers.arrayElement and faker.helpers.arrayElements](https://v9.fakerjs.dev/guide/upgrading.html#faker-helpers-arrayelement-and-faker-helpers-arrayelements) feels out of place as it is not a removal but a breaking change. It should be moved to a section in [Breaking Changes to Specific Methods](https://v9.fakerjs.dev/guide/upgrading.html#breaking-changes-to-specific-methods). It should also emphasise that invalid input values are not possible due to typescript.
- The naming schema in [Renamed Locale Definitions](https://v9.fakerjs.dev/guide/upgrading.html#renamed-locale-definitions) should make  clear that `category_name` and `entry_name` are singular. This might be something that should be included in another guide as well.
- [Breaking Changes to Specific Methods](https://v9.fakerjs.dev/guide/upgrading.html#breaking-changes-to-specific-methods)
  - The last part in [Birthdate New Default Mode](https://v9.fakerjs.dev/guide/upgrading.html#birthdate-new-default-mode) read weirdly. 
    > We also improved the error messages in case of invalid min/max age/year ranges.
  - "create a new issue" in [Separate Timezone Methods](https://v9.fakerjs.dev/guide/upgrading.html#separate-timezone-methods) should have a link to the GitHub
  - [Images Have Random Options by Default](https://v9.fakerjs.dev/guide/upgrading.html#images-have-random-options-by-default) is very repetitive. Wewrite to a more generic description with a table pointing out the options to restore the old behaviour.
  - [Phone Number style Replaces Explicit format](https://v9.fakerjs.dev/guide/upgrading.html#phone-number-style-replaces-explicit-format) feels duplicated to [Phone Module](https://v9.fakerjs.dev/guide/upgrading.html#phone-module).

xDivisionByZerox:
Fixes #3007.

- #3007
- [Web-Preview](https://deploy-preview-3027.fakerjs.dev/guide/upgrading.html)


docs/guide/upgrading.md
```diff
@@ -82,15 +82,23 @@ import {
...
-const f32 = new SimpleFaker({ randomizer: generateMersenne32Randomizer() });
-f32.seed(123);
-const r32 = f32.helpers.multiple(() => f32.number.int(10), { count: 10 });
+const oldFaker = new SimpleFaker({
+  randomizer: generateMersenne32Randomizer(),
+});
+oldFaker.seed(123);
+const oldRandomizer = oldFaker.helpers.multiple(() => oldFaker.number.int(10), {
+  count: 10,
+});
```
ST-DDT: It looks like I remembered this wrong. These are indeed values and not ranomizers.

docs/guide/upgrading.md
```diff
@@ -713,18 +719,3 @@ This affects:
 - The `format` property of `faker.color.cmyk()`, `faker.color.hsl()`, `faker.color.hwb()`, `faker.color.lab()`, `faker.color.lch()` must be one of `'binary' | 'css' | 'decimal'` if provided
 - The `variant` property of `faker.location.countryCode()` must be one of `alpha-2`, `alpha-3`, `numeric` if provided
 - The `casing` property of `faker.string.alpha()` and `faker.string.alphanumeric()` must be one of `'upper' | 'lower' | 'mixed'` if provided
-
-### Phone Number `style` Replaces Explicit `format`
```
matthewmayer, line 717: I feel we could keep a simple version of this with a link to the full docs.

xDivisionByZerox: I was thinking about a "Whats new in v9.0" blog post for our website. Are we talking about the same thing?

docs/guide/upgrading.md
```diff
@@ -129,10 +139,6 @@ const faker = new Faker({
 });
 ```
```
ST-DDT, line 137:
```suggestion

### Restructured dist folder

The `dist` folder now contains minified and chunked files for CJS, because we switched to [tsup](https://tsup.egoist.dev) for the bundling process.
So it is no longer possible to use `@faker-js/faker/dist/cjs/...`.
However, as we officially support only `exports` defined via `package.json`, this should not affect your code.

```

ST-DDT (approved):

Shinigami92 (approved):

ejcheng (approved):

docs/guide/upgrading.md
matthewmayer: I feel any time we remove anything without an obvious replacement (like unique or phone format) we should document the replacement options in a lot of detail.

Shinigami92: **Team Decision**

We want to announce this in the upcoming blog post, because this is nothing related to migration

Issue #21, opened 2022-01-08
Discord Community Server


Shinigami92:
Create and link a Discord community server.

(If you need a moderator for the start and initial setup, I can help)

clarkerican:
What is everyone's take on Slack vs Discord?  I'm happy to help set either one up!

damienwebdev:
We have the Daffodil Discord server https://discord.gg/EcF7XFpr as a temporary server if we need to get stuff done quickly. I don't intend for this to be the forever home of FakerJs though,

Shinigami92:
What's Daffodil? I asked for creating a Faker.js community Discord Server like e.g. we have for Vite or Vue or Quasar

Example:
Vite: https://chat.vitejs.dev <a href="https://chat.vitejs.dev"><img src="https://img.shields.io/badge/chat-discord-blue?style=flat&logo=discord" alt="discord chat"></a>

damienwebdev:
Daffodil is a project that depends on Faker. I was just using it as a temporary server for quick correspondence until we set up a new one.

Use this instead: https://discord.gg/U6JxmrEm for a temporary one.

PR #3038, opened 2024-08-05
infra: fix preflight comment for external contributors

ST-DDT:
This PR fixes the preflight comments for external contributors.

PRs that are created from third party repositories are run with read only permissions, thus the CI run is unable to post the comment to the PR.

There are two solutions to that issue:

1. Use the pull request target event
2. Run a secondary CI step with increased permissions

The pull_request_target event based CI pipeline is considered dangerous if it uses the checked out code and since the pipeline actually uses the PRs code in order to determine whether it is complete or not, this is not an option.
So this PR uses the second option and runs an independent CI step using the code from next to run the script to update the comment.

## Example

- PR: https://github.com/ST-DDT/faker/pull/21
- Check-Action: https://github.com/ST-DDT/faker/actions/runs/10255812338/job/28373547559
- Comment-Action: https://github.com/ST-DDT/faker/actions/runs/10255829985/job/28373604237
- PR-Comment: https://github.com/ST-DDT/faker/pull/21#issuecomment-2269893046
  grafik

I hope that I tested this sufficiently and this isn't just a flawed test setup.
It would be good, if you could test this by creating a PR with preflight diffs against `ST-DDT/faker:next` (optimally by using your own forks of faker instead of from the main repo)

- [ST-DDT/faker@`0f10028` (#21)](https://github.com/ST-DDT/faker/pull/21/commits/0f10028cda939acaa4ef37396d9722994249cf88)


Shinigami92:
as with your other PR, please first adjust the PR title to something matching with the history

ST-DDT:
> as with your other PR, please first adjust the PR title to something matching with the history

Done

Shinigami92 (dismissed):

ejcheng (dismissed):

.github/workflows/commentCodeGeneration.ts
```diff
@@ -7,42 +7,60 @@ import type { context as ctx, GitHub } from '@actions/github/lib/utils';
...
+    state: 'open',
+    head: head_branch,
+  });
+
+  if (prs.length === 0) {
+    console.log(`No PRs found for branch ${head_branch}`);
+    return;
+  }
+
+  const pr_number = prs[0].number;
```
xDivisionByZerox, line 31: Instead of doing an additional http request, could you no try to access the [`pull_request` field on the `context`](https://github.com/actions/toolkit/blob/f003268b3250d192cf66f306694b34a278011d9b/packages/github/src/interfaces.ts#L24)? Alternativly the [`issue` getter](https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts#L57) (as previously used) should work as well, sice this will always return a numeric value.

.github/workflows/preflight-comment.yml
```diff
@@ -0,0 +1,45 @@
...
+          node-version: 22
+          cache: 'pnpm'
+
+      - name: Install deps
+        run: pnpm install
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Transpile ts
+        run: pnpm tsup-node .github/workflows/commentCodeGeneration.ts --format cjs --clean false --out-dir .github/workflows
```
xDivisionByZerox, line 38: To optimize our process and conserve energy, could we consider writing the file directly in JavaScript with JSDoc comments for type safety, instead of transpiling it each time?

.github/workflows/preflight-comment.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set node version to 22
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
+        with:
+          node-version: 22
+          cache: 'pnpm'
+
+      - name: Install deps
+        run: pnpm install
+        env:
+          CYPRESS_INSTALL_BINARY: 0
```
xDivisionByZerox, line 35: Do we need to install dependencies here? The script is not using any AFAICT.

xDivisionByZerox (commented):
The current PR desciption doesn't provide information **why** the setup did not work previously. I can follow the steps done in the PR, but can't confirm for myself **why** this fixes the problem. Could you be so kind to give a short explanation in the PR description, please.

.github/workflows/preflight-comment.yml
Shinigami92: I won't really like to do that 🫤 
However we could try the new experimental node feature `node --experimental-strip-types .github/workflows/commentCodeGeneration.ts`, if we have node v22.6.0 available here

.github/workflows/commentCodeGeneration.ts
ST-DDT: The pull request field in the event was always empty in my tests. 😕

ST-DDT:
> The current PR desciption doesn't provide information **why** the setup did not work previously. I can follow the steps done in the PR, but can't confirm for myself **why** this fixes the problem. Could you be so kind to give a short explanation in the PR description, please.

Sure, will add it when I'm back.
TLDR: External PRs dont have any permissions (to comment on the PR).

.github/workflows/preflight-comment.yml
ST-DDT: It uses tsup(-node).

.github/workflows/preflight-comment.yml
ST-DDT: There might be ways to import the code via tsx. For now, I only moved it from one workflow to another.
I'll have another look soon.

xDivisionByZerox: > I won't really like to do that

Why? What do we gain from using typescript in this specific szenario?

.github/workflows/preflight-comment.yml
xDivisionByZerox: Yeah, this comment only makes sense in combination with [my other comment](https://github.com/faker-js/faker/pull/3038#discussion_r1711472685).

.github/workflows/commentCodeGeneration.ts
xDivisionByZerox: The getter for "issue" as well?

.github/workflows/preflight-comment.yml
Shinigami92: > > I won't really like to do that
> 
> Why? What do we gain from using typescript in this specific szenario?

Cause I (assume) wont like to `allowJs: true` in tsconfig and live with a pure ts only codebase

xDivisionByZerox:
FYI: I tested this implementation against @ST-DDT's fork and [it worked as expected](https://github.com/ST-DDT/faker/pull/22#issuecomment-2278124984).

.github/workflows/preflight-comment.yml
xDivisionByZerox: > Cause I (assume) wont like to allowJs: true in tsconfig [...]

That would not be requireed since only `/src` is tracked there 🤔 :

https://github.com/faker-js/faker/blob/354ff4af8b7e3d34125e3052208b5800f6a5ecd9/tsconfig.build.json#L15

We don't have any TS-Config that tracks files in `/.github`. Correct me if I'm wrong.

> [...] live with a pure ts only codebase

I see the benifit for our codebase, but fail to see the benifit over JSDoc enhanced JS in simple scripts. Thats why I asked for reasoning for this specific scenario. Whats the benifit of having a "pure ts only codebase"? This feels like a ADHD thing regarding the GH statistics (???) :

image

Shinigami92: > > Cause I (assume) wont like to allowJs: true in tsconfig [...]
> 
> That would not be requireed since only `/src` is tracked there 🤔 :
> 
> https://github.com/faker-js/faker/blob/354ff4af8b7e3d34125e3052208b5800f6a5ecd9/tsconfig.build.json#L15
> 
> We don't have any TS-Config that tracks files in `/.github`. Correct me if I'm wrong.

Here is the "correction":

https://github.com/faker-js/faker/blob/354ff4af8b7e3d34125e3052208b5800f6a5ecd9/package.json#L80

- https://github.com/faker-js/faker/blob/next/tsconfig.json

---

I pray that some day `.prettierrc.js` and `eslint.config.js` can be a `.ts` file, as every single other tool allows nowadays to also have a `.ts` config file

Edit: for prettier we can follow https://github.com/prettier/prettier/issues/16119

- https://github.com/prettier/prettier/issues/16119

And for eslint, I found a whole RFC 👀 
- https://github.com/eslint/rfcs/tree/main/designs/2024-support-ts-config-files

xDivisionByZerox: > Here is the "correction": [...]

Thank you for that. I considered the default `include` to be empty (`[]`), but with your comment I verified my knowledge and noticed that this is only the case if you provide `files` as well. 

> [!Note]
> Thats one of these occasions where I'd prefer verbose configurations over implicitly using defaults.

Would be easy enought to exclude, TBF.

> I pray that some day .prettierrc.js and eslint.config.js can be a .ts file, as every single other tool allows nowadays to also have a .ts config file

Again, **why**? What benifit do you gain here? We are arguing over an additional compilation step each time the action is run, for a file that is likely to not change for the next year. I'm don't deny the possibility to use TS here, but at this point I really what to know **why**. "Because it's TS" is not a valid technical argument here IMO.

ST-DDT: Maybe it is easier to just inline it.

Shinigami92: you have the stronger arguments right now, so feel free

in the meanwhile I can check if `actions/github-script` can be patched to allow typescript out of the box.
right now it looks like we even use legacy cjs here... maybe we can at least use todays standard esm

Shinigami92: Oh look, there are already issues for
- esm: https://github.com/actions/github-script/issues/457
- ts: https://github.com/actions/github-script/issues/294

Shinigami92: One last point from my side: I'm a bit unsure how this is related to this PR
Maybe extract into own PR so we can refer more easily in the future

xDivisionByZerox: > in the meanwhile I can check if actions/github-script can be patched to allow typescript out of the box.

Excluding the compilation tep by having GH actions accept TS file/code nativly would (mostly) negate all argument I made so far which would bring me on board with that language choice for this task.

> [...] I'm a bit unsure how this is related to this PR

I believe this is relevant because these are completely new files (both TypeScript and GitHub Action). This made me think that downloading all (dev-)dependencies and compiling the script file, which is very unlikely to change, adds overhead and complexity, which is unnecessary in my opinion.

I'm fine talking about the topic of infrastructure design in a separate issue/PR.

Shinigami92: > I believe this is relevant because these are completely new files (both TypeScript and GitHub Action). [...]

Sorry but wrong again 🤔 
`.github/workflows/commentCodeGeneration.ts` is and was a `.ts` file for around 2 years now
- initial commit: https://github.com/faker-js/faker/blob/f934792db4c3c1b44550b79cee893aa545164c59/.github/workflows/commentCodeGeneration.ts

xDivisionByZerox: You are correct. The TS file is not new, only the YAML (`.github/workflows/preflight-comment.yml`) that defines the steps is. So my addition "both TypeScript and GitHub Action" is incorrect, sorry about that.

.github/workflows/commentCodeGeneration.ts
ST-DDT: Yes:

> Issue `{ owner: 'ST-DDT', repo: 'faker', number: undefined }`

- https://github.com/ST-DDT/faker/pull/23
- https://github.com/ST-DDT/faker/commit/2ccb840ac170313f11921857899dc106c554d50f
- https://github.com/ST-DDT/faker/actions/runs/10376203914/job/28727594034

ST-DDT:
Ready for review

Shinigami92 (approved):

.github/workflows/commentCodeGeneration.ts
xDivisionByZerox: Interesting, I'll look into this seperatly.

[I found an interesting discussion in the checkout action repo, tho.](https://github.com/actions/checkout/issues/58#issuecomment-663103947)

xDivisionByZerox (approved):

ejcheng (approved):

PR #3039, opened 2024-08-05
infra: improve CI preflight warning message

### LINKED ISSUE #1840: github-actions bot also triggers when build step failed
### Pre-Checks

### Describe the bug

This is an `infra` bug in our CI

The github-actions bot also triggers when build step failed
Then it suggest to run `pnpm run generate:locales`, `pnpm run generate:api-docs`, and `pnpm run test -u`
But these are totally wrong suggestions as the build step already failed

I see two solutions:
1. Do not print the github-actions bot when the build or anything before that already failed (my preferred solution)
2. Listing all comments that needs to be done to fix the CI (this could get very spammy and unnecessary verbose)

### Minimal reproduction code

_No response_

### Additional Context

image

### Environment Info

```shell
-
```

- [ ] CJS
- [ ] ESM

### Used Package Manager

pnpm

ST-DDT:
Fixes #1840

- #1840

Suggestions for other/better messages are welcome.

## Preview

grafik


.github/workflows/commentCodeGeneration.ts
```diff
@@ -21,7 +21,14 @@ export async function script(
     issue_number: context.issue.number,
   });
 
-  const body = `Uncommitted changes were detected after runnning <code>generate:*</code> commands.\nPlease run <code>pnpm run preflight</code> to generate/update the related files, and commit them.`;
+  const body = `GitHub Actions has detected that this PR is incomplete.
```
Shinigami92: **question:** "incomplete" to what?
Maybe we should add a little bit more detail here, like "incomplete with the contribution assumptions" or anything fancier like that

Shinigami92:
The capital `CI` in PR title looks weird to me, maybe align on something from the history: https://github.com/faker-js/faker/commits/next/.github/workflows/commentCodeGeneration.ts

.github/workflows/commentCodeGeneration.ts
ST-DDT: I have trouble understanding your question and the suggested alternative.

Incomplete => something is missing.
The next sections explains what the user is supposed to do.

Shinigami92: And I have trouble understanding the current sentence as it feels incomplete 😅 

How maybe about something like this?

```suggestion
  const body = `GitHub Actions has detected that there are missing steps that needs to be executed before an approval can be given.
```

matthewmayer: how about "Github Actions found some issues when trying to build the code in this PR."

Shinigami92: > how about "Github Actions found some issues when trying to build the code in this PR."

I don't like mixing the word issue with such things, because GitHub uses issues already as a term for "ticket"
So often I replace "issue" with "problem", so it can't be confused 
However I find my last proposal still a tiny bit better, as it also provides a clear context/reason/goal of **why** this needs to be tackled.

xDivisionByZerox: I like the reworded suggestion from @matthewmayer:

> "Github Actions found some problems when trying to build the code in this PR."

@Shinigami92 Your comment suggests that an approval will be given once the CI completes. However, that's not always the case. For instance, if a PR is opened without a connected issue or if we decide against an implementation after multiple review sessions, an approval might not be granted. This is why I lean towards Matthew's suggestion.

Shinigami92: That's a REALLY good point @xDivisionByZerox 👍

Shinigami92: So how about this?

```suggestion
  const body = `GitHub Actions raised some problems by trying to process the code in this PR.
```

ST-DDT: - Issue => problem - ok
- Raise problem - I would rather go for found problems
- To build/processing code - This may be considered more than building/pure code (test snapshots).

From the suggestions so far I like @matthewmayer's modified version the most.

How about
```suggestion
  const body = `GitHub Actions has found some problems running the preflight checks.
```
Alternatively,
```suggestion
  const body = `GitHub Actions has found some problems processing the PR.
```

Shinigami92: > How about [...]

Both suggestions are good 👍

xDivisionByZerox: > How about [...]

I prefer the first suggestion, but without the question mark (`?`) at the end.

Shinigami92 (approved):

matthewmayer (approved):

xDivisionByZerox (approved):

PR #3042, opened 2024-08-08
ci: publish continuous releases

prisis:
This would enable preview releases (like netlify/vercel previews) but for faker as package

## Examples
Something like this would be possible:

npm i https://pkg.pr.new/@faker-js/faker@173
And this would install the package from pull request number 173.

npm i https://pkg.pr.new/@faker-js/faker@902743b
And this would bring the commit version of the package.

## Motivation
This way, it'd be way easier to try out new pull requests, features and bug fixes. As it allows for a smoother feedback loop on the development cycle.

---

Now we only need to install the [app](https://github.com/apps/pkg-pr-new) so we can have the pr being released!


.github/workflows/pr.yml
```diff
@@ -65,3 +65,42 @@ jobs:
...
+
+      - name: Install deps
+        run: pnpm install
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Generate code
+        id: generate
+        run: |
+          pnpm run generate:locales
```
prisis, line 99: Do i need this?

xDivisionByZerox:
[GitHub let's you checkout PRs](https://docs.github.com/de/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally?tool=cli) as well. If you don't want to use the GH cli, you can also add the PR requester's remote and checkout their branch. Lastly, you could apply the raw patch to your local version.
Why do we need this? I really dislike the thought of fludding some registry (be it npm or anything else) for the sake of "convenience" (that already has a working process). Keep in mind that each release is 10 mb of storage.

prisis:
This would allow you to install only the bundle package, no need for git or build process, just npm i, i a different project to check if something was fixed, otherwise you need to play around with symlink, pnpm link etc

Plus you can run all playground/examples on every MR

Shinigami92:
**Team Decision**

Right now we do not have any internal use case and also don't know any consumer related feedback. Do you (@prisis) have any need or know someone else that has the need? Otherwise we would be in favor of closing this until there is a need.

prisis:
Just seeing it as a simple solution to test all branches without needing to checkout the branch and test for example a fix for the times, I just install it and have a fast response if it works. 

In the end it's a proposal

ST-DDT:
Currently, I either can review a PR by just looking at it or I would like to play with the actual code a bit to see whether I can improve it. When I do that I can also use our playground to test the various aspects of the change, stackblitz doesn't work well for me to test faker features (I mainly see it as a demo).
AFAICT the others think the same regarding this (not sure about stackblitz).

So we would like to close this for now, until there is demand for this from either the community or us.
Thanks for creating this proposal regardless ❤️

PR #3051, opened 2024-08-12
infra: update renovate's eslint group

ST-DDT:

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -36,9 +36,10 @@
     {
       "groupName": "eslint",
       "matchPackagePrefixes": [
-        "@typescript-eslint/",
         "@eslint-types/",
-        "eslint"
+        "@eslint/",
+        "eslint",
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -36,9 +36,10 @@
     {
       "groupName": "eslint",
       "matchPackagePrefixes": [
-        "@typescript-eslint/",
         "@eslint-types/",
-        "eslint"
+        "@eslint/",
+        "eslint",
```
xDivisionByZerox, line 41: <s>What about the other \@eslint* packages, which have some other suffixes? These should be in here as well I think.

https://github.com/faker-js/faker/blob/d2fe40610f6b007c147b139a9c8feea037280549/package.json#L95-L101  
</s>

NVM, I cant read

xDivisionByZerox (approved):

Shinigami92 (approved):

PR #3057, opened 2024-08-15
feat(location): en_CA postcodes by state

matthewmayer:
Implements postcode_by_state for en_CA


Shinigami92:
FFR: This issue was originally reported by @zombiebrainz on Discord

src/locales/en_CA/location/postcode_by_state.ts
```diff
@@ -0,0 +1,21 @@
+// https://en.wikipedia.org/wiki/Postal_codes_in_Canada
+// The basic format is A1A 1A1
+// First letter is province/territory-specific
+// Other letters do not allow D, F, I, O, Q, U
+const suffix = '[0-9][A-HJ-NPRSTVW-Z] [0-9][A-HJ-NPRSTVW-Z][0-9]';
```
ST-DDT: The letters D and F are not allowed.

```suggestion
const suffix = '[0-9][ABCEGHJ-NPRSTVW-Z] [0-9][ABCEGHJ-NPRSTVW-Z][0-9]';
```

ST-DDT (changes requested):

src/locales/en_CA/location/postcode_by_state.ts
```diff
@@ -0,0 +1,21 @@
...
+  BC: `{{helpers.fromRegExp(V${suffix})}}`, // British Columbia
+  MB: `{{helpers.fromRegExp(R${suffix})}}`, // Manitoba
+  NB: `{{helpers.fromRegExp(E${suffix})}}`, // New Brunswick
+  NL: `{{helpers.fromRegExp(A${suffix})}}`, // Newfoundland and Labrador
+  NS: `{{helpers.fromRegExp(B${suffix})}}`, // Nova Scotia
+  NT: `{{helpers.fromRegExp(X${suffix})}}`, // Northwest Territories
+  NU: `{{helpers.fromRegExp(X${suffix})}}`, // Nunavut
+  ON: `{{helpers.fromRegExp([KLMNP]${suffix})}}`, // Ontario
+  PE: `{{helpers.fromRegExp(C${suffix})}}`, // Prince Edward Island
+  QC: `{{helpers.fromRegExp([HJG]${suffix})}}`, // Quebec
```
ST-DDT: minor: IMO we should keep them alphabetically.

```suggestion
  QC: `{{helpers.fromRegExp([GHJ]${suffix})}}`, // Quebec
```

ST-DDT (dismissed):

ST-DDT (approved):

src/locales/en_CA/location/postcode_by_state.ts
```diff
@@ -0,0 +1,21 @@
+// https://en.wikipedia.org/wiki/Postal_codes_in_Canada
+// The basic format is A1A 1A1
+// First letter is province/territory-specific
+// Other letters do not allow D, F, I, O, Q, U
+const suffix = '[0-9][ABCEGHJ-NPRSTVW-Z] [0-9][ABCEGHJ-NPRSTVW-Z][0-9]';
+
+export default {
+  AB: `{{helpers.fromRegExp(T${suffix})}}`, // Alberta
```
Shinigami92, line 8: **thought (if-minor):** I'm a bit confused that there are no added tests or at least generated snapshot results 🤔 But maybe this is really tested somehow behind the scene

However, should it be `helpers.fromRegExp('T${suffix}')` (with quotes) instead?! Or do we have some magic going on 🤔

ST-DDT: `faker.helpers.fake` "magic".
`'` does not work because it would fail json parsing. `"` would work, but it doesnt change anything here.

ST-DDT: Offtopic: For v10 I plan to replace most of our internal fake patterns with something like this `(fakerCore) -> fromRegExp(fakerCore, 'T${suffix}')`.
So we dont have to parse our own code anymore.

Ref: 
- #2667

matthewmayer: > **thought (if-minor):** I'm a bit confused that there are no added tests or at least generated snapshot results 🤔 But maybe this is really tested somehow behind the scene
> 
> 
> 
> However, should it be `helpers.fromRegExp('T${suffix}')` (with quotes) instead?! Or do we have some magic going on 🤔 

The snapshot tests are currently nearly all for default en locale only. We could add some for non English locales but might be hard to know where to stop(presumably not every method in every locale as it would be very slow)

ST-DDT: > But maybe this is really tested somehow behind the scene

It is tested here: https://github.com/faker-js/faker/blob/next/test/all-functional.spec.ts

Shinigami92 (approved):

PR #3069, opened 2024-08-21
infra: switch to @stylistic/eslint-plugin

### LINKED ISSUE #3068: Switching to stylistic eslint plugin
As `typescript-eslint` v8 removes some rules we are using in favor of https://eslint.style/, we want to implement e.g. `@stylistic/eslint-plugin-ts` so we can still benefit from rules like [`padding-line-between-statements`](https://eslint.style/rules/default/padding-line-between-statements)

This needs to be done, before we can merge #3064, so we don't loose the checks for these rules.

This issue was created out of a thread over in #3064:

    **todo (blocking):** before simply removing this lint rule, we need to add https://eslint.style/rules/ts/padding-line-between-statements

_Originally posted by @Shinigami92 in https://github.com/faker-js/faker/pull/3064#discussion_r1722393401_

In typescript-eslint v8 has been removed as it has been moved to the eslint-stylistic package.
This issues serves as a reminder to add the rule back and to unblock the dependency update: https://github.com/faker-js/faker/pull/3064

- https://github.com/faker-js/faker/pull/3064

Shinigami92:
closes #3068 

unblocks #3064

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -132,7 +132,6 @@ export default tseslint.config(
       '@stylistic': stylistic,
     },
     rules: {
-      '@stylistic/semi': 'error',
       '@stylistic/padding-line-between-statements': [
         'error',
         { blankLine: 'always', prev: 'block-like', next: '*' },
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -138,6 +138,7 @@ export default tseslint.config(
       ],
     },
   },
+  //#endregion
 
   //#region deprecation
   {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -2,6 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unsafe-argument */
 import { fixupPluginRules, includeIgnoreFile } from '@eslint/compat';
 import eslint from '@eslint/js';
+import stylistic from '@stylistic/eslint-plugin';
 import eslintPluginDeprecation from 'eslint-plugin-deprecation';
 import eslintPluginJsdoc from 'eslint-plugin-jsdoc';
 import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';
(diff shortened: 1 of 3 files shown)


eslint.config.js
```diff
@@ -129,6 +126,20 @@ export default tseslint.config(
   },
   //#endregion
 
+  //#region stylistic
+  {
+    plugins: {
+      '@stylistic': stylistic,
+    },
+    rules: {
+      '@stylistic/semi': 'error',
```
Shinigami92: **thought:** not yet sure, if we should re-enable `semi`, as this is already covered by prettier.  
We could decide to only enable rules which are not covered by prettier.  
Maybe discuss this in tomorrows meeting?

Shinigami92 (commented):
tested locally by removing some blank lines, and seems to work fine  
I followed the migration guide on their docs and decided to take the "easy" path and use the full combined plugin variant, where we can enable more rules in the future  
It looks like there are no "recommended" rule sets as in other plugins, because their directive is more like "these are stylistic rules which are turned on individually"

eslint.config.js
Shinigami92: we dont need to configure already prettier formatted things

```suggestion
```

eslint.config.js
```diff
@@ -129,6 +126,19 @@ export default tseslint.config(
...
+    plugins: {
+      '@stylistic': stylistic,
+    },
+    rules: {
+      '@stylistic/padding-line-between-statements': [
+        'error',
+        { blankLine: 'always', prev: 'block-like', next: '*' },
+      ],
+    },
+  },
```
ST-DDT, line 140: This is missing the endregion comment.

eslint.config.js
```diff
@@ -129,6 +126,19 @@ export default tseslint.config(
...
+
```
Shinigami92, line 142: **suggestion:** 
```suggestion
  //#endregion

```

eslint.config.js
Shinigami92: nice catch 💪

xDivisionByZerox (approved):

ST-DDT (approved):

PR #3070, opened 2024-08-22
feat(number): add romanNumeral method

### LINKED ISSUE #2649: Add support for generating Roman Numerals

It would be useful to be able to generate roman numerals, e.g. for sections of a fake document, or as a fake pattern to be used in another pattern like "Super Bowl XXVII".

### Suggested solution

```js
faker.number.romanNumeral({min:1, max: 100}) //XVII
```
min would default to 1 (you can't write 0 in roman numerals). max would probably be limited to 4999 as roman numerals are not as well-defined above 5000.
Maybe lowercase/uppercase options?

Not localizable. Since roman numerals are always written in latin script.

### Alternative

write custom conversion code eg. 

```js
convertToRomanNumerals(faker.number.int({min:1, max:4999}))
```

where convertToRomanNumerals is a suitable implementation from https://stackoverflow.com/questions/9083037/convert-a-number-into-a-roman-numeral-in-javascript

### Additional context

_No response_

AmaanRS:
Adds #2649

- #2649

---

Adds a method for generating random romanNumerals:

````ts
faker.number.romanNumeral() // "CMXCIII"
faker.number.romanNumeral(5) // "III"
faker.number.romanNumeral({ min: 10 }) // "XCIX"
faker.number.romanNumeral({ max: 20 }) // "XVII"
faker.number.romanNumeral({ min: 5, max: 10 }) // "VII"
````

[Documentation](https://deploy-preview-3070.fakerjs.dev/api/number.html#romannumeral)

### CODE REVIEW

File: src/modules/number/index.ts [hunk 1]
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
 
     return min + offset;
   }
+
+  /**
+   * Returns a Roman Numeral in [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#string_type) format.
```
```suggestion
   * Returns a roman numeral in [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#string_type) format.
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
...
+   * faker.number.romanNumeral(5) // "V"
+   * faker.number.romanNumeral({ min: 10 }) // "XI"
+   * faker.number.romanNumeral({ max: 20 }) // "III"
+   * faker.number.romanNumeral({ min: 5, max: 10 }) // "VII"
+   *
+   */
+
+  romanNumeral(options: number | { min?: number; max?: number } = {}): string {
+    const DEFAULT_MIN = 1;
+    const DEFAULT_MAX = 4999;
```
ejcheng: What do we think about these default bounds?

ejcheng (dismissed):
Please add tests for `romanNumeral`

src/modules/number/index.ts
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @param options.max Upper bound for generated roman numerals. Defaults to `4999`.
+   *
+   * @throws When `min` is greater than `max`.
+   * @throws When `min`,`max` is not a number.
+   * @throws When `min` is less than `1`.
+   * @throws When `max` is greater than `4999`.
+   *
+   * @example
+   * faker.number.romanNumeral() // "IV"
+   * faker.number.romanNumeral(5) // "V"
```
Shinigami92: **suggestion:** ok, so some initial feedback also from my side
we should find another example value here, to make it more clear that `5 not => V`
```suggestion
   * faker.number.romanNumeral(5) // "III"
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @param options.min Lower bound for generated roman numerals. Defaults to `1`.
+   * @param options.max Upper bound for generated roman numerals. Defaults to `4999`.
+   *
+   * @throws When `min` is greater than `max`.
+   * @throws When `min`,`max` is not a number.
+   * @throws When `min` is less than `1`.
+   * @throws When `max` is greater than `4999`.
+   *
+   * @example
+   * faker.number.romanNumeral() // "IV"
```
Shinigami92: **suggestion:** it's possible but nearly unrealistic that an example value between 1 and 4999 will result in IV
you should pick a "more random" example

src/modules/number/index.ts
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
...
+   * faker.number.romanNumeral(5) // "V"
+   * faker.number.romanNumeral({ min: 10 }) // "XI"
```
Shinigami92: **suggestion:** also for this

src/modules/number/index.ts
```diff
@@ -443,4 +443,113 @@ export class NumberModule extends SimpleModuleBase {
...
+      'IV',
+      'V',
+      'VI',
+      'VII',
+      'VIII',
+      'IX',
+    ];
+
+    let roman = '';
+    let i = 3;
```
Shinigami92: **thought:** this looks like a very magic number to me, why is this `3`?

Shinigami92 (dismissed):

src/modules/number/index.ts
matthewmayer: This code comes from https://blog.stevenlevithan.com/archives/javascript-roman-numeral-converter

I think it would be good to extract the actual "int to roman numeral" part to a separate function and include a comment referencing the blog post.

src/modules/number/index.ts
AmaanRS: I have changed the max to 3999, check the new draft

src/modules/number/index.ts
AmaanRS: I tested the code written in https://blog.stevenlevithan.com/archives/javascript-roman-numeral-converter but i found out that it worked only for values in range 1 to 999 so i changed the code, i took the new logic from here https://www.30secondsofcode.org/js/s/to-roman-numeral/

AmaanRS:
In this commit 7abd2c7b262a31d5a699a17f99dcfab9d4650afd i created a new function intToRomanNumeral so where should i place it ?

src/modules/number/index.ts
```diff
@@ -443,4 +443,93 @@ export class NumberModule extends SimpleModuleBase {
...
+    ];
+
+    const intToRomanNumeral = (num: number): string =>
+      lookup.reduce((acc: string, [k, v]: [string, number]) => {
+        acc += k.repeat(Math.floor(num / v));
+        num = num % v;
+        return acc;
+      }, '');
+
+    return intToRomanNumeral(num);
```
Shinigami92: **question:** why wrapping it in a function in the first place?

you could potentially just do:

```suggestion
    return lookup.reduce((acc, [k, v]) => {
        acc += k.repeat(Math.floor(num / v));
        num = num % v;
        return acc;
      }, '');
```

#nottested

src/modules/number/index.ts
```diff
@@ -443,4 +443,93 @@ export class NumberModule extends SimpleModuleBase {
...
+  /**
+   * Returns a roman numeral in [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#string_type) format.
+   * The bounds are inclusive.
+   *
+   * @param options Maximum value or options object.
+   * @param options.min Lower bound for generated roman numerals. Defaults to `1`.
+   * @param options.max Upper bound for generated roman numerals. Defaults to `3999`.
+   *
+   * @throws When `min` is greater than `max`.
+   * @throws When `min`,`max` is not a number.
```
Shinigami92: **suggestion:** 
```suggestion
   * @throws When `min`, `max` is not a number.
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,93 @@ export class NumberModule extends SimpleModuleBase {
...
+    const DEFAULT_MIN = 1;
+    const DEFAULT_MAX = 3999;
+
+    if (typeof options === 'number') {
+      options = {
+        max: options,
+      };
+    }
+
+    let num = DEFAULT_MIN;
```
matthewmayer: this line is not needed as num is always set at line 505 or 507

src/modules/number/index.ts
```diff
@@ -443,4 +443,93 @@ export class NumberModule extends SimpleModuleBase {
...
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be larger then min ${min}.`);
+    }
+
+    if (max === min) {
+      num = min;
+    } else {
+      num = this.int({ min, max });
+    }
```
matthewmayer: there's no need to special-case the `max==min`. just simplify down to

`let num = this.int({ min, max });`

src/modules/number/index.ts
matthewmayer: note we don't allow .reduce() https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md so this should probably just be replaced with a simple loop

matthewmayer:
```js
    let result = '';
    for (const [k, v] of lookup) {
      result += k.repeat(Math.floor(num / v));
      num = num % v;
    }
    return result;
```

AmaanRS: > This code comes from https://blog.stevenlevithan.com/archives/javascript-roman-numeral-converter
> 
> I think it would be good to extract the actual "int to roman numeral" part to a separate function and include a comment referencing the blog post.

Should i create a new function,if yes where should i place it ?

AmaanRS:
How should i write random seeded test for "it('should generate a Roman numeral between 1 and 10')" ? Since i don't what answer will be.

AmaanRS:
And there will be no value range test ?

src/modules/number/index.ts
matthewmayer: You're not using that function any more so I think it's fine to inline it as suggested above

matthewmayer:
There's a few ways you could do it. For 1 to 10 you could check if it's any of the values from I to X by listing all 10 values. 

You could also check it only contains suitable characters. 

Also add tests that check hitting each of the errors and check appropriate error is thrown.

AmaanRS:
Please let me know if there are any changes with the code and test if not i will run preflight and then commit the code

matthewmayer:
Ideally please run preflight after every change. This will help catch any minor syntax errors etc so the reviewers can focus on the logic of your code rather than nitpicking.

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @throws When `min` is less than `1`.
+   * @throws When `max` is greater than `3999`.
+   *
+   * @example
+   * faker.number.romanNumeral() // "CMXCIII"
+   * faker.number.romanNumeral(5) // "III"
+   * faker.number.romanNumeral({ min: 10 }) // "XCIX"
+   * faker.number.romanNumeral({ max: 20 }) // "XVII"
+   * faker.number.romanNumeral({ min: 5, max: 10 }) // "VII"
+   *
```
ST-DDT, line 466:
```suggestion
   *
   * @since 9.1.0
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+   */
+
```
```suggestion
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+    let result = '';
+
+    if (typeof options === 'number') {
+      options = {
+        max: options,
+      };
+    }
+
+    const min = Number(options.min ?? DEFAULT_MIN);
+    const max = Number(options.max ?? DEFAULT_MAX);
```
```suggestion
    const { min = 1, max = 3999 } = options;
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+
+    if (max > DEFAULT_MAX) {
+      throw new FakerError(`Max value ${max} should be 3999 or lesser.`);
+    }
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be larger then min ${min}.`);
+    }
+
+    num = this.int({ min, max });
```
ST-DDT: Please declare variables late/shortly before using it.

```suggestion
    let num = this.int({ min, max });
    let result = '';
```

test/modules/number.spec.ts
```diff
@@ -625,6 +635,80 @@ describe('number', () => {
...
+        expect(() => {
+          faker.number.romanNumeral({ max: '2q' as any });
+        }).toThrow(new FakerError('Max value must be a valid number.'));
+      });
+
+      it('should throw when max is NaN', () => {
+        expect(() => {
+          faker.number.romanNumeral({ max: NaN });
+        }).toThrow(new FakerError('Max value must be a valid number.'));
+      });
```
ST-DDT: We usually don't include tests for these and don't handle it in the implementation either.
The string parts aren't possible with proper usage of TypeScript (which we assume our users or their tooling use) or are able to read the resulting error messages in case of bad values.

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+      throw new FakerError(`Min value ${min} should be 1 or greater.`);
+    }
+
+    if (max > DEFAULT_MAX) {
+      throw new FakerError(`Max value ${max} should be 3999 or lesser.`);
+    }
+
+    if (max < min) {
+      throw new FakerError(`Max ${max} should be larger then min ${min}.`);
+    }
```
ST-DDT: This is already covered by `faker.number.int()`.

src/modules/number/index.ts
```diff
@@ -443,4 +443,88 @@ export class NumberModule extends SimpleModuleBase {
...
+    const min = Number(options.min ?? DEFAULT_MIN);
+    const max = Number(options.max ?? DEFAULT_MAX);
+
+    if (typeof min !== 'number' || isNaN(min)) {
+      throw new FakerError('Min value must be a valid number.');
+    }
+
+    if (typeof max !== 'number' || isNaN(max)) {
+      throw new FakerError('Max value must be a valid number.');
+    }
```
ST-DDT: We don't include these check, since we assume that our users either use TypeScript, a type aware IDE, or are able to read the native error messages, created when a bad value is passed.

AmaanRS:
I ran into an error during preflight ```FakerApiDocsProcessingError: Failed to process parameter options at src/modules/number/index.ts:468:16 : Expected exactly one element for jsdocs, got 0```

matthewmayer:
> Expected exactly one element for jsdocs

its not a very intuitive error message but i think the problem is that you're missing a `@since` JSDoc tag

add

`@since 9.1.0` at the bottom of the JSDoc as this likely wont be in the 9.0.0 release.

ST-DDT:
> I ran into an error during preflight `FakerApiDocsProcessingError: Failed to process parameter options at src/modules/number/index.ts:468:16 : Expected exactly one element for jsdocs, got 0`

The error is not caused by `@since`. It is caused by the `options.min/max` parameter not having jsdocs.
I'll improve the error message.

````txt
FakerApiDocsProcessingError: Failed to process property 'options.min' at src/modules/number/index.ts:467:36 : Expected exactly one element for jsdocs, got 0
````

The options object should be defined like this:

````ts
  romanNumeral(
    options:
      | number
      | {
          /**
           * Lower bound for generated number.
           *
           * @default 1
           */
          min?: number;
          /**
           * Upper bound for generated number.
           *
           * @default 3999
           */
          max?: number;
        } = {}
  ): string {
````

File: src/modules/number/index.ts [hunk 18]
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
 
     return min + offset;
   }
+
+  /**
+   * Returns a roman numeral in [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#string_type) format.
```
matthewmayer: The link to the String type seems unnecessary

src/modules/number/index.ts
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
...
+    }
+
+    const { min = 1, max = 3999 } = options;
+
+    if (min < DEFAULT_MIN) {
+      throw new FakerError(`Min value ${min} should be 1 or greater.`);
+    }
+
+    if (max > DEFAULT_MAX) {
+      throw new FakerError(`Max value ${max} should be 3999 or lesser.`);
```
matthewmayer: lesser->less

matthewmayer (changes requested):

matthewmayer (dismissed):

src/modules/number/index.ts
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
...
+          /**
+           * Upper bound for generated number.
+           *
+           * @default 3999
+           */
+          max?: number;
+        } = {}
+  ): string {
+    const DEFAULT_MIN = 1;
+    const DEFAULT_MAX = 3999;
```
Shinigami92, line 488: **suggestion:** if we already define these default values here, we can reuse them in line 496, 499 and 503

src/modules/number/index.ts
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
...
+    if (typeof options === 'number') {
+      options = {
+        max: options,
+      };
+    }
+
+    const { min = DEFAULT_MIN, max = DEFAULT_MAX } = options;
+
+    if (min < DEFAULT_MIN) {
+      throw new FakerError(`Min value ${min} should be 1 or greater.`);
```
Shinigami92: **suggestion:** 
```suggestion
      throw new FakerError(`Min value ${min} should be ${DEFAULT_MIN} or greater.`);
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
...
+    }
+
+    if (max > DEFAULT_MAX) {
+      throw new FakerError(`Max value ${max} should be 3999 or less.`);
```
Shinigami92: **suggestion:** 
```suggestion
      throw new FakerError(`Max value ${max} should be ${DEFAULT_MAX} or less.`);
```

src/modules/number/index.ts
```diff
@@ -443,4 +443,91 @@ export class NumberModule extends SimpleModuleBase {
...
+      ['V', 5],
+      ['IV', 4],
+      ['I', 1],
+    ];
+
+    let result = '';
+
+    for (const [k, v] of lookup) {
+      result += k.repeat(Math.floor(num / v));
+      num = num % v;
```
ST-DDT: Maybe:

```suggestion
      num %= v;
```

test/modules/number.spec.ts
```diff
@@ -625,6 +635,56 @@ describe('number', () => {
...
+          6: 'VI',
+          7: 'VII',
+          8: 'VIII',
+          9: 'IX',
+          10: 'X',
+        };
+
+        const roman = faker.number.romanNumeral({ min: 1, max: 10 });
+
+        expect(Object.values(values)).toContain(roman);
```
ST-DDT: Are the keys of `values` used for anything?

ST-DDT:
@AmaanRS Could you please address the requested changes?

test/modules/number.spec.ts
AmaanRS: Nothing, expect for returning them as roman numeral string

ST-DDT: Then maybe use an array directly.

AmaanRS: Done!!

test/modules/number.spec.ts
```diff
@@ -625,6 +635,56 @@ describe('number', () => {
...
+          'VII',
+          'VIII',
+          'IX',
+          'X',
+        ];
+
+        const roman = faker.number.romanNumeral({ min: 1, max: 10 });
+
+        expect(values).toContain(roman);
+      });
```
ST-DDT: Sorry, I didn't think of this earlier.
I changed this test to check for the actual values:

```suggestion
      it.each(
        Object.entries({
          I: 1,
          IV: 4,
          IX: 9,
          X: 10,
          XC: 90,
          XCIX: 99,
          XXVII: 27,
          CCLXIII: 263,
          DXXXVI: 536,
          DCCXIX: 719,
          MDCCCLI: 1851,
          MDCCCXCII: 1892,
          MMCLXXXIII: 2183,
          MMCMXLIII: 2943,
          MMMDCCLXVI: 3766,
          MMMDCCLXXIV: 3774,
          MMMCMXCIX: 3999,
        })
      )(
        'should generate a Roman numeral %s for value %d',
        (expected: string, value: number) => {
          const mock = vi.spyOn(faker.number, 'int');
          mock.mockReturnValue(value);

          const actual = faker.number.romanNumeral();

          mock.mockRestore();

          expect(actual).toBe(expected);
        }
      );
```

A few manually selected ones and a few random ones.

ST-DDT (commented):
One last change, then this is good to go.

ST-DDT:
Please also change the description of the PR to something explaining the feature in more detail. A few sentences are sufficient.

test/modules/number.spec.ts
AmaanRS: Done!!

ST-DDT (dismissed):

src/modules/number/index.ts
```diff
@@ -443,4 +443,95 @@ export class NumberModule extends SimpleModuleBase {
...
+   * @throws When `max` is greater than `3999`.
+   *
+   * @example
+   * faker.number.romanNumeral() // "CMXCIII"
+   * faker.number.romanNumeral(5) // "III"
+   * faker.number.romanNumeral({ min: 10 }) // "XCIX"
+   * faker.number.romanNumeral({ max: 20 }) // "XVII"
+   * faker.number.romanNumeral({ min: 5, max: 10 }) // "VII"
+   *
+   * @since 9.1.0
```
```suggestion
   * @since 9.2.0
```

ST-DDT (approved):

ST-DDT:
@Shinigami92 and @import-brain Could you please re-review this PR (or dismiss your stale reviews)?

ST-DDT:
@AmaanRS Thanks for your first contribution. ❤️ 
I hope you had a good experience contributing.
If you have any suggestions on how to improve things, please let us know.

This feature will be released in Faker v9.2, which will probably release shortly (a week to a few weeks).

PR #3082, opened 2024-09-01
infra: improve error messages for parameter properties

### LINKED ISSUE #3079: Non-intuitive error if options property doesn't have jsdocs
```FakerApiDocsProcessingError: Failed to process parameter options at src/modules/number/index.ts:468:16 : Expected exactly one element for jsdocs, got 0```

_Originally posted by @AmaanRS in https://github.com/faker-js/faker/issues/3070#issuecomment-2323336805_

### PR DESCRIPTION
**Pull request by ST-DDT**
Partially fixes https://github.com/faker-js/faker/issues/3079

- https://github.com/faker-js/faker/issues/3079

Adds distinct error references for option parameter's properties and outputs them to the github action errors.

### CODE REVIEW

**Comment by ST-DDT** [comment]
I also added an error notification to the script:

grafik

**Comment by ST-DDT** [comment]
Any wishes regarding the two error message parts (mainText + causeText)?

File: scripts/consts.ts [hunk 1]
```diff
@@ -0,0 +1,3 @@
+import { env } from 'node:process';
+
+export const CI_PREFLIGHT = env.CI_PREFLIGHT === 'true';
```
Shinigami92: **question:** why did you sourced this out into its own file?

_Context:_
I personally don't like semi-global const files, where everything just gets stored.
Instead we could just use `env.CI_PREFLIGHT` directly (potentially no need to check for `'true'`)

Shinigami92 (commented):
I would potentially still be a bit confused when I would encounter such a `causeText`, but I see that you just pass forward the `error.message`. So for now this is fine for me.

scripts/consts.ts
ST-DDT: > **question:** why did you sourced this out into its own file?

Because I don't like random variables scattered across the codebase. So since we now use it more then once, I just put it into a dedicated variable with proper types (And theoretically could but documentation on to explain what it is used for).
I could rename it to `env.ts`, but I'm not sure that would make it any better.

I was thinking of changing the definition back to `!!env.CI_PREFLIGHT` though (to match the previous behaviour).

ST-DDT:
> I would potentially still be a bit confused when I would encounter such a `causeText`

The causeText is just defined in a different location.

https://github.com/faker-js/faker/blob/17f570e05ea413af691bb81b61dd8c28130a47e1/scripts/apidocs/utils/value-checks.ts#L4

Would the following be easier to understand?

````ts
`Expected exactly one ${property} element, got ${input.length}`
`Expected exactly one jsdoc element, got 0`
````

(I have to check the other usages still)

matthewmayer:
I think the real confusion is that it is not intuitive that you have to write the JSDocs for the option parameters twice, once at the top of the method and once directly above each option. So this is an improvement, but still not sure that a new contributor confronted with this message would know how to fix it.

scripts/consts.ts
Shinigami92: I would definitely prefer a `env.ts` over `const.ts`, because `const.ts` leads to "lets just put this here", where as `env.ts` is explicitly for environments.
We could then even go a step further and make use of [envalid](https://www.npmjs.com/package/envalid). With that, only checked envs are processed and no other envs slip through. Also it converts automatically e.g. a `'true'` to a real boolean `true`.

ST-DDT:
How about an error like this:

> `FakerApiDocsProcessingError: Failed to process property 'options.min' at src/modules/number/index.ts:48:11 : Expected exactly one jsdocs element, got 0. Please ensure that each method signature has JSDocs, and that all properties of option/object parameters are documented with both @param tags and inline JSDocs.`

scripts/consts.ts
ST-DDT: Renamed, there is no need to add a library for just 1 env variable.

Shinigami92:
> How about an error like this:
> 
> > `FakerApiDocsProcessingError: Failed to process property 'options.min' at src/modules/number/index.ts:48:11 : Expected exactly one jsdocs element, got 0. Please ensure that each method signature has JSDocs, and that all properties of option/object parameters are documented with both @param tags and inline JSDocs.`

This is BY FAR EXORBITANT much better 😍

Shinigami92 (approved):

xDivisionByZerox (approved):
Great work @ST-DDT 👏

matthewmayer (approved):

wedouglas:
> I think the real confusion is that it is not intuitive that you have to write the JSDocs for the option parameters twice, once at the top of the method and once directly above each option. So this is an improvement, but still not sure that a new contributor confronted with this message would know how to fix it.

@matthewmayer No response even after the court order?

PR #3090, opened 2024-09-05
docs: use defineProps deconstruction

ST-DDT:
Simplifies the vue components definitions by using deconstructors.


docs/.vitepress/components/api-docs/method.vue
```diff
@@ -4,51 +4,60 @@ import { slugify } from '../../shared/utils/slugify';
...
+  description,
+  since,
+  parameters,
+  returns,
+  throws,
+  signature,
+  examples,
+  seeAlsos,
+  sourcePath,
+} = method;
```
Shinigami92, line 19: **nitpick:** we are loosing reactivity here. However, this might not be a problem, as the component gets mounted with its initial state and never changes its state (as far as I assume)  

more correct would be something like `toRefs(toRef(() => method))` (I did not test that, it only comes from my gut feeling)

Shinigami92 (commented):
The new syntax feels still weird to me and I would also like to explore how current state of IDE support is. Also we sadly did not benefit here in these examples of the real new feature of assigning default values and omit the use of `withDefaults` 🙁

xDivisionByZerox (approved):

docs/.vitepress/components/api-docs/method.vue
ST-DDT: Our data aren't reactive/refs, so there is no difference either way.

Shinigami92 (approved):

PR #3092, opened 2024-09-05
infra(netlify): remove ignore script

ST-DDT:
Removes the netlify ignore check.

- There are some cases where it would be nice if there was a preview of the changes (aka as some dependency changes).
- The ignore script does not reduce the number of notifications and we aren't close to the build limits either

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,15 +5,6 @@
 [build]
   publish = "docs/.vitepress/dist"
   command = "pnpm docs:build:ci"
-  ignore = '''
-    if [ "$CONTEXT" != "deploy-preview" ]; then
-      # Always build when not a PR preview
-      exit 1;
-    else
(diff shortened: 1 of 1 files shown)


Shinigami92:
> There are some cases where it would be nice if there was a preview of the changes (aka as some dependency changes).

So should we just remove `pnpm-lock.yaml` from the check?

ST-DDT:
> > There are some cases where it would be nice if there was a preview of the changes (aka as some dependency changes).
> 
> So should we just remove `pnpm-lock.yaml` from the check?

Then it triggers in 75% of cases, so there is no need for the added complexity for no gain.
(The difference is only the locale PRs)

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3093, opened 2024-09-06
fix: emit cts types

### LINKED ISSUE #3078: warning in package.json during build
There is a warning in the package.json when the code is build using tsup.

- https://github.com/faker-js/faker/actions/runs/10638166909/job/29493558767

````txt
Warning: G] The condition "default" here will never be used as it comes after both "import" and "require" [package.json]

    package.json:48:6:
      48 │       "default": "./dist/index.js"
         ╵       ~~~~~~~~~

  The "import" condition comes earlier and will be used for all "import" statements:

    package.json:46:6:
      46 │       "import": "./dist/index.js",
         ╵       ~~~~~~~~

  The "require" condition comes earlier and will be used for all "require" calls:

    package.json:47:6:
      47 │       "require": "./dist/index.cjs",
         ╵       ~~~~~~~~~

Warning: G] The condition "default" here will never be used as it comes after both "import" and "require" [package.json]

    package.json:54:6:
      54 │       "default": "./dist/locale/*.js"
         ╵       ~~~~~~~~~

  The "import" condition comes earlier and will be used for all "import" statements:

    package.json:52:6:
      52 │       "import": "./dist/locale/*.js",
         ╵       ~~~~~~~~

  The "require" condition comes earlier and will be used for all "require" calls:

    package.json:53:6:
      53 │       "require": "./dist/locale/*.cjs",
         ╵       ~~~~~~~~~
````

Related code:

https://github.com/faker-js/faker/blob/17f570e05ea413af691bb81b61dd8c28130a47e1/package.json#L44-L49
https://github.com/faker-js/faker/blob/17f570e05ea413af691bb81b61dd8c28130a47e1/package.json#L50-L55

### LINKED ISSUE #3087: [9.0.0] Types are not properly emitted
### Pre-Checks

### Describe the bug

@faker-js/faker@9.0.0 does not emit typescript types properly. Consider [this tool](https://arethetypeswrong.github.io/?p=%40faker-js%2Ffaker%409.0.0) as reference.

Consider https://github.com/grafana/pyroscope-nodejs/issues/37 as reference.

### Minimal reproduction code

Simply install the library in NodeJS >= 16 and try to import it in a cjs module.

tsc will complain with the following error:

```
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@faker-js/faker")' call instead.
```

### Additional Context

Having a look at the `package.json`:

```json
{
  "exports": {
    ".": {
      "import": "./dist/index.js",
      "require": "./dist/index.cjs",
      "default": "./dist/index.js"
    },
    "./locale/*": {
      "types": "./dist/types/locale/*.d.ts",
      "import": "./dist/locale/*.js",
      "require": "./dist/locale/*.cjs",
    },
  }
}
```

This is wrong. Types emitted at `./dist/types/` are esm types, breaking cjs requires. Instead of doing this, types should be emitted in both cjs and esm builds so typescript can infer the right types. Consider https://github.com/grafana/pyroscope-nodejs/issues/37 as reference

### Environment Info

```shell
System:
    OS: Linux 6.8 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 24.45 GB / 31.24 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.9.0/bin/yarn
    npm: 10.8.2 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 9.9.0 - ~/.nvm/versions/node/v20.9.0/bin/pnpm
  Browsers:
    Chrome: 127.0.6533.119
```

- [X] CJS
- [ ] ESM

### Used Package Manager

pnpm

### LINKED ISSUE #3104: faker verision 9.0.0  : typescript  Problem when i Importing Faker Package as ESM
### Pre-Checks

### Describe the bug

When trying to import the Faker package as an ESM, my IDE (VSCode) displays the following error:

```
The current file is a CommonJS module, whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider using a dynamic 'import("@faker-js/faker")' call instead.
To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/home/victor/www/nlw-pocket/server/package.json'.
```

Even though I am already using ESM in my project, I tried reverting to the previous version of Faker with:

```bash
npm install @faker-js/faker@8.4.1 -D
```

and it worked well. However, when I switch back to the current version 9.0.0, my IDE continues to show this error.

---

observation:
both versions work  well

OS: windows 11 with wsl2

image

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
OS: windows 11 with wsl2
package menager: npm 

{
 "@faker-js/faker": "^9.0.0",
    "tsx": "^4.19.0",
    "typescript": "^5.6.2"
}
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

Shinigami92:
closes #3087


Shinigami92:
@faker-js/maintainers how should we handle following?
> test/faker.spec.ts(21,40): error TS2307: Cannot find module '..' or its corresponding type declarations.

Shinigami92:
The only difference between the old types, is that the `.d.ts` have `.js` suffixes and the `.d.cts` have `.cjs` suffixes. That's it 🤷
Everything else is just an exact copy, resulting in an additional ~500kb to be downloaded by million of users weekly...

Shinigami92:
It looks like that with this change the moduleResolution Node/Node10 is not usable anymore and the build fails with:

```node
index.ts:2:34 - error TS2307: Cannot find module '@faker-js/faker/locale/de' or its corresponding type declarations.
  There are types at 'D:/shinigami/OpenSource/faker-playground/node_modules/@faker-js/faker/dist/locale/de.d.ts',
  but this result could not be resolved under your current 'moduleResolution' setting.
  Consider updating to 'node16', 'nodenext', or 'bundler'.
```

image

For whatever reason, this only affects `/locale/*` imports

scripts/apidocs/project.ts
```diff
@@ -4,6 +4,6 @@ import { Project } from 'ts-morph';
 export function getProject(options: Partial<ProjectOptions> = {}): Project {
   return new Project({
     ...options,
-    tsConfigFilePath: options.tsConfigFilePath ?? 'tsconfig.build.json',
+    tsConfigFilePath: options.tsConfigFilePath ?? 'tsconfig.json',
```
ST-DDT, line 7: Hi, can you please explain why this value was changed?

ST-DDT:
> @faker-js/maintainers how should we handle following?
> 
> > test/faker.spec.ts(21,40): error TS2307: Cannot find module '..' or its corresponding type declarations.

It's gone, if you build the project first.

scripts/apidocs/project.ts
Shinigami92: Because my original intent was/is to do not have a `tsconfig.build.json` at all anymore, because everything is now managed by `tsup`. And because `tsup` does not have any reference to `tsconfig.build.json` but (I assume) reads the `tsconfig.json`, I already changed this.
But this is also one of the reason why this PR is still in draft.

Shinigami92:
> > @faker-js/maintainers how should we handle following?
> > > test/faker.spec.ts(21,40): error TS2307: Cannot find module '..' or its corresponding type declarations.
> 
> It's gone, if you build the project first.

But previously we only build the types and not the whole project. Do we really want now to build the whole project just to check the types? Or do we want to add a `@ts-expect-error / @ts-ignore` on this line?

ST-DDT:
> But previously we only build the types and not the whole project. Do we really want now to build the whole project just to check the types?

No, we don't have to build it.

> Or do we want to add a `@ts-expect-error / @ts-ignore` on this line?

We can do that, but I would like to understand why it doesn't work anymore.

ST-DDT:
> It looks like that with this change the moduleResolution Node/Node10 is not usable anymore and the build fails with:
> 
> ```js
> index.ts:2:34 - error TS2307: Cannot find module '@faker-js/faker/locale/de' or its corresponding type declarations.
>   There are types at 'D:/shinigami/OpenSource/faker-playground/node_modules/@faker-js/faker/dist/locale/de.d.ts',
>   but this result could not be resolved under your current 'moduleResolution' setting.
>   Consider updating to 'node16', 'nodenext', or 'bundler'.
> ```
> 
> image
> 
> For whatever reason, this only affects `/locale/*` imports

If you keep the type stuff, it works:

````ts
  "types": "index.d.ts",
  "typesVersions": {
    ">=5.0": {
      "*": [
        "dist/*"
      ]
    }
  },
````

**Comment by ST-DDT** [comment]
We have to add tests/playgrounds for all of this, so that it doesn't silently break.

File: package.json [hunk 2]
```diff
@@ -32,26 +32,26 @@
...
   "module": "dist/index.js",
-  "types": "index.d.ts",
-  "typesVersions": {
-    ">=5.0": {
-      "*": [
-        "dist/types/*"
-      ]
-    }
-  },
   "exports": {
```
mshima, line 65: node10 doesn’t understand exports so it will not resolve without typesVersions.
types field is not required.

```suggestion
  "typesVersions": {
    ">=5.0": {
      ".": [
        "./dist/index.d.ts"
      ],
      "locale/*": [
        "./dist/locale/*.d.ts"
      ]
    }
  },
  "exports": {
```

mshima:
I played with this branch against https://github.com/faker-js/faker/pull/3095 and the missing step is to add typesVersions field again.

package.json
```diff
@@ -63,8 +73,7 @@
     "clean": "rimraf coverage .eslintcache dist docs/.vitepress/cache docs/.vitepress/dist node_modules",
     "build:clean": "rimraf dist",
     "build:code": "tsup-node",
-    "build:types": "tsc --project tsconfig.build.json",
```
ST-DDT: If you make the file obsolete, please remove it.

Shinigami92:
Should we merge https://github.com/faker-js/faker/pull/3097 before we make such bigger changes to `package.json`?

- https://github.com/faker-js/faker/pull/3097

xDivisionByZerox:
Running the current PR against the playground I get a [similar error @Shinigami92 got before](https://github.com/faker-js/faker/pull/3093#issuecomment-2333983188). This time for the root import:



```node
Could not find a declaration file for module '@faker-js/faker'. 
'C:/repos/faker/dist/index.cjs' implicitly has an 'any' type.
There are types at 'c:/repos/faker-playground/node_modules/@faker-js/faker/dist/index.d.ts', 
but this result could not be resolved under your current 'moduleResolution' setting. 
Consider updating to 'node16', 'nodenext', or 'bundler'.ts(7016)
```

Steps to reproduce:

1. Checkput faker and faker-playground next to eachother
2. Clean install dependencies in faker
3. Build faker using `pnpm run build`
4. Override faker dependency in playground
5. Clean install dependencies in faker-playground
6. Run build in faker-playground using `pnpm run build`

Shinigami92:
Thanks @xDivisionByZerox for the review, I will have a look into this probably this evening

package.json
```diff
@@ -83,8 +90,11 @@
   "types": "index.d.ts",
```
mshima: Types value is wrong
```suggestion
  "types": "dist/index.d.ts",
```

package.json
```diff
@@ -83,8 +90,11 @@
...
   "typesVersions": {
     ">=5.0": {
-      "*": [
-        "dist/types/*"
+      ".": [
+        "./dist/index.d.ts"
+      ],
```
mshima, line 95: This is not needed with types field
```suggestion
```

Shinigami92: Is this not needed for Node10? Or in general (NodeNext, Node16 and other moduleResolutions)? Is "types" taking somehow precedence? Or is it the other way around?

package.json
Shinigami92: Oh my god I think I got it now 💡 

https://github.com/faker-js/faker/blob/18ab2c7b3c8fa78c46d4abafc07c3219f41caa46/package.json#L86

previously we had `*` here, but now we use `.`
so previously the `*` was put in front of everything, but the `.` is now a relative start from here
and this is why `dist/` is now required and fixes the current problem

Shinigami92:
> Repeating [#3093 (comment)](https://github.com/faker-js/faker/pull/3093?notification_referrer_id=NT_kwDOAG3Lq7MxMjI4NDk2OTcwODo3MTk1NTYz#discussion_r1752136938)

Oh my god I think I got it now 💡 

https://github.com/faker-js/faker/blob/18ab2c7b3c8fa78c46d4abafc07c3219f41caa46/package.json#L86

previously we had `*` here, but now we use `.`
so previously the `*` was put in front of everything, but the `.` is now a relative start from here
and this is why `dist/` is now required and fixes the current problem

package.json
mshima: > Is this not needed for Node10?

Node10 will use `types` in root for `.`.

> Or in general (NodeNext, Node16 and other moduleResolutions)? Is "types" taking somehow precedence? Or is it the other way around?

`exports` fields takes precedence for other moduleResolutions. Since specific types exists, `typesVersions` are ignored for them.

Shinigami92:
This PR now runs green against https://github.com/faker-js/playground/pull/7 on my local machine

notaphplover:
I just tested it against my issue repro repo and worked properly :smiley:

package.json
```diff
@@ -65,26 +64,37 @@
...
+        "types": "./dist/locale/*.d.ts",
+        "default": "./dist/locale/*.js"
+      }
     },
     "./package.json": "./package.json"
   },
   "main": "dist/index.cjs",
   "module": "dist/index.js",
-  "types": "index.d.ts",
+  "types": "dist/index.d.ts",
```
ST-DDT, line 90: Why is this no longer shortened? How can I test this and typesVersions?

tsconfig.build.json
```diff
@@ -1,16 +0,0 @@
...
-    "target": "ES2022",
-    "noEmit": false,
-    "emitDeclarationOnly": true,
-    "rootDir": "src",
-    "outDir": "dist/types",
-
-    // This negates what is set in the extended tsconfig.json
-    "skipLibCheck": false,
-    "allowSyntheticDefaultImports": false,
-    "resolveJsonModule": false
```
ST-DDT, line 13: Are these config options actually obsolete or should we enable them?
If they are obsolete why did we add them in the first place?

test/faker.spec.ts
```diff
@@ -18,6 +18,8 @@ describe('faker', () => {
       .filter((key) => typeof console[key] === 'function')
       .map((methodName) => vi.spyOn(console, methodName));
 
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: Types might or might not exist, depending on the runned scripts
```
ST-DDT:
```suggestion
    // @ts-ignore: Types may or may not exist, depending on whether the project was built first.
```

package.json
mshima: In current main `typesVersion` will remap root `types` to `dist/types/index.d.ts` due to `*` glob.
https://github.com/faker-js/faker/blob/18ab2c7b3c8fa78c46d4abafc07c3219f41caa46/package.json#L86-L88

Since the '*' glob was removed from `typesVersion` root types should map to the full path.

tsconfig.build.json
Shinigami92: No, they are turned on explicitly in `tsconfig.json`, here they were turned of for the build process, but this is now done by `tsup` as well.

ST-DDT (approved):

ST-DDT:
**Team Decision**

If all works well, we will merge and release this soon.
Additional testing and reviews are very much appreciated. ❤️

mshima (approved):
An additional recommendation is to use .d.cts types in root `types` and `typesVersions`. They are only used in node10 resolutions and probably will be used in commonjs projects.
But I haven’t tried that against playground.

ST-DDT:
> An additional recommendation is to use .d.cts types in root `types` and `typesVersions`. They are only used in node10 resolutions and probably will be used in commonjs projects. But I haven’t tried that against playground.

Thanks for the information. Do you have a source for that recommendation?
I was unable to find it and chatgpt seemed to be confused regarding that question.

mshima:
> Thanks for the information. Do you have a source for that recommendation?
> I was unable to find it and chatgpt seemed to be confused regarding that question.

https://www.typescriptlang.org/tsconfig/#moduleResolution
```
'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use node10 in modern code.
```
According to this, node10 will be used in legacy code that only supports CommonJS.
So a CommonJS type is probably better.

xDivisionByZerox (approved):

PR #3094, opened 2024-09-07
infra: replace tsup with packem

### LINKED ISSUE #3087: [9.0.0] Types are not properly emitted

prisis:
This is a different approach to fix #3087 

- Switched tsup with [packem](https://github.com/visulima/packem/tree/alpha)
- Reduced package size from `npm notice package size: 2.1 MB` `npm notice unpacked size: 8.6 MB` to `npm notice package size: 2.0 MB` `npm notice unpacked size: 6.8 MB` (hard to display, the newest version has more type files..) -> run `npm pack --dry-run` to validate it
- Fixed types for
  - "@faker-js/faker"  `node10` `node16 (from CJS)` `node16 (from ESM)` `bundler`
  - "@faker-js/faker/locale/*"  `node10` `node16 (from CJS)` `node16 (from ESM)` `bundler`
  - "@faker-js/faker/package.json"  `node10` `node16 (from CJS)` `node16 (from ESM)` `bundler`
- Changed exports to support the new types files
- The `typesVersions` are generated based on https://github.com/andrewbranch/example-subpath-exports-ts-compat/tree/main/examples/node_modules/types-versions-wildcards, this is handle by `packem`


ST-DDT:
Node10 works fine for me, what needs to be fixed?

prisis:
shini wanted that i split this MR into 2 -> shini did open new prs with the changes.
 
- [x] i need to make a change in packem to not normalize the package.json on the fly, (currently changing the code for this)
- [x] i need to remove ^ from the versions
- [x] need to move the typesVersions on the correct place

prisis:
~Should i move `attw` into a new PR?~ Will move it into a new PR

package.json
```diff
@@ -31,27 +31,105 @@
   "license": "MIT",
   "type": "module",
   "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "index.d.ts",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.mts",
   "typesVersions": {
-    ">=5.0": {
+    "*": {
```
prisis: Im not sure about this part, should it only allow the supported typescript version used in the root package or should it be configurable?

.github/workflows/ci.yml
```diff
@@ -173,8 +173,8 @@ jobs:
         env:
           CYPRESS_INSTALL_BINARY: 0
 
-      - name: Build types
-        run: pnpm run build:types
+      - name: Build
+        run: pnpm run build
```
prisis, line 177: working currently to allow packem, only generate types, this could be change later into `pnpm run build --dts-only`

package.json
prisis: This can be changed with the config file

```
rollup: {
        node10Compatibility: {
            writeToPackageJson: true,
            typeScriptVersion: '*'
        },
}
```

Shinigami92: > Im not sure about this part, should it only allow the supported typescript version used in the root package or should it be configurable?

we definitely use TS 5.0+ syntax (generic const templates), so we can be sure that everything below TS 5 is not supported

package.json
```diff
@@ -34,27 +34,105 @@
...
     ".": {
-      "types": "./dist/types/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs",
-      "default": "./dist/index.js"
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      },
+      "import": {
```
Shinigami92, line 69: **suggestion:** IMO this should be default, just to be really safe and we don't forget any unpopular exports (e.g. "node" or "browser")
better they ship incompatible versions but could be worked around somehow, as instead they are not shipped at all

same for line 132

package.json
```diff
@@ -123,10 +198,11 @@
     "prettier": "3.3.3",
     "prettier-plugin-organize-imports": "4.0.0",
     "rimraf": "5.0.10",
+    "esbuild": "0.23.1",
+    "@visulima/packem": "1.0.0-alpha.104",
```
Shinigami92: **nitpick:** pls run sort imports

ST-DDT:
I get the following warning during build:

> WARNING   [packem] [plugin:patch-types] packem_shared/airline-!~{001}~.d.cts contains confusing identifier names
- Casing$1

````ts
type Casing = 'upper' | 'lower' | 'mixed';
[...]
type Casing$1 = 'lower' | 'upper' | 'mixed';
````

ST-DDT:
I get the following error during playground build:

> Syntax Error: Thread Loader (Worker 0)
> ~\git\Faker\faker\dist\packem_shared\Faker-cIuF47cl.mjs: Static class blocks are not enabled. Please add `@babel/plugin-transform-class-static-block` to your configuration.

ST-DDT:
The build process also edits my package.json:

````diff
 "typesVersions": {
    ">=5.0": {
      ".": [
-        "./dist/index.d.ts"
+        "./dist\\index.d.ts"
      ],
      "locale/*": [
-        "./dist/locale/*.d.ts"
+        "./dist\\locale\\*.d.ts"
      ]
    }
  },
````

ST-DDT:
**Team Decision**

We will first merge #3093, but we keep this as an alternative once the remaining issues are resolved.

ST-DDT:
We should decide whether we want to proceed here or not.

ST-DDT:
**Team Decision**

- The current variant works for now and is battle tested
- We don't have capacity to push this to the finish line
- We plan to remove tsup/any build tool for v10 in favor of plain tsc

PR #3095, opened 2024-09-07
infra(ci): run against playground

mshima:
Main types `index.d.ts` doesn’t exist.
Fix types to match '.' exports.

ref: https://github.com/jhipster/generator-jhipster/pull/27179


test/faker.spec.ts
```diff
@@ -18,9 +18,6 @@ describe('faker', () => {
       .filter((key) => typeof console[key] === 'function')
       .map((methodName) => vi.spyOn(console, methodName));
 
-    const file: unknown = await import('..');
-    expect(file).toBeDefined();
-
```
mshima, line 23: Package import cannot be tested in vitest test, dist folder does not exist.

mshima:
Same fix is applied in https://github.com/faker-js/faker/pull/3094.
Alternative fix is applied in https://github.com/faker-js/faker/pull/3093 (removes types from root)

The above PRs tries to fix CommonJS types too, while this fixes esm types.

Shinigami92:
you can checkout https://github.com/faker-js/playground and test your changes

test/faker.spec.ts
Shinigami92: We cannot simply remove such a test, we added it on purpose

mshima: tsc should be executed before or type checks should be ignored.
Since index.d.ts does not exist, it considers that types does not exist.

mshima:
> you can checkout https://github.com/faker-js/playground and test your changes

Why don’t integrate playground on ci tests?

Shinigami92:
> > you can checkout [faker-js/playground](https://github.com/faker-js/playground) and test your changes
> 
> Why don’t integrate playground on ci tests?

PR welcome, go and try for it

Shinigami92:
@mshima thanks, this is a huge step forward and I wished for something like this a long time ❤️ 
you need to rebind/relink the repo faker to the playground test run
so something like modifying pnpm override before installing the playground deps

mshima:
@Shinigami92 I am on mobile now, so I cannot rebase.
I think it’s only missing steps names now.
Feel free to modify the PR.

Playground is passing 🚀

.github/workflows/playground.yml
```diff
@@ -0,0 +1,49 @@
+name: Playground
```
ST-DDT: Maybe edit/rename https://github.com/faker-js/faker/blob/next/.github/workflows/check-release-pr.yml instead?

mshima:
Playground seems to be missing pure node ts samples.

package.json
```diff
@@ -32,7 +32,6 @@
   "type": "module",
   "main": "dist/index.cjs",
   "module": "dist/index.js",
-  "types": "index.d.ts",
```
prisis: Why did you remove the types? 

This is needed for the node10 -> yeah currently its broken

mshima: PR was marked as invalid so I am testing playground behavior.

mshima: added again

test/faker.spec.ts
mshima: Should be added again ignoring type checks?

mshima: I think it should be tested in bundling process.

Shinigami92: as far as I know, we imported this to also test that it does not log stuff as a side effect
IMO you should readd it and potentially just mark with `@ts-ignore` for example

package.json
Shinigami92: are you sure that this works? because last time we touched this, we found out that `typesVersions` is taken first and so `dist/types/*` + `index.d.ts` => `dist/types/index.d.ts`

.github/workflows/playground.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Build Faker
+        run: pnpm install && pnpm build
+        working-directory: faker
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Check Playground
+        run: |
+          sed -i -e "s/overrides-for-dev/overrides/g" package.json
```
Shinigami92: **suggestion:** 
```suggestion
          sed -i -e "s/overrides-for-release/overrides/g" package.json
```

because `file:` is better than `link:`

package.json
prisis: The logic is a bit funny; When i was researching how to do it in packem, found a lot of different descriptions

The best way (to support all cases, happy to be corrected) is to have the "types" for the main export and the typesVersions for the subpath resolution.

.github/workflows/playground.yml
```diff
@@ -0,0 +1,49 @@
...
+          cat package.json
+          pnpm install --no-frozen-lockfile
+          pnpm -r --workspace-concurrency 1 test
```
Shinigami92: **suggestion:** maybe also add the build step

```suggestion
          pnpm -r --workspace-concurrency 1 build
          pnpm -r --workspace-concurrency 1 test
```

test/faker.spec.ts
Shinigami92: Maybe have a look into https://github.com/faker-js/faker/pull/1966

- https://github.com/faker-js/faker/pull/1966

package.json
Shinigami92: Maybe have a look into https://github.com/faker-js/faker/pull/1966

- https://github.com/faker-js/faker/pull/1966

test/faker.spec.ts
mshima: This is importing a mocked module, not the real faker.

Shinigami92:
AND NOW we really see the result! 🎉 

image

Right now, this behavior is correct! because we have this issue

test/faker.spec.ts
Shinigami92: > This is importing a mocked module, not the real faker.

This should import `dist/index.js` based on `package.json` module field

mshima: Maybe vitest maps the module.

test/faker.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('faker', () => {
       .filter((key) => typeof console[key] === 'function')
       .map((methodName) => vi.spyOn(console, methodName));
 
+    // @ts-ignore
```
Shinigami92: **suggestion:** 
```suggestion
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
```

test/faker.spec.ts
mshima: We cannot use expect error like eslint suggests since it won’t break if dist types can be on place

Shinigami92: > We cannot use expect error like eslint suggests since it won’t break if dist types can be on place

See https://github.com/faker-js/faker/pull/3095#discussion_r1748908325 😉

.github/workflows/playground.yml
Shinigami92: Maybe not "release", because I think this runs everytime

package.json
mshima: index.d.ts doesn’t exit so it’s probably a noop.

mshima:
For some reason mapping types to non existing index.d.ts fixes playground.

In the end it just adds the playground workflow

Shinigami92 (dismissed):

prisis (dismissed):

test/faker.spec.ts
```diff
@@ -18,6 +18,8 @@ describe('faker', () => {
       .filter((key) => typeof console[key] === 'function')
       .map((methodName) => vi.spyOn(console, methodName));
 
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
mshima:
```suggestion
    // @ts-ignore package types depends on been built
```

mshima: I will let maintainers to merge

ST-DDT:
Please note that this will create a bi-directional dependency between two separate repositories.

Shinigami92:
> Please note that this will create a bi-directional dependency between two separate repositories.

Do you see any problem with this? IMO it's fine, because playground is in out org as well.

.github/workflows/playground.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Build Faker
+        run: pnpm install && pnpm build
+        working-directory: faker
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+
+      - name: Check Playground
+        run: |
+          sed -i -e "s/overrides-for-release/overrides/g" package.json
+          cat package.json
```
xDivisionByZerox: Do we need to print the file here or is this from previous debugging sessions?

xDivisionByZerox (commented):
If we activate the playground for all PRs why would we need [check-release-pr.yml](https://github.com/faker-js/faker/blob/next/.github/workflows/check-release-pr.yml) anymore? 

> Do you see any problem with this? IMO it's fine, because playground is in out org as well.

[We had a reason](https://github.com/faker-js/faker/pull/2988#issuecomment-2207029693) why we didn't do this previously. Now a random contributor might need to apply changes in the playground onto of the changes in faker when they work a breaking change feature.

mshima:
IMO running playground in CI:
- reduces maintenance burden by avoiding regressions and requiring manual checks: https://github.com/faker-js/faker/pull/3093#issuecomment-2339729022.
- improves contribution experience by not having to ask to manually check against playground: https://github.com/faker-js/faker/pull/3095#issuecomment-2335533878

> We could do that but it has a side effect, that all features that are used in the playground that might experience breaking changes need to be updated to the new feature before it is available as a dependency. This is also the case for the releasenPR but that those are even rarer.

- the workflow can be ignored if it's not required to pass.
- contributor will be warned about breaking changes or unwanted breaking changes.

Moving playground to the main repository can be considered too.

Shinigami92:
> * the workflow can be ignored if it's not required to pass.
> * contributor will be warned about breaking changes or unwanted breaking changes.

I agree 👍 we can mark the CI check as "not required" and if it breaks something, we are just "informed"

.github/workflows/playground.yml
mshima:
```suggestion
```

Shinigami92 (dismissed):

ST-DDT:
> I agree 👍 we can mark the CI check as "not required" and if it breaks something, we are just "informed"

Please note that that will only work once and will cause confusion for future PRs (that will fail the CI even if they didn't break anything because next is already "broken").

I would rather move the playgrounds inside the main repo to avoid these bidirectional dependencies.

ST-DDT:
**Team Decision**

We don't want to add bi-directional dependencies between the playground and the faker repo.
We think that moving the playground to the main repo is the best solution for that, but that would require extensive work e.g. converting the repo into a mono-repo, what we don't want to do yet.

We would like to still run the pipeline on demand somehow e.g. using a comment, but we are currently unsure about the exact expectation we have for the pipeline. We will talk about this again in a future team meeting.

ST-DDT:
If you have an idea, on how to move the playground to the main repo feel free to open a PR with a suggestion.

ST-DDT:
We are looking into restructuring our project setup to solve this issue. For now, you don't have to update this PR.
Thanks for bringing this to our attention.

xDivisionByZerox:
I'm closing this, since we introduced the ability to automatically test our Faker against the Faker playground in #2988 as part of the restructuring of our automatic release process.

- #2988

PR #3109, opened 2024-09-14
refactor(image): deprecate avatarLegacy for removal

### LINKED ISSUE #2934: Prefer a stable CDN over IPFS
### Pre-Checks

### Describe the bug

I am using `faker.image.avatar()` method to get a random avatar URL.

The method returns a random IPFS URL: `https://cloudflare-ipfs.com/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/269.jpg`

But to my surprise, the Cloudflare IPFS Gateway returns 504 error:

image

The IPFS gateway itself is working, but the asset can not be fetched by the gateway from the IPFS network.

### Minimal reproduction code

_No response_

### Additional Context

IMHO IPFS Gateway might not be that stable after all, `faker-js` could publish the random assets on a dedicated GitHub repo and use jsDelivr (or any other CDN) to deliver them.

### Environment Info

```shell
System:
    OS: Linux 6.8 Fedora Linux 40 (Workstation Edition)
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 11.70 GB / 31.03 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 18.19.0 - /run/user/1000/fnm_multishells/174660_1717401217872/bin/node
    npm: 10.2.3 - /run/user/1000/fnm_multishells/174660_1717401217872/bin/npm
    pnpm: 8.15.8 - /run/user/1000/fnm_multishells/174660_1717401217872/bin/pnpm
    bun: 1.1.10 - ~/.npm-global/bin/bun
  Browsers:
    Chrome: 125.0.6422.141
```

- [X] CJS
- [ ] ESM

### Used Package Manager

pnpm

ST-DDT:
Makes #2934 obsolete

- #2934 

---

The `faker.image.avatarLegacy()` method no longer produces useful results, so we should remove it soon.

Theoretically, we could just remove it, but I'm not sure that is a good solution since it is technically a breaking change.

https://deploy-preview-3109.fakerjs.dev/api/image.html#avatarlegacy


matthewmayer:
Could we make it an alias to avatar for the time being?

xDivisionByZerox (approved):

xDivisionByZerox:
> Could we make it an alias to avatar for the time being?

I wouldn't do that, since that takes away some of the incentive to actually migrate.

matthewmayer:
Shipping a version with urls we know to be broken doesn't seem ideal either though.

ST-DDT:
If someone uses the legacy method, they very intentionally used that method because they wanted that image set and not the GH avatars.
I also considered throwing an error instead but that would break other peoples applications.
Not having images will hopefully bring them to investigate why and replace the method.

I agree that this situation is not ideal, but I don't see a solution that wouldn't break our tests or specification.

We should add a test, to ensure that our generated links are still valid (if they are expected to be).

Shinigami92:
Right now, these would e.g. "fail" with an



when used for demo apps

And this method is used in my previous company. I could ask/notify a friend there the next few days and tell them, that they need to find a new way to fake human faces.

ST-DDT:
> And this method is used in my previous company. I could ask/notify a friend there the next few days and tell them, that they need to find a new way to fake human faces.

Potential Workaround:

````ts
faker.image.urlLoremFlickr({ category: 'portrait', width: 400, height: 400 })
````

**Comment by Shinigami92** [comment]
> > And this method is used in my previous company. I could ask/notify a friend there the next few days and tell them, that they need to find a new way to fake human faces.
> 
> Potential Workaround:
> 
> ```ts
> faker.image.urlLoremFlickr({ category: 'portrait', width: 400, height: 400 })
> ```

Nah, this wont work for the data they need. They need real human faces. Like ID card pictures.

And this.... is not something like that XD

<details>
<summary>Details</summary>




</details>

**Comment by matthewmayer** [comment]
We could consider pointing to a static image with a message that avatarLegacy is deprecated? More useful than a broken image.

**Comment by matthewmayer** [comment]
I'd like to get a basic version of https://github.com/faker-js/faker/issues/465 implemented for 9.1.0 so we have a more useful replacement to point people to.

**Comment by Shinigami92** [comment]
> We could consider pointing to a static image with a message that avatarLegacy is deprecated? More useful than a broken image.

Sounds like a quick good work around for the moment.

> I'd like to get a basic version of #465 implemented for 9.1.0 so we have a more useful replacement to point people to.

Not sure if 9.1, but potentially a 9.X.

**Comment by ST-DDT** [comment]
> > We could consider pointing to a static image with a message that avatarLegacy is deprecated? More useful than a broken image.
> 
> Sounds like a quick good work around for the moment.

Which image should I use instead?

Something like this?

grafik

**Comment by Shinigami92** [comment]
> Which image should I use instead?
> 
> Something like this?
> 
> grafik

Partly yes, but no...
`faker.image.avatar()` does not solve the same purpose and is more targeted for "just" internet profile pictures => avatars
So maybe `faker.image.avatarLegacy() is broken, pls help/contribute`

**Comment by ST-DDT** [comment]
> So maybe `faker.image.avatarLegacy() is broken, pls help/contribute`

Maybe like this?

> faker.image.avatarLegacy() is borken
> use faker.image.avatar() instead
> or consider contributing to #465

**Comment by Shinigami92** [comment]
> > So maybe `faker.image.avatarLegacy() is broken, pls help/contribute`
> 
> Maybe like this?
> 
> > faker.image.avatarLegacy() is borken
> > use faker.image.avatar() instead
> > or consider contributing to #465

could be a compromise

maybe `use faker.image.avatar() instead for PFP` (https://www.dictionary.com/e/acronyms/pfp/)

**Comment by ST-DDT** [comment]
> maybe `use faker.image.avatar() instead for PFP` ([dictionary.com/e/acronyms/pfp](https://www.dictionary.com/e/acronyms/pfp/))

The fact that you have to put a dictionary there kind of disqualifies that abbreviation.
IMO the term avatar is well understood and more known than PFP.

**Comment by matthewmayer** [comment]
Maybe open a new GH discussion and just embed the full URL to that discussion in the image. Easier to edit that with a longer description of possible alternatives than trying to squeeze it into an image.

**Comment by matthewmayer** [comment]
I discovered the images are still available on an alternative IPFS endpoint at https://gateway.pinata.cloud/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/ if we want to retrieve the images from there.

**Comment by ST-DDT** [comment]
> I discovered the images are still available on an alternative IPFS endpoint at https://gateway.pinata.cloud/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/ if we want to retrieve the images from there.

Those images have too many issues. We wont bring them back.

**Comment by ST-DDT** [comment]
**Team Decision**

We will deprecate the method without changing the target urls.

- If we replace the generated links with a custom image, we have to host that image somewhere FOREVER
- We don't plan on backporting the fix to older methods either, it will just provide a link that is no longer working.
- The method will not throw an error itself in case somebody only needs the link (the string value)

**Review by matthewmayer — APPROVED** [review]

**Review by ejcheng — APPROVED** [review]

File: src/modules/image/index.ts [hunk 1]
```diff
@@ -55,9 +52,17 @@ export class ImageModule extends ModuleBase {
...
    * @since 8.0.0
+   *
+   * @deprecated The links are no longer working. Use `avatar()` instead.
    */
-  // This implementation will change in the future when we tackle https://github.com/faker-js/faker/issues/465.
   avatarLegacy(): string {
+    deprecated({
+      deprecated: 'faker.image.avatarLegacy()',
+      proposed: 'faker.image.avatar()',
+      since: '9.0.1',
```
Shinigami92, line 62: **todo:** we need to fix this `since` before releasing next version, because 9.0.1 is already published

src/modules/image/index.ts
```diff
@@ -55,9 +52,17 @@ export class ImageModule extends ModuleBase {
...
    *
    * @since 8.0.0
+   *
+   * @deprecated The links are no longer working. Use `avatar()` instead.
    */
-  // This implementation will change in the future when we tackle https://github.com/faker-js/faker/issues/465.
   avatarLegacy(): string {
+    deprecated({
+      deprecated: 'faker.image.avatarLegacy()',
+      proposed: 'faker.image.avatar()',
```
Shinigami92, line 61: **issue:** `faker.image.avatar()` is not a direct valid replacement, because it serves a slightly different purpose than it did before

src/modules/image/index.ts
```diff
@@ -55,9 +52,17 @@ export class ImageModule extends ModuleBase {
...
+      since: '9.0.1',
+      until: '10.0.0',
+    });
+
     return `https://cloudflare-ipfs.com/ipfs/Qmd3W5DuhgHirLHGVixi6V76LhCkZUz6pnFt5AJBiyvHye/avatar/${this.faker.number.int(
       1249
     )}.jpg`;
```
Shinigami92, line 68: **suggestion:** @matthewmayer had a potentially good idea here: https://github.com/faker-js/faker/pull/3109#issuecomment-2353009329
Should we consider it? If not, why?

Shinigami92 (commented):
I would like to add, that I was at a conference while the "team decision" was made.

src/modules/image/index.ts
ST-DDT: It is definition wise. The old images happen to be useful for a different purpose. We dont have a replacement for that secondary purpose yet. If we add a method that solves that usecase we can add the alternative with that PR.

src/modules/image/index.ts
ST-DDT: Are you referring to the discussion or the image/the link within?

src/modules/image/index.ts
ST-DDT: See also our meeting notes.

src/modules/image/index.ts
ST-DDT: Fixed in #3124

src/modules/image/index.ts
Shinigami92: The link. So the idea is to `return 'https://github.com/faker-js/faker/discussions/XXXX';` where XXXX is the new discussion.
This will still result then in a missing image tag.

e.g.:

ST-DDT: Changing the current link breaks our tests and might be breaking for some of our users.
This isn't meant as a strict rejection, just a reminder that every change breaks someones workflow (similar to how avatar() would break the usecase for needing portraits, even though it isn't defined anywhere).
That's one of the reasons, why I would to be very specific with the expectations we have for the results of a method.
(aka should we provide metadata or would these serve a purpose not aligned with the method that serves the images)

https://xkcd.com/1172/

On the other hand, if the users don't even look at our docs and their own logs, why would they read a random link in the browser?

Shinigami92: Did not considered these points 👍👌 
So we could then instead throw a runtime warning with a link 🤔

PR #3119, opened 2024-09-16
infra: publish releases with provenance

ST-DDT:
This PR adds release providence to our releases on npm.

## Example:

[{FF85472C-BA5B-4A54-B066-1852C76586BC}](https://www.npmjs.com/package/@vitest/mocker)

I haven't tested the pipeline changes myself, but I verified with various other libraries, that the syntax should work.

I decided to use an ENV variable to set the flag as I consider it easier to read than the other alternatives (--flag, .npmrc).
Also it does not prevent publishing from our own computers, if ever needed again.

## Example projects

- https://github.com/nuxt/nuxt/commit/ec1503731b1cabb331c28726bba3fb4d4e2d3af8
- https://github.com/toeverything/blocksuite/commit/b6cb9111f9414b2de6b595ac3060ba0f14304f2f
- https://github.com/vitest-dev/vitest/blob/main/.github/workflows/publish.yml#L10 + https://github.com/vitest-dev/vitest/commit/08824602aa10962668ca20ab64b4f8ba79fd27d9

## See also

- https://github.com/npm/provenance
- https://docs.npmjs.com/cli/v10/commands/npm-publish#provenance


Shinigami92 (approved):
Cool

xDivisionByZerox (approved):

PR #3127, opened 2024-09-21
test: verify the generated image links are working

### LINKED ISSUE #3112: Add tests to methods that produce links to ensure they still work
All methods that produce links pointing to actual resources should have a test ensuring the link resolves with a HTTP-200 (redirects to an actual page are allowed).

This mainly applies to the image module, other modules have to be checked for those links.

The tests ensure we notice broken methods sooner (prior to a release).

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/issues/3109#issuecomment-2351131970_

ST-DDT:
Fixes #3112

- #3112 

This PR adds an integration test for the image module/all methods that return a url that actually points to something to ensure the url is working as expected.

The integration tests are run once a week or on demand.


Shinigami92:
The current provided solution feels like mixin unit with integration tests. IMO we should split them from each other. That way, because we rely here also on external sources, we could execute this in a scheduled workflow e.g. every week or so, so we also reduce the "ddos" outgoing from this repo.

test/modules/image.spec.ts
```diff
@@ -92,23 +142,22 @@ describe('image', () => {
...
-      expect(avatarUrl).toBeTypeOf('string');
-      expect(avatarUrl).toMatch(/^https:\/\//);
-      expect(() => new URL(avatarUrl)).not.toThrow();
+      await assertWebAddress(avatarUrl);
     });
   });
 
   describe('avatarGitHub', () => {
-    it('should return a random avatar url from GitHub', () => {
+    it('should return a random avatar url from GitHub', async () => {
```
matthewmayer: Isn't this going to be flaky? As not all GitHub ids return real users? (Eg a deleted user).

ST-DDT: I didn't run into that issue yet. If it is actually an issue, I would rather know this, because it means the method is broken or unreliable.

matthewmayer: apparently if a user is deleted
e.g. https://api.github.com/user/13

then the avatar URL just returns a Github logo
https://avatars.githubusercontent.com/u/13

ST-DDT: It is still a working image, so I consider this a non issue.

ST-DDT:
**Team Decision**

- We will move these tests to `integration-test`/`test/integration`
- We will run the integration test once a week

ST-DDT:
@xDivisionByZerox Like this or should I really use a separate folder structure?

xDivisionByZerox:
> @xDivisionByZerox Like this or should I really use a separate folder structure?

Thats how I do it in my projects, **but** I have my test files next to the actual implementation.

- directory
  - file.ts
  - file.spec.ts (unit tests)
  - file.(integration|int|it)[.-]spec.ts
  - file.(e2e|cy)[.-]spec.ts

The structure feels natural to **me**.

Since the tests are already in a separate directory, one might suggest splitting integration and unit tests into two directories as well:

- test
  -  integration
    - file.(integration|int|it)?[.-]spec.ts
  -  unit
    - file.spec.ts

On advantage would be that the test setup file might be more linear, since the file patter can be a diretory instead of a "crazy" pattern. Additionally, if integration tests do not have to follow the same structure like unit tests do, this might be a good approche as well.

ST-DDT:
There are currently too many open PRs to move the unit tests to a subfolder.
So I will only move the integration tests to a subfolder.

xDivisionByZerox (approved):

Issue #3107, opened 2024-09-12
`faker.time` module to simplify time (not datetime) usecases


douglasg14b:
### Clear and concise description of the problem

Generating fakes specifically for times, not necessarily dates. These will only produce times, and not consider dates.

The date fucntions do perform this utility to some degree but require additional work to achieve which could be baked into a time module directly.

### Suggested solution

in module `faker.date` or in a new module `faker.time` it would have fucntions for:

- `faker.time.between`
- `faker.time.after`
- `faker.time.range`
- ....etc

These should accept a format argument that allows the time output to be formatted as 24/12h time (or others)

### Alternative

Users build their own time fucntions that work on top of `faker.date`

### Additional context

Naive examples from my own use:

```ts
function getFakeTime(options?: { from?: Date; to?: Date } = {}) {
    const from = options?.from || '2024-01-01T00:00:00.000Z';
    const to = options?.to || '2024-01-01T23:59:59.999Z';

    return faker.date.between({from, to});
}

function getFake24hTime(time?: Date) {
    const dateTime = time || getFakeTime();
    return dateTime.toISOString().slice(11, 16);
}

function getFake24hTimeRange() {
    const from = getFakeTime();
    const to = getFakeTime({from});
    return {
        from: getFake24hTime(from),
        to: getFake24hTime(to),
    }
}
```

ST-DDT:
> The date functions do perform this utility to some degree but require additional work to achieve which could be baked into a time module directly.

Could you explain which issues you are encountering with the date module when generating times other than `date.toISOString().slice(11, 16)`?

Maybe these issues also apply to the date module in general and just haven't been raised yet.

---

> * faker.time.range

> ```ts
> return {
>         from: getFake24hTime(from),
>         to: getFake24hTime(to),
>     }
> ```

IMO it is unlikely that we add a method like this, because it is likely a user dependent structure. e.g. maybe you use `{from,to}`, others might want to use `{start,end}` or `from-to`.

As a workaround for that, I can recommend the [`faker.date.betweens({ from, to, count: 2)`](https://fakerjs.dev/api/date.html#betweens) function (or from the time module, if we add it). It returns a sorted list, so you can just use it like this:

````ts
const [ rangeFrom, rangeTo ] = faker.date.betweens({ from, to, count: 2);
return {
    from: toTime(rangeFrom),
    to: toTime(rangeTo ),
}
````

---

> These should accept a format argument that allows the time output to be formatted as 24/12h time (or others)

[...]

PR #3128, opened 2024-09-22
docs(date): add hints how to generate date ranges

ST-DDT:
Picks up a suggestion from #3107 and explains how date ranges can be generated easily.

- #3107

Preview: https://deploy-preview-3128.fakerjs.dev/api/date.html#overview

grafik

Not sure, about the phrasing and location of the text.


src/modules/date/index.ts
```diff
@@ -567,6 +567,11 @@ export class SimpleDateModule extends SimpleModuleBase {
  *
  * For more control, any of these methods can be customized with further options, or use [`between()`](https://fakerjs.dev/api/date.html#between) to generate a single date between two dates, or [`betweens()`](https://fakerjs.dev/api/date.html#betweens) for multiple dates.
  *
+ * If you need to generate a date range (start-end), you can do so using either of these two methods:
+ *
+ * - `const [start, end] = faker.date.betweens({ from, to, count: 2 });`
```
xDivisionByZerox: This will be problematic when a user uses TS strict mode (as we expect) with the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) flag (which I'd argue to be a strict option as well). [You can verify this in stackblitz yourself.](https://stackblitz.com/edit/faker-js-date-range-hpszb?file=index.ts&view=editor)

ST-DDT: AFAIK that one isnt part of simply strict but extended strict. Otherwise faker would have that enabled as well. And IIRC it causes a bunch of issues with regex and dynamic arrays as the ts type system is badly suited for those. Thanks for including a stackblitz for verification.

If you have the extended strict, then you can use the other example.

What should I do now? Remove it? Flip them? Create another alternative? Propose a new method for that usecase?

xDivisionByZerox: First and foremost, I wanted to provide visibility for this potential issue as I'm not sure myself what to do with it. Maybe discuss this on Thursday?

xDivisionByZerox: **Team Decision:**

We switch the example entries so the more generic one will be first. The destruction example will also receive a comment about the potential problem with `noUncheckedIndexedAccess`.

ST-DDT: Applied (although I don't particularity like the comment in its current form)

Shinigami92 (approved):

PR #2557, opened 2023-11-21
refactor(helper): deprecate replaceSymbolWithNumber

suyashgulati:
Fixes #1994

Changes done:
- deprecated replaceSymbolWithNumber
- created a new legacy method legacyReplaceSymbolWithNumber
- methods using replaceSymbolWithNumber now use the legacy method instead

- no update on unit tests required

PR #3130, opened 2024-09-22
refactor(internet): rename userName method to username

### LINKED ISSUE #3023: Rename `userName` to `username`
The [`userName`](https://fakerjs.dev/api/internet.html#username) method in the internet module should be named `username` – consider deprecating the former.

https://english.stackexchange.com/questions/43436/username-user-name-or-user-name

suyashgulati:
Aims to resolve #3023

As the team has decided to user username as a convention over the current userName method name.
This PR refactors the code to use username instead.


docs/guide/upgrading.md
```diff
@@ -445,7 +445,7 @@ Removed deprecated internet methods
 | `faker.internet.avatar()`                                      | `faker.image.avatarLegacy()` or `faker.image.avatar()`            |
 | `faker.internet.email(firstName, lastName, provider, options)` | `faker.internet.email({ firstName, lastName, provider, ... })`    |
 | `faker.internet.exampleEmail(firstName, lastName, options)`    | `faker.internet.exampleEmail({ firstName, lastName, ... })`       |
-| `faker.internet.userName(firstName, lastName)`                 | `faker.internet.userName({ firstName, lastName })`                |
```
matthewmayer: this shouldn't be changed

src/modules/internet/index.ts
```diff
@@ -186,18 +186,18 @@ export class InternetModule extends ModuleBase {
...
+   * faker.internet.username({ firstName: 'Jeanne', lastName: 'Doe'}) // 'Jeanne_Doe98'
+   * faker.internet.username({ firstName: 'John', lastName: 'Doe' }) // 'John.Doe'
+   * faker.internet.username({ firstName: 'Hélene', lastName: 'Müller' }) // 'Helene_Muller11'
+   * faker.internet.username({ firstName: 'Фёдор', lastName: 'Достоевский' }) // 'Fedor.Dostoevskii50'
+   * faker.internet.username({ firstName: '大羽', lastName: '陳' }) // 'hlzp8d.tpv45' - note neither name is used
    *
    * @since 2.0.1
    */
-  userName(
+  username(
```
matthewmayer, line 330: you need to leave the old version as an alias, with a `@see` to the new name, and mark it `@deprecated` until v10

The new method should be marked `@since 9.1.0`

test/modules/internet.spec.ts
```diff
@@ -51,7 +51,7 @@ describe('internet', () => {
         });
     });
 
-    t.describe('userName', (t) => {
+    t.describe('username', (t) => {
```
matthewmayer, line 69: need to test both old and new method names during v9.

src/modules/internet/index.ts
matthewmayer: see https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#deprecation-workflow

suyashgulati:
I wasn't aware the old one needs to be deprecated, I expected it to be refactor issue.
I will make changes.
thanks @matthewmayer

suyashgulati:
@matthewmayer I have made the changes and deprecated the userName 
But if I am using testing both new and deprecated code I am not able to pass the `preflight` locally.

Screenshot 2024-09-23 at 2 41 39 AM

Its showing cannot use deprecated code.
Can you suggest what do to in that case?

xDivisionByZerox:
The error messages are correct as eslint is warning you about the usage of deprecated code. In this specific case, it is obviously ok, since you just maked this method as deprecated. We have this rule to make sure that we are no longer using the deprecated method in any of the live source code.

For the test files you can use a comment to disable the check on the next line. An example would currently be in the image module test where `avatarLegacy` was deprecated:
https://github.com/faker-js/faker/blob/c15b48864ac6cd7a5cb1ba23e4e555c9cdfe1328/test/modules/image.spec.ts#L116-L124

suyashgulati:
Issue: Unit Tests are failing at checking the deprecated tag on new `username `method
RCA: I found out we are creating a new ephemeral files to execute examples from the js doc in `test/scripts/apidocs/verify-jsdoc-tags.spec.ts` being written [here](https://github.com/faker-js/faker/blob/6d133b13abab1ef9cfb72b5f3ad9136ecd4b41ba/test/scripts/apidocs/verify-jsdoc-tags.spec.ts#L143)

As `userName` and `username` have same name but different case the execution file for `username` is not getting created reason being MacOS and windows have case insensitive file systems.
Because of that when running examples for `username` it picks up examples of `userName` instead and shows the deprecated console.warn.
Reference: https://github.com/nodejs/help/issues/2116

Please suggest a way around this.

matthewmayer:
Great detective work!

ST-DDT:
You can use the following as a workaround:

https://github.com/faker-js/faker/blob/6e72712ffe878a9b30b6bf8a83239775adf1f176/test/scripts/apidocs/verify-jsdoc-tags.spec.ts#L33

````ts
function resolvePathToMethodFile(
  moduleName: string,
  methodName: string,
  signature: number
): string {
  const dir = resolveDirToModule(moduleName);

  // TODO @ST-DDT 2024-09-23: Remove this in v10
  if (methodName === 'userName') {
    methodName = 'userName2';
  }

  return resolve(dir, `${methodName}_${signature}.ts`);
}
````

File: src/modules/internet/index.ts [hunk 4]
```diff
@@ -183,7 +184,7 @@ export class InternetModule extends ModuleBase {
    * @param options.firstName The optional first name to use. If not specified, a random one will be chosen.
    * @param options.lastName The optional last name to use. If not specified, a random one will be chosen.
    *
-   * @see faker.internet.displayName(): For generating an Unicode display name.
+   * @see faker.internet.username(): For generating an Unicode display name.
```
Shinigami92: **issue:** I think this change was not intended

src/modules/internet/index.ts
```diff
@@ -196,6 +197,8 @@ export class InternetModule extends ModuleBase {
    * faker.internet.userName({ firstName: '大羽', lastName: '陳' }) // 'hlzp8d.tpv45' - note neither name is used
    *
    * @since 2.0.1
+   *
+   * @deprecated Use `faker.internet.username()` instead. Example: `faker.internet.username();`
```
Shinigami92: **question:** Do we have such an example in a `@deprecated` anywhere else?

suyashgulati: Thanks @Shinigami92 for the review.

Docs also mentioned similar thing:
```
/**
 * @see faker.cat.random()
 *
 * @deprecated Use `faker.cat.random()` instead.
 */
get cat() {
  deprecated({
    deprecated: 'faker.animal.cat',
  });
  return 'cat';
}
```


Also I used similar deprecation message in an [old PR](https://github.com/faker-js/faker/pull/2557/files#diff-3f6b4fc424ba2cedcbf5f612ba256d248f0c05f46314eed1dfa81590bcd429fdR250).

If not, can you suggest the right way?

README.md
```diff
@@ -65,7 +65,7 @@ const { faker } = require('@faker-js/faker');
 export function createRandomUser() {
   return {
     userId: faker.string.uuid(),
-    username: faker.internet.userName(),
+    username: faker.internet.username(),
```
matthewmayer: We have had problems before where people copy paste from the README and are confused when it doesn't work (especially during the gap between the change landing on next and the release). I wonder if we should add a comment to this line like

`// before version 9.1.0, use userName()`

src/modules/internet/index.ts
```diff
@@ -212,6 +215,56 @@ export class InternetModule extends ModuleBase {
        */
       lastName?: string;
     } = {}
+  ): string {
+    deprecated({
+      deprecated: 'faker.internet.userName()',
+      proposed: 'faker.internet.username()',
+      since: '2.0.1',
```
matthewmayer: I think this is supposed to be the first version it was deprecated ie 9.1.0?)

matthewmayer:
I noticed on
https://deploy-preview-3130.fakerjs.dev/api/

both username and userName link to https://deploy-preview-3130.fakerjs.dev/api/internet.html#username (as the hash fragments are lowercased)

src/modules/internet/index.ts
matthewmayer: i think just change
<code>
@deprecated Use `faker.internet.username()` instead. Example: `faker.internet.username();`
</code>
to simply
<code>
@deprecated Use `faker.internet.username()` instead.
</code>

xDivisionByZerox:
> I noticed on [deploy-preview-3130.fakerjs.dev/api](https://deploy-preview-3130.fakerjs.dev/api/)
> 
> both username and userName link to [deploy-preview-3130.fakerjs.dev/api/internet.html#username](https://deploy-preview-3130.fakerjs.dev/api/internet.html#username) (as the hash fragments are lowercased)

For me it two different ones:

- username: https://deploy-preview-3130.fakerjs.dev/api/internet.html#username
- userName: https://deploy-preview-3130.fakerjs.dev/api/internet.html#username-1

Am I missing something?

matthewmayer:
The links from the https://deploy-preview-3130.fakerjs.dev/api/ page both go to the same place.

ST-DDT:
> The links from the https://deploy-preview-3130.fakerjs.dev/api/ page both go to the same place.

I pushed a fix that should solve that issue.

src/modules/internet/index.ts
ST-DDT:
```suggestion
   * @deprecated Use `faker.internet.username()` instead.
```

suyashgulati: Resolved

docs/guide/upgrading.md
ST-DDT: I would like to put this up for consideration again.
Why should we not do this?
What do the others think?

ST-DDT (dismissed):

matthewmayer (dismissed):

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):

ejcheng (dismissed):

ST-DDT:
Dear @suyashgulati,

Thank you for your patience and contributions so far. We wanted to provide you with a brief update.

We are pleased to inform you that we have addressed all the known bugs in version 9.0. We will now take a short break and resume work around October 7, 2024, to begin merging new features for v9.1, including yours.

In the meantime, we hope you have a great time, and we sincerely appreciate your ongoing support and contributions.

Best regards,
your FakerJS Team

ST-DDT (approved):

ST-DDT:
Thanks for your contribution.
This change will be included in the next faker release v9.1.0.

PR #3144, opened 2024-09-25
fix(image): fix dataUri with type svg-base64 in browsers

ST-DDT:
Fixes #3143

- #3143

Replaces the problematic `Buffer.from(input).toString('base64')` call with a custom replacement that also works in browsers.
Also adds an eslint rule that should prevent this error in the future. 

Similar to the `toBase64URL` method from #2936

- #2936


eslint.config.ts
```diff
@@ -205,6 +204,7 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
     files: ['src/**/*.ts'],
     rules: {
       'jsdoc/require-jsdoc': 'error',
+      'no-restricted-globals': ['error', 'Intl', 'Buffer'],
```
ST-DDT: We could extend this to all globals that aren't available in the browser.

````ts
import globals from 'globals';

const browserGlobals = new Set(Object.keys(globals.browser));
const disallowedGlobals = Object.keys(globals.node).filter(
  (e) => !browserGlobals.has(e)
);
````

```suggestion
      'no-restricted-globals': ['error', 'Intl', ...disallowedGlobals],
```

matthewmayer (commented):
Could we use Buffer.from where available (as it's faster) and only fallback to the custom code when needed?

eslint.config.ts
xDivisionByZerox: This looks promising. I would argue that the `Set` is not required since there cannot be duplicated object keys.  Instead of `browserGlobals.has(e)` you could then use `browserGlobals.includes(e)`.

---

Would this have drastic decreases in performace for the lint step? 🤔

ST-DDT: > Would this have drastic decreases in performace for the lint step? 🤔

I haven't tested that yet, but I kind of don't like it because I have to add a new dependency for that.
It works for the cypress tests out of the box, not sure which property is responsible for that.

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3154, opened 2024-10-01
feat(locale): add Swedish(sv) counties

johanholmerin:
Based on the list at https://sv.wikipedia.org/wiki/Sveriges_l%C3%A4n#Lista_%C3%B6ver_Sveriges_l%C3%A4n

Note: although län is the administrative area that translates to [counties](https://en.wikipedia.org/wiki/Counties_of_Sweden), it would also be useful to have support for [municipalities](https://en.wikipedia.org/wiki/Municipalities_of_Sweden)(kommuner), since that's whats used for addresses in Sweden.


ejcheng (dismissed):

matthewmayer:
Generally state() returns the "level 1 administrative areas" and county() returns the "level 2 administrative areas" regardless of what they are actually called in that language. 

Does that make sense for sv?

johanholmerin:
Thanks @matthewmayer, that helped make things clearer. I've switched to using municipalities(kommuner) instead(based on the official list at https://skr.se/skr/tjanster/kommunerochregioner/kommunerlista.1246.html). The län were already available as state().

xDivisionByZerox (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #3162, opened 2024-10-08
feat(location): add `continent` method

### LINKED ISSUE #3164: `faker.location.continent` for generating random (real) continents

I am currently generating fake data for results from an OpenAPI spec. The spec has an entity that is location data. It includes a continent. See https://github.com/planet-a-ventures/affinity-node/blob/4df29ba8e97f02715f381da76efe13146463d0a1/openapi/2024-10-08.json#L2827

As a developer using faker I want to be able to generate random continent names.

### Suggested solution

A new `locations.continent` method.

### Alternative

_No response_

### Additional context

Implementation is here: https://github.com/faker-js/faker/pull/3162

joscha:
Adds a new function to the `location` module that generates the name of a random [continent](https://en.wikipedia.org/wiki/Continent).


src/modules/location/index.ts
```diff
@@ -216,6 +216,20 @@ export class LocationModule extends ModuleBase {
     );
   }
 
+  /**
+   * Returns a random continent name.
+   *
+   * @example
+   * faker.location.continent() // 'Asia'
+   *
+   * @since 9.0.4
```
ST-DDT: This will be 9.1.0 at the earliest.

ST-DDT (commented):
Please note, that we (usually) require an issue first.

https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md

src/modules/location/index.ts
joscha:
```suggestion
   * @since 9.1.0
```

joscha:
> Please note, that we (usually) require an issue first.

~Do we need that in this case? It feels like an extremely simple case. If we don't want it, I can close the PR, no hurt feelings, if we do want it there is not a lot more background to it :)~

See https://github.com/faker-js/faker/issues/3164

matthewmayer:
I feel this is such a harmless addition we can probably skip the usual
- make issue
- wait for 10 upvotes
- implement 

I can't really think of any way this api could be extended. And it's a trivial amount of data. 

We have spent a long time improving the code and processes. I think in various 8.x releases we can take advantage of that and add some "low-hanging fruit" methods.

ST-DDT:
In this case it is probably fine (thats why I put the usually in parethesis).

One lesson we learned for sure is that lot location stuff seems simple, but the usecases vary a lot. Some people might need only the name, others might only need the code, and some need both related to each other (e.g. countries).

I did a quick google search and found this list of continent codes.
- https://github.com/datasets/continent-codes/blob/main/data/continent-codes.csv
- https://datahub.io/core/continent-codes
I have yet to check if these are an official thing.
(And I have never heart of them before either)

What do you/the others think?

joscha:
> * https://github.com/datasets/continent-codes/blob/main/data/continent-codes.csv
> * https://datahub.io/core/continent-codes
>   I have yet to check if these are an official thing.
>   (And I have never heart of them before either)

I've never heard about continent codes before, but it makes sense there would be some.

Just one note: the two data sources above have continents like the UN interprets them, see [here](https://en.wikipedia.org/wiki/Oceania#:~:text=The%20CIA%20World%20Factbook%20also,also%20used%20for%20statistical%20purposes.). This means that for example they have `Oceania` instead of `Australia` (my [data source](https://en.wikipedia.org/wiki/Continent)).
From Wikipedia:

> [Oceanic islands](https://en.wikipedia.org/wiki/Island#Oceanic_islands) are occasionally grouped with a nearby continent to divide all the world's land into geographical regions. Under this scheme, most of the island countries and territories in the [Pacific Ocean](https://en.wikipedia.org/wiki/Pacific_Ocean) are grouped together with the continent of Australia to form the geographical region [Oceania](https://en.wikipedia.org/wiki/Oceania).[[2]](https://en.wikipedia.org/wiki/Continent#cite_note-2)

I am not hooked on either; happy to change it to whatever fits best with the rest of faker's data, as long as we have a list of continents in the end.

matthewmayer:
I have never heard of the continent codes before (or seen them in use in any software)

matthewmayer:
I have never heard of the continent codes before (or seen them in use in any software)

So I think just the names are fine.

test/modules/location.spec.ts
```diff
@@ -77,6 +77,8 @@ describe('location', () => {
 
     t.it('country');
 
+    t.it('continent');
```
Shinigami92, line 80: **todo:** please also add a "NON_SEEDED_BASED_RUN" case below

src/locales/en/location/continent.ts
```diff
@@ -0,0 +1,9 @@
+export default [
+  'Asia',
+  'Africa',
+  'North America',
+  'South America',
+  'Antarctica',
+  'Europe',
+  'Australia',
```
Shinigami92: **suggestion:** please sort these values alphabetically

Shinigami92 (commented):
because `country` is also just `string[]` right now, I would suggest to do the same with `continent` for now

test/modules/location.spec.ts
ST-DDT: AFAICT country and county don't have them either (yet).

src/locales/en/location/continent.ts
joscha: 34a701f7

test/modules/location.spec.ts
joscha: 4e8634b3

test/modules/location.spec.ts
```diff
@@ -145,6 +147,17 @@ describe('location', () => {
   describe.each(times(NON_SEEDED_BASED_RUN).map(() => faker.seed()))(
     'random seeded tests for seed %i',
     () => {
+      describe('continent()', () => {
+        it('returns random continent', () => {
+          const continent = faker.location.continent();
+
+          expect(continent).toBeTruthy();
+          expect(continent).toMatch(
+            /^(Africa|Antarctica|Asia|Australia|Europe|North America|South America)$/
```
joscha: can widen this to a `^[A-Z][a-z]+(:? [A-Z][a-z]+)?$` regex, however this is not meant to generate a fantasy name but a fixed list. Given that the next supercontinent is expected to form in about 250 million years, as continents are currently drifting apart, this regex should be quite stable and not change anytime soon.

src/definitions/location.ts
```diff
@@ -41,6 +41,11 @@ export type LocationDefinition = LocaleEntry<{
    */
   country: string[];
 
+  /**
+   * The names of all continents.
+   */
+  continent: string[];
```
ST-DDT: Please move the definitions and the method so that they arent between country and countryCode.

test/modules/location.spec.ts
Shinigami92: Alternatively you use the exact same strategy as we did for all such cases like e.g. in the company module ([example](https://github.com/faker-js/faker/blob/d43d68dd11c81d641042cf38b3260592b9a96419/test/modules/company.spec.ts#L67))

So just
```ts
expect(actual).toBeTruthy();
expect(actual).toBeTypeOf('string');
expect(faker.definitions.location.continent).toContain(actual);
```

test/modules/location.spec.ts
Shinigami92: > AFAICT country and county don't have them either (yet).

You are right, somehow we missed these 🤔 strange

test/modules/location.spec.ts
joscha: 0a16dc78

src/definitions/location.ts
joscha: 02d4c5a3

ST-DDT (approved):

ejcheng (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

xDivisionByZerox:
I'll merge this PR since it has sufficent approvals.

Thank you for contribution @joscha. Your change will be included in the upcoming version 9.1.0 of Faker.

joscha:
Thank you all for your reviews and help to get this in.

PR #3168, opened 2024-10-10
docs: support browser console based exploration

### LINKED ISSUE #3129: Add lazy loadable faker to our documentation
Continuation of #2426

- #2426

I sometimes like to test a few possibilities before suggesting them. However, I don't always have access to my main computer to launch my local playground, so I usually use my browser/our website to copy and use the browser usage to the console to run the scripts.
Since the playground embeds the output into the website, it feels more complicated than just using the browser console.
How about adding a console log to our website that explains how to test faker code from anywhere on the website?

````ts
console.warn("If you would like to test faker in the browser console, you can do so using `enableFaker()`");
console.warn("If you would like to test faker in a playground, visit https://new.fakerjs.dev");
async function enableFaker() {
    const imported = await import('https://cdn.jsdelivr.net/npm/@faker-js/faker@'+version+'/+esm');
    Object.assign(window, imported);
    console.log("You can now start using faker:");
    console.log("e.g. `faker.food.description()` or `fakerZH_CN.person.firstName()`");
    console.log("For other languages please refer to https://fakerjs.dev/guide/localization.html#available-locales");
    return imported;
}
````

Lazy loading faker like this would avoid issues caused by the large package size.

I'm not sure, where we would have to place this snippet for vitepress to pick it up, but I would like to hear your feedback first.

### PR DESCRIPTION
**Pull request by ST-DDT**
Fixes: #3129

- #3129

---

This PR allows users of our website to load faker into the browser console using a simple command.

> `await enableFaker()`

You can test this https://deploy-preview-3168.fakerjs.dev/

### CODE REVIEW

**Review by xDivisionByZerox — DISMISSED** [review]
Cool, didn't know top level wait was a thing in the browser console. 🤷‍♀️

**Comment by ST-DDT** [comment]
> Cool, didn't know top level wait was a thing in the browser console. 🤷‍♀️

Well, it works on my machine 😉 !

https://caniuse.com/mdn-javascript_operators_await_top_level

**Comment by matthewmayer** [comment]
Suggest we could add a menu item to the Try It dropdown like this

image

Linking to an explanatory page explaining how to open Developer Tools and try it.

**Comment by ST-DDT** [comment]
> Suggest we could add a menu item to the Try It dropdown like this
> image
> 
> Linking to an explanatory page explaining how to open Developer Tools and try it.

Good idea and thanks for reminding me, I should have read the original issue again.

**Review by Shinigami92 — DISMISSED** [review]

**Comment by matthewmayer** [comment]
Could it print the current version of Faker to the console after you call

`await enableFaker()`

for example
```
You can now start using Faker 9.0.3:
e.g. `faker.food.description()` or `fakerZH_CN.person.firstName()`
For other languages please refer to https://fakerjs.dev/guide/localization.html#available-locales
For a full list of all methods please refer to https://fakerjs.dev/api/
```

This might avoid confusion e.g. if you are using next.fakerjs.dev and then are confused when a 9.1.0 method that is documented there is not yet available.

File: docs/.vitepress/config.ts [hunk 1]
```diff
@@ -149,6 +150,25 @@ const config: UserConfig<DefaultTheme.Config> = {
         href: 'https://fosstodon.org/@faker_js',
       },
     ],
+    [
+      'script',
+      {
+        id: 'browser-console-faker',
+      },
+      `
+console.warn('If you would like to test faker in the browser console, you can do so using \`await enableFaker()\`');
```
ST-DDT: `Faker` or `faker`?

```suggestion
console.warn('If you would like to test Faker in the browser console, you can do so using \`await enableFaker()\`');
```

(here and the other locations)

matthewmayer: Prefer Faker

docs/.vitepress/config.ts
```diff
@@ -149,6 +150,25 @@ const config: UserConfig<DefaultTheme.Config> = {
...
       },
     ],
+    [
+      'script',
+      {
+        id: 'browser-console-faker',
+      },
+      `
+console.warn('If you would like to test Faker in the browser console, you can do so using \`await enableFaker()\`');
+console.warn('If you would like to test Faker in a playground, visit https://new.fakerjs.dev');
```
xDivisionByZerox: 1. Can we group these logs to not have them displayed as separate entries:

```suggestion
console.warn('If you would like to test Faker in the browser console, you can do so using \`await enableFaker()\`.\nIf you would like to test Faker in a playground, visit https://new.fakerjs.dev.');
```

| Before | After |
| ------- | ----- |
|  |  |

---

2. You used `console.warn` here. 

My impression is that you wanted to make the message stand out. The disadvantage of `console.warn` is that the user might have [filtered their console by log level](https://developer.chrome.com/docs/devtools/console/log#level) (the link is for chrome but all major browser support this). Additionally I find the stack trace being show irritating.

I'd suggest using [colored console output](https://developer.mozilla.org/en-US/docs/Web/API/console#styling_console_output) via a simple log instead:

```suggestion
console.log('%cIf you would like to test Faker in the browser console, you can do so using \`await enableFaker()\`', 'background: rgba(16, 183, 127, 0.14); color: rgba(255, 255, 245, 0.86); padding: 0.5rem;');
console.log('%cIf you would like to test Faker in a playground, visit https://new.fakerjs.dev', 'background: rgba(16, 183, 127, 0.14); color: rgba(255, 255, 245, 0.86); padding: 0.5rem; display: inline-block;');
```

| Before | After |
| ------- | ----- |
|  |  |

I've used the colors that are set for out "Note" blocks and applied a little bit of padding that scales based on the console font-size to take the users zoom in the console into account. The `display: inline-block` is for additional support in firefox without changing the behaviour in chromium browsers.

---

Suggestions together:

```suggestion
console.log('%cIf you would like to test Faker in the browser console, you can do so using \`await enableFaker()\`.\nIf you would like to test Faker in a playground, visit https://new.fakerjs.dev.', 'background: rgba(16, 183, 127, 0.14); color: rgba(255, 255, 245, 0.86); padding: 0.5rem; display: inline-block;');
```

| Before | After - Chromium | After - Firefox | 
| ------- | ------------------- | --------------- |
|  |  |  |

docs/.vitepress/config.ts
```diff
@@ -149,6 +150,25 @@ const config: UserConfig<DefaultTheme.Config> = {
...
+async function enableFaker() {
+  const imported = await import('https://cdn.jsdelivr.net/npm/@faker-js/faker@${version}/+esm');
+  Object.assign(globalThis, imported);
+  console.log('You can now start using Faker v${version}:');
+  console.log('e.g. \`faker.food.description()\` or \`fakerZH_CN.person.firstName()\`');
+  console.log('For other languages please refer to https://fakerjs.dev/guide/localization.html#available-locales');
+  console.log('For a full list of all methods please refer to https://fakerjs.dev/api/');
```
xDivisionByZerox: Can we apply the same changes here as I stated [in my previous comment](https://github.com/faker-js/faker/pull/3168#discussion_r1797691240) (group and style)?

docs/.vitepress/config.ts
matthewmayer: Today I learned you can style console.log with CSS!

https://www.bennadel.com/blog/3941-styling-console-log-output-formatting-with-css.htm

ST-DDT: Not sure about this, but done.

docs/.vitepress/config.ts
ST-DDT: Done

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3170, opened 2024-10-11
refactor(location): allow postcode_by_state to take an array of patterns

matthewmayer:
ref #3169

This change will make it easier to handle ZIP codes where there are "holes" in the ranges.

For example Hawaii could use two ranges like this to exclude 96799 (American Samoa)
```
HI: [
    '{{number.int({"min": 96701,"max": 96798})}}',
    '{{number.int({"min": 96801,"max": 96898})}}',
  ],
```

Since `faker.helpers.fake` accepts an array, just need to change the type. Will change the actual definitions in a followup PR.


ST-DDT (approved):

xDivisionByZerox (approved):
Sounds reasonable. Would we change the definition property type to only `string[]` in v10?

ST-DDT:
I'd prefer using actual code instead of string patterns for v10.

- #3176

Shinigami92 (approved):

PR #3172, opened 2024-10-11
infra: show eslint progress

Shinigami92:
why? cause it's cool  
no but seriously, with that locally you can see how long you potentially need to wait anymore

docs: https://www.npmjs.com/package/eslint-plugin-file-progress | https://github.com/sibiraj-s/eslint-plugin-file-progress


ST-DDT:
It slows down a full run of eslint by 3seconds.

79 seconds -> 82 seconds, but it feels more responsive because actually see it doing something.

eslint.config.ts
```diff
@@ -198,7 +199,16 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
...
   eslintPluginPrettierRecommended,
-  //#endregion,
+  //#endregion
+
+  {
+    plugins: {
+      'file-progress': eslintPluginFileProgress,
+    },
+    rules: {
+      'file-progress/activate': 1,
```
ST-DDT: Please use 'error' like all the other entries in this file do.

ST-DDT (dismissed):
An optional but nice little gimmick.

eslint.config.ts
Shinigami92: I used the valid 1 syntax here because this is the document way of doing it and also it means "turned on"
Sadly true is not a valid option
Only 0,1,2 and off,warn,error
I also don't really like the eslint way of props, but at least this does make slightly more sense to me that active:error 😕

ST-DDT (dismissed):

ST-DDT (dismissed):

ST-DDT (dismissed):

eslint.config.ts
ST-DDT: I added an issue to ask for a recommended config preset to use instead.

- https://github.com/sibiraj-s/eslint-plugin-file-progress/issues/8

matthewmayer (dismissed):

ST-DDT:
@matthewmayer Sorry, they just released an update that fixes Node 18 support and ships it's own types.

ST-DDT:
There seems to be an issue with the new versions.

sibiraj-s:
> There seems to be an issue with the new versions.

Can you try v2.1.1, I was manually typing the types, not generated. I made some mistakes.

ST-DDT:
Fixed

ST-DDT (dismissed):

eslint.config.ts
```diff
@@ -200,7 +201,12 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
 
   //#region prettier
   eslintPluginPrettierRecommended,
-  //#endregion,
+  //#endregion
+
+  //#region file-progress
+  // https://www.npmjs.com/package/eslint-plugin-file-progress
```
Shinigami92: **nitpick (non-blocking):** IMO this line can be removed, as we do not add such a line for any other plugin

ST-DDT: I considered adding it for the other plugins as well to simplify config lookups.

Shinigami92: feels over-engineered to me 🤷 
But as it is mostly your codebase in the last month(s), I will leave that decision to you
image

ST-DDT:
```suggestion
```

ST-DDT: I found a better way to achieve that:

````sh
pnpm eslint --inspect-config --flag unstable_ts_config
````

ST-DDT (approved):

Shinigami92:
Can't approve, because historically I opened the PR
But count me as approved ✅

PR #3188, opened 2024-10-16
infra: remove preflight failure comment

ST-DDT:
This PR removes the preflight failure comment.

The github CI failure in combination with the error annotations are sufficient to guide the users.

- https://github.com/faker-js/faker/blob/ee45594a88e85f9dc888eac08c7ebd5979a765d1/.github/workflows/pr.yml#L50


.github/workflows/commentCodeGeneration.ts
Shinigami92, line 1: **question:** could you please explain why it is required to have the TypeScript checked file inlined in the yml file? IMO we are loosing some benefits by doing this.

.github/workflows/commentCodeGeneration.ts
```diff
@@ -1,73 +0,0 @@
...
-    state: 'open',
-    head: head_branch,
-  });
-
-  if (prs.length === 0) {
-    console.log(`No PRs found for branch ${head_branch}`);
-    return;
-  }
-
-  const pr_number = prs[0].number;
```
xDivisionByZerox, line 31: I'm pretty sure this is the actual problem with the script integration, no? Instead of getting the PR by arbitrarily fetching the API and choosing the first element, it should be retrieved from the context the pipeline is run in. :thinking:

.github/workflows/commentCodeGeneration.ts
ST-DDT: Using a file that is checked in is a security risk when using `pull_request_target`.
As it can be modified by an attacking party.

.github/workflows/commentCodeGeneration.ts
ST-DDT: Any idea how to do that?

AFAICT we cannot directly call another workflow with parameters because it would still run with (insufficient) third-party permissions and we cannot access the outputs of the other workflow if there is no dependency between them. And for that they need to be in the same file.

xDivisionByZerox: I was mistaken, thinking that the inclined script was a copy-paste from this one. In the inlined script, you correctly use `context.issue.number` instead of hitting the REST api.
My comment is obsolete with that conclusion.

.github/workflows/commentCodeGeneration.ts
Shinigami92: And that's not possible with the yml file?

ST-DDT: AFAIK even if you replace the file in your PR, it still takes the one from the default branch.

Shinigami92 (dismissed):

ST-DDT:
**Team Decision**

- We will remove the comment, the error annotations are sufficient

ST-DDT:
Ready for review

xDivisionByZerox (dismissed):

xDivisionByZerox (dismissed):

PR #3189, opened 2024-10-16
feat: de-proxy locale access

### LINKED ISSUE #3155: Support older browsers without Proxy
We're using Faker in browser, and need to support older browsers that do not have [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) (including all IE versions). As Faker relies on that for the [LocaleProxy](https://github.com/faker-js/faker/blob/9537dfddba882bd93d9a429697fd44bc72428426/src/internal/locale-proxy.ts), it doesn't work in those browsers.

Is there any scope for Faker to be able to support those browsers? I don't have any suggestions for what that would be, but maybe some sort of browser feature detection or disabling the locale functionality.

Google's [proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) doesn't cover the `has` or `deleteProperty` handlers, so doesn't work for Faker's use case. (Unless an option is to stop using those handlers?)

ST-DDT:
Second part of the standalone module function feature #2667 

- #2667

Fixes #3155

- #3155

Requires #2838

- #2838

## Overview

1. Introduce FakerCore #2838
2. De-proxy locale access #3189
3. De-class config function #3748 (seed, setDefaultRefDate, ...)
4. De-module functions #3748 (airline, ...)

## Description

Instead of calling `this.faker.definitions.category.entry`, we will now use `resolveLocaleData(this.faker.fakerCore, 'category', 'entry')`.
Please note that this will get simplified with the next PR as `this.faker.fakerCore` will get replaced with just `fakerCore`.
This change makes the functions independent from the main faker instance, at least locale access wise.
That makes `resolveLocaleData` the first effective "standalone module function", as it no longer needs faker, only the fakerCore is needed.

As an additional benefit it also fixes #3155

- #3155 (as long as the full faker instance isn't created/the faker.definitions isn't accessed (when using lazy loading))

## Recreating the PR

(Partially outdated)

- Cherry-pick the first two commits.
- Replace `this\.faker\.definitions\.(\w+)\.(\w+)` with `assertLocaleData(this.faker.fakerCore.locale.$1?.$2, '$1.$2')` in `src`.

<details>
<summary><h2>Next Steps (Future PR)</h2></summary>

The next step after this PR is converting the other methods to standalone module functions.

### Current

````ts
export class AirlineModule extends ModuleBase {
  airport(): Airport {
    return this.faker.helpers.arrayElement(
      assertLocaleData(this.faker.fakerCore.locale.airline.airport, 'airline', 'airport')
    );
  }
}
````

### Future

````ts
function airport(fakerCore: FakerCore): Airport {
  return arrayElement(fakerCore, 
    assertLocaleData(fakerCore.locale.airline.airport, 'airline', 'airport')
  );
}

export class AirlineModule extends ModuleBase {
  airport(): Airport {
    // return airport(this.faker.fakerCore);
    return airport(this.fakerCore);
  }
}
````

</details>

### CODE REVIEW

**Comment by ST-DDT** [comment]
Should I push this PR forward (as a pre-step of #3748) or should I drop it in favor of #3748 (or its parts)?
(Affects all modules that have locale files, could likely be split into one PR per module as well)

TLDR: This PR intends to do:

- Replace `this.faker\.definitions\.(\w+)\.(\w+)` and similar \
  with `assertLocaleData(this.faker.fakerCore.locale.$1.$2, '$1.$2')`

#3748 does the same just with `this.faker.fakerCore` -> `fakerCore` (during the move to separate files).

**Comment by Shinigami92** [comment]
> Should I push this PR forward (as a pre-step of #3748) or should I drop it in favor of #3748 (or its parts)? [...]


- #3748 touches 359 files right now, that are at least 259 files to much for me to make a proper review
so everything that helps to chunk the PRs down to <100 touched files per PR is a win for reviewers

**Comment by matthewmayer** [comment]
Agreed, if there are ways to chunk the more mechanical changes into seperate PRs like this that make the final PR smaller, and it doesn't break the build, then go for it.

**Comment by ST-DDT** [comment]
I consider splitting this PR into 3 smaller PRs.

- feat: introduce resolveLocaleData
- refactor: replace Proxy locale access with resolveLocaleData
- ?: deprecate LocaleProxy

This PR is theoretically ready for review, but I would like to fix/simplify some code outside of this PR to reduce the diff (complexity).

**Comment by ST-DDT** [comment]
**Decision Needed**

Which of the three should we do?

1) `const data = resolveLocaleData(core, 'database', 'collation');`
   Autocomplete works, but unable to jump directly to the definitions.
2) `const data = assertLocaleData(core.locale.database?.coallation, 'database.coallation');`
   Basically a fancy `require` check. More verbose but also more flexible. Currently internal.
3) Add both variants, but internally use only one of them.

File: src/faker.ts [hunk 1]
```diff
@@ -82,6 +88,11 @@ export class Faker extends SimpleFaker {
...
   readonly word: WordModule = new WordModule(this);
 
+  /**
+   * The raw definitions for the current locale.
+   *
+   * @deprecated Use `faker.fakerCore.locale` instead.
+   */
   get rawDefinitions(): LocaleDefinition {
     return this.fakerCore.locale;
   }
```
xDivisionByZerox: The following goes for faker.rawDefinitions, as well as faker.definitions.
 
1. When following the deprecation workflow, accessors must also print a warning during runtime. This behavior is missing right now.
1. Do we need to deprecate these accessors? While the calling of "resolveLocale" is a great tool internally, I see it as downgrade for the users DX, since they are now required to import an additional function and invoke it.

src/registry.ts
```diff
@@ -0,0 +1,19 @@
+import type { FakerCore } from '.';
+import { utilsModule as utils } from './utils/registry';
+
+// TODO @ST-DDT 2026-04-12: This file will be auto generated in a future PR.
+
+/**
+ * Global Registry for the Faker library, containing all (SMF-aware) module registries.
```
xDivisionByZerox: We might want to be explicit what "SMF" is. We are working in the faker context and are well aware of our terminology - users are not and might get confused.

src/definitions/definitions.ts
```diff
@@ -29,29 +29,49 @@ export type LocaleEntry<TCategoryDefinition extends Record<string, unknown>> = {
...
+  word: WordDefinition;
+};
+
+/**
+ * Helper type to undo `LocaleEntry<T>` wrapping.
+ */
+type DefinedLocaleEntry<T> = T extends LocaleEntry<infer U> ? U : never;
+
+/**
+ * Helper type containing the well known locale definitions as used by this library, assuming all values are present.
```
xDivisionByZerox:
```suggestion
 * Helper type containing the well known locale definitions as used by the Faker library, assuming all values are present.
```

xDivisionByZerox (commented):
I know that this PR is still in draft mode - still wanted to mention some things I found while getting a general overview of this PR.

---

> **Decision Needed**

What exactly is required to be decided in your request? I look to me that you are giving and are not asking a question. :thinking:

ST-DDT:
> > **Decision Needed**
> 
> What exactly is required to be decided in your request? I look to me that you are giving and are not asking a question. 🤔

https://github.com/faker-js/faker/pull/3189#issuecomment-4231810491 Edited

src/registry.ts
ST-DDT:
```suggestion
 * Global Registry for the Faker library, containing all module registries.
```

src/faker.ts
ST-DDT: > 1. When following the deprecation workflow, accessors must also print a warning during runtime. This behavior is missing right now.

I'll add it, if we actually go for the deprecation.

> 2. Do we need to deprecate these accessors? While the calling of "resolveLocale" is a great tool internally, I see it as downgrade for the users DX, since they are now required to import an additional function and invoke it.

This depends.

`faker.definitions` currently uses the `LocaleProxy`. 

- If we deprecate the proxy, we need to deprecate the field well.
  We could redefine `faker.definitions === faker.rawDefinitions` in v11 and only deprecate rawDefinitions for removal now.
- If we keep all three of them, then we have to make sure, that we actually solve #3155 (for our code only).

src/registry.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Global Registry for the Faker library, containing all module registries.
+ */
+type FakerRegistry = Record<
+  string,
+  Record<string, (fakerCore: FakerCore, ...args: unknown[]) => unknown>
+>;
+
+/**
+ * Global Registry for the Faker library, containing all (SMF-aware) module registries.
+ */
```
Shinigami92: **suggestion:** 

```suggestion
/**
 * Global Registry for the Faker library, containing all module registries
 * of standalone module functions (see #2667).
 */
```

src/utils/registry.ts
```diff
@@ -0,0 +1,9 @@
+import { resolveLocaleData } from './resolve-locale-data';
+
+// TODO @ST-DDT 2026-04-12: Add future SMF utils
+/**
+ * Registry Module containing (SMF-aware) utility functions for the Faker library.
+ */
```
Shinigami92: **suggestion:** 

```suggestion
/**
 * Registry Module containing standalone module function (see #2667)
 * utility functions for the Faker library.
 */
```

Shinigami92 (commented):
While reading `src/modules/location/index.ts`, I noticed that `direction()`, `cardinalDirection()`, and `ordinalDirection()` all call `resolveLocaleData(this.faker.fakerCore, 'location', 'direction')` multiple times per invocation (up to four in `direction()`).

Since `resolveLocaleData` is a pure read from `core.locale` with no randomizer involvement, caching the result in a local is snapshot-safe — I verified locally that the full `test/modules/location.spec.ts` suite (2407 tests, including seeded snapshots) passes unchanged. It also matches the style already in place in `date.month`, `date.weekday`, `system.fileType`, and `system.fileExt`, where the lookup is hoisted into a local.

Suggested shape for the three methods:

```ts
direction(options: { abbreviated?: boolean } = {}): string {
  const { abbreviated = false } = options;
  const direction = resolveLocaleData(
    this.faker.fakerCore,
    'location',
    'direction'
  );

  if (!abbreviated) {
    return this.faker.helpers.arrayElement([
      ...direction.cardinal,
      ...direction.ordinal,
    ]);
  }

  return this.faker.helpers.arrayElement([
    ...direction.cardinal_abbr,
    ...direction.ordinal_abbr,
  ]);
}

cardinalDirection(options: { abbreviated?: boolean } = {}): string {
  const { abbreviated = false } = options;
  const direction = resolveLocaleData(
    this.faker.fakerCore,
    'location',
    'direction'
  );

  if (!abbreviated) {
    return this.faker.helpers.arrayElement(direction.cardinal);
  }

  return this.faker.helpers.arrayElement(direction.cardinal_abbr);
}

ordinalDirection(options: { abbreviated?: boolean } = {}): string {
  const { abbreviated = false } = options;
  const direction = resolveLocaleData(
    this.faker.fakerCore,
    'location',
    'direction'
  );

  if (!abbreviated) {
    return this.faker.helpers.arrayElement(direction.ordinal);
  }

  return this.faker.helpers.arrayElement(direction.ordinal_abbr);
}
```

---

+1 to xDivisionByZerox's point on spelling out "SMF" — worth noting the same acronym also appears in `src/utils/registry.ts:5` ("Registry Module containing (SMF-aware) utility functions…"), so the fix needs to land in both spots.

These doc comments ship in the published typedocs, so readers outside the maintainer group will hit them without context for "SMF".

src/registry.ts
ST-DDT:
```suggestion
/**
 * Global registry for the Faker library, containing all module registries with their standalone module functions.
 */
```

src/utils/registry.ts
ST-DDT:
```suggestion
/**
 * Registry rodule containing all standalone utility functions for the Faker library.
 */
```

ST-DDT:
> While reading `src/modules/location/index.ts`, I noticed that `direction()`, `cardinalDirection()`, and `ordinalDirection()` all call `resolveLocaleData(this.faker.fakerCore, 'location', 'direction')` multiple times per invocation (up to four in `direction()`).

Split into a separate PR:

- https://github.com/faker-js/faker/pull/3809

ST-DDT:
> +1 to xDivisionByZerox's point on spelling out "SMF" — worth noting the same acronym also appears in `src/utils/registry.ts:5` ("Registry Module containing (SMF-aware) utility functions…"), so the fix needs to land in both spots.
> 
> These doc comments ship in the published typedocs, so readers outside the maintainer group will hit them without context for "SMF".

Changed.

Is "standalone module functions" a term that outsiders do understand or should we choose a different one?

src/utils/registry.ts
```diff
@@ -2,7 +2,7 @@ import { resolveLocaleData } from './resolve-locale-data';
 
 // TODO @ST-DDT 2026-04-12: Add future SMF utils
 /**
- * Registry Module containing (SMF-aware) utility functions for the Faker library.
+ * Registry rodule containing all standalone utility functions for the Faker library.
```
xDivisionByZerox: You had a typo in you inline suggestion.

```suggestion
 * Registry module containing all standalone utility functions for the Faker library.
```

xDivisionByZerox:
> **Decision Needed**
>
> Which of the three should we do?
>
> 1) `const data = resolveLocaleData(core, 'database', 'collation');`
>   Autocomplete works, but unable to jump directly to the definitions.
> 2) `const data = assertLocaleData(core.locale.database?.coallation, 'database.coallation');`
>   Basically a fancy `require` check. More verbose but also more flexible. Currently internal.
> 3) Add both variants, but internally use only one of them.

**Team Decision**

We go with option 2 for now. The internal API usage can be verbose, we can use tooling to make our lives easier if required.
The public API for users will be designed in a separate manner.

src/faker.ts
xDivisionByZerox: **Team Decision**

Deprecations will not be done in this PR.

src/modules/airline/index.ts
```diff
@@ -88,7 +89,11 @@ export class AirlineModule extends ModuleBase {
    */
   airport(): Airport {
     return this.faker.helpers.arrayElement(
-      this.faker.definitions.airline.airport
+      assertLocaleData(
+        this.faker.fakerCore.locale.airline?.airport,
+        'airline',
+        'airport'
+      )
     );
```
ST-DDT, line 96: Which variant do you prefer?

````ts
    return this.faker.helpers.arrayElement(
      assertLocaleData(
        this.faker.fakerCore.locale.airline?.airport,
        'airline',
        'airport'
      )
    );
````

or

````ts
    return this.faker.helpers.arrayElement(
      assertLocaleData(
        this.faker.fakerCore.locale.airline?.airport,
        'airline.airport'
      )
    );
````

xDivisionByZerox: There is no reason to spread the key for the error over multiple arguments. Providing the entire key chain as one string should be sufficent.

ST-DDT:
Ready for review.

xDivisionByZerox (dismissed):
Looks good :+1:

ST-DDT:
I just thought about the following:
If we don't support browsers without Proxy anyway, then we could just use the LocaleProxy in the FakerCore.

- Then we wouldn't need this PR at all.
- We would continue to have the short redundancy free calls `fakerCore.locale.airline.airport` that will still throw.

Shinigami92:
Proxy seems fine, go wild

image

ST-DDT:
I created an alternative to this PR:

- https://github.com/faker-js/faker/pull/3820#issuecomment-4276438874

ST-DDT:
We will continue to use our `LocaleProxy` instead of this PR.

See also:

- https://github.com/faker-js/faker/issues/3155#issuecomment-4283721228

PR #3190, opened 2024-10-17
docs: optimize line breaks in default objects

ST-DDT:
Fixes the display of objects in default values.

**Before**

grafik

**After**

grafik

---

 See also:

- https://github.com/faker-js/faker/pull/2936/files#r1649768213


Shinigami92 (approved):
😸

scripts/apidocs/processing/jsdocs.ts
```diff
@@ -50,7 +50,11 @@ export function getParameterTags(jsdocs: JSDoc): Record<string, JSDocTag> {
 }
 
 export function getDefault(jsdocs: JSDoc): string | undefined {
-  return getOptionalTagFromJSDoc(jsdocs, `default`);
+  return (
+    getOptionalTagFromJSDoc(jsdocs, `default`)
+      // Prevent line breaks between the key and the value { foo: 'bar' }
+      ?.replaceAll(': ', ': ')
```
xDivisionByZerox, line 56: I'm not sure about this. Using any other representation of these spaces would make much more sense for me for visiblitiy purposes. The unschooled eye might think that this is a noop when instead it replaces spaces with [non-breaking-spaces](https://www.ascii-code.com/character/nbsp), 🤔

ST-DDT: That's why I added the comment.
It isn't highlighted on GitHub, but VSCode highlights it.

{EA7B297A-1D15-4662-A037-6CAD337FE8AF}

PR #3191, opened 2024-10-17
infra(eslint): limit usage of globals in src

### LINKED ISSUE #3145: Prevent usage of features not broadly available
Related to #3143

- #3143

We should adjust our build or lint tooling to report bad usage of globals/features that aren't compatible with both modern browsers and node JS versions.

See here for an initial attempt:

- https://github.com/faker-js/faker/pull/3144#discussion_r1775884619

ST-DDT:
Fixes #3145

- #3145

Ban the usage of globals in our `src` folder.
All usages of these globals have to be marked with a eslint disable rule making them easier to spot.


eslint.config.ts
```diff
@@ -200,6 +200,17 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
...
 
   //#region overrides
+  {
+    // Must override the config from typescript-eslint
+    files: ['src/**/*.ts'],
+    rules: {
+      'no-undef': 'error',
+    },
+    languageOptions: {
+      // Don't allow any globals in our TypeScript files
```
ST-DDT:
```suggestion
      // Don't allow any globals in our TypeScript files - unless explicitly ignored
```

eslint.config.ts
```diff
@@ -200,6 +200,17 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
...
+      'no-undef': 'error',
+    },
+    languageOptions: {
+      // Don't allow any globals in our TypeScript files - unless explicitly ignored
+      globals: {},
+    },
+  },
   {
     files: ['src/**/*.ts'],
     rules: {
```
Shinigami92: **question:** why not merging these both together? they addressing the same file glob

ST-DDT: Fixed. (The config originally looked different and I only simplified it without checking for existing groups)

Shinigami92 (approved):

xDivisionByZerox:
What happened to the original implementation where globals that are not specific to an environment were allowed?

ST-DDT:
> What happened to the original implementation where globals that are not specific to an environment were allowed?

It was never added to our source and this solution was recommended to me by members of eslint.

ST-DDT:
Also AFAIK we don't actually want to use any additional globals anyway.

xDivisionByZerox (approved):

PR #3192, opened 2024-10-18
docs(guide): remove esModuleInterop flag from config

ST-DDT:
AFAICT our users don't need the `esModuleInterop` (at least for our code).
So we should remove them from our guide and I also removed them from our [playground](https://github.com/faker-js/playground/commit/03d7893769480105f2abc4da5d5429770e5d8a89).

Could you please verify that there aren't any cases that I missed where our users need it?


xDivisionByZerox (approved):
Due to us using named exports I don't see who this flag would be required for our users.

xDivisionByZerox:
For reference: https://www.typescriptlang.org/tsconfig/#esModuleInterop

PR #3194, opened 2024-10-18
infra: reduce log spam in generate-locales

ST-DDT:
IIRC we talked (at some point) about reducing the log spam from the generate:locales script.

So this script removes the `Running data normalization for: file` logs.


xDivisionByZerox (approved):

Shinigami92:
I'm not so sure about this one
With and without it takes both 9 seconds on my machine (`pnpm run generate:locales`)

But these logs show at least somehow the actual progress as well as prettier and the other PR for file-progress in eslint does.
However what I feel much more as log-spam is `pnpm run generate:api-docs` 🤔 

What we could do to `generate:locales` is potentially to reduce colors like the prettier CI does if there was no actual change to the file cause of cache (or something like that)

PR #3196, opened 2024-10-19
feat(animal): add petName method

### LINKED ISSUE #2206: Pet names

There aren't pet names

### Suggested solution

Add pet names

### Alternative

_No response_

### Additional context

_No response_

suyashgulati:
To resolve #2206

Adds new method `petName` on animal module


suyashgulati:
Need maintainers' help on resolving the test failure:

```
Error: The locale data for 'animal.petName' are missing in this locale.
  Please contribute the missing data to the project or use a locale/Faker instance that has these data.
  For more information see https://fakerjs.dev/guide/localization.html
```
  
I am not sure what needs to be done for this. Do I need to add any other locale than en to resolve this?

src/modules/animal/index.ts
```diff
@@ -72,6 +72,20 @@ export class AnimalModule extends ModuleBase {
     return this.faker.helpers.arrayElement(this.faker.definitions.animal.lion);
   }
 
+  /**
+   * Returns a random pet name.
+   *
+   * @example
+   * faker.animal.petName() // 'Coco'
+   *
+   * @since 5.5.0
```
matthewmayer: Should be `@since 9.2.0` as that will be next minor release

test/modules/animal.spec.ts
```diff
@@ -17,6 +17,7 @@ const functionNames = [
   'horse',
   'insect',
   'lion',
+  'petName',
```
matthewmayer: the problem with the test failure is that these tests assume that the definition name exactly matches the method name. But in this case it (correctly) doesn't - petName vs pet_name. I would remove petName from the functionNames array and write a seperate test for it.

matthewmayer (changes requested):

test/modules/animal.spec.ts
ST-DDT: I'll have a look at it.

ST-DDT:
Blocked by: #3197

- #3197

test/modules/animal.spec.ts
ST-DDT: -> #3197

ST-DDT:
This is no longer blocked, but you have to adjust the test file.

test/modules/animal.spec.ts
```diff
@@ -133,6 +134,13 @@ describe('animal', () => {
           expect(faker.definitions.animal.type).toContain(actual);
         });
       });
+
+      describe('petName()', () => {
+        it('should return random value from type array', () => {
```
matthewmayer: From pet name array not type array?

src/locales/en/animal/pet_name.ts
```diff
@@ -0,0 +1,45 @@
+export default [
+  'Ace',
+  'Archie',
+  'Bailey',
+  'Bandit',
+  'Bear',
```
matthewmayer: Maybe omit this just because it's confusing if this is the value (You might think you have the wrong method?)

matthewmayer (changes requested):

test/modules/animal.spec.ts
ST-DDT:
```suggestion
        it('should return random value from pet name array', () => {
```

src/locales/en/animal/pet_name.ts
ST-DDT:
```suggestion
```

ST-DDT (approved):

matthewmayer (approved):

ejcheng (approved):

ST-DDT:
Your additions will be available in Faker v9.2.0.
While I don't have an exact release date for it yet, we will likely release it somewhat soonish (a week - a few weeks).

PR #3197, opened 2024-10-19
test: properly convert function name to locale entry name

ST-DDT:
Currently, the animal tests use the function name as the name of the entry.
This however doesn't work when the name uses camelCase such as `petName`.

This PR changes the locale data lookup to convert the camelCase function name to a snake_case locale entry name before using it.

Required for

- #3196 

### Alternative

- #3198


matthewmayer:
Would it be better just to explicitly test each method in turn? Its repetitive but maybe easier to understand. Most other simple modules, even ones which are just helpers.arrayElement wrappers like `faker.book` just test every method one by one.

matthewmayer:
> Would it be better just to explicitly test each method in turn? Its repetitive but maybe easier to understand. Most other simple modules, even ones which are just helpers.arrayElement wrappers like `faker.book` just test every method one by one.

for comparison i made  #3198 - i find this easier to read and more consistent with other modules.

ST-DDT:
> for comparison i made #3198 - i find this easier to read and more consistent with other modules.

Thanks for creating the PR for comparison.
IMO the current loop based version is easier/faster to fully understand, but I don't have a strong opinion either way.

PR #3198, opened 2024-10-19
test(animal): unroll animal test loop

matthewmayer:
Unroll the test loop for animal module

### Alternative

- #3197


test/modules/animal.spec.ts
```diff
@@ -5,36 +5,130 @@ import { times } from './../support/times';
...
+    t.it('crocodilia');
+    t.it('dog');
+    t.it('fish');
+    t.it('horse');
+    t.it('insect');
+    t.it('lion');
+    t.it('rabbit');
+    t.it('rodent');
+    t.it('snake');
+    t.it('type');
```
ST-DDT: Any reason not to use `t.itEach` here?

ST-DDT (dismissed):

suyashgulati (approved):
This way makes the tests more verbose.

matthewmayer:
It's deliberately more verbose. I feel that test code shouldn't be too "clever". 

I would argue if the tests were in this format it would have been easier to see what had gone wrong when copy pasting the new test for petName. The compiler / autocomplete would guide you that you needed pet_name not petName for the definition.

xDivisionByZerox (approved):
I have to agree with @matthewmayer here. Test code is not supposed to be clever. Thinks like the DRY principle should not be applied to make the intend of the test more clear. 

In the worst case test case isolation might even break by "clever" helper functions, since these functions are often not tested themself. While this is not the case here, I like to  bring awareness of the posibilities for these situations.

ST-DDT (approved):
It is a for loop. I wouldn't call it clever by any means.
A for loop enforces a basic level of consistency across all methods, that a copy paste approach does not.

Ultimately, it doesn't really matter to me.
So I'm fine either way.

Shinigami92 (approved):

PR #3201, opened 2024-10-20
refactor(finance): deprecate maskedNumber for removal

### LINKED ISSUE #2504: Deprecate finance.maskedNumber for removal
We want to deprecate the [`finance.maskedNumber()`](https://fakerjs.dev/api/finance.html#maskednumber) method for removal.

### Reasons for removal

- The method has/had a bug for 6 months and it hasn't been reported at all -> low/no actual usage.
- Configuring the method is more time and code consuming than just adding the parenthesis and ellipsis manually instead.
- There is no real standard for anonymizing these data. Developers will likely have their own anonymizing function that they use for their actual data.
- The way the data are anonymized is very specific, in most cases the data are replaced by `*` instead of being outright removed.
- There are way more data that you might want to mask such as e-mails, ibans, credit card numbers, that cannot be generated by the method.

### Suggested workaround

````ts
const data = faker.finance.iban(); // "DE31070946979005090431"
const maskedData = data.replace(/(?<=.{4})\w(?=.{2})/g, '*'); // "DE31****************31"
````

### See also

- https://github.com/faker-js/faker/pull/2494

### PR DESCRIPTION
**Pull request by ST-DDT**
Fixes #2504

- #2504

---

The masking of important information is usually implemented in the application itself, so there is no need for us to provide a method ourselves.
This is especially true, since the method has lots of options and would need even more to actually represent real masked data.

As a workaround, you can use the following:

- `faker.finance.creditCardNumber().replaceAll(/(?<=.{4}).(?=.{2})/g, '*')`
- `faker.finance.iban().replaceAll(/(?<=.{4})\w(?=.{2})/g, '*')`

### Popularity-Data from #3215

- `faker.finance.maskedNumber`: 59

### CODE REVIEW

File: src/modules/finance/index.ts [hunk 1]
```diff
@@ -961,9 +978,9 @@ export class FinanceModule extends ModuleBase {
     const company = this.faker.company.name();
     const transactionType = this.transactionType();
     const account = this.accountNumber();
-    const card = this.maskedNumber();
+    const card = this.creditCardNumber().replaceAll(/(?<=.{4}).(?=.{2})/g, '*');
```
xDivisionByZerox: This will not produce the same output as you can see in the snapshot tests.

xDivisionByZerox:
Dispite [my comment in #3202](https://github.com/faker-js/faker/pull/3202#issuecomment-2452995670), I think deprecating this method is valid:

1. **Proofen** low to none actual use
2. The utility this method provides is arguably confusing and niche

xDivisionByZerox (dismissed):

test/modules/__snapshots__/finance.spec.ts.snap
```diff
@@ -260,6 +260,6 @@ exports[`finance > 1337 > pin > with length option 1`] = `"2124352971"`;
 
 exports[`finance > 1337 > routingNumber 1`] = `"212435298"`;
 
-exports[`finance > 1337 > transactionDescription 1`] = `"invoice transaction at Gottlieb - Koelpin using card ending with ***(...9477) for HUF 262.02 in account ***52971361"`;
+exports[`finance > 1337 > transactionDescription 1`] = `"invoice transaction at Gottlieb - Koelpin using card ending with 4773*******13 for JMD 262.02 in account ***52971361"`;
```
matthewmayer: the phrasing of transactionDescription "ending with" doesn't really make sense if both the start and end of the masked number are shown.

matthewmayer (changes requested):

test/modules/__snapshots__/finance.spec.ts.snap
matthewmayer: So I suggest either we mask all but the last digits like before, or delete "ending with".

src/modules/finance/index.ts
```diff
@@ -322,6 +331,14 @@ export class FinanceModule extends ModuleBase {
           ellipsis?: boolean;
         } = {}
   ): string {
+    deprecated({
+      deprecated: 'faker.finance.maskedNumber()',
+      proposed:
+        "faker.finance.iban().replace(/(?<=.{4})\\w(?=.{2})/g, '*') or a similar approach",
+      since: '9.1.0',
```
```suggestion
      since: '9.3.0',
```

test/modules/__snapshots__/finance.spec.ts.snap
ST-DDT: Should be fixed now.

src/modules/finance/index.ts
ST-DDT: I adjusted the example in the jsdocs accordingly.

matthewmayer (approved):

PR #3202, opened 2024-10-20
feat(finance): use fake patterns for transactionDescription

ST-DDT:
This PR changes the `faker.finance.transactionDescription()` method to use fake patterns instead of a hard coded string.

### Popularity-Data from #3215

- `faker.finance.transactionDescription`: 20


ST-DDT:
If you still need the method, please inline/copy the implementation:

````ts
transactionDescription(): string {
    const amount = faker.finance.amount();
    const company = faker.company.name();
    const transactionType = faker.finance.transactionType();
    const account = faker.finance.accountNumber();
    const card = faker.finance.creditCardNumber().replaceAll(/(?<=.{4}).(?=.{2})/g, '*');
    const currency = faker.finance.currencyCode();

    return `${transactionType} transaction at ${company} using card ending with ***${card} for ${currency} ${amount} in account ***${account}`;
  }
````

**Review by Shinigami92 — DISMISSED** [review]

File: src/modules/finance/index.ts [hunk 1]
```diff
@@ -955,8 +956,17 @@ export class FinanceModule extends ModuleBase {
...
    *
    * @since 5.1.0
+   *
+   * @deprecated Refer to [PR #3202](https://github.com/faker-js/faker/pull/3202#issuecomment-2425147798) for alternatives ways to generate a transaction descriptions.
    */
   transactionDescription(): string {
+    deprecated({
+      deprecated: 'faker.finance.transactionDescription()',
+      proposed: 'an alternative from PR #3202',
+      since: '9.1.0',
```
```suggestion
      since: '9.2.0',
```

xDivisionByZerox:
> ### Popularity-Data from #3215
>
> - `faker.finance.transactionDescription`: 20

[A quick GH search](https://github.com/search?q=faker.finance.transactionDescription+language%3ATypeScript+OR+faker.finance.transactionDescription+language%3AJavaScript+&type=code) says otherwise 🤔 
I have to admit that some of the references are pretty plain and wouldn't suffer from the method going exstinct.

> I would consider this a method generating a complex object although it is a string. Which makes the method our of scope for Faker.

I actually kind of agree with you here. BUT:

1. we had no complains about the methods implementation so far (as far as i know)
2. we did a lot of restructuring and deprecating during v8.0 which did upset some people as we know (2 million people are still on latest v7)

Having this method in our codebase doesn't cause a big enought pain point to call for emidiate action in my opinion.

matthewmayer:
I tend to agree this doesnt cause any harm and should be left alone. Perhaps we could add a note this is discouraged for new code without formally deprecating it.

ST-DDT:
> I tend to agree this doesnt cause any harm

The method only produces English output and isn't usable outside of the English locale.
Worse for non English locales this produces mixed language results.

> `fakerZH_CN.finance.transactionDescription()`
> "支付 transaction at 诸州市智杰网络科技有限公司 using card ending with ***(...5783) for GBP 680.89 in account ***18156282"

Note: I'm not against soft-deprecating this method, I just wanted to raise awareness of this fact.

---

If we go for soft-deprecation, how about a text like this:

> **Soft-Deprecated:** New code should not use this method as it generates a text that has no real-world use other than being an English text.
> If you need just a text, please consider using the [`lorem`](https://fakerjs.dev/api/lorem.html) module or [`faker.food.description()`](https://fakerjs.dev/api/food.html#description)
> The Faker team considers this method out of scope for the library as it returns a dynamic complex object disguised as a string.
> Since the method doesn't cause any harm, we will not remove it yet.

ST-DDT:
Or alternatively, we convert it to a proper localized method.

ST-DDT:
**Team Decision**

- We will not deprecate this method.
- We will refactor the current implementation to use localized fake patterns instead of a hard-coded parameterized English string.

ST-DDT:
Ready for review.

matthewmayer (approved):

PR #3203, opened 2024-10-20
infra(unicorn): permanently disable no-object-as-default-parameter

ST-DDT:
Ref: #2439

- #2439

---

Permanently disables the [`unicorn/disable no-object-as-default-parameter`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/disable no-object-as-default-parameter.md) lint rule.

We assume that our users use typescript so they would always pass all required parameters.
Even if they don't all our methods handle that with basic defaults.

Affected code:

````txt
~/faker/src/modules/lorem/index.ts
  135:5  error  Do not use an object literal as default for parameter `wordCount`      unicorn/no-object-as-default-parameter
  203:5  error  Do not use an object literal as default for parameter `sentenceCount`  unicorn/no-object-as-default-parameter
  361:5  error  Do not use an object literal as default for parameter `lineCount`      unicorn/no-object-as-default-parameter

~/faker/test/scripts/apidocs/method.example.ts
  270:5  error  Do not use an object literal as default for parameter `a`  unicorn/no-object-as-default-parameter
  305:5  error  Do not use an object literal as default for parameter `a`  unicorn/no-object-as-default-parameter
  325:5  error  Do not use an object literal as default for parameter `a`  unicorn/no-object-as-default-parameter
````

### Alternative Solution

Change the implementation of the methods to not use signature level defaults.

````diff
sentence(
-    wordCount:
+    wordCount?:
      | number
      | {
          /**
           * The minimum number of words to generate.
           */
          min: number;
          /**
           * The maximum number of words to generate.
           */
          max: number;
-        } = { min: 3, max: 10 }
+        }
  ): string {
-    const sentence = this.words(wordCount);
+    const sentence = this.words(wordCount ?? { min: 3, max: 10 });
...
````

### CODE REVIEW

**Comment by ST-DDT** [comment]
I'm actually not sure which variant I prefer.

**Comment by xDivisionByZerox** [comment]
The alternative solution would actually "catch" accidental `null` inputs from our users:

I compressed the signature for simplicitiy sake:

<table>
<tr>
<th>Object default</th>
<th>Inline default</th>
<tr>
<tr>
<td>

```ts
sentence({
    min: number;
    max: number;
} = { min: 3, max: 10 }): string {
    const sentence = this.words(wordCount);
    // ...
}
sentence(); // works
sentence(undefined); // works
sentence(null); // errors!!!
```

</td>

<td>

```ts
sentence({
    min: number;
    max: number;
}): string {
    const sentence = this.words(wordCount ?? { min: 3, max: 10 });
    // ...
}
sentence(); // works
sentence(undefined); // works
sentence(null); // works
```

</td>
</tr>
</table>

I jsut wanted to point that out.

ST-DDT:
You could say the same for other methods such as `faker.number.int(null)`.
That's why we recommend strict mode.

https://github.com/faker-js/faker/blob/af1dbcdbf363a12c448834dff031a698c8925e2e/docs/guide/usage.md?plain=1#L107

ST-DDT:
Also AFAICT:

> `faker.lorem.sentence(null)`
> "Tenax vilitas harum." 

https://github.com/faker-js/faker/blob/af1dbcdbf363a12c448834dff031a698c8925e2e/src/modules/helpers/index.ts#L1087
(That method doesn't follow our usual options deconstruction though)

matthewmayer:
I also don't have a strong preference. Just want to point out according to https://github.com/faker-js/faker/discussions/3215#discussioncomment-11013917 the faker.lorem.* methods are some of the most heavily used in Faker so we should be very careful not to accidentally change runtime behavior even if it's a subtle change for a edge case.

xDivisionByZerox (approved):

PR #3205, opened 2024-10-20
refactor(locale): remove inaccessible files

### LINKED ISSUE #1612: Remove en_AU_ocker locale

Seems this locale was originally added as a jokey way to add Australian slang names

https://en.wikipedia.org/wiki/Ocker

`['Bazza', 'Bluey', 'Davo', 'Johno', 'Shano', 'Shazza'];`

But there's no API to access this:
https://github.com/faker-js/faker/blob/next/src/locales/en_AU_ocker/person/ocker_first_name.ts

So the only way is via `faker.helpers.fake`

> faker.setLocale("en_AU_ocker")
> faker.helpers.fake("{{person.ocker_first_name}}")
'Bluey'

This seems not worth maintaining a whole locale for, it gets in the way when trying to fix l10n bugs. Maybe remove for 8.0?

### Suggested solution

Remove en_AU_ocker

### Alternative

If there are sentimental reasons for keeping it, it could be kept (I'm not against Easter eggs per se)

### Additional context

_No response_

ST-DDT:
Fixes #1612

- #1612

---

Removes the inaccessible files:

- src/locales/en_AU_ocker/location/region.ts
- src/locales/en_AU_ocker/person/ocker_first_name.ts

The files are theoretically available via faker.helpers.fake, but I couldn't find any calls to that online and I don't expect there to be any either.


matthewmayer (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):
This could be seen as a fix as well. I'd say that all files that are not defined by the definitions are not part of the offical api and thus have no right regarding percistency throughout different versions.

ST-DDT:
I use `refactor(locale)` for now because it shows up as changed locales in the changelog.

PR #3207, opened 2024-10-20
docs(guide): add complex overwrites example

### LINKED ISSUE #1508: Document how to write factory functions with value overwrites

Most who use faker end up wrapping faker methods into "factory functions" that compose those methods into usable objects that represent their underlying data needs.

Example in faker docs itself: https://fakerjs.dev/guide/usage.html#create-complex-objects

```tsx

class User { ... }

function createRandomUser(): User {
  return {
    _id: faker.datatype.uuid(),
    avatar: faker.image.avatar(),
    birthday: faker.date.birthdate(),
    firstName: faker.name.firstName(),
    lastName: faker.name.lastName(),
    sex: faker.name.sexType(),
    subscriptionTier: faker.helpers.arrayElement(['free', 'basic', 'business']),
  };
}

const user = createRandomUser();
```

There are several concepts in this pattern that could be extracted into a relatively lightweight method that could ship along with faker giving first class support and a strong foundation for this use case being implemented "correctly" (I'll lay out my argument for why there should be a correct way to do this"

### Suggested solution

We implemented a version of this in our team internally and would like to open source it. Currently there is no option to opens source within our company, so I wanted to propose we add it directly into faker. The method is implemented and tested already and the actual code is very short so it would likely be a low maintenance addition.

Below I've added our README we use to document the method. The actual method is just a few lines and is just a simple deep merge. It is currently implemented using `deepmerge` but could be refactored to be without dependencies.

Most of the magic is in the custom types. These types enforce good best practices when creating complex objects by leveraging typescript to enforce only required properties to be defined by default and giving an optional override argument to the factory function to allow for defining optional properties.

<----- Here is the README for our implementation ----->

---
title: Writing Tests with Mock Factories
---

### Getting started

When you want to create a new mock, create a `fixtures` folder in the component directory and
place an `index.ts` in it. Mocks can be shared between unit tests and stories.

Folder structure

```bash
 📂Component
 ┣ 📂tests
 ┣ 📂fixtures
 ┃ ┗ index.ts
 ┣ Component.tsx
 ┣ index.ts
```

Then create the mock by passing required properties:

```tsx
// component/fixtures/index.ts
import {ComponentProps} from 'react';
import {createMockFactory} from 'tests/utilities';

import Component from '../component';

type Props = ComponentProps<typeof Component>;

export const mockComponentProps = createMockFactory<Props>(() => {
  return {
    property: 'some default string',
    otherProperty: faker.random.number(),
  };
});
```

Finally, use the mock into your test file or stories:

```tsx
// component/tests/component.test.tsx or
// component/tests/component.stories.tsx
import {mockComponentProps} from '../fixtures';

// Get values from mock
const defaultProps = mockComponentProps();

// Or use customize values
const customProps = mockComponentProps({
  property: 'some custom string',
});
```

**Important**: do not export the mock factories through the component. Doing so might cause the
staging and production builds to fail, as the `/tests` folder is not resolved in those builds.

Conventions:

- name mock factories with `mock${InterfaceName}` so an interface with name `FooBar` would have a
  mock factory named `mockFooBar`
- name component mock factories with suffix `Props` so component with name `Foo` would have a
  mock factory named `mockFooProps`
- name the folder where mocks are exported `fixtures` and export named mock factories from an
  index.ts file in this folder.
- mock factories that should be considered public should be imported from the component fixtures
  `components/MyComponent/fixtures` index.ts file.

### `defaults` argument

Use `createMockFactory<T>` to generate reusable, composable, and strongly typed mock data objects
in your stories or unit tests. To create a mock factory, just pass a typescript interface and
your `defaults` data to the `createMockFactory` method. This method will type check your data,
to make sure it correctly implements all and only all required properties of the interface.

The `defaults` argument will accept either a static object, or a function that will be called
each time a mock is created, allowing for randomized or dynamic data. The default will also restrict
your default values to required props only, so in the example below, `optional` must be set to
undefined in the "default" data set. This ensures that the base set of data for an object covers all
and only all of the required properties, making the "default" case more accurately reflect how
typescript will handle your interface.

```tsx
interface Props {
  disabled: boolean;
  optional?: string;
  nested: MoneyObject;
}

const mockProps = mockMyInterface<Props>({
  disabled: false,
  nested: {
    amount: 10,
    currencyCode: 'USD',
  },
});
```

### `overrideValues` argument

Continuing with the example from above, call `mockProps()` to generate mock objects from the
factory.

`mockProps()` is a generic function, so it also accepts a type ReturnType, the default value
for ReturnType is, well T.

What does that mean? It means that if you call `mockProps()` with no generic argument, it assumes
you want the function to return a value of the interface the mock is mocking. You can also pass
in any interface that extends `T`.

That means you could pass in `DeepOmitOptional<T>` and this would mean mockProps() would only be
allowed to return required properties. you could also pass in `T & SuperT` where `SuperT` adds
additional properties.

Here is an example:

```tsx
interface Person {
  name: string;
  age: number;
  isActive?: boolean;
}

const mockPerson = createMockFactory<Person>(() => ({
  name: faker.name.firstName(),
  age: faker.random.number({min: 0, max: 100}),
}));

interface PersonWithLastName extends Person {
  lastName: string;
}

// mockPerson() only cares that you fulfil the base interface Person.
// so PersonWithLastName can reuse the mock factory of Person, with complete type safety.
mockPerson<PersonWithLastName>({
  ...mockPerson(),
  lastName: 'wow',
});

// required person this will throw an error because isActive is not required on type Person
// type-error isActive of type boolean is not assignable to type undefined
mockPerson<DeepOmitOptional<Person>>({
  isActive: true,
});
```

The `overrideValues` argument that is a `DeepPartial` of your interface.

```tsx
interface Props {
  string: string;
  optional?: number;
  nested: {
    optional?: boolean;
  };
}
// Good
const part: DeepPartial<Props> = {nested: {optional: false}};
```

Using a partial of your interface allows each mock instance to override specific properties
and merge them with the values from `defaults`. A deep partial allows the properties to
be overwritten at any level in the data structure.

```tsx
interface Props {
  string: string;
  optional?: number;
  nested: {
    optional?: boolean;
  };
}

const bad: DeepPartial<DeepOmitOptional<Props>> = {nested: {optional: false}};
// Bad - optional is optional, so must be undefined

const good: DeepPartial<DeepOmitOptional<Props>> = {
  string: 'banana',
  nested: {optional: undefined},
};
// Good - optional can be initialized as an undefined value.

const alsoGood: DeepPartial<DeepOmitOptional<Props>> = {
  string: 'banana',
  nested: {},
};
// Good - optional can be omitted from nested entirely.
```

Typescript will throw an error that the property optional incorrectly implements the type
expected, because types boolean and undefined are incompatible.

NOTE:

`Your IDE will make creating mock objects really easy if you have the typescript language service enabled. This will allow for visual feedback in your editor, as well as autocomplete of your data objects.`

### Composing mocks

You can compose mock factories into larger mock factories. This can reduce code duplication
and make your data much more consistent and reliable. Let's say you define a mockFactory for
`MoneyObject`:

```tsx
const mockMoneyObject = createMockFactory<MoneyObject>(() => ({
  amount: randomMoneyAmount(),
  currency: 'USD',
}));
```

You could then use mockMoneyObject(); in the mockProps factory default argument.

```tsx
const mockProps = createMockFactory<Props>({
  disabled: false,
  nested: mockMoneyObject(),
});
```

Now each time you create a mock from mockProps, the nested property will use mockMoneyObject to
define its data. Because nested expects the type `DeepOmitOptional<MockMoneyObject>` our call to
`mockMoneyObject()` must return only required data.

Instead of having to continuously redefine values for nested interfaces, you can define just a
single mock factory for a typescript interface, and share the mocking logic into any number of use
cases that depend on that interface. Notice though, that if you don't explicitly use
`DeepOmitOptional<T>()` in these composition cases, the value being passed to a parent "defaults"
argument, will throw an error on the property, instead of the value.

```tsx
import {createMockFactory} from './createMockFactory';

interface Address {
  street: string;
  zipCode?: number;
}

const mockAddress = createMockFactory<Address>({
  street: faker.address.street(),
});

interface Person {
  name: string;
  address: Address;
}

// in this case the error will focus on address because it returns the incorrect data set,
// but it won't tell you that zipCode is the issue in a clear precise manner.
createMockFactory<Person>({
  name: faker.random.name(),
  address: mockAddress({
    zipCode: 12345,
  }),
});

// in this case the error will drill down to reveal the zipCode value should be undefined.
createMockFactory<Person>({
  name: faker.random.name(),
  address: mockAddress<DeepOmitOptional<Address>>({
    zipCode: 12345,
  }),
});
```

### Compare to alternatives

You might ask, why use a fancy factory method to create some mock objects? Why not just use
static values? Let's compare two code snippets. One using a more basic, 'on-the-fly' getter
function, versus a mock factory.

1. 'on-the-fly' mock getter

```tsx
function getMockData(input?: Partial<Props>): Props {
  return merge(
    {
      disabled: false,
      nested: {
        amount: 10,
        currencyCode: 'USD',
      },
    },
    input,
  );
}
```

2. mock factory

```tsx
const mockData = createMockFactory<Props>(() => ({
  disabled: faker.random.boolean(),
  nested: mockMoneyObject(),
}));
```

The key differences are that in the second case you rely on less boilerplate code, you achieve a
much higher type safety, and you have much more flexibility in how you design your mocks. Default
values exclude optional properties, giving you very reliable mock data. Using a factory method for
generating the mocks themselves makes the solution scalable and easy to adopt.
Allowing static and dynamic objects retains the option of simple static mocks, but further increases
the reliability of the mock with dynamic values, since you don't have to rely on a single value,
but a range of values that match the type criterion.

### Alternative

I could open source this under a separate repo, but it just feels so close to faker it made sense to at least open the discussion if this would be something that could belong in faker.

### Additional context

_No response_

ST-DDT:
Fixes #1508

- #1508

---

The PR adds an additional example to the complex objects creation documentation.

- [Preview](https://deploy-preview-3207.fakerjs.dev/guide/usage.html#create-complex-objects)


Shinigami92 (approved):

matthewmayer:
Hmm this seems more like a "how to use typescript" than anything faker-specific.

Shinigami92:
> Hmm this seems more like a "how to use typescript" than anything faker-specific. 

What do you mean? Had you a look into the diff?
This is a strategy totally valid to be used in plain JS only 👀
You won't need to add the Partial to the override parameter and it would still work the same.

ST-DDT:
The entire section is more about raising awareness of concepts you could use with faker.
The [NodeJS-Section](https://deploy-preview-3207.fakerjs.dev/guide/usage.html#node-js) explains how you can use faker and the [Create Complex Objects section](https://deploy-preview-3207.fakerjs.dev/guide/usage.html#create-complex-objects) expands on this with ways to structure your methods for complex objects.

A secondary goal is push users into this style of creating "faker" methods, so that they can more easily switch to [standalone module functions](https://github.com/faker-js/faker/issues/2667) (see the code examples in the alternatives section) if they would like to when we introduce them in v10.

xDivisionByZerox (approved):
I'm fine with this, since there was an explicit call out from some community members for this type of documentation.

However, I agree with Matthew that this content seems more related to general programming practices rather than faker specifically. Every programmer should be able to identify these situations as the a requirement for creation design pattern. 
Our documentation primarily focuses on the factory pattern. Is it necessary to prime consumers towards a specific programming style, when other approaches, such as the builder pattern, might be better suited for their needs? While it’s not inherently bad, it might limit the flexibility for users to choose the best method for their particular use case.

Furthermore, I fear that the inclusion of documentation that focuses on the developer interaction **with** faker instead of the usage **of** faker might set expectations for our community that we don't want to get involved in.

matthewmayer:
I'll grudgingly approve 😂

Sometimes you see these kind of worked examples in a kind of "cookbook" section. Could be worth considering in future.

matthewmayer (approved):

PR #3218, opened 2024-10-24
feat(locale): add books on pt_br

guilherme-codes:
It Adds the book locales to pt_BR.


matthewmayer (changes requested):
Please sort all entries alphabetically

matthewmayer:
Also 

Error: Duplicated values are [Obra poética]

If you can run the preflight command locally that will do the sorting for you and spot minor errors like this.

ST-DDT (approved):

matthewmayer (approved):

ejcheng (approved):

PR #3220, opened 2024-10-24
feat: add initial seed parameter to constructors

### LINKED ISSUE #3008: Allow Faker to Take Initial Seed on Construction

As a developer I want to be able to  directly provide a seed when creating a new faker instance. Currently I have to write a separate line mutation the faker instance to configure it the way I want. I feel like it would be good DX to directly create a desired faker instance without having to call additional configuration methods.

### Suggested solution

Add `initialSeed` property to the faker constructor, which will set the seed of the internal randomizer instance on creation.

```ts
const faker = new Faker({ locale: [base] });
faker.seed(123);

// alternative
const faker = new Faker({ locale: [base], initialSeed: 123 });
```

### Alternative

Make `Faker#seed` chainable and return a reference to the current instance. That way I can at leas write it in a shorter way.

```ts
// now
const faker = new Faker({ locale: [base] });
faker.seed(123);

// alternative
const faker = new Faker({ locale: [base] }).seed(123);
```

### Additional context

Same could be argued for `setDefaultRefDate` which could be present as a parameter in Fakers constructor.

ST-DDT:
Fixes #3008

- #3008

---

Adds the ability to add the seed directly to the constructor:

````ts
const customFaker = new SimpleFaker({ seed: 123 });
````

instead of 

````ts
const customFaker = new SimpleFaker();
customFaker.seed(123);
````

Documentation:

- https://deploy-preview-3220.fakerjs.dev/api/simpleFaker.html#constructor
- https://deploy-preview-3220.fakerjs.dev/api/faker.html#constructor

### CODE REVIEW

File: src/simple-faker.ts [hunk 1]
```diff
@@ -120,9 +126,24 @@ export class SimpleFaker {
...
        */
       randomizer?: Randomizer;
+
+      /**
+       * The initial seed to use.
+       * The seed can be used to generate reproducible values.
+       *
+       * Refer to the `seed()` method for more information.
+       *
+       * This option is ignored if `randomizer` is specified.
```
xDivisionByZerox: I would not ignore the option if a randomizer is provided. The interfaces requires a "seed()" method, which I would call in the constructor.

src/simple-faker.ts
```diff
@@ -120,11 +125,25 @@ export class SimpleFaker {
...
        */
       randomizer?: Randomizer;
+
+      /**
+       * The initial seed to use.
+       * The seed can be used to generate reproducible values.
+       *
+       * Refer to the `seed()` method for more information.
+       *
+       * @default randomSeed()
```
ST-DDT: Remove and write as plain text so that we don'T reference an internal method in the docs.

ST-DDT: Done

xDivisionByZerox (approved):

Issue #2390, opened 2023-09-11
Invalid zip/postal codes when using locale en_GB


mikeybinns:
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md).
- [X] Read the [docs](https://fakerjs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/faker-js/faker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Faker issue and not related to a combination with another package.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/faker-js/faker/discussions) or join our [Discord Chat Server](https://chat.fakerjs.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.
- [ ] I am willing to provide a PR.

### Describe the bug

Very closely related to #1416 but for a different locale, so I though a new ticket would be appropriate.

Great Britain post codes also have some specific rules which means that some of the generated postcodes are invalid.

I have a REGEX string for validating if a GB post code is valid, so I'm sharing it here as hopefully it will help, even if only with automated testing. This regex is created based on [this gov.uk document](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/611951/Appendix_C_ILR_2017_to_2018_v1_Published_28April17.pdf)
```ts
/^([A-Z]{1,2}[0-9]{1,2}|[A-Z]{1,2}[0-9]{1}[A-Z]{1})([0-9]{1}[ABDEFGHJLNPQRSTUWXYZ]{2})$/gi,
```

I think the biggest tripping block for Faker here is that the final two letters cannot be C I K M O V.

For those who come across this issue, here's a workaround to get you going in modern TypeScript:
```ts
/**
 * Generate a valid UK postcode using faker.
 * 
 * Faker's implementation is too generic to work with strict 
 * post code validators, so we must check the values returned 
 * to see if they are valid
 * @see https://github.com/faker-js/faker/issues/2390
 * 
 * @returns A promise which resolves to a valid UK postcode
 */
async function generatePostCode() {
  return new Promise<string>((resolve) => {
    let postcode = "";
    while (
      postcode
[...]

PR #3223, opened 2024-10-26
refactor(location): replace zipCode format parameter with style

ST-DDT:
First part of #2390

- #2390

---

Deprecates the `format` parameter in favor of using the native locale patterns or using `faker.helpers.replaceSymbols` directly.


src/modules/location/index.ts
```diff
@@ -71,14 +165,29 @@ export class LocationModule extends ModuleBase {
...
-      format = [format];
+    const { format } = options;
+
+    if (format != null) {
+      deprecated({
+        deprecated:
+          'faker.location.zipCode(format) and faker.location.zipCode({ format })',
+        proposed:
+          'faker.location.zipCode(), faker.location.zipCode({ state }), or faker.helpers.replaceSymbols(format)',
+        since: '9.1.0',
```
```suggestion
        since: '9.2.0',
```

matthewmayer:
Hmm

Given that the default en locale uses the patterns #####, #####-#### I would imagine that a very common case would be forcing 5 digit ZIPs using format:"#####"

We might want to provide better guidance for this scenario.

ST-DDT:
> Hmm
> 
> Given that the default en locale uses the patterns #####, #####-#### I would imagine that a very common case would be forcing 5 digit ZIPs using format:"#####"
> 
> We might want to provide better guidance for this scenario.

I'm not sure what you are asking for here.
Should I add an explicit example for `zipCode('#####')` to `replaceSymbols('#####')` or what?
IMO that is sufficiently covered by the deprecation message already:

````ts
https://github.com/faker-js/faker/blob/ebd0dd04a6055c8878f82c565d6ad13aa34ddd40/src/modules/location/index.ts#L55
````

I could change the runtime message to include the actual format parameter though:

````diff
- 'faker.location.zipCode(), faker.location.zipCode({ state }), or faker.helpers.replaceSymbols(format)',
+ `faker.location.zipCode(), faker.location.zipCode({ state }), or faker.helpers.replaceSymbols('${format}')`,
````

I hope the main usecase for that method is without the format parameter using the patterns provided by the locale.

Here are some calls using the format parameter.

- https://github.com/search?q=%22faker.location.zipCode%28%27%22&type=code (175)
- https://github.com/search?q=%22faker.location.zipCode%28%5C%22%22&type=code (62)

And without format parameter

- https://github.com/search?q=%22faker.location.zipCode%28%29%22&type=code (1.7k)
- https://github.com/search?q=%22faker.location.zipCode%28%7B%22&type=code (42)

**Comment by matthewmayer** [comment]
I think my issue is that `zipCode("#####")` is expressive in a way that none of the suggested replacements are

For example if I have code like

```js
const address = [faker.location.streetAddress(), faker.location.city(), faker.location.state(), faker.location.zipCode("#####")].join("\n")
```

I feel to someone reading the code that's more intuitive than a call to say faker.string.numeric() 

I think "I only want 5 digit zip codes" should be something the zip code method can handle.

ST-DDT:
One of the reasons I don't like that pattern is that it focuses on the wrong issue.
If you want to have a five digit postcode, why don't you use the one provided by your locale?
Does it have more than one variant? aka one short format and one long format?
Isn't the actual issue that there is no "format"/variant option to select between the two?

matthewmayer:
Yes agreed. Having an option where ZIP+4 is only used if an extended:true parameter is used might be good. 

ZIP+4 is rarely used in practice in the US.

ST-DDT:
> Yes agreed. Having an option where ZIP+4 is only used if an extended:true parameter is used might be good.
> 
> ZIP+4 is rarely used in practice in the US.

Unfortunately, I have no clue how the US zip code system works. Other than speeding up international mail delivery (SC) by a week/50% if I use a 9(?) digit zip code instead of the usual 5(?).

Does other countries have similar systems?

matthewmayer:
ChatGPT says:

Most Americans do not know their ZIP+4 code from memory and typically only use the standard five-digit ZIP code in everyday contexts. The ZIP+4 code system is primarily used by the USPS and businesses involved in bulk mailing or shipping, where accuracy and speed are essential. Here’s why ZIP+4 codes are not widely known or commonly used by the general public:

	1.	Limited Necessity: For most personal mail and packages, the five-digit ZIP code is sufficient. USPS delivers accurately without requiring the extra four digits.
	2.	Lack of Familiarity: ZIP+4 codes aren’t widely taught, and they’re not typically requested for most purposes, like filling out forms, so people rarely commit them to memory.
	3.	Complexity: ZIP+4 codes can be specific to buildings, floors, or even particular departments within large companies or institutions. For individuals at residential addresses, the ZIP+4 may not be consistently applied or is often only referenced in specific settings.
	4.	Automation: Online platforms and shipping services that benefit from ZIP+4 often auto-generate or auto-fill it during the address verification process. This eliminates the need for individuals to know or use their ZIP+4 manually.

In short, unless a ZIP+4 code is directly needed for business or government purposes, most people only remember and use the five-digit ZIP.

ST-DDT:
Do you know any other country that has multiple variants of zip codes?

matthewmayer:
Not really. In the UK for example postcodes are like "CB1 1DQ" and in some cases the first part "CB1" might be used to refer to an area like "central Cambridge ". But the default postcode is definitely the long/full one.

ST-DDT:
Would `'short' | 'long'` be sufficient to distinguish between them?

I'll lookup our other locale's zip patterns for this as well.

matthewmayer:
The question is what is the default. For US zip we want short to be default. But in theory what if there's an another locale where the regular zip code is the long version but there's a short version which is a useful alternative.

ST-DDT:
> The question is what is the default.

How about: The default is a random one. That way you can check how your system behaves with unexpected values. If you want a specific format you will look up the options and then pick one that you assume does the thing you need.

ST-DDT:
**Team Decision**

1. We will deprecate the format parameter in favor of a `style: 'short' | 'long'` parameters.
2. We noticed that user facing `zipCode` API uses `postcode` locale data internally. We considered renaming the method, but at the current time we will not implement that. We will discuss this again in the next team meeting
3. We would like to merge postcode and postcode_by_state into a single object `postcode: Record<LiteralUnion<'national'/'*'>, PostCodeDefinitions>` + change from replaceSymbols to fake patterns (This will be done in a later PR to reduce visual diff in this PR)

The postcode definition might look like this:

````ts
type PostCodeDefinitions =
  | {
      pattern: string[];
      short_pattern?: never;
      long_pattern?: never;
    }
  | {
      pattern?: never;
      short_pattern: string[];
      long_pattern: string[];
    };
````

If the `style` parameter is specified, but the current locale only supports one format, then the parameter is ignored.
If the `style` parameter is not specified and the current locale uses multiple formats, then a random one will be picked.

PR #3229, opened 2024-10-26
test: patch and filter seeded run stacktraces

ST-DDT:
As discussed during today's team-meeting:

This PR changes the error location shown by seeded runs.

1. For errors during faker method invocation, the error inside the method is shown as primary location, with a new secondary location of the location of the seeded runs invocation.
   - I also omitted the error location inside locale proxies to highlight the actual access of the locale data.
2. For snapshot diff errors, the error is shown on the call location of the seeded runs.

I'm not sure about the actual stacktrace filtering. Should we only use the stacktrace expansion?

<details>
<summary><h2>Preview</h2></summary>

You can observe two differences in each image category:

1. The error location
2. The stacktrace below the error location

Variants:

- Old => without out this PR
- Stacktrace Expansion => when reverting the vitest.config.ts changes)
- Stacktrace Expansion And Filtering => as implemented by this PR

#### Old

error-old

#### Stacktrace Expansion

error-se

#### Stacktrace Expansion And Filtering

error-se-f

#### Old

snapshot-old

#### Stacktrace Expansion

snapshot-se

#### Stacktrace Expansion And Filtering

snapshot-se-f

</details>


ST-DDT:
If you want to test this changes yourself, you can use this patch:

````diff
     t.describe('zipCode', (t) => {
       t.it('noArgs')
         .it('with string', '###')
-        .it('with format option', { format: '###-###' });
+        .it('with format option', { format: '###-####' })
      // TODO @Shinigami92 2024-03-15: These are currently commented out because non-default locales are currently not supported
-      // .it('with state option', { state: 'CA' })
+        .it('with state option', { state: 'CA' });
       // .it('with options', { state: 'CA', format: '###-###' });
````

File: test/support/seeded-runs.ts [hunk 1]
```diff
@@ -288,6 +300,54 @@ class TestGenerator<
...
+ * @param extraStackFrames The additional stack frames to add after this file.
+ */
+function patchExtraStackFrames(
+  error: unknown,
+  extraStackFrames: () => string[]
+): unknown {
+  if (error instanceof Error && error.stack != null) {
+    const stack = error.stack.split('\n');
+    const index = stack.findLastIndex((e) =>
+      e.replaceAll('\\', '/').includes('/test/support/')
```
Shinigami92: **suggestion (non-blocking):** add a comment that this is a fix for windows separators

test/support/seeded-runs.ts
```diff
@@ -233,7 +241,8 @@ class TestGenerator<
     const tester: MethodTester<TModule[TMethodName]> = {
       it(name: string, ...args: Parameters<TModule[TMethodName]>) {
         expectVariantNotTested(name);
-        vi_it(name, () => callAndVerify(method, args));
+        const extraStackFrames = collectExtraStackFrames(name.length + 7);
```
Shinigami92: **question (non-blocking):** what's this magic number? `7`

Shinigami92 (dismissed):

test/support/seeded-runs.ts
ST-DDT: `it('', `

Shinigami92: we assume that `test('', )` will never be used.
does it also work for e.g. `it.each` and stuff like that? No need to fix it now, but if it occurs, we can adjust it later.

test/support/seeded-runs.ts
ST-DDT: Fixed

test/support/seeded-runs.ts
ST-DDT: This is our custom test framework here, so there is only `t.it('name', ...args)` to call it.
I now use the length of the full string instead of plain math.

Shinigami92 (approved):

Issue #3176, opened 2024-10-12
Proposal: Replace internal fake patterns with a resolver function


ST-DDT:
Continuation of/Requires: #2667, #2838

- #2667
- #2838

Partially inspired by: #1832, #2664

- #1832
- #2664

---

### Clear and concise description of the problem

Glossary:

- fake patterns: String patterns used by faker.helpers.fake
- resolver functions: functions that take `fakerCore` as an argument and return a value based on a seed

Currently, some faker internal methods use fake patterns to generate some complex values based on certain parameters.
However parsing fake patterns is slow and error prone.

### Suggested solution

Introduce a new helper method: `faker.helpers.resolve(fns: ResolverFunction[])`

````ts
type ResolverFunction<T> = (core: FakerCore) => T;

function resolve<T>(core: FakerCore, fns: ResolverFunction<T>[]): T{
    const fn = arrayElement(core, fns);
    return fn(core);
}
````

This could later be extended to support parameters as well.

#### Fake-Usage (OLD)

````ts
export const last_name_pattern = [
	'{{person.last_name}}',
	'{{person.last_name}}-{{person.last_name}}',
];

function lastName(core: FakerCore) {
	return fake(core, core.locale.person.last_name_pattern);
}
````

#### Resolve-Usage (NEW)

````ts
export const last_name_resolvers = [
	(core) => arrayElement(core, core.locale.person.last_name),
	(core) => arrayElement(core, core.locale.person.last_name) + '-' + arrayElement(core, core.locale.person.last_name),
];

function lastName(core: FakerCore) {
	return resolve(core, core.locale.person.last_name_resolvers);
}
````

#### Pros

- Compile time errors instead of runtime errors
- Supports auto completion when creating the resolver functions
- Significantly better performance
- No issues with json encoded parameters

#### Cons

- Breaking change for locale data
- Methods that use the new resolve patterns are harder to populate using a DB/file/string only store

#### Affected Methods

- company.name
- food.description
- food.dish
- location.zipCode
- location.city
- location.street
- location.streetAddress
- location.secondayAddress
- person.lastName
- person.bio
- person.jobTitle

And potentially:

- finance.iban
- ....

### Alternatives

Alternatively, we could follow this pattern for each method individually.

````ts
export const last_name_resolvers = [
	(core) => arrayElement(core, core.locale.person.last_name),
	(core) => arrayElement(core, core.locale.person.last_name) + '-' + arrayElement(core, core.locale.person.last_name),
];

[...]

PR #3233, opened 2024-10-29
refactor(location): use postcode fake patterns

ST-DDT:
Second part of #2390

- #2390

Blocked by #3223

- #3223

---

Replaces the zipcode symbol patterns with fake patterns similar to when using the state parameter.

<details>
<summary>Locale Data Transformation Script</summary>

The locale data have been transformed using the following script:

````ts
// rename.ts
import { glob } from 'glob';
import { rmSync, writeFileSync } from 'node:fs';
import { faker } from './src';

const files = await glob('src/locales/*/location/postcode.ts');

for (const file of files) {
  console.log(file);
  let { default: content } = (await import(`./${file}`)) as {
    default: string | string[] | null;
  };

  rmSync(file, { force: true });

  if (typeof content === 'string') {
    content = [content];
  }

  if (content !== null) {
    content = content.map((pattern) =>
      pattern
        .replaceAll(/(#+)/g, ({ length }) =>
          length === 1 ? '{{string.numeric}}' : `{{string.numeric(${length})}}`
        )
        .replaceAll(/(\?+)/g, ({ length }) =>
          length === 1
            ? '{{string.alpha({ "casing": "upper" })}}'
            : `{{string.alpha({ "casing": "upper" , "length": ${length}})}}`
        )
        .replaceAll(/(\*+)/g, ({ length }) =>
          length === 1
            ? '{{string.alphanumeric({ "casing": "upper" })}}'
            : `{{string.alphanumeric({ "casing": "upper" , "length": ${length}})}}`
        )
    );
    for (const pattern of content) {
      console.log(faker.helpers.fake(pattern));
    }
  }

  writeFileSync(
    file.replace('.ts', '_pattern.ts'),
    `export default ${JSON.stringify(content, null, 2)};`
  );
}
````

`pnpm tsx rename.ts` followed by `pnpm run generate:locales && pnpm run format`

</details>


matthewmayer:
A lot of these seem very verbose compared to what they are replacing. 

In some cases would it make sense to just keep the fake pattern as '{{helpers.replaceSymbols(old pattern}}'

Some could also be simplified by fromRegexp or arrayElement

If we are renaming all the definitions should be change it to zipcode instead of postcode? It's weird that the definition name is different to method name.

ST-DDT:
I can switch back to replaceSymbol syntax no problem.

One reason for the change is, I don't like that method/it's format because it isn't obvious what the symbol represents.
IMO the code is now easier to read and it would even by easier to read, when/if we switch to [resolver functions](https://github.com/faker-js/faker/issues/3176)

In the future, I'd like to deprecate the replaceSymbol methods in favor of regex based methods.
I didn't use regex based implementations yet, because it is too slow for now.

ST-DDT:
> If we are renaming all the definitions should be change it to zipcode instead of postcode? It's weird that the definition name is different to method name.

I'm not sure about the renaming myself yet either.
Should we rename the method or the locale definitions?

ST-DDT:
Note: This is a placeholder PR (at least in my eyes). I plan to replace it (and it's children) with a resolver function in v11.

Requirements:

- #3748 
- #3176 

I mostly keep this around for the fixes in the child PRs.

PR #3221, opened 2024-10-24
docs: expose documentation for randomizers

ST-DDT:
Currently, we sometimes mention the `generateMersenne32Randomizer` and `generateMersenne53Randomizer` methods, but we don't have actual documentation for them on our website.

This PR adds both methods to the Utilities category.

Documentation:

- https://deploy-preview-3221.fakerjs.dev/api/utils.html#generatemersenne32randomizer
- https://deploy-preview-3221.fakerjs.dev/api/utils.html#generatemersenne53randomizer

PR #3242, opened 2024-11-02
docs: expose documentation for all utilities

ST-DDT:
Extension of #3221

- #3221

---

Treats all functions inside the utils folder as public utilities that need documentation.

Also adds a test to ensure that the elements don't change unexpectedly.


xDivisionByZerox (approved):
I'm fine with the change in general. I'm unsure whether including all function in the snapshot is necessary. 🤔 
This concern is not blocking, tho.

ST-DDT:
> I'm unsure whether including all function in the snapshot is necessary. 🤔

I don't think it is necessary either, but currently we don't have any tests that cover the completeness of the docs generation.
This is my attempt to cover this. (I did start with just the utilities module at first, but picking that from the page list is harder than just checking the whole set.)

PR #3250, opened 2024-11-04
infra(commit-and-tag-version): auto-bump version in usage-guide

ST-DDT:
Automates another step from the release process:

> Update [docs/guide/usage.md version pin](https://github.com/faker-js/faker/blob/next/docs/guide/usage.md?plain=1#L82)

### Preview

- https://github.com/ST-DDT/faker/commit/3335e70b40fc596591c44d6073bfd0dbb62bf026

### Limitations

Doesn't seem to work on windows maybe because it launches a new CMD instead of (git-)bash
(I don't expect us to do releases locally again though)

## Alternatives Considered

I also considered using a js script for the replacing, but that requires converting the `.versionrc.json` to `.versionrc.js`.
It might be possible to use a ts file instead, if we tsc "compile" it before delegating to commit-and-tag-version in `pnpm run release`.


xDivisionByZerox:
Cool that you found this, but could you guide me towards documentation on the "scripts" property? I cannot seem to find any references towards that [in their documentation](https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/master/versions/2.2.0/README.md). 🤔

ST-DDT:
Its part of the commit and tag version plugin config:

- https://github.com/absolute-version/commit-and-tag-version?tab=readme-ov-file#lifecycle-scripts

I'm not sure why there are two different documentations and I consider this one hard to read/understand.

You can try the script in a linux docker container.

- `pnpm run release`

xDivisionByZerox (approved):

PR #3254, opened 2024-11-09
refactor(locale): improve zh_CN vehicle manufacturers

Heuluck:
Use the correct translations, and add some vehicle manufacturers to zh_CN


ST-DDT (changes requested):
Please run `pnpm run preflight` once and commit the changes.

ST-DDT (approved):
Thank you for your PR and quick fixes ❤️ .

I checked with Google translate and it looks good to me, but I would love to have a native speaker review this as well.

Heuluck:
It was my first pr so thanks for helping me correcting my mistake

xDivisionByZerox (approved):
Confirmed the changes as well by using a translation program.
In the worst case (we don't find any native speaker), I'd say this is good enought and can be changed by later contributors if values are undesired.

ST-DDT:

The changes will be included in the next release, but we don't a release schedule for it yet.

PR #3255, opened 2024-11-10
infra(unicorn): consistent-function-scoping

ST-DDT:
Ref: #2439

- #2439

---

Move inline functions without scope to the root of the file as standalone functions.

Related:

- #2667 (standalone module functions increase the ease of usuability of these functions, but also have inplications on where these methods needs to be defined e.g. in file or in utils/separate files)


Shinigami92 (approved):

test/modules/number.spec.ts
```diff
@@ -5,6 +5,14 @@ import { seededTests } from '../support/seeded-runs';
 import { MERSENNE_MAX_VALUE } from '../utils/mersenne-test-utils';
 import { times } from './../support/times';
 
+function isFloat(value: number): boolean {
+  return value % 1 !== 0;
```
matthewmayer, line 9: I don't really get the point of this check. 0 is a valid float.

ST-DDT: It is more of a check for hasDecimalPart.

PR #3256, opened 2024-11-10
refactor(helpers): introduce allowHuge flag

### LINKED ISSUE #2695: Improve DX for methods that use all memory
### Pre-Checks

### Describe the bug

Using a big integer as min or max in `faker.string.numeric` in length options, is not handled gracefully and causes `Javascript heap out of memory` crash in V8. It should at least validate the inputs and provide a graceful exception message. Even better; provide big int support in options.

### Minimal reproduction code

```
const { faker } = require("@faker-js/faker");

const fake = {
    bigRange: faker.string.numeric({ length: { min: 1000000000, max: 2000000000 }})
}

console.log("bigRange:", fake.bigRange);
```

### Additional Context

_No response_

### Environment Info

```shell
System:
  OS: macOS (Apple silicon)
Binaries:
  Node: 20.9.0
  npm: 10.1.10
npmPackages:
  @faker-js/faker: 8.4.1
```

- [X] CJS
- [ ] ESM

### Used Package Manager

npm

ST-DDT:
Second part of #2695

- #2695

---

Adds a `allowHuge` flag to `faker.helpers.multiple`. If not set, the number of elements is limited to < 1 Mio.
I considered lowering the limit to just 100k as even 1 Mio elements take quite the while to generate.

Once this PR is accepted I'll continue with the string and lorem(?) module.

### DIFF
--- a/src/internal/check-huge.ts
+++ b/src/internal/check-huge.ts
@@ -1,0 +1,23 @@
+import { FakerError } from '../errors/faker-error';
+
+/**
+ * Check if the count is potentially too huge to be executable in a reasonable time or using default memory limits.
+ *
+ * @internal
+ *
+ * @param count The count to check.
+ * @param name The name of the parameter.
+ * @param allowHuge Whether to allow huge counts.
+ */
+export function checkHuge(
+  count: number | { max: number },
+  name: string,
+  allowHuge: boolean = false
+): void {
+  const maxCount = typeof count === 'number' ? count : count.max;
+  if (!allowHuge && maxCount >= 1e6) {
+    throw new FakerError(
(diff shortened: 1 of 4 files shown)


matthewmayer:
Should this maybe just be a warning rather than an error?

ST-DDT:
> Should this maybe just be a warning rather than an error?

For now, I opted for an error as there is a strong connection between the error and the call location. A warning just appears in the log and needs extra context for assiciating it with the source.
It is just a suggestion to resolve the linked issue at this point and nothing final.

If the others think, this should be a warning or should (not) be there at all. I would like to hear about it as well.

ST-DDT:
**Team Decision**

We will not add an `allowHuge` or similar flag.
JS's `string.repeat` and other similar methods don't have an `allowHuge` flag either.
The main performance impact is from the repeated nested method calls and not by the huge list itself.

We will try to improve the performance of methods generating strings though.

PR #3258, opened 2024-11-12
docs: faker.animal.type now has 44 possible animals

s-inu:
```ts
import { faker } from "@faker-js/faker";

const animals = faker.helpers.uniqueArray(faker.animal.type, 1e3);

const animalSet = new Set(animals);

console.log(animals.length); // 44
console.log(animalSet.size); // 44
```

### DIFF
--- a/docs/guide/unique.md
+++ b/docs/guide/unique.md
@@ -10,7 +10,7 @@ faker.animal.type(); //'horse'
 faker.animal.type(); //'horse'
 ```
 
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):

xDivisionByZerox (approved):
Good catch 👍

matthewmayer:
Maybe we could switch to `faker.location.continent()` which should be a fairly stable length for the next few million years.

ejcheng (approved):

ST-DDT:
> Maybe we could switch to `faker.location.continent()` which should be a fairly stable length for the next few million years.

IMO 44 entries is sufficiently small to get the point across.
I wouldn't mind a PR to change the method reference to that either.

PR #3126, opened 2024-09-21
feat(image): add AI-generated avatars

matthewmayer:
Draft PR to explore the best API for this. CDN URL will need to be replaced with final one.

Uses 100 male and 100 female images generated as part of discussion here https://github.com/faker-js/faker/issues/465#issuecomment-2356107803

fix #465
fix #2173

PR #3259, opened 2024-11-12
feat(person): `sexType` can return `'generic'`

### LINKED ISSUE #3058: Clean up `generic` prefixes
`generic` is intended for cases that don't represent male or female specifically.
In this case it looks like it is just the union of male and female, so it can now be omitted, because it will fallback to the gendered version automatically if it cannot find generic ones.

_Originally posted by @ST-DDT in https://github.com/faker-js/faker/pull/3020#discussion_r1719049653_

ST-DDT:
Fixes #3058

- #3058

Next steps: #3266

- #3266

Meeting-Notes: 2024-11-07

---

This PR tries to lay the foundation of what is considered generic and what is specifically sexed:

- Generic: Values that can be attributed to/used by either of male or female persons. (e.g. `Dr.`)
- Female/Male: Values that can be attributed to specifically female/male (e.g. `Mrs.`/`Mr.`)

It also changes the way how definitions are selected:

- If generic is requested: Then only values that are specifically generic are returned, unless there are no such values.
- If female/male is requested: Then the method will mostly (80%) return female/male values with some (20%) generic values sprinkled in. Since generic can explicitly be both, the generic elements don't cause any issues.

Please check whether the definitions/jsdoc are clear regarding their intentions and if you consider this definition to be correct.
We will clean up the locale data in separate PRs as this might affect lots of files/locales.


src/modules/person/index.ts
```diff
@@ -247,7 +281,7 @@ export class PersonModule extends ModuleBase {
    * @since 8.0.0
    */
   sexType(): SexType {
-    return this.faker.helpers.enumValue(Sex);
+    return this.faker.helpers.arrayElement([Sex.Female, Sex.Male]);
```
Shinigami92, line 306: **issue (blocking):** I need to reject this line-change, as now the method that is used to return one of the possible Sex values only returns a limited sub-set of the possible Sex values. This is not only a breaking change in terms of "expectations" what the method is doing and the cases it is used for, but also is now different to any other kind of `*Type(): *` faker method.

Shinigami92 (dismissed):

src/modules/person/index.ts
Shinigami92: For reference:

https://github.com/faker-js/faker/blob/1633c8deb8f6f41151b39b62689ae3e02ab295bc/src/modules/airline/index.ts#L213-L215
https://github.com/faker-js/faker/blob/1633c8deb8f6f41151b39b62689ae3e02ab295bc/src/modules/color/index.ts#L236-L238
https://github.com/faker-js/faker/blob/1633c8deb8f6f41151b39b62689ae3e02ab295bc/src/modules/color/index.ts#L248-L250

_not sure if these are the only examples, I just searched the first 3 modules_

ST-DDT: FFR: Adding `generic` to the return values could be considered a breaking change too. As this method's return values are specifically intended for programmatic use.

xDivisionByZerox: > This is not only a breaking change in terms of "expectations" what the method is doing and the cases it is used for [...]

I understand your concern about the change potentially breaking expectations. However, technically speaking, this isn't a breaking change as the function's behavior remains consistent with its previous version. I can agree that it might _break_ expectations. If possible, I'd be happy if you could avoid using technical words out of context in arguments.

> [...] but also is now different to any other kind of *Type(): * faker method.

While I'm not entirly sure what qualifies as "*Type(): *" functions for you, but I think `animal.type` might already deviate from the uniformity you are aiming for:

https://github.com/faker-js/faker/blob/1633c8deb8f6f41151b39b62689ae3e02ab295bc/src/modules/animal/index.ts#L201-L203

src/modules/person/index.ts
```diff
@@ -20,30 +36,48 @@ export type SexType = `${Sex}`;
  */
 function selectDefinition<T>(
   faker: Faker,
-  sex: SexType | undefined,
+  sex: SexType = faker.helpers.enumValue(Sex),
```
matthewmayer: doesn't this mean that there is effectively a 1/3 chance of getting the generic option when no parameter is passed (e.g. `faker.person.prefix()`), plus the 20%.

matthewmayer: For example with
```js
export default {
  generic: ['Dr.'],
  female: ['Mrs.', 'Ms.', 'Miss'],
  male: ['Mr.'],
};
```
```js
faker.helpers.multiple(()=>faker.person.prefix(), {count:1000}).filter(a=>a=="Dr.").length 
// 459 
```

which is higher than i would want

ST-DDT: Should the `sex` parameter default to the `sexType()` method then?
Or do you have a different suggestion to solve this imbalance?

matthewmayer: Yes that seems logical.

ST-DDT: If there are no objections, I'll change it to that later today/this week.

src/modules/person/index.ts
ST-DDT: Should I change the method's jsdocs/return type to make it more clear?

Shinigami92: I might not have the time today to accurately respond, sorry.
I would suggest I attend the next meeting and we shortly discuss and I explain on voice.
See you hopefully tomorrow / please wait with merging this PR.

src/modules/person/index.ts
ST-DDT: Changed

matthewmayer:
I understand wanting to keep this PR small but I think it's difficult to decide if this is the right approach without reviewing what locale files will need to be updated too. 

To clarify with this approach would we require the "generic", "female" and "male" options never have any items in common?

ST-DDT:
**Team Decision**

We will change the sexType method to return all SexType values.
A parameter `includeGeneric: boolean = false` can be used to control the inclusion of `generic` in the results.
We will check in v10 whether we want to change the default for `includeGeneric` based on code usage?

ST-DDT: Will create a stacked PR for the locale changes.

- Things in both female and male -> generic
- Things in generic -> remove from female and male
- Without manual changes
- The stacked PR might be later split in smaller chunks to help with reviewability

src/definitions/person.ts
```diff
@@ -1,12 +1,27 @@
...
+       */
       generic?: T[];
+      /**
+       * Values that are primarily attributable to only females.
+       */
       male: T[];
+      /**
+       * Values that are primarily attributable to only males.
+       */
       female: T[];
```
ST-DDT: Should I change the order here to alphabetical as well?

Shinigami92: I would appreciate that

ST-DDT:
> ST-DDT: Will create a stacked PR for the locale changes.
>
> [...]
>
> - Without manual changes
> - The stacked PR might be later split in smaller chunks to help with reviewability

@xDivisionByZerox Currently, the person files aren't sorted, so this by itself causes a massive diff.
Should we sort them first? Would you like to create a PR for that or should I?

(Maybe start with only sorting first, because the cleanup PR will use the list for filtering: See https://github.com/faker-js/faker/pull/3266#issuecomment-2478817148)

src/definitions/person.ts
ST-DDT: Done

xDivisionByZerox:
> @xDivisionByZerox Currently, the person files aren't sorted, so this by itself causes a massive diff.
> Should we sort them first? Would you like to create a PR for that or should I?

Please go ahead. I'm not available during the weekend, sadly.

ST-DDT:
Done: https://github.com/faker-js/faker/pull/3269

- https://github.com/faker-js/faker/pull/3269

ST-DDT:
Updated the distribution to match the result from our team meeting.

- https://github.com/faker-js/faker/pull/3266#issuecomment-2555258729

Ready for review.

src/definitions/person.ts
```diff
@@ -1,15 +1,30 @@
...
 
+/**
+ * Entries that are dependent on a person's sex.
+ */
 export type PersonEntryDefinition<T> =
   | {
+      /**
+       * Values that are primarily attributable to only males.
+       */
+      female: T[];
```
ejcheng, line 11:
```suggestion
      /**
       * Values that are primarily attributable to only females.
       */
      female: T[];
```

src/definitions/person.ts
```diff
@@ -1,15 +1,30 @@
...
+      /**
+       * Values that cannot clearly be attributed to a specific sex or are used for both sexes.
+       */
       generic?: T[];
+      /**
+       * Values that are primarily attributable to only females.
+       */
       male: T[];
```
ejcheng, line 19:
```suggestion
      /**
       * Values that are primarily attributable to only males.
       */
      male: T[];
```

src/definitions/person.ts
```diff
@@ -1,15 +1,30 @@
...
-      female: T[];
     }
   | {
+      female?: never;
+      /**
+       * Values that cannot clearly be attributed to a specific sex or are used both sexes.
```
ejcheng:
```suggestion
       * Values that cannot clearly be attributed to a specific sex or are used for both sexes.
```

xDivisionByZerox:
Note https://github.com/faker-js/faker/pull/3126/files#r1945082472 for implementation details!

src/modules/person/index.ts
```diff
@@ -2,11 +2,27 @@ import type { Faker } from '../..';
...
+ */
 export enum Sex {
+  /**
+   * Is used for values that are primarily attributable to only females.
+   */
   Female = 'female',
+  /**
+   * Is used for values that cannot clearly be attributed to a specific sex or are used for both sexes.
+   */
+  Generic = 'generic',
```
ST-DDT, line 16: TODO: Handle `faker.image.personPortrain({ sex: 'generic' })`

- https://github.com/faker-js/faker/pull/3126#discussion_r1945082472

ST-DDT:
Read for review

test/modules/__snapshots__/person.spec.ts.snap
```diff
@@ -52,19 +56,19 @@ exports[`person > 42 > zodiacSign 1`] = `"Gemini"`;
...
-exports[`person > 1211 > firstName > with sex 1`] = `"Todd"`;
+exports[`person > 1211 > firstName > with sex 1`] = `"Skye"`;
 
-exports[`person > 1211 > fullName > noArgs 1`] = `"Steve Zieme"`;
+exports[`person > 1211 > fullName > noArgs 1`] = `"Dane Osinski-Paucek"`;
 
 exports[`person > 1211 > fullName > with all (sex) 1`] = `"John Doe PhD"`;
 
-exports[`person > 1211 > fullName > with firstName 1`] = `"Mr. John Fahey MD"`;
+exports[`person > 1211 > fullName > with firstName 1`] = `"Mrs. John Fahey DDS"`;
```
matthewmayer, line 67: This looks wrong. Mrs John?

ST-DDT: It is that way because only the first mame is given:

https://github.com/faker-js/faker/blob/f20b83c3d2c342358e766d28eb268112876b6568/test/modules/person.spec.ts#L37

matthewmayer (dismissed):

src/modules/person/index.ts
```diff
@@ -20,30 +36,48 @@ export type SexType = `${Sex}`;
...
+    return binary;
   }
+
+  return (
+    generic ??
+    // The last statement should never happen at run time. At this point in time,
+    // the entry will satisfy at least (generic || (female && male)).
+    // TS is not able to infer the type correctly.
+    []
+  );
```
xDivisionByZerox, line 80: The "catch" case here, that basically duplicates the generic part, could be prevented by using a switch with an explicit handle for `'female' | 'male'` and an implicit one (via the default case) for generic and the unhandled scenario when `binary == null`. I'd be in favor of structuring like this. Other ideas?

xDivisionByZerox: I've tried to make sense of this comment after half a year and decided that whatever I tried to point out here is not important enough. The current implementation works and while being a little awkward with the duplicated generic handling, is quite easy to understand which is great for the maintainability in this area of the project. So I'm closing this discussion.

xDivisionByZerox (dismissed):

matthewmayer:
Looks good to me, but I think let's wait till after 10.2.0 is released to merge it. 

Then this as well as the follow on locale changes can be part of 10.3.0.

xDivisionByZerox:
Thanks for the input @matthewmayer :+1:
I'll mark this PR with https://github.com/faker-js/faker/labels/do%20NOT%20merge%20yet to prevent us from merging it for now.

matthewmayer:
Happy for this to merge now 10.2 is out, targeting 10.3 release along with associated cleanup of the locale data.

matthewmayer (dismissed):

test/modules/image.spec.ts
```diff
@@ -121,6 +121,19 @@ describe('image', () => {
...
+    it.each(Object.values(Sex))(
+      'should return a random avatar url from AI for %s',
+      (sex) => {
+        const imageUrl = faker.image.personPortrait({ sex });
+
+        expect(imageUrl).toBeTypeOf('string');
+        expect(imageUrl).toMatch(
+          /^https:\/\/cdn\.jsdelivr\.net\/gh\/faker-js\/assets-person-portrait\/(female|male)\/512\/\d{1,2}\.jpg$/
+        );
+        expect(() => new URL(imageUrl)).not.toThrow();
```
Shinigami92: **chore:** 

```suggestion
        expect(() => new URL(imageUrl)).not.toThrowError();
```

Shinigami92 (approved):
Approving this in the hope that `includeGeneric` will become someday default `true`, for political reasons, diversity and a better world I can live in.

xDivisionByZerox (approved):

xDivisionByZerox:
We might want to follow up with a draft PR that changes the default of `includeGeneric` to `true` as a breaking change for the major version (whenever this will be). Or at least create an issue for it?

Also, this PR will not show up in the changelog due to being a "refactor". We might want to rephrase the title to "feat(person): add optional parameter includeGeneric to sexType" or something similar. What do you think?

Shinigami92:
> We might want to follow up with a draft PR that changes the default of `includeGeneric` to `true` as a breaking change for the major version (whenever this will be). Or at least create an issue for it?

I can do this/take this. However I think we have a lot of time for that, as I'm not sure when we will start thinking about v11.

> Also, this PR will not show up in the changelog due to being a "refactor". We might want to rephrase the title to "feat(person): add optional parameter includeGeneric to sexType" or something similar. What do you think?

👍 really good idea

---

Edit: I will open a Issue/PR when this PR here got merged

xDivisionByZerox:
> We might want to follow up with a draft PR that changes the default of includeGeneric to true as a breaking change for the major version (whenever this will be). Or at least create an issue for it?

FYI: I've opened #3694, so we don't forget.

Issue #3261, opened 2024-11-14
Lorem EN `word.ts` contains a single character in the pool


konarx:
### Pre-Checks

- [X] Follow our [Code of Conduct](https://github.com/faker-js/faker/blob/next/CODE_OF_CONDUCT.md).
- [X] Read the [Contributing Guidelines](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md).
- [X] Read the [docs](https://fakerjs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/faker-js/faker/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Faker issue and not related to a combination with another package.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/faker-js/faker/discussions) or join our [Discord Chat Server](https://chat.fakerjs.dev).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.
- [ ] I am willing to provide a PR.

### Describe the bug

I was using `lorem.word` for testing, and I had a failing test. I was confident enough to ping the developer:
-"Hey, you have a bug here."
-"No, you are providing a single character when I expect a word of at least two characters."
-"No, I don't. I use a library that specifically uses random words, not chars".
-"Yes, you are. Here, see the payload you provided yourself."
-"Ah..."
And here I am :) There is an 'a' character [here](https://github.com/faker-js/faker/blob/next/src/locales/en/lorem/word.ts#L2), which is NOT a word, so I do not think it should be in this pool.

### Environment Info

```shell
-
```

### Which module system do you use?

- [ ] CJS
- [ ] ESM

### Used Package Manager

npm

ST-DDT:
FFR: 

https://github.com/faker-js/faker/blob/467bd83dbd2f34dbd5080c45de18901823f83469/src/locales/en/lorem/word.ts#L2

The English locale does have other one character (non-lorem) words as well (e.g. most prominently `I` and `a`).
Do you consider those to be words?
To be clear, I'm not against removing it, I just wish to understand your usecase a bit more.
Because if we remove `a` from the list, then maybe someone else considers 2 letter words to be too short.

If you need words of a certain length, have you tried [`faker.word.sample({length: { min: 2, max: 1000 }})`](https://fakerjs.dev/api/word.html#sample)?
Or do you specifically need a similar feature for the lorem words?

[...]

PR #3262, opened 2024-11-14
refactor(word): cleanup filter-word-list-by-length.ts

ST-DDT:
Cleanup some long forgotten code in the words module.

The first part was not used because we shipped TS4 at that time.
The later now uses an explicit nullish check instead of a falsy check.

Is this a chore or a refactor?

Found while looking at: #3261

- #3261


Shinigami92:
It's a `refactor` if `length` was previously explicitly checked against `0`

Shinigami92 (approved):

PR #3272, opened 2024-11-19
infra(unicorn): prefer-export-from

ST-DDT:
Ref: #2439

- #2439

---

Enables the [`unicorn/prefer-export-from`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-export-from.md) lint rule.

I used the playground to check whether this affects tree shaking somehow, and it doesn't have a noticeable impact on the bundle size.


src/locales/base/location/time_zone.ts
```diff
@@ -1,4 +1,3 @@
-import timeZone from '../date/time_zone';
-
 // The base locale uses any possible time zone.
-export default timeZone;
+
```
ST-DDT:
```suggestion
```

src/locales/hr/location/county.ts
```diff
@@ -1,3 +1,3 @@
 // Since first-level administrative regions of Croatia are called "counties" we return them both from the state() and county() methods
-import state from './state';
-export default state;
+
```
ST-DDT:
```suggestion
```

PR #3269, opened 2024-11-16
refactor(locale): sort person data

ST-DDT:
~~Follow-on to #2265~~ (the actual follow on will be after #3266)

- ~~#2265~~

Preparation for #3266

- #3266

---

Sorts (and uniques) the entries in the `person` module definitions in all locales.

The first commit contains the script changes that were used to apply the changes.

PR #3279, opened 2024-11-25
refactor(locale): split up Spanish generic first names

ST-DDT:
Preparation for #3269

- #3269

---

Splits and deduplicates the Spanish generic first names  into male and female.
I also had to fix the street name patterns, as the generic set no longer exists.

This PR does not sort the split up lists to reduce the diff.


xDivisionByZerox (approved):
Lists contained 227 unique names before and they contain 227 unique names now. LGTM 👍

PR #3284, opened 2024-11-26
docs(phone): update faker.phone.number() example

### LINKED ISSUE #3283: The `faker.phone.number()` is returned differently than in the documentation.
### Pre-Checks

### Describe the bug

I read the [documentation](https://fakerjs.dev/api/phone.html#number) and thought that faker.phone.number() would return a string of the form xxx-xxx-xxxx, but it doesn't.

Need to supplement the documentation so that readers can fully expect the value returned by `faker.phone.number()`.

### Minimal reproduction code

image

### Additional Context

_No response_

### Environment Info

```shell
System:
  OS: macOS 14.1.1
  CPU: (8) arm64 Apple M2
  Memory: 111.69 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.9.0 - /opt/homebrew/bin/node
  Yarn: 1.22.21 - /opt/homebrew/bin/yarn
  npm: 10.8.3 - /opt/homebrew/bin/npm
  pnpm: 9.5.0 - /opt/homebrew/bin/pnpm
  bun: 1.1.26 - ~/.bun/bin/bun
Browsers:
  Chrome: 131.0.6778.86
  Safari: 17.1
npmPackages:
  @faker-js/faker: ^9.2.0 => 9.2.0
```

- [x] CJS
- [x] ESM

### Used Package Manager

npm

sounmind:
- Fixes #3283

### DIFF
--- a/src/modules/phone/index.ts
+++ b/src/modules/phone/index.ts
@@ -19,7 +19,7 @@ export class PhoneModule extends ModuleBase {
    * @see faker.helpers.fromRegExp(): For generating a phone number matching a regular expression.
    *
    * @example
-   * faker.phone.number() // '961-770-7727'
+   * faker.phone.number() // '961-770-7727' (default 'human' style)
    * faker.phone.number({ style: 'human' }) // '555.770.7727 x1234'
    * faker.phone.number({ style: 'national' }) // '(961) 770-7727'
    * faker.phone.number({ style: 'international' }) // '+15551234567'


src/modules/phone/index.ts
```diff
@@ -19,7 +19,7 @@ export class PhoneModule extends ModuleBase {
    * @see faker.helpers.fromRegExp(): For generating a phone number matching a regular expression.
    *
    * @example
-   * faker.phone.number() // '961-770-7727'
+   * faker.phone.number() // '961-770-7727' (default 'human' style)
```
xDivisionByZerox, line 22: This would require a change on ALL our methods for consistency purposes. Also, what happens when the options object get more parameters in the future? Do you intend to list all default values for each parameter here? Take a look at one of the potential worst cases: [internet.jwt()](https://fakerjs.dev/api/internet.html#jwt). Would you honestly make the call to add the following chang eto the example?

```diff
     * @example
-    * faker.internet.jwt() // 'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE3MTg2MTM3MTIsImV4cCI6MTcxODYzMzY3OSwibmJmIjoxNjk3MjYzNjMwLCJpc3MiOiJEb3lsZSBhbmQgU29ucyIsInN1YiI6IjYxYWRkYWFmLWY4MjktNDkzZS1iNTI1LTJjMGJkNjkzOTdjNyIsImF1ZCI6IjczNjcyMjVjLWIwMWMtNGE1My1hYzQyLTYwOWJkZmI1MzBiOCIsImp0aSI6IjU2Y2ZkZjAxLWRhMzMtNGUxNi04MzJiLTFlYTk3ZGY1MTQ2YSJ9.5iUgaCaFVPZ8d1QD0xMjoeJbmPVyUfKfoRQ6Njzm5MLp5F4UMh5REbPCrW70fAkr'
+    * faker.internet.jwt() // 'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE3MTg2MTM3MTIsImV4cCI6MTcxODYzMzY3OSwibmJmIjoxNjk3MjYzNjMwLCJpc3MiOiJEb3lsZSBhbmQgU29ucyIsInN1YiI6IjYxYWRkYWFmLWY4MjktNDkzZS1iNTI1LTJjMGJkNjkzOTdjNyIsImF1ZCI6IjczNjcyMjVjLWIwMWMtNGE1My1hYzQyLTYwOWJkZmI1MzBiOCIsImp0aSI6IjU2Y2ZkZjAxLWRhMzMtNGUxNi04MzJiLTFlYTk3ZGY1MTQ2YSJ9.5iUgaCaFVPZ8d1QD0xMjoeJbmPVyUfKfoRQ6Njzm5MLp5F4UMh5REbPCrW70fAkr' (default { alg: faker.internet.jwtAlgorithm(), typ: 'JWT' } header, { iat: faker.date.recent(), exp: faker.date.soon(), nbf: faker.date.anytime(), iss: faker.company.name(), sub: faker.string.uuid(), aud: faker.string.uuid(), jti: faker.string.uuid() } payload, faker.defaultRefDate() refDate )
```

The formatting of line alone is a headache in itself.

Line 33 perfectly describes describes the default behaviour of the function. There is nothing to change here IMO.

https://github.com/faker-js/faker/blob/b390432626811d2af32086b16151bc4c55634c60/src/modules/phone/index.ts#L33

Please also note the definition of the word example:

> one of a number of things, or a part of something, taken to show the character of the whole:
> _This painting is an example of his early work._
>
> Synonyms: specimen, sample
> [...]

<sub>[Source: dictionary.com](https://www.dictionary.com/browse/example)</sub>

---

Although, I have to thank you. By writing this comment I noticed that the default example for `internet.jwt` was missing it's return value:

https://github.com/faker-js/faker/blob/b390432626811d2af32086b16151bc4c55634c60/src/modules/internet/index.ts#L1101

xDivisionByZerox (changes requested):

src/modules/phone/index.ts
sounmind: I overlooked the potential impact on document consistency.
I also didn’t realize there could be examples like JWT.
The definition of “example” was helpful as well.

Thank you for explaining with rich context!
I now understand the current way the document is expressed.

There aren’t many people like me giving similar feedback, so I think it’s better not to make this change. I'll close this PR soon.
I’m really glad, though, that this feedback incidentally helped catch the mistake regarding the jwt return value!

matthewmayer: What if we just had two lines with two different examples.

xDivisionByZerox: > There aren’t many people like me giving similar feedback, so I think it’s better not to make this change. I'll close this PR soon.

It's unfortunate because we require feedback like this to understand the community's needs better. While we can't accept every request or make every change suggested by community members, such insights are invaluable for planning the project's future. 

Please don't feel discouraged by the rejection of this particular change. I hope to see your contributions (in any size and form) in the future.

> What if we just had two lines with two different examples.

I believe this would lead to inconsistencies in the documentation as well. 

If I had to identify the root cause of this issue, I would say that the example value for the default seems "too perfect". How about we change it to another human-styled phone number that isn't in the form of `xxx-xxx-xxxx`?

ST-DDT: I also considered adding a refresh button to the examples replacing the values in the `// comments`.

xDivisionByZerox: > I also considered adding a refresh button to the examples replacing the values in the `// comments`.

This sounds like super cool idea! Could you extract that idea into it own issue? That way we can start dedicated discussions, make arguments and collect implementation ideas in its own context.

ST-DDT: Done: https://github.com/faker-js/faker/issues/3287

- https://github.com/faker-js/faker/issues/3287

PR #3288, opened 2024-11-28
refactor: replace MersenneTwister implementation with a pure-rand based one

ST-DDT:
This PR replaces our `MersenneTwister` implementation with one modified from https://github.com/dubzzz/pure-rand.
This PR does not change the underlying algorithm/results per seed.

# Reasons for the change

- Improved performance
- Avoid long license preamble
- (Slightly improved readability)

# Improved Performance

|  MersenneTwister  |               Performance | Compared to old (32bit) |
| :---------------: | ------------------------: | ----------------------: |
| pure-rand (32bit) | 61,218,398 ops/sec ±2.47% |                    3.85 |
|    old (32bit)    | 15,896,596 ops/sec ±0.85% |                    1.00 |
|    old (53bit)    |  8,582,587 ops/sec ±1.07% |                    0.53 |
|    new (32bit)    | 67,500,658 ops/sec ±2.67% |                    4.25 |
|    new (53bit)    | 48,467,687 ops/sec ±1.59% |                    3.05 |

(higher is better)

Fastest is `new (32bit)` with 4.25 times the speed of faker's old implementation.

This especially improves the performance for our - current default - 53 bit variant with a speed up by 5.65 times.

One of the main reasons for the speed up is that `pure-rand`'s implementation uses existing `Math` functions intended to work with `int32` instead of manually clamping the values to the 32bit bounds after each operation. Since the new implementation is based on that, we also get these improvements.

- [Benchmark Source](https://github.com/user-attachments/files/17952606/mersenne-bench.zip)

# Why not use `pure-rand` as a dependency

1. [pure-rand](https://www.npmjs.com/package/pure-rand)'s implementation is based on the official spec, it does not have our `initByArray` method and I was unable to find that in the official spec to verify that it is compliant. @dubzzz feel free to use our new implementation of that method for your library.
2. Adding the code to our source allows us to apply some optimizations to it, that aren't back-portable to pure-rand
   - Avoid array copy during use instead of being immutable
   - Add the methods to generate float32/float53 instead of just int32
   - Add `initByArray` method and maybe later `initByString`/`initByAny` that depend on it
3. I also tried vendoring the file, but TS's `private static member functions` are up to 75% slower when compiled with target ESNext in comparison to `standalone functions` for some reason. It is the other way round when compiled with ES3/5 (used by pure-rand).
4. Lastly, we don't have any dependencies and thus it would be our first

---

@dubzzz We already tried to contact you via e-mail to check whether you are fine with our plans to use your implementation as a base for ours. Please let us know what you think.
The code remains MIT licensed, we just wanted to make sure you aren't blindsided by this change.


dubzzz:
Clearly ok for me 👍
If I can help on providing missing helpers directly from pure-rand, let me know. Not for this release but maybe later in the future

Shinigami92 (approved):
Definitely an approval from my side, because it looks like that no snapshot were altered and so it is a "one to one move forward"

matthewmayer:
Could there be some logic in adding pure-rand as a devDependency and adding a few tests that show that the output of our function is identical to the output from pure-rand? To guard against any possible divergence in future.

ejcheng (approved):

xDivisionByZerox:
> Could there be some logic in adding pure-rand as a devDependency and adding a few tests that show that the output of our function is identical to the output from pure-rand? To guard against any possible divergence in future. 

That's not what we are aiming for here. Our goal is to have a faster implementation of mersenne. We only use pure-rand's implementation as a basis.

test/utils/mersenne.spec.ts
```diff
@@ -16,69 +16,63 @@ import {
...
-}
-
 describe('MersenneTwister19937', () => {
-  describe('genrandInt32()', () => {
+  describe('nextU32()', () => {
     it('should be able to return 0', () => {
-      const twister = newTwister(257678572);
+      const twister = new MersenneTwister19937(257678572);
 
       // There is no single value seed that can produce 0 in the first call
```
xDivisionByZerox, line 24: **Soft blocking:** I'm still running the implementation for all possible numbers to see if there is a faster approche after this rewrite. I know that there shouldn't be, but I still want to confirm that.

xDivisionByZerox (changes requested):

test/utils/mersenne.spec.ts
dubzzz: By the way is it true that there is no seed able to return a sequence starting at zero?

Regarding the tests for randomness I did some on pure-rand. Maybe you can check on that side to see how I tested it. At least it could give ideas.

xDivisionByZerox: @ST-DDT was running the tests for the old implementation and did not find any value. Since both implementations should result in the same outcome, I'm not very hopeful  to find something. I'll comment an update when I find the first seed with a minimal sequence resulting in 0.

ST-DDT: > By the way is it true that there is no seed able to return a sequence starting at zero?

There is no singular seed for that. You can achieve this with an appropriately selected state (which we don't support), and maybe using the seed array, but there are just too many combinations to try and find one.

xDivisionByZerox (approved):

dubzzz:
@ST-DDT It took me a while to come back to this thread, but I started to make some changes on pure-rand that are addressing parts of your concerns.

Version 8 embraces mutable by default while still offering an easy way to have pure generators. Distributions tailored for float32 and float53 already reached main, and will certainly be soon part of 8.1.0.

The init by array is not yet a thing but I plan to work on it along with providing ways to jump on the mersenne twister generator.

Regarding the no dependency thing I get your point totally. I know that it's also the position of effect-ts, but they made an exception for fast-check (a project of mine).

_Just wanted to give you a heads up on the new version_

ST-DDT:
Thanks for the update!

PR #3301, opened 2024-12-01
docs(api): add refresh button to examples

### LINKED ISSUE #3287: Add refresh button to the api docs examples

Currently, the documentation only provides a single example per invocation, this could result in wrong expectations of method return values.

See:

- https://github.com/faker-js/faker/issues/3283

### Suggested solution

Adding a refresh button to the example sections:

https://github.com/user-attachments/assets/f3b38e5b-a256-4114-9c62-0c969bfa7fdd

For that to work, we have to 

- trigger downloading of the latest faker version (on first press)
- change our apidocs json data to be js/ts, so that we can store actual code in there
- add a function `() => unknown[]` to the export data from the actual examples
- take each result JSON.stringify it and insert into the code block

### Alternative

Add more examples for the no args examples.

### Additional context

Original issue:

- https://github.com/faker-js/faker/issues/3283

Original fix suggestion:

- https://github.com/faker-js/faker/pull/3284

Alternative suggestion for a fix:

- https://github.com/faker-js/faker/pull/3284#discussion_r1859214745

ST-DDT:
Fixes #3287

- #3287

---

Adds a refresh button to the examples without requiring changes to our examples.

All `faker.` invocations are recorded and pasted as a comment after the invocation.

[Preview](https://deploy-preview-3301.fakerjs.dev/api/lorem.html#paragraphs)


ST-DDT:
This PR is ready for reviewing and feedback (except some minor details).
But I would like to create some test before merge, so I marked this as a draft for now.

docs/api/.gitignore
ST-DDT, line 1: Eslint does not know this .gitignore exists and thus processes the generated api docs files.
So I moved the contents from here to the main .gitignore.

matthewmayer (commented):
The quote styling for strings is different before and after pressing refresh

Screenshot 2024-12-02 at 00 00 14

Screenshot 2024-12-02 at 00 00 18

ST-DDT:
The same applies to `,` vs `, ` in arrays.

For the quotes it is a relatively easy fix.
Not sure for arrays.

Do you think either of the two issues should be fixed?

matthewmayer:
> The same applies to `,` vs `, ` in arrays.
> 
> For the quotes it is a relatively easy fix. Not sure for arrays.
> 
> Do you think either of the two issues should be fixed?

preferably yes, the visual glitch when the first refresh happens is a little disconcerting

matthewmayer:
Also noticed the refresh button animates weirdly (the circle moves around as its not centered). My preference would be for a more informative button like "Load more examples..." as suggested before, but welcome to suggest other wordings/placements

ST-DDT:
> Also noticed the refresh button animates weirdly (the circle moves around as its not centered).

That's one of the reasons, this PR is a draft.
I'll probably create an svg to fix that.

> My preference would be for a more informative button like "Load more examples..." as suggested before, but welcome to suggest other wordings/placements

I personally prefer the icon, but that just might be because I have a hard time visualizing the text button/it's placement.
Maybe with a screenshot it would be easier to visualize.

If we go for text we should propably replace "more" with "new" because they dont get appended.

I do plan on adding a hover text/title to it though.

ST-DDT:
How readable do you think the current implementation is?

matthewmayer:
> How readable, do you consider the current implementation?

i dont think its very intuitive what would happen if you press it (and bear in mind hover titles dont work on mobile). So i'd prefer something more explicit.

matthewmayer:
Maybe something like this
"Load new examples"
"Load random examples"
"Randomize examples"

image

xDivisionByZerox:
> i dont think its very intuitive what would happen if you press it (and bear in mind hover titles dont work on mobile). So i'd prefer something more explicit.

I understand your perspective, but I respectfully disagree. The reload symbol is a widely recognized icon in UI/UX design, and it's generally understood by users. Our documentation is aimed at developers, who are likely familiar with this symbol. 

Regarding the mobile issue, our primary focus has been on optimizing the desktop experience. While mobile functionality is important, the majority of our users access the documentation on desktop. Therefore, we prioritize desktop usability. I appreciate your feedback, but I believe the current design is effective for our target audience.

ST-DDT:
> How readable, do you consider the current implementation?

Well, I meant the actual source code, not the visualization. I was recently told that my code is hard to read/maintain for other developers.
So I appreciate any feedback. Is the complexity of the code appropriate for the complexity of the task?

ST-DDT:
> preferably yes, the visual glitch when the first refresh happens is a little disconcerting

Should I 
- dynamically download/embed prettier to format the example results
- or should we generally switch to JSON like result examples?

matthewmayer:
Pulling in prettier client-side for this sounds like overkill. I think using JSON-like results with double quotes in the example output sounds simpler.

docs/.vitepress/components/api-docs/format.ts
ST-DDT, line 1: This file is only here to demonstrate, how it would look like with async loaded prettier.
I was unable to provide a sane way to convert the data to text without writing an entire stringify library.

We do have the following options:

- JSON.stringify (and change the native examples)
- JSON.stringify + custom  (currently buggy for JSON objects) https://github.com/faker-js/faker/blob/36104dab447f6797e8cf22809bc29f3f2436dd5a/docs/.vitepress/components/api-docs/method.vue#L111-L124
- JSON.stringify + prettier https://github.com/faker-js/faker/blob/a7d7177d34d98cb11dd2e22eda99f7b5be3e67a0/docs/.vitepress/components/api-docs/format.ts#L7-L15
- custom stringify 

Lets discuss this in the next team meeting.

ST-DDT:
I noticed some inconsistent formatting in our jsdoc examples as well, so for now I decided to remove the formatting part from the PR and I will create a separate issue for consistent formatting of our examples both in our source code as well during refresh.
To keep the visual diff low on the website, I will keep some basic replaces in place.

cypress/e2e/example-refresh.cy.ts
```diff
@@ -0,0 +1,32 @@
+describe('example-refresh', () => {
+  it('should refresh the page', () => {
```
ST-DDT:
```suggestion
  it('should refresh the example', () => {
```

ST-DDT:
Ready for review

xDivisionByZerox:
I did some explorative testing on the preview and found that the `location.language` examples produce an error in the console:

- [Link to docs](https://deploy-preview-3301.fakerjs.dev/api/location.html#language)

```txt
framework.CKFbI6s_.js:13  TypeError: Cannot read properties of undefined (reading 'lastElementChild')
    at J (method.DAik2iG1.js:1:2549)
    at Proxy.O (method.DAik2iG1.js:1:2981)
    at a (method.DAik2iG1.js:1:1350)
    at Qt (framework.CKFbI6s_.js:13:38)
    at He (framework.CKFbI6s_.js:13:108)
    at HTMLButtonElement.n (framework.CKFbI6s_.js:18:7223)
```

Otherwise the experice feels pretty clean. One weird thing might be the examples of [`faker.seed`](https://deploy-preview-3301.fakerjs.dev/api/faker.html#seed) since it is designed to return the same value for a designated seed. I don't think there is some special handling neccessary, tho.

ST-DDT:
> I did some explorative testing on the preview and found that the `location.language` examples produce an error in the console:
> 
> * [Link to docs](https://deploy-preview-3301.fakerjs.dev/api/location.html#language)
> 
> 
> ```
> framework.CKFbI6s_.js:13  TypeError: Cannot read properties of undefined (reading 'lastElementChild')
>     at J (method.DAik2iG1.js:1:2549)
>     at Proxy.O (method.DAik2iG1.js:1:2981)
>     at a (method.DAik2iG1.js:1:1350)
>     at Qt (framework.CKFbI6s_.js:13:38)
>     at He (framework.CKFbI6s_.js:13:108)
>     at HTMLButtonElement.n (framework.CKFbI6s_.js:18:7223)
> ```

The issue was caused by the additional property after the function call.

`faker.location.language().name`

For now, I added support for a single property after the function call.
(And improved the handling when no end can be detected.)

You can test this with the airline module, just add a property to one of the examples there.
(I don't know a method that has them yet other than location.language)
faker.location.language now works to a certain extend (aka it will display TypeErrors because the method does not exist in the latest release).

There are 3 ways to handle these type errors:

- disable refresh buttons on new methods (currentVersion < since)
  (This doesn't handle breaking changes/new options)
- or change the error name to "not released yet"
- or use the built version of faker instead of the released one (separate PR)

Especially for new methods/during their initial review, it might be helpful to be able to show more results with a simple click.
Note: Always review the PR code changes first, before viewing any preview.

matthewmayer:
If we can find a way where it works for non-released methods even deploy previews I think that's a nice addition to help with QA.

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
   throws,
   signature,
   examples,
+  refresh,
   seeAlsos,
   sourcePath,
 } = method;
 
+const code = ref<HTMLDivElement | null>(null);
```
Shinigami92: **nitpick (non-blocking):** just some DX improvements that e.g. help reducing clutter in IDEs when using inlayHints.

```suggestion
const code: Ref<HTMLDivElement | null> = ref(null);
```

or

```suggestion
const code = useTemplateRef('code');
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -1,8 +1,11 @@
 <script setup lang="ts">
+import { computed, ref } from 'vue';
```
Shinigami92: **nitpick (non-blocking):** 
```suggestion
import type { ComputedRef, Ref } from 'vue';
import { computed, ref } from 'vue';
```

or

```suggestion
import type { ComputedRef, Ref } from 'vue';
import { computed, ref, useTemplateRef } from 'vue';
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
...
   signature,
   examples,
+  refresh,
   seeAlsos,
   sourcePath,
 } = method;
 
+const code = ref<HTMLDivElement | null>(null);
+const codeBlock = computed(() => code.value?.querySelector('div pre code'));
+let codeLines: Element[] | undefined;
```
Shinigami92: **nitpick (non-blocking):** 
```suggestion
const codeLines: Ref<Element[] | undefined> = ref();
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
...
+      lastElementChild.textContent += ' ';
+    }
+  }
+
+  return result;
+}
+
+async function onRefresh() {
+  if (refresh != null && codeBlock.value != null) {
+    codeLines ??= initRefresh();
```
Shinigami92: **nitpick (non-blocking):** 
```suggestion
    codeLines.value ??= initRefresh();
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
...
+      !lastElementChild.textContent?.endsWith(' ')
+    ) {
+      lastElementChild.textContent += ' ';
+    }
+  }
+
+  return result;
+}
+
+async function onRefresh() {
```
Shinigami92: **nitpick:** 
```suggestion
async function onRefresh(): Promise<void> {
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
...
+
+    // Remove old comments
+    codeBlock.value
+      .querySelectorAll('.comment-delete-marker')
+      .forEach((el) => el.remove());
+
+    // Insert new comments
+    for (let i = 0; i < results.length; i++) {
+      const result = results[i];
+      const domLine = codeLines[i];
```
Shinigami92: **nitpick (non-blocking):** 
```suggestion
      const domLine = codeLines.value[i];
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -51,8 +157,14 @@ function seeAlsoToUrl(see: string): string {
...
     <div v-html="signature" />
 
-    <h3>Examples</h3>
-    <div v-html="examples" />
+    <h3 class="inline">Examples</h3>
+    <RefreshButton
+      class="refresh"
+      v-if="refresh != null"
+      style="margin-left: 0.5em"
+      :refresh="onRefresh"
```
Shinigami92, line 165: **nitpick:** 
```suggestion
      @refresh="onRefresh"
```

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -14,10 +17,113 @@ const {
...
+    // Remove old comments
+    codeBlock.value
+      .querySelectorAll('.comment-delete-marker')
+      .forEach((el) => el.remove());
+
+    // Insert new comments
+    for (let i = 0; i < results.length; i++) {
+      const result = results[i];
+      const domLine = codeLines[i];
+      const prettyResult = await formatResult(result);
```
Shinigami92: **nitpick:** 
```suggestion
      const prettyResult = formatResult(result);
```

docs/.vitepress/components/api-docs/refresh-button.vue
```diff
@@ -0,0 +1,68 @@
...
+  spinning.value = true;
+  await Promise.all([refresh(), delay(100)]);
+  spinning.value = false;
+}
+
+// Extra delay to make the spinning effect more visible
+// Some examples barely/don't change, so the spinning is the only visible effect
+function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
Shinigami92: **nitpick:** 
```suggestion
import { useTimeoutFn } from '@vueuse/core';
import { ref } from 'vue';

const emit = defineEmits<{
  refresh: [];
}>();

const spinning = ref(false);

async function onRefresh(): Promise<void> {
  if (spinning.value) {
    return;
  }

  spinning.value = true;
  emit('refresh');

  // Extra delay to make the spinning effect more visible
  // Some examples barely/don't change, so the spinning is the only visible effect
  useTimeoutFn(() => {
    spinning.value = false;
  }, 100);
}
```

docs/.vitepress/components/api-docs/method.vue
ST-DDT: Note to self: This follows the vue syntax similar to `@click` on the html button itself.

docs/.vitepress/components/api-docs/refresh-button.vue
ST-DDT: This breaks the functionality of the button, as it no longer waits for the refresh to complete.

docs/.vitepress/components/api-docs/method.vue
ST-DDT: This suggestion is blocked by: https://github.com/faker-js/faker/pull/3301/files#r1894602053

ST-DDT:
Ready for review.

Shinigami92:
I don't know what is happening on my local machine, but the CSS animation does not get triggered 😕 
However when I try the code as an example in JSFiddle, there it works?!
I currently have `Chrome Version 131.0.6778.205 (Official Build) (64-Bit)` and also `Firefox 133.0.3 (64-Bit)`
`Windows Version 23H2 (Build 22631.4602)`

On my iOS iPhone the animation works 👀 That is very confusing...

docs/.vitepress/components/api-docs/refresh-button.vue
Shinigami92: As discussed in Voice Chat, maybe for now lets just add a dev-comment at line 4 (`defineProps`) that this is not suggested Vue best practice, but just works for now.  
And maybe if someone finds the time we can tackle and improve later, but get this PR into production.

Also because of the non working animations on my localhost, I cant debug it right now 🙁

ST-DDT: Created a comment

Shinigami92 (approved):

PR #3314, opened 2024-12-03
chore: align variable naming style of internet user agent

ST-DDT:
Preparation for #2439

- #2439

Related:

- #3312
- #3313
- #3314
- #3315
- #3316
- #3317
- #3318
- #3319

---

Most of our code uses camelCase variables instead of snake_case.

This PR changes the `internet/user-agent.ts` to follow the same style.

---

This change is split from the no abbreviation lint PR, to reduce diff and I consider this slightly more valid than most of the other changes required for that.

PR #3315, opened 2024-12-03
chore: improve variable naming of system module

ST-DDT:
Preparation for #2439

- #2439

Related:

- #3312
- #3313
- #3314
- #3315
- #3316
- #3317
- #3318
- #3319

---

Improves the variable naming in the `system` module.
Feel free to suggest better names.

---

This change is split from the no abbreviation lint PR, to reduce diff and I consider this slightly more valid than most of the other changes required for that.

### DIFF
--- a/src/modules/system/index.ts
+++ b/src/modules/system/index.ts
@@ -77,21 +77,21 @@ export class SystemModule extends ModuleBase {
       .toLowerCase()
       .replaceAll(/\W/g, '_');
 
-    const extensionsStr = this.faker.helpers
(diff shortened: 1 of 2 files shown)


matthewmayer (approved):

xDivisionByZerox (approved):
While I don't have a problem with the change suggested, I also did not have any problems with the previous source code. This to me feels like making a change for the sake of changing the codebase. I don't want to block so I'll still approve.

ejcheng (approved):

PR #3318, opened 2024-12-03
chore: improve parameter naming of keys.ts

ST-DDT:
Preparation for #2439

- #2439

Related:

- #3312
- #3313
- #3314
- #3315
- #3316
- #3317
- #3318
- #3319

---

Improves the variable naming in the `internet` module.

---

This change is split from the no abbreviation lint PR, to reduce diff and I consider this slightly more valid than most of the other changes required for that.

### DIFF
--- a/src/internal/keys.ts
+++ b/src/internal/keys.ts
@@ -6,8 +6,8 @@
  *
  * @internal
  *
- * @param obj The object to get the keys of.
+ * @param object The object to get the keys of.
  */
-export function keys<T extends object>(obj: T): Array<keyof T> {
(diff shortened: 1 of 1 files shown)


xDivisionByZerox:
I'm not sure about this change since I don't see any advantage in this parameter name change. Just like `i` is an acceptable abbreviated variable name for counting in itterations, `obj` is perfectly fine for arguments that represent a gerneric object IMO.

If we would like to completly activate the the `prevent-abbreviations` rule I'd propably not argue. [But sice we are not](https://github.com/faker-js/faker/pull/3320), I don't see a reason to change this part of our code.

PR #3327, opened 2024-12-05
refactor(locale): fix truncated song names with commas

matthewmayer:
likely accidentally lost when originally imported

ref #996 and https://manghammath.com/80s%20Hits/The%20Top%201000%20Songs%20of%20AllTime.pdf


ST-DDT (approved):
Why doesnt this cause snapshot updates? Are we just lucky?

matthewmayer:
> Why doesnt this cause snapshot updates? Are we just lucky?

yep, its mostly just slight renames so didn't affect the snapshotted values

xDivisionByZerox (approved):

PR #3329, opened 2024-12-05
refactor(locale): fix various locale data with trailing spaces

matthewmayer:
Remove data which incorrectly ends with a trailing space.

in the case of ar/location/state and hr/location/street_name, this was also a duplicate of a line without a space, so removed.


matthewmayer:
although this breaks the rule of keeping PRs to a single module, i think its quick to review as a batch.

ST-DDT (approved):
Should we add this to our normalize data script?

matthewmayer:
> Should we add this to our normalize data script?

There are a few cases where trailing or leading spaces seem intentional such as some prefixes and suffixes. 

So perhaps not.

src/locales/ar/location/state.ts
```diff
@@ -20,7 +20,6 @@ export default [
   'عرابة',
   'رام الله',
   'طولكرم',
-  'بئر السبع ',
```
xDivisionByZerox, line 23: Why did you remove this entry entirly instead of stripping the reailing space?

ST-DDT: AFAICT it is already present in line 6.

https://github.com/faker-js/faker/blob/c9284dc210b78ef289eb1967bd1e6d76be0e27aa/src/locales/ar/location/state.ts#L6

xDivisionByZerox: Allright, I did not see this, since GH only show 3 more lines around the change area. Thanks for the clearification 👍

xDivisionByZerox (approved):

Issue #2305, opened 2023-08-10
Refine `country_code` definitions in all locales to ensure only relevant data is included


ejcheng:
Our `country_code` definitions are inconsistent at this time. We should consider going through each locale's `country_code` definitions and ensure that the only entries in that locale are countries that are relevant (proximity, etc) to the selected locale. Also, we would need to refine our JSDocs to explain this.

This issue is an expanded form of #1861.

- de_CH #2304
- fr_CH- doesn't look like any changes are needed

xDivisionByZerox:
**Team decision:**

We generally want this.
More generic locales (`en`, `de`) should contain a wider range of relevant contry codes. The more specific a locale gets, the more specific the country codes should be as well.

All locale groups changes should be provided in a separate PR to allow for more fine grained research on those locales.

xDivisionByZerox:
I checked all locales and their implementation looks fine. To check them out yourself you can use the following reproduction:

1. Go to our website: https://fakerjs.dev/
2. Open the dev console
3. Use the following script to get all country codes by locale:
   ```js
   await enableFaker();
   Object.entries(allLocales).reduce((result, [localeName, localeData]) => {
     if(localeData.location?.country_code) {
       result[localeName] = localeData.location?.country_code
     }
     return result;
   }, {});
   ```

As of Faker v9.8.0 you will receive the following result:

<details><summary>result.json</summary>

```json
{
  "base": [
    {
      "alpha2": "AD",
      "alpha3": "AND",
      "numeric": "020"
    },
    {
      "alpha2": "AE",
      "alpha3": "ARE",
      "numeric": "784"
    },
    {
      "alpha2": "AF",
      "alpha3": "AFG",
      "numeric": "004"
    },
    {
      "alpha2": "AG",
      "alpha3": "ATG",
      "numeric": "028"
    },
    {
      "alpha2": "AI",
      "alpha3": "AIA",
      "numeric": "660"
    },
    {
      "alpha2": "AL",
      "alpha3": "ALB",
      "numeric": "008"
    },
    {
      "alpha2": "AM",
      "alpha3": "ARM",
      "numeric": "051"
    },
    {
      "alpha2": "AO",
      "alpha3": "AGO",
      "numeric": "024"
    },
    {
      "alpha2": "AQ",
      "alpha3": "ATA",
      "numeric": "010"
    },
    {
      "alpha2": "AR",
      "alpha3": "ARG",
      "numeric": "032"
    },
    {
[...]

PR #3333, opened 2024-12-10
feat(location): add list of spoken languages

### LINKED ISSUE #1548: Add language

I am looking to add random languages for a field "Spoken Languages"

### Suggested solution

Under faker.name please add language, like spoken language (e.g. English, Chinese, Spanish)

or

in faker.random.locale() you could add the option to return language name instead of the 2 letter code only

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #2622: Add method to return random languages (ISO-639 data)

I am currently working on a project where I would like to return a random ISO-639 code in mock data. It would be great for faker to support this internally similar to `location.countryCode`

### Suggested solution

I'm not sure which module it may fit in well today, or if a new module may make the most sense, but my proposal would be to add a `language` or `languageCode` method within faker. The method would take as input a `variant` based on the ISO-639 standard, defaulting to ISO-639-1 (2 character) codes when called without parameters.

Link to the standard: https://www.loc.gov/standards/iso639-2/php/code_list.php

The variants I propose to be implemented are:
1. `639-1` - to return `639-1` 2 character codes
2. `639-2` - to return `639-2` 3 character codes

The additional sets - `639-3`, `639-4`, `639-5` - may also be implemented. Similarly, the English, French, and German language names as defined in the standard could be added as variants.

Alternative to the specific sets, an `alpha-2` and `alpha-3` variant approach could be taken where `alpha-2` would be the 639-1 code set and `alpha-3` may be the 639-2 set.

### Alternative

_No response_

### Additional context

_No response_

UmairJibran:
closes #1548
closes #2622

For Ref:
- [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
- [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
- [ISO 639-2 Language Code List](https://www.loc.gov/standards/iso639-2/php/code_list.php)

Languages added for
- English [en]
- Urdu [ur]


matthewmayer:
I think based on discussion at https://github.com/faker-js/faker/issues/2622 we wanted this to return an object with not just the language name but also the iso 2 and 3 letter codes where available.

UmairJibran:
Thanks @matthewmayer I didn't read that discussion, I'll update it later tonight

UmairJibran:
It's ready for review

src/modules/location/index.ts
```diff
@@ -628,4 +628,22 @@ export class LocationModule extends ModuleBase {
...
+   * @see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
+   * @see [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
+   * @see [ISO 639-2 Language Code List](https://www.loc.gov/standards/iso639-2/php/code_list.php)
+   *
+   * @example
+   * faker.location.language() // { alpha2: 'en', alpha3: 'eng', name: 'English' }
+   *
+   * @since 9.3.0
+   */
+  language(): { alpha2: string; alpha3: string; name: string } {
```
ST-DDT: In other methods like these we have used and exported a dedicated type for this, to make it easier to use/reference in user code:

https://github.com/faker-js/faker/blob/8b2976f3479db445a2a7b66606110009495478f4/src/modules/finance/index.ts#L461
https://github.com/faker-js/faker/blob/8b2976f3479db445a2a7b66606110009495478f4/src/modules/finance/index.ts#L15

src/modules/location/index.ts
```diff
@@ -628,4 +628,22 @@ export class LocationModule extends ModuleBase {
...
+   * Returns a random Spoken Language.
+   *
+   * @see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
+   * @see [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
+   * @see [ISO 639-2 Language Code List](https://www.loc.gov/standards/iso639-2/php/code_list.php)
+   *
+   * @example
+   * faker.location.language() // { alpha2: 'en', alpha3: 'eng', name: 'English' }
+   *
+   * @since 9.3.0
```
ST-DDT: Faker v9.3.0 is already released, so it must be `v9.4.0` (or later).

src/modules/location/index.ts
```diff
@@ -1,6 +1,26 @@
...
+ * Represents a language with its full name, 2 character ISO 639-1 code, and 3 character ISO 639-2 code.
+ */
+export interface Language {
+  /**
+   * The full name for the Language (e.g. `English`).
+   */
+  name: string;
+
+  /**
+   * The 2 character ISO 639-1 code.
```
```suggestion
   * The 2 character [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) code.
```

src/modules/location/index.ts
```diff
@@ -1,6 +1,26 @@
...
+   */
+  alpha2: string;
+
+  /**
+   * The 3 character ISO 639-2 code.
```
```suggestion
   * The 3 character [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2) code.
```

src/definitions/location.ts
```diff
@@ -147,4 +148,14 @@ export type LocationDefinition = LocaleEntry<{
    * @see [IANA Time Zone Database](https://www.iana.org/time-zones)
    */
   time_zone: string[];
+
+  /**
+   * A list of languages.
+   * Each language has a name, an ISO 639-1 code, and an ISO 639-2 code.
+   *
+   * @see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
+   * @see [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
```
ST-DDT: IMO these are better suited on the properties.

```suggestion
```

src/definitions/location.ts
```diff
@@ -147,4 +148,14 @@ export type LocationDefinition = LocaleEntry<{
    * @see [IANA Time Zone Database](https://www.iana.org/time-zones)
    */
   time_zone: string[];
+
+  /**
+   * A list of languages.
+   * Each language has a name, an ISO 639-1 code, and an ISO 639-2 code.
```
ST-DDT: Optional: IMO this isn't needed here, what do the others think?

```suggestion
```

src/definitions/location.ts
```diff
@@ -147,4 +148,14 @@ export type LocationDefinition = LocaleEntry<{
    * @see [IANA Time Zone Database](https://www.iana.org/time-zones)
    */
   time_zone: string[];
+
+  /**
+   * A list of languages.
```
ST-DDT: Should we use any languages here or just the ones spoken in the locale?

```suggestion
   * A list of spoken languages.
```

src/modules/location/index.ts
```diff
@@ -1,6 +1,26 @@
 import { FakerError } from '../../errors/faker-error';
 import { ModuleBase } from '../../internal/module-base';
 
+/**
+ * Represents a language with its full name, 2 character ISO 639-1 code, and 3 character ISO 639-2 code.
+ */
+export interface Language {
+  /**
+   * The full name for the Language (e.g. `English`).
```
```suggestion
   * The full name for the language (e.g. `English`).
```

src/modules/location/index.ts
```diff
@@ -628,4 +648,22 @@ export class LocationModule extends ModuleBase {
       this.faker.definitions.location.time_zone
     );
   }
+
+  /**
+   * Returns a random Spoken Language.
```
```suggestion
   * Returns a random spoken language.
```

ST-DDT (commented):
Thanks for your quick responses.
I have found a few more things to potentially improve this.

My biggest question mark regarding this feature is which values the method is supposed to return.

- The languages commonly known/spoken by/relevant for the populance of a locale
- "All" languages regardless of applicability for the locale (current)

(This is relevant for the countryCode method as well #2305)

- #2305

matthewmayer:
I'd say all languages.

test/modules/location.spec.ts
```diff
@@ -415,6 +417,17 @@ describe('location', () => {
           expect(faker.definitions.location.time_zone).toContain(actual);
         });
       });
+
+      describe('language()', () => {
+        it('should return a random language', () => {
+          const actual = faker.location.language();
+          expect(actual.name).toBeTruthy();
+          expect(actual.alpha2).toBeTruthy();
```
matthewmayer, line 425: Maybe we can check it has 2 characters?

src/modules/location/index.ts
```diff
@@ -628,4 +648,22 @@ export class LocationModule extends ModuleBase {
...
+   *
+   * @see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
+   * @see [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
+   * @see [ISO 639-2 Language Code List](https://www.loc.gov/standards/iso639-2/php/code_list.php)
+   *
+   * @example
+   * faker.location.language() // { alpha2: 'en', alpha3: 'eng', name: 'English' }
```
matthewmayer: Can we add an example that drills down to the sub items like

faker.location.language().name

ST-DDT: Maybe one additional for each property.

matthewmayer:
I think we should include at least all languages faker is localised into. I think the following are missing. 

1.	Afrikaans (af)
	2.	Azerbaijani (az)
	3.	Maldivian (dv)
	4.	Farsi/Persian (fa)
	5.	Latvian (lv)
	6.	Indonesian (id)
	7.	Nepali (ne)
	8.	Thai (th)
	9.	Uzbek (uz)
	10.	Yoruba (yo)

src/definitions/location.ts
UmairJibran: I will leave this one out for a while, I thought it would be better suited to include it. We can remove it at the very end.

UmairJibran:
I have pushed the latest changes.

@ST-DDT @matthewmayer thanks a lot guys for going through the changes in smaller iterations!

for #2305 I'm not sure how to approach it, do I hard coded which `language` and `countryCode` goes where? I could work on that but I would want to take it up in a different PR.

ST-DDT (dismissed):

matthewmayer:
> I have pushed the latest changes.
> 
> @ST-DDT @matthewmayer thanks a lot guys for going through the changes in smaller iterations!
> 
> for #2305 I'm not sure how to approach it, do I hard coded which `language` and `countryCode` goes where? I could work on that but I would want to take it up in a different PR.

yeah dont worry about #2305 for now that would certainly be a different PR if we decide to do it

src/modules/location/index.ts
```diff
@@ -628,4 +648,25 @@ export class LocationModule extends ModuleBase {
...
+
+  /**
+   * Returns a random spoken language.
+   *
+   * @see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)
+   * @see [ISO 639-2](https://en.wikipedia.org/wiki/ISO_639-2)
+   * @see [ISO 639-2 Language Code List](https://www.loc.gov/standards/iso639-2/php/code_list.php)
+   *
+   * @example
+   * faker.location.language() // { alpha2: 'en', alpha3: 'eng', name: 'English' }
```
matthewmayer: i think we should consider NOT having the default example data be English/en/eng. For example if the example was German/de then it would be more informative - it hints to you that for example that the names of the languages are in the name of the current locale (ie it's "German" not "Deutsch").

UmairJibran: yeah that makes more sense, It will also show that for ISO 639-2 we are using the **terminology** code and not the **bibliographic** code.

For the example of German language:


| English Name | Localized Name | ISO 639-2 | ISO 639-3 (b) | ISO 639-3(t) |
|---|---|---|---|---|
| German | Deutsch | de | ger | deu |

And the object for German would be
```JSON
{
    "name": "German",
    "alpha2": "de",
    "alpha3": "deu"
}
```
and not
```JSON
{
    "name": "German",
    "alpha2": "de",
    "alpha3": "ger"
}
```

matthewmayer (approved):

ST-DDT (approved):

PR #3336, opened 2024-12-14
fix(finance): correct Discover card number format

### LINKED ISSUE #2350: Verify all credit card issuer patterns are valid with validator.js

julsql:
Partially fixes https://github.com/faker-js/faker/issues/2350

This PR addresses an issue with the Discover card number format in the finance module. The previous implementation allowed invalid card number formats that did not align with Discover's official specifications. Discover card number can't have 20-digit card numbers.
https://www.discoverglobalnetwork.com/content/dam/discover/en_us/dgn/docs/IPP-VAR-Enabler-Compliance.pdf

Changes Introduced:

- Removed incorrect 20-digit Discover card numbers.

No breaking changes are introduced.


ST-DDT:
Have you checked the other locales as well?

julsql:
> Have you checked the other locales as well?

You're right, I've missed one. I've just added it.

ST-DDT:
Could you please link the official spec for the discover card numbers for easier verification of your changes?

ST-DDT:
If possible, please also add a test that uses validatorjs' isCreditcard with provider discover.

https://github.com/faker-js/faker/blob/5436111712afc292004216f25d5e3eec0705cc56/test/modules/finance.spec.ts#L478-L480

https://github.com/validatorjs/validator.js/blob/86911d817fbcd177c295b13d7ca0fdff87341191/src/lib/isCreditCard.js#L7

Is the validation pattern there correct?

matthewmayer:
> If possible, please also add a test that uses validatorjs' isCreditcard with provider discover.
> 
> https://github.com/faker-js/faker/blob/5436111712afc292004216f25d5e3eec0705cc56/test/modules/finance.spec.ts#L478-L480
> 
> https://github.com/validatorjs/validator.js/blob/86911d817fbcd177c295b13d7ca0fdff87341191/src/lib/isCreditCard.js#L7
> 
> Is the validation pattern there correct?

When i previously tested this validator.js incorrectly disallows the `64[4-9]#-####-####-###L` pattern

https://github.com/faker-js/faker/issues/2350#issuecomment-1710276802

julsql:
> > If possible, please also add a test that uses validatorjs' isCreditcard with provider discover.
> > https://github.com/faker-js/faker/blob/5436111712afc292004216f25d5e3eec0705cc56/test/modules/finance.spec.ts#L478-L480
> > 
> > https://github.com/validatorjs/validator.js/blob/86911d817fbcd177c295b13d7ca0fdff87341191/src/lib/isCreditCard.js#L7
> > Is the validation pattern there correct?
> 
> When i previously tested this validator.js incorrectly disallows the `64[4-9]#-####-####-###L` pattern
> 
> [#2350 (comment)](https://github.com/faker-js/faker/issues/2350#issuecomment-1710276802)

I've tried with validator.js but indeed it throws an error for the `64[4-9]#-####-####-###L` pattern. After checking on the web the pattern seems correct (https://neapay.com/post/bin-list-range-for-mastercard-visa-amex-diners-discover-jcb-cup_94.html)

Should I add a test using validatorjs' isCreditcard and remove the `64[4-9]#-####-####-###L` pattern as well? Or I leave it that way and wait for validator.js to be corrected?

ST-DDT:
> Or I leave it that way and wait for validator.js to be corrected?

We can omit the check for now and add it once validatorjs fixes their pattern.
Is there a PR for that fix in validatorjs?

julsql:
> We can omit the check for now and add it once validatorjs fixes their pattern. Is there a PR for that fix in validatorjs?

I looked for an existing PR in validatorjs addressing this issue, but I couldn't find any.

ST-DDT:
> looked for an existing PR in validatorjs addressing this issue, but I couldn't find any.

If you would like to, you could create it when you have time.

ST-DDT (approved):

matthewmayer:
> > looked for an existing PR in validatorjs addressing this issue, but I couldn't find any.
> 
> 
> 
> If you would like to, you could create it when you have time.

I'm not sure the official info from Discover provides enough info for a canonical validation regex. It shows the correct prefixes but not what lengths are valid for each prefix?

julsql:
> I'm not sure the official info from Discover provides enough info for a canonical validation regex. It shows the correct prefixes but not what lengths are valid for each prefix?

Based on the official documentation from Discover, we know that the valid lengths for cards with the prefix 644-9 are between 16 and 19 digits. I think it should be enough to build the regex in validator.js? I will open an issue when I will have time.

matthewmayer:
Right I mean it's fine if faker.js only generates a small subset of valid Discover card numbers.  But validator.js should ideally validate all valid card numbers and I'm not sure we have detailed enough info to create a PR on validator.js to do that.

matthewmayer (approved):

PR #3343, opened 2024-12-28
docs: use embedded build for preview examples

### LINKED ISSUE #3285: `internet.jwt()` fails on documentation website
### Pre-Checks

### Describe the bug

When running `faker.internet.jwt()`  on our documentation site, you receive an runtime error:

```js
+esm:13 Uncaught TypeError: Unknown encoding: base64url
    at Uint8Array.Yi (+esm:13:5072)
    at Uint8Array.a [as toString] (+esm:13:5949)
    at ir (chunk-3QROWZCZ.js:1:5396)
    at kr.jwt (chunk-3QROWZCZ.js:1:18786)
    at <anonymous>:1:16
```

This might affect other browser environments as well (at least the ones using the same CDN).

### Minimal reproduction code

Go to https://fakerjs.dev and open the dev tools.

Run the following:

```js
await enableFaker();
faker.internet.jwt();
```

### Additional Context

Our [check for the existance of `Buffer`](https://github.com/faker-js/faker/blob/b390432626811d2af32086b16151bc4c55634c60/src/internal/base64.ts#L42) fails because the file (https://cdn.jsdelivr.net/npm/@faker-js/faker@9.2.0/+esm) loaded from the cdn includes a Buffer polyfill:

image

### Environment Info

```shell
Tested in Microsoft Edge - Version 131.0.2903.63 (Offizielles Build) (64-Bit)
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

### LINKED ISSUE #3342: Refresh examples button doesn't work for unreleased methods
Noted at https://github.com/faker-js/faker/pull/3301#issuecomment-2552541517

For example the refresh button on next.fakerjs.dev for faker.location.language() doesn't work yet as the docs are for 9.4.0 but the JS pulled in is for 9.3.0

It would be nice if this could work to allow for easier testing of the examples for new methods.

ST-DDT:
Fixes #3342, https://github.com/faker-js/faker/pull/3301#issuecomment-2552541517

- #3342
- https://github.com/faker-js/faker/pull/3301#issuecomment-2552541517
- #3285

---

This bundles the project into a single file and makes it available to the browser console for `next` and PRs.
To keep traffic low on our website and help counting the usage of the versions, the release sites keep using the jsdelivr CDN links.

You can test it using the new [language](https://deploy-preview-3343.fakerjs.dev/api/location.html#language) method.

grafik

grafik
grafik


package.json
```diff
@@ -10,10 +10,11 @@
     "generate": "run-s generate:locales generate:api-docs",
     "generate:api-docs": "tsx ./scripts/apidocs.ts",
     "generate:locales": "tsx ./scripts/generate-locales.ts",
-    "docs:build": "run-s generate:api-docs docs:build:run",
+    "docs:build": "run-s generate:api-docs docs:build:embedded docs:build:run",
+    "docs:build:embedded": "tsup-node --entry.faker src/index.ts --format esm --outDir docs/public --no-dts --no-splitting  --no-clean",
```
ST-DDT: Is `embedded` the right term here? Any other suggestions?

tsup.config.ts
```diff
@@ -14,4 +14,7 @@ export default defineConfig({
   minify: true,
   sourcemap: false,
   splitting: true,
+  esbuildOptions(options) {
+    options.charset = 'utf8'; // Prevent Unicode escaping
```
ST-DDT: This reduces the size from the (docs) shipped `faker.js` file from 20MB to 2.7MB (due to not using unicode escape sequences!? Not sure whether there are other things at play as well).
Does that break anything for our users (in our normal builds)?
Or is there a better way to do this?

For comparison:

| Branch | `/dist` | `/docs/public/faker.js` |
| :---: | :---: | :---: |
| `next` | 8.6 MB | 20 MB |
| `this` | 6.6 MB | 2.7 MB |

matthewmayer:
This appears to fix `jwt()` too - although I'm not sure why it fails as that was added in 9.1.0... https://next.fakerjs.dev/api/internet.html#jwt

image

ST-DDT:
Thats caused by the bad Buffer polyfil. Not sure where we discussed that.

EDIT:

- #3285

package.json
xDivisionByZerox: Seams correct enought. Since this script manipulates the [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) the faker lib is loaded from you could also embrace the web standart: "docs:build:same-origin" or "docs:build:no-cors". I don't thinkk it matters to much tho.

tsup.config.ts
xDivisionByZerox: I'd say we make this change and see if anyone complains? I currently do not see how this would affect our users. Maybe do this change separatly? That way its not hidden in a seemingly unrelated PR?

xDivisionByZerox (commented):
Could you extend the [CONTRIBUTING page] with the new script as well?

ST-DDT:
> Could you extend the [CONTRIBUTING page] with the new script as well?

The other doc scripts aren't listed there either and IMO the command isn't supposed to be called by anybody directly.

package.json
ST-DDT: 🤔

tsup.config.ts
ST-DDT: Sounds like a plan.

@Shinigami92 You know more that I regarding building JS, are you aware of any issues related to this?
And why has it such a huge impact on size only for the embedded code?
Any ideas?

Shinigami92: > Sounds like a plan.
> 
> @Shinigami92 You know more that I regarding building JS, are you aware of any issues related to this? And why has it such a huge impact on size only for the embedded code? Any ideas?

You sure it does not come from the `--no-splitting`? Sounds weird to me 🤷

ST-DDT: Thanks for the input.

I removed the no-splitting option, as it doesn't have an impact on the result anyway.
I also removed the charset option, as it no longer changes the resulting file size as severely.
2.29MB vs 3.66MB.

I asked chatgpt for potential issues with the option, and while it couldn't name any actual issues, it still strongly recommended sticking to the default.
So maybe we shouldn't change the option outside of breaking changes/major versions.

ST-DDT:
Could someone please review this? I would like to do a release of Faker as part of a presentation on Wednesday 2025-01-15.

xDivisionByZerox (approved):

Issue #3335, opened 2024-12-14
Consistent formatting of examples (return values)


ST-DDT:
Continuation of #3287

- #3287 

---

The examples are inconsistently formatted, both in our source code and when running the refresh on the website.

- string without quotes:
  https://github.com/faker-js/faker/blob/00be2e195ef2944a87ec4e7c1a073e2eaf86c8ac/src/modules/finance/index.ts#L126-L128
- single line string newline:
  https://github.com/faker-js/faker/blob/00be2e195ef2944a87ec4e7c1a073e2eaf86c8ac/src/modules/finance/index.ts#L971-L972
- single line string inline:
  https://github.com/faker-js/faker/blob/3c7abb55e68fcbcf41560539a15845e7c8882765/src/modules/food/index.ts#L43
- array without leading/ending space: 
  https://github.com/faker-js/faker/blob/3c7abb55e68fcbcf41560539a15845e7c8882765/src/modules/helpers/index.ts#L821
- array with leading/ending space:
  https://github.com/faker-js/faker/blob/3c7abb55e68fcbcf41560539a15845e7c8882765/src/modules/helpers/index.ts#L619

Also some examples seem to refer to old parameter defaults or locale data as well.

We should define a function that formats our examples consistently.
Potentially automatically via script.

ST-DDT:
Blocked by #3301 

- #3301

ST-DDT:
What is (in your opinion) the expected formatting for:

- Objects:
  1. `{ 'name': 'French', 'alpha2': 'fr', 'alpha3': 'fra' }`
  2. `{ name: 'French', alpha2: 'fr', alpha3: 'fra' }`
  3. `{name: 'French', alpha2: 'fr', alpha3: 'fra'}`
  4. `{\n  name: 'French',\n  alpha2: 'fr',\n  alpha3: 'fra'\n}`
  5. other
- Arrays:
  1. `[ 'b', 'c', 'a' ]`
  2. `['b', 'c', 'a']`
  3. `[\n  'b',\n  'c',\n  'a'\n]`
  4. other

ST-DDT:
How should it be implemented:

1. via [`prettier`](https://github.com/faker-js/faker/blob/a7d7177d34d98cb11dd2e22eda99f7b5be3e67a0/docs/.vitepress/components/api-docs/format.ts) (asynchronously downloaded on demand)
[...]

PR #3348, opened 2024-12-28
docs: fix examples which return string/number

matthewmayer:
Partial manual fix for #3335 for clarifying the return type of examples
- unquoted strings
- incorrectly quoted numbers
- doubled up examples

### DIFF
--- a/src/modules/date/index.ts
+++ b/src/modules/date/index.ts
@@ -192,12 +192,12 @@ export class SimpleDateModule extends SimpleModuleBase {
    * @example
    * faker.date.betweens({ from: '2020-01-01T00:00:00.000Z', to: '2030-01-01T00:00:00.000Z' })
    * // [
-   * //   2022-07-02T06:00:00.000Z,
-   * //   2024-12-31T12:00:00.000Z,
-   * //   2027-07-02T18:00:00.000Z
+   * //   '2022-07-02T06:00:00.000Z',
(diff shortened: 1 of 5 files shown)


ST-DDT (approved):

src/modules/date/index.ts
```diff
@@ -192,12 +192,12 @@ export class SimpleDateModule extends SimpleModuleBase {
    * @example
    * faker.date.betweens({ from: '2020-01-01T00:00:00.000Z', to: '2030-01-01T00:00:00.000Z' })
    * // [
-   * //   2022-07-02T06:00:00.000Z,
-   * //   2024-12-31T12:00:00.000Z,
-   * //   2027-07-02T18:00:00.000Z
+   * //   '2022-07-02T06:00:00.000Z',
```
ST-DDT, line 195: Note(non-blocking): This changes the value to how the refresh button shows them (currently), but makes it hard to detect, that these are actual `Date` objects.
Approved for now, for consistency, but will change again soon.

See also: 

- https://github.com/faker-js/faker/issues/3335#issuecomment-2564401247

matthewmayer: agree, Dates should probably be unquoted like in the REPL

ST-DDT:
I'll merge this early so that it doesn't conflict with the example formatting PR and so I have the latest improvements available.

Issue #1862, opened 2023-02-20
Adding Random probability distribution function


maxime4000:
### Clear and concise description of the problem

So I'm seeding a database with faker. I have field that allow array of some type. I want to generate multiple array, but with different size. Some where the array is empty, some where the array has 1 elements and some where the array has multiple elements.

Most of the case will have one element in the array, but I also want to test limit case, so having a way to generate Random distributed data would be nice. 

```ts
const isEmpty = faker.datatype.boolean(); // ~50%
const isOneElement = faker.datatype.boolean(); // ~25%
const length = faker.datatype.number(100); // ~25%
const array = 
isEmpty ? []
        : isOneElement 
          ? [getFakerFunction(field)] 
          : Array.from({length}, () => getFakerFunction(field));
return array
```

Let's said that I'm faking an array of value and I want some length to be more common than others. It's common to have an array of length 1 to 3 but it's very rare to have an array of 100. I would like to have a random probability distribution function for this.

### Suggested solution

In my case, I'm looking for a random exponential distribution. 
- Length 1 has 40% chance to happen,
- Length 2 => 30%
- Length 3 => 20%
- and so on...

The function would accept an argument like this:
```ts
type ExponentialDistributionOptions = {
	min?: number;
	max?: number;
	precision?:number;
	curveSettings: {
		deviation?:number;
		mean?: number;
		// ...
	}
}
```
And would generate a number using the distribution called. 
I would expect to call faker.random.exponentialDistribution({min: 0, max: 100, curveSettings: {...}}) and the number generated from this would have more chance to be closer to 0 than closer to 100. On a scale of 1000 random value generated, we could see few value with a number close to 100. 

I wouldn't limit the feature to only exponential distribution, I would also add gaussian distribution, Rayleigh distribution, gamma distribution, etc... 

### Additional context

I'm not sure if what I'm asking is out of scope for faker, but at the same time, faker is generating data from a random value. Why would faker couldn't generate number base on some probability of that number to be generated? 

Btw, I'm no mathematician, so I might be incorrect with what I explain, but I still think faker could add some random probability distribution function.

ST-DDT:
Do you refer to something like this?

````ts
function exponentialDistributionNumber(start = 1, stepScale = 2, stepProbability = 0.5, limit = Number.MAX_SAFE_INTEGER) {
    let max = start;
    while(faker.datatype.boolean(stepProbability) && max < limit) {
        max *= stepScale;
    }
[...]

PR #3349, opened 2024-12-28
fix(system): semver parts should not be limited to 0-9

matthewmayer:
Semver components shouldnt be limited to one digit

Assuming all semvers are in the form x.y.z with single digits is likely to cause bugs with sorting etc

This PR allows y and z to go up to 20 for more variety, while keeping the numbers realistic


ST-DDT (approved):
IMO versions numbers are a perfect use case for numbers with exponentially decreasing probability for higher version numbers.

- https://github.com/faker-js/faker/issues/1862

xDivisionByZerox (approved):

PR #3363, opened 2025-01-03
fix(number): bigint shouldn't throw when max is negative

soc221b:


xDivisionByZerox:
I confirmed that this is a bug in the current version (v9.3.0) of Faker. Adding the acceped label.

ST-DDT:
I'm not sure about it.
Afaict all our number methods return 0-max if only max is specified. So either we have to change all methods to behave like this or the user has to specify both min and max (and maybe we have to adjust the error message accordingly).
I personally prefer the later as arbitarily designating boundaries like this feels wrong. The user supposed knows their intend better than we do, so we shouldnt guess it.

> Defaults to `0n` or `max - 999999999999999n` if `max` is negative.

Vs

> FakerError: Min not specified when using negative max option.

Or something similar.

If you just need a negative value you could just use `-faker.number.bigInt()`

matthewmayer:
> I'm not sure about it.
> 
> Afaict all our number methods return 0-max if only max is specified. So either we have to change all methods to behave like this or the user has to specify both min and max (and maybe we have to adjust the error message accordingly).
> 
> I personally prefer the later as arbitarily designating boundaries like this feels wrong. The user supposed knows their intend better than we do, so we shouldnt guess it.
> 
> 
> 
> > Defaults to `0n` or `max - 999999999999999n` if `max` is negative.
> 
> 
> 
> Vs
> 
> 
> 
> > FakerError: Min not specified when using negative max option.
> 
> 
> 
> Or something similar.
> 
> 
> 
> If you just need a negative value you could just use `-faker.number.bigInt()`

I agree

I think it's fairly intuitive what `faker.number.int({max:10})` returns

I don't think it's intuitive what 

`faker.number.int({max:-10}) ` returns

ST-DDT:
@soc221b Could you please explain your thoughts regarding this change?
(Or the original issue you ran into?)

ST-DDT:
I changed the target branch from `v9` to `next`, as `v9` is our release branch and all our development happens on the `next` branch.
(Unless we specifically back port changes to older versions)

soc221b:
Thanks for your comments about this.

---

> Could you please explain your thoughts regarding this change?

I personally think giving a negative max is a valid usage even it doesn't seem intuitive.

> the original issue you ran into?

I'm building a library to generate fake data based on faker.js and ran into this issue: https://github.com/soc221b/zod-schema-faker/blob/1.1.0/src/zod-bigint-faker.ts#L26-L29

---

I created an example of form validation using `<input max="-10">`, it works as expected:
https://stackblitz.com/edit/does-negative-max-work?file=index.html

ST-DDT:
> I'm building a library to generate fake data based on faker.js and ran into this issue: https://github.com/soc221b/zod-schema-faker/blob/1.1.0/src/zod-bigint-faker.ts#L26-L29

I've looked at your code and noticed that you also have some code for multipleOf (similar to our `faker.number.int({ multipleOf })`).
Would you like to create a feature request issue for that as well?

soc221b:
> Would you like to create a feature request issue for that as well?

Sure, I'll create another PR for that. 😃

ST-DDT:
Workaround for `faker.number.bigInt({ max: -1n})`:

````ts
-faker.number.bigInt({min: 1})
````

or with full context:

````ts
if (min == null && max != null && max < 0) {
  return -faker.number.bigInt({ min: -max });
}
````

ST-DDT:
See also: https://github.com/faker-js/faker/issues/3364

- https://github.com/faker-js/faker/issues/3364

xDivisionByZerox:
**Team Decision:**

We'd like to close this PR due to multiple reasons:

1. The desired features has an easy to use workaround provided [above](https://github.com/faker-js/faker/pull/3363#issuecomment-2569176519).
2. The feature would lead to inconsistencies in our API, specifically the methods in the `NumberModule`.
3. The feature would result in none intuitive behaviour from the changed method.

If you feel that we are misstaken in our actions, please feel free to discuss further in the PR or open another issue.

---

We would gladly accept your PR for the `multipleOf` argument in the `number.bigInt` method which you [suggested previously](https://github.com/faker-js/faker/pull/3363#issuecomment-2569489089).

PR #3371, opened 2025-01-13
refactor(locales): fix chemical element names in zh_CN

tuanzisama:
In the current locale, some simplified Chinese translations have issues. I will fix them in this PR.

BTW, is it better to translate more content? For example, aireline, commerce or others. If so, I would be interested in adding more fake data related to Simplified and Traditional Chinese.


ST-DDT (approved):
Looks like the old values were of by one.
Nicely spotted.

ST-DDT:
> BTW, is it better to translate more content? For example, aireline, commerce or others. If so, I would be interested in adding more fake data related to Simplified and Traditional Chinese.

Any additional translations are welcome. If you keep the locale additions/changes to a single category/module (airline, science, word), then this helps with reviewing.

https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#adding-new-locale-or-updating-existing-one

xDivisionByZerox (approved):
Just in case someone else need an external source to cross compare the values: https://en.wikipedia.org/wiki/Template:Periodic_table_(Chinese_names)

PR #3375, opened 2025-01-18
feat(number): add distributor functions

### LINKED ISSUE #1862: Adding Random probability distribution function

ST-DDT:
Implements #1862

- #1862

---

Adds a `distributor` option to faker.number.int and float. The distributor function can be used to influence the distribution of the generated values. E.g. uniformDistributor and exponentialDistributor.

- [Preview Distributors](https://deploy-preview-3375.fakerjs.dev/api/distributors.html)
- [Preview NumberModule](https://deploy-preview-3375.fakerjs.dev/api/number.html#int)

---

The following table shows the rough distribution of values generated using 
`exponentialDistributor({ base: x })`:

|  Result   | Base 0.1 | Base 0.5 | Base 1 | Base 2 | Base 10 |
| :-------: | -------: | -------: | -----: | -----: | ------: |
| 0.0 - 0.1 |     4.1% |     7.4% |  10.0% |  13.8% |   27.8% |
| 0.1 - 0.2 |     4.5% |     7.8% |  10.0% |  12.5% |   16.9% |
| 0.2 - 0.3 |     5.0% |     8.2% |  10.0% |  11.5% |   12.1% |
| 0.3 - 0.4 |     5.7% |     8.7% |  10.0% |  10.7% |    9.4% |
| 0.4 - 0.5 |     6.6% |     9.3% |  10.0% |  10.0% |    7.8% |
| 0.5 - 0.6 |     7.8% |     9.9% |  10.0% |   9.3% |    6.6% |
| 0.6 - 0.7 |     9.4% |    10.7% |  10.0% |   8.8% |    5.7% |
| 0.7 - 0.8 |    12.1% |    11.5% |  10.0% |   8.2% |    5.0% |
| 0.8 - 0.9 |    16.9% |    12.6% |  10.0% |   7.8% |    4.5% |
| 0.9 - 1.0 |    27.9% |    13.8% |  10.0% |   7.5% |    4.1% |

The exponential distribution is achieved by using `base ** exponent` where exponent is a random float between 0 and 1.
The result is then stretched evenly to fit to `min` and `max`.


src/modules/number/index.ts
```diff
@@ -534,4 +534,165 @@ export class NumberModule extends SimpleModuleBase {
...
+   * |   0   |     4.1% |     7.4% |  10.0% |  13.8% |   27.8% |
+   * |   1   |     4.5% |     7.8% |  10.0% |  12.5% |   16.9% |
+   * |   2   |     5.0% |     8.2% |  10.0% |  11.5% |   12.1% |
+   * |   3   |     5.7% |     8.7% |  10.0% |  10.7% |    9.4% |
+   * |   4   |     6.6% |     9.3% |  10.0% |  10.0% |    7.8% |
+   * |   5   |     7.8% |     9.9% |  10.0% |   9.3% |    6.6% |
+   * |   6   |     9.4% |    10.7% |  10.0% |   8.8% |    5.7% |
+   * |   7   |    12.1% |    11.5% |  10.0% |   8.2% |    5.0% |
+   * |   8   |    16.9% |    12.6% |  10.0% |   7.8% |    4.5% |
+   * |   9   |    27.9% |    13.8% |  10.0% |   7.5% |    4.1% |
```
ST-DDT: Should we have these tables here or in the parameter descriptions?

If we have them here we can show them on the documentation.
If we have them in the options we cannot display them.

IMO we should not have them in both places.

ST-DDT:
I'm also not sure whether `exponentialDistribution` is the correct name for the function.

matthewmayer:
I don't like calling it "exponentialDistribution " because you're not returning a distribution, you are returning a single value from that distribution.

So I'd call it exponentialValue or floatExponential or something similar to emphasise the thing that is returned is a float or a value.

src/modules/number/index.ts
```diff
@@ -534,4 +534,165 @@ export class NumberModule extends SimpleModuleBase {
 
     return result;
   }
+
+  /**
+   * Generates a random number between `min` and `max` using an exponential distribution.
```
matthewmayer:
```suggestion
   * Generates a random float between `min` and `max` using an exponential distribution.
```

Emphasise its a float NOT an int (since the tables below might mislead people if they dont notice the values have been `floor`ed)

src/modules/number/index.ts
```diff
@@ -534,4 +534,165 @@ export class NumberModule extends SimpleModuleBase {
...
+   *
+   * @throws If `base` is less than or equal to `0`.
+   * @throws If `max` is less than `min`.
+   *
+   * @example
+   * faker.number.exponentialDistribution() // 0.41928964795957224
+   * faker.number.exponentialDistribution(10) // 1.656598169056771
+   * faker.number.exponentialDistribution({ min: 10, max: 100 }) // 88.7273250669911
+   * faker.number.exponentialDistribution({ min: 0, max: 100, base: 10 }) // 6.9442760672808745
+   * faker.number.exponentialDistribution({ min: 0, max: 100, bias: 10 }) // 67.03715679154617
```
matthewmayer: i'm not sure adding a second way (bias) of achieving the same results is that useful; would prefer picking either bias or base.

src/modules/number/index.ts
ST-DDT: Would using `Math.floor(result)` as a table header solve this issue as well?
(I'm not against changing number to float)

src/modules/number/index.ts
ST-DDT: `base` is the mathematical basis that the function uses, however it is kind of hard to use because to invert the effect you have to `base = 1/base` and big base => probably low value.
Bias is intended as a non mathematical way to define the base, where if you want to invert the value you can just `bias = -bias` and big bias => probably big value.

I would prefer to keep base, because it is the actual foundation as the implementation,
and I would prefer to keep bias, due to its ease of use.
But that's just my opinion and could be changed.

ST-DDT:
> you are returning a single value from that distribution.

I'm not even sure that is the case.

> So I'd call it exponentialValue or floatExponential or something similar to emphasise the thing that is returned is a float or a value.

Good idea. I'm not finally sure which term I would like to use but stepping away from calling it `exponentialDistribution` is the correct choice.

I'm currently considering one of these:

- exponential
- exponentialValue
- exponentialNumber
- exponentialFloat
- floatExponential
- exponentiallySpread (or something similar)
- powerLaw___ (<- ChatGpt)

src/modules/number/index.ts
matthewmayer: Or you could show ranges in the table like "0 to 1" or "0  <= x < 1"?

ST-DDT: Done.

src/modules/number/index.ts
```diff
@@ -534,4 +534,165 @@ export class NumberModule extends SimpleModuleBase {
...
+   *
+   * @since 9.5.0
+   */
+  exponentialDistribution(
```
xDivisionByZerox: Some notes from my side (I'll leave a comment at the code to have a separate discussion block):

1. The functions seems out of place in this module. All function names so far have had a technical nature to them while this one is more mathmatically named. Following that schema `number.int` would be `number.natural` or `number.float` would be `number.rational`. Not that dramatic, but something I noticed after we tried organizing the modules in v8 so neatly.
2.  Is it correct to make a new function at all. Reading the original issue, the usecase described a distribution of integers along an exponential scale. This function however returns floats. IMO, integrating the parameters into the existing methods is the most logical option. The default could be set to return a unionized value distribution (base = 1) to not introduce breaking changes.
  2.1.  When the parameters are intrgrated into the other functions they should maybe grouped in a separate, optional object. Since I would consider this API pretty advanced for the usual user, I'd "hide" it in an extra nested layer. Additionally, I could think of faker providing some utility function or constants that have predefined, commonly used distributions:
```ts
import { faker, exponentialDistibution, linearDistibution } from '@faker-js/faker';

console.log(exponentialDistibution()); // { bias: -1, base: 10 }
console.log(exponentialDistibution({ inverted: true })); // { bias: -1, base: 0.1 }

faker.nuber.int({ distibution: exponentialDistibution() });
faker.nuber.float({ distibution: exponentialDistibution({ inverted: true }) });
faker.nuber.float({ distibution: linearDistibution() });
```

What do you think? 🤔

matthewmayer: i think there is logic in making it an option for faker.number.float() etc instead

The only downside is when the really basic functions like float() and int() have too many options it could make it overwhelming for new users to read the documentation.

src/modules/number/index.ts
ejcheng: I think we should leave the tables here so they show up in the docs.

src/modules/number/index.ts
xDivisionByZerox: I generally agree on your reasoning. Thats why I propsed an additional nested layer in the options. That way the arguments `bias` and `base` do not bleed into the first level of options. Additionally, adding some overloads to the function could work. Make one overload (the default) with less arguments in the options that are used more often and one overload with an options object that represents a superset of the default, simple one.

```ts
type SimpleOptions = { min: number; max: number };
type AdvancedOptions = SimpleOptions & { distibution: { base: number; bias: number } };

class NumberModule {
  int(): number; // no args = default args
  int(options: AdvancedOptions): number; // advanced arguments, put first to not get picked up by IDE (IDE usually picks last overload as default)
  int(options: SimpleOptions): number; // simple arguments, last to be the default
  int(options?: AdvancedOptions = {}): number { // implementation can just use optional advanced options since it includes all possible cases
    // ...
  }

  // ...
}
``` 

_The types are separated for representation purposes here, I'm not sure if our docs already allow for non inline type definitions._

ST-DDT: The `distribution: exponentialDistribution({base: 2})` option seems like a good solution.
`type Distribution = (options: {random, min, max}): number`
That way other developers can add their own distributions as needed.
I'll try it soon.

I fear `AdvancedOptions` might be complicated to implement for our api docs.
Currently we just use the last (api) signature.

https://github.com/faker-js/faker/blob/51dddb8327d2d6d8e2342405e008453cac990b44/scripts/apidocs/output/page.ts#L207

xDivisionByZerox: > That way other developers can add their own distributions as needed.

Thats the idea. Although, I'd propably provide some common use caes out of the box. Thats not a requirement for me, tho.

> I fear AdvancedOptions might be complicated to implement for our api docs.
Currently we just use the last (api) signature.

No worries. Lets try to focus on designing the best possible API for our users first. We can discuss the handling of the documentation afterwards, IMO.

ST-DDT: I might have run into a small issue here for now.

While linear and exponential/power-law distributions only require a single random value to generate. A normal distribution requires two.

So should we mark the interface as unstable for now and pass an entire faker instance in there or should we wait for v10 and then directly use `FakerCore`s?

matthewmayer: There's a ton of different probability distributions and they might all have different numbers  of parameters and traditional names for their parameters: https://en.wikipedia.org/wiki/Probability_distribution - not to mention some make sense for discrete distributions (like ints) and some for continuous distributions (like floats).

So i dont think you'll be able to have a standardised way to refer to the parameters.

xDivisionByZerox: That's a good argument. I was probably looking too naively at the problem at hand.
It would still be awesome if we could solve the problem without introducing a new method. I stated the module specification problem in my first comment.

ST-DDT: Thats why I try to implement the distribution option as a function, not data.

````ts
function int(options: { distribution: (fakerCore) => number, ....}) {
````

`f() ∈ [0, 1)`

Comment (matthewmayer) [reply src/modules/number/index.ts]: another thing that occurred to me; in many cases when generating a small finite number of different ints, developers may find it easier to just use weightedArrayElement and just manually set the percentage chance of getting each value.

So i think this is actually much more useful in the float case, when you can't easily do that.

Comment (ST-DDT) [reply src/modules/number/index.ts]: One of the usecases for this method is the regexp method: specifically for `*` lengths.

**Comment by xDivisionByZerox** [comment]
**Team decision**

The feature will be implemented by extending the already existing methods (`number.int()`, `number.float()`, etc). An additional parameter will be added. The parameter will be a function that exepts a randomizer instance and will return a biased randomized value. The return value will be used by the original function to "push" the value in the direction the distribution desires.

We are currently unsure how to name the parameter. While distribution is okay, we are not sure if this is the mathmatical correct term.
The name "spread" was discussed, but discarded since it might confuse users with the JS spread operator.
We will think about other names until the next meeting. If we do not come up with any, the current name of "distribution" is fine for us.

**Comment by ST-DDT** [comment]
Ready for review.

I changed the implementations to use a `distributor` parameter.
`distributor` is close enough to `distribution` so that you can guess what the option does, without actually running into the mathematical specification.

File: src/distributors/uniform.ts [hunk 5]
```diff
@@ -0,0 +1,41 @@
...
+ * distributor(randomizer) // 0.28282249581185814
+ * distributor(randomizer) // 0.017890944117802343
+ *
+ * @since 9.6.0
+ */
+export function uniformDistributor(): Distributor {
+  return UNIFORM_DISTRIBUTOR;
+}
+
+const UNIFORM_DISTRIBUTOR: Distributor = ({ next }) => next();
```
xDivisionByZerox, line 41: I thought quite long about it. What exactly was your reasoning to extract the uniform function implementation into it's own function expression? The only benifit I was able to see, would be function reference comparison:

```ts
const a = uniformDistributor();
const b = uniformDistributor();
a === b // true
```

Was that your reasoning? If yes, is this really desired? 🤔

ST-DDT: The value is exposed as fn because the other distributions are as well.

The value is defined as a const because it is used as default for the number functions. This might be related to my usual backend development habits.

The a===b part is unintentional.

test/modules/number.spec.ts
xDivisionByZerox, line 1: To not couple the number tests with the implementation details of our given distributor functions, we should probably use custom, "deterministic" ones in theses test cases. Additionally, we should add separate test cases for the provided distributor functions. For example: The exponential distributor is currently not being checked for a negative "base" input.

scripts/apidocs/utils/markdown.ts
```diff
@@ -49,6 +58,18 @@ function comparableSanitizedHtml(html: string): string {
     .replaceAll(' ', '');
 }
 
+/**
+ * Wraps the given code in a code block.
```
xDivisionByZerox:
```suggestion
 * Wraps the given code in a markdown code block.
```

src/distributors/distributor.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns Generates a random float between 0 (inclusive) and 1 (exclusive).
+ *
+ * @example
+ * import { Distributor, Randomizer, faker } from '@faker-js/faker';
+ *
+ * const alwaysMin: Distributor = () => 0;
+ * const uniform: Distributor = (randomizer: Randomizer) => randomizer.next();
+ *
+ * faker.number.int({ min: 2, max: 10, distributor: alwaysMin }); // 2
+ * faker.number.int({ min: 0, max: 10, distributor: uniform }); // 5
```
xDivisionByZerox, line 33: The exemplary calls should probably be repeated to better visualize the results:

```suggestion
 * const alwaysMin: Distributor = () => 0;
 * faker.number.int({ min: 2, max: 10, distributor: alwaysMin }); // 2
 * faker.number.int({ min: 2, max: 10, distributor: alwaysMin }); // 2
 * faker.number.int({ min: 2, max: 10, distributor: alwaysMin }); // 2
 
 * const uniform: Distributor = (randomizer: Randomizer) => randomizer.next();
 * faker.number.int({ min: 0, max: 10, distributor: uniform }); // 5
 * faker.number.int({ min: 0, max: 10, distributor: uniform }); // 2
 * faker.number.int({ min: 0, max: 10, distributor: uniform }); // 9
```

src/distributors/distributor.ts
```diff
@@ -0,0 +1,35 @@
+import type { Randomizer } from '../randomizer';
+
+/**
+ * A function that determines the distribution of generated values.
+ * Values generated by a randomizer are considered uniformly distributed, distributor functions can be used to change this.
+ * If many results are collected the results form a limited distribution between `0` and `1`.
+ * So an exponential distributor will values resemble a limited exponential distribution.
```
xDivisionByZerox:
```suggestion
 * So an exponential distributor's values will resemble a limited exponential distribution.
```

src/distributors/exponential.ts
```diff
@@ -0,0 +1,122 @@
+import { FakerError } from '../errors/faker-error';
+import type { Distributor } from './distributor';
+import { uniformDistributor } from './uniform';
+
+/**
+ * Creates a new function that generates power-law/exponentially distributed values.
```
xDivisionByZerox, line 6: Maybe we can add a more accessible explanation what this means as well.

```suggestion
 * Creates a new function that generates exponentially distributed values.
 * This means that values on the edges (min or max based on configuration) are more likely to be generated.
```

This needs to be put in the second signature as well.

src/distributors/exponential.ts
```diff
@@ -0,0 +1,122 @@
...
+         * The higher/more positive the `bias`, the more likely the number will be closer to the maximum value.
+         * The lower/more negative the `bias`, the more likely the number will be closer to the minimum value.
+         * Values of `0` will generate a uniform distribution.
+         * Can alternatively be configured using the `base` option.
+         *
+         * @default -1
+         */
+        bias?: number;
+      }
+): Distributor;
```
xDivisionByZerox, line 90: Was there a reason why we didn't made these two dedicated functions? 

The advantages would be that it should be more obvious to the end user that you can not pass bias AND base at the same time. Technically this is already the way, but the signatures could more detailed examples based on the input that is possible. The signature containing the bias argument could only contain table with the bias values. Same with the base. 

The disadvantage would be that we would still require a third combined signature (same as we have right now) for our documentation generation.

xDivisionByZerox (commented):
The more I'm looking at this PR, the more I'm falling in love with this feature's design. ❤️ 

I found some small things that I was noticing, while trying to get back into this topic. @ST-DDT are you still available or can I take over the PR?

ST-DDT:
> @ST-DDT are you still available or can I take over the PR?

Feel free to take over.

src/distributors/exponential.ts
ST-DDT: AFAICT this is a power law distribution and not a exponential distribution in the mathematical sense.

> This means that values on one side of the of the value range are far more likely than the values on the other side of the range (depending on configuration).

Or something similar.

src/distributors/exponential.ts
ST-DDT: Splitting the options into two separate signatures is fine.
Maybe we did this for discoverability of both options or maybe this wasn't finished in the first place.

matthewmayer:
What is the point of the uniformDistributor? Isn't that the same as not passing a distributor at all?

matthewmayer:
Should we provide a sample normal distribution function (it wouldn't be a true normal distribution because it's bounded at the min and max). 

But I think it would be a fairly common use case to want something like "generate me heights in centimetres, with a mean of 170cm and standard deviation of 7cm." So I think we could provide a "recipe" for that kind of distribution even if it's not strictly normal.

ST-DDT:
> What is the point of the uniformDistributor? Isn't that the same as not passing a distributor at all?

To allow for using a non-nullish distributor value.
But I dont particulary care.

matthewmayer:
> > What is the point of the uniformDistributor? Isn't that the same as not passing a distributor at all?
> 
> 
> 
> To allow for using a non-nullish distributor value.
> 
> But I particulary care.

That's fine maybe we could just clarify that in the documentation. Otherwise a developer may set the option and wonder why "nothing happens".

ST-DDT:
EDIT/fix: 
But I dont particulary care.

ST-DDT:
> Should we provide a sample normal distribution function (it wouldn't be a true normal distribution because it's bounded at the min and max).
> 
> But I think it would be a fairly common use case to want something like "generate me heights in centimetres, with a mean of 170cm and standard deviation of 7cm." So I think we could provide a "recipe" for that kind of distribution even if it's not strictly normal.

You might want to start without it first.
It is tricky in regard to boundary matching and you might want to have a mathematician have a look at it so it doesnt deviate too much from the expectations.
Aka:
- how do we prevent values outside of the 0-1 range, because a normal distribution goes -infinity to +infinity
- how does that change if we configure the "width" or "center" of the distribution
- how can we help users configure it more easily, where is "170cm height" or "7cm std diviation" in the scale between 0 and 1.

matthewmayer:
I am a mathematician 😀 I will have a go at something that would meet expectations.

src/distributors/exponential.ts
xDivisionByZerox: While working on the separation of the signatures, I've noticed that could make it harder for users to dynamically provide the options object. This is for multiple reasons:

- When using multiple signatures, you can no longer infer the parameter type correctly:
   ```ts
   type BaseOrBias = Parameters<typeof exponentialDistributor>[0]; // depending on which signature is first, but for example { bias?: number | undefined; } | undefined
   ```
- Even if you type the options union yourself and provided a value, you will run into an inference problem:
   ```ts
   type BaseOrBias = { base: number } | { bias: number };

   function getBaseOrBias(): BaseOrBias {
     // implementation
   }

   exponentialDistributor(getBaseOrBias());
   ```
   Results in this error:
   > No overload matches this call.
  Overload 1 of 2, '(options?: { base?: number | undefined; } | undefined): Distributor', gave the following error.
    Argument of type 'BaseOrBias' is not assignable to parameter of type '{ base?: number | undefined; } | undefined'.
      Type '{ bias: number; }' has no properties in common with type '{ base?: number | undefined; }'.
  Overload 2 of 2, '(options?: { bias?: number | undefined; } | undefined): Distributor', gave the following error.
    Argument of type 'BaseOrBias' is not assignable to parameter of type '{ bias?: number | undefined; } | undefined'.
      Type '{ base: number; }' has no properties in common with type '{ bias?: number | undefined; }'.ts(2769)
exponential.ts(128, 17): The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.

Because of that, I'd say we keep the options as a union.

src/distributors/exponential.ts
xDivisionByZerox: @matthewmayer can you give insight here? I did not study mathematics, so I can only give perspective from a user that is clueless about these things.

ST-DDT:
Any update on this? I can resolve the merge conflicts for this.

xDivisionByZerox:
@ST-DDT I think no one really had time to do additional reviews after I updated the last version.
I've resolved a lot of comments from my previous review. 

I'm still not sure about the exponential distribution functions documentation (see https://github.com/faker-js/faker/pull/3375#discussion_r2153125317).
I also still think that we should add dedicated tests for the distributors itself (see: https://github.com/faker-js/faker/pull/3375#discussion_r2153071889).

I resolved the merge conflicts for now and can take a look at the additional test cases.

scripts/apidocs/processing/class.ts
```diff
@@ -196,6 +198,34 @@ export function processProjectUtilities(project: Project): RawApiDocsPage {
...
 }
 
+// Distributors
+
+export function processProjectDistributors(project: Project): RawApiDocsPage {
+  console.log(`- Distributors`);
+
+  const distributor = required(
+    project
+      .getSourceFile('src/distributors/distributor.ts')
```
xDivisionByZerox, line 208: _Leaving the comment on this, since it displays the file path in a meaningful way (and the discussion can be closed afterwards)._

Should these functions really be located directly in the "src" directory? Our project structure so far has been pretty clean/aligned with the public api.
I feel like these should be put in the "src/util" directory. Opinions?

ST-DDT: `src/randomizer.ts` is there as well.

xDivisionByZerox: And so is `faker.ts` - but these are core elements you simply cant omit when working with Faker, Distributors or optional when interacting with our library.

ST-DDT: I'm not sure I can follow your argument.
You can "omit" randomizers when working with faker.
We could put randomizers into its own folder, but for a long time it was only one thing so it didn't need to be.
Distributors have more variation than randomizers, so they have their own file each.
But each file only contains one entry so the usual nested structure for modules does not make sense to me.

grafik

Which leads to the last point, could it be in utils?
Sure, but for me utils is the place where you put stuff, that you don't know where else to put them.
In the api docs I wouldn't want distributors under utilities alone, maybe with randomizers there as well, it might be consistent.
If you define soundly what a faker utility is, then we should apply it to both source and docs.

One way to define it would be "everything, that is not Faker or one of its modules".

then it would look like this (ignoring order):

- Utilities (not clickable)
  - Distributors
  - Randomizers
  - Functions (e.g. mergeLocales)
  - Enums (e.g. SexType)
  - Types (LocaleDefinitions)

IMO 3 entries is the minimum to warrant a nested structure.

This may be an entire point by itself to discuss, just keep in mind, that the internal file structure is not part of the public API yet, so we can change it later as well.

xDivisionByZerox: I'm closing this discussion. After rereading it, I feel like the comment was unnecessarily nitpicky. 
I still think that we lack a clear architectural plan on organizing files in a meaningful way (for anything other than modules). But this is the case because there is no clear rule to do so as of right now. Starting this discussion in this feature PR does not achieve the goal of discussion and save-guarding the feature itself. Instead it creates artificial discussion point for things outside of the feature scope.

src/distributors/exponential.ts
xDivisionByZerox: I'm closing this. Users can follow through with feedback if the description of the function is not clear enought.

test/modules/number.spec.ts
xDivisionByZerox: @ST-DDT I saw that you added the tests for the distributors themself. Thanks for that.

Should we still change the test implementation for the number module by using a deterministic distributor for the tests in this module? Most of the tests cases test the integration of a correctly set up `exponentialDistributor`:

- int() and float()
  - should generate a number with low base
  - should generate a number with high base
  - should generate a number with low bias
  - should generate a number with high bia

As far as I can tell, these function should not case about theses specifics. The test can be simplified to a single one: "should generate a number based on the provided distributor" (or similar).
In this test we could use a deterministic distributor (like one that always returns 0) to predict the desired outcome:

```ts
it('should generate a number based on the provided distributor', () => {
  let distributorCall = 0;  
  const distributor = () => distributorCall++ % 4 === 0 ? 1 : 0;
  const results = Array.from({ length: 10 }, () => 0);
    results[faker.number.int({ max: 9, distributor })]++;
  }

  expect(results[0]).toBe(750);
  expect(results[9]).toBe(250);
});
```

xDivisionByZerox (commented):
Other than the comment in https://github.com/faker-js/faker/pull/3375#discussion_r3101530999 I couldn't find anything that is blocking from my side. After I got some other opinions on the matter, I'm feeling confident in approving this.

test/modules/number.spec.ts
ST-DDT: Fixed

ST-DDT:
> Other than the comment in [#3375 (comment)](https://github.com/faker-js/faker/pull/3375#discussion_r3101530999) I couldn't find anything that is blocking from my side.

Fixed

Shinigami92 (approved):

xDivisionByZerox (approved):

xDivisionByZerox:
Thank you for bringing this to an end @ST-DDT! :pray:

PR #3393, opened 2025-02-06
test: enable test to generate Number.MAX_SAFE_INTEGER

ST-DDT:
I checked the codebase for todos and found this disabled test, so I fixed and enabled it.

The tests have been added in v8.4.0

- #2470

And should have been enabled in v9.0.0-alpha.0 two months later

- #2357

### DIFF
--- a/test/modules/number.spec.ts
+++ b/test/modules/number.spec.ts
@@ -710,8 +710,7 @@ describe('number', () => {
         expect(actual).toBe(0);
       });
 
-      // TODO @ST-DDT 2023-10-12: This requires a randomizer with 53 bits of precision
-      it.todo('should be able to return MAX_SAFE_INTEGER', () => {
+      it('should be able to return MAX_SAFE_INTEGER', () => {
         randomizer.next = () => MERSENNE_MAX_VALUE;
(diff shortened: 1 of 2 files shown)


Shinigami92 (approved):

Shinigami92:
Just a little thingy I observed here in the PR: Looks like the `Randomizer#next` function is not sealed, at least not at TypeScript level 🤔 Should we think about it to mark it as `readonly` so someone needs to explicitly use `// @ts-expect-error: force override`?

ST-DDT:
We could do that 🤷‍♂️ , but we havent sealed any other object/module either.

Shinigami92:
> We could do that 🤷‍♂️ , but we havent sealed any other object/module either.

 👀

Shinigami92:
> > We could do that 🤷‍♂️ , but we havent sealed any other object/module either.
> 
> [img] 👀

I played a bit around with it, and it makes to many bad side-effects in tests, snapshots and other things like auto-doc-get. So I decided against touching it. 👌

xDivisionByZerox (approved):

PR #3395, opened 2025-02-07
test: retry verify tag 3 times

Shinigami92:
Retries the test 3 times, because locally it errors for me very often when I do a first run.

Looks like this even happens on GitHub CI: https://github.com/faker-js/faker/actions/runs/13203808509/job/36862047625?pr=3394#step:7:3071


ST-DDT:
Pros: Fixes issues with some tests
Cons: Makes the test less useful to detect random errors

I think the cons don't matter that much as the changed tests have a different purpose.

ST-DDT (approved):

ejcheng (approved):

PR #3400, opened 2025-02-12
fix: test before using Buffers

### LINKED ISSUE #3285: `internet.jwt()` fails on documentation website

ST-DDT:
Fixes #3285

---

Adds a feature check before committing to use the Buffer methods.

## Alternative

Remove the Buffer "shortcut" check and usage from the base64url method.

### DIFF
--- a/src/internal/base64.ts
+++ b/src/internal/base64.ts
@@ -14,7 +14,7 @@
  * @example const encodedHeader = toBase64(JSON.stringify(header));
  */
 export const toBase64: (input: string) => string =
-  typeof Buffer === 'undefined'
+  typeof Buffer === 'undefined' || !bufferFeatureCheck('base64')
     ? (input) => {
         const utf8Bytes = new TextEncoder().encode(input);
         const binaryString = Array.from(utf8Bytes, (byte) =>
@@ -39,10 +39,24 @@ export const toBase64: (input: string) => string =
  * @example const encodedHeader = toBase64Url(JSON.stringify(header));
  */
 export const toBase64Url: (input: string) => string =
-  typeof Buffer === 'undefined'
+  typeof Buffer === 'undefined' || !bufferFeatureCheck('base64url')
     ? (input) =>
         toBase64(input)
           .replaceAll('+', '-')
           .replaceAll('/', '_')
           .replaceAll(/=+$/g, '')
     : (input) => Buffer.from(input).toString('base64url');
+
+/**
+ * Checks whether the environment supports the given encoding on the `Buffer` class.
+ * This is required because some `Buffer` polyfills do not support all encodings.
+ *
+ * @param encoding The encoding to check.
+ */
+function bufferFeatureCheck(encoding: 'base64' | 'base64url'): boolean {
+  try {
+    return typeof Buffer.from('test').toString(encoding) === 'string';
+  } catch {
+    return false;
+  }
+}
--- a/src/internal/base64.ts
+++ b/src/internal/base64.ts
@@ -53,7 +53,7 @@ export const toBase64Url: (input: string) => string =
  *
  * @param encoding The encoding to check.
  */
-function bufferFeatureCheck(encoding: 'base64' | 'base64url'): boolean {
+function bufferFeatureCheck(encoding: BufferEncoding): boolean {
   try {
     return typeof Buffer.from('test').toString(encoding) === 'string';
   } catch {
--- a/src/internal/base64.ts
+++ b/src/internal/base64.ts
@@ -53,6 +53,7 @@ export const toBase64Url: (input: string) => string =
  *
  * @param encoding The encoding to check.
  *
+ * @see https://www.npmjs.com/package/buffer
  * @see https://github.com/feross/buffer/issues/309
  */
 function bufferFeatureCheck(encoding: BufferEncoding): boolean {
--- a/src/internal/base64.ts
+++ b/src/internal/base64.ts
@@ -52,6 +52,8 @@ export const toBase64Url: (input: string) => string =
  * This is required because some `Buffer` polyfills do not support all encodings.
  *
  * @param encoding The encoding to check.
+ *
+ * @see https://github.com/feross/buffer/issues/309
  */
 function bufferFeatureCheck(encoding: BufferEncoding): boolean {
   try {


src/internal/base64.ts
```diff
@@ -14,7 +14,7 @@
  * @example const encodedHeader = toBase64(JSON.stringify(header));
  */
 export const toBase64: (input: string) => string =
-  typeof Buffer === 'undefined'
+  typeof Buffer === 'undefined' || !bufferFeatureCheck('base64')
```
xDivisionByZerox, line 17: Can we put the feaure check result in a constant on top of the file without any export statements? The original idea of using the Buffer methods was a minimal performace gain if it exist. Running this check every call might not complety negate the original performance gain, but will at least diminish it slightly.

xDivisionByZerox (commented):
I'm honestly fine with both solutions. Having the explicit check function with documentation linking to the related upstream issues, might remind someone in the future to look into this and remove the workaround. On the other side, most of us are propably suvscribed to the external issues as well and will get notified anyway.

src/internal/base64.ts
ST-DDT: I'm confused what you are referring to here. 
The check is only executed once, because it is used for the initialization of the function constant and not executed as part of the function itself.

````ts
const method =
  condition
  ? impl1
  : impl2
````

Comment (Shinigami92) [reply src/internal/base64.ts:17]: Not sure if I read something wrong here, but I agree with @xDivisionByZerox  
Right now `toBase64` is called every time in `image.dataUri`.

Comment (ST-DDT) [reply src/internal/base64.ts:17]: I added a log to the buffer feature check:

````diff
 function bufferFeatureCheck(encoding: BufferEncoding): boolean {
+  console.log('bufferFeatureCheck', encoding);
   try {
````

Then I ran the method a few times:

````ts

console.log(faker.image.dataUri());
console.log(faker.image.dataUri());
console.log(faker.image.dataUri());
````

And I got this result:

````txt
$ pnpm exe

> faker-test@ exe ~/git/Faker/faker-playground
> tsx index.ts

bufferFeatureCheck base64
bufferFeatureCheck base64url
data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZlcnNpb249IjEuMSIgYmFzZVByb2ZpbGU9ImZ1bGwiIHdpZHRoPSI4NSIgaGVpZ2h0PSIyMTQ3Ij48cmVjdCB3aWR0aD0iMTAwJSIgaGVpZ2h0PSIxMDAlIiBmaWxsPSIjZWZlYWI0Ii8+PHRleHQgeD0iNDIuNSIgeT0iMTA3My41IiBmb250LXNpemU9IjIwIiBhbGlnbm1lbnQtYmFzZWxpbmU9Im1pZGRsZSIgdGV4dC1hbmNob3I9Im1pZGRsZSIgZmlsbD0id2hpdGUiPjg1eDIxNDc8L3RleHQ+PC9zdmc+
data:image/svg+xml;charset=UTF-8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20version%3D%221.1%22%20baseProfile%3D%22full%22%20width%3D%221614%22%20height%3D%221660%22%3E%3Crect%20width%3D%22100%25%22%20height%3D%22100%25%22%20fill%3D%22%239a7cef%22%2F%3E%3Ctext%20x%3D%22807%22%20y%3D%22830%22%20font-size%3D%2220%22%20alignment-baseline%3D%22middle%22%20text-anchor%3D%22middle%22%20fill%3D%22white%22%3E1614x1660%3C%2Ftext%3E%3C%2Fsvg%3E
data:image/svg+xml;charset=UTF-8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20version%3D%221.1%22%20baseProfile%3D%22full%22%20width%3D%22636%22%20height%3D%22550%22%3E%3Crect%20width%3D%22100%25%22%20height%3D%22100%25%22%20fill%3D%22%23aabfd7%22%2F%3E%3Ctext%20x%3D%22318%22%20y%3D%22275%22%20font-size%3D%2220%22%20alignment-baseline%3D%22middle%22%20text-anchor%3D%22middle%22%20fill%3D%22white%22%3E636x550%3C%2Ftext%3E%3C%2Fsvg%3E
````

As you can see, the `bufferFeatureCheck` was only executed once for each variant.

Comment (Shinigami92) [reply src/internal/base64.ts:17]: Ah, now I got it! I was confused to the used JS syntax. I usually never `export const fnName = ...` anon-functions but always prefer `export function fnName(...`
This is a good example where it would make the code more readable.

Comment (ST-DDT) [reply src/internal/base64.ts:17]: > This is a good example where it would make the code more readable.

If you know a way to rewrite it to achieve the same behavior...

Comment (xDivisionByZerox) [reply src/internal/base64.ts:17]: > I was confused to the used JS syntax.

Same here. I did not see that the availablitly check was not part of the functions body.

> If you know a way to rewrite it to achieve the same behavior.

I don't think that neccessary, TBH. It was an honest mistake. This diff wasn't in the reviewer favor as well. At least for the "toBase64" part. I'll verify my believes more thoroughly next time, before writing a suggestion.

**Review by xDivisionByZerox — APPROVED** [review]

File: src/internal/base64.ts (hunk 1 above)
Comment (ST-DDT) [reply src/internal/base64.ts:17]: > > If you know a way to rewrite it to achieve the same behavior.
>
> I don't think that necessary [...] I'll verify my believes more thoroughly next time, before writing a suggestion.

No problem, explanations (including those of my intentions) aren't my forte unfortunately.
I hope my statement didn't sound harsh as I was merely curious how else to achieve this.
I do have an idea, but that wouldn't help with code readability.

**Comment by Shinigami92** [comment]
My opinionated solution:

```ts
/* eslint-disable no-undef -- This file serves as a compatibility layer between environments */

/**
 * Checks whether the environment supports the given encoding on the `Buffer` class.
 * This is required because some `Buffer` polyfills do not support all encodings.
 *
 * @param encoding The encoding to check.
 *
 * @see https://www.npmjs.com/package/buffer
 * @see https://github.com/feross/buffer/issues/309
 */
function bufferFeatureCheck(encoding: BufferEncoding): boolean {
  try {
    return typeof Buffer.from("test").toString(encoding) === "string";
  } catch {
    return false;
  }
}

const SUPPORTS_BASE64_BUFFER =
  typeof Buffer === "undefined" || !bufferFeatureCheck("base64");
const SUPPORTS_BASE64_URL_BUFFER =
  typeof Buffer === "undefined" || !bufferFeatureCheck("base64url");

/**
 * This works the same as `Buffer.from(input).toString('base64')`
 * to work on both Node.js and browser environment.
 *
 * @internal
 *
 * @param input The string to encode to Base64.
 *
 * @returns Base64 encoded string.
 *
 * @see https://datatracker.ietf.org/doc/html/rfc4648
 *
 * @example const encodedHeader = toBase64(JSON.stringify(header));
 */
export function toBase64(input: string): string {
  if (SUPPORTS_BASE64_BUFFER) {
    return Buffer.from(input).toString("base64");
  }

  const utf8Bytes = new TextEncoder().encode(input);
  const binaryString = Array.from(utf8Bytes, (byte) =>
    String.fromCodePoint(byte)
  ).join("");
  return btoa(binaryString);
}

/**
 * This works the same as `Buffer.from(input).toString('base64url')`
 * to work on both Node.js and browser environment.
 *
 * @internal
 *
 * @param input The string to encode to Base64 URL.
 *
 * @returns Base64 URL encoded string.
 *
 * @see https://datatracker.ietf.org/doc/html/rfc4648
 *
 * @example const encodedHeader = toBase64Url(JSON.stringify(header));
 */
export function toBase64Url(input: string): string {
  if (SUPPORTS_BASE64_URL_BUFFER) {
    return Buffer.from(input).toString("base64url");
  }

  return toBase64(input)
    .replaceAll("+", "-")
    .replaceAll("/", "_")
    .replaceAll(/=+$/g, "");
}
```

I know this moves the check into the function, but makes it (to me) much more readable. I don't think that the condition has a to bad performance compared to actually craft the `base64` string.

Edit: we might even be able to remove the `typeof Buffer === "undefined"`, because the try-catch would catch it anyway 😅

PR #3402, opened 2025-02-16
feat(number): bigint multipleOf

soc221b:
Adds the `multipleOf` option to `faker.number.bigInt()`:

- [Preview NumberModule](https://deploy-preview-3402.fakerjs.dev/api/number.html#bigint)


src/modules/number/index.ts
```diff
@@ -406,6 +406,12 @@ export class NumberModule extends SimpleModuleBase {
            * @default min + 999999999999999n
            */
           max?: bigint | number | string | boolean;
+          /**
+           * The generated bigint will be a multiple of this parameter.
```
ST-DDT, line 414: Please add this to `@param options.multipleOf` as well + add an example with the multipleOf option set.

ST-DDT:
I'll look at it more closely soon.
Have a nice weekend

ST-DDT:
I finally had some time to look into you PR and run some tests locally.
I found a few cases that behave differently from the existing methods:

````ts
faker.number.int({ 'min': 6, 'max': 9, 'multipleOf': 5 }) // Error: No suitable integer value between 6 and 9 found.
faker.number.bigInt({ 'min': 6, 'max': 9, 'multipleOf': 5 }) // 10
````

````ts
faker.number.int({ 'min': -9, 'max': 0, 'multipleOf': 5 }) // -5
faker.number.bigInt({ 'min': -9, 'max': 0, 'multipleOf': 5 }) // Error: Multiple of 5n should be less than or equal to max 0n.
````

Please fix these cases and add tests for them.

File: src/modules/number/index.ts [hunk 2]
```diff
@@ -421,17 +431,40 @@ export class NumberModule extends SimpleModuleBase {
...
+        ? min / multipleOf
+        : BigInt(0) <= min
+          ? min / multipleOf + BigInt(1)
+          : min / multipleOf;
+    const effectiveMax =
+      max % multipleOf === BigInt(0)
+        ? max / multipleOf
+        : BigInt(0) <= max
+          ? max / multipleOf
+          : max / multipleOf - BigInt(1);
```
ST-DDT: IMO this can be simplified a bit (and extended with a comment for why we do it):

```suggestion
        const effectiveMin = min / multipleOf + (min % multipleOf > 0n ? 1n : 0n); // Math.ceil(min / multipleOf)
        const effectiveMax = max / multipleOf - (max % multipleOf < 0n ? 1n : 0n); // Math.floor(max / multipleOf)
```

src/modules/number/index.ts
```diff
@@ -421,17 +431,40 @@ export class NumberModule extends SimpleModuleBase {
...
     }
 
-    const delta = max - min;
+    if (effectiveMax < effectiveMin) {
+      throw new FakerError(
+        `No suitable bigint value between ${min} and ${max} found.`
+      );
+    }
 
+    const delta = effectiveMax - effectiveMin;
```
ST-DDT: Maybe move the +1 here same as in `int()`.

```suggestion
    const delta = effectiveMax - effectiveMin + 1n; // +1 for inclusive max bounds and even distribution
```

src/modules/number/index.ts
```diff
@@ -440,8 +473,7 @@ export class NumberModule extends SimpleModuleBase {
         })
       ) %
       (delta + BigInt(1));
```
```suggestion
      ) % delta;
```

src/modules/number/index.ts
```diff
@@ -421,17 +431,40 @@ export class NumberModule extends SimpleModuleBase {
...
+          : max / multipleOf - BigInt(1);
+
+    if (effectiveMin === effectiveMax) {
+      return effectiveMin * multipleOf;
     }
 
-    const delta = max - min;
+    if (effectiveMax < effectiveMin) {
+      throw new FakerError(
+        `No suitable bigint value between ${min} and ${max} found.`
```
ST-DDT, line 453: @faker-js/maintainers @faker-js/members 
Not sure about this one. `BigInt` or `bigint`?

```suggestion
        `No suitable BigInt value between ${min} and ${max} found.`
```

If we change it here, we should also change it in the jsdocs above.

src/modules/number/index.ts
```diff
@@ -375,14 +375,18 @@ export class NumberModule extends SimpleModuleBase {
    * @param options Maximum value or options object.
    * @param options.min Lower bound for generated bigint. Defaults to `0n`.
    * @param options.max Upper bound for generated bigint. Defaults to `min + 999999999999999n`.
+   * @param options.multipleOf The generated bigint will be a multiple of this parameter. Defaults to `1n`.
    *
    * @throws When `min` is greater than `max`.
+   * @throws When `multipleOf` is not a positive bigint.
+   * @throws When `multipleOf` is greater than `max`.
```
ST-DDT: This error is no longer the case.

```suggestion
```

src/modules/number/index.ts
```diff
@@ -375,14 +375,18 @@ export class NumberModule extends SimpleModuleBase {
    * @param options Maximum value or options object.
    * @param options.min Lower bound for generated bigint. Defaults to `0n`.
    * @param options.max Upper bound for generated bigint. Defaults to `min + 999999999999999n`.
+   * @param options.multipleOf The generated bigint will be a multiple of this parameter. Defaults to `1n`.
    *
    * @throws When `min` is greater than `max`.
+   * @throws When `multipleOf` is not a positive bigint.
```
ST-DDT, line 381: The multiple of parameter can technically be defined as string...

```suggestion
   * @throws When `multipleOf` is not positive.
```

ST-DDT (commented):
Sorry, for the slow review. It looks good to me, just a few suggestions to make it easier to read.

soc221b:
> Sorry, for the slow review. It looks good to me, just a few suggestions to make it easier to read.

Thank you for your review. I have made the changes you suggested.

ST-DDT (approved):

ST-DDT:

PR #3404, opened 2025-02-17
feat(finance): add ISO 4217 numerical codes to Currency object

### LINKED ISSUE #2289: Add numeric currency code by ISO 4217 standard

Right now `finance.currency` has a structure of `{ code: 'USD', name: 'US Dollar', symbol: '$' }`, but in many financial application I usually see numeric currency code usage, instead of string (so `840` instead of `USD`).

### Suggested solution

The suggestion is to add `numericCode` (probably needs a better name) to the currency object inside `faker.finance`.

I can try to implement that by myself

### Alternative

_No response_

### Additional context

_No response_

Nfloc:
Address issue #2289 

added numericCode parameter to currency object in all languages (put zero for all non ISO currency (cypto currency)).

added currencyNumericCode to finance definition.

added currencyNumericCode testing to finance.spec.ts

changed descriptions to match new parameter.

added random currencyNumericCode method.


src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
...
   }
 
+  /**
+   * Returns a random currency number ID.
+   * (The ISO 4217 numerical code for the currency (e.g. `US Dollar` -> 840 ))
+   *
+   * @example
+   * faker.finance.currencyNumID() // 840
+   *
+   * @since 9.4.0
```
matthewmayer: Next release will be 9.6.0

src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
     return symbol;
   }
 
+  /**
+   * Returns a random currency number ID.
+   * (The ISO 4217 numerical code for the currency (e.g. `US Dollar` -> 840 ))
+   *
+   * @example
+   * faker.finance.currencyNumID() // 840
```
matthewmayer: Don't like numID as a name

Suggest

numericCode
numericalCode
number

src/locales/fa/finance/currency.ts
```diff
@@ -3,135 +3,162 @@ export default [
     name: 'درهم امارات',
     code: 'AED',
     symbol: '',
+    numID: 784, // UAE Dirham
```
ST-DDT: Sometimes you have a comment here and sometimes you dont.
Would you mind sharing your intend for the comment or is it just a dev artifact?

src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
...
+   *
+   * @since 9.4.0
+   */
+  currencyNumID(): number {
```
ST-DDT: I'm not sure we need a new method as `faker.finance.currency().numID` is easy enough.
What do you think?

Shinigami92: When you have a look into the code above this function implementation, you will find that we currently also have functions for the other `currency*` functions. IMO we should at least be consistent, so I'm fine with having that new extra function.

matthewmayer: i agree for consistency with other methods lets keep the extra method.

src/modules/finance/index.ts
matthewmayer: https://fakerjs.dev/api/location#countrycode uses 'numeric' so i think `numericCode`/`currencyNumericCode()` would probably be most consistent?

src/modules/finance/index.ts
matthewmayer: Should it actually return a string rather than a number? It appears the codes are usually presented as three digits with leading zeros, e.g. AUD is `036` not `36`?

https://en.wikipedia.org/wiki/ISO_4217#cite_note-SOS217-1

image

src/locales/fa/finance/currency.ts
Nfloc: Mostly a dev artifact. 
For all the non-english languages I put the English translations so it was easier to read and compare.

edit: all translation comments removed

src/modules/finance/index.ts
Nfloc: Good question, in the original posted issue the wording indicating that it would be a number instead of a string.


- 419140da62a0f21134aed2ff18cb6c3f

Let me know opinions on what it should be, I'm not super familiar with financial applications.

src/modules/finance/index.ts
```diff
@@ -457,14 +462,15 @@ export class FinanceModule extends ModuleBase {
   }
 
   /**
-   * Returns a random currency object, containing `code`, `name `and `symbol` properties.
+   * Returns a random currency object, containing `code`, `name`, `symbol`, and `numericCode` properties.
    *
    * @see faker.finance.currencyCode(): For generating specifically the currency code.
    * @see faker.finance.currencyName(): For generating specifically the currency name.
    * @see faker.finance.currencySymbol(): For generating specifically the currency symbol.
+   * @see faker.finance.numericCode(): For generating specifically the currency numerical ID.
```
```suggestion
   * @see faker.finance.currencyNumericCode(): For generating specifically the currency numerical ID.
```

src/modules/finance/index.ts
```diff
@@ -27,6 +27,11 @@ export interface Currency {
    * The symbol for the currency (e.g. `$`).
    */
   symbol: string;
+
+  /**
+   * The ISO 4217 numeric code for the currency (e.g. `840`).
+   */
+  numericCode: number;
```
ST-DDT: Since the result is always 3 characters long and doesn't need the usual number operations like addition, we should use a string here.

Nfloc:
Should I update my original pull request description to fit the new parameter names?

src/locales/fa/finance/currency.ts
```diff
@@ -1,137 +1,932 @@
 export default [
   {
-    name: 'درهم امارات',
+    name: '阿联酋迪拉姆',
```
ST-DDT: I think you unintentionally mixed the names up in this file.

src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
     return symbol;
   }
 
+  /**
+   * Returns a random currency number ID.
+   * (The ISO 4217 numerical code for the currency (e.g. `US Dollar` -> '840' ))
```
```suggestion
   * (The ISO 4217 numerical code for the currency (e.g. `US Dollar` -> `'840'` ))
```

src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
     return symbol;
   }
 
+  /**
+   * Returns a random currency number ID.
```
```suggestion
   * Returns a random currency numeric code.
```

src/modules/finance/index.ts
```diff
@@ -457,14 +462,15 @@ export class FinanceModule extends ModuleBase {
   }
 
   /**
-   * Returns a random currency object, containing `code`, `name `and `symbol` properties.
+   * Returns a random currency object, containing `code`, `name`, `symbol`, and `numericCode` properties.
    *
    * @see faker.finance.currencyCode(): For generating specifically the currency code.
    * @see faker.finance.currencyName(): For generating specifically the currency name.
    * @see faker.finance.currencySymbol(): For generating specifically the currency symbol.
+   * @see faker.finance.currencyNumericCode(): For generating specifically the currency numerical ID.
```
```suggestion
   * @see faker.finance.currencyNumericCode(): For generating specifically the currency numeric code.
```

test/modules/finance.spec.ts
```diff
@@ -326,6 +328,14 @@ describe('finance', () => {
         });
       });
 
+      describe('currencyNumericCode()', () => {
+        it('should return a string', () => {
+          const currencyNumericCode = faker.finance.currencyNumericCode();
+
+          expect(currencyNumericCode).toBeTypeOf('string');
```
matthewmayer, line 335: also check its 3 characters long?

ST-DDT (dismissed):

matthewmayer (dismissed):

src/modules/finance/index.ts
```diff
@@ -516,6 +522,19 @@ export class FinanceModule extends ModuleBase {
     return symbol;
   }
 
+  /**
+   * Returns a random currency numeric code.
+   * (The ISO 4217 numerical code for the currency (e.g. `US Dollar` -> `'840'` ))
```
xDivisionByZerox: Are the parenthesis required here? Additionally, since this function returns a random code, it should be "a currency" not "the curreny":


```suggestion
   * The ISO 4217 numerical code for a currency (e.g. `US Dollar` -> `'840'` )
```

src/definitions/finance.ts
```diff
@@ -18,7 +18,7 @@ export type FinanceDefinition = LocaleEntry<{
   credit_card: { [issuer: string]: string[] };
 
   /**
-   * Currencies including their name, code and symbol (e.g. `US Dollar` / `USD` / `$`).
+   * Currencies including their name, code, symbol and ISO numericCode (e.g. `US Dollar` / `USD` / `$` / '840').
```
xDivisionByZerox: This should be written in plain English and not a reference to the property name:


```suggestion
   * Currencies including their name, code, symbol and ISO numerical code (e.g. `US Dollar` / `USD` / `$` / '840').
```

ST-DDT (approved):

ST-DDT:

PR #3417, opened 2025-03-01
fix(number): don't ignore multipleOf in float when min=max

ST-DDT:
Currently, `faker.number.float({ min: 6, max: 6, multipleOf: 5 })` does not throw, even though 6 is not a multiple of 5.

This PR removes the shortcut when `min === max` bypassing the multipleOf checks.
Now, it throws an error similar to `int()` (and soon `bigInt()`).

---

This issue was discovered while reviewing #3402

- #3402


src/modules/number/index.ts
```diff
@@ -113,7 +113,7 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.fractionDigits The maximum number of digits to appear after the decimal point, for example `2` will round to 2 decimal points.  Only one of `multipleOf` or `fractionDigits` should be passed.
    *
    * @throws When `min` is greater than `max`.
-   * @throws When `multipleOf` is negative.
+   * @throws When `multipleOf` is not a positive float.
```
ST-DDT: Float or number?

```suggestion
   * @throws When `multipleOf` is not a positive number.
```

xDivisionByZerox:
I'm unsure how to continue here. By semantic versioning definition this change should be considered breaking. On the other side this feature is implemented for at least one year and the issue with this misaligned behaviour was never raised.

matthewmayer:
It's a bugfix and I think the old behavior is unexpected and undocumented so I think patch-level is ok.

src/modules/number/index.ts
```diff
@@ -113,7 +113,7 @@ export class NumberModule extends SimpleModuleBase {
    * @param options.fractionDigits The maximum number of digits to appear after the decimal point, for example `2` will round to 2 decimal points.  Only one of `multipleOf` or `fractionDigits` should be passed.
    *
    * @throws When `min` is greater than `max`.
-   * @throws When `multipleOf` is negative.
+   * @throws When `multipleOf` is not a positive number.
```
Shinigami92, line 116: **question (non-blocking):** what's the difference between `is negative` and `is not a positive number`?

test/modules/number.spec.ts
```diff
@@ -156,6 +156,18 @@ describe('number', () => {
...
 
+      it('throws for impossible multipleOf where min=max', () => {
+        const input = {
+          min: 11,
+          max: 11,
+          multipleOf: 10,
+        };
+
+        expect(() => faker.number.int(input)).toThrow(
+          new FakerError('No suitable integer value between 11 and 11 found.')
```
Shinigami92, line 167: **thought:** as the "upper bound" is "exclusive", IMO this PR is valid and a correct implementation. I will approve ✅

Shinigami92 (approved):

src/modules/number/index.ts
xDivisionByZerox: "Negative" evalues to to less than zero.
"not a positive number" includes zero (since its neighter positive, nor negative).

The method throws when the "multipleOf" parmeter is zero. So not positive is more correct.

xDivisionByZerox (approved):

xDivisionByZerox:
> It's a bugfix and I think the old behavior is unexpected and undocumented so I think patch-level is ok.

Thanks for giving your thought on that. The problem is that semantic versioning doesn't  care about a behavior being unexpected. But I partially agree, so I'll approve as well.

PR #3429, opened 2025-03-01
fix(finance): routingNumber now uses real FederalReserveRoutingSymbol from lookup table.

### LINKED ISSUE #3290: Fix finance.routingNumber() implementation
### Problem Description

There are rules regarding what is and isn't possibly a legitimate routing number according to the American Bankers Association and/or the Federal Reserve. As of now, faker complies with only a _part_ of this ruleset (the check digit validation), but based on a few tests I ran locally, the Federal Reserve Routing Symbol (the first 4 digits) and the Institution Identifier (the next 4 digits) of FinanceModule.routingNumber have failed to correspond to an actual, usable-in-the-wild value.

**Note:** This proposal is primarily concerned with modifications that would produce a routing number with a valid Federal Reserve Routing Symbol (first 4 digits), and _not_ the Institution Identifier (next 4 digits). Also note that, despite using the words "in the wild", that is not the goal of this proposal, but rather to produce a routing number that is significantly closer to what a real, usable routing number would be than what faker currently produces.

### Suggested solution (if the proposed data meets all requirements for licensing/copyright)

[Link to "Proposed Data"](https://www.frbservices.org/resources/fees/check-key-to-routing-numbers.html)

In FinanceModule.routingNumber():

**digits [0-3] (Federal Reserve Routing Symbol)** = Pull from a static lookup table derived from the proposed data.
**digits [4-7] (Institution Identifier)** = 4 random digits. (I am not proposing to attempt to solve this problem.)
**digits[8] (Check Digit)** = derived from digits[0-7]

### Alternative (if the proposed data does NOT meet requirements for licensing/copyright)

**digits [0-1] (Federal Reserve District Identitifer)** = Random number between 1 and 12, left-padded with a zero. This would correspond to the 12 federal reserve banks, and would be better than nothing at all. 

**Note:** '00' is reserved for U.S. government use and, in my opinion, should not be allowed to be produced as [I assume] almost anybody using this library is not going to be concerned with routing numbers starting with '00', but that may be a false assumption; you tell me.

**digits [2-3] (Federal Reserve Branch Office City Identifier)** = 2 random digits, left-padded with a zero.

**Note:** [The wikipedia page for ABA Routing Numbers](https://en.wikipedia.org/wiki/ABA_routing_transit_number#ABA_prefix_table) lists what appears to be all of these, but no source is cited, and therefore cannot be used as a source of truth. Furthermore, it lists a 3-digit value '101' as belonging to American territories such as American Samoa, Puerto Rico, Guam, etc., which doesn't make much sense. Also, this isn't strictly necessary for the proposal, anyway.

The rest is the same as the primary proposed solution:

**digits [4-7] (Institution Identifier)** = 4 random digits.
**digits[8] (Check Digit)** = derived from digits[0-7]

### Additional context

The trouble with implementing a feature like this is that _technically_, the "rules" surrounding what does or does not constitute a routing number which corresponds to a real financial institution are always in flux. However, _practically_, the majority of these rules either have never changed since the Federal Reserve's inception in 1913, as far as I can tell anyway (e.g. routing numbers starting in '80' for traveler's checks), or have not changed in such a long time that they can be effectively treated as though they've never changed (e.g. "thrift" institutions codes have not changed since the mid-1980s). Therefore, even if the updates to the rules vs this library were not fully in sync, there would still be a vastly improved chance of getting a valid, real routing number than otherwise. The main part which I believe can be improved is that of the Federal Reserve Routing Symbol, which is simply the first 4 characters of the routing number, which is further broken down as the first 2 characters designating the Federal Reserve District and the second 2 characters designating the city of the reserve branch within that district. These do not change very often (I'm not entirely sure, after some research, the last time they have).

You can check validation [here](https://routingnumber.aba.com/Search1.aspx) (https://routingnumber.aba.com/Search1.aspx) to see for yourself (one simple test is to use '522814402' ([the routing number given in the comment above the function definition](https://github.com/faker-js/faker/blob/42083ac91e97dcc5aa3f25d18f89560b5b5f4c76/src/modules/finance/index.ts#L174)), but this is a business and using that API is of course not viable for an open-source product. I mentioned this simply as a quick way to verify the legitimacy of the need for this feature.

And just in case this point is raised, I suppose I can address it with my own opinion now. It is arguable that since faker's purpose is to produce fake data anyway, it shouldn't need to be a real financial institution at all. However, I am of the opinion that since test code will inevitably use portions of production code, and it cannot be assumed that that production code was written in a way that makes it feasible to isolate its mechanics from its semantics, it should not be a concern for those using this library to need to differentiate between a valid (real-world) routing number and an invalid routing number that simply passes check digit validation.

Furthermore, it may be best simply to augment the existing functionality of routingNumber() to just return one with either its first 4 digits from the lookup table or the first 2 constrained to between '01' - '12', and not have an optional boolean argument at all. After all, using the proposed boolean switch should simply produce a routing number that is a member of a subset of all the routing numbers that it currently produces, so any valid routing number when using the switch should also be among the set of those produced without it.

woldaker:
Fixes #3290

---

finance.routingNumber is now comprised as follows:

Digits 0-3 = Pulled from locales/base/finance/federal_reserve_routing_symbol.ts
Digits 4-7 = Random
Digit 8 = Check digit. Calculated from the previous 8 digits as before.

Added more test conditions to the randomly-seeded finance.routingNumber test.


matthewmayer:
while routingNumber is perhaps not the ideal name, the costs of changing it probably outweigh just leaving it for now. If in the future there'a a request to add a different type of routing number, we could consider changing it then.

matthewmayer (commented):
Let's add a little more detail to the method description

e.g. 
```
"Generates a random routing number."
```
=>

```
"Generates a random [ABA routing number](https://en.wikipedia.org/wiki/ABA_routing_transit_number)"
```

woldaker:
Done. I used one of the routing numbers from the test snapshot for the example string.

woldaker:
Also, I forgot to mention... I couldn't figure out how to link [the original issue](https://github.com/faker-js/faker/issues/3290) to this PR. Sorry I am a bit new to contributing to public github repos.

woldaker:
Also, as I'm sure you've noticed, I skipped any form of testing whether the 2nd group of 2 digits at indexes 2-3 actually matched up with one of the numbers listed in the base finance locale. Of course were I to do that, what I'd really be doing is simply checking if `listOfFederalReserveRoutingSymbolsFromBaseLocale.includes(routingNumber.substring(0, 4))`, but it appeared as though performing that sort of a check in the tests was not really what y'all wanted. I also considered splitting the first set of 2 digits up from the second set of 2 digits (the institution identifier), which if I had gone in that direction I would have encapsulated the FederalReserveRoutingSymbol into an object and then would have needed to instead make `src/locales/base/finance/routing_number.ts` and then in there do `export const federal_reserve_district = [...];` and `export const institution_identifier = (federalReserveDistrict: number) => {...};` or something similar, and then it's no longer a simple lookup table anymore, and I thought all of it was just a bit too overkill and was venturing into over-engineering territory, so I decided against that.

Edit: On a site note, it was interesting to find out that validate.js's `isAbaRouting()` function does not currently perform any validation on the aspects of an ABA routing number that this PR addresses. I looked at its source. All it's really doing is validating the check digit.

Edit 2: After looking at [the validator.js source](https://github.com/validatorjs/validator.js/blob/master/src/lib/isAbaRouting.js) again, I had missed this regex at the top:

`^(?!(1[3-9])|(20)|(3[3-9])|(4[0-9])|(5[0-9])|(60)|(7[3-9])|(8[1-9])|(9[0-2])|(9[3-9]))[0-9]{9}$`

Which _is_ checking (though not obviously, of course) that the federal reserve district is correct.
The last bit `[0-9]{9}` is verifying that it's a total of 9 digits long, which is great, but then you have the entire section before that which, after analyzing it further, is checking the first two digits to make sure that it falls under one of these categories:

Quoted from the wikipedia page on ABA routing numbers:

- 00 is used by the United States Government
- 01 through 12 are the "normal" routing numbers, and correspond to the 12 Federal Reserve Banks. For example, 0260-0959-3 is the routing number for Bank of America incoming wires in New York, with the initial "02" indicating the Federal Reserve Bank of New York.
- 21 through 32 were assigned only to [thrift institutions](https://en.wikipedia.org/wiki/Cooperative_banking) (e.g. credit unions and savings banks) through 1985, but are no longer assigned (thrifts are assigned normal 01–12 numbers). Currently they are still used by the thrift institutions, or their successors, and correspond to the normal routing number, plus 20. (For example, 2260-7352-3 is the routing number for Grand Adirondack Federal Credit Union in New York, with the initial "22" corresponding to "02" (New York Fed) plus "20" (thrift).)
- 61 through 72 are special purpose routing numbers designated for use by non-bank payment processors and clearinghouses and are termed Electronic Transaction Identifiers (ETIs), and correspond to the normal routing number, plus 60.
- 80 is used for traveler's checks

So that regex is effectively trying to be as exhaustive as possible, almost like an official spec would, except it's a little bizarre for two reasons:

One is that it is allowing 000000000, which passes even the check digit validation and is definitely not a valid routing number.
Two is the very strange use of `(9[0-2])|(9[3-9])`. I know, that's a little nit-picky, but what were the authors doing there exactly? Why not just `(9[0-9])`?

So tl;dr the call to `isAbaRouting()` in the test does handle the check digit (and the length=9) for us, but the additional check afterward to see if the first two digits fall within the range of 01-12 are still necessary so as to _disallow_ results y'all aren't interested in, like the thrift/legacy numbers which are no longer issued, traveler's checks, and the 00 govt-only ones.

test/modules/finance.spec.ts
```diff
@@ -149,11 +150,24 @@ describe('finance', () => {
       });
 
       describe('routingNumber()', () => {
-        it('should return a string', () => {
-          const routingNumber = faker.finance.routingNumber();
+        const routingNumber = faker.finance.routingNumber();
+
+        const firstTwoDigits = routingNumber.substring(0, 2);
+        const federalReserveDistrict = Number.parseInt(firstTwoDigits);
```
ST-DDT: Please move these into the the `it()` test blocks.

test/modules/finance.spec.ts
```diff
@@ -149,11 +150,24 @@ describe('finance', () => {
...
 
+        it('should return a string', () => {
           expect(routingNumber).toBeTypeOf('string');
         });
+
+        it('should pass check digit validation', () => {
+          expect(routingNumber).toSatisfy(isAbaRouting);
+        });
```
ST-DDT: IMO these two test cases should be joined into one.

woldaker: You can see from validator's source that [it is already asserting](https://github.com/validatorjs/validator.js/blob/a7d5cffe8930ad5c734c50d963d3bc544af4d3cc/src/lib/isAbaRouting.js#L9) [that it is a string](https://github.com/validatorjs/validator.js/blob/a7d5cffe8930ad5c734c50d963d3bc544af4d3cc/src/lib/util/assertString.js#L2). Does that make the first check completely unnecessary? The one argument I can see for saying NO to that is that that would leave the validity of faker.js's tests dependent upon the implementation details of validator.js.

test/modules/finance.spec.ts
woldaker: Okay. I think I did it that way for DRY purposes, as you'd need to re-construct `routingNumber` (or any other variable used in multiple `it()` blocks, but that's the only one this applies to here) each time you needed to reference it inside of the `it()` block, but if that's what you want, that's fine. I don't quite understand what's wrong with doing it the way I did it, but that's fine. If you care to explain I'd be happy to know though.

test/modules/finance.spec.ts
ST-DDT: The string check makes the result in the tests easier to grasp even if you dont read the jsdocs of either method.

woldaker:
Strange. I keep getting this today:
image

which ultimately ends up failing the preflight:
image

When I first went to make the changes you had requested, pnpm complained about being updated, so I ended up doing `pnpm setup` followed by `pnpm add -g @pnpm/exe` which gave an error saying that I had to remove the existing pnpm.exe first, so I did `Remove-Item C:\path\to\my\pnpm.exe` and then reran `pnpm add -g @pnpm/exe` and then it worked. But now it just fails preflight every time.

Any ideas?

ST-DDT:
The timeouts happen randomly unfortunately. Even if we increase it to 30s.
We arent sure why this happens, but you can ignore the timeouts for now.

As for the pnpm update, you might have to update node to the latest (LTS) version. That solved it for me.
I use `corepack enable` to manage the pnpm version in use.

test/modules/finance.spec.ts
ST-DDT: Vitest uses callbacks (`() => {}`) to run the tests. And to re-run them.
If the values are declared outside the callback then they wont change for subsequent executions. We run these tests multiple times to ensure they dont pass by chance (or at least reduce the probability).

https://github.com/faker-js/faker/blob/f591459aff030172915db2c7ba1f251e705898d8/test/modules/finance.spec.ts#L107-L109

woldaker:
Okay that makes me feel better that I didn't do something and break it unwittingly. I'll try again after work today.

woldaker:
Okay I tried again and it didn't time out this time. Pushed up the changes you mentioned.

woldaker:
I was checking back in to see if there was any further action y'all needed me to take to get the PR completed. I see that there was a conflict in the finance definitions file, so I resolved that. Just let me know if there's anything else I should do. Thanks.

xDivisionByZerox:
@woldaker Thanks for taking action on your own already. Please note that actually merging your PR might take some time. This is not because we don't like your feature but have to carefully consider multiple side effects a feature might have and how well it integrates into the entire API of faker. Lastly, we are obviously working on later major features for the entire Faker project. We also need to consider how new features would integrate in these future plans.
So, to summarize, don't worry if your PR is not getting any action for some time. I appreciate your patience :)

woldaker:
Understood. If I can be of any further help, just say the word, and I'll
keep an eye out.

test/modules/finance.spec.ts
```diff
@@ -150,10 +151,22 @@ describe('finance', () => {
       });
 
       describe('routingNumber()', () => {
-        it('should return a string', () => {
+        it('should return a string and pass check digit validation', () => {
```
xDivisionByZerox:
```suggestion
        it('should return a valid ABA routing number', () => {
```

src/locales/base/finance/federal_reserve_routing_symbol.ts
```diff
@@ -0,0 +1,168 @@
+// Source: https://www.frbservices.org/resources/fees/check-key-to-routing-numbers.html
+// Last Synced with Source: 2025-03-01
```
xDivisionByZerox: We can get the general information based on the commit date. I strongly believe that we would update these values (if required) before merging a PR that changes them.

xDivisionByZerox (commented):
It's been a while, but I've finally had the chance to review your changes. 🎉 

Overall, the implementation looks good to me. I have just two minor suggestions - these aren't strictly necessary, but they could help reduce the overall complexity introduced by this PR.

One key point I'd like to discuss with the Faker team is whether the routing number data should be included in the base locale. Since routing numbers are specific to the U.S. by design, users who intentionally select the base locale typically do so to minimize the amount of imported data at startup.

Additionally, the routing number function is a legacy implementation that does not fully align with Faker's current standards. If this function were introduced today, it might be designed with a more general-purpose name - such as `bankCode` - to accommodate multiple locales. To be clear, I’m **not** suggesting deprecating the function in favor of a more generic alternative! However, given the tight coupling between its business logic and the American locale, I find it challenging to justify returning values for other locales.

Would it make sense to expose the locale data exclusively to the default `en` - and thus default - locale? After all, when calling this method from, for example, the `de` locale, would users really expect an ABA routing number? 🤔 FYI the German equivalent of "routing number" would be the ["Bankleitzahl"](https://en.wikipedia.org/wiki/Bankleitzahl).

Looking forward to your thoughts @faker-js/members!

test/modules/finance.spec.ts
woldaker: This is true. US ABA routing numbers are for domestic transfers so that makes total sense to put them into the en_US locale. Internationally, you'd use an IBAN & SWIFT (or some other "routing number" equivalent for the country you're sending it to like GBDSC (UK) or CACPA (Canada)).

I can make those changes at some point this week after work.

woldaker:
This is true. US ABA routing numbers are for domestic transfers so that makes total sense to put them into the en_US locale. Internationally, you'd use an IBAN & SWIFT (or some other "routing number" equivalent for the country you're sending it to like GBDSC (UK) or CACPA (Canada)).

I can make those changes at some point this week after work.

test/modules/finance.spec.ts
xDivisionByZerox: > I can make those changes at some point this week after work.

Just making sure:
You only "need" to do the changes I specifically marked in the PR. You don't need to move the locale data just now. I'd like further feedback from other team members for now. 🙂

xDivisionByZerox:
> I can make those changes at some point this week after work.

Just making sure:
You only "need" to do the changes I specifically marked in the PR. You don't need to move the locale data just now. I'd like further feedback from other team members for now. 🙂

src/locales/base/finance/federal_reserve_routing_symbol.ts
woldaker: I should leave the source URL at the top though right?

woldaker: You're saying just remove the // Last Synced comment right?

xDivisionByZerox: Yes. Providing Source is very appreciated. Only the last synced line is obsolete IMO.

woldaker:
I am so sorry, work got busy and I completely forgot to finish this up! Taking another look at it now...

woldaker:
As an update, I took another look at this last night. I am not even on the same machine anymore that I was last time, so I didn't have pnpm, etc. After reinstalling what I think was everything I needed, pnpm run preflight kept failing with:

```
 FAIL  test/faker.spec.ts > faker > should not log anything on startup
Error: Test timed out in 5000ms.
...
 ELIFECYCLE  Command failed with exit code 1.
ERROR: "test:update-snapshots" exited with 1.
 ELIFECYCLE  Command failed with exit code 1.
```

so I'll have to take another look at it after work today.

woldaker:
Okay I was finally able to get it working. I think it was just that intermittent failure that nobody knows the cause of that was mentioned earlier. I made [the changes as suggested here](https://github.com/faker-js/faker/pull/3429#pullrequestreview-2933416949).

One thing I must say is that I _absolutely agree_ that the routing number table I added does **not** belong in base. It belongs in any locale where US is the region. However, that brings up a different issue altogether: when someone from the UK uses faker and calls RoutingNumber(), I would think they should get a GBDSC number (the UK equivalent of the ABA), which doesn't look anything like a US ABA routing number (in the regex/string-format sense. They do both fulfill the same purpose: to identify a specific bank both in brand and in branch location physically). You can find a pretty decent list of various country's bank routing number formats here: [https://developer.gs.com/docs/services/transaction-banking/routing-codes/](https://developer.gs.com/docs/services/transaction-banking/routing-codes/).

~~Is the plan to do the same for all of them? I presume not. So that brings into question the wisdom of this PR in general. If we do this for US routing numbers, why doesn't every country get the same treatment? That would be a lot of work, of course; far too much work than anyone on the faker team (I presume) would consider to be worth the cost.~~

EDIT: I realized that that last paragraph would only be true if every country actually had their own RoutingNumber() function. This PR as simply an improvement to the current RoutingNumber() function, which only returns numbers formatted as a US ABA routing number anyway, falls outside the realm of what I had originally said. I think it is still an improvement that doesn't hurt or affect anything else.

Let me know your thoughts.

xDivisionByZerox:
> EDIT: I realized that that last paragraph would only be true if every country actually had their own RoutingNumber() function. This PR as simply an improvement to the current RoutingNumber() function, which only returns numbers formatted as a US ABA routing number anyway, falls outside the realm of what I had originally said. I think it is still an improvement that doesn't hurt or affect anything else.

Big edit right there. I appreciate the effort you put into thinking about this feature in a wider sense. As pointed out in [my previous comment](https://github.com/faker-js/faker/pull/3429#pullrequestreview-2933416949) the `routingNumber()` method is a legacy method that we got form the previous code base back in 2022. This means that it's API doesn't really align with the design choices made in modern (post v7) Faker. Ultimately the Faker team (or a community member) probably needs to design a more generic API that suites all locales for this use case.

But the change you present is targeted on the validity of the current resulting values of the `routingNumber()` method - which is a valid change/suggestion in it off itself. So let's focus on that.

I don't know when I or one for the team members will have time to review this PR again, but I didn't want to leave you hanging without at least an answer since I saw that you had a previous issue with your development setup, to which no one responded to.

ST-DDT (approved):
IMO the current implementation looks good to me.

I think, we could put the array in a const in the finance module directly, since it is only ever used by the routing number and I think rarely change.  #3748 might be able to solve the tree shaking concerns. 
Unless there is a need to frequently update them/replace them with a custom set to satisfy a common need. @woldaker @faker-js/maintainers What do you think?

As for replacing the implememtation to conform with other locales.
I dont think we should do that. Instead we could add a separate method for those. Though most of them should have been replaced with bic anyway AFAIK.
Maybe we could rename the method to abaRoutingNumber following our usual process in a separate PR.

Shinigami92 (approved):

ST-DDT:
> I think, we could put the array in a const in the finance module directly

We can do that in a later refactor (same with most data from the base locale).

PR #3436, opened 2025-03-09
test: extract cjs require tests to .spec.cts

Shinigami92:
I just go a great idea over night! 💡 
What if we just move all `require(*)`-cjs calls into `.spec.cts` files. That way we can more safely use the `require`-nodejs keyword and ensure that it really works.
But not only this, when we will remove cjs support in v10, we could ensure this way at least for newer versions of nodejs, that it still works with their new `require(esm)` https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require


test/require.spec.cts
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable @typescript-eslint/no-require-imports, unicorn/prefer-module */
```
xDivisionByZerox: Maybe disable the rule for all *.cts files globally? I mean it doesn't really make sense to have it active there. 🤔

test/require.spec.cts
```diff
@@ -0,0 +1,49 @@
...
+const { describe, expect, it, vi } = await import('vitest');
+const { allLocales, SimpleFaker } = require('../dist/index.cjs');
+
+describe('require (cjs)', () => {
+  describe.each(
+    Object.keys(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      allLocales
+    )
```
xDivisionByZerox, line 9: Can you use our internal `keys` function (from `'src/internal/keys.ts'`)? That way we don't need to disable the line from linting.

xDivisionByZerox (commented):
> What if we just move all require(*)-cjs calls into .spec.cts files. That way we can more safely use the require-nodejs keyword and ensure that it really works.

What do you mean by "ensure that it really works"? Why does it not work right now? 🤔 


> But not only this, when we will remove cjs support in v10, we could ensure this way at least for newer versions of nodejs, that it still works with their new require(esm) [nodejs.org/api/modules.html#loading-ecmascript-modules-using-require](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require)

I don't see how moving the test in a separate file ensures this behavior. As far as I can tell this would require our code base to use explicit `*.m(t|j)s` file extensions, which is not the case right now. Am I missing something? Could you elaborate on this?

Shinigami92:
> > What if we just move all require(*)-cjs calls into .spec.cts files. That way we can more safely use the require-nodejs keyword and ensure that it really works.
> 
> What do you mean by "ensure that it really works"? Why does it not work right now? 🤔

Before this PR, we used `require` in a normal `ts` file which gets processed by vitest/vite and therefore potentially gets optimized (not sure if this really happens!)
When we now use `.cts` instead, we ensure that the file is running in pure common-js context.
https://github.com/faker-js/faker/pull/3436/files#diff-f8cd4ea46aecbba86ee9cba1a851bca523aca5e63bf9fa1866e3604a9d2dec26L9
https://github.com/faker-js/faker/blob/1e551c5f47e292d3dc6bbefae9eaf75664a052a1/test/locale-imports.spec.ts#L9

> > But not only this, when we will remove cjs support in v10, we could ensure this way at least for newer versions of nodejs, that it still works with their new require(esm) [nodejs.org/api/modules.html#loading-ecmascript-modules-using-require](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require)
> 
> I don't see how moving the test in a separate file ensures this behavior. As far as I can tell this would require our code base to use explicit `*.m(t|j)s` file extensions, which is not the case right now. Am I missing something? Could you elaborate on this?

No, we wont need to ship `.mjs` files, because everything is controlled be `package.json`'s `"type": "module"` and therefore the hole package will get consumed as pure esm package.
In combination with NodeJS's new [require(esm)](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require), this finally get solved. And our `.spec.cts` file can cover that.

test/require.spec.cts
Shinigami92: ~~I'll try 👀~~

Oh, I think I did it that way, because it does not use imports from `src` at all, but is more an integration/e2e-test depending on the build output
That way, the test also fully ensures that nothing gets "optimized" but reads the final bundled sourcecode

Shinigami92:
@xDivisionByZerox in the future (v10) we might need to run this `.spec.cts` test with Node v20.17+ and the `--experimental-require-module` or Node v23+
(see https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require and open the `History`)

And then we need to switch the required calls to depend on v10's esm bundled code
```diff
- require('../dist/index.cjs')
+ require('../dist/index.js')

- require(`../dist/locale/${locale}.cjs`)
+ require(`../dist/locale/${locale}.js`)
```

Shinigami92:
@xDivisionByZerox just an info, but https://github.com/faker-js/faker/pull/3540#issuecomment-3016734241 fails with node 18
But this PR here would also still run with node 18, because it actively still tests it with native cjs support and does not fake a require inside an esm file

Do you still wont like this cts test file change?

xDivisionByZerox (approved):
> @xDivisionByZerox just an info, but https://github.com/faker-js/faker/pull/3540#issuecomment-3016734241 fails with node 18
> But this PR here would also still run with node 18, because it actively still tests it with native cjs support and does not fake a require inside an esm file
>
> Do you still wont like this cts test file change?

That was the argument (or proof of concept) I was waiting for 👍

PR #3452, opened 2025-03-31
docs: add remarks about external sources

Shinigami92:
This PR introduces a new JSDoc tag we can use to mark methods depending on external resources not in control of FakerJS itself.

---

I looked up for several possible JSDoc tags and found [@remarks](https://api-extractor.com/pages/tsdoc/tag_remarks/) to be the one that potentially fits the best. Sadly it is not standardized but at least its listed and used already in some other Projects out there.
ChatGPT also told me after a quick chat that it is one of the best options.


eslint.config.ts
```diff
@@ -172,6 +172,12 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
   {
     name: 'jsdoc overrides',
     rules: {
+      'jsdoc/check-tag-names': [
+        'error',
+        {
+          definedTags: ['remarks'],
+        },
+      ],
```
Shinigami92: **note:** [`structuredTags`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/check-tag-names.md#repos-sticky-header:~:text=of%20settings.jsdoc.-,structuredTags,-automatically%20considered%20as) is defining all the standard tags, so we don't need to

Shinigami92:
@matthewmayer as you are mostly involved in the docs right now, would you like to leave some feedback?

matthewmayer:
Do you intend this to be surfaced in the web docs?

Shinigami92:
> Do you intend this to be surfaced in the web docs?

On the long run yes 👍 
But maybe I need help for that, or someone can help us implementing it there in the doc gen.
For now I would just introduce the baseline for this "feature"

matthewmayer:
Perhaps add a description of what the tag is used for to https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#jsdocs

Shinigami92:
> Perhaps add a description of what the tag is used for to https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#jsdocs

Good idea, I will try to add that in next few days

Shinigami92:
@matthewmayer Added contribution docs and more remarks to the JSDocs (potentially all, but I'm also just a human and might have missed one 😹)

CONTRIBUTING.md
```diff
@@ -296,6 +296,11 @@ function foo(options: { test: string }) {
 </tr>
 </table>
 
+> Further JSDoc tags are in use for specific cases.
+
+- `@internal` - If the method is not intended to be used by the end user, e.g. a helper function
+- `@remarks` - If the method depends on external data not in control of Faker, e.g. a third-party image provider
+
```
ejcheng:
```suggestion
> Other JSDoc tags are in use for specific cases.

- `@internal` - If the method is not intended to be used by the end user, e.g. a helper function.
- `@remarks` - If the method depends on external data not being controlled by Faker, e.g. a third-party image provider.

```

ejcheng (dismissed):

xDivisionByZerox:
I've added some web facing documentation extensions.




I've also renamed the tag to the singular form "remark" in 627a6810945b61571d84cc5fd7d03eb2964296d0. I found it was more reflecting the actual use case and allows the tag to be included multiple times (if so desired). If you do not agree with this change, feel free to bring forward your concerns.

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -145,6 +146,11 @@ function seeAlsoToUrl(see: string): string {
 
     <div v-html="description"></div>
 
+    <div v-if="remark" class="danger custom-block">
```
matthewmayer: i think "danger" color is too scary. Perhaps should be info or tip?

Shinigami92: Fully agree with @matthewmayer 👍
This should be just a hint for the user and is nothing bad.

Shinigami92:
> I've also renamed the tag to the singular form "remark" in 627a6810945b61571d84cc5fd7d03eb2964296d0. I found it was more reflecting the actual use case and allows the tag to be included multiple times (if so desired). If you do not agree with this change, feel free to bring forward your concerns.

As we are already not using a standardized tag anyway, these are some really good arguments 👍

matthewmayer:
Do you think this tag would also be appropriate on the faker.internet.email and faker.phone.number methods to indicate that these may generate emails/phone numbers which belong to real people, so one shouldn't send emails or phone numbers to them. 

I think we mention that elsewhere in the documentation but not directly on these methods.

Shinigami92:
> Do you think this tag would also be appropriate on the faker.internet.email and faker.phone.number methods to indicate that these may generate emails/phone numbers which belong to real people, so one shouldn't send emails or phone numbers to them.
> 
> I think we mention that elsewhere in the documentation but not directly on these methods.

I'm not sure if we should do this. Thoughts I have about this are:
- start small, do not overuse
- we do not use it for `image.personPortrait` as this is in our control

however, if others agree with you @matthewmayer, I wont block that decision 🙂

docs/.vitepress/components/api-docs/method.vue
xDivisionByZerox: Totally fine with me. I was going for "danger" since it has the most contrast in dark mode. I didn't have any deeper thought behind that implementation.
Would you also adjust the title in that case?

Shinigami92: > Would you also adjust the title in that case?

Yeah, not directly sure what, but maybe we already have some other similar blocks like this and then take that over here

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -145,6 +146,11 @@ function seeAlsoToUrl(see: string): string {
 
     <div v-html="description"></div>
 
+    <div v-if="remark" class="tip custom-block">
+      <p class="custom-block-title">Attention!</p>
```
Shinigami92: **suggestion:** 
lets just go with 

```suggestion
      <p class="custom-block-title">Note</p>
```

xDivisionByZerox (approved):

matthewmayer (approved):

PR #3464, opened 2025-04-02
fix(iban): more strict pattern for IE and PS

### LINKED ISSUE #3463: Discrepancies with iban generation for PS and IE
Looks like there have been some changes to isIBAN in `validator` which cause failure

```js
//validator@13.12.0
validator.isIBAN("IE06 0192 5502 8613 0093 62") //true

//validator@13.15.0
validator.isIBAN("IE06 0192 5502 8613 0093 62") //false
```

this is the fix in question https://github.com/validatorjs/validator.js/pull/2518/files

_Originally posted by @matthewmayer in https://github.com/faker-js/faker/issues/3456#issuecomment-2771112952_

xDivisionByZerox:
# Description

This PR fixes #3463.

- #3463

## How to Test

To verify the change you can open the [deploy review](https://deploy-preview-3464.fakerjs.dev/) of the docs and past the following in the dev console:

```js
// load faker
await enableFaker();

// load validator's isIban function
const { default: isIban } = await import('https://unpkg.com/validator@13.15.0/es/lib/isIBAN.js');

// define testing function
function testRandomIban() {
  for (const localeCode of ['IE', 'PS']) {
    for (const iban of faker.helpers.multiple(() => faker.finance.iban({ countryCode: localeCode }), { count: 100 })) {
      if(!isIban(iban)) {
        throw new Error('Found IBAN thats still invalid: ' + iban);
      }
    }
  }

  console.log('Yey! All generated ibans were valid :)')
}

// run the test once
testRandomIban();
```

The expected output is the log that tells you that all ibans were valid. If you want to repeat the test (with newly generated ibans) you can simply call `testRandomIban()` in the dev console again.

### DIFF
--- a/src/modules/finance/iban.ts
+++ b/src/modules/finance/iban.ts
@@ -500,7 +500,7 @@ const iban: Iban = {
       total: 22,
       bban: [
         {
-          type: 'c',
+          type: 'a',
           count: 4,
         },
         {
@@ -862,7 +862,7 @@ const iban: Iban = {
       total: 29,
       bban: [
         {
-          type: 'c',
+          type: 'a',
           count: 4,
         },
         {


Shinigami92 (approved):
This PR changes the used algorithm from type c to a which then does not generate digits anymore

https://github.com/faker-js/faker/blob/048c32581bd5d15f333ef63e4ca968e02793373d/src/modules/finance/index.ts#L921-L929

xDivisionByZerox:
> This PR changes the used algorithm from type c to a which then does not generate digits anymore

I guess you answered that because the implementation wasn't that clear? At least, that's what I was thinking when starting to work on the issue. I'm certain we could update this code (in a separate PR) to make it more readable. I'm also not sure if we need all featured that the Iban data structure provides :shrug:

Shinigami92:
> > This PR changes the used algorithm from type c to a which then does not generate digits anymore
> 
> I guess you answered that because the implementation wasn't that clear? At least, that's what I was thinking when starting to work on the issue. I'm certain we could update this code (in a separate PR) to make it more readable. I'm also not sure if we need all featured that the Iban data structure provides :shrug:
> 

Yes, all good 
It was just and only for clarification 
There is no questionmark or similar on my statement 😄

matthewmayer (approved):

matthewmayer:
> > This PR changes the used algorithm from type c to a which then does not generate digits anymore
> 
> I guess you answered that because the implementation wasn't that clear? At least, that's what I was thinking when starting to work on the issue. I'm certain we could update this code (in a separate PR) to make it more readable. I'm also not sure if we need all featured that the Iban data structure provides :shrug:
> 

I agree the IBAN code is quite incomprehensible. Even some minor changes like defining a constant for "c" and "a" at top of the file would help readability

xDivisionByZerox:
I agree with both of you. I'll work on a second PR to make this more readable. I'm just wondering whether we should consider performance metrics here. While I could immediately think of multiple ways of how the implementation could be written more readable, all would definitely be slower (mostly due to multiple iterations over the expected pattern).
I'm always torn when it comes to performance in the faker library. What do you think? Or should I open a separate issue to discuss this?

Shinigami92:
> [...] I'm always torn when it comes to performance in the faker library. What do you think? Or should I open a separate issue to discuss this?

I'm also most times in favor of readability and maintainability over performance. JS is not the language of choice when it comes to processing millions of data efficiently anyway.

PR #3467, opened 2025-04-06
fix(locale): weight common chinese surnames more frequently

### LINKED ISSUE #3466: using "common last_name" when generate chinese name
# problem

https://github.com/faker-js/faker/blob/next/src/locales/zh_CN/person/last_name.ts

In China, there are over 1,000 last name, but most of them are very rare. In everyday life, commonly seen last name probably number fewer than 200(maybe even less). However, this [list](https://github.com/faker-js/faker/blob/next/src/locales/zh_CN/person/last_name.ts) includes too many last name, most of which are uncommon. As a result, many of the generated names don't even look like a Chinese name at first glance—because the last name are just too rare. Many of the last name in the current list are rarely encountered in real life—some people may go their whole lives without ever meeting someone with one of these last name.

I tried the online demo several times, and each time the generated names felt unnatural. They often included very rare last name that made the names look unrealistic at a glance.

Image

# suggest

王 李 张 刘 陈 杨 黄 赵 吴 周 徐 孙 马 朱 胡 郭 何 林 高 罗 郑 梁 谢 宋 唐 许 邓 韩 冯 曹 彭 曾 肖 田 董 潘 袁 蔡 蒋 余 于 杜 叶 程 魏 苏 吕 丁 任 卢 姚 沈 钟 姜 崔 谭 陆 汪 范 廖 石 金 韦 贾 夏 付 方 邹 熊 白 孟 秦 邱 侯 江 尹 薛 闫 雷 龙 黎 史 陶 贺 毛 段 郝 顾 龚 邵 覃 武 钱 戴 严 莫 孔 常 汤 赖 萧 傅 阎 包 康 伍 施 万 洪 庞 樊 季 庄 殷 温 倪 翟 申 向 齐 乔 文 安 易 颜 牛 岳 简 骆 毕 章 鲁 关 葛 柳 俞 聂 蓝 祝 柴 纪 焦 祁 耿 邢 尚 芦 成 涂 麦 欧阳 上官 皇甫 令狐 诸葛 司徒 司马 申屠 夏侯 慕容 尉迟 长孙

I suggest using only the most common Chinese last name. These last name cover over 85% of the population in China and already represent the most natural and familiar combinations used in everyday life.

matthewmayer:
POC to fix #3466

Weights 20 common Chinese surnames to appear 80% of the time


Shinigami92 (approved):

src/locales/zh_CN/person/last_name.ts
```diff
@@ -1,4 +1,26 @@
 export default {
+  generic_common: [
```
xDivisionByZerox, line 2: I'm not sure about this extra key as it is not included in the module definitions type contract. I'd be in favor to allow (not enforce!) weighted lists in the name schemas in general. I could see this being introduced as a workaround at most for now.

yyz945947732:
[**zh_TW**](https://github.com/faker-js/faker/blob/next/src/locales/zh_TW/person/last_name.ts)  (Taiwan) also only includes a subset of common surnames, and it works quite well. so I think **zh_CN** could do the same — weighting may not be necessary (consider `generic_common` is not a module type). just removing the rare surnames and keeping the common ones might be enough.

If you're aligned with this idea, I'm happy to submit a PR for the change (pick surnames), which would only change the [locals](https://github.com/faker-js/faker/blob/next/src/locales/zh_CN/person/last_name.ts).

matthewmayer:
I'm not 100% sure this is the best approach (versus just using common surnames) though I'd say the fact that generic_common is not in the module definition is not enough to disqualify it as an approach, there are plenty of other edge cases in the locales which use non-standard keys.

For example company.legal_entity_type and company.category used in various definitions and locales.

ST-DDT (approved):

ST-DDT:
For me, both weighting and/or reducing the number of rare last names in that list are valid sulutions.

ST-DDT:
Looks like you have to run preflight on this PR once more.

ST-DDT:
What should we do with this PR? Merge or close?

PR #3482, opened 2025-04-15
test: use validator@13.15.0 with isULID, isISO31661Numeric, isISO15924

matthewmayer:
By upgrading @types/validator to @13.15.0 we can take advantage of the new functions isULID, isISO31661Numeric and isISO15924 from validator to improve and simplify tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "@types/node": "22.14.1",
     "@types/sanitize-html": "2.15.0",
     "@types/semver": "7.7.0",
-    "@types/validator": "13.12.3",
+    "@types/validator": "13.15.0",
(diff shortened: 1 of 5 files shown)


Shinigami92 (approved):

xDivisionByZerox (approved):
Are the checks prior to the validation function still useful? Since validator works with regex under the hood most of the time, I'm not seeing how having two regex is useful for us.

matthewmayer:
> Are the checks prior to the validation function still useful? Since validator works with regex under the hood most of the time, I'm not seeing how having two regex is useful for us.

I don't see much point in removing them

PR #3483, opened 2025-04-15
docs: more than 70 locales

matthewmayer:
Now there are 70+ locales!


Shinigami92 (approved):

xDivisionByZerox (approved):
Thank you for taking the time to update this. I'm not sure if we need to update the docs on every factor 10  milestone, tho. Maybe 100 as the next one?

PR #3485, opened 2025-04-20
fix(locale): ko modified street_name to street_name_part

seoahan:
Fixed street pattern usage for ko locale #2062 
`street_name.ts` modified to `street_name_part.ts`. `street_pattern.ts` correspondingly modified. 

Additionally: 
Updated data to more accurately represent state names and street names.  

Street names: 
Added data to `street_name_part.ts` (formerly `street_name.ts`).
Added and modified data in `street_suffix.ts`. The previous suffixes in `street_suffix.ts` were not suffixes for street names, but rather for zones/areas. Modified the suffixes for accuracy, and added additional suffixes involving numbers as common for Korean street names. 

State names: 
Changes to state names have been moved to PR #3487

PR #3487, opened 2025-04-21
refactor(locale): ko state data update

seoahan:
Split state changes from street changes PR as [requested](https://github.com/faker-js/faker/pull/3485#issuecomment-2817530266). 

Expanded state names to their full names in `state.ts` .


src/locales/ko/location/state.ts
xDivisionByZerox, line 1: Looking at the [list of provinces](https://en.wikipedia.org/wiki/Provinces_of_Korea#Provincial_divisions_since_the_division_of_Korea) (the first-level administrative area) of Korea, I can't seem to find the same list as you provided. 

Can you provide an (preferable English) source for states/provinces so I can cross-reference them?

seoahan: I expanded the names of the states to their full name, not replaced them entirely with new states.  

In english (and often in Korean), the states are usually referred by their shortened names (what I would consider abbreviations). But they do have official "full" names that are used for addresses and such. 
In the list that you provided, if you click on the names of the provinces in the first column, you can see their "full" names on the right hand box under "transcriptions".

For example, this is for the the first province (Busan):
Busan 부산 (釜山)
Metropolitan city
Busan Metropolitan City 부산광역시

"Busan Metropolitan City (부산광역시)" would be the full name of Busan.

xDivisionByZerox: I see. Thank you for explaining this to me.

---

For others that want to verify, I found the [ISO 3166 page for south korea](https://en.wikipedia.org/wiki/ISO_3166-2:KR). There the names are easier to cross reference.

xDivisionByZerox (approved):
The change itself looks good. Although, I noticed that all the states are from south Korea only (as they where before). Technically, the locale `ko` is not bound to south Korea specifically, but all koran speaking countries. We might need to revisit this in the future and introduce a `ko_KR` locale. That way the locale data would more accurately describe its semantic definition. 
Since the entries did exist before this change was proposed, I don't see the necessity to do this immediately.

matthewmayer:
> The change itself looks good. Although, I noticed that all the states are from south Korea only (as they where before). Technically, the locale `ko` is not bound to south Korea specifically, but all koran speaking countries. We might need to revisit this in the future and introduce a `ko_KR` locale. That way the locale data would more accurately describe its semantic definition. 
> 
> Since the entries did exist before this change was proposed, I don't see the necessity to do this immediately.

That is also the case for a bunch of other locales like fr and ru and jp and de. So I agree that's a fight for another day 😀

matthewmayer (approved):

xDivisionByZerox:
Thank you for your contribution @seoahan.
I see that this was your first contribution to our project (that got merged).
Would you mind answering some questions regarding your contribution? This way we can understand the potential struggles or confusions new contributors might have when getting into the project. If you don't want to answer, please ignore this comment:

1. What did motivate you to do your contribution?
1. Did you read our documentation regarding contributing before starting to work on your contribution?
1. How clear was the process of working on your contribution (before the PR) for you? Was there ever a time while working in the project where you didn't know what to do? If so how did you solve your problem?
1. How clear was the process of the code review (PR) for you? Was there ever a time you felt unsure about the current situation?
1. Anything else you'd like to share with the Faker team?

PR #3520, opened 2025-05-29
docs: add info for seeding dates

### LINKED ISSUE #3519: Time of date is not seedable
### Pre-Checks

### Describe the bug

When creating a random date without a refDate and with a seed, the time of the date still changes randomly.

### Minimal reproduction code

```
import { base, de, en, Faker } from "@faker-js/faker";

const faker1 = new Faker({ locale: [de, en, base], seed: 1 });
const faker2 = new Faker({ locale: [de, en, base], seed: 1 });

console.log(faker1.date.past());
setTimeout(() => console.log(faker2.date.past()), 1_000);
// Times of dates will be different by roughly one second
```

### Additional Context

This is I think because `past` just gets the time from `refDate` which is set to now by default:

https://github.com/faker-js/faker/blob/549d71cf3361cca787d685a1c053ef8135e13641/src/simple-faker.ts#L35

### Environment Info

```shell
System:
    OS: macOS 15.4.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 10.81 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - ~/.nvm/versions/node/v22.15.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.15.0/bin/npm
    pnpm: 10.8.1 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 136.0.7103.114
```

- [ ] CJS
- [x] ESM

### Used Package Manager

pnpm

dbartholomae:
Fixes #3519


src/simple-faker.ts
```diff
@@ -150,7 +150,8 @@ export class SimpleFaker {
    * Sets the seed or generates a new one.
    *
    * Please note that generated values are dependent on both the seed and the
-   * number of calls that have been made since it was set.
+   * number of calls that have been made since it was set. If you are using dates,
+   * you will also need to configure them separately.
```
Shinigami92, line 157: Is this potentially duplicated to line 167/168?

README.md
```diff
@@ -126,7 +126,7 @@ export const faker = new Faker({
 
 ## ⚙️ Setting a randomness seed
 
-If you want consistent results, you can set your own seed:
+If you want consistent results, you can set your own seed. There are additional considerations if using faker.date methods, see [Reproducible Results](https://fakerjs.dev/guide/usage.html#reproducible-results).
```
Shinigami92:
```suggestion
If you want consistent results, you can set your own seed. There are additional considerations if using `faker.date` methods, see [Reproducible Results](https://fakerjs.dev/guide/usage.html#reproducible-results).
```

src/simple-faker.ts
dbartholomae: It is, but without it, the current version contradicts what is linked to in lines 167/168, or is at least misleading. But since lines 167/168 exist, I didn't add an explicit link and kept the sentence short.

dbartholomae: I would also say that this is only necessary because we use the term "seed" which carries assumptions with it (like the idea that setting a seed alone already leads to reproducible results), so since we don't fully follow the default usage of the term "seed" we need to be extra specific here.

Shinigami92: if @xDivisionByZerox gives her approval, I will give mine too 👍 (cause otherwise if we do additional commits, the approvals would get stale anyway)

dbartholomae: Thanks! :)

xDivisionByZerox: Apologies for the late review. I was not available during the weekend. I'll take a closer look, later this day.

dbartholomae: Thanks! This is purely documentation, so not super time-critical :)

README.md
```diff
@@ -126,7 +126,7 @@ export const faker = new Faker({
 
 ## ⚙️ Setting a randomness seed
 
-If you want consistent results, you can set your own seed:
+If you want consistent results, you can set your own seed. There are additional considerations if using `faker.date` methods, see [Reproducible Results](https://fakerjs.dev/guide/usage.html#reproducible-results).
```
xDivisionByZerox: Since I'm not native in English I can tell you _why_ exactly, but the sentence feels off. I'd suggest a rephrasing:

```suggestion
If you want consistent results, you can set your own seed. If you are using `faker.date` methods, there are additional considerations. See [Reproducible Results](https://fakerjs.dev/guide/usage.html#reproducible-results).
```

src/simple-faker.ts
xDivisionByZerox: I don't mind the explicit reminder in written form here.

xDivisionByZerox (approved):

pkuczynski (approved):

PR #3529, opened 2025-06-08
chore(deps): switch to prettier-plugin-pkg

Shinigami92:
Together with @JounQin I contributed to an alternative package of `prettier-plugin-packagejson`. It is written in TypeScript and has less sub-dependencies but archiving the same result but with the option (I contributed) to ignore specific keys like `scripts` so it does not forcefully sort things you wont have sorted.

E.g. contributors or maintainers could be ordered in chronological order.

Beside the less sub-dependencies, I personally prefer to read in the scripts section the scripts grouped by a personal favored order. e.g. the preflight is at the bottom, cause it is then easier to pin-point with the eyes instead of searching it between all the other script.
Also test, test:update-snapshots, coverage, integration-test and cypress are grouped together, cause they all contribute to testing.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -4,7 +4,7 @@
  * @type {import('prettier').Config}
  */
 export default {
-  plugins: ['prettier-plugin-organize-imports', 'prettier-plugin-packagejson'],
+  plugins: ['prettier-plugin-organize-imports', 'prettier-plugin-pkg'],
   singleQuote: true,
   trailingComma: 'es5',
   overrides: [
@@ -27,7 +27,9 @@ export default {
     {
       files: 'package.json',
       options: {
+        packageSortOrderPreset: 'npm-plus',
         packageSortOrder: ['name', 'version', 'description', 'scripts'],
+        packageIgnoreSort: ['scripts'],
       },
     },
   ],
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "npm-run-all2": "7.0.2",
     "prettier": "3.5.3",
     "prettier-plugin-organize-imports": "4.1.0",
-    "prettier-plugin-packagejson": "2.5.10",
+    "prettier-plugin-pkg": "0.21.1",
     "rimraf": "5.0.10",
     "sanitize-html": "2.15.0",
     "semver": "7.7.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -77,9 +77,9 @@ importers:
       prettier-plugin-organize-imports:
         specifier: 4.1.0
         version: 4.1.0(prettier@3.5.3)(typescript@5.8.3)(vue-tsc@2.2.8(typescript@5.8.3))
-      prettier-plugin-packagejson:
-        specifier: 2.5.10
-        version: 2.5.10(prettier@3.5.3)
+      prettier-plugin-pkg:
+        specifier: 0.21.1
+        version: 0.21.1(prettier@3.5.3)
       rimraf:
         specifier: 5.0.10
         version: 5.0.10
@@ -702,10 +702,6 @@ packages:
   '@pkgjs/parseargs@0.11.0':
     resolution: {integrity: sha512-+1VkjdD0QBLPodGrJUeqarH8VAIvQODIbwh9XpP5Syisf7YoQgsJKPNFoqqLQlu+VQ/tVSshMR6loPMn8U+dPg==}
     engines: {node: '>=14'}
-
-  '@pkgr/core@0.1.2':
-    resolution: {integrity: sha512-fdDH1LSGfZdTH2sxdpVMw31BanV28K/Gry0cVFxaNP77neJSkd82mM8ErPNYs9e+0O7SdHBLTDzDgwUuy18RnQ==}
-    engines: {node: ^12.20.0 || ^14.18.0 || >=16.0.0}
 
   '@pkgr/core@0.2.4':
     resolution: {integrity: sha512-ROFF39F6ZrnzSUEmQQZUar0Jt4xVoP9WnDRdWwF4NNcXs3xBTLgBUDoOwW141y1jP+S8nahIbdxbFC7IShw9Iw==}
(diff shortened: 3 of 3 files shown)


xDivisionByZerox:
I have problems with applying the argumentation chain from your description to the "problem" at hand. 
Your reasoning regarding script sorting is based on personal preference rather than objective criteria, without presenting a factual conclusion on the matter. The one argument I would agree with is that scripts that belong contextually together, should be put next to each other. But then the question arises for me: _Why don’t they share the same prefix?_ By renaming the scripts to reflect their contextual usage, they would naturally group together:

| Old  Name | New Name |
| ---------- | -----------|
| test | test |
| test:update-snapshots | test:update-snapshots | 
| coverage | test:coverage |
| integration-test | test:integration | 
| cypress | test:cypress or test:e2e | 

That said, I’m unsure why this plugin is necessary in the first place. Why must the fields be rearranged every time the formatter runs? If maintaining a specific structure in `package.json` is important, it can be established once (as seen in #3098) and upheld through processes like code review. Additionally, after a project’s infrastructure is set up, `package.json` typically doesn’t undergo frequent structural changes.

If this PR ties to resolve the "blocking" dependency update PR https://github.com/faker-js/faker/pull/3499#discussion_r2080152633 I'd rather vote to remove the plugin entirely.

Shinigami92:
The plugin is especially there in the first place so that if a dependency (normal or dev) is instantly correctly ordered without thinking about it.

I'm not against grouping them by `test:` prefixes, go for it and do a PR if this makes you happy :)

Potentially we have all our opinions and need to find a way to live with it. But if you have hard broke feelings by a conflict in mind of how sort orders should be or should not be done, then okay, do a PR and remove the plugin completely and I will approve.

However, I'm very unhappy about the forced change within a bugfix release of a plugin going against my opinions.
I just put the energy into contributing into `prettier-plugin-pkg` and I'm happy with the results and use it in packages I fully control. So I already benefit from it.

But anyway, doing something in Faker is becoming always a fight and start to not have hard feelings about it anymore 🤷 

I gave my arguments why I did it, you are not satisfied with it, so be it.

Shinigami92:
@xDivisionByZerox sorry... I hope we can solve the problem somehow. But I gave some arguments that I hope you can understand like the visually eye-scanning for the preflight command.
I just don't like that nearly always you encounter with some vetos where potentially is no problem for you.
I have the problem that it conflicts with that the dependency tries to forcefully sort scripts where the scripts are more important to also be read by humans. And yeah adding prefixes can potentially solve it to some degree, but even then e.g. I would like to have unit tests run before integration tests or e2e tests.
So when we now have e.g. 

```
test:unit
test:coverage
test:integration
test:e2e
```

the plugin would still ignore what humans can thinks and organize into this structure and just forcefully sort it by alphabetical order. And I really dont like that.

But on the other hand, having all the other things sorted automatically like dependencies and files and keywords etc, is helping us maintainers to not further think about it, but let the plugin do its job.

When we want a sorting and/or prefixing change in our scripts, then lets discuss it in a separate thread. I'm fine with that.
But just doing a veto where I put now a day of work into it by contributing fells a bit painful tbh.

xDivisionByZerox:
1. I did not use the word "veto" in any of my comments. The word "veto" is a strong absolute stand when coming to an agreements. So to be very clear I did not and do not **veto** this change.
2. I truly do not understand how having a discussion "becomes always a fight" for you. Like do you always get your way at work? How are you working in a team if not through constructive argumentations on a conceptual level? I am not fighting you. I provide counter arguments to your stand to see if your argumentation can withstand them. If not, your point couldn't have been that great could it? I'm not here to fight with you. I'm here to question if a change does truly improve the project for the users or maintainer or not. 
3. Nobody asked you to do that work. There was no reason, other than your personal matters in disliking a specific sorting in an established package. I agree that changing the sorting of your library in a patch version is off-putting, BUT this library is an opinionated approach on sorting fields in the file. If you do not agree with the opinion, that package is not for you. I know that this will not be as big of an argument, but prettier did this in a non-major version before as well and we still use it 🤷‍♀️

Shinigami92:
> 3. Nobody asked you to do that work. There was no reason, other than your personal matters in disliking a specific sorting in an established package. I agree that changing the sorting of your library in a patch version is off-putting, BUT this library is an opinionated approach on sorting fields in the file. If you do not agree with the opinion, that package is not for you. I know that this will not be as big of an argument, but prettier did this in a non-major version before as well and we still use it 🤷‍♀️

But the new/alternative package now has an option for this, plus it has less sub-dependencies, and it archives the same goal as the other plugin in its prior version where the forcement did not happen.

---

I think the issue I have with your comments is that they just and only come up with counter arguments to check if my already given arguments are valid and stand their point. Instead I would like to get some constructive critic.

However I do not see any blocker between adding the `test:` prefixes AND use the new alternative package.

However removing the old package... this is to me a destructive critic. And maybe this raged my anger.

I assume I misread into the `I'd rather vote to remove the plugin` and mistakenly remembered in my head that you wrote veto somewhere in your text, without rechecking it.
But good that you don't have a veto.

---

tldr conclusion:

my arguments for why doing this change:

+ less sub-dependencies
+ the plugin has its own opinionation by default, but allows to configure it with out opinions
+ the package was written in typescript (definitely something optional)
+ the package is part of the ecosystem/community I live in (definitely something optional)

these are my points, I do not have more

removing the old plugin without a replacement is also not a constructive option to me, cause IMO ...

> That said, I’m unsure why this plugin is necessary in the first place. Why must the fields be rearranged every time the formatter runs? If maintaining a specific structure in package.json is important, it can be established once (as seen in https://github.com/faker-js/faker/pull/3098) and upheld through processes like code review. Additionally, after a project’s infrastructure is set up, package.json typically doesn’t undergo frequent structural changes.

... is not very constructive. We added it so random contributors e.g. do not accidentally put a new devDependency somewhere by hand. sorting the package.json in a given structure reduces marge conflicts.

---

Maybe when I now think more and more about it, one of my main issues was that your comment started with `I have problems with ...` as an introduction. Which instantly puts the whole context into a position of negativity.
Just a tip, you don't need to read it, but maybe it can help: https://blog.wordvice.com/grammar-avoid-double-negatives/
Maybe this helps a bit to understand how it come over to me.

xDivisionByZerox (approved):
As stated in the last team meeting:

I previously had the incorrect assumption that any `npm install` command would automatically sort the dependency fields. If that were the case, the necessity of this PR would be nonexistent for me. Since that isn't the case, I don’t have a strong preference for the specific order and wouldn’t block switching to a different sorting plugin.

Shini pointed out several advantages of the new plugin over the old one. While I don’t fully agree with every argument made, I do recognize that any improvement - even a minor one - is better than staying stagnant.

ejcheng (approved):

PR #3530, opened 2025-06-15
infra: update eslint config

Shinigami92:
While configuring projects at company work, I copied over some huge parts of Faker's structure into the company projects. While doing that, I found out that some parts are not working as intended or outdated.

This PR fixes some issues with out current `eslint.config.ts`


eslint.config.ts
```diff
@@ -1,6 +1,6 @@
 import { includeIgnoreFile } from '@eslint/compat';
 import eslint from '@eslint/js';
-import stylistic from '@stylistic/eslint-plugin';
+import eslintPluginStylistic from '@stylistic/eslint-plugin';
```
Shinigami92, line 3: **note:** unifying naming convention by using `eslintPlugin<PluginName>`

eslint.config.ts
```diff
@@ -38,11 +38,17 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
   },
   //#endregion
 
+  //#region prettier
+  // Prettier recommended should be the first plugin to allow overriding opinionated rules below in any section
+  eslintPluginPrettierRecommended,
```
Shinigami92: **note:** `eslintPluginPrettierRecommended` overrides some basic rules (via eslint-config-prettier) https://github.com/prettier/eslint-config-prettier/blob/4c9489339d37bf96d31e0596e64bb8d4cb4308ef/index.js#L8-L33

When loading the recommended set to late, we would need to override specifically set things down below in another block. But right now we tempted to organize the config in a way that each language, or plugin, or project-specific-part-overrides are in their region-blocks, so they can be easily found by human.

eslint.config.ts
```diff
@@ -147,15 +153,15 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
   //#endregion
 
   //#region unicorn
-  eslintPluginUnicorn.configs['flat/recommended'],
+  eslintPluginUnicorn.configs.recommended,
```
Shinigami92, line 150: **note:** `eslintPluginUnicorn.configs['flat/recommended']` got deprecated and even marked with _strikethrough_ in VSCode

eslint.config.ts
```diff
@@ -147,15 +153,15 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
...
   {
     name: 'unicorn overrides',
     rules: {
       'unicorn/import-style': 'off', // subjective & doesn't do anything for us
       'unicorn/no-array-callback-reference': 'off', // reduces readability
       'unicorn/no-nested-ternary': 'off', // incompatible with prettier
-      'unicorn/no-object-as-default-parameter': 'off', // https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2199
       'unicorn/no-null': 'off', // incompatible with TypeScript
+      'unicorn/no-object-as-default-parameter': 'off', // https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2199
```
Shinigami92, line 158: **note:** rules are sorted in alphabetical order to reduce merge conflicts

eslint.config.ts
```diff
@@ -249,7 +251,7 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
       vitest: eslintPluginVitest,
     },
     rules: {
-      'deprecation/deprecation': 'off',
+      '@typescript-eslint/no-deprecated': 'off',
```
Shinigami92, line 264: **note:** this was a leftover and we don't even have `deprecation` plugin installed anymore. `typescript-eslint` get us covered here.

eslint.config.ts
xDivisionByZerox: The Plugins integration guide clearly states that it should be placed last, or at least after relevant plugins, to get the change to actually override the rules  conflicting with prettier. By moving the rule up in the definition, this utlity is lost. I would put it back at it's orginial position, or even completly at the end.

xDivisionByZerox:
Only one comment. The rest looks fine to me. 👍

eslint.config.ts
Shinigami92: Good recherche 👍 

Then I need to configure that one explicitly at the bottom with an extra comment
https://github.com/prettier/eslint-config-prettier#curly

xDivisionByZerox (approved):

xDivisionByZerox:
Since this PR is Infrastructure related, I'd say we can merge this without a second review before the usual 7 day limit. Dependency PR are handled in a similar manner.

Shinigami92:
I'm fine with that

PR #3536, opened 2025-06-28
infra: stop using node 18

### LINKED ISSUE #3514: Remove Node v18 Support
# Description

NodeJS version 18 has reached the end of its maintenance phase. As usual, we should move away from this version in the next major version of Faker.

- [NodeJS Release Schedule](https://nodejs.org/en/about/previous-releases)

Shinigami92:
This PR does not touch user facing code, but only removes node 18 pipelines and allows to update the rimraf package which is only internally used anyway.
~~The only prepared used facing change was made to the guide and tells the user to use at least node 20~~, but this PR does not yet remove node 18 support, as tsup still targets node 18.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node_version: [18, 20, 22, 24]
+        node_version: [20, 22, 24]
       fail-fast: false
     timeout-minutes: 10
 


docs/guide/index.md
```diff
@@ -15,7 +15,7 @@ This documentation only covers the JavaScript implementation of Faker.
 
 ## Environments
 
-You can run Faker in the browser, or in Node.js. Faker v9.0 requires Node.js version 18 or above. Both ESM and CommonJS imports are available.
```
matthewmayer: This doesn't seem accurate. Could we write something like 

Faker v9 requires Node.js version 18 or above. Both ESM and CommonJS imports are available. Faker v9 will be the last major version to support Node 18, so for future compatibilty it is recommended to use Node 20, 22 or 24.

Shinigami92: I have an inner conflict with this. In the past I can't remember that we changed docs for "older" versions to tell what will happen in future majors.
But on the other side, my suggested change did not changed to much and just encourage users to use supported version of node.
So I think I will just revert this line change fully.

We should finally start freezing maintenance and push work on v10.
Otherwise I feel like we will stay on v9 for several months and block our "major"-version development process.

docs/guide/index.md
```diff
@@ -15,7 +15,7 @@ This documentation only covers the JavaScript implementation of Faker.
...
+You can run Faker in the browser, or in Node.js. Faker v9.0 requires Node.js version 20 or above. Both ESM and CommonJS imports are available.
```
Shinigami92:
```suggestion
You can run Faker in the browser, or in Node.js. Faker v9.0 requires Node.js version 18 or above. Both ESM and CommonJS imports are available.
```

ejcheng (dismissed):

xDivisionByZerox (approved):

PR #3538, opened 2025-06-28
test: add check for const number

Shinigami92:
There was just a confusion about how TS works, so I added a second test to clarify the difference and resolved the TODO


xDivisionByZerox (approved):
I get why this is happening and why the extra test works "as expected" - but doesn't that basically mean that the `const` in the `maybe()` generic is basically doing nothing. 

https://github.com/faker-js/faker/blob/b174135e0921f13cdf5886235cd7e918e5588f32/src/modules/helpers/index.ts#L767

From my understanding the `const` clause (don't know if that is the right term) can never be meat in this specific scenario.

Shinigami92:
> I get why this is happening and why the extra test works "as expected" - but doesn't that basically mean that the `const` in the `maybe()` generic is basically doing nothing.
> 
> https://github.com/faker-js/faker/blob/b174135e0921f13cdf5886235cd7e918e5588f32/src/modules/helpers/index.ts#L767
> 
> From my understanding the `const` clause (don't know if that is the right term) can never be meat in this specific scenario.

it has a different behavior for e.g. arrays
see other tests below

xDivisionByZerox:
> it has a different behavior for e.g. arrays

That's true. But I'm talking about this specific case. The generic is inferring a return type of a function that is passed in. The function itself could be inferred as `const` by typescript. But the return type can never as it. itself would need to statically refer to it's return type as const, from my understanding. That's also what you showed by the second test.

xDivisionByZerox:
Anyways, my question was regarding the implementation of the `maybe` method, not the test that is being fixed here. So I'm merging this.

PR #2869, opened 2024-05-01
docs(web): add contributing pages

xDivisionByZerox:
The goal of this PR is to fix #1513.

Similar to [ESLint's Contribution documentation](https://eslint.org/docs/latest/contribute/), this PR adds a new "Contributing" Section to our Webdocumentation. The overall goal of this (as well as follow up PRs) was to migrate the contents of the CONTRIBUTION file on GitHub with a full fledged documentation on our website. Having the the additional mark up power of vitepress lets us structure these docs in a more abstract why than to dump everything into a single play markdown file.
Since this has already been questioned in the comments, I'll hope to discuss the matter with you in the PR or a team meeting.

My plan is to add an additional "Active Development" page later on. 
This way the rest of the CONTRIUTION section can be addresses in a meaningmul way, without making the "Submit a Pull Request" page more overwhelming than it already is.
I'm open to additional suggestion, tho.

Do not forget to check out the new docs via the [netlify deploy preview](https://github.com/faker-js/faker/pull/2869#issuecomment-2088904691).

---

> [!NOTE] 
> I'll admit that I've used the help of some AI tools to structure the overall file hierarchy. Since I am not native in English, I do hope this reduces the need for spelling and grammar fixes drastically as well.

> [!IMPORTANT]
> My commit messages are marked as "Unverified" right now, since my PC broke down last Thursday. I'm currently on a temporary device which I am to lazy to configure. Sorry for the inconvenience.

PR #3541, opened 2025-06-29
infra: utilize git clean

Shinigami92:
This PR utilizes `git clean` https://git-scm.com/docs/git-clean which is usable in any environment.
Benefits:
1. we don't need `rimraf` at all anymore
2. no untracked files get missed, because git knows what is in `.gitignore`

I switched already to this strategy in some private and some company projects, and it is working really well 👍 
Also if we someday need to exclude anything from the command, it is possible to use [`--exclude`](https://git-scm.com/docs/git-clean#Documentation/git-clean.txt---excludeltpatterngt) (which I also use for e.g. `*.local` so some local untracked env files wont get removed in a company project)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "9.8.0",
   "description": "Generate massive amounts of fake contextual data",
   "scripts": {
-    "clean": "rimraf coverage .eslintcache dist docs/.vitepress/cache docs/.vitepress/dist node_modules",
-    "build:clean": "rimraf dist",
+    "clean": "git clean -fdx --exclude _local",
+    "build:clean": "git clean -fX dist",
     "build:code": "tsup-node",
     "build": "run-s build:clean build:code",
     "generate": "run-s generate:locales generate:api-docs",
@@ -130,7 +130,6 @@
     "prettier": "3.5.3",
     "prettier-plugin-organize-imports": "4.1.0",
     "prettier-plugin-pkg": "0.21.1",
-    "rimraf": "5.0.10",
     "sanitize-html": "2.15.0",
     "semver": "7.7.1",
     "ts-morph": "26.0.0",
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -58,12 +58,6 @@
     {
       "groupName": "doc-dependencies",
       "matchPackageNames": ["ts-morph", "vitepress"]
-    },
-    {
-      // Disable updates for rimraf until Node 18 support is dropped
-      "groupName": "disabled",
-      "matchPackageNames": ["rimraf"],
-      "enabled": false
     }
   ],
   "stopUpdatingLabel": "s: on hold",
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,5 @@ yarn.lock
 
 cypress/screenshots
 cypress/videos
+
+_local/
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -3,8 +3,8 @@
   "version": "9.8.0",
   "description": "Generate massive amounts of fake contextual data",
   "scripts": {
-    "clean": "rimraf coverage .eslintcache dist docs/.vitepress/cache docs/.vitepress/dist node_modules",
-    "build:clean": "rimraf dist",
+    "clean": "git clean -fdx",
+    "build:clean": "git clean -fdx dist",
```
xDivisionByZerox: From the documentation:

> [...] If a <pathspec> is specified, -d is irrelevant; [...]

Since "dist" is given as a path, `-d` is technically not required.

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "9.8.0",
   "description": "Generate massive amounts of fake contextual data",
   "scripts": {
-    "clean": "rimraf coverage .eslintcache dist docs/.vitepress/cache docs/.vitepress/dist node_modules",
-    "build:clean": "rimraf dist",
+    "clean": "git clean -fdx",
```
xDivisionByZerox: > 2. no untracked files get missed, because git knows what is in .gitignore

If I put a new (and by that "untracked") file in the project - for example `scripts/test.ts` - it will get removed by this command. Either I misunderstood what you were trying to say or this is not true. You you might wanted to use the `-X` (uppercase) option here, which will only remove files wich are actually ignored by GH.

---

Anyway, I'm keeping a separate `scripts/custom/` directory with scripts that helped me testing stuff over the years. I put this directory in my local `.git/info/excludes` file. This command will be removed so I'd appreciate if we could specify the paths git should clean explicitly. I also tested that the directory would get indeed get removed - but you can confirm this yourself by:
1. Creating a new untracked file - for example `scripts/custom/test.ts`
2. Put the file (or directory if you created a new one) in your local excludes - `scripts/custom/`
3. Run the clean command
   ```shell
   $ git clean -dfX # even with the uppercase -X
   Removing scripts/custom
   ```

xDivisionByZerox (dismissed):

package.json
Shinigami92:
```suggestion
    "build:clean": "git clean -fx dist",
```

confirm

package.json
Shinigami92: Mh... the only way I currently see would be to add `--exclude scripts/custom`
Should I add this for your case?

Alternative we could also decide for a pattern. e.g. `--exclude **/_local/` 👀

xDivisionByZerox: > Mh... the only way I currently see would be to add --exclude scripts/custom
Should I add this for your case?

Hm, it feels kinda hacky to add this just because I would have a specific use case. One could argue that I can simply not run these npm scripts 🤷‍♀️ 

> Alternative we could also decide for a pattern. e.g. --exclude **/_local/ 👀

This more generic solution sounds much better to me. This would be something we should mention in [the new developer guide](https://github.com/faker-js/faker/pull/2869) tho^^

Would you be fine with adding this exclusion pattern? I would take that solution over not having no option to have excluded local files at all.
This pattern should probably be included in the `.gitignore` file as well in that case.

Shinigami92: updated
please not that it did not work with `-X` but I needed to switch back now to `-x`

tested locally with `-n`

xDivisionByZerox (approved):
lgtm 👍

PR #3560, opened 2025-07-05
refactor(word)!: change default error strategy to 'fail'

### LINKED ISSUE #3263: Change default word filtering strategy to fail
https://github.com/faker-js/faker/blob/d31579b2593de48ea01015ad1cd2c46207a67a94/test/modules/word.spec.ts#L66-L73

IMO if the user specifies a specific length for their words, then by default it should fail if the request cannot be fulfilled.
If the user wants a different behavior, then they can still define an explicit error handling strategy.

Found while looking at: #3261

- #3261

xDivisionByZerox:
This PR fixes issue #3263 

- #3263


Shinigami92 (dismissed):

docs/guide/upgrading.md
```diff
@@ -61,3 +61,22 @@ Some methods do not have exact replacements, so check your code carefully.
 | `faker.image.urlPlaceholder` | `faker.image.dataUri`                                    |
 | `faker.finance.maskedNumber` | See [#3201](https://github.com/faker-js/faker/pull/3201) |
 | `faker.image.avatarLegacy`   | `faker.image.avatar`                                     |
+
+### Word Methods Default Resolution Strategy
+
+The default resolution strategy for the methods in the word module changed to 'fail'.
+This means that method in the word module will fail if no words for your input criteria exist.
```
matthewmayer: Perhapa "will throw an error" instead of "will fail"? To emphasise what "fail" means.

xDivisionByZerox: Good suggestion. That will bring more clarity to the actual runtime behavior.

```suggestion
This means that method in the word module will throw an error if no words for your input criteria exist.
```

Shinigami92 (dismissed):

docs/guide/upgrading.md
```diff
@@ -61,3 +61,22 @@ Some methods do not have exact replacements, so check your code carefully.
 | `faker.image.urlPlaceholder` | `faker.image.dataUri`                                    |
 | `faker.finance.maskedNumber` | See [#3201](https://github.com/faker-js/faker/pull/3201) |
 | `faker.image.avatarLegacy`   | `faker.image.avatar`                                     |
+
+### Word Methods Default Resolution Strategy
+
+The default resolution strategy for the methods in the word module changed to 'fail'.
+This means that method in the word module will throw an error if no words for your input criteria exist.
```
matthewmayer:
```suggestion
This means that methods in the word module will throw an error if no words for your input criteria exist.
```

docs/guide/upgrading.md
```diff
@@ -61,3 +61,22 @@ Some methods do not have exact replacements, so check your code carefully.
...
+
```
matthewmayer, line 76: I think adding a practical example would be helpful:

For example:
```ts
// There are no nouns between 20-25 characters long in the word list
faker.word.noun({length:{min:20, max:25}})
// In v9, this would return a random noun of any length, like 'plastic'
// In v10, this throws an error `[Error]: No words found that match the given length.`
```

xDivisionByZerox: I've added the section but adjusted the error message, since it will be a "FakerError".

Shinigami92 (dismissed):

matthewmayer:
@xDivisionByZerox looks like the example wasnt actually added in 81c8d29 ?

xDivisionByZerox:
Thanks for checking again. I musted have messed up the rebase when pulling the merged next branch. The example should be up now.

Shinigami92 (approved):

matthewmayer (approved):

PR #3583, opened 2025-08-03
feat(phone): mobile style for phone.number

### LINKED ISSUE #3582: Allow generating only mobile numbers for faker.phone.number()

We've had various requests to restrict faker.phone.number() to only return mobile/cellphone numbers e.g. https://github.com/faker-js/faker/discussions/2883 and #3574 

In many locales phone numbers have a fixed format e.g. in en_GB they are always 07##########. It's common for country-specific websites to require a mobile number for signup.

### Suggested solution

A new style "mobile" in the faker.phone.number() method which will return a nationally formatted mobile number with no spaces, e.g. 071234512345 for en_GB or 0812345678 for en_ZA.

In locales where this doesn't make sense e.g. en/en_US where cellphone numbers are indistinguishable from landlines, we would not provide data and this would throw.

### Alternative

There is some existing data stored in src/*/cell_phone/formats for some locales however 
- This is not currently exposed in a method, it's only accessible via faker.fake and is undocumented
- In many cases the data is many years out of date.

The naming could alternatively be say `cell` or `cellphone` or `nationalMobile`

### Additional context

I have added a sample PR at https://github.com/faker-js/faker/pull/3583

matthewmayer:
fix #3582

Note in this initial PR i made the code and test changes, plus migrated over existing data from locales/*/cell_phone - for each locale i manually checked the data formats against wikipedia and updated/simplified where necessary.

Could add additional PRs for other locales later.


test/modules/phone.spec.ts
ST-DDT: Should we change the default en instance to be able to produce mobile numbers?
(including for seeded tests)

ST-DDT (dismissed):

test/modules/phone.spec.ts
xDivisionByZerox: I feel like we should strive to make the default locale be able to be used for all features (knowing that this isn't the case right now).

matthewmayer: USA numbers don't have a way to distinguish landlines and mobile numbers.

matthewmayer: i think it would have to be something like this for consistency with other `en` locale phone_numbers (which are US for historical reasons).

```js
# NANPA does not distinguish mobile phone numbers, so this would include fixed-line phone numbers too
export default ['!##!######']
```

ST-DDT: We provide test data, so as long as it could be a mobile phone number this should be sufficient.

matthewmayer: have added en

ST-DDT (dismissed):

test/modules/phone.spec.ts
Shinigami92: **todo (if-minor):** are we missing a `.it('with mobile style', { style: 'mobile' })` here, or was it omitted intentionally?

matthewmayer: now we added fallback support for en it makes sense to add it

Shinigami92 (approved):

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3585, opened 2025-08-05
docs: split custom Faker instance example

pkuczynski:
Currently two examples are joined and it looks quite confusing.

Screenshot 2025-08-05 at 12 27 39

https://deploy-preview-3585.fakerjs.dev/api/faker.html#faker


pkuczynski:
Initial attempt does not look as intended:

Screenshot 2025-08-05 at 12 56 23

pkuczynski:
Still not perfect, but I think a little bit more clear

image

Is there a way to split examples into two _boxes_?

pkuczynski:
Removing second `@example` tag was the only way to achieve a bit of clarity. Still I think being able to break into two separate boxes would be best:

image

ST-DDT:
> Is there a way to split examples into two _boxes_?

Yes, that would require some changes to the api-docs scripts.

Class-Level:

https://github.com/faker-js/faker/blob/61e4c45ac9176823e5e2a01dc1cfb761d3bb01d5/scripts/apidocs/output/page.ts#L76

Method-Level:

https://github.com/faker-js/faker/blob/61e4c45ac9176823e5e2a01dc1cfb761d3bb01d5/scripts/apidocs/output/page.ts#L195

Note: The later might have impact on the refresh examples button.

matthewmayer:
I think it's a bit tricky to read partly because of the other commented out code. (Like the CJS imports and the seed). So comments are being used both as a header and to show optional alternatives. 

Maybe some of the commented lines could be removed or split to another example.

ST-DDT:
I just revisted this and maybe I have a new idea on how to address this.
Instead of merging the examples with a newline, we could change the joiner to
`\n//-----------------------------------------`.

Then we could put a comment in the first line of each example.

E.g.

// Default Usecase
...
// ------------------------------ added by docs script
// Custom usecase for smaller bundle size
...

Maybe we can even replace --- with an hr in post processing.

What do you think?

xDivisionByZerox:
Superseded by #3817.

- #3817

PR #3616, opened 2025-09-25
docs: instructions to solve v10 compatibility issues with jest

### LINKED ISSUE #3606: Faker v10 ESM-only breaks compatibility with Jest
MAINTAINER EDIT:

> [!NOTE]
> See the [current workaround in this discussion](https://github.com/faker-js/faker/issues/3606#issuecomment-3233612736).

---

**Problem Description**

The recent release of Faker v10 introduced a breaking change by removing CommonJS support entirely, making it ESM-only. This breaks compatibility with many existing TypeScript/JavaScript projects that use Jest and
other testing frameworks in CommonJS environments.

**Real-world Impact**

I'm experiencing this issue in a TypeScript monorepo with the following setup:

- NestJS API (CommonJS-based)
- Database package using Faker in factories for testing
- Jest for unit testing with complex configuration (env-cmd, tsconfig-paths, etc.)

After upgrading to Faker v10, all tests fail with:

Cannot use import statement outside a module
Must use import to load ES Module: @faker-js/faker/dist/index.js

**Technical Analysis**

**Why ESM-only is problematic:**

1. Jest ecosystem compatibility: Jest's ESM support is still experimental and fragile
2. Complex project configurations: Projects with env-cmd, path mapping, and custom setups break
3. Framework limitations: NestJS and many other frameworks still primarily use CommonJS
4. Forced ecosystem migration: Requires migrating entire testing stack, not just faker

**Current workarounds are insufficient:**

- Transform ignore patterns: Doesn't work reliably with complex Jest setups
- Jest ESM experimental: Unstable and breaks other tooling
- Manual mocking: Defeats the purpose of using faker
- Downgrading to v9: Only temporary solution

**Comparison with Other Libraries**

Most mature libraries handle this transition better by maintaining dual package exports:
- TypeScript itself maintains both ESM and CommonJS
- Lodash provides separate packages
- Many testing utilities kept backward compatibility

**Suggested Solution: Restore CommonJS Support**

Provide dual exports like many other libraries:

```
{
  "exports": {
    ".": {
      "import": "./dist/esm/index.js",
      "require": "./dist/cjs/index.js"
    }
  }
}
```

**Business Impact**

**This breaking change affects:**

- Enterprise projects with complex testing setups
- CI/CD pipelines that suddenly fail
- Development teams forced to spend time on tooling instead of features
- Library adoption - teams may switch to alternatives

**Conclusion**

While ESM is the future, forcing adoption by breaking existing projects is counterproductive. The faker library is widely used in testing environments that are inherently conservative and slow to change.

Please consider restoring CommonJS support or at least providing clearer migration paths for complex real-world setups.

---

**Environment:**
- @faker-js/faker: v10.0.0
- TypeScript: 5.9.x
- Jest: 30.0.x
- Node.js: 24.3.x

pkuczynski:


docs/guide/upgrading.md
```diff
@@ -47,11 +47,37 @@ As mentioned in the previous section, CJS can still be used if you use a modern
...
+
+// Exclude from transformation all files in `node_modules`, except `@faker-js`
+transformIgnorePatterns: [
+  // npm
+  'node_modules/(?!@faker-js).+',
+
+  // pnpm
+  'node_modules/.pnpm/.+/node_modules/(?!@faker-js).+'
+],
+```
```
xDivisionByZerox, line 80: The change itself looks fine. I was thinking about representing the code block for the config as a [Code Group](https://vitepress.dev/guide/markdown#code-groups) to better differentiate between npm and pnpm packages. 
BUT, this documentation only represents a migration option for `ts-jest`. By default Jest uses [babel-jest](https://github.com/jestjs/jest/tree/main/packages/babel-jest#setup). Also, the transformer configuration itself is not that important - only that you have one defined for `*.js` files. That's why I would probably replace the "transform" section in a more generic way, like this:

```ts
// Transform both `ts` and `js` files. Defining only `ts` would not be enough, as we also need to transform @faker-js
transform: {
  '^.+\\.(t|j)s$': 'ts-jest', // use the transformer and configuration of your choice
}

// Exclude from transformation all files in `node_modules`, except `@faker-js`
transformIgnorePatterns: [
  // npm
  'node_modules/(?!@faker-js).+',

  // pnpm
  'node_modules/.pnpm/.+/node_modules/(?!@faker-js).+'
],
```

matthewmayer:
Maybe we should try to keep this short and put a longer description and workarounds on https://fakerjs.dev/guide/frameworks.html

xDivisionByZerox:
@pkuczynski are you still interested in providing a solution here?

ST-DDT:
What is the status on this PR?

xDivisionByZerox:
I will close this as stale.

pkuczynski:
Hey @xDivisionByZerox and @ST-DDT - sorry for late reply. Not sure what you wanted me to do here? This documentation change felt complete from my POV?

docs/guide/upgrading.md
```diff
@@ -47,11 +47,37 @@ As mentioned in the previous section, CJS can still be used if you use a modern
...
+
+```ts
+// Transform both `ts` and `js` files. Defining only `ts` would not be enough, as we also need to transform @faker-js
+transform: {
+  '^.+\\.(t|j)s$': 'ts-jest',
+
+  // or when you pass more settings:
+  '^.+\\.(t|j)s$': [
+    'ts-jest',
+    // ... other setttings
```
ST-DDT, line 68: Our playground seems to specify slightly more parameters depending on the setup:

https://github.com/faker-js/playground/blob/main/playgrounds/jest-cjs/jest.config.js

If these are required for using faker, then we should mention them here as well.

Maybe also add a link to the playground itself.

docs/guide/upgrading.md
ST-DDT: I think the code groups would be worth exploring for this case.
Although I would consider them optional, because we haven't used them much elsewhere AFAIK.

docs/guide/upgrading.md
pkuczynski: The setup I posted was a minimal configuration which worked with my project. Not sure, mayne the other settings are more tailored to a different setup or added later, after I created my original PR?

Nonetheless, I linked the playground now...

docs/guide/upgrading.md
pkuczynski: @xDivisionByZerox not sure about it being more generic makes a lot of sense. I think most of the people would be able to easily follow the diff between npm, pnpm, etc. And I do not know how to use code groups? If you show me example, I can certainly do that, although not sure if this is worth the hassle...

ST-DDT (approved):

docs/guide/upgrading.md
xDivisionByZerox: All good. Since this has been open for a while I'm fine with the implementation as is. First get the helping documentation out so people have an easier time. Additional UX features can be added in a later iteration. :)

xDivisionByZerox (approved):

Shinigami92 (approved):

xDivisionByZerox:
> Maybe we should try to keep this short and put a longer description and workarounds on [fakerjs.dev/guide/frameworks.html](https://fakerjs.dev/guide/frameworks.html)

I think this is worth exploring, since new users will most likely check this section instead of the migration guide. I'll provide a follow up PR that provides additional context in the frameworks section.

docs/guide/upgrading.md
pkuczynski: Yeah, I have should been more active on this. Apologies mate... Too much on my plate recently...

xDivisionByZerox: All good. :v:
This more comment on my constant nitpicking towards myself. I'm sorry if you interpreted it as comment towards you.

PR #3617, opened 2025-09-26
feat(locale): add counties & states to nb_NO location

jylep:
Adding missing top-level administrative equivalent of states for Norwegian locale (fylker)

- Adds fallback to counties for states
- Adds counties

https://en.wikipedia.org/wiki/Counties_of_Norway


xDivisionByZerox (commented):
Thank you for your PR @jylep.

I see in the PR description that your intend was to add "top-level administrative" for Norway. I can see in the linked sources that this administrative level is called  "county" in Norway. So I can see why you would put them into the "county.ts" file.

Please take a look at the descriptions of the `county` and `state` property in the location definition:

- https://github.com/faker-js/faker/blob/b547045a55a86f327940852ce8b8df3c3a656109/src/definitions/location.ts#L55-L58
- https://github.com/faker-js/faker/blob/b547045a55a86f327940852ce8b8df3c3a656109/src/definitions/location.ts#L65-L68

As you can see `state` represents the first-level administrative areas while `county` represents the second-level administrative areas. This has sadly been the case for quite some time due to the historic nature of the project heavily defaulting to the US state for structuring locale data sets.

After reading on the [administrative divisions of Norway](https://en.wikipedia.org/wiki/Norway#Administrative_divisions) I would suggest putting the Norwegen "counties" (fylker) into Fakers "state" (first-level) property, while the Norwegen "municipalities" (kommune) would match Fakers "county" (second-level) property.

I can see that this might cause confusion due to the naming collision between Fakers naming definition and the real words application of these words. But, I'd say it is much better in the long run, to stay consistent and honor the specification of a fields when adding locale data to it. I can see the names of these entries change in the future to a more generic term, as we had this exact problem multiple times now.

Please tell me your opinion on the matter and thank you again for taking the initiative and time to making Faker better for anyone by providing these locale data.

matthewmayer:
I tend to agree. For consistency let's put 1at level admin areas in state() and 2nd level in county() regardless of the official name.

jylep:
Added kommuner (municipalities) as well for counties. 👍

matthewmayer (dismissed):

xDivisionByZerox:
The test seem to be failing for multiple reasons:

1. You locale data contain duplicate entries.
2. You didn't update the snapshot tests.
3. The formatter expects a different code styel.

To solve problem 1, it should be enough to regenerate the locales via the npm script `generate:locales`:

```shell
pnpm run generate:locales

# OR

npm run generate:locales
```

For the second problem you need to run the test with the `-u` flag. For this you can use the npm script `test:update-snapshots`:

```shell
pnpm run test:update-snapshots

# OR

npm run test:update-snapshots
```

The third problem can be solved by running the formatter locally:

```shell
pnpm run format

# OR

npm run format
```

@jylep Would you be so kind and run these scripts on your locale setup and commit the resulting changes afterwards? After that, the tests should be fixed.

matthewmayer (dismissed):

ejcheng (dismissed):

jylep:
I tried to run tests locally to see what is causing the failing tests but I got way more errors that are likely unrelated (dynamic import errors)

The CI logs are not really giving me any info on what could be wrong. Just in case I ran the `generate:locales`, `test:update-snapshot` and `format` scripts without any resulting change.

matthewmayer:
Dynamic import errors suggest you may be using an old version of node. 

Faker.js v10 requires a minimum of Node.js v20.19.0, v22.13.0, or v24.0.0.

matthewmayer (approved):

Shinigami92 (approved):

PR #3618, opened 2025-09-27
feat(location): Hungarian support for city name, full street address, continents, countries and directions (hu)

lupus78:
HU Location support for city name, full street address, continents, countries and directions.

includes:

- all countries and continents in hungarian
- directions in hungarina
- all hungarian settlement names as city name
- most commong street names and suffixes
- postal code pattern
- building number pattern


lupus78:
I see the test error. If I'm right, the test thinks the "Ü" is not a valid hungarian character? It is, how do I solve it?

xDivisionByZerox:
> I see the test error. If I'm right, the test thinks the "Ü" is not a valid hungarian character? It is, how do I solve it?

Nearly correct. The letter "Ü" is currently not listed in the list of "used" characters in the locale data. We use these lists to track all characters in a locale and look for "suspicious characters". You do not need to worry, since these lists are autogenerated through snapshot tests. 

All you need to do is run the test with the `-u` flag to tell vitest to update the snapshots. You can do this by running the following command locally:

```shell
pnpm run test:update-snapshots
```

matthewmayer (approved):

ejcheng (approved):

Shinigami92 (approved):

PR #3632, opened 2025-10-11
chore: migrate to eslint defineConfig

Shinigami92:
this will unstuck #3624 
- #3624

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -1,5 +1,4 @@
 import { includeIgnoreFile } from '@eslint/compat';
-import { defineConfig } from '@eslint/config-helpers';
 import eslint from '@eslint/js';
 import eslintPluginStylistic from '@stylistic/eslint-plugin';
 import eslintPluginVitest from '@vitest/eslint-plugin';
@@ -7,6 +6,7 @@ import eslintPluginFileProgress from 'eslint-plugin-file-progress';
 import eslintPluginJsdoc from 'eslint-plugin-jsdoc';
 import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';
 import eslintPluginUnicorn from 'eslint-plugin-unicorn';
+import { defineConfig } from 'eslint/config';
 import { resolve } from 'node:path';
 import tseslint from 'typescript-eslint';
 
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -1,4 +1,6 @@
 import { includeIgnoreFile } from '@eslint/compat';
+import type { ConfigWithExtendsArray } from '@eslint/config-helpers';
+import { defineConfig } from '@eslint/config-helpers';
 import eslint from '@eslint/js';
 import eslintPluginStylistic from '@stylistic/eslint-plugin';
 import eslintPluginVitest from '@vitest/eslint-plugin';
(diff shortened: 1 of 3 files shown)


eslint.config.ts
```diff
@@ -255,6 +257,7 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
     name: 'test/**/*.ts overrides',
     files: ['test/**/*.spec.ts', 'test/**/*.spec.cts', 'test/**/*.spec.d.ts'],
     plugins: {
+      // @ts-expect-error: weird type error
       vitest: eslintPluginVitest,
```
xDivisionByZerox, line 260: This is probably related to https://github.com/vitest-dev/eslint-plugin-vitest/issues/771 🤔

Shinigami92: @xDivisionByZerox  Are you in favor to wait for a fix for this type error? Or can we just merge as is. It's working fine over in node-pg-migrate for a while already.
We will get notified by `@ts-expect-error:` when there is no type error anymore, due to a renovate update.

xDivisionByZerox: > We will get notified by @ts-expect-error: when there is no type error anymore, due to a renovate update.

I'm fine with the current implementation. As you said, since you used `@ts-expect-error` we should be required to track this on our own, but get notified automatically on a later dependency update.

eslint.config.ts
```diff
@@ -51,7 +53,7 @@ const config: ReturnType<typeof tseslint.config> = tseslint.config(
   //#endregion
 
   //#region typescript-eslint
-  ...tseslint.configs.strictTypeChecked,
+  ...(tseslint.configs.strictTypeChecked as ConfigWithExtendsArray),
```
xDivisionByZerox: [According to the ts-eslint docs](https://typescript-eslint.io/users/configs/#strict-type-checked), this config can be provided without spreading. Then the type assertion should not be required:


```suggestion
  tseslint.configs.strictTypeChecked,
```

The resulting configs are the same from my quick testing using eslint's `--print-config` flag.

xDivisionByZerox: That way we don't need the extra dependency.

Shinigami92: The extra dependency is needed anyway, because this is from what eslint's `defineConfig` is coming from

xDivisionByZerox: What? Why that? `defineConfig` is exposed from the `eslint` package under the `eslint/config` path [as by eslint's offical documentation](https://eslint.org/docs/latest/use/configure/configuration-files). 🤔

Shinigami92: This might be new 😲

xDivisionByZerox: The PR https://github.com/eslint/eslint/pull/19487 introduced the `definedConfig` function in the "eslint" package, which was first introduced [v9.22.0 of ESLint](https://github.com/eslint/eslint/releases/tag/v9.22.0) - so about 7 months ago as of writing this comment.

---

I was curious since the way I described it has "always" been the way I've setup ESLint in my enterprise work environment (since the migration to v9 was approved). So this is not a "your way is wrong, my  way is right" kind of thing, just an interesting to know information. :)

xDivisionByZerox (approved):

PR #3648, opened 2025-11-03
feat: Add support for UPC

### LINKED ISSUE #3446: Add a new method faker.commerce.upc() to generate a valid Universal Product Code (UPC).
## Description

A **UPC (Universal Product Code)** is a standardized barcode system used to identify retail products globally. It consists of **12 numeric digits**, where:

- The first **11 digits** are the product identifier.
- The last **digit** is a **check digit**, calculated using the **Modulo 10 algorithm**.

Adding this feature will allow developers to generate fake UPC codes for testing purposes, simulate retail data, and mock API responses.

- More about the UPC: https://en.wikipedia.org/wiki/Universal_Product_Code 

It is helpful for the testing of E-commerce websites or barcode generator systems

### Suggested solution

In module commerce.ts, we could provide the following implementation (basic code need to improve it)

```typescript
export function upc(): string {
  const digits = Array.from({ length: 11 }, () => Math.floor(Math.random() * 10)).join('');
  const checkDigit = calculateCheckDigit(digits);
  return `${digits}${checkDigit}`;
}

function calculateCheckDigit(digits: string): number {
  const sum = digits.split('').reduce((acc, num, idx) => {
    return acc + (idx % 2 === 0 ? parseInt(num) * 3 : parseInt(num));
  }, 0);
  return (10 - (sum % 10)) % 10;
}
```

### Alternative

The alternate method is `isbn()` but it is different from the UPC

### Additional context

_No response_

Dhanush-K-Gowda:
This PR fixes #3446
Have introduced support for [UPC ](https://en.wikipedia.org/wiki/Universal_Product_Code)


src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
...
+  upc(prefix = ''): string {
+    const clean = (prefix || '').replaceAll(/\D/g, '');
+    if (clean.length > 11) {
+      throw new FakerError('Prefix must be at most 11 numeric digits');
+    }
+
+    const remaining = 11 - clean.length;
+    const rand = Array.from({ length: remaining }, () =>
+      this.faker.number.int({ min: 0, max: 9 })
+    ).join('');
```
matthewmayer: You can use faker.string.numeric() for this
 https://fakerjs.dev/api/string.html#numeric

src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
...
+   *
+   * @see upc
+   *
+   * @example
+   * faker.commerce.calculateCheckDigit('12345678901') // 2
+   * faker.commerce.calculateCheckDigit('01234567890') // 5
+   *
+   * @since 10.1.0
+   */
+  calculateCheckDigit(digits: string): number {
```
matthewmayer: this doesn't need to be public. Compare how we handle check digits in isbn()

src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
...
+   *
+   * @throws {FakerError} If `prefix` contains more than 11 digits after sanitization.
+   *
+   * @see calculateCheckDigit
+   *
+   * @example
+   * faker.commerce.upc() // '036000291452'
+   * faker.commerce.upc('01234') // '012345678905'
+   *
+   * @since 10.1.0
```
matthewmayer: next available version is 10.2.0

src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
 
     return data.join(separator);
   }
+
+  /**
+   * Returns a valid [UPC‑A](https://en.wikipedia.org/wiki/Universal_Product_Code) (12 digits).
```
matthewmayer, line 357: Please add test cases to test/modules/commerce.spec.ts to cover all parameters and possible errors

src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
...
+   * @see calculateCheckDigit
+   *
+   * @example
+   * faker.commerce.upc() // '036000291452'
+   * faker.commerce.upc('01234') // '012345678905'
+   *
+   * @since 10.1.0
+   */
+  upc(prefix = ''): string {
+    const clean = (prefix || '').replaceAll(/\D/g, '');
```
matthewmayer: i think throwing an error if the prefix contains non-digits might be clearer (e.g. developer types `faker.commerce.upc({prefix:"X"})` and then is confused why it does nothing)

matthewmayer (changes requested):

src/modules/commerce/index.ts
```diff
@@ -349,4 +351,76 @@ export class CommerceModule extends ModuleBase {
...
   }
+
+  /**
+   * Returns a valid [UPC‑A](https://en.wikipedia.org/wiki/Universal_Product_Code) (12 digits).
+   *
+   * When a `prefix` is provided, it is sanitized to digits and padded with random
+   * digits so that the body has 11 digits. The 12th digit (check digit) is computed
+   * using the Modulo 10 algorithm.
+   *
+   * @param prefix Optional numeric prefix for the UPC body (0–11 digits). Non‑digits are ignored.
```
matthewmayer: in general we prefer options to be in a options object for clarity and to allow the signature to be extended in a backwards compatible way later, ie `faker.commerce.upc({prefix: "123"})` not `faker.commerce.upc("123")`

matthewmayer (changes requested):

src/modules/commerce/index.ts
Dhanush-K-Gowda: sure will implement all the suggested changes soon

matthewmayer:
please update snapshot tests using `pnpm run test -u`

In general you can run `pnpm run preflight` locally to lint and test everything locally to avoid CI errors.

Dhanush-K-Gowda:
> please update snapshot tests using `pnpm run test -u`
> 
> In general you can run `pnpm run preflight` locally to lint and test everything locally to avoid CI errors.

I'm really sorry,  I ran `pnpm run preflight` on my local Ubuntu machine and didn’t encounter any errors. I’m not sure what’s causing these test failures here, but I’ll look into it and follow up shortly.

Dhanush-K-Gowda:
Just realized the CI builds were failing because I hadn’t included the snapshots in the PR. Added them now , should work fine

test/modules/commerce.spec.ts
```diff
@@ -6,6 +6,31 @@ import { times } from './../support/times';
...
+  let idx = 0;
+  for (const digit of body) {
+    const n = Number.parseInt(digit, 10);
+    sum += n * (idx % 2 === 0 ? 3 : 1);
+    idx++;
+  }
+
+  const calculatedCheck = (10 - (sum % 10)) % 10;
+  return calculatedCheck === checkDigit;
+}
```
matthewmayer, line 20: rather than have to reimplement the checksum logic in the test, we could import the `isEAN` function from the validator.js library (similar to we do for isISBN) as if you prefix 0 to a UPC you get an EAN. Then it could simply be:

```
function verifyUPCCheckDigit(upc: string): boolean {
  if (!/^\d{12}$/.test(upc)) {
    return false;
  }

  return isEAN(`0${upc}`);
}
```

test/modules/commerce.spec.ts
```diff
@@ -247,6 +280,195 @@ describe('commerce', () => {
...
+          expect(upc, 'UPC should be exactly 12 digits').toHaveLength(12);
+          expect(upc, 'UPC should contain only digits').toMatch(/^\d{12}$/);
+          expect(
+            upc.startsWith(prefix),
+            'UPC should start with the provided prefix'
+          ).toBe(true);
+          expect(
+            verifyUPCCheckDigit(upc),
+            'UPC check digit should be valid'
+          ).toBe(true);
```
matthewmayer: in this case there's only one possible result, so we could just check the result is '012345678905'?

test/modules/commerce.spec.ts
```diff
@@ -247,6 +280,195 @@ describe('commerce', () => {
...
+        });
+
+        it('should generate different UPCs on multiple calls', () => {
+          const upc1 = faker.commerce.upc();
+          const upc2 = faker.commerce.upc();
+
+          // While it's theoretically possible to get the same UPC twice,
+          // it's highly unlikely with 11 digits of randomness
+          expect(upc1).not.toBe(upc2);
+        });
```
matthewmayer: lets remove this. users shouldn't expect uniqueness from faker methods in general.

test/modules/commerce.spec.ts
```diff
@@ -247,6 +280,195 @@ describe('commerce', () => {
...
+            { count: 100 }
+          );
+
+          for (const upc of results) {
+            expect(upc).toHaveLength(12);
+            expect(upc).toMatch(/^\d{12}$/);
+            expect(upc.startsWith('01234')).toBe(true);
+            expect(verifyUPCCheckDigit(upc)).toBe(true);
+          }
+        });
```
matthewmayer: i think dont need this. calling the methods 100 times is overkill :)

matthewmayer:
CI has passed now, thanks! Just have some nitpicks on the tests.

Shinigami92 (commented):
Made a general review, and thanks a lot @matthewmayer for taking over the detailed review here 👍 
When the points from Matt are addressed, I can give my approval ✅

Dhanush-K-Gowda:
All changes are in. Thanks for checking it out. @matthewmayer @Shinigami92

matthewmayer (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):
Looks good. Thank you for implementing this feature. 👍

PR #3651, opened 2025-11-06
test: remove flaky test in finance.spec.ts

matthewmayer:
e.g. https://github.com/faker-js/faker/actions/runs/18304593051/job/52144979374#step:7:1


xDivisionByZerox (approved):
Works for now 👍

Shinigami92:
> Works for now 👍 

Should we maybe just disable it conditionally on os==win?

xDivisionByZerox:
> Should we maybe just disable it conditionally on os==win?

I would leave it as is. Just fix it at some point. 🤷‍♀️

PR #3654, opened 2025-11-09
docs: add a doc page for each locale

matthewmayer:
Experimental idea: add a doc page for each locale. This would be an easy way to show off how Faker works in various languages, and maybe also be useful for SEO?

Screenshot 2025-11-09 at 19 20 20

Screenshot 2025-11-09 at 19 20 13


Shinigami92 (commented):
Interesting idea. A bummer right now is that when visiting e.g. https://deploy-preview-3654.fakerjs.dev/locales/de.html the left sidebar is gone, and so as a user I loose UX / context.

Shinigami92:
LGTM, but CI error

matthewmayer:
Don't really understand the error

Shinigami92:
> Don't really understand the error



Some scripts that are running are checking and generating stuff in the docs folder. And if there is a diff mismatch, then it fails as there is e.g. a non-committed change.

matthewmayer:
Fixed thanks

Shinigami92 (dismissed):

scripts/generate-locales.ts
```diff
@@ -456,23 +531,26 @@ for (const locale of locales) {
...
+    locale === 'base' ? '`base`' : `[\`${locale}\`](/locales/${locale}.md)`;
+  localizationLocales += `| ${linkedLocale} | ${localeTitle} | \`${localizedFaker}\` |\n`;
 
   promises.push(
     // src/locale/<locale>.ts
     // eslint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
     generateLocaleFile(locale),
 
+    // eslint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
+    generateLocaleDocumentation(locale),
```
xDivisionByZerox, line 553: For consistency - could you add a comment with the output path over the eslint-disable comment?


```suggestion
    // /docs/locales/*.md
    // eslint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
    generateLocaleDocumentation(locale),
```

package.json
```diff
@@ -10,7 +10,7 @@
     "generate": "run-s generate:locales generate:api-docs",
     "generate:api-docs": "tsx ./scripts/apidocs.ts",
     "generate:locales": "tsx ./scripts/generate-locales.ts",
-    "docs:build": "run-s generate:api-docs docs:build:embedded docs:build:run",
+    "docs:build": "run-s generate:api-docs generate:locales docs:build:embedded docs:build:run",
```
xDivisionByZerox, line 13: The "generate:locales" part should be added to the "docs:dev" script as well, no?

docs/.vitepress/components/api-docs/locale.vue
```diff
@@ -0,0 +1,247 @@
...
+function resolveFakerInstances(fakerModule: FakerModule): Map<string, unknown> {
+  const result = new Map<string, unknown>();
+
+  const targetInstance = (fakerModule as Record<string, unknown>)[target];
+  if (targetInstance != null) {
+    result.set(target, targetInstance);
+  }
+
+  return result;
+}
```
xDivisionByZerox: In which case would there be more than one instance that gets resolved? I tried some user tests, but this function always resolved a map with only one instance. If that's always the case, why not return the instance directly?

docs/.vitepress/components/api-docs/locale.vue
```diff
@@ -0,0 +1,247 @@
...
+
+  pendingFrame = requestAnimationFrame(() => {
+    pendingFrame = null;
+    void enhanceCodeBlocks();
+  });
+}
+
+async function enhanceCodeBlocks(): Promise<void> {
+  await nextTick();
+  await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
```
xDivisionByZerox: Why are these ticks required? I tried locally with them being removed and everything seemed to work fine. 🤔

docs/.vitepress/components/api-docs/locale.vue
```diff
@@ -0,0 +1,247 @@
...
+}
+
+function newCommentSpan(content: string): string {
+  return `<span class="comment-delete-marker" style="--shiki-light:#6A737D;--shiki-dark:#6A737D">// ${content}</span>`;
+}
+
+watch(
+  () => route.path,
+  () => scheduleEnhancement()
+);
```
xDivisionByZerox: Does this every trigger? You use the component with `v-once` in the locale pages. 🤔

xDivisionByZerox (commented):
This looks like a great feature. 👍 
I have some questions (sorry, Vue noob here) and suggestions. See comments.

---

Some general extensions I could see being added to this feature in the future:

- use the already existing "RefreshButton" component to reload the examples inline
- use Slots or TemplateRefs for easier/more fine grained post processing
- improve UX by giving the user feedback on which page they are (currently no entry is highlighted on the sidebar)

docs/.vitepress/components/api-docs/locale.vue
matthewmayer: removed

docs/.vitepress/components/api-docs/locale.vue
matthewmayer: removed

docs/.vitepress/components/api-docs/locale.vue
matthewmayer: simplified

package.json
matthewmayer: added

matthewmayer:
i did consider using the RefreshButton but the code seems sufficiently different for it to not really help that much. For example, on the other docs pages we render the results which are already in the code block comments first, and only load the client-side JS on demand to refresh. 

But for locale pages we want to dynamically load on first load of the page.

xDivisionByZerox (approved):

xDivisionByZerox:
I'll merge this merge since Shini previously approved the PR. They are currently on vacation, so they cannot renew their approval.

PR #3656, opened 2025-11-13
feat(locale): add da states (regions)

matthewmayer:
Since 2007 Denmark has had 5 https://en.wikipedia.org/wiki/Regions_of_Denmark#List_of_regions


xDivisionByZerox (commented):
LGTM 👍 

_We could also take a look at [Municipality (Kommuner)](https://en.wikipedia.org/wiki/Municipalities_of_Denmark) for `county` in the future._

xDivisionByZerox (approved):
LGTM 👍 

_We could also take a look at [Municipality (Kommuner)](https://en.wikipedia.org/wiki/Municipalities_of_Denmark) for `county` in the future._

PR #3659, opened 2025-11-15
feat(locale): enhance ta_IN locale with internet, location, and person data

Pradeepg2445:
This PR enhances the `ta_IN` locale with the following additions:

- Added internet definitions (domain_suffix, free_email)
- Added location data (city names, state list, postcode, street patterns)
- Expanded person names (first_name, last_name, patterns)
- Updated locale index to include new modules

All changes pass preflight checks and follow the existing locale structure.


src/locales/ta_IN/person/last_name_pattern.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  generic: [
+    { value: '{{person.last_name.generic}}', weight: 95 },
+    {
+      value: '{{person.last_name.generic}}-{{person.last_name.generic}}',
```
matthewmayer: Are hyphenated surnames like this common in Tamil?

Pradeepg2445: Thanks for the review!

Hyphenated surnames are not common in Tamil. I added the pattern only to follow the structure used in some other locales, but I can remove it to keep the data accurate for ta_IN.

I’ll update the PR and push the changes shortly.

matthewmayer: Ok you can make it 100% on the single last name pattern then.

Pradeepg2445: Thanks again for the review!
I’ve addressed all the pending feedback and pushed the updated changes.
Please let me know if anything else is needed from my side.

ST-DDT:
Sorry for the late reply. Would it be possible to split this PR into smaller chunks? E.g. one per module (person, commerce), one after another?
That would make it way easier to review.

https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#adding-new-locale-or-updating-existing-one

ST-DDT:
Please split this PR into smaller PRs (one per module).

Pradeepg2445:
OK, I will update you as soon as possible.

PR #3660, opened 2025-11-18
fix(locale): ku_kmr_latin endonym should be in Latin script

mazyar-kawa02:


matthewmayer:
Please rerun 

pnpm run preflight 

And commit the changes.

mazyar-kawa02:
@matthewmayer Could you check now? Thanks for everything!

matthewmayer (approved):

xDivisionByZerox (approved):
Since this is the "*_latin" variant of the locale, the endonym should also be in latin script. Good Catch 👍

PR #3664, opened 2025-11-26
fix(locale): remove some negative hu words

matthewmayer:
per https://github.com/faker-js/faker/discussions/1631

ChatGPT says:

**parasztország** — derogatory; literally “peasant country,” used to insult a country as backward or uncultured.

**parasztos** — context-dependent; literally “peasant-like.” Neutral when referring to traditional/folk style, but derogatory when describing someone’s behavior as rude or uncouth.

**retardált** — strongly derogatory; literally “retarded.” Offensive slang for someone perceived as intellectually disabled.


xDivisionByZerox (approved):
Fine by me. These lists contain enough words, so removing these won't hurt even if it turns out in the future, that the contextual negativity wasn't that bad.

PR #3673, opened 2025-12-14
docs: fix lack of contrast in developer console log message in light mode

### LINKED ISSUE #3672: Lack of contrast in developer console log message in light mode
https://github.com/faker-js/faker/blob/1cf3991c180ddeb75253d1850f085147a04ee3bf/docs/.vitepress/config.ts

At least in Chrome, the color choices for the `If you would like to test Faker in the browser console, you can do so using 'await enableFaker()'` log in the console is very hard to read due to lack of contrast.

Image 

```
const logStyle = 'background: rgba(16, 183, 127, 0.14); color: rgba(255, 255, 245, 0.86); padding: 0.5rem; display: inline-block;';
```

matthewmayer:
fix #3672

light mode
Screenshot 2025-12-14 at 20 21 28

dark mode
Screenshot 2025-12-14 at 20 21 51

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -151,7 +151,7 @@ const config = defineConfig({
       'script',
       { id: 'browser-console-faker' },
       `
-const logStyle = 'background: rgba(16, 183, 127, 0.14); color: rgba(255, 255, 245, 0.86); padding: 0.5rem; display: inline-block;';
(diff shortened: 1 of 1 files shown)


xDivisionByZerox (approved):
This seems to work. Thank you :+1:

Shinigami92 (approved):

PR #3674, opened 2025-12-18
docs: add sponsors

Shinigami92:
Adding a sponsoring section to honor partners.
~~It is not final yet in which groups we split them.~~


Shinigami92:
**Team Decision:** @Shinigami92 will try out the https://graphql-docs-v2.opencollective.com/access and dynamically query in vitepress clientside the active sponsors and cache that for ~1 month clientside, so the limit of 10r/min are not exceeded

matthewmayer:
I don't see anything new on the deploy preview?

Shinigami92:
> I don't see anything new on the deploy preview?

lower right corner

ST-DDT:
In case this doesn't load, we should handle it more gracefully.

- maybe silence the error
- and maybe hide the sponsor header (since it is empty).

empty

---

Also the initial load looks different than later updates:

| Initial | After |
| --- | --- |
| initial | later |


(fixes itself on window resize)

xDivisionByZerox (approved):
Looks good enough for now. We can iterate in the future if we desire to do so.

docs/.vitepress/components/AsideSponsors.vue
```diff
@@ -0,0 +1,107 @@
...
+
+const sponsors = computed<Sponsor[]>(
+  () =>
+    backersStorage.value?.backers
+      .map((backer) => ({
+        name: backer.account.name ?? backer.account.slug,
+        img: backer.account.imageUrl ?? '',
+        url: `https://opencollective.com/${backer.account.slug}`,
+      }))
+      .slice(0, 10) || []
```
ST-DDT, line 42: Please filter/reduce the data before writing it to local storage.

ST-DDT (changes requested):
This bug still exists:

- https://github.com/faker-js/faker/pull/3674#issuecomment-4158473259

docs/.vitepress/components/AsideSponsors.vue
```diff
@@ -0,0 +1,107 @@
...
+  lastUpdated: 0,
+  backers: [],
+});
+
+const sponsors = computed<Sponsor[]>(
+  () =>
+    backersStorage.value?.backers
+      .map((backer) => ({
+        name: backer.account.name ?? backer.account.slug,
+        img: backer.account.imageUrl ?? '',
```
ST-DDT: Also, we should filter accounts without an image:

Image

docs/.vitepress/components/AsideSponsors.vue
Shinigami92: I stored the full data on purpose, so if we need to adjust data or want to display more or less later, we don't have a breaking change but just can access the cached data as is.

ST-DDT: I don't expect the used data to change. But 🤷

docs/.vitepress/components/AsideSponsors.vue
```diff
@@ -0,0 +1,117 @@
...
+    now.getMonth() !== lastUpdated.getMonth()
+  ) {
+    try {
+      const backers = await fetchBackers();
+      backersStorage.value = {
+        lastUpdated: Date.now(),
+        backers,
+      };
+    } catch {
+      // Silently ignore fetch errors; the section will hide itself when empty.
```
ST-DDT: Maybe cache the failure, to avoid using up all api limits.

ST-DDT (dismissed):
Technically ok, but not really my preference to have.

docs/.vitepress/components/AsideSponsors.vue
Shinigami92: added a backoff
https://graphql-docs-v2.opencollective.com/access#authentication

docs/.vitepress/components/AsideSponsors.vue
```diff
@@ -0,0 +1,130 @@
...
+      backersStorage.value = {
+        lastUpdated: Date.now(),
+        backers,
+      };
+    } catch {
+      // Cache the failure so we don't burn the OpenCollective quota
+      // on every page load while the API is unavailable.
+      backersStorage.value = {
+        ...backersStorage.value,
+        lastFailedAt: Date.now(),
```
ST-DDT: Maybe merge lastUpdated and lastFailed into updateAfter/expiresAt. That will also simplify the check on each refresh.

If (now > updateAfter)

Shinigami92: Good idea, the check would definitely get simpler. Before I refactor, I want to make sure we agree on the tradeoff:

**Current shape (`lastUpdated` + `lastFailedAt`):** stores *facts* — when we last fetched, when we last failed. Refresh policy is computed from those facts on each load.

**Proposed shape (`refreshAfter` / `expiresAt`):** stores *the decision* — when we may next try. One comparison, fewer lines.

What we'd lose by collapsing them:

- **Debuggability** — popping open DevTools no longer tells you "last fetched 12 days ago, last failed 30s ago"; you only see when the next attempt is scheduled.
- **Policy flexibility** — if we ever change the cadence (different backoff, weekly instead of monthly), cached entries would still encode the *old* `refreshAfter` until the next attempt overwrites them. With facts-in-storage we'd recompute fresh policy on every load.
- **"Start of next month" becomes explicit** — today the `getMonth() !== lastUpdated.getMonth()` check is fuzzy and TZ-agnostic. With `refreshAfter` we have to pick: local TZ start-of-month? UTC? "now + 30 days"?

For a docs-site sponsor widget where the cache lifetime is short and we're unlikely to debug cache forensics or change the refresh cadence, I think the simpler check is the better call. But it's your taste preference more than mine, so happy to go either way — should I refactor?

ST-DDT: This thing is getting way to complicated for what it is supposed to do, which adds to my dislike of this feature.

I prefer short and simple (expiresAt)

- I can tweak the local storage to not display it for me ever
- It is slighly more performant, not that it matters

Shinigami92: I feel like you making it complicated, I was asking for if we can just iterate and/or live with the previous solution, but you keep adding and adding change requests / suggestions.

I gave this answer on purpose, because you are not thinking about up and downsides (or at least I don't get the feeling you do.
I even just found out that the request rate-limit is 10 per minute. So IMO we are already overengineering this...

Shinigami92: tbh... I just reverted my last commit now and moved back to previous implementation. I feel like we are overengineering here by far to much for just a simple thing.
If you want to improve it, I would like to kindly ask you open up a follow-up PR for another iteration.

xDivisionByZerox (approved):
I repeat my previous review.

PR #3678, opened 2025-12-20
feat(finance): add IR iban

ds1371dani:
Currently the finance > iban is missing IR locale

after i added the patterns, my test didn't pass because of validator.js isIBAN validation (it had wrong pattern)

validator.js was fixed in this [PR](https://github.com/validatorjs/validator.js/pull/2634) and so the validator package in faker was also updated


xDivisionByZerox:
Hi @ds1371dani. Thanks for this contribution :raised_hands:.

We are handing development on the "next" branch. Please only provide PRs against it. I did go ahead and updated the target branch. This introduced conflicts in the package file, since the next branch is already using version 13.15.23 of validator. Would you be so kind and resolve the conflicts by merging "origin/next" into your branch or rebasing your branch onto "origin/next"?
Although, I'm trying to get renovate to update the package file on "next" right now, to include this dependency update separately. Thanks for bringing this valuable validator.js update to our attention.

Also, I se that the snapshot tests are failing. Would you be so kind and resolve that issue as well? You can find infos about this problem in our [contribution guide on our website](https://fakerjs.dev/contributing/submit-a-pull-request.html#step-4-run-preflight-checks). To resolve this concrete issue all you need to run is `pnpm run test:update-snapshots`.

xDivisionByZerox:
I worked though renovate requirements and let it create #3680 to update validator.js - that you don't need to resolve the conflicts and can simply revert the dependency update commit.

ds1371dani:
Thanks @xDivisionByZerox for your quick answer, i rebased onto "origin/next" branch and also ran "preflight" script and rolled back validator package version

I initially made my PR against v10 branch because i preferred this update to be included in the current major version. after merging into next branch when will it be released?

xDivisionByZerox:
> I initially made my PR against v10 branch because i preferred this update to be included in the current major version. after merging into next branch when will it be released?

This will happen with the next release. We rarely do backport, only for docs and security related issues.
We do not have a date for the next release in mind, but in the team meeting this week (18th of December) we agreed that we want to schedule another release soon. I'll be looking through open PRs and merge the once that are complete, so that we might be able to bring out another version before the end of the year.

I sadly can't give you a more concrete day. Please remember that the entire Faker team is doing this project in their free time and likely wants to spend time with their family and friends just as anyone else. Thanks for understanding :heart_hands:

test/modules/finance-iban.spec.ts
```diff
@@ -377,6 +377,31 @@ describe('finance_iban', () => {
...
+
+          expect(
+            26,
+            `IR IBAN would be 26 chars length, given is ${iban.length}`
+          ).toBe(iban.length);
+
+          expect(
+            32,
+            `IR formatted IBAN would be 32 chars length, given is ${ibanFormated.length}`
+          ).toBe(ibanFormated.length);
```
xDivisionByZerox, line 398: **For other reviewer:**

Usually I would suggest the following here: 

>By using the `toHaveLength()` matcher, you do not need to provide an additional error message, since the matcher already take care of that. This would also align more closely with other test setups."
>
>```suggestion
>          expect(iban).toHaveLength(26);
>          expect(ibanFormated).toHaveLength(32);
>```

But looking at the other test implementations for iban, this is not the case. So we should stay consistent with the other test in this file for now.
We should probably update the test file to use more self descripting matchers in a separate PR. By doing this we benefit from better, "native" error messages and a more coherent usage of matchers across all test files.

xDivisionByZerox (approved):
Direct wikipeda link for other reviewers to verify the schema:

- https://en.wikipedia.org/wiki/International_Bank_Account_Number#IBAN_formats_by_country (under the first table in the section "Aspirational country codes for International Bank Account Number")

PR #3692, opened 2026-01-07
docs: update emoji for Date feature in README

ViKing-py:

PR #3693, opened 2026-01-08
Minor improvemnts

ViKing-py:


xDivisionByZerox:
What exactly is the improvement here? You simply rewrote the existing code using slightly different syntax - the overall use-case stays identical. I would have expected an analysis + suggestion in the PR description if you provide an "improvement" without a prior issue. Please provide this information otherwise I'm going to have to close your PR.

Furthermore, you already provided a PR in which you suggest changing the icon (FFR see #3692). This does not need to be in this PR as well.

ST-DDT:
This PR seems to be abandonned -> Closing.
Feel free to reach out, if you want to continue this.

PR #3701, opened 2026-01-22
feat(string): support uuid v7

### LINKED ISSUE #3699: Support for UUIDv7 Generation

As a developer using Faker I would like the ability to generate compliant UUIDv7 values that I can use with Zod (https://zod.dev). I noticed the issue raised about `ULID` (https://github.com/faker-js/faker/issues/2648) in which organisation members were open to adding the functionality when the requirement was raised. (https://github.com/faker-js/faker/issues/2648#issuecomment-1936328128) (https://github.com/faker-js/faker/issues/2648#issuecomment-1936322810)

### Suggested solution

Introduction of `faker.string.uuidv7()` or `faker.string.uuid({ version: 7 })` or similar within the `string` module.

### Alternative

I have considered extending the package with a custom implementation but feel that this would be beneficial to the wider community.

### Additional context

UUIDv7 Specification: https://www.rfc-editor.org/rfc/rfc9562#name-uuid-version-7

xDivisionByZerox:
# Description

This PR solves #3699.

The  PR introduces support for generating UUID v7 values via `faker.string.uuid({ version: 7 })`. This extends the existing function with a time ordered variant based on the Unix timestamp. To preserve the current behavior, an explicit signature for `faker.string.uuid({ version: 4 })` (the default) has also been added.

Since UUID v7 is time‑based, the new API allows providing a reference date (`refDate`) argument as well.

A few implementation choices differ slightly from the usual patterns in this project:

- The core functionality is implemented in standalone functions that are then used by the module. This approach has several advantages:
  - It keeps both the UUID algorithm and the public API easier to read.
  - It makes the public API more naturally extensible for future UUID versions.
  - It simplifies a potential future transition toward standalone functions (see #2667).
- The UUID API now exposes multiple signatures. The goal is to make the relationship between the `refDate` argument and the selected version clearer. Another benefit is that the default signature can retain its original `since` tag (v8.0.0).  
  For documentation generation to work correctly, an additional signature mirroring the implementation is required, and this one must appear last. You can verify this in the [documentation preview for this PR](https://fakerjs.dev#TODO).

## Link

- #3699
- [RFC 9562 - UUID version 7](https://www.rfc-editor.org/rfc/rfc9562#name-uuid-version-7)


src/modules/string/index.ts
```diff
@@ -692,18 +693,105 @@ export class StringModule extends SimpleModuleBase {
...
+     *
+     * @default faker.defaultRefDate()
+     */
+    refDate: string | Date | number;
+  }): string;
+  /**
+   * Returns a UUID ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
+   *
+   * @param options An optional options object.
+   * @param options.version  The specific UUID version to use. Defaults to `4`.
```
Shinigami92: **chore:** 

```suggestion
   * @param options.version The specific UUID version to use. Defaults to `4`.
```

src/modules/string/index.ts
```diff
@@ -692,18 +693,105 @@ export class StringModule extends SimpleModuleBase {
...
+    /**
+     * The specific UUID version to use.
+     */
+    version: 4;
+  }): string;
+  /**
+   * Returns a UUID v7 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
+   *
+   * @param options An options object.
+   * @param options.version  The specific UUID version to use.
```
Shinigami92: **chore:** 

```suggestion
   * @param options.version The specific UUID version to use.
```

src/modules/string/index.ts
```diff
@@ -692,18 +693,105 @@ export class StringModule extends SimpleModuleBase {
...
+   * faker.string.uuid() // '4136cd0b-d90b-4af7-b485-5d1ded8db252'
+   *
+   * @since 8.0.0
+   */
+  uuid(): string;
   /**
    * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
    *
+   * @param options An options object.
+   * @param options.version  The specific UUID version to use.
```
Shinigami92: **chore:** 

```suggestion
   * @param options.version The specific UUID version to use.
```

Shinigami92 (dismissed):

src/modules/string/index.ts
```diff
@@ -692,18 +693,105 @@ export class StringModule extends SimpleModuleBase {
...
+     */
+    refDate: string | Date | number;
+  }): string;
+  /**
+   * Returns a UUID ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
+   *
+   * @param options An optional options object.
+   * @param options.version The specific UUID version to use. Defaults to `4`.
+   * @param options.refDate The timestamp to encode into the UUID.
+   * This parameter is only relevant for time based UUID's.
```
matthewmayer:
```suggestion
   * This parameter is only relevant for UUID v7.
```

matthewmayer (commented):
should we also mention this as one of the methods listed in 
https://fakerjs.dev/guide/usage#reproducible-results

src/modules/string/index.ts
xDivisionByZerox: UUID v1 and v6 are also time based. I wanted to write it in a way that immediately would be correct for future implementations as well. But I guess since we only offer v7 for now we can be more explicit.

xDivisionByZerox:
> should we also mention this as one of the methods listed in [fakerjs.dev/guide/usage#reproducible-results](https://fakerjs.dev/guide/usage#reproducible-results)

I don't think this is necessary, since these are only examples and not a comprehensible list. Or did you mean to replace one of the examples from the date module with this method, to get more diversity in the examples?

matthewmayer:
> > should we also mention this as one of the methods listed in [fakerjs.dev/guide/usage#reproducible-results](https://fakerjs.dev/guide/usage#reproducible-results)
> 
> 
> 
> I don't think this is necessary, since these are only examples and not a comprehensible list. Or did you mean to replace one of the examples from the date module with this method, to get more diversity in the examples?

Yeah good idea

src/modules/string/index.ts
```diff
@@ -692,18 +693,105 @@ export class StringModule extends SimpleModuleBase {
...
-      .replaceAll('x', () => this.faker.number.hex({ min: 0x0, max: 0xf }))
-      .replaceAll('y', () => this.faker.number.hex({ min: 0x8, max: 0xb }));
+  uuid(options?: {
+    /**
+     * The specific UUID version to use.
+     */
+    version?: 4 | 7;
+    /**
+     * The timestamp to encode into the UUID.
+     * This parameter is only relevant for time based UUID's.
```
matthewmayer: This one needs a fix too for consistency.

xDivisionByZerox: You are right. You gave me a suggestion block and I took the easy way of simply committing it inside of GH without looking around first. I'll do better next time. :)

Shinigami92 (approved):

matthewmayer (approved):

Issue #3325, opened 2024-12-05
Add automated tests for fake patterns


ST-DDT:
### Clear and concise description of the problem

If we add or change fake patterns (or have to review them), we have to manually check for all possible combinations of patterns.

### Suggested solution

Add an automated test that checks fake patterns for errors including errors in nested fake patterns.

### Alternative

Keep testing by hand.

### Additional context

We have to find a way to detect all fake patterns even if they aren't suffixed with pattern.

Found in: https://github.com/faker-js/faker/pull/3324/commits/bfbd0b608fdcf05966cc1f286d315b9d5f7ac26e

- https://github.com/faker-js/faker/pull/3324/commits/bfbd0b608fdcf05966cc1f286d315b9d5f7ac26e

matthewmayer:
Sounds useful; though what errors specifically would you test for? 

The typo i made could have been caught by checking for a balanced number of } and { characters, but not sure how common that would be.

atzzCokeK:
@ST-DDT @matthewmayer Hi! I'd like to work on this issue if this issue is still alive. I did some investigation and benchmarking.

Pattern detection: Recursively walk all locale definition values and collect every string containing {{. This finds 472 unique patterns across all locales, regardless of file naming.

## Two types of patterns exist:

- 431 patterns use {{module.method}} syntax, resolved via helpers.fake()
- 41 patterns use {{keyword}} syntax, resolved via helpers.mustache() with a manual data map

The mustache()-based patterns are found in two places:
- hacker.phrase — uses {{verb}}, {{noun}}, etc. This could potentially be refactored to fake()-style ({{hacker.verb}}) as a separate effort.
- person.fullName — uses {{person.firstName}}, {{person.lastName}}, etc. This might not be replaced with fake() easily because it injects caller-provided values (e.g. a specific firstName or sex passed as options).

## Proposed test (two layers):

1. {/} balance check on all patterns (~1.4ms) — catches brace typos like #3324
faker.helpers.fake() execution on the 431 fake-style patterns, per locale (~21ms) — also catches invalid method refs and bad args
2. The mustache-style patterns would only get layer 1 for now. Both layers together run in under 25ms — negligible CI impact.

Does this approach sound good?

atzzCokeK:
In hacker.phrase, the patterns could be refactored from `{{verb}}` to `{{hacker.verb}}` , which would unify the resolution method. Would a separate PR for that be welcome? If an issue would be preferred, I'm happy to create one instead.

[...]

PR #3736, opened 2026-02-07
refactor(hacker): use helpers.fake() instead of helpers.mustache() in phrase()

atzzCokeK:
## Description
Refactors `hacker.phrase()` to use `helpers.fake()` instead of `helpers.mustache()`.

## Context
`hacker.phrase()` used a manual data map to resolve placeholders like `{{verb}}`. But since `helpers.fake()` can resolve `{{hacker.verb}}` directly, the manual mapping is unnecessary.

(`person.fullName()` still uses `mustache()` — it needs to pass in caller-provided values, so that one stays as-is.)

Related: https://github.com/faker-js/faker/issues/3325#issuecomment-3863590282

### Changes

- Locale data (en, eo, nl, ru, zh_CN): `{{verb}}` → `{{hacker.verb}}`, etc.
- `phrase()`: removed data map, now just calls `helpers.fake()`
- Updated JSDoc and snapshots


matthewmayer (dismissed):

ST-DDT (dismissed):

src/definitions/hacker.ts
```diff
@@ -26,10 +26,10 @@ export type HackerDefinition = LocaleEntry<{
 
   /**
    * Some phrases that will be injected with random hacker words.
-   * May use any of the HackerDefinition keys wrapped in double braces
-   * (e.g. `I'm {{ingverb}} {{adjective}} {{noun}}`).
+   * May use any of the hacker module methods prefixed with `hacker.` wrapped in double braces
+   * (e.g. `I'm {{hacker.ingverb}} {{hacker.adjective}} {{hacker.noun}}`).
```
xDivisionByZerox: I would align this with descriptions of other fake patterns:

```suggestion
   * A list of faker patterns to generate a hacker phrase.
```

Some other descriptions in the definitions that represent fake pattern lists:

- https://github.com/faker-js/faker/blob/2a4b15cff44cec578b23b579d9bd8603887024ac/src/definitions/company.ts#L32-L35
- https://github.com/faker-js/faker/blob/2a4b15cff44cec578b23b579d9bd8603887024ac/src/definitions/food.ts#L9-L12
- https://github.com/faker-js/faker/blob/2a4b15cff44cec578b23b579d9bd8603887024ac/src/definitions/location.ts#L105-L108
- https://github.com/faker-js/faker/blob/2a4b15cff44cec578b23b579d9bd8603887024ac/src/definitions/location.ts#L125-L138

xDivisionByZerox (commented):
Only one small suggestion. The other change looks good. It should reduce the overall complexity, where different locale patterns are resolved in many different ways.

src/definitions/hacker.ts
atzzCokeK: THX, Great suggestion!:>

xDivisionByZerox: It seems like I set the lines for the suggestion incorrect - apologies.
I meant to actually replace the entire description with my suggestion. Mentioning "methods prefixed with `hacker.`" is a little misleading since all modules can actually be used with `fake()`.

atzzCokeK: 416f2c7d611b1b3cb7fb54b92dae482b0e9b65fb

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3742, opened 2026-02-15
feat: append korean names

i-am-a-dolphin:
Ref: https://yiunsr.tistory.com/885


ST-DDT:
Hi @i-am-a-dolphin ,

Faker usually limits lists of names to ~1k entries to save memory/disk space a bit.
Is there a reason why you would add all names you have obtained prgrammatically?
(if the current set needs to be refreshed, then thats a valid reason, we should just stick to ~1k names)

Also, have you checked whether that particular source may be subject to copyright/ownership issues?
(the names itself aren't copyright-able, but specific collections might be without own transformation)

ST-DDT:
@i-am-a-dolphin Are you still interested in this PR?

ST-DDT:
This PR appears to have gotten stale. If you are stil interested in this please let us know and we can reopen this one.

PR #3743, opened 2026-02-16
infra: trusted publishing via OIDC token

### LINKED ISSUE #3593: Switch to trusted publishing for npm
Like https://github.com/prettier/eslint-plugin-prettier/pull/757 we should switch to [Trusted publishing for npm packages](https://docs.npmjs.com/trusted-publishers)

Shinigami92:
Remove manual token, switch to real OIDC token

⚠️ We need to find out if dist-tag still works 🤔 

> Edit @xDivisionByZerox => it doesn't

if this really works, we can remove the npm token completely

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -46,16 +46,9 @@ jobs:
           echo -e "$(grep -P '## 📙 API Documentation' -B 10000 README.md)\n\n- [Getting Started Guide](https://fakerjs.dev/guide/)\n- [API Reference](https://fakerjs.dev/api/)\n\n$(grep -P '## 🚀 Features' -A 10000 README.md)" > README.md
           sed -i "s|/fakerjs.dev/|/v$RELEASE_MAJOR.fakerjs.dev/|g" README.md
 
-      - name: Set publishing config
-        run: pnpm config set '//registry.npmjs.org/:_authToken' "${NPM_AUTH_TOKEN}"
-        env:
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
-
       - name: Publish
         run: |
-          pnpm publish --tag next --no-git-checks
-        env:
-          NPM_CONFIG_PROVENANCE: true
+          pnpm publish --access public --tag next --no-git-checks --provenance
 
       - name: Set latest/next dist-tag
         run: |
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -38,8 +38,6 @@ jobs:
           echo "RELEASE_VERSION=$RELEASE_VERSION" >> $GITHUB_ENV
           RELEASE_MAJOR=$(jq -r '.version | split(".")[0]' package.json)
           echo "RELEASE_MAJOR=$RELEASE_MAJOR" >> $GITHUB_ENV
-          DIST_TAG=$(jq -r '.version | if split("-")[1] == null then "latest" else "next" end' package.json)
-          echo "DIST_TAG=$DIST_TAG" >> $GITHUB_ENV
 
       - name: Prepare README
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release.yml
```diff
@@ -46,16 +46,9 @@ jobs:
...
-        run: pnpm config set '//registry.npmjs.org/:_authToken' "${NPM_AUTH_TOKEN}"
-        env:
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
-
       - name: Publish
         run: |
-          pnpm publish --tag next --no-git-checks
-        env:
-          NPM_CONFIG_PROVENANCE: true
+          pnpm publish --access public --tag next --no-git-checks --provenance
```
xDivisionByZerox, line 49: You added the config for `access`. The value  you added is the same as the default value. Usually, I'm the one asking for explicit configurations - even if it would be equivalent to the default values. May I ask why you decided to provide it explicitly this time?

xDivisionByZerox (commented):
When I was doing the latest release (v10.3.0), the pipeline failed for the step of setting an additional dist-tag. (see: [Run 21767280993](https://github.com/faker-js/faker/actions/runs/21767280993/job/62806193328)).
According to the [npm documentation](https://docs.npmjs.com/trusted-publishers#limitations-and-future-improvements) this operation is not available via this authentication method. So we eighter go back to setting these tags manually or create a very fine grained access token (only for tag releases - no publishes) for our bot account.

.github/workflows/publish-release.yml
Shinigami92: simply to align with all my other workflows in my other projects, and to omit bypassing through indirect hacks (imagine someone changes somehow .npmrc or something like that)

ST-DDT: I personally would omit it, but IMO adding it doesn't matter that much.
If you have a hacked repository/files, then you have other problems as well (and problably shouldn't release in the first place).

ST-DDT (dismissed):
I pretty much prefer OIDC based auth over token auth.

Maybe use the two step approach until npm either adds support for publish with multiple tags or adds auth for dist-tag.

As an alternative, we drop the next tag entirely. It is kind of hard to search for it, but I couldn't find it any usages of it. This might be my prefered solution.

Shinigami92:
> I pretty much prefer OIDC based auth over token auth.
> 
> Maybe use the two step approach until npm either adds support for publish with multiple tags or adds auth for dist-tag.
> 
> As an alternative, we drop the next tag entirely. It is kind of hard to search for it, but I couldn't find it any usages of it. This might be my prefered solution.

I will discuss this later with @xDivisionByZerox (maybe just next team meeting, we are not in rush with this PR), but yeah, maybe we either split the OIDC publish workflows and the dist-tag pipeline from each other.
We have than a safe publish pipeline and a token based dist-tag pipeline where we need to set the token every 90days

ST-DDT:
One way to handle the token expiry cases, is spliting the dist-tagging into a separate workflow.
It still gets triggered by the release workflow, but if it fails, the token can be replaced and then the workflow part rerun without running into the re-release issue.

.github/workflows/publish-release.yml
xDivisionByZerox: I honestly don't mind having it. AS I pointed out in my comment - I'm all for explicit configs. I was just wondering, since I'm usually in the minority with this opinion.

I think Shini was referring to the fact that CLI argument usually overwrite all implicit settings. These settings might come from a centralized config file (".npmrc") and are able to be changed dynamically. Explicitly setting the config here, omits this attack vector since an attack would need to change this file in particular. A "random" contributor changing the CI files would certainly gather more attention that them changing the magic happening in the doc generation n scripts (which would be able to change the npm config during the doc generation step in the pipeline).
That being said:

- This attack vector exist regardless, so other parts of the CI might be affected if an attacker would choose to go this route
- An attacker has little to no interest in changing the "access" of the released library. What would they even change it to? Make our releases private? xD

.github/workflows/publish-release.yml
xDivisionByZerox: This step should probably removed for now. I have no problem setting the tags manually. We need to do so for the post release process (set `stable` tag one week after release) anyways.

xDivisionByZerox: I removed the step since it makes our release always fail, currently. This means that currently the last step (merging branches) needs to be done manually as well.

xDivisionByZerox (approved):

ST-DDT (approved):

PR #3748, opened 2026-02-28
feat: complete standalone modular functions transformation

### LINKED ISSUE #1118: Export individual API modules on their own

### Suggested solution

```typescript

```
vs
```typescript

const noun = word.noun();
```

### Alternative

```typescript
```

```typescript
```

### Additional context

_No response_

### LINKED ISSUE #2667: Proposal: Design for standalone module functions

### LINKED ISSUE #3867: Design Consideration: Standalone Module Function this parameter
This is more of a design consideration for discussion than an actual proposal. (ADR?)

Instead of explicitly passing fakerCore to each SMF, we could use a `this: FakerCore` parameter.

````ts
// Explicit Parameter
function firstName(fakerCore: FakerCore) {
  return arrayElement(fakerCore, fakerCore.locale.person.first_name);
}

// Direct Use A
fristName(enCore)

// Direct Use B
const enFirstName = bindTo(firstName, enCore);

enFirstName();

// Indirect Use A
function mockUser() {
  return { firstName: firstName(enCore) };
}

mockUser();

// Indirect Use B
function mockUser(fakerCore = enCore) {
  return { firstName: firstName(fakerCore) };
}

mockUser();
````

````ts
// This Parameter
function firstName(this: FakerCore) {
  return arrayElement(this.locale.person.first_name);
}

// Direct Use
const enFirstName = firstName.bindTo(enCore);

enFirstName();

// Indirect Use
function mockUser() {
  return { firstName: firstName() };
}

const enMockUser = mockUser.bindTo(enCore);

enMockUser();
````

Pros:

- One less parameter to pass
- No need to import localized core during use
- Unverified: No javadoc information loss on bind in comparison to custom bind implementations

Cons:

- Usage of a SMFs requires calling bindTo
- Most uses get encapsulated in a function call anyway, so there difference in usage is hardly noticible anyway
- Miniscule RAM overhead.
- Harder to deal with multiple cores/languages in one fn.
- Harder to detect whether `this` has been bound correctly.
- Possible `this` collision if weaving faker and external methods

I think we talked about it before, but I couldnt find any issues/docs/comments that contain any this vs parameter discussion.

Did I miss anything?
Which do you prefer?

### PR DESCRIPTION
**Pull request by ST-DDT**
Final part of #2667, Continuation of #4073

- #2667
- #4073

---

Removes the temp transformation script - completing the migration process.
Reused the existing PR to "preserve" all comments that were made here.

<details>
<summary>Original description</summary>

The goal of this PR is to convert our class/module based system to one, that uses standalone module functions and use them as the implementation of modules.

Non-goals of this PR: 

- create or expose nano bound module functions e.g. enFirstName (except for test purposes).

This PR will be created/achieve in the following steps:

- [x] Create FakerCore
- [x] Create transformation script (deleted before merge)
- [x] Apply transformation script (for early preview of code)
- [x] Fixup references to constants/types/enums, that would be to hard to implement via script
- [x] Run tree-shake tests `faker.person.firstName()` vs `firstName(englishCore)` vs `firstName(nanoCore)`
- [x] Create `generate-module-tree` script, that rebuilds the modules using the standalone functions
- [ ] Update test suite, to use the standalone module function instead?
- [ ] ~Update docs generation script to handle the new structure~
- [ ] Add faker vs standalone toggle to docs page for usage examples?

How to review:

- Expect force pushes during the early stages of this PR (to keep the generated parts separate from manually written code/changes)
- Review what the transform script does, not how it is implemented, as it gets deleted later on.
- Revert the transform commit, apply the script and compare for deltas.
- Check the transformed methods for issues (e.g. name conflict handling const vs imports)
- Check the manual fixup commit for where stuff gets/should be moved
- Run tests on your machine

### Partial alternatives

- #2838 - Slightly different but mostly compatible implementation of FakerCore

</details>

### CODE REVIEW

**Comment by ST-DDT** [comment]
How should we distinguish between module functions and helper code for modules?

e.g. https://github.com/faker-js/faker/blob/next/src/modules/finance/bitcoin.ts

I considered prefixing the helper files with `_`. 

So that:

- `src/modules/*/index.ts` -> module index
- `src/modules/*/[a-z-]+.ts` -> api
- `src/modules/*/_[a-z-]+.ts` -> helpers

As an alternative, we could add an `@api` tag to the jsdocs.

What do you think?

**Comment by ST-DDT** [comment]
Which variant do you prefer?

A)
````ts
resolveLocaleData(fakerCore, 'commerce', 'product_name')
````

or B)

````ts
assertLocaleData(fakerCore.definitions.commerce?.product_name, 'commerce.product_name')
````

A is shorter, but you can click on B to navigate to the source definitions. Also auto completion works better with B.

**Comment by xDivisionByZerox** [comment]
> Which variant do you prefer?
> 
> A)
> 
> ```ts
> resolveLocaleData(fakerCore, 'commerce', 'product_name')
> ```
> 
> or B)
> 
> ```ts
> assertLocaleData(fakerCore.definitions.commerce?.product_name, 'commerce.product_name')
> ```
> 
> A is shorter, but you can click on B to navigate to the source definitions. Also auto completion works better with B.

I need some context here:

- What would be the signature of both functions?
- Why is the string required in version B?
- Are both functions able to resolve entries that are nested more deeply than the standard module/entry way? (think person.firstname.female)

---

Also I have the strong feeling that the standalone module function PRs would get more traction from the core team if they could be done in smaller chunks. Like maybe start with one module and let everyone experience of it works. Since modules have to stay anyway, there would be no harm in simply adding one module at the time. We could even release them as "developer preview" to gather intel on how they "feel" in every day coding.

Also, I'm aware that the "transform-once" script exists, but:

- its currently not reliable (see failing tests)
- its hard to understand what is actually happening
- I'll check the generated files anyway to check if they align with the expected code style

I'm not saying it doesn't work! Throwing it at the entire code base at once might be a bit overwhelming I'd say.
WE did the same thing for the locale data normalization for example.

**Comment by ST-DDT** [comment]
> I need some context here:

https://github.com/faker-js/faker/blob/626865fdbe3f7ca1b6ce7f6ac7fe38c7de316707/src/internal/locale-proxy.ts#L57-L63
https://github.com/faker-js/faker/blob/43f5d71cc6280c51481178fe3fe4a5aac0924a42/src/utils/resolve-locale-data.ts#L18-L37

It needs to the string to include it in the actual error message.

> Are both functions able to resolve entries that are nested more deeply than the standard module/entry way? (think person.firstname.female)

Theoretically yes, but that's not what they are for.
The goal here is to extract the entries from the locale definitions and then pass them to the next process step usually arrayElement. Currently, all our locale data are structured that way.
Please note that definitions.person.first_name is the part that may be absent, everything below that we expect to follow the given type spec. We do not check whether only female is set and male is absent, we have a PersonEntryDefinition that we expect every locale to follow on that level.
If we users want to model their custom code differently, then they can do that either way.

The longer I think about this, the more I prefer B.

---

> Also, I'm aware that the "transform-once" script exists, but its currently not reliable (see failing tests)

It was never meant to do all the work.

> Also I have the strong feeling that the standalone module function PRs would get more traction from the core team if they could be done in smaller chunks. Like maybe start with one module and let everyone experience of it works.

> https://github.com/faker-js/faker/issues/2667#issuecomment-1945478127
> I would really love to see a working implementation that results in <<< 600 KiB like you proposed here, before actually rewriting the whole src code base.

This is kind of chicken and egg problem.
In order for showing that it works, I would like to show a real example e.g. the person module.
To implement that, I need the helpers module, that requires the number module, which requires some utils and the actual core.

Sounds simple, right?

The helper module contains fake. In the current implementation I mostly ignored how fake "knows" which methods exists.
We might have to add a module registry or something similar in the future.

Then there is this tiny detail called jsdocs verification tests, that I need to adjust as well.
I was able to dodge most of the api docs generation stuff for now.

Then there are naming conflicts with module methods e.g. lorem.word and word.word.

Should the module classes already start using the standalone methods?
This will only work if standalone fake works.

TLDR: Consider this a POC. I transformed the entire code base, to see which problems we need to handle.
This produces a working code base, that you can build youself locally and run any tests you would like to run with it.

For an in depth review, I can split this PR later into one per module, but we likely cannot do any releases once the first one lands until all of them are in (unless you don't make them public api).

FYI: Existing tests are passing now, but we need to consider how we change/duplicate them, to ensure both the standalone and the faker tree based ones have a desent coverage.

**Comment by ST-DDT** [comment]
I did some tests with the current state of this PR to determine whether it results in smaller bundle sizes:

Just add the following exports to the faker's index.ts:

````ts
// Note this won't be the final export names
export { default as enPerson } from './locales/en/person';
export { avatar } from './modules/image/avatar';
export { urlLoremFlickr } from './modules/image/url-lorem-flickr';
export { firstName } from './modules/person/first-name';
export { lastName } from './modules/person/last-name';
````

Then I used the vue/vite-esm playground to test the bundle size.

### FTF

````vue
<script setup lang="ts">
import { Faker, enPerson, FakerError } from "@faker-js/faker";
import { ref } from "vue";

const faker = new Faker({
  locale: { person: enPerson },
});

const fullName = ref(`${faker.person.firstName()} ${faker.person.lastName()}`);
const avatarUrl = ref(faker.image.avatar());
const natureImageUrl = ref(
  faker.image.urlLoremFlickr({ category: "nature" }) + new FakerError("test")
);
</script>
````

> dist/assets/index-mO_qTbs-.js   182.21 kB │ gzip: 61.69 kB

grafik

### SMF

````vue
<script setup lang="ts">
import { firstName, lastName, avatar, urlLoremFlickr, enPerson, FakerError, createFakerCore } from "@faker-js/faker";
import { ref } from "vue";

const fakerCore = createFakerCore({
  definitions: { person: enPerson },
});

const fullName = ref(`${firstName(fakerCore)} ${lastName(fakerCore)}`);
const avatarUrl = ref(avatar(fakerCore));
const natureImageUrl = ref(
  urlLoremFlickr(fakerCore, { category: "nature" }) + new FakerError("test")
);
</script>
````

> dist/assets/index-GwB6agm0.js   136.59 kB │ gzip: 50.37 kB

grafik

### Hard-Coded

````vue
<script setup lang="ts">
import { ref } from "vue";

const fullName = ref(`Firstname LastName`);
const avatarUrl = ref("https://cdn.jsdelivr.net/gh/faker-js/assets-person-portrait/female/512/75.jpg");
const natureImageUrl = ref(
  "https://loremflickr.com/2185/2550/nature?lock=7872953778781783Error:%20test"
);
</script>
````

> dist/assets/index-DGlXtW4P.js   60.04 kB │ gzip: 24.07 kB


### Result

The standalone module functions (SMF) take less space in the final bundle (-46 kB | gzip: -11 kB).
If you ignore vue and the other stuff, then its 30% smaller. 
(And for whatever reason, we still ship extra parts of faker, maybe since definening a module class is a side effect?)

**Comment by ST-DDT** [comment]
Interesting finding with createFakerCore:

````ts
return {
    definitions: Array.isArray(definitions)
      ? mergeLocales(definitions)
      : (definitions ?? {}),
    randomizer: randomizer ?? ({} as Randomizer),
    config: config ?? {},
  }
````

-> Some side effect, causes `simpleFaker` to be in the final bundle.

vs

````ts
return {
    definitions: Array.isArray(definitions)
      ? definitions[0]
      : (definitions ?? {}),
    randomizer: randomizer ?? ({} as Randomizer),
    config: config ?? {},
  }
````

-> Not included 

mergeLocales doesn't even use/reference simpleFaker! (Same with generateMersenne53Randomizer)

🤔 WHY!?

**Comment by xDivisionByZerox** [comment]
> This is kind of chicken and egg problem.
> In order for showing that it works, I would like to show a real example e.g. the person module.
To implement that, I need the helpers module, that requires the number module, which requires some utils and the actual core.
>
> Sounds simple, right?
>
> The helper module contains fake. In the current implementation I mostly ignored how fake "knows" which methods exists.
> We might have to add a module registry or something similar in the future.
>
> Then there is this tiny detail called jsdocs verification tests, that I need to adjust as well.
> I was able to dodge most of the api docs generation stuff for now.
>
> Then there are naming conflicts with module methods e.g. lorem.word and word.word.
>
> Should the module classes already start using the standalone methods?
> This will only work if standalone fake works.
> 
> TLDR: Consider this a POC. I transformed the entire code base, to see which problems we need to handle.
> This produces a working code base, that you can build youself locally and run any tests you would like to run with it.
>
> For an in depth review, I can split this PR later into one per module, but we likely cannot do any releases once the first one lands until all of them are in (unless you don't make them public api).

Thanks for the in depth explanation. I was already aware about the "person => helper => number" situation. That why I honestly had the number module in mind for this experiment. I was not aware of the faker.helper.fake situation.

**Comment by ST-DDT** [comment]
I switched from `resolveLocaleData` to `assertLocaleData`.

**Comment by ST-DDT** [comment]
I just noticed that my current implementation of a moduleRegistry for fake leads to a circular dependency.
The fake method references the global registry, , the global registry loads the methods via the module indexes, the helperEntries are defined in the index, but since the class is already initializing the helpers registry entry gets set to undefined and fake cannot call any helper methods directly. And if you execute the requests in a different order, the other modules are missing. I'm not exatly sure about my explanation, but I receive undefined modules in the registry.
I have to think about how to avoid that, maybe I will create a "registry.ts" file per module...

Depending on your point of view all methods that use fake experience a breaking change.

For now, I call all internal fake methods in the least breaking way, while still allowing for some level of tree shaking. Due to circular dependencies it is probably impossible to implememt in an entirely non-breaking way.

We might be able to relieve some of the pain points by switching from fake patterns to resolver functions internally. But that would be a breaking change and a separate PR for sure.

See also 

- #3176 

I'll push the latest version once I get the fake issue resolved.

**Comment by ST-DDT** [comment]
It's working...

**Comment by Shinigami92** [comment]
> It's working...

fantastic 🚀 

is it possible to clean this PR up and/or spliterate it into smaller PR-chunks? e.g. extract preparing chores which does not affect directly the implementation of this PR, but is required for it

e.g. I see some utilities like casing-utils (which I usually do by https://www.npmjs.com/package/change-case), also some file namings now with a prefixed underscore

**Comment by ST-DDT** [comment]
As for cleaning this PR, not sure how to do that (except for squashing a few commits here and there).

---

I would like to split this PR into smaller pieces, but not sure whether some commits needs to be kept together.

e.g.

- basics/utils
- scripts 
- helpers, int, string
- other modules (each)
- generate-module-tree + fixes

---

Before we start merging a bunch of stuff, there are a few things that we need to consider:

- Do we consider the changes to fake "breaking"?
- How do we export the actual SMFs?
- Do we export the module registries?

**Comment by Shinigami92** [comment]
# PR #3748 Review: Standalone Module Functions

> Reviewed by @Shinigami92 / Claude Max with `/effort high` | 2026-03-22

---

## Summary

This PR introduces a fundamental architectural shift from class-based module methods to standalone functions, enabling tree-shaking and smaller bundle sizes. The approach is sound - the ~30% reduction in bundle size for targeted imports is compelling. I've done a deep review of the core infrastructure, the transformation pattern, and the resulting code. Below are my findings, concerns, and suggestions.

---

## Architecture & Design

The overall architecture is clean and well-thought-out:

```
FakerCore (data container) -> Standalone Functions -> Module Classes (thin wrappers)
```

The `FakerCore` interface is minimal and elegant - just `definitions`, `randomizer`, and `config`. The separation of the data container from the class hierarchy is the right move for tree-shaking.

I especially like how the per-function mini-registries for `fake()` calls solve the circular dependency problem (e.g., `streetAddress` only importing `{ street, buildingNumber, secondaryAddress }` rather than the full `locationModule`). This is the kind of design that makes tree-shaking actually work.

---

## Critical Issues

### 1. `fake()` entrypoints divergence - Behavioral difference between standalone and class API

This is the most significant concern in the PR.

**Class-based `fake()`** (`src/modules/helpers/index.ts:734-737`):
```ts
entrypoints: ReadonlyArray<unknown> = [
  moduleRegistry,
  this.faker.rawDefinitions,
]
```

**Standalone `fake()`** (`src/modules/helpers/fake.ts:181`):
```ts
entrypoints: ReadonlyArray<unknown> = [fakerCore.definitions]
```

The class version resolves `{{person.firstName}}` through `moduleRegistry.person.firstName` (a function), while the standalone version would try to resolve it as `definitions.person.firstName` (which is `first_name` in locale data, not `firstName` - so it would fail or behave differently).

This means:
- `faker.helpers.fake('{{person.firstName}}')` works (resolves via moduleRegistry)
- `fake(fakerCore, '{{person.firstName}}')` would **not** work with the default entrypoints

Users migrating from class API to standalone would hit this silently. The JSDoc for standalone `fake()` says "Defaults to `[ fakerCore.definitions ]`" but the class version docs say "Defaults to the moduleRegistry and locale data." - this should be made more prominent as a key difference, or ideally the standalone version should also default to include the moduleRegistry (accepting the tree-shaking tradeoff for the `fake()` function specifically).

**Suggestion:** Consider making the standalone `fake()` default entrypoints match the class version, or at least add a very prominent warning in the JSDoc. Since `fake()` is inherently dynamic and can't be tree-shaken anyway, including `moduleRegistry` in its defaults is the pragmatic choice.

### 2. `fakerToCore` uses `rawDefinitions` - bypasses locale proxy validation

In `src/internal/faker-to-core.ts:14`:
```ts
definitions: faker instanceof Faker ? faker.rawDefinitions : {},
```

The Faker class wraps definitions in a `LocaleProxy` (`createLocaleProxy(this.rawDefinitions)`) that provides helpful error messages with documentation links when locale data is missing. By using `rawDefinitions`, standalone functions lose this DX benefit.

The standalone functions do use `assertLocaleData()` which throws on missing data, but the proxy provides additional niceties (prevents accidental writes, gives contextual error messages for missing categories).

**Suggestion:** Consider whether `fakerToCore` should use `faker.definitions` (the proxy-wrapped version) instead. This would preserve the error reporting quality users are used to. If not, document this as a known DX trade-off.

### 3. Missing public API exports

`src/index.ts` exports `moduleRegistry` but does NOT export:
- `createFakerCore`
- `FakerCore` (type)
- `FakerConfig` (type)
- `fakerToCore`
- `setDefaultRefDate` / `getDefaultRefDate`

Without these, users **cannot actually use standalone functions** in a meaningful way. I understand this is a POC, but for reviewability it would be helpful to see the intended public API surface. Even if unexported for now, a comment or TODO noting the planned exports would help reviewers reason about the final shape.

**Suggestion:** At minimum, add TODO comments in `src/index.ts` indicating which new symbols are planned for export.

### 4. `@ts-expect-error` for private field access in `fakerToCore`

`src/internal/faker-to-core.ts:15-19`:
```ts
randomizer: faker._randomizer,
config: {
  defaultRefDate: faker._defaultRefDate,
},
```

This creates a fragile coupling - any rename of these private fields would silently break at runtime (TypeScript would flag the `@ts-expect-error` as unnecessary, but only if someone notices).

**Suggestion:** Consider adding a `toCore()` method to `SimpleFaker`/`Faker` instead, which can access its own private fields safely. Or make `_randomizer` and `_defaultRefDate` protected/internal rather than private.

---

## Moderate Concerns

### 5. Tree-shaking blocker in `createFakerCore`

As ST-DDT discovered in the PR comments, importing `mergeLocales` in `createFakerCore` pulls `simpleFaker` into the bundle via side effects. This partially undermines the tree-shaking benefit.

`src/faker-core.ts:4-5`:
```ts
import { mergeLocales } from './utils/merge-locales';
import { generateMersenne53Randomizer } from './utils/mersenne';
```

**Suggestion:** Consider lazy imports or making `mergeLocales` optional:
```ts
export function createFakerCore(options) {
  const { definitions, randomizer, config } = options;
  const resolved = Array.isArray(definitions)
    ? (await import('./utils/merge-locales')).mergeLocales(definitions)
    : (definitions ?? {});
  // ...
}
```
Or split into `createFakerCore` (simple, no mergeLocales) and `createFakerCoreFromLocales` (with merging).

### 6. Transform script fragility

The `generate-module-tree.ts` script uses string replacements for JSDoc transformation (`src/scripts/generate-module-tree.ts:148-169`):

```ts
.replace(' * @param fakerCore The FakerCore to use.\n', '')
.replaceAll(/ +\*\n +\*\n/g, ' *\n')
.replaceAll(new RegExp(`${methodName}\\(fakerCore(?:, ?)?`, 'g'), ...)
```

This is fragile - any deviation in JSDoc formatting (extra whitespace, different param descriptions) could cause silent failures. The `restoreFakerTreeInvocations` regex logic (`lines 108-117`) is also complex and could produce incorrect results with edge cases.

**Suggestion:** Consider using ts-morph's AST manipulation for JSDoc transformations instead of regex string replacements where possible. Also: should the transform scripts (`tranform-once.ts`) be deleted before merge as mentioned in the PR description?

### 7. The `tranform-once.ts` typo and code duplication

The filename has a typo: `tranform-once.ts` (missing 's' in "transform"). Also, it duplicates the case conversion utilities that also exist in `scripts/shared/character-case.ts` (lines 133-147 of the script vs the shared module).

**Suggestion:** Fix the typo if the file is being kept, and deduplicate the case conversion utilities. Per the PR description this script gets deleted before merge, so this may be moot.

### 8. `bio()` relying on `fake()` without explicit entrypoints

`src/modules/person/bio.ts:21`:
```ts
return fake(fakerCore, bio_pattern);
```

The `bio_pattern` locale data contains patterns like `{{person.bio_part}}`, `{{internet.emoji}}`, `{{word.noun}}`. Since `fake()` defaults to `[fakerCore.definitions]`, these resolve as locale data paths - which works today. But this creates an implicit contract: **any locale data pattern used without explicit entrypoints must only reference other locale data paths, never module functions.**

If a locale contributor adds a pattern like `{{person.firstName}}` to `bio_pattern`, it would break standalone `bio()` while still working via `faker.person.bio()` (which uses `moduleRegistry` as entrypoint).

**Suggestion:** Add a comment in the standalone `fake()` JSDoc or a lint rule that flags standalone `fake()` calls without explicit entrypoints, or at least document this constraint for locale contributors.

---

## Minor Issues / Suggestions

### 9. Inconsistent module naming in registries

Registry keys don't prefix with module name (good), but the import aliases do:
```ts
import { bear as animalBear } from './bear';
export const animalModule = {
  bear: animalBear,  // key is 'bear', not 'animalBear'
};
```

This is correct behavior, but the verbose aliasing adds visual noise. Since each registry file is in its own module directory, the aliasing isn't strictly necessary - a simpler approach would be:

```ts
import { bear } from './bear';
export const animalModule = { bear };
```

The aliasing is needed in `index.ts` (to avoid conflicts), but not in `registry.ts`. Though I understand this may be generated code, so simplifying the generator might be the better approach.

### 10. `FakerConfig` is very minimal

```ts
export interface FakerConfig {
  defaultRefDate?: () => Date;
}
```

For the POC this is fine, but worth thinking ahead about what else might land here (timeZone, locale preferences, etc.) to ensure the interface stays clean.

### 11. No type export for the module registry shape

The `moduleRegistry` object is exported but its type is inferred. Users doing advanced things (like wrapping or extending it) would benefit from an explicit type:

```ts
export type ModuleRegistry = typeof moduleRegistry;
```

### 12. JSDoc `@since 10.4.0` on new APIs

`createFakerCore` is tagged `@since 10.4.0`. Make sure this version number is correct when merging.

---

## Questions for Discussion

1. **Should standalone functions be public API in this PR, or should this land as internal-only first?** The PR description mentions this is a POC. If we merge as internal-only, we can iterate on the API surface without breaking changes.

2. **How should we handle the `fake()` entrypoints divergence?** Three options:
   - a) Standalone `fake()` defaults to `[moduleRegistry, fakerCore.definitions]` (matches class, kills tree-shaking for `fake`)
   - b) Keep the divergence but document clearly
   - c) Make `fake()` require explicit entrypoints (no default) to force awareness

3. **Per xDivisionByZerox's suggestion** about splitting into smaller PRs per module - is there a plan to do this before merge? I agree the overall approach is solid, but reviewing 359 files at once is challenging. If we could merge the infrastructure (`FakerCore`, `fakerToCore`, `module-registry`, scripts) first, then add modules incrementally, the review quality would be higher.

4. **The `mergeLocales` tree-shaking issue** - has a workaround been found for the side-effect that pulls `simpleFaker` into the bundle?

---

## What Looks Good

- The `FakerCore` interface is clean and minimal
- The `assertLocaleData()` pattern is consistent and provides good error messages
- Inline mini-registries for `fake()` calls (e.g., `streetAddress`, `company.name`, `productDescription`) are elegant and tree-shake-friendly
- The `_` prefix convention for internal helpers (`_eval.ts`, `_luhn-check.ts`, `_char-mappings.ts`, `_select-definition.ts`) is clear
- Bundle size results are promising (~30% reduction)
- The generate-module-tree script approach ensures consistency across all modules
- All existing tests pass, and coverage actually increased slightly

---

## TL;DR

The architecture is solid and the bundle size wins are real. The main concerns are:

1. **`fake()` behaves differently** between standalone and class API (critical to document or fix)
2. **Missing public API exports** make standalone functions unusable for end users
3. **`fakerToCore` bypasses locale proxy** validation (DX trade-off)
4. **Private field access via `@ts-expect-error`** is fragile

I'd recommend addressing items 1 and 4 before merge, and having a plan for items 2 and 3. The split-by-module approach @xDivisionByZerox suggested would also make the final review more thorough.

ST-DDT:
@Shinigami92 Sorry for the question, but is that your opinion/an analysis backed by you personally or just a copy and paste from Claude's analysis of this PR?

Shinigami92:
> @Shinigami92 Sorry for the question, but is that your opinion/an analysis backed by you personally or just a copy and paste from Claude's analysis of this PR?

second ⚠️ 
It was a try to make a helpful review for a near 400 file-change PR to provide you with hopefully helpful infos or creative directions
please let me know if you don't want reviews like this, and/or if anything about this comment is specifically bad, so I can e.g. improve prompts next time

ST-DDT:
I'm fine with reviews from AI tools, just add a hint that this does or does not reflect your personal opinions.
Some people use AI to help with formulating their ideas/translating their responses/actually sprinkle their own analysises/opinions throughout the text, so explicitly stating your intentions with posting it helps to clarify that.

I'm aware of all/most of the points brought up by the review, thats why I asked for explicit opinions here:
https://github.com/faker-js/faker/pull/3748#issuecomment-4070405620

Shinigami92:
> I'm fine with reviews from AI tools, just add a hint that this does or does not reflect your personal opinions. Some people use AI to help with formulating their ideas/translating their responses/actually sprinkle their own analysises/opinions throughout the text, so explicitly stating your intentions with posting it helps to clarify that.
> 
> I'm aware of all/most of the points brought up by the review, thats why I asked for explicit opinions here: [#3748 (comment)](https://github.com/faker-js/faker/pull/3748#issuecomment-4070405620)

ah now I understand where your comment was coming from 👍 
yeah you are right, I will keep care about this next time

very good that you are aware of almost everything claude discovered
for the comment, I think we need a human-meeting together with @xDivisionByZerox, because everything else would be to complex and comment-write-intensive

ST-DDT:
FYI: I updated the FakerCore PR:

- https://github.com/faker-js/faker/pull/2838

It uses a slightly different implementation then the one used in this PR but both are generally compatible.
I can modify either PR depending on which variant your prefer.

This/SMF PR:

- `Faker` -> `fakerToCore()` -> `FakerCore`
  https://github.com/faker-js/faker/blob/8fa7e270fdc130fa34e644dc282364e52f038baf/src/internal/faker-to-core.ts#L12-L22

FakerCore PR:

- `faker.fakerCore` permanent field in class
  https://github.com/faker-js/faker/blob/6dada60781938d5e76f87062abf0be9d1f9cf7b1/src/simple-faker.ts#L123-L125

ST-DDT:
I updated this to the latest "agreed upon" state.
All methods that use `fake`, currently rely on full `Faker` instances (no registry).
This will only be resolved in v11.

I'll start splitting this into smaller PRs soon.

ST-DDT:
Note: I "rebased" this PR, but I still need to verify all (non-automated) changes.

src/modules/airline/index.ts
```diff
@@ -1 +1,12 @@
+/**
+ * IATA stands for [International Air Transport Association](https://iata.org).
+ * It's the trade association for the world's airlines and it is
+ * responsible for setting standards relating to many aspects of airline
+ * operations.
+ */
```
Shinigami92, line 6: **polish:** I think this is in the wrong place

ST-DDT: I think it is in the correct place as it affects all of airline.
If you disagree, where do you think this should be and what are your criteria for choosing it?

Shinigami92: I have two thoughts about it, one of them applied already previously outside this PR:
1. The JSDoc is and was at a place where it is not really near a place about "IATA", but "just" at the top of the file.
   - this might even be overlooked by an AI, because they tend to read rows +- X-lines of offset. if then such a documentation is not near, it might be overlooked.
2. now it is even in an index file, which have more or less the only purpose to guard module stuff up to their folder-branch.
   - the fear of overlook just even increased, because it is not even in a place where this info is valuable.

ST-DDT: I think, we follow some fundamentally different trains of thoughts here.

IMO AIs know what IATA airline/airport codes are (or dont need to know). Same for most humans that need airline code as well.
So this is mostly there for those that want to jump into their imports (and I assume it will be via the SMF index file) before delving further down. 
Those that jump into a module/method will likely get thrown down to where the implementstion starts never seeing the head code in the first place.
It might be that we dont need the info at root level anyway and/or just need to replace all IATA references with wiki links directly.

Shinigami92: > […] It might be that we dont need the info at root level anyway and ~~/or~~ just need to replace all IATA references with wiki links directly.

understood and agree

ST-DDT: Agree to 

- A) keeping it at the index 
- B) or spreading it to all IATA references

?

Shinigami92: I updated my comment above to make my sentence more obvious 🙂  
sorry for confusion

ST-DDT: Then I'll do that as part of a spearate PR.

ST-DDT:
I verified that I didn't miss any changes due to the rebase.

<details>
<summary>smfdiff for rebase</summary>

````bash
smfdiff() {
  local show_green=0

  if [ "$1" = "--green" ] || [ "$1" = "-g" ]; then
    show_green=1
    shift
  fi

  local dir="$1"

  if [ -z "$dir" ]; then
    echo "Usage: smfdiff [--green|-g] path/to/dir" >&2
    return 2
  fi

  local file="${dir%/}/module.ts"

  if [ ! -f "$file" ]; then
    echo "Error: file not found: $file" >&2
    return 1
  fi

  local cutoff
  cutoff="$(date -d '2026-06-25 00:00:00' +%s)"

  git blame --line-porcelain -- "$file" |
  awk -v cutoff="$cutoff" -v show_green="$show_green" '
    BEGIN {
      new_count = 0
      old_count = 0
    }

    /^author-time / {
      is_new = ($2 >= cutoff)
    }

    /^\t/ {
      text = substr($0, 2)

      if (text ~ /^export class / || text ~ /^ \* ### Overview/) {
        exit
      }

      if (text ~ /^import /) {
        next
      }

      if (is_new) {
        new_count++
        if (show_green) {
          printf "🟢 %s\n", text
        } else {
          printf "%s\n", text
        }
      } else {
        old_count++
        printf "🔴 %s\n", text
      }
    }

    END {
      if (new_count > 0) {
        printf "\n🟢🟢🟢 %d line(s)\n", new_count
      } else {
        printf "\n🔴🔴🔴 0 lines(s)\n"
      }
    }
  '
}
````

(Though I have to use /module.ts going forward)

</details>

I will now update the core split PR.

src/modules/airline/index.ts
ST-DDT: Fixed in code, will be resolved with the next rebase.

ST-DDT:
I updated this one last time. I'll split this into smaller pieces / update the core part soon.

I noticed, one major downside with the switch to oxc tooling:
It cannot (correctly/fully) remove unused imports. Which makes coded transformation tasks such as this way harder.
I even considered temporarily adding back the required dependencies to run the old cleanup tooling for this task.

The final generate-module-tree script is still somewhat repeatable, but you may have to manually cleanup the imports afterwards.

I tried the autofix, but it didn't work fully for me.
Roughly 50% off all files fixed in "refactor: transform x module" + "final cleanup" commits are added by this issue.

I spend an awful lot of effort in order to keep the manual changes to the code to a minimum, so please don't cause big merge conflicts.

I will update the core modules PR tomorrow:

 - #3857 

Feel free to try to bring the oxc tooling to drop unneeded imports automatically, but make sure it does it properly as doing it by halves doesn't do my script any favor. You can easily try this by reverting to `chore: run transform script (dirty)` and triggering the cleanup.

Shinigami92:
That is interesting, because at work we do not have such issues with oxc. I need to check that I did not forgot something 🤔

However also the review I can start earliest this afternoon. You might have already noticed that I often do not work on faker in working hours as I have a company and a private Mac not sharing the same ssh keys. So my private Mac is at home, and I'm in coworking space 🤷

ST-DDT:
No hurry. When creating this PR, my internal estimates pointed to the end of 2026 or later for the last PR of this series to be merged.
And hopefully v12 having a stable public SMF API.

ST-DDT:
I updated the script with the following improvements:

- imports are added only when needed, reducing the manual cleanup commits
- SMFs now have `@since 11.0.0`, the original methods retain their since from their module-function signature (index based).
  - The since may get lost or shuffled, if signatures in the SMF are changed.
  - `@since 11.0.0 (8.4.0)` is not supported
- SMFs are now flagged as experimental
- `generate:module-tree` is now safe (no diff) to run repeatedly
- add auto generated comment to modules
- moved `legacyReplaceSymbolWithNumber` to own file
- `sample` -> `moduleSample`

Checkpoints (preflight green):

- `refactor: _prefix module helpers`
- `feat: introduce module to standalone transform script` (not apply)
- `refactor: transform word module`
- `infra: introduce generate-module-tree script` (not apply)`
- `chore: remove temp exports` )last commit)

The scripts can be run via:

- `pnpm tsx scripts/temp-transform-once.ts `
- `pnpm run generate:module-tree [module]`

If you don't wish to review all, you can simplify this using this command:

````sh
git diff 5753990 -- src/modules/
````

(Diff to previous version)

Shinigami92 (approved):
👍

PR #3773, opened 2026-03-29
infra: migrate from eslint to oxlint

Shinigami92:
## Summary

This PR replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter) - a much faster linter written in Rust by the oxc project.

**What changed:**

- Removed 9 ESLint-related packages, added `oxlint` 1.78.0 + `oxlint-tsgolint` 7.0.2001 (for type-aware linting)
- Kept `eslint-plugin-jsdoc` and run it through oxlint's [JS plugin bridge](https://oxc.rs/docs/guide/usage/linter/js-plugins.html) - oxlint's native jsdoc plugin is missing too many of the rules we rely on (see below)
- Migrated all rule configurations, file overrides, and ignore patterns to `oxlint.config.ts`
- Converted `eslint-disable` comments to `oxlint-disable` where the rule exists, removed them where it doesn't
- Updated VS Code extension recommendation, devcontainer, renovate grouping, documentation, and `.gitignore`
- Prettier stays as-is (runs standalone via `pnpm prettier`), a follow-up PR will look into oxfmt
- Type-aware linting is enabled via `oxlint-tsgolint`, so rules like `consistent-type-exports`, `restrict-plus-operands`, `switch-exhaustiveness-check` etc. all work with full type information

**Performance:** The lint step in CI went from **52s** ([`next` branch](https://github.com/faker-js/faker/actions/runs/23672232244/job/68967928424)) to **~4s** - a **~12x speedup** - with type-aware linting enabled. Without the jsdoc JS plugin oxlint finishes in ~1s; running `eslint-plugin-jsdoc` through Node is what costs the other ~3s, and it is worth it (see below).

## Rule sets: explicit instead of presets

The ESLint config extended full presets (`eslint:recommended`, `tseslint.configs.strictTypeChecked`, `unicorn/recommended`, `jsdoc/flat/recommended-typescript-error`). oxlint has no equivalent presets - it groups rules into categories (`correctness`, `suspicious`, `pedantic`, `style`, `restriction`, ...), and those categories do **not** line up with the ESLint presets. Enabling only `correctness` + `suspicious` would silently stop enforcing a large part of what we had before, e.g. `typescript/no-misused-promises`, `typescript/only-throw-error`, `jsdoc/require-param-description`, `unicorn/filename-case` and ~90 further unicorn rules.

Therefore every rule of the previous presets that oxlint implements is now listed explicitly in `oxlint.config.ts`. That makes the config longer, but it is an exact, reviewable statement of what we enforce - and it is verified: the whole repository lints clean with it.

## jsdoc: `eslint-plugin-jsdoc` via oxlint's JS plugin bridge

oxlint's native jsdoc plugin implements only about half of what `jsdoc/flat/recommended-typescript-error` gave us. Missing are, among others, `require-jsdoc` (our guard for public API docs), `sort-tags`, `check-param-names`, `valid-types`, `check-alignment`, `multiline-blocks`, `no-multi-asterisks`, `require-returns-check`, `require-yields-check` and `check-values` (tracked in [oxc#1170](https://github.com/oxc-project/oxc/issues/1170)).

Instead of losing those, we keep `eslint-plugin-jsdoc` (same version as before, 64.1.0) and load it through oxlint's [JS plugin](https://oxc.rs/docs/guide/usage/linter/js-plugins.html) support:

```ts
plugins: ['typescript', 'unicorn', 'oxc', 'vitest'], // no native 'jsdoc'
jsPlugins: [{ name: 'jsdoc-js', specifier: 'eslint-plugin-jsdoc' }],
settings: { jsdoc: { mode: 'typescript' } },
```

- The name `jsdoc` is reserved by oxlint's native plugin, so the JS plugin is aliased to `jsdoc-js` - rules and inline disables are written as `jsdoc-js/<rule>`
- Rule options, `settings` and auto-fixes all work: `pnpm run lint --fix` sorts JSDoc tags again, exactly like before
- Trade-offs: JS plugins are still in alpha (not subject to semver), linting takes ~4s instead of ~1s, and `eslint-plugin-jsdoc` pulls `eslint` itself back in as an auto-installed peer dependency (it imports `ESLint` in its entry module) - it is never executed, oxlint only loads the rule implementations

## Rules that are not enforced anymore

Not every ESLint rule has an oxlint equivalent yet. Below is what we're temporarily losing.

### typescript rules not implemented (need type-aware linting via tsgolint)

| Rule | What it does | Tracking |
|---|---|---|
| `typescript/naming-convention` | Enforces our PascalCase for types/classes, `T`-prefix for type params convention | Closed in oxc ([oxc#6361](https://github.com/oxc-project/oxc/issues/6361)) because it needs type info, open in [tsgolint#186](https://github.com/oxc-project/tsgolint/issues/186) |

### unicorn rules not implemented by oxlint

These `eslint-disable` directives were removed because oxlint does not know the rules (and would report the directives as unused):

| Rule | What it guards |
|---|---|
| `unicorn/prefer-https` | `http:` URLs in results (`src/modules/internet/module.ts`) |
| `unicorn/no-this-outside-of-class` | `this` in the vitest matcher extensions (`test/vitest-extensions.ts`) |
| `unicorn/no-declarations-before-early-exit` | declaration ordering (`src/modules/commerce/module.ts`) |
| `unicorn/no-top-level-assignment-in-function` | lazy module-renderer init (`scripts/shared/markdown.ts`) |
| `unicorn/prefer-iterator-to-array` | `matchAll` spread, needs Node 22+ (`test/scripts/apidocs/verify-jsdoc-tags.spec.ts`) |
| `unicorn/no-nonstandard-builtin-properties` | `Error.prototype.stack` access (`test/support/seeded-runs.ts`) |
| `unicorn/no-top-level-side-effects` | `console.log` in vitest configs (`vitest.config.ts`, `vitest.it-config.ts`) |

`unicorn/no-unreadable-array-destructuring` **is** implemented and enabled, but oxlint only flags consecutive ignored values, not the swap-via-destructuring pattern that `eslint-plugin-unicorn` v73 flags - so the disable comment in `src/modules/helpers/module.ts` was dropped as well.

### Intentionally dropped (not needed with oxlint)

| What | Why dropped |
|---|---|
| `@stylistic/padding-line-between-statements` | Stylistic formatting concern - oxlint has no `@stylistic` plugin. Closed as "not planned" ([oxc#11435](https://github.com/oxc-project/oxc/issues/11435)). Will be covered by oxfmt in the future. |
| `eslint-plugin-file-progress` | CI convenience only (progress bar during linting). Not needed - the whole lint run takes a few seconds. |
| `eslint-plugin-prettier` | Prettier is now run standalone via `pnpm prettier --check .` in CI. This was just the bridge between ESLint and Prettier. Will be replaced by oxfmt in a future PR. |

### Rules disabled to preserve parity with the ESLint config

oxlint puts a few rules into `correctness`/`suspicious` that were **not** part of our ESLint config, or implements them more strictly than the ESLint counterpart. They are turned off so this PR does not introduce unrelated violations - they can be evaluated separately:

| Rule | Why disabled |
|---|---|
| `no-shadow`, `no-new`, `no-underscore-dangle` | Not in our ESLint config |
| `no-useless-constructor` | oxlint doesn't skip constructors that only declare parameter properties (`typescript-eslint` does), which would flag 4 valid module constructors |
| `typescript/consistent-return` | Not in our ESLint config |
| `typescript/no-unsafe-type-assertion`, `typescript/require-array-sort-compare` | Not in our ESLint config |
| `vitest/require-to-throw-message` | Not part of the vitest recommended set we used |
| `unicorn/consistent-function-scoping` | In `eslint-plugin-unicorn`'s recommended set, but oxlint's implementation also flags non-capturing local/arrow functions, which would introduce new violations |

## What we gain

- **`.vue` files are linted now.** ESLint never looked at them (no vue parser configured). oxlint does, which surfaced a handful of small issues that are fixed in this PR: a useless escape in `method.vue`, `String#replace(/g)` → `replaceAll`, implicit length checks and `(i) => i` filters in `ApiIndex.vue`, and `Array#reverse()` → `toReversed()` in `refreshable-code.vue`. One `no-non-null-assertion` in `AsideSponsors.vue` is suppressed inline.
- **`cypress/**` is linted now** - it only needed `vitest/valid-expect` turned off, because cypress ships its own chai-based `expect`.
- `unicorn/prefer-spread` still flags `String#split('')` (`eslint-plugin-unicorn` v73 no longer does), so `vin.split('')` became `[...vin]` in `src/modules/vehicle/module.ts` - equivalent for VIN input, covered by the existing vehicle tests.

## Other notes

- `no-undef` with an empty `env` is kept for `src/**/*.ts` (the ESLint config did this via `languageOptions.globals: {}`), so library sources still can't reach for environment globals unnoticed. The two intentional usages (`base64.ts`, `deprecated.ts`) keep their disable comments.
- `typescript/no-unsafe-return` is enabled, with one inline disable in `docs/.vitepress/versions.ts` because tsgolint fails to resolve the type of `Array#toSorted`.
- `test/require.spec.cts` is ignored by oxlint due to a parser limitation with top-level `await` in `.cts` (CommonJS) files - it's the only remaining migration-related ignore.
- JS plugins don't run on `.vue` files (oxlint doesn't hand custom-parser files to them), so the jsdoc rules skip our vue components - same as before, ESLint didn't lint them either.
- oxlint automatically respects `.gitignore`, so no need to duplicate those patterns in `ignorePatterns`.

## Test plan

- [x] `pnpm run lint` passes with 0 errors (type-aware enabled). One intentional `vitest/warn-todo` warning remains on the Windows-flaky `it.todo` in `finance.spec.ts`; CI runs `oxlint` without `--deny-warnings`, so it is non-fatal
- [x] `pnpm run ts-check` passes
- [x] `pnpm prettier --check .` passes
- [x] `pnpm run test` passes (55 test files)
- [x] CI pipeline passes - lint job in ~1s vs 52s on `next`


docs/api/ApiIndex.vue
```diff
@@ -31,17 +31,17 @@ const filtered = computed(() => {
           const matchedHeaders = item.headers.filter(
             ({ text, anchor }) => matches(text) || matches(anchor)
           );
-          return matchedHeaders.length
+          return matchedHeaders.length > 0
             ? { text: item.text, link: item.link, headers: matchedHeaders }
             : null;
         })
-        .filter((i) => i);
+        .filter((i) => i !== null);
```
ST-DDT: We usually use the nullish check instead.

```suggestion
        .filter((i) => i != null);
```

docs/api/ApiIndex.vue
```diff
@@ -31,17 +31,17 @@ const filtered = computed(() => {
...
 
-      return matchedGroups.length
+      return matchedGroups.length > 0
         ? { text: section.text, items: matchedGroups }
         : null;
     })
-    .filter((i) => i) as APIGroup[];
+    .filter((i) => i !== null) as APIGroup[];
```
ST-DDT: Same here

test/modules/date.spec.ts
```diff
@@ -1033,7 +1033,7 @@ describe('date', () => {
     });
 
     it('should not use the refDateSource when refDate is provided (with function)', () => {
-      const spy: () => Date = vi.fn();
+      const spy: () => Date = vi.fn<() => Date>();
```
ST-DDT: If we pass the type to the generics then we likely don't need it on the variable itself.

scripts/generate-locales.ts
```diff
@@ -510,15 +510,14 @@ for (const locale of locales) {
...
+    // oxlint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
     generateLocaleFile(locale),
 
     // /docs/locales/*.md
-    // eslint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
+    // oxlint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
     generateLocaleDocumentation(locale),
 
     // src/locales/**/index.ts
-    // eslint-disable-next-line unicorn/prefer-top-level-await -- Disabled for performance
```
ST-DDT, line 521: Why does this one not require the suppression?
(This is more about reliability + predictability, than this line in particular)

docs/api/ApiIndex.vue
Shinigami92: good catch 👍

test/modules/date.spec.ts
Shinigami92: good idea

scripts/generate-locales.ts
Shinigami92: I also saw that but wasn't enough interested to investigate, will try to find it out now that you requested it 👌

Shinigami92: 🤖 No suppression needed here, because oxlint's `unicorn/prefer-top-level-await` bails out for functions that are referenced more than once and this one is recursive.

ST-DDT (dismissed):
Well, then it is likely just a quirk of oxlint/unicorn.
I don't consider the reason, good or 🤖trusty but I also disagree with unicorns latest recommended rules, so that isn't anything unique in the eslint/oxlint/unicorn world.

The speed up is nice. If we encounter any major issues, we can revert back to eslint.

docs/.vitepress/components/AsideSponsors.vue
```diff
@@ -36,6 +36,7 @@ const sponsors = computed<Sponsor[]>(
       .filter((backer) => backer.account.imageUrl)
       .map((backer) => ({
         name: backer.account.name ?? backer.account.slug,
+        // oxlint-disable-next-line typescript/no-non-null-assertion -- Guaranteed by the filter above
         img: backer.account.imageUrl!,
```
xDivisionByZerox, line 40: nitpick: This could probably be "solved" by creating a dedicated type gaaurd

```ts
function isBackerWithImage(value: Backer): value is Backer & { account: { imageUrl: string }  } {
  return value.account.imageUrl != null;
}

// further down
backersStorage.value?.backers
     .filter(isBackerWithImage)
     .map((backer) => ({
       name: backer.account.name ?? backer.account.slug,
        // imageUrl should not be nullish anymore
       img: backer.account.imageUrl!,
```

oxlint.config.ts
```diff
@@ -0,0 +1,397 @@
...
+    //#endregion
+
+    //#region typescript
+    // name: 'typescript overrides'
+    'typescript/array-type': [
+      'error',
+      { default: 'array-simple', readonly: 'generic' },
+    ],
+    'typescript/ban-ts-comment': 'error',
+    'typescript/consistent-return': 'off', // not in the original ESLint config
```
xDivisionByZerox: Are these "not in original ..." comments required (I saw 4; 2 with a separation concern)? This seems like a heads up from your agent. After this PR is merged, this comment would not add any meaningful information IMO.

Shinigami92: yes, I left them in for easier review, I can remove them tomorrow

docs/.vitepress/components/AsideSponsors.vue
Shinigami92: will look into this tomorrow

src/internal/locale-proxy.ts
```diff
@@ -45,7 +45,7 @@ function isLocaleProxy(value: unknown): value is LocaleProxy {
   return (
     value != null &&
     typeof value === 'object' &&
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // oxlint-disable-next-line typescript/no-explicit-any
     (value as any)?.[LOCALE_PROXY_TAG] === true
```
ST-DDT: Maybe:

```suggestion
    (value as LocaleProxy)[LOCALE_PROXY_TAG] === true
```

src/internal/bind-this-to-member-functions.ts
ST-DDT:
```suggestion
export function bindThisToMemberFunctions<TClass extends { new (): TClass }>(
  instance: InstanceType<TClass>
): void {
  let p = Object.getPrototypeOf(instance);
  do {
    for (const name of Object.getOwnPropertyNames(p) as Array<
      keyof typeof instance
    >) {
```

ST-DDT (dismissed):

src/internal/bind-this-to-member-functions.ts
```diff
@@ -11,13 +11,20 @@
  * const someMethod = someModule.someMethod;
  * someMethod(); // Works
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// oxlint-disable-next-line typescript/no-explicit-any
 export function bindThisToMemberFunctions<TClass extends { new (): any }>(
   instance: InstanceType<TClass>
 ): void {
   let p = Object.getPrototypeOf(instance);
   do {
```
Shinigami92: **suggestion:** 

```suggestion
```

src/internal/bind-this-to-member-functions.ts
Shinigami92: looks like that suggestion was corrupted, fixed it with another suggestion in webui

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3783, opened 2026-03-31
feat(date): add ability to provide year range for `past` and `future`

### LINKED ISSUE #3653: past/future date functions to support min years

currently `past` and `future` supports only option to set max range in years
In case test needs short date in past/future (`.toISOString().slice(0, 10)` magic 🪄) you can still get date of current day

### Suggested solution

to avoid adjusting reference date or manually adding +/- days it would be nice to support additional config to set min diff

e.g.
```
faker.date.past({yearsMin: 1, years: 2}) // at least 1 year but not more than 2 years in the past

faker.date.future({yearsMin: 1, years: 2}) // at least 1 year but not more than 2 years in the future
```

I guess same would apply for `soon` and `recent`

### Alternative

I can use date-fns to add/remove random number of days but it feels bit to ugly

```ts
addDays(
          new Date(),
          faker.number.int({ min: 1, max: 20 }),
        )
          .toISOString()
          .slice(0, 10)
```

### Additional context

_No response_

xDivisionByZerox:
# Description

This PR fixes #3653.

## Changes

- Add a small helper function to convert a number of years to is ms representation. This is mostly done for readability purposes.
- Extend the `years` argument in `date.past()` and `date.future()` to be able to provide a range
  - For the naming and JSDoc approach I took inspiration from the string module
  - `year.max` must be greater than `years.min` - this will be enforced by `date.between()` (the underlying implementation) as well, but I added an additional check to give a nicer error message
- Added `throw` tags to the JSDocs of the functions - these where previously missing even tho the functions were able to throw
- Added tests for the new behavior
  - Aligned the exist test of `date.future` with the one from `date.past` since they were more strict

> [!NOTE]
> I quite enjoyed implementing an actual feature again. I would go ahead and implement the same feature for `date.recent` and `date.soon` in a separate PR if the team allows it.

## Links

- #3653


src/modules/date/index.ts
```diff
@@ -75,32 +103,46 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+    const { refDate = this.faker.defaultRefDate() } = options;
+    let { years = 1 } = options;
+    if (typeof years === 'number') {
+      years = { min: 0, max: years };
+    }
 
-    if (years <= 0) {
+    if (years.max <= 0) {
       throw new FakerError('Years must be greater than 0.');
     }
```
xDivisionByZerox, line 114: I was not sure if we should really add this, but should there also be an error if `years.min` is less than `0`? If that is the case someone would likely want to call `date.future` instead. Same goes for the implementation of `date.future` as well.

src/modules/date/index.ts
```diff
@@ -47,14 +56,18 @@ export class SimpleDateModule extends SimpleModuleBase {
...
    *
+   * @throws {FakerError} If `years.max` is less than 0.
+   * @throws {FakerError} If `years.min` is greater or equal than `years.max`.
+   *
    * @see faker.date.recent(): For generating dates in the recent past (days instead of years).
    *
    * @example
    * faker.date.past() // '2021-12-03T05:40:44.408Z'
    * faker.date.past({ years: 10 }) // '2017-10-25T21:34:19.488Z'
+   * faker.date.future({ years: { min: 4, max: 7 } }) // '2022-12-12T03:43:16.434Z'
```
xDivisionByZerox:
```suggestion
   * faker.date.past({ years: { min: 4, max: 7 } }) // '2022-12-12T03:43:16.434Z'
```

test/modules/__snapshots__/date.spec.ts.snap
```diff
@@ -81,7 +81,9 @@ exports[`date > 42 > future > with only number refDate 1`] = `2021-07-08T10:07:3
 
 exports[`date > 42 > future > with only string refDate 1`] = `2021-07-08T10:07:33.524Z`;
 
-exports[`date > 42 > future > with value 1`] = `2024-11-19T18:52:08.216Z`;
+exports[`date > 42 > future > with value numeric 1`] = `2024-11-19T18:52:08.216Z`;
```
xDivisionByZerox, line 84: Sadly the diff does not show that only the name of the test changed for this line. I wanted to emphasizes that this feature does not change the generated values per seed!

src/modules/date/index.ts
```diff
@@ -47,14 +56,18 @@ export class SimpleDateModule extends SimpleModuleBase {
    * Generates a random date in the past.
    *
    * @param options The optional options object.
-   * @param options.years The range of years the date may be in the past. Defaults to `1`.
+   * @param options.years The range of years the date may be in the past. Either as a fixed amount of years or as a year range. Defaults to `1`.
```
ST-DDT, line 59: This sounds a bit redundant with the first sentence.

Maybe use:

1) The range of years the date may be in the past. Either as `0-n` years in the past or `min-max` years in the past. Defaults to `1`.

ST-DDT (commented):
We should apply the adjustments to recent and soon as well.

xDivisionByZerox:
> We should apply the adjustments to recent and soon as well.

As I said in the PR description, I would suggest providing an additional PR for this. Alternatively, the PR title could be extended even further.
All I want is that the concrete functions are named in the autogenerated changelog when releasing.

src/modules/date/index.ts
xDivisionByZerox: I'll let an other reviewer decide this one. I'm not sure about the suggestion you made, but I can also not find arguments against it, other than "I personally don't like it".

src/modules/date/index.ts
ST-DDT: if the year is negative, then its not in the past.
Maybe just add it to be safe. 

Instead of checking max, you could just check min, since max needs to be > min anyway.

src/modules/date/index.ts
ST-DDT: Alternatively, one of these:

2) Specifies how many years in the past the date may be. Accepts either a single value (`0–n`) or a range (`min–max`). Defaults to 1.

3) Defines the allowable range of years in the past for the date. Can be provided as a single number (`0-n` years ago) or as a range (`min-max` years ago). Defaults to 1.

src/modules/date/index.ts
```diff
@@ -75,32 +103,46 @@ export class SimpleDateModule extends SimpleModuleBase {
...
+      );
+    }
+
     const time = toDate(refDate).getTime();
 
     return this.between({
-      from: time - years * 365 * 24 * 3600 * 1000,
-      to: time - 1000,
+      from: time - yearsToMs(years.max),
+      to: time - yearsToMs(years.min) - 1000,
```
ST-DDT, line 126: I'm also considering increasing the safety margin of this from 1s to 1d. recent/soon stay with 1s.

ST-DDT:
> As I said in the PR description, I would suggest providing an additional PR for this. Alternatively, the PR title could be extended even further.

Sorry, have missed that part of the description. Separate PR would be more clear.

src/modules/date/index.ts
xDivisionByZerox: Would that be considered breaking? If so, I'd provide a separate "fix" PR, since this PR is only adding a feature. It should (IMO) not change existing behavior.

ST-DDT: I wouldnt call ot breaking.
Separate PR, if we do it.

ST-DDT (approved):

Shinigami92 (approved):

PR #3786, opened 2026-04-04
refactor: optimize core modules for stability and type safety

hiSandog:
## Summary

This PR introduces 4 optimizations to core modules that improve stability, type safety, and code maintainability.

#### 1. Convert `fake()` from recursive to iterative
- **File:** `src/modules/helpers/index.ts`
- **Why:** The recursive implementation of `fake()` could cause stack overflow when processing deeply nested template patterns (e.g. maliciously crafted input).
- **How:** Rewrote the method as a simple `while(true)` loop that iteratively replaces template tags.

#### 2. Add upper bound to `*` and `+` quantifier limits
- **File:** `src/modules/helpers/index.ts`
- **Why:** The exponential growth pattern (`limit *= 2`) had no ceiling. In theory, an unlucky sequence of random `true` returns from `datatype.boolean()` could generate astronomically large limits.
- **How:** Capped the limit at 256 for both quantifiers, which is more than sufficient for realistic use cases.

#### 3. Extract `fakeFromPattern()` utility
- **File:** `src/internal/fake-from-pattern.ts` (new), `src/modules/hacker/index.ts`
- **Why:** Multiple modules contain identical boilerplate calls like `this.faker.helpers.fake(this.faker.definitions.X.pattern)`. Extracting this to a utility function improves consistency and reduces repetition.
- **How:** Created a new internal helper function and updated the hacker module's `phrase()` method as the first consumer.

#### 4. Tighten `Callable` type
- **File:** `src/internal/types.ts`
- **Why:** The `Callable` type used `...args: any[]` which bypasses all type checking on function parameters.
- **How:** Changed to `...args: readonly unknown[]` for improved type safety while maintaining compatibility with all existing call sites.

---

All existing tests pass.


ST-DDT:
@hiSandog Please let us know, when you think this PR is ready for review.

matthewmayer:
3 seems somewhat like premature abstraction. Do you think it makes code easier to read or maintain?

hiSandog:
@ST-DDT Thanks for keeping this on my radar! I've addressed matthewmayer's feedback by removing the \`fakeFromPattern` helper — agreed that wrapping a single use was unnecessary abstraction. The rest of the changes are the stack overflow fix and the \`Callable` type cleanup, both of which are ready for review. Let me know if there's anything else to adjust!

PR #3795, opened 2026-04-07
fix(helpers): ignore wrapper characters in fromRegExp

ST-DDT:
Fixes #3663

- #3663

---

Ignore the anchor characters at the start and end of regex patterns.

````ts
faker.helpers.fromRegExp(/^foo$/)
// old: ^foo$
// new: foo
````

### CODE REVIEW

**Comment by matthewmayer** [comment]
this is arguably breaking, although seems doubtful anyone was relying on this?

File: src/modules/helpers/index.ts [hunk 1]
```diff
@@ -371,7 +371,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
     if (pattern instanceof RegExp) {
       isCaseInsensitive = pattern.flags.includes('i');
       pattern = pattern.toString();
-      pattern = /\/(.+?)\//.exec(pattern)?.[1] ?? ''; // Remove frontslash from front and back of RegExp
+      pattern = /\/\^?(.+?)\$?\//.exec(pattern)?.[1] ?? ''; // Remove slash and anchors from front and back of RegExp
     }
```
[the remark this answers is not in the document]
ST-DDT: This pattern only ignores ^ and $, the method itself does not support escape characters.
So we ignore them here as well.

test/modules/helpers.spec.ts
```diff
@@ -689,6 +689,11 @@ describe('helpers', () => {
           expect(actual).toHaveLength(9);
           expect(actual).toMatch(/^[A-D0-9]{4}-[A-D0-9]{4}$/i);
         });
+
+        it('hides regex wrapper characters', () => {
+          const actual = faker.helpers.fromRegExp(/^foo$/i);
+          expect(actual).toBe('foo');
+        });
```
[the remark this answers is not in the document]
ST-DDT: The method does not handle escaped characters at all. So we won't add a test for that either.

ST-DDT:
The method basically only implements a third of all regex features. So we could declare this as `wont't fix`, but ^$ are common regex wrappers IMO..

Once the standalone module functions PR is merged, then I'll consider adding a `faker-plugins-string-regexp2factory` or similar package, that properly parses the regex into an AST (using a lib) in order to generate the output.

hiSandog:
The regex `/\/\^?(.+?)\$?\/` correctly strips optional `^` and `'$'` anchors from the RegExp string representation. 

One edge case to consider: if someone passes `/^^foo$$/` (multiple anchors, which is valid regex but redundant), the current pattern would produce `^foo$` since `^?` and `$?` only match zero or one occurrence. This is unlikely in practice, but using `^\^*` and `\$*` (zero or more) would be more robust:

```ts
pattern = /\/\^*(.+?)\$*\//.exec(pattern)?.[1] ?? '';
```

Given that `fromRegExp` only implements a subset of regex features (as noted in the discussion above), the current fix is pragmatic for the common `/^foo$/` case.

ST-DDT:
@faker-js/maintainers Thoughts?

Shinigami92:
> @faker-js/maintainers Thoughts?

LGTM, this is a bug fix, not a breaking change - nobody should be relying on literal `^`/`$` leaking into the output.

**suggestion:** consider using `pattern.source` instead of parsing `toString()`, it gives you the pattern without slashes/flags directly:

```ts
if (pattern instanceof RegExp) {
  isCaseInsensitive = pattern.flags.includes('i');
  pattern = pattern.source.replace(/^\^/, '').replace(/\$$/, '');
}
```

a bit more readable and avoids the slash-parsing regex entirely.
Either way, happy to approve.

Shinigami92 (dismissed):

src/modules/helpers/index.ts
```diff
@@ -370,8 +370,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
 
     if (pattern instanceof RegExp) {
       isCaseInsensitive = pattern.flags.includes('i');
-      pattern = pattern.toString();
-      pattern = /\/(.+?)\//.exec(pattern)?.[1] ?? ''; // Remove frontslash from front and back of RegExp
+      pattern = pattern.source.replace(/^\^/, '').replace(/\$$/, '');
```
xDivisionByZerox: > > The regex `/\/\^?(.+?)\$?\/` correctly strips optional `^` and `'$'` anchors from the RegExp string representation. 
> > 
> > One edge case to consider: if someone passes `/^^foo$$/` (multiple anchors, which is valid regex but redundant), the current pattern would produce `^foo$` since `^?` and `$?` only match zero or one occurrence. > > This is unlikely in practice, but using `^\^*` and `\$*` (zero or more) would be more robust:
> > 
> > ```ts
> > pattern = /\/\^*(.+?)\$*\//.exec(pattern)?.[1] ?? '';
> > ```
> > 
> > Given that `fromRegExp` only implements a subset of regex features (as noted in the discussion above), the current fix is pragmatic for the common `/^foo$/` case.
>
> @faker-js/maintainers Thoughts?

Lets just strip all of the wrapper characters in the front and back.

ST-DDT: The current replacements do exactly that. `/^\^/` -> `''` (Replace ^ at the start of the string)

xDivisionByZerox: Yes, but only the first one, no? The finding from @hiSandog was that multiple characters would NOT correctly get removed. Afterwards you asked for thought, so my thought is to strip **ALL** wrapper characters in the beginning and end.

Edit: Or am I am I missing something? From the replacement I'm seeing, only one character gets removed.

ST-DDT: I originally considered multi wrapper a won't fix. But I changed it now.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3796, opened 2026-04-07
feat(locale): add mn_MN_cyrl (Mongolian) locale

btseee:
As a native Mongolian speaker I'm adding full Mongolian localization.

Баярлалаа (Thank you)


matthewmayer:
Баярлалаа!

For completely new locales, we recommend to split into multiple PRs as it makes it much easier to review. For the first PR please include only the `metadata` and one or two simple modules e.g. `faker.color` or `faker.date`. This lets us resolve any issues about locale naming, scripts etc.

You can then follow up with additional PRs for other modules.

btseee:
Dear @matthewmayer,

As you requested i only entered Color and Date module with metadata now.

Please let me know what you think.

matthewmayer:
Content looks fine. My main question to the maintainers before we proceed is does it make sense to go for `mn` or `mn_MN` (Mongolia) - more recent locales we have tended to use the language_country version as it gives us more flexibility in the future (say if someone wants to contribute a version for Mongolian speakers in another country with appropriate phone and location data).

btseee:
mn_MN works and is more precise.

One more follow-up: in Mongolia, two writing systems are in use , Cyrillic (Bulgarian-based) which we currently did and traditional Mongolian script. Would it be better to distinguish these at the locale level (e.g., mn_Cyrl_MN vs mn_Mong_MN), or keep them unified under mn_MN and handle script differences at a different layer in the future?

Example Mongolian script: 
image

matthewmayer:
We typically do that as a suffix e.g. existing locales:

```
uz_UZ_latin
sr_RS_latin
```

So if you think a future mn_MN_mong is possible, we should do
```
mn_MN_cyrl
```

btseee:
Perfect. Than let's proceed with `mn_MN_cyrl` for now. Should i do this change or should you?

matthewmayer:
You should :)

matthewmayer:
Note you'll also need to add some extra keys to metadata. See sr_RS_latin for an example.

btseee:
@matthewmayer It's corrected now thank you.

src/locales/mn_MN_cyrl/color/human.ts
```diff
@@ -0,0 +1,22 @@
+export default [
+  'алтан',
+  'бараан хөх',
+  'зэс',
+  'мөнгөн',
+  'нил',
+  'ногоон',
+  'оранж',
```
ST-DDT: My translator suggestions улбар шар instead.

src/locales/mn_MN_cyrl/color/human.ts
```diff
@@ -0,0 +1,22 @@
+export default [
+  'алтан',
+  'бараан хөх',
+  'зэс',
+  'мөнгөн',
+  'нил',
```
ST-DDT: Maybe нил ягаан?

src/locales/mn_MN_cyrl/color/human.ts
```diff
@@ -0,0 +1,22 @@
...
+  'хар',
+  'хүрэн',
+  'хөх',
+  'цагаан',
+  'цайвар ногоон',
+  'цайвар улаан',
+  'шар',
+  'ягаан',
+  'ягаан улаан',
+  'ягаанавар',
```
ST-DDT: My translator tool had issues with this. Please check if correct.

ST-DDT (dismissed):
My translation tool had some trouble with some of the colors, but I'm not a native speaker so please check if there are any issues. Otherwise this looks good to me.

btseee:
@ST-DDT rewritten color module completely with English translation as comment.

matthewmayer:
you'll need to update the test snapshots one more time with `pnpm run test:update-snapshots`

ST-DDT (dismissed):

ST-DDT:
Please run `pnpm run preflight`

matthewmayer (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #3797, opened 2026-04-07
feat(locale): improved en_IE postal_address

matthewmayer:
Note en_IE uses county not state

### DIFF
--- a/src/locales/en_IE/location/index.ts
+++ b/src/locales/en_IE/location/index.ts
@@ -5,12 +5,14 @@
 import type { LocationDefinition } from '../../..';
 import city_pattern from './city_pattern';
 import county from './county';
+import postal_address from './postal_address';
 import postcode from './postcode';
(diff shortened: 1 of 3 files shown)


ST-DDT (approved):

Shinigami92 (approved):

hiSandog:
The `postal_address` templates reference `{{location.secondaryAddress}}` in the first pattern, but `en_IE` doesn't define a `secondary_address.ts` file. When that template is selected, `secondaryAddress` will fallback to the base locale or return unexpected values.

Two options:
1. Add an `secondary_address.ts` file for `en_IE` with Irish-specific secondary address formats (e.g., flat numbers, apartment references)
2. Remove the `{{location.secondaryAddress}}` line from the template if it's not commonly needed

Also, the `Co. {{location.county}}` format is correct for Irish addresses — good use of the traditional "County" abbreviation.

PR #3798, opened 2026-04-07
feat(locale): Add postal_address and improved secondary_address for es

matthewmayer:
Notes
- typical Spanish address uses City or City&Province (county) not Comunidades Autónomas (state)
- more realistic floor/room data
- added clarifying comments


ST-DDT (approved):

Shinigami92 (approved):

hiSandog:
The `postal_address` templates correctly reference `{{location.county}}` for the province level, which aligns with the comment added to `county.ts` identifying these as "Provincias" rather than autonomous communities.

One suggestion: consider adding a clarifying comment in `postal_address.ts` explaining why `county` is used instead of `state` for the regional level, similar to how the PR description notes that Spanish addresses use City/Province rather than Comunidades Autónomas. This helps future maintainers understand the design choice when reviewing the template patterns.

PR #3802, opened 2026-04-11
fix: link release checklist in release PR workflow

hiSandog:
## Summary
- replace the placeholder release checklist TODO in the release PR workflow
- look up the current open c: release issue and include its URL when present
- omit the checklist section when no release checklist issue is open

## Validation
- pnpm exec prettier --check .github/workflows/prepare-release-pr.yml


ST-DDT:
Cool idea. How does this behave if the release checklist issue does not exist yet?

hiSandog:
Good question! If the release checklist issue does not exist yet, `gh issue list` returns an empty array, so `RELEASE_CHECKLIST_URL` becomes an empty string. The `if [ -n "$RELEASE_CHECKLIST_URL" ]` guard then skips the checklist section entirely — the PR body just contains the release version and the manual tasks checkbox, with no broken or placeholder link.

In practice, the checklist issue is created as part of the release prep process (it is an issue labeled `c: release`), so by the time the release PR workflow runs, the issue should already exist. But even if it does not, the workflow handles it gracefully without errors.

ST-DDT:
Should we cancel/fail the process instead?
@faker-js/maintainers What do you think?

.github/workflows/prepare-release-pr.yml
```diff
@@ -64,19 +64,22 @@ jobs:
 
       - name: Create draft PR
         run: |
+          RELEASE_CHECKLIST_URL="$(gh issue list --label 'c: release' --state open --limit 1 --json url --jq '.[0].url // ""')"
```
xDivisionByZerox: This is where I would default to the old text.


```suggestion
          RELEASE_CHECKLIST_URL="$(gh issue list --label 'c: release' --state open --limit 1 --json url --jq '.[0].url // "TODO add link to issue"')"
```

xDivisionByZerox (commented):
> Should we cancel/fail the process instead?
> @faker-js/maintainers What do you think?

Usually I would be in favor of failing the PR creation, because not having a release issue would mean that someone is fundamentally going against our internal release process.
But since that would require reverting all other steps as well, which would increase the complexity of the workflow file more that we gain from the change, I'd say that we simply default to the old text inside the jq expression (see https://github.com/faker-js/faker/pull/3802/changes#r3069395214). 
By doing that, the instantiation of the `PR_BODY` variable should also be more straight forward, since `RELEASE_CHECKLIST_URL` should not be empty anymore.

---

Does this change evaluate to incorrect issue numbers if we ever forget to close an old release issue? I don't think that this is worth addressing if it does, since the chance of this happening is very slim. I'm just curious if it would. I guess it wouldn't due to the `--limit 1` and/or the default sort order being "newest first" :thinking:

hiSandog:
@xDivisionByZerox Good call! Applied your suggestion — the jq expression now defaults to `"TODO add link to issue"` instead of an empty string, which simplifies the PR_BODY construction by removing the if/else guard entirely. Pushed in d1ea36ba.

.github/workflows/prepare-release-pr.yml
```diff
@@ -64,19 +64,18 @@ jobs:
...
 
-          - [ ] Completed manual changes/tasks for this release
+            - [ ] Completed manual changes/tasks for this release
 
-          ---
+            ---
 
-          - Checklist: TODO add link to issue
-          "
+            - Checklist: $RELEASE_CHECKLIST_URL"
```
ST-DDT: Please retain the original indentation here, as it gets reflected in the final PR description.
```suggestion
          - [ ] Completed manual changes/tasks for this release

          ---

          - Checklist: $RELEASE_CHECKLIST_URL
          "
```

ST-DDT (approved):

xDivisionByZerox (approved):

hiSandog:
Just checking in on this PR. Could you please merge it at your convenience? Thanks a lot! @xDivisionByZerox

PR #3805, opened 2026-04-12
chore: import types from the source file directly

ST-DDT:
Import directly from the source file to avoid automatically adding new imports to it, causing side effects.


Shinigami92 (approved):

xDivisionByZerox (approved):
Is there a linter rule to restrict the importing from the index.ts file for all other files by any change? I can see this slipping in again in later refactors due to IDEs usually preferring to pick the shortest import path that is available.

Issue #3810, opened 2026-04-16
Add more categories for static images


matthewmayer:
## Background

In older versions of Faker, the `faker.image` module was heavily reliant on multiple third-party image placeholder services e.g. Unsplash, LoremPixel, LoremFlickr and Picsum. Some of these allowed requesting images by categories, e.g. you could generate images right using simple methods like faker.image.cats(), faker.image.nature(), or faker.image.business().

The full list of old category methods was `abstract, animals, business, cats, city, food, nature, nightlife, people, sports. technics, transport`

Unfortunately over time, several of these providers became rate-limited or went down permanently. As of v10.x, only Picsum survives and it doesn't support categorized images.

In v9.5.0 we tried a new approach for the `faker.image.personPortrait` API - https://v10.fakerjs.dev/api/image.html#personportrait - it references static images, originally generated by an opensource GenAI image modal, in a [seperate repo](https://github.com/faker-js/assets-person-portrait) served via the jsdelivr CDN.

This new approach seems to have worked smoothly, and has quietly become very popular, with 1.7M requests/week made to the assets-person-portrait repository.

## Proposal

It seems there is an opportunity to add more similar categorized static images that are not dependent on a specific third party provider.

Similar to `faker.image.personPortrait` we would support a fixed range of sizes (square only) and perhaps some simple parameters.

This issue can be for general discussion of the API and upvoting. If you have a specific method to propose e.g. `faker.image.animal({type:"cat|dog"})` please add as a seperate sub-issue and reference this one.

ST-DDT:
**Community Input Required**

If you need additional modules feel free to create new sub-issues (feature request).
Please include a short summary with:

- what you need them for
- and whether you would like to have specific parameters
- and whether they must meet specific criteria

PR #3815, opened 2026-04-17
docs(image): link to tracking issue

ST-DDT:
Adds a link to #3810 to the image module.

- #3810

Preview:

- https://deploy-preview-3815.fakerjs.dev/api/image.html

IDE:
IDE

Website:
WEBSITE

### DIFF
--- a/src/modules/image/index.ts
+++ b/src/modules/image/index.ts
@@ -15,6 +15,9 @@ import type { SexType } from '../person';
  * For a random user avatar image, use [`avatar()`](https://fakerjs.dev/api/image.html#avatar), or [`personPortrait()`](https://fakerjs.dev/api/image.html#personportrait) which has more control over the size and sex of the person.
  *
  * For full control over the returned image URL, use [`faker.helpers.arrayElement()`](https://fakerjs.dev/api/helpers.html#arrayelement) with your own array of image URLs.
+ *
+ * ::: info Hint
+ * If you think an image method/category is missing, please [open an issue/vote for an existing one](https://github.com/faker-js/faker/issues/3810).
  */
 export class ImageModule extends ModuleBase {
   /**


Shinigami92 (approved):

matthewmayer (approved):

src/modules/image/index.ts
```diff
@@ -15,6 +15,9 @@ import type { SexType } from '../person';
  * For a random user avatar image, use [`avatar()`](https://fakerjs.dev/api/image.html#avatar), or [`personPortrait()`](https://fakerjs.dev/api/image.html#personportrait) which has more control over the size and sex of the person.
  *
  * For full control over the returned image URL, use [`faker.helpers.arrayElement()`](https://fakerjs.dev/api/helpers.html#arrayelement) with your own array of image URLs.
+ *
+ * ::: info Hint
```
xDivisionByZerox, line 19: All other `::: info` Block have "Note" as the title if only one word follows.

---

I don't know how our internal parser handles this syntax, but usually we close the message boxes at the end in markdown files.

ST-DDT: I used Hint here to distinguish it from the other (green) note blocks.

---

I don't know exactly which tool gets confused, but we already have a `:::` in the generated docs page.
If I add a closing block, then the `::::` show up as is. I forgot what v-pre does for us here.

https://github.com/faker-js/faker/blob/e1f6ce251d3bad350abfcd4e225c7ff3dcefc4f7/scripts/apidocs/output/page.ts#L63-L79

Not sure how to tell it that those are different blocks.

xDivisionByZerox: Hm, alright. Fine by me :+1:
Thanks for explaining the issue to me.

xDivisionByZerox (approved):

PR #3817, opened 2026-04-18
docs: render multiple @example blocks as tabbed code groups

Shinigami92:
potential replacement for #3585

- #3585

image

https://deploy-preview-3817.fakerjs.dev/api/faker.html

### DIFF
--- a/docs/.vitepress/components/api-docs/refreshable-code.vue
+++ b/docs/.vitepress/components/api-docs/refreshable-code.vue
@@ -14,7 +14,9 @@ const {
 }>();
 
 const code = useTemplateRef('code');
-const codeBlock = computed(() => code.value?.querySelector('div pre code'));
+// Scope to the wrapper so that multiple `<pre><code>` blocks (one per tab in a
+// VitePress code-group) are all covered. `.line` selectors below rely on this.
+const codeBlock = computed(() => code.value);
 const codeLines = ref<Element[]>();
 
 function initRefresh(): Element[] {
--- a/docs/.vitepress/components/api-docs/refreshable-code.vue
+++ b/docs/.vitepress/components/api-docs/refreshable-code.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, onMounted, ref, useTemplateRef } from 'vue';
+import { onMounted, ref, useTemplateRef } from 'vue';
 import { formatResult } from './format';
 import RefreshButton from './refresh-button.vue';
 
@@ -14,16 +14,13 @@ const {
 }>();
 
 const code = useTemplateRef('code');
-// Scope to the wrapper so that multiple `<pre><code>` blocks (one per tab in a
-// VitePress code-group) are all covered. `.line` selectors below rely on this.
-const codeBlock = computed(() => code.value);
 const codeLines = ref<Element[]>();
 
 function initRefresh(): Element[] {
-  if (codeBlock.value == null) {
+  if (code.value == null) {
     return [];
   }
-  const domLines = codeBlock.value.querySelectorAll('.line');
+  const domLines = code.value.querySelectorAll('.line');
   let lineIndex = 0;
   const result: Element[] = [];
   while (lineIndex < domLines.length) {
@@ -84,13 +81,13 @@ function initRefresh(): Element[] {
 }
 
 async function onRefresh(): Promise<void> {
-  if (refresh != null && codeBlock.value != null) {
+  if (refresh != null && code.value != null) {
     codeLines.value ??= initRefresh();
 
     const results = await refresh();
 
     // Remove old comments
-    codeBlock.value
+    code.value
       .querySelectorAll('.comment-delete-marker')
       .forEach((el) => el.remove());
 
--- a/scripts/apidocs/output/page.ts
+++ b/scripts/apidocs/output/page.ts
@@ -2,7 +2,12 @@ import { writeFileSync } from 'node:fs';
 import { resolve } from 'node:path';
 import type { ApiDocsMethod } from '../../../docs/.vitepress/components/api-docs/method';
 import { formatMarkdown, formatTypescript } from '../../shared/format';
-import { adjustUrls, codeToHtml, mdToHtml } from '../../shared/markdown';
+import {
+  adjustUrls,
+  codeGroupToHtml,
+  codeToHtml,
+  mdToHtml,
+} from '../../shared/markdown';
 import { FILE_PATH_API_DOCS } from '../../shared/paths';
 import { toRefreshableCode } from '../../shared/refreshable-code';
 import type { RawApiDocsPage } from '../processing/class';
@@ -74,7 +79,7 @@ async function writePageMarkdown(page: RawApiDocsPage): Promise<void> {
 
   ${adjustUrls(description)}
 
-  ${examples.length === 0 ? '' : `<div class="examples">${await codeToHtml(examples.join('\n'))}</div>`}
+  ${examples.length === 0 ? '' : `<div class="examples">${await codeGroupToHtml(examples)}</div>`}
 
   :::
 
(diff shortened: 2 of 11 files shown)


scripts/shared/markdown.ts
```diff
@@ -81,6 +86,32 @@ export async function codeToHtml(code: string): Promise<string> {
...
+export async function codeGroupToHtml(codes: string[]): Promise<string> {
+  if (codes.length <= 1) {
+    return codeToHtml(codes.join('\n'));
+  }
+
+  const delimiter = '```';
+  const blocks = codes
+    .map(
+      (code, index) =>
+        `${delimiter}ts [Example ${index + 1}]\n${code}\n${delimiter}`
```
ST-DDT: IMO we should require titles if multiple examples are present:
If you have them as a comment, then it doesn't negatively impact IDE tooltips.
IMO it also helps clarifing what that individual example is used for.

````ts
// Title
[...]
````

**Review by ST-DDT — COMMENTED** [review]
Could you please add multi example test cases to our tests?

- https://github.com/faker-js/faker/blob/04f21d30cb86596b41e29ba539721e416c8245a9/test/scripts/apidocs/class.example.ts#L27-L33
- https://github.com/faker-js/faker/blob/04f21d30cb86596b41e29ba539721e416c8245a9/test/scripts/apidocs/method.example.ts#L338-L348

I hope that these will get picked up automatically, if you just add them to the file.

File: test/scripts/shared/markdown.spec.ts [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+  beforeAll(async () => {
+    await initMarkdownRenderer();
+  });
+
+  describe('codeGroupToHtml()', () => {
+    it('renders a plain code block for a single example', async () => {
+      const html = await codeGroupToHtml(['const a = 1;']);
+
+      expect(html).not.toContain('vp-code-group');
+      expect(html).toContain('const');
```
ST-DDT: How about using a snapshot test here as well?

test/scripts/shared/markdown.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    it('renders a tabbed code group with titles from leading comments', async () => {
+      const html = await codeGroupToHtml([
+        '// First title\nconst a = 1;',
+        '// Second title\nconst b = 2;',
+      ]);
+
+      expect(html).toContain('vp-code-group');
+      expect(html).toContain('First title');
+      expect(html).toContain('Second title');
```
ST-DDT: How about using a snapshot test here as well?

scripts/shared/markdown.ts
```diff
@@ -81,6 +86,57 @@ export async function codeToHtml(code: string): Promise<string> {
...
+}
+
+const codeGroupTitleCommentRegex = /^\s*\/\/\s*(.+?)\s*$/;
+
+function extractCodeGroupTitle(
+  code: string,
+  index: number
+): { title: string; body: string } {
+  const newlineIndex = code.indexOf('\n');
+  const firstLine = newlineIndex === -1 ? code : code.slice(0, newlineIndex);
```
ST-DDT: If the example only consists of the title, then we should fail anyway.

```suggestion
  const firstLine = newlineIndex === -1 ? '' : code.slice(0, newlineIndex);
```

scripts/shared/markdown.ts
```diff
@@ -81,6 +86,57 @@ export async function codeToHtml(code: string): Promise<string> {
...
+  const match = codeGroupTitleCommentRegex.exec(firstLine);
+  if (match == null) {
```
ST-DDT:
```suggestion
  if (!firstLine.startsWith('// ') {
```

scripts/shared/markdown.ts
```diff
@@ -81,6 +86,57 @@ export async function codeToHtml(code: string): Promise<string> {
...
+    throw new Error(
+      `Example ${index + 1} in a multi-example block must start with a \`// Title\` line comment to label the code-group tab, but got: ${JSON.stringify(firstLine)}`
+    );
+  }
+
+  const body = newlineIndex === -1 ? '' : code.slice(newlineIndex + 1);
+
+  return { title: match[1], body };
```
ST-DDT:
```suggestion
  return { title: firstLine.substring(3), body };
```

scripts/shared/markdown.ts
```diff
@@ -81,6 +86,57 @@ export async function codeToHtml(code: string): Promise<string> {
...
+  const newlineIndex = code.indexOf('\n');
+  const firstLine = newlineIndex === -1 ? code : code.slice(0, newlineIndex);
+  const match = codeGroupTitleCommentRegex.exec(firstLine);
+  if (match == null) {
+    throw new Error(
+      `Example ${index + 1} in a multi-example block must start with a \`// Title\` line comment to label the code-group tab, but got: ${JSON.stringify(firstLine)}`
+    );
+  }
+
+  const body = newlineIndex === -1 ? '' : code.slice(newlineIndex + 1);
```
ST-DDT: In the worst case it will return the full body, but we shouldn't run into that case here anyway.

```suggestion
  const body = code.slice(newlineIndex + 1);
```

scripts/shared/markdown.ts
```diff
@@ -81,6 +86,57 @@ export async function codeToHtml(code: string): Promise<string> {
...
+function extractCodeGroupTitle(
+  code: string,
+  index: number
+): { title: string; body: string } {
+  const newlineIndex = code.indexOf('\n');
+  const firstLine = newlineIndex === -1 ? code : code.slice(0, newlineIndex);
+  const match = codeGroupTitleCommentRegex.exec(firstLine);
+  if (match == null) {
+    throw new Error(
+      `Example ${index + 1} in a multi-example block must start with a \`// Title\` line comment to label the code-group tab, but got: ${JSON.stringify(firstLine)}`
```
ST-DDT: Is from the call context clear which method the example belongs to?
If not, maybe print the full example, to make finding it easier.

scripts/apidocs/output/page.ts
```diff
@@ -199,7 +204,7 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
       throws.length === 0 ? undefined : await mdToHtml(throws.join('\n'), true),
     returns: returns.text,
     signature: await codeToHtml(formattedSignature),
-    examples: await codeToHtml(examples.join('\n')),
+    examples: await codeGroupToHtml(examples),
```
ST-DDT, line 207: This causes conflicts with the refresh button, since the code group likely uses a different tree structure.

> Uncaught (in promise) TypeError: can't access property "insertAdjacentHTML", domLine is undefined

Tested with helper module:

````ts
  /**
   * Replaces the symbols and patterns in a credit card schema including Luhn checksum.
   *
   * This method supports both range patterns `[4-9]` as well as the patterns used by `replaceSymbolWithNumber()`.
   * `L` will be replaced with the appropriate Luhn checksum.
   *
   * @param string The credit card format pattern. Defaults to `'6453-####-####-####-###L'`.
   * @param symbol The symbol to replace with a digit. Defaults to `'#'`.
   *
   * @example
   * // Test1
   * faker.helpers.replaceCreditCardSymbols() // '6453-4876-8626-8995-3771'
   * faker.helpers.replaceCreditCardSymbols('1234-[4-9]-##!!-L') // '1234-9-5298-2'
   * @example
   * // Test2
   * faker.helpers.replaceCreditCardSymbols() // '6453-4876-8626-8995-3771'
   * faker.helpers.replaceCreditCardSymbols('1234-[4-9]-##!!-L') // '1234-9-5298-2'
   *
   * @since 5.0.0
   */
  replaceCreditCardSymbols(
````

File: docs/.vitepress/components/api-docs/refreshable-code.vue [hunk 10]
```diff
@@ -14,7 +14,9 @@ const {
 }>();
 
 const code = useTemplateRef('code');
-const codeBlock = computed(() => code.value?.querySelector('div pre code'));
+// Scope to the wrapper so that multiple `<pre><code>` blocks (one per tab in a
+// VitePress code-group) are all covered. `.line` selectors below rely on this.
+const codeBlock = computed(() => code.value);
```
ST-DDT: This works, but now `codeBlock .value === code.value` making the variable kind of redundant.

ST-DDT (approved):

matthewmayer (approved):

xDivisionByZerox (approved):
Look cool. Thanks for that. :+1:

matthewmayer:
We should follow up to see if there are any other example blocks which could benefit from being split like this.

PR #3818, opened 2026-04-19
feat: add first standalone module functions

### LINKED ISSUE #1118: Export individual API modules on their own

### LINKED ISSUE #2667: Proposal: Design for standalone module functions

ST-DDT:
Partially implements #2667
Split from #3748 to decrease change size
Forces #3155 into `Won't fix`

- #2667
- #3748
- #3155

---

This PR introduces the first standalone module functions for improved feature (p)review.

- [getDefaultRefDate](https://deploy-preview-3818.fakerjs.dev/api/utils.html#getdefaultrefdate)
- [setDefaultRefDate](https://deploy-preview-3818.fakerjs.dev/api/utils.html#setdefaultrefdate)


ST-DDT:
Note to self: apidocs example refresh does not work ("past" workaround temporarily needed here as well)

src/utils/registry.ts
```diff
@@ -0,0 +1,10 @@
+import { getDefaultRefDate } from './get-default-ref-date';
+import { setDefaultRefDate } from './set-default-ref-date';
+
+/**
+ * Registry module containing all standalone utility functions for the Faker library.
+ */
+export const utilsModule = {
+  getDefaultRefDate,
+  setDefaultRefDate,
+};
```
Shinigami92: **suggestion:** does it make sense here to use `as const` for better strict typing?

```suggestion
} as const;
```

test/scripts/apidocs/page.spec.ts
```diff
@@ -112,4 +112,19 @@ describe('toRefreshFunction', () => {
     // then
     expect(result).toMatchSnapshot();
   });
+
+  it('should handle SMF calls', async () => {
```
Shinigami92: **polish:** I'm still conflicted if we should bring the abbreviation into the source code  
Contributors might be confused by the meaning, as we have zero documentation about that term

src/utils/registry.ts
ST-DDT: I'll have to check. I think I didnt add const to allow adding more custom functions in user code.
Maybe I have to add a `satisfies` though.

test/scripts/apidocs/page.spec.ts
ST-DDT: I'll change that. Some of this is due to us not having a public name for this yet.

ST-DDT:
Everything fixed.

src/registry.types.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Global Registry for the Faker library, containing all module registries.
+ */
+export type FakerRegistry = Record<string, ModuleRegistry>;
+
+/**
+ * Per module registry containing the module's standalone functions.
+ */
+export type ModuleRegistry = Record<
+  string,
+  (fakerCore: FakerCore, ...args: never[]) => unknown
```
Shinigami92: **issue:** what do you want to express with `...args: never[]`?
This should either be dropped, or maybe you want e.g. `...args: readonly unknown[]` 🤔

scripts/apidocs/output/page.ts
```diff
@@ -103,20 +112,29 @@ async function writePageMarkdown(page: RawApiDocsPage): Promise<void> {
...
+  page: RawApiDocsPage,
+  registryHints: Record<string, string> = {}
+): Promise<void> {
   const { camelTitle, methods } = page;
   const pageData: Record<string, ApiDocsMethod> = Object.fromEntries(
     await Promise.all(
       methods.map(async (method) => [method.name, await toMethodData(method)])
     )
   );
+  const priorizedRegistryHints = {
```
Shinigami92: **typo:** prioritizedRegistryHints

src/registry.types.ts
```diff
@@ -0,0 +1,14 @@
+import type { FakerCore } from '.';
```
Shinigami92: **suggestion:** 

```suggestion
import type { FakerCore } from './core';
```

src/registry.types.ts
ST-DDT: I used unknown first, but then I couldn't assign and also got a lint warning, that I should use never instead.
Removing it worked as well.

src/utils/get-default-ref-date.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ * fakerCore.randomizer.seed(1234);
+ *
+ * // Default behavior
+ * // setDefaultRefDate(fakerCore);
+ * past(fakerCore); // Changes based on the current date/time
+ *
+ * // Use a static ref date
+ * setDefaultRefDate(fakerCore, new Date('2020-01-01'));
+ * past(fakerCore); // Reproducible '2019-07-03T08:27:58.118Z'
```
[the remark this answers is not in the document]
ST-DDT: We temporarily create the `past` const in the docs script itself util the date module is transformed.

src/utils/set-default-ref-date.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @example
+ * fakerCore.randomizer.seed(1234);
+ *
+ * // Default behavior
+ * // setDefaultRefDate(fakerCore);
+ * past(fakerCore); // Changes based on the current date/time
+ *
+ * // Use a static ref date
+ * setDefaultRefDate(fakerCore, new Date('2020-01-01'));
+ * past(fakerCore); // Reproducible '2019-07-03T08:27:58.118Z'
```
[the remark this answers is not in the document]
ST-DDT: We temporarily create the `past` const in the docs script itself util the date module is transformed.

src/registry.types.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Global Registry for the Faker library, containing all module registries.
+ */
+export type FakerRegistry = Record<string, ModuleRegistry>;
+
+/**
+ * Per module registry containing the module's standalone functions.
+ */
+export type ModuleRegistry = Record<string, (fakerCore: FakerCore) => unknown>;
```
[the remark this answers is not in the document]
ST-DDT: Since this is only required as safe-guard during init/set this isn't actually needed.
Using unknown here causes issues. Alternatively I could use any, but that would require an eslint disable comment.

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
```diff
@@ -139,25 +164,90 @@ ${examples}`;
...
                 if (!examples.includes('import ')) {
-                  const imports = [
+                  const rootImports =
                     // collect the imports for the various locales e.g. fakerDE_CH
-                    ...new Set(examples.match(/(?<!\.)faker[^.-]*(?=\.)/g)),
-                  ];
+                    new Set(examples.match(/(?<!\.)faker\w*(?=\.)/g));
+
+                  const functionImports = [
+                    ...new Set(examples.match(/\b(\w+)\((faker\.)?fakerCore/g)),
```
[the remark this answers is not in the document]
ST-DDT: `new Set(null/undefined)` behave the same as `new Set()`

src/index.ts
```diff
@@ -87,6 +87,7 @@ export type { SystemModule } from './modules/system';
 export type { VehicleModule } from './modules/vehicle';
 export type { WordModule } from './modules/word';
 export type { Randomizer } from './randomizer';
+export { fakerRegistry } from './registry';
 export { SimpleFaker, simpleFaker } from './simple-faker';
 export { mergeLocales } from './utils/merge-locales';
```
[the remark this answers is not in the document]
ST-DDT: Should we decide now or later where/how to export the standalone functions?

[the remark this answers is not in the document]
Shinigami92: `utils/ref-date` sounds good to me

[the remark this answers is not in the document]
ST-DDT: Do you refer to entrypoints here or something else?

So:

`@faker-js/faker/utils/ref-date` and thus `@faker-js/faker/modules/person/first-name`?

or better something shorter?

`@faker-js/faker/utils` and thus `@faker-js/faker/modules/person`?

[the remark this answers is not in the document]
Shinigami92: welp, I read the suggestion wrongly. nevermind.
I thought `getDefaultRefDate, setDefaultRefDate` is previously in `merge-locales` and should be moved to `ref-date`.

so... no, we can export them later

Shinigami92 (dismissed):

src/utils/get-default-ref-date.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * // Use a ref date that changes every time it is used
+ * let clock = new Date("2020-01-01").getTime();
+ * setDefaultRefDate(fakerCore, () => {
+ *   clock += 1000; // +1s
+ *   return new Date(clock);
+ * });
+ *
+ * getDefaultRefDate(fakerCore) // 2020-01-01T00:00:01Z
+ * getDefaultRefDate(fakerCore) // 2020-01-01T00:00:02Z
```
xDivisionByZerox, line 33: Would this be perfect to apply our new code group for multiple examples?

_Same for setDefaultRefDate obviously._

```suggestion
 * @example
 * // Default behavior
 * fakerCore.randomizer.seed(1234);
 *
 * // setDefaultRefDate(fakerCore);
 * past(fakerCore); // Changes based on the current date/time
 * @example
 * // Static ref date
 * fakerCore.randomizer.seed(1234);
 *
 * setDefaultRefDate(fakerCore, new Date('2020-01-01'));
 * past(fakerCore); // Reproducible '2019-07-03T08:27:58.118Z'
 * @example
 * // Ref date that changes every time it is used
 * fakerCore.randomizer.seed(1234);
 *
 * let clock = new Date("2020-01-01").getTime();
 * setDefaultRefDate(fakerCore, () => {
 *   clock += 1000; // +1s
 *   return new Date(clock);
 * });
 *
 * getDefaultRefDate(fakerCore) // 2020-01-01T00:00:01Z
 * getDefaultRefDate(fakerCore) // 2020-01-01T00:00:02Z
```

Image

src/utils/get-default-ref-date.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @since 10.5.0
+ */
+export function getDefaultRefDate(fakerCore: FakerCore): Date {
+  return fakerCore.config.defaultRefDate?.() ?? DEFAULT_REF_DATE_SOURCE();
+}
```
xDivisionByZerox, line 39: From a design perspective: Why are these core utility functions standalone instead of being located inside the faker core instance? 
Have have a potential answer to that question but would like to have this verified.

src/simple-faker.ts
```diff
@@ -7,8 +7,8 @@ import { SimpleHelpersModule } from './modules/helpers';
 import { SimpleLocationModule } from './modules/location';
 import { NumberModule } from './modules/number';
 import { StringModule } from './modules/string';
-
-export const DEFAULT_REF_DATE_SOURCE: () => Date = () => new Date();
+import { getDefaultRefDate } from './utils/get-default-ref-date';
+import { setDefaultRefDate as utilsSetDefaultRefDate } from './utils/set-default-ref-date';
```
xDivisionByZerox: You renamed the setter but not the getter on the import. Is there a reason for doing so?

src/utils/get-default-ref-date.ts
```diff
@@ -0,0 +1,39 @@
+import type { FakerCore } from '../core';
+
+export const DEFAULT_REF_DATE_SOURCE: () => Date = () => new Date();
```
xDivisionByZerox: I know that the previous DEFAULT_REF_DATE_SOURCE was also exported from the faker file, but why exactly are we doing this? `getDefaultRefDate` is supposed to hide the implementation detail (and existence) of the source of the default ref date, no?
I also didn't find any references of this variable being used somewhere else.

src/registry.ts
xDivisionByZerox, line 1: From my understanding, this file will contain a map of all smf in the future. Wouldn't that go directly against the goal of improving tree shakeability? 
The moment the registry reference is used anywhere, it should be impossible to remove any function from the final build artifact. Am I missing something?

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
```diff
@@ -85,6 +105,7 @@ describe('check docs completeness', () => {
 });
 
 describe('verify JSDoc tags', () => {
+  //#region Per Module
```
xDivisionByZerox, line 108: What do you achieve with these regions? The following describe block can just a easy be collapsed if desired, no? Same goes for most of the other regions.

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
```diff
@@ -139,25 +164,90 @@ ${examples}`;
...
               });
+              //#endregion
 
+              //#region Description
               it('verify description', () => {
                 assertDescription(signature.description);
               });
+              //#endregion
 
+              //#region Example
```
xDivisionByZerox: This should propably align with the the actual jsdoc tag:


```suggestion
              //#region @example
```

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
```diff
@@ -186,7 +276,9 @@ ${examples}`;
                   ).resolves.toBeDefined();
                 }
               );
+              //#endregion
 
+              //#region Deprecated
```
xDivisionByZerox: This should propably align with the the actual jsdoc tag:


```suggestion
              //#region @deprecated
```

scripts/apidocs/output/page.ts
```diff
@@ -28,18 +28,27 @@ editLink: false
  * @param pages The pages to write.
  */
 export async function writePages(pages: RawApiDocsPage[]): Promise<void> {
-  await Promise.all(pages.map(writePage));
+  const registryHints: Record<string, string> = Object.fromEntries(
+    pages.flatMap((page) =>
+      page.methods.map((method) => [method.name, page.camelTitle])
+    )
+  );
```
xDivisionByZerox, line 38: Just so I understand correct: This is a reverse mapping, where each function name is mapped to it's owning module, correct?

src/utils/get-default-ref-date.ts
ST-DDT: Might have been a leftover from previous iterations/tests.
Fixed.

src/utils/get-default-ref-date.ts
ST-DDT: Changed

src/simple-faker.ts
ST-DDT: For disambiguity:

`faker.defaultRefDate` -> `getDefaultRefDate`
`faler.setDefaultRefDate` -> `setDefaultRefDate` ⚠️ -> `utilsSetDefaultRefDate`

Or should I keep `defaultRefDate` also for the standalone function?

src/utils/get-default-ref-date.ts
ST-DDT: There exists multiple reasons, some became obsolete in various iterations.

1. For constistency with potential future (user) additions to the config methods: e.g. `get/setRegExpDefaultCharacterSet`. This is one of the reasons why I have dedicated an entire config object to this. We would still need to publish it for user augmentation, but that can be done later.
2. All methods interacting with faker(core) instances become standalone module functions. There isn't really a huge difference between `faker.defaultRefDate` and `faker.date.now()`.
3. Similarity to the now obsolete `resolveLocaleData` SMF
4. FakerCore is not a class and thus does not have easily created/attached methods. Originally, I intended FakerCore to be a loosely typed interface that you could throw whatever into, but this has proven hard to use, so I opted for the slightly more ridgid `createFakerCore`. It is intentionally not a class , so that you don't have to extend it for custom functions.
5. Having (custom) functions on the core itself would prevent it from being forked/derived for nested calls. Not sure if that is still a goal. #1499

src/registry.ts
ST-DDT: > The moment the registry reference is used anywhere, it should be impossible to remove any function from the final build artifact.

Sadly yes.

This is needed for fake (and our docs examples). 

Thats why I put the following comment on there:

> You normally don't need to access this registry unless you want to call `fake()` or other custom code needing to lookup standalone functions.

If you want full faker function support with fake you need to pass it to standalone fake.
The `Faker` fake will pass it to fake by default.

I plan to replace all internal fake calls with resolver functions in a future PR, to avoid most of the negative impact of fake/the registry.

- #3176 

Note: "Currently", none of the internal standalone fake calls use the registry, instead they manually reference all needed methods and build a subset of available methods.

- https://github.com/faker-js/faker/blob/344b3e5408a4963099b3ee8583cbf3c8aef58fb1/src/modules/food/description.ts#L16-L19

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
ST-DDT: IMO it is easier to navigate using the side panel:

grafik

I can remove them if you don't like them.

scripts/apidocs/output/page.ts
ST-DDT: Yes, some of the examples reference other methods. In order to avoid having to add all imports, I search for the method by name and then add needed "import"/registry-lookup.

Alternatives:

 - `@exampleImports`
 - `@example import { foo } from './bar'`
 - `#use example-import:`
 - `const hiddenMeta = ...`

for each example/method. The later would be invisible for the user, but I would read/parse them via AST.

If you have a preference, or other suggestion. Please let me know.

test/scripts/apidocs/verify-jsdoc-tags.spec.ts
Shinigami92: Feel free to stay with them as is, I also like this kind of sidemap navigation

xDivisionByZerox: I honestly didn't know this was a think, because I do not have the side panel active in my IDE (only scrollbar). I can see that some might find this helpful, so I'm not blocking this.

scripts/apidocs/output/page.ts
xDivisionByZerox: No alternatives required. I really just tried to understand what was going on in the referenced lines and wanted to make sure I understood correctly. :+1:

src/simple-faker.ts
xDivisionByZerox: > Or should I keep defaultRefDate also for the standalone function?

No, for the standalone functions get/set prefixes make sense. 
I completely understand the requirement for "set*" to receive an additional prefix, I would have expected for `getDefaultRefDate` to also receive the `util` prefix additionally.  That way it is clear that `utilsGetDefaultRefDate` and `utilsSetDefaultRefDate` belong together. I would argue that that mental chain is somewhat broken right now.

This is more of a **nitpick**, so if you disagree with my take feel free to resolve this comment on your own.

ST-DDT: Sorry, I missunderstood your "renamed" reference here, since technically both have been renamed.
Just in different stages.
Prefix added for consistency during imports.

src/registry.ts
xDivisionByZerox: **Considerations:**

1. Fake is not implemented in this PR => the registry is not required as well - at least now
2. I'd would say that fake does not require a SMF alternative - the module is giving the fake function the restrictions it required. The STM variant could be unintuitive for users and a potential food gun (regarding tree shakability) on top of that.

**Suggestion:**

Remove the registry in this PR. At this point in time, I'm not considering a fake SMF in the final API. If the community requires it to exist, I'm open to reopen the discussion regarding the matter.

For the SMF that require `fake`, we can use a temporary faker instance: `new Faker(fakerCore).helpers.fake(pattern)`. In the next major version, fake patterns inside the locale data will be replace by resolve functions (`(fakerCore) => firstName(fakerCore)` instead of `'{{firstName}}'`). 
See #3176.

- #3176

ST-DDT: Fixed/registry removed

src/utils/get-default-ref-date.ts
ST-DDT: @xDivisionByZerox and I talked abot this today. No changes needed here.

Talking points:

- put methods directly on core or config: 
  - Would make custom user config options harder to implement/inconsistent in implementation.
  - Would require a dedicated createConfig call.
- config options name conflicts between different extensions:
  - Idea: Use symbols instead of string keys
    - Pros: Unique/No name conflicts
    - Negative: Harder to pre-initialize the config, before passing it to `createFakerCore`
      To allow this, we would need to make the symbol public and then it isn't any more safe then a string.
- Expose FakerConfig for type augmentation later alongside howto docs

plus probably some more, that I forgot about

ST-DDT:
Ready for review.

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3819, opened 2026-04-19
feat: introduce resolveLocaleData utility function

ST-DDT:
Extracted from #3189 for future discussion
Blocked by #3818 

- #3189
- #3818

---

Adds the `resolveLocaleData` utility function.

Intended for users of our library to check whether their custom data are present.

````ts
@example
resolveLocaleData(fakerCore, 'food', 'fruit'); // ['apple', 'apricot', ...]
arrayElement(fakerCore, resolveLocaleData(fakerCore, 'food', 'fruit')); // 'cherry'
````

PR #3820, opened 2026-04-19
refactor(core): expose core.locale as LocaleProxy

ST-DDT:
~Alternative to #3189 and #3819~

- ~#3189~
- ~#3819~

---

- Expose the locale data as `LocaleProxy` in `FakerCore`.
- New `raw` attribute to get access to the raw LocaleDefinitions (to avoid issues with locale data existence checks)

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1,5 +1,7 @@
 import type { FakerConfig } from './config';
 import type { LocaleDefinition } from './definitions';
+import type { LocaleProxy } from './internal/locale-proxy';
+import { createLocaleProxy } from './internal/locale-proxy';
 import type { Randomizer } from './randomizer';
 import { mergeLocales } from './utils/merge-locales';
 import { generateMersenne53Randomizer } from './utils/mersenne';
@@ -13,7 +15,7 @@ export interface FakerCore {
    *
    * Always present, but it might be empty if the locale data is not available.
    */
-  readonly locale: LocaleDefinition;
+  readonly locale: LocaleProxy;
 
   /**
    * The randomizer used to generate random values.
@@ -32,7 +34,7 @@ export interface FakerOptions {
    *
    * @default {}
    */
-  locale?: LocaleDefinition | LocaleDefinition[];
+  locale?: LocaleProxy | LocaleDefinition | LocaleDefinition[];
   /**
    * The randomizer used to generate random values.
    *
@@ -95,7 +97,9 @@ export function createFakerCore(options: FakerOptions = {}): FakerCore {
   }
 
   return {
-    locale: Array.isArray(locale) ? mergeLocales(locale) : locale,
+    locale: createLocaleProxy(
+      Array.isArray(locale) ? mergeLocales(locale) : locale
+    ),
     randomizer,
     config,
   };
--- a/src/internal/locale-proxy.ts
+++ b/src/internal/locale-proxy.ts
@@ -1,12 +1,21 @@
 import type { LocaleDefinition } from '../definitions';
 import { FakerError } from '../errors/faker-error';
 
+const LOCALE_PROXY_TAG = Symbol('FakerLocaleProxy');
+
 /**
  * A proxy for LocaleDefinition that marks all properties as required and throws an error when an entry is accessed that is not defined.
  */
-export type LocaleProxy = Readonly<{
-  [key in keyof LocaleDefinition]-?: LocaleProxyCategory<LocaleDefinition[key]>;
-}>;
+export type LocaleProxy = Readonly<
+  {
+    [key in keyof LocaleDefinition]-?: LocaleProxyCategory<
+      LocaleDefinition[key]
+    >;
+  } & {
+    raw: LocaleDefinition;
+    [LOCALE_PROXY_TAG]: true;
+  }
+>;
 
 type LocaleProxyCategory<T> = Readonly<{
   [key in keyof T]-?: LocaleProxyEntry<T[key]>;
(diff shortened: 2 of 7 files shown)


ST-DDT:
I marked this as P2, because it needs to be decided on prior to the next release.

ST-DDT:
Benefits of this over #3189:

- No duplidated keys
- Implicit locale data checks

Benefits of #3189 over this:

- Does not use Proxy
- Slightly less RAM usage

Shinigami92 (dismissed):

src/internal/locale-proxy.ts
```diff
@@ -18,13 +27,31 @@ const throwReadOnlyError: () => never = () => {
...
+ * Checks if the given value is a LocaleProxy.
+ *
+ * @param value The value to check.
+ *
+ * @returns  True if the value is a LocaleProxy, false otherwise.
+ */
+function isLocaleProxy(value: unknown): value is LocaleProxy {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (value as any)?.[LOCALE_PROXY_TAG] === true;
+}
```
xDivisionByZerox, line 51: This will throw a runtime error if `value` of value `null | undefined`. Since you allow theses values by defining `value` as `unknown` I'd say its fair to assume that this can happen. You might want to only allow `object` values to be passed as an argument in this function.

src/internal/locale-proxy.ts
```diff
@@ -33,25 +60,29 @@ export function createLocaleProxy(locale: LocaleDefinition): LocaleProxy {
...
+        if (categoryName === LOCALE_PROXY_TAG) {
+          return true;
+        }
+
         return target[categoryName];
       }
 
-      if (categoryName in proxies) {
-        return proxies[categoryName];
+      if (categoryName === 'nodeType') {
```
xDivisionByZerox, line 84: For my understanding: What is this "nodeType" accessor? I know that it was in the previous implementation, but I don't remember what it was used for. Sorry.

src/internal/locale-proxy.ts
```diff
@@ -1,12 +1,21 @@
...
-export type LocaleProxy = Readonly<{
-  [key in keyof LocaleDefinition]-?: LocaleProxyCategory<LocaleDefinition[key]>;
-}>;
+export type LocaleProxy = Readonly<
+  {
+    [key in keyof LocaleDefinition]-?: LocaleProxyCategory<
+      LocaleDefinition[key]
+    >;
+  } & {
+    raw: LocaleDefinition;
```
xDivisionByZerox, line 19: `LocaleDefinition` can be any key from what I can see. Setting "raw" here might overwrite a real custom module for our users. Instead you could do one of two things:

1. These "hidden" properties are usually prefixed with some additional special character. Think "__proto__" for example. So you could rename it to "__raw" to make it less unlikely to overwrite an actual property.
2. You already created a special symbol for the proxy identity (which is very clever btw).Why not do this for "raw" as well?

xDivisionByZerox, line 19: If I see this correctly, this property will publicly exposed. In this case it should get some documentation.

src/core.ts
```diff
@@ -32,7 +34,7 @@ export interface FakerOptions {
    *
    * @default {}
    */
-  locale?: LocaleDefinition | LocaleDefinition[];
+  locale?: LocaleProxy | LocaleDefinition | LocaleDefinition[];
```
xDivisionByZerox, line 37: In which case will a faker core ever NOT hold a LocaleProxy?

src/internal/locale-proxy.ts
ST-DDT: Fixed

src/internal/locale-proxy.ts
ST-DDT: 1. I don't consider it hidden, because it is the recommended way to check if a field exists without throwing an error.
   I consider `raw` to be an unlikely module name. any number of _ would reduce that even further, but it would also discourage the use for the intended purpose.

2) IMO Symbol access is messy as you need to specifically import it.

**Note:** This is not a hard "no" to a rename, but there is intention in the current design/name (`rawDefinitions`).
There are plenty of alternative ways to design this instead.

ST-DDT: Done

src/core.ts
ST-DDT: These are `FakerOptions` not `FakerCore`.
FakerOptions are used to create a new FakerCore and locale data may be absent/empty for that particular core.

```txt
FakerOptions -> FakerCore -> Faker
locale?      -> locale!   -> definitions!
```

src/internal/locale-proxy.ts
ST-DDT: I think this is the reason:

- https://github.com/faker-js/faker/pull/2092

Shinigami92 (approved):

src/internal/locale-proxy.ts
xDivisionByZerox: Thank you. seems reasonable enought.

src/core.ts
xDivisionByZerox: You are totally right. I extended the diff between line 5 and 13 instead of 19 and 32, which made it seem like this was the definition of the FakerCore Interface. My bad.

src/internal/locale-proxy.ts
xDivisionByZerox: Your arguments are reasonable enough for me to take back these suggestion. Especially the symbol access, which seemed clever in my mind but would not be very good DX as you pointed out.

xDivisionByZerox (approved):

PR #3822, opened 2026-04-21
fix(locale): remove citation brackets from snake names

AustralianRaven:
## What

Remove citation brackets (e.g. `[4]`, `[5]`) from two snake names in the English locale data:

- `'Brown spotted pitviper[4]'` → `'Brown spotted pitviper'`
- `'Pointed-scaled pit viper[5]'` → `'Pointed-scaled pit viper'`

These brackets were reference footnote markers copied from the Wikipedia source and should not appear in the generated data.
See [this Wikipedia page](https://en.wikipedia.org/wiki/Protobothrops_mucrosquamatus) for an example:
image

## Why

Introduced in ~~[03b965a](https://github.com/faker-js/faker/commit/03b965abfde57be617a1ae7ba715604e149a803d)~~ [1914b96](https://github.com/faker-js/faker/commit/1914b965c22b61a82e0aaf97e560899cd81bfee9), the two entries contained Wikipedia-style citation numbers (`[4]` and `[5]`) that leaked into the locale file. This causes `faker.animal.snake()` to occasionally return strings like `"Brown spotted pitviper[4]"`, which is not a valid snake name.

## Related issues

No issue — this is a data fix per the contributing guide.


matthewmayer:
Good catch. Was actually introduced in 1914b965

Seems to be the only such references in the locales folder

image

matthewmayer:
Please run `pnpm run preflight` to regenerate the test snapshots.

AustralianRaven:
> Please run `pnpm run preflight` to regenerate the test snapshots.

Whoops, sorry. This has now been done.

ST-DDT (approved):

ST-DDT:
Nicely spottet. Thanks for your contribution ❤️

xDivisionByZerox (approved):
Thanks for changing the introduction commit on your own. I was about to say something regarding the matter xD
Good catch regardless.

matthewmayer (approved):

PR #3827, opened 2026-04-28
fix(date): incorrect date boundary in past and future

ST-DDT:
Fix incorrect date boundaries in `past()` and `future()` causing random pipeline failures.

The issue is mainly caused by leap years as those have more days than usual, that didn't get accounted for.

### DIFF
--- a/src/modules/date/index.ts
+++ b/src/modules/date/index.ts
@@ -6,15 +6,6 @@ import { assertLocaleData } from '../../internal/locale-proxy';
 import { SimpleModuleBase } from '../../internal/module-base';
 
 /**
- * Small helper function to convert a number of years to an amount of milliseconds.
- *
- * @param years The number of years to convert to milliseconds.
- */
-function yearsToMs(years: number): number {
-  return years * 365 * 24 * 3600 * 1000;
-}
-
-/**
  * Module to generate dates (without methods requiring localized data).
  */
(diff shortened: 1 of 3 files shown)


xDivisionByZerox (commented):
Would there be a chance to add test that force the correct leap year handling, to prevent regressions regarding the matter in the future?

ST-DDT:
There are already tests for that - thats why we have the random failures.

Should I add tests with a custom randomizer to force max/min values?

Shinigami92:
I agree with @xDivisionByZerox, we currently see snapshot changes here, but not the actual covered fix.
I would be in favor if we could add a test with statically passed refDate so mimic the exact behavior in a before broken case.

ST-DDT:
So basically:

https://github.com/faker-js/faker/blob/f286dfdb046be8387b3104bb9d6503ad5112fef8/test/modules/date.spec.ts#L194-L211

Just with a static ref date instead of the implicit `now`?

Shinigami92:
you should make a copy of that test-case, not update it

xDivisionByZerox:
Yes, just like the test you referenced, but with a static ref date. You might need to create a separate faker instance for that case, because the ref date would be persisted and have side effects on the other tests of you use the same instance.

ST-DDT:
Done

Shinigami92 (approved):
nice, thx 👍

PR #2787, opened 2024-04-05
refactor(locale)!: remove deprecated locale aliases

Shinigami92:

PR #3828, opened 2026-04-28
refactor(locale): rename ne to en_NP

### LINKED ISSUE #3655: ne locale is not in Nepali
### Pre-Checks

### Describe the bug

`ne` locale is not actually in Nepali (written in `Deva` Devanagari script). For example the cities and names at https://github.com/faker-js/faker/blob/next/src/locales/ne/location/city_name.ts or https://github.com/faker-js/faker/blob/next/src/locales/ne/person/first_name.ts are English transliterations

I feel the current data should be moved to `en_NP` leaving `ne` free for a proper Devanagari locale.

Locale was originally added here
https://github.com/faker-js/faker/commit/0048b9f6cfd69378ca3ba2f588444e7bb8095d23

### Minimal reproduction code

_No response_

### Additional Context

_No response_

### Environment Info

```shell
na
```

- [x] CJS
- [ ] ESM

### Used Package Manager

npm

ST-DDT:
Fixes #3655

- #3655
- Based on #2787

---

- Rename ne to en_NP.
- Deprecate ne faker.


Shinigami92:
This is a major breaking change, is it? I think this needs to wait for v11.
Similar as we did here: https://v9.fakerjs.dev/guide/upgrading#locale-aliases

ST-DDT:
> This is a major breaking change, is it?

Not yet. This only deprecates the old locale.

````ts
import { faker } from '@faker-js/faker/locale/ne'"
````

still works.

The actually breaking change is the removal of the old alias.

Or should we rename it in v11 without prior deprecation?

Shinigami92:
Maybe just plan it in for v11, that way we have one commit/noise less (also for the customer facing changelog)

matthewmayer:
Couldn't we fix this in a non-semver-major way by:
1. Add the current ne data to en_NP
2. Replace Latin names for things like city names in the existing `ne` locale with Nepali names in Devangari script

Issue #1443, opened 2022-10-13
Extract type definitions that might be used across modules


Shinigami92:
We have some types like `Casing` that can be reused be multiple module functions.
Also we might want to extract a `Range` or an interface that defines a length or range that can then be used by all functions that have a length or min-max signature.

Decision to discuss: We might want to change all length-only signatures to also accept a min-max option.

xDivisionByZerox:
@ST-DDT Does the rewrite of your docs generation support the resolving of JSDoc information of custom types? Or can we at least make it work? 
If I remember correctly this was the previously the  problem why we were not able to use custom types. At least for objects interfaces/types that contained properties with JSDocs.

ST-DDT:
IIRC the current implementation shows the name of the type aka Casing, but it should already resolve it as shadow type but it isnt shown yet. I have somewhat forgotten about the docs rewrite to change the visualization of the docs. If I have some free time, I'll look into it this week.

ST-DDT:
Just found this issue. What still needs to be done here?

xDivisionByZerox:
We have Union Types that define the same thing in multipel places. For example:

- https://github.com/faker-js/faker/blob/f286dfdb046be8387b3104bb9d6503ad5112fef8/src/modules/string/index.ts#L8
- https://github.com/faker-js/faker/blob/f286dfdb046be8387b3104bb9d6503ad5112fef8/src/modules/color/index.ts#L42

Another idea was to provide a unified interface for ranges that we have in a lot of places (`number | { min?: number, max?: number }`). That way we would not repeat the documentation in multiple places and had a more unified API.
So far we have not done it, because the autogenerate documentation would not be able to handle it (if I remember correctly).

ST-DDT:
If you tell me, what you need/expect, maybe I can look into it (apidocs part).

ST-DDT:
@Shinigami92 Anything left to do here?

PR #3829, opened 2026-04-29
refactor(types): extract shared Casing type

Shinigami92:
Part of #1443.

Extracts the duplicated `Casing` type into a new public `src/utils/types.ts` so it can be imported once and reused across modules — and so consumers can pre-construct typed arguments before passing them to faker methods.

## Changes

- New `src/utils/types.ts` exporting `Casing = 'upper' | 'lower' | 'mixed'` with JSDoc.
- `src/modules/string/index.ts` and `src/modules/color/index.ts` no longer declare their own `Casing`; both now `import type { Casing } from '../../utils/types'`.
- `src/modules/finance/bitcoin.ts` re-points its `Casing` import from `../string` to `../../utils/types`.
- `src/index.ts` re-exports `Casing` from `./types` instead of from `./modules/color`.
- Apidocs snapshot updated to reflect the unified union order (`'upper' | 'lower' | 'mixed'`), which matches the existing `switch (casing)` order in both modules.

## Notes

- No runtime behavior change — type-only refactor.
- Public API surface is unchanged: `Casing` is still exported from the package root.
- Follow-up extractions (`NumberRange`, etc.) will be opened as separate PRs.


Shinigami92:
@ST-DDT please check this draft PR in terms of api-docs. you can takeover this PR and commit directly into it from here on if required. Otherwise I can just undraft.

ST-DDT:
AFAICT this has no impact on apidocs.

ST-DDT (dismissed):

src/types.ts
xDivisionByZerox, line 1: We already have a file `src/internal/types.ts`. For consistency I'd propse to move this file under the `src/utils` directory. (internals = not exported in public API; utils = exported in public API)

xDivisionByZerox (commented):
> > @ST-DDT please check this draft PR in terms of api-docs. you can takeover this PR and commit directly into it from here on if required. Otherwise I can just undraft.
> AFAICT this has no impact on apidocs.

From my understanding, this part was never the part that made trouble with the API generation. It was more the question of how we structure our publicly exposed types. The second part in the referenced issue (create types for shared configurations) was more the problem. But the api documentation generation went through so many iterations to harden its functionality that it might also work out of the box now.

src/types.ts
Shinigami92: I did it exactly for that on purpose, Casing should be a exported non internal type

image

xDivisionByZerox: But you have the types file directly in the `src` directory right now. Currently, only files directly related to the faker class. Parts of the code that can be used when using the faker library, but are not directly related to the faker class, are usually located inside the `src/utils` directory. That's why I was proposing to move it there as well.

| Currently | Proposed |
|---------- | ---------- |
| `src/types.ts` | `src/utils/types.ts` |

Shinigami92: Oh good catch, it was supposed to go into utils

Shinigami92:
Well, that AI was not very helpful :(
But need to do when have time

xDivisionByZerox (approved):

ST-DDT (approved):

PR #3838, opened 2026-05-02
feat: introduce NumberRange type

Shinigami92:
_the following PR description is claude generated_

## Summary

Introduces a shared, publicly exported `NumberRange` interface and uses it to replace the ~40 inline `{ min: number; max: number }` shapes scattered across the module method signatures.

```ts
export interface NumberRange {
  /** The minimum value (inclusive). */
  min: number;
  /** The maximum value (inclusive). */
  max: number;
}
```

```ts
import { faker, type NumberRange } from '@faker-js/faker';

const count: NumberRange = { min: 1, max: 10 };

faker.lorem.words(count);
faker.helpers.arrayElements(['a', 'b', 'c'], count);
faker.string.alpha({ length: count });
```

Refs #1443.

## Motivation

- **Shorter, more readable signatures.** Method signatures that previously inlined a multi-line `number | { min; max }` object now read as `number | NumberRange`.
- **A reusable, importable type for consumers.** Users can declare a single `NumberRange` value and pass it to any range-accepting method, instead of re-declaring the shape inline or reverse-engineering it via `Parameters<...>`.
- **Deduplication.** One canonical definition replaces ~40 copies of the same shape.

## Design decision (Option 3)

The discussion weighed several approaches (a single shared type vs. per-concept named types such as `WordCountRange`, hybrid inline/shared shapes, and intersection types). We are going with **Option 3: a single shared `NumberRange`**.

Rationale:

- For the consumer, `NumberRange` and a hypothetical `WordCountRange` are structurally identical - extra per-concept types mostly add mental overhead and import churn.
- Starting with one type is the forward-compatible choice: a more specific type can be **added later non-breakingly** if users ask for it, whereas removing redundant types after the fact would be a breaking change.

## Tradeoffs

- The per-field JSDoc that previously lived on each inline `min`/`max` (e.g. *"The minimum number of words to generate."*) is replaced by the generic `NumberRange` field docs (*"The minimum value (inclusive)."*). The method-level `@param x.min` / `@param x.max` lines are unaffected.
- **Known follow-up:** the API-docs generator no longer renders the `min`/`max` sub-parameters (and their per-field defaults) for parameters typed as `NumberRange`. This needs to be addressed separately in the apidoc generation - see the discussion thread.

### DIFF
--- a/scripts/apidocs/processing/type.ts
+++ b/scripts/apidocs/processing/type.ts
@@ -91,6 +91,8 @@ export function getTypeText(
           : newSimpleType('string');
 
         return newUnionType([displayType, baseType]);
+      } else if (name === 'NumberRange') {
+        return newSimpleType('{ min: number; max: number }');
       }
 
       const typeParameters = typeArguments.map((t) => getTypeText(t, options));
--- a/src/index.ts
+++ b/src/index.ts
@@ -92,4 +92,4 @@ export {
   generateMersenne32Randomizer,
   generateMersenne53Randomizer,
 } from './utils/mersenne';
-export type { Casing } from './utils/types';
+export type { Casing, NumberRange } from './utils/types';
--- a/src/modules/airline/module.ts
+++ b/src/modules/airline/module.ts
@@ -1,4 +1,5 @@
 import { ModuleBase } from '../../internal/module-base';
+import type { NumberRange } from '../../utils/types';
 
 export enum Aircraft {
   Narrowbody = 'narrowbody',
@@ -240,18 +241,7 @@ export class AirlineModule extends ModuleBase {
        *
        * @default { min: 1, max: 4 }
        */
-      length?:
-        | number
-        | {
-            /**
-             * The minimum number of digits to generate.
-             */
-            min: number;
-            /**
-             * The maximum number of digits to generate.
-             */
-            max: number;
-          };
+      length?: number | NumberRange;
       /**
        * Whether to pad the flight number up to 4 digits with leading zeros.
        *
--- a/src/modules/date/module.ts
+++ b/src/modules/date/module.ts
@@ -4,6 +4,7 @@ import type { Faker } from '../../faker';
 import { toDate } from '../../internal/date';
 import { assertLocaleData } from '../../internal/locale-proxy';
 import { SimpleModuleBase } from '../../internal/module-base';
+import type { NumberRange } from '../../utils/types';
 
 /**
  * Module to generate dates (without methods requiring localized data).
@@ -70,22 +71,7 @@ export class SimpleDateModule extends SimpleModuleBase {
        *
        * @default 1
        */
-      years?:
-        | number
-        | {
-            /**
-             * The minimum amount of years the date should be in the past.
-             *
-             * @default 0
-             */
-            min: number;
-            /**
-             * The maximum amount of years the date should be in the past.
-             *
-             * @default 1
-             */
-            max: number;
-          };
+      years?: number | NumberRange;
       /**
        * The date to use as reference point for the newly generated date.
        *
@@ -149,22 +135,7 @@ export class SimpleDateModule extends SimpleModuleBase {
        *
        * @default 1
        */
-      years?:
-        | number
-        | {
-            /**
-             * The minimum amount of years the date should be in the future.
-             *
-             * @default 0
-             */
-            min: number;
-            /**
-             * The maximum amount of years the date should be in the future.
-             *
-             * @default 1
-             */
-            max: number;
-          };
+      years?: number | NumberRange;
       /**
        * The date to use as reference point for the newly generated date.
        *
@@ -280,18 +251,7 @@ export class SimpleDateModule extends SimpleModuleBase {
      *
      * @default 3
      */
-    count?:
-      | number
-      | {
-          /**
-           * The minimum number of dates to generate.
-           */
-          min: number;
-          /**
-           * The maximum number of dates to generate.
-           */
-          max: number;
-        };
+    count?: number | NumberRange;
   }): Date[] {
     const { from, to, count = 3 } = options;
     return this.faker.helpers
--- a/src/modules/finance/bitcoin.ts
+++ b/src/modules/finance/bitcoin.ts
@@ -1,4 +1,4 @@
-import type { Casing } from '../../utils/types';
+import type { Casing, NumberRange } from '../../utils/types';
 
 /**
  * The bitcoin address families.
@@ -30,7 +30,7 @@ export type BitcoinNetworkType = `${BitcoinNetwork}`;
 
 type BitcoinAddressOptions = {
   prefix: Record<BitcoinNetworkType, string>;
-  length: { min: number; max: number };
+  length: NumberRange;
   casing: Casing;
   exclude: string;
 };
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1,6 +1,7 @@
 import type { Faker, SimpleFaker } from '../..';
 import { FakerError } from '../../errors/faker-error';
 import { SimpleModuleBase } from '../../internal/module-base';
+import type { NumberRange } from '../../utils/types';
 import { fakeEval } from './eval';
 import { luhnCheckValue } from './luhn-check';
 
@@ -1029,18 +1030,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    */
   arrayElements<const T>(
     array: ReadonlyArray<T>,
-    count?:
-      | number
-      | {
-          /**
-           * The minimum number of elements to pick.
-           */
-          min: number;
-          /**
-           * The maximum number of elements to pick.
-           */
-          max: number;
-        }
+    count?: number | NumberRange
   ): T[] {
     if (array.length === 0) {
       return [];
@@ -1117,20 +1107,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    *
    * @since 8.0.0
    */
-  rangeToNumber(
-    numberOrRange:
-      | number
-      | {
-          /**
-           * The minimum value for the range.
-           */
-          min: number;
-          /**
-           * The maximum value for the range.
-           */
-          max: number;
-        }
-  ): number {
+  rangeToNumber(numberOrRange: number | NumberRange): number {
     if (typeof numberOrRange === 'number') {
       return numberOrRange;
     }
@@ -1163,18 +1140,7 @@ export class SimpleHelpersModule extends SimpleModuleBase {
        *
        * @default 3
        */
-      count?:
-        | number
-        | {
-            /**
-             * The minimum value for the range.
-             */
-            min: number;
-            /**
-             * The maximum value for the range.
-             */
-            max: number;
-          };
+      count?: number | NumberRange;
     } = {}
   ): TResult[] {
     const count = this.rangeToNumber(options.count ?? 3);
--- a/src/modules/lorem/module.ts
+++ b/src/modules/lorem/module.ts
@@ -1,4 +1,5 @@
 import { ModuleBase } from '../../internal/module-base';
+import type { NumberRange } from '../../utils/types';
 import { filterWordListByLength } from '../word/filter-word-list-by-length';
 
 /**
@@ -47,18 +48,7 @@ export class LoremModule extends ModuleBase {
            *
            * @default 1
            */
-          length?:
-            | number
-            | {
-                /**
-                 * The minimum length of the word.
-                 */
-                min: number;
-                /**
-                 * The maximum length of the word.
-                 */
-                max: number;
-              };
+          length?: number | NumberRange;
           /**
            * The strategy to apply when no words with a matching length are found.
            *
@@ -101,20 +91,7 @@ export class LoremModule extends ModuleBase {
    *
    * @since 2.0.1
    */
-  words(
-    wordCount:
-      | number
-      | {
-          /**
-           * The minimum number of words to generate.
-           */
-          min: number;
-          /**
-           * The maximum number of words to generate.
-           */
-          max: number;
-        } = 3
-  ): string {
+  words(wordCount: number | NumberRange = 3): string {
     return this.faker.helpers
       .multiple(() => this.word(), { count: wordCount })
       .join(' ');
@@ -134,20 +111,7 @@ export class LoremModule extends ModuleBase {
    *
    * @since 2.0.1
    */
-  sentence(
-    wordCount:
-      | number
-      | {
-          /**
-           * The minimum number of words to generate.
-           */
-          min: number;
-          /**
-           * The maximum number of words to generate.
-           */
-          max: number;
-        } = { min: 3, max: 10 }
-  ): string {
+  sentence(wordCount: number | NumberRange = { min: 3, max: 10 }): string {
     const sentence = this.words(wordCount);
     return `${sentence.charAt(0).toUpperCase() + sentence.substring(1)}.`;
   }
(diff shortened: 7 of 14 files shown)


Shinigami92:
@ST-DDT @xDivisionByZerox — looking for input on a tradeoff this PR introduces.

## Context

This PR replaces ~40 inline `{ min: number; max: number }` shapes with a shared `NumberRange` interface. As a side effect, the per-site type-level JSDoc on `min`/`max` (e.g. *"The minimum number of words to generate."*) is gone. The method-level `@param wordCount.min` / `@param wordCount.max` lines are still there, but the JSDoc that surfaces in VSCode hover on the field itself is now the generic `NumberRange` description (*"The minimum value (inclusive)."*).

## Options

### 1. Per-concept named ranges
Define small types per use-case, each carrying its own JSDoc:
```ts
export interface WordCount extends NumberRange {
  /** The minimum number of words to generate. */
  min: number;
  /** The maximum number of words to generate. */
  max: number;
}
```
- ✅ Tooltips recovered, semantically meaningful named types for consumers.
- ❌ ~10 new exported types; API surface grows.

### 2. Hybrid: inline shape for surface methods, `NumberRange` internally
Keep `NumberRange` for plumbing (`helpers.rangeToNumber`, `helpers.multiple`) and revert to inline `{ /** doc */ min; /** doc */ max }` on user-facing methods (`lorem.words`, `string.alpha`, …).
- ✅ Pragmatic — best tooltips where users actually hover.
- ❌ Inconsistent style across the codebase.

### 3. Accept it, lean on `@param`
Ship as-is. TS LSP can surface `@param x.min` JSDoc on destructured fields in some cases, and the rendered docs site has the info via `@param`.
- ✅ Smallest diff, simplest.
- ❌ Tooltip regression in VSCode for the inline `{ min, max }` form.

### 4. Re-declare `min` / `max` inline on each method, typed as `NumberRange`-compatible
```ts
words(
  wordCount: number | (NumberRange & {
    /** The minimum number of words to generate. */
    min: number;
    /** The maximum number of words to generate. */
    max: number;
  }) = 3
): string;
```
- ✅ `NumberRange` stays publicly importable for consumers, no per-concept type explosion, per-method tooltips preserved.
- ❌ The methods themselves don't gain dedup — each still re-declares `min`/`max` with JSDoc; `NumberRange` is more of a public-API contract than an implementation deduplication.

---

## My take

I lean toward **option 4**: it keeps `NumberRange` publicly available for consumers (which is part of the point per #1443), doesn't bloat the API with a specific type for every method, and we still get the tooltip docs at hover sites.

However it _extends_ the tooltip, not overrides it:


**Option 1** is also acceptable as a fallback — it gives consumers semantically meaningful types — but I'd rather have one shared `NumberRange` than ten specific ones.

It does not have the extended JSDoc, but only the specific (which might be a pro-argument for this option)


I don't like **option 2**: the inconsistency between "surface" and "internal" methods feels like a smell to me.

I don't like **option 3**: the per-site tooltip docs were an intentional team decision earlier, and I don't want to silently regress that.

Thoughts?

ST-DDT:
IMO The most important benefit of this change is the shortened method signature => simplified readability.
Most users understand what a range is.

There are a few questions that we need to search answers for first:

A) Do our users understand that a range is {min, max} and NOT [min, max] with minimal effort?

  - If they need to see that it is an object at first glance, then this PR is obsolete.
  - If not and hovering the type is acceptable, then we should decide B+C

B) Do our users have a need for types or do they inline the arguments anyway?
   Are our users likely to use our type, instead of declaring their own?
   Do we do it for the readability of our signature?

C) Is the context closeness of numberRange.min/max important to our users?
   aka do they need to know that it is a `WordRange` vs a `number` Range?

---

1. IMO this kind of defeats the purpose of this change.
   If needed, our users can create a type for all our parameters themselves (either manually or via `Parameters`)
   The parameter itself is called wordCount with type `NumberRange`, IMO it is clear what it does.
   If we go that route anyway, then we should keep the `Range` as part of the name.

2. Maybe. IMO this is the smallest and least interruptive. However the benefit is likely negible.
   Depends on (C): If it is important, then this maybe the max extend this PR can go.
   Our users are still able to use our conviencience type. Even if there is no direct link between the two.
   Though I doubt, that anyone will (know of and) use the NumberRange conviencience type then and instead users might define their own copy of the type or access it directly via `Parameters[0]`.
   Also we should keep compatibility with that type in mind when implementing more features.

3. Maybe.
   Depends on (C): If it is not important, then this is what we should do.
   This would be my recommended way.

4. No. This negatively impacts readability of the signature. Especially the multiple unions and joins increase mental complexity quite a bit.

5. "Keep as is in next"
   This would be my next recommendation in line.

Order of preference: 3, 5, 2, 1, 4 
(3, 5, 2 are close together in preference, and they mostly depend on the goal of/requirements behind this change)

---

Diff-Preview

The current apidoc generation process does not display. Min and max params anymore:

| [**Old**](https://fakerjs.dev/api/helpers.html#arrayelements) | [**New**](https://deploy-preview-3838.fakerjs.dev/api/helpers.html#arrayelements) |
| --- | --- |
| grafik | grafik |
| grafik | grafik |

**Note:** when the range currently has a default for either value, then we no longer display it in the jsdocs after this PR is merged.

---

> I don't like option 3: the per-site tooltip docs were an intentional team decision earlier, and I don't want to silently regress that.

Isn't the whole point of this issue/PR to rethink this? If not, what is it?

---

- Ref: https://github.com/faker-js/faker/issues/1443

Shinigami92:
@ST-DDT thanks for the feedback 👍 

> Isn't the whole point of this issue/PR to rethink this? If not, what is it?

The most valueable feature for consumers would be to use

```ts
import { type NumberRange, faker} from '@faker-js/faker';

const options: NumberRange = {
//             ^^^^^^^^^^^
  min: 1,
  max: 10
};

//                   vvvvvvv passing typed options
faker.module.method1(options)
faker.module.method2(options)
faker.module.method3(options)
```

---

So thinking about this, it would indeed be the best option to use Option 1, even when this is a bit verbose on our side.

Shinigami92:
The workaround right now is something very complex like:
```ts

type NumberRange = Exclude<Parameters<faker.module.method>[0], number>
```

_pseudo code, not tested_

ST-DDT:
> So thinking about this, it would indeed be the best option to use Option 1, even when this is a bit verbose on our side.

Why do you think this distinction is important for our users?

````ts
const options: NumberRange = {
  min: 1,
  max: 10
};

faker.module.method1(options)
faker.module.method2(options)
faker.module.method3(options)
````

vs

````ts
const options: WordCountRange = {
  min: 1,
  max: 10
};

faker.module.method1(options)
faker.module.method2(options)
faker.module.method3(options)
````

IMO both types are identical. And adding more imports/different types just increases the mental costs to remember how they are defined.

Shinigami92:
You are right on the first look, but when in an IDE the second one (`WordCountRange`) will give the min and max a specific tooltip, instead of a generic.

ST-DDT:
IMO in order to move this forward, I would suggest starting with one type and if users complain, add another one.
Since `NumberRange` === `WordCountRange`, adding a new type would be breaking, while removing the redundant one would be.

Shinigami92:
## Decision: locking in Option 3 (single shared `NumberRange`)

After the discussion, we'll proceed with **Option 3** — a single shared, publicly exported `NumberRange` interface — which is what this PR currently implements.

Reasoning (per @ST-DDT):
- `NumberRange` and a per-concept type like `WordCountRange` are structurally identical for consumers; extra named types mostly add import/mental overhead.
- Starting with one type is forward-compatible: a more specific type can be **added later non-breakingly** if users actually ask for it, whereas removing redundant types afterwards would be breaking.

This trades away the per-field `min`/`max` JSDoc at hover sites in favor of the generic `NumberRange` field docs, which we accept.

One remaining follow-up before this is ready: the apidoc generator no longer renders the `min`/`max` sub-parameters (and their per-field defaults) for `NumberRange`-typed params. I'll handle that separately.

_this comment was fully posted by claude, I told claude to write into a markdown file, but it ignored it and directly posted it 🙁_
image

ST-DDT (dismissed):

xDivisionByZerox (dismissed):
Currently the NumberRange type is resolved as a type name in the docs, instead of "exploding" the type declaration into more primitive types:



If we want t change that and instead show the underlying "object"-type we would need to resolve it via it's symbol (type => symbol => declaration [type=alias] => type node => text). We do not seem to have tests for parameters so I wasn't able to test this out of the box. 

---

_Edit: Should be somewhere here where we would need to change that:_
https://github.com/faker-js/faker/blob/fd33b723c1155bcaeb501a1eb1031b8b2917ba64/scripts/apidocs/processing/type.ts#L145-L156

ST-DDT:
<details>
<summary>Here the docs patch (click to expand)</summary>

````diff
diff --git a/scripts/apidocs/processing/type.ts b/scripts/apidocs/processing/type.ts
index 25ec30c96..2c0310e20 100644
--- a/scripts/apidocs/processing/type.ts
+++ b/scripts/apidocs/processing/type.ts
@@ -91,6 +91,8 @@ export function getTypeText(
           : newSimpleType('string');
 
         return newUnionType([displayType, baseType]);
+      } else if (name === 'NumberRange') {
+        return newSimpleType('{ min: number; max: number }');
       }
 
       const typeParameters = typeArguments.map((t) => getTypeText(t, options));
diff --git a/test/scripts/apidocs/__snapshots__/method.spec.ts.snap b/test/scripts/apidocs/__snapshots__/method.spec.ts.snap
index 7be83ab71..e85e9c37c 100644
--- a/test/scripts/apidocs/__snapshots__/method.spec.ts.snap
+++ b/test/scripts/apidocs/__snapshots__/method.spec.ts.snap
@@ -20,6 +20,7 @@ exports[`method > expected and actual methods are equal 1`] = `
   "methodWithThrows",
   "multiParamMethod",
   "noParamMethod",
+  "numberRangeParamMethod",
   "optionalStringParamMethod",
   "optionsInlineParamMethodWithDefaults",
   "optionsInterfaceParamMethodWithDefaults",
@@ -342,7 +343,7 @@ exports[`method > processMethodLike(literalUnionParamMethod) 1`] = `
         },
         {
           "default": undefined,
-          "description": "Value \`'a'\` or \`'b'\` or an array thereof.",
+          "description": "\`'a'\` or \`'b'\` or an array thereof.",
           "name": "mixed",
           "type": {
             "text": "'a' | 'b' | string | Array<'a' | 'b' | string>",
@@ -395,7 +396,7 @@ exports[`method > processMethodLike(literalUnionParamMethod) 1`] = `
         },
         {
           "default": undefined,
-          "description": "Value \`'a'\` or \`'b'\` or an array thereof.",
+          "description": "\`'a'\` or \`'b'\` or an array thereof.",
           "name": "namedMixed",
           "type": {
             "text": "'a' | 'b' | string | Array<AB | string>",
@@ -983,6 +984,111 @@ exports[`method > processMethodLike(noParamMethod) 1`] = `
 }
 `;
 
+exports[`method > processMethodLike(numberRangeParamMethod) 1`] = `
+{
+  "name": "numberRangeParamMethod",
+  "signatures": [
+    {
+      "deprecated": undefined,
+      "description": "Test with NumberRange.",
+      "examples": [],
+      "experimental": undefined,
+      "parameters": [
+        {
+          "default": undefined,
+          "description": "\`{min: 1, max: 10}\`.",
+          "name": "value",
+          "type": {
+            "text": "{ min: number; max: number }",
+            "type": "simple",
+          },
+        },
+        {
+          "default": undefined,
+          "description": "Array of \`{min: 1, max: 10}\`.",
+          "name": "array",
+          "type": {
+            "text": "Array<{ min: number; max: number }>",
+            "type": "generic",
+            "typeParameters": [
+              {
+                "text": "{ min: number; max: number }",
+                "type": "simple",
+              },
+            ],
+          },
+        },
+        {
+          "default": undefined,
+          "description": "The options parameter.",
+          "name": "options",
+          "type": {
+            "text": "{ ... }",
+            "type": "simple",
+          },
+        },
+        {
+          "default": undefined,
+          "description": "The count parameter.",
+          "name": "options.count",
+          "type": {
+            "text": "{ min: number; max: number }",
+            "type": "simple",
+          },
+        },
+        {
+          "default": undefined,
+          "description": "\`{min: 1, max: 10}\` or an array thereof.",
+          "name": "mixed",
+          "type": {
+            "text": "{ min: number; max: number } | Array<{ min: number; max: number }>",
+            "type": "union",
+            "types": [
+              {
+                "text": "{ min: number; max: number }",
+                "type": "simple",
+              },
+              {
+                "text": "Array<{ min: number; max: number }>",
+                "type": "generic",
+                "typeParameters": [
+                  {
+                    "text": "{ min: number; max: number }",
+                    "type": "simple",
+                  },
+                ],
+              },
+            ],
+          },
+        },
+      ],
+      "remarks": [],
+      "returns": {
+        "text": "string",
+        "type": "simple",
+      },
+      "seeAlsos": [],
+      "signature": "function numberRangeParamMethod(
+    value: NumberRange,
+    array: ReadonlyArray<NumberRange>,
+    options: {
+      /** The count parameter. */
+      count: NumberRange;
+    },
+    mixed: NumberRange | ReadonlyArray<NumberRange>
+  ): string;",
+      "since": "1.0.0",
+      "throws": [],
+    },
+  ],
+  "source": {
+    "column": -1,
+    "filePath": "test/scripts/apidocs/method.example.ts",
+    "line": -1,
+  },
+}
+`;
+
 exports[`method > processMethodLike(optionalStringParamMethod) 1`] = `
 {
   "name": "optionalStringParamMethod",
diff --git a/test/scripts/apidocs/method.example.ts b/test/scripts/apidocs/method.example.ts
index 1651129e9..3aeebb86e 100644
--- a/test/scripts/apidocs/method.example.ts
+++ b/test/scripts/apidocs/method.example.ts
@@ -1,4 +1,4 @@
-import type { Casing, ColorFormat } from '../../../src';
+import type { Casing, ColorFormat, NumberRange } from '../../../src';
 import { FakerError } from '../../../src/errors/faker-error';
 import type { LiteralUnion } from '../../../src/internal/types';
 import type { AlphaNumericChar } from '../../../src/modules/string';
@@ -182,8 +182,8 @@ export class SignatureTest {
    * @param namedValue `'a'` or `'b'`.
    * @param array Array of `'a'` or `'b'`.
    * @param namedArray Array of `'a'` or `'b'`.
-   * @param mixed Value `'a'` or `'b'` or an array thereof.
-   * @param namedMixed Value `'a'` or `'b'` or an array thereof.
+   * @param mixed `'a'` or `'b'` or an array thereof.
+   * @param namedMixed `'a'` or `'b'` or an array thereof.
    *
    * @since 1.0.0
    */
@@ -205,6 +205,34 @@ export class SignatureTest {
     );
   }
 
+  /**
+   * Test with NumberRange.
+   *
+   * @param value `{min: 1, max: 10}`.
+   * @param array Array of `{min: 1, max: 10}`.
+   * @param options The options parameter.
+   * @param options.count `{min: 1, max: 10}`.
+   * @param mixed `{min: 1, max: 10}` or an array thereof.
+   *
+   * @since 1.0.0
+   */
+  numberRangeParamMethod(
+    value: NumberRange,
+    array: ReadonlyArray<NumberRange>,
+    options: {
+      /** The count parameter. */
+      count: NumberRange;
+    },
+    mixed: NumberRange | ReadonlyArray<NumberRange>
+  ): string {
+    return JSON.stringify({
+      value,
+      array,
+      options,
+      mixed,
+    });
+  }
+
   /**
    * Test with a Record parameter.
    *
````

</details>

Feel free to apply now or in a separate PR.

Shinigami92:
@ST-DDT example: https://deploy-preview-3838.fakerjs.dev/api/airline.html#flightnumber
well done, good work 👍

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3849, opened 2026-05-14
feat(locale): add Welsh (cy) county definitions

IestynGage:
Based on the counties/county boroughs that are in Wales.

This does have a implication on the states list.  One could say Wales' first level of administraion is counties/county boroughs. But if you did `.county()` and it returned something else it would be confusing. I suppose for states it could be England, Nothern Ireland, Scotland and Wales (but in welsh of course)?


matthewmayer:
That's similar to what en_GB does. I can see some logic in having "Cymru" as the only entry for state() in cy.

I'd also suggest implementing the new postal address pattern to give a sensible full address that uses county() in Welsh.

ST-DDT (approved):

Shinigami92 (approved):

IestynGage:
> That's similar to what en_GB does. I can see some logic in having "Cymru" as the only entry for state() in cy.
> 
> I'd also suggest implementing the new postal address pattern to give a sensible full address that uses county() in Welsh.

Sure, happy to do that next :)

ST-DDT:

PR #3850, opened 2026-05-15
feat(locale): add Welsh (cy) postal address definitions

IestynGage:
Follows https://github.com/faker-js/faker/pull/3849.

Postal address definitions as ask by @matthewmayer in https://github.com/faker-js/faker/pull/3849#issuecomment-4457250292. Uses the new county, based on the `EN_GB` address.

PR #3851, opened 2026-05-19
feat(locale): add Welsh (cy) postcode definitions

IestynGage:
Based on the `en_GB` postcode definitions with a prefix as recommened by @matthewmayer in https://github.com/faker-js/faker/pull/3850#issuecomment-4484704496

### DIFF
--- a/src/locales/cy/location/postcode.ts
+++ b/src/locales/cy/location/postcode.ts
@@ -1,0 +1,6 @@
+const suffixes = ['## ##?', '# #??'];
+const prefixes = ['CF', 'CH', 'HR', 'LD', 'LL', 'NP', 'SA', 'SY'];
+
+export default prefixes.flatMap((prefix) =>
+  suffixes.map((suffix) => prefix + suffix)
+);
--- a/src/locales/cy/location/index.ts
+++ b/src/locales/cy/location/index.ts
@@ -6,11 +6,13 @@ import type { LocationDefinition } from '../../..';
 import county from './county';
 import direction from './direction';
 import postal_address from './postal_address';
+import postcode from './postcode';
 
 const location: LocationDefinition = {
   county,
   direction,
   postal_address,
+  postcode,
 };
 
 export default location;
--- a/test/__snapshots__/locale-data.spec.ts.snap
+++ b/test/__snapshots__/locale-data.spec.ts.snap
@@ -8,7 +8,7 @@ exports[`locale-data > should only have known characters 1`] = `
   "base": " ()+,-./:;ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz",
   "bn_BD": " (),-ঁংঅআইঈউএওকখগঘঙচছজঝঞটঠডঢণতথদধনপফবভমযরলশষসহ়ািীুূৃেৈোৌ্ড়য়",
   "cs_CZ": " #()+-.ABCDEFGHIJKLMNOPRSTUVWXZabcdefghijklmnopqrstuvwxyzÁÍÚáéíóöúüýČčĎďěňŘřŠšťůűŽž",
-  "cy": " ,-ABCDEFGHILMNOPRSTWYabcdefghijklmnoprstuwyôŵ",
+  "cy": " #,-?ABCDEFGHILMNOPRSTWYabcdefghijklmnoprstuwyôŵ",
   "da": " !"#()+,-./ABCDEFGHIJKLMNOPQRSTUVWYZabcdefghijklmnopqrstuvwxyzÅÆØãåæçéíø",
   "de": " #&'()+,-.ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyzÄÖÜßàãäéíöúü",
   "de_AT": " #&()+,-.ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyzÄÖÜßãäéíöúü",


src/locales/cy/location/postcode.ts
```diff
@@ -0,0 +1,4 @@
+export default [
+  '{{location.postcode_prefix}}# #??',
+  '{{location.postcode_prefix}}## #??',
+];
```
IestynGage: The `en_GB` is just `export default ['??# #??', '??## #??'];` I suppose you can also add a similiar prefix to `en_GB` as well. But the postcodes for England, NI, Scotland, Wales, crown dependecies and other special postcodes.

ST-DDT (changes requested):
Please make sure to test your changes and verify that they return the expected results ⚠️

````ts
fakerCY.location.zipCode() // "{{location.postcode_prefix}}76 2MM"
````

Currently, the zipcode implementation does not support fake patterns (for default patterns).

This is planned for #3233 / #3176 

Until then, you have to cross-product them yourself.

File: src/locales/cy/location/postcode.ts (hunk 1 above)
Comment (ST-DDT) [reply src/locales/cy/location/postcode.ts]: See also #2390 

- #2390

**Comment by IestynGage** [comment]
> Please make sure to test your changes and verify that they return the expected results ⚠️
> 
> ```ts
> fakerCY.location.zipCode() // "{{location.postcode_prefix}}76 2MM"
> ```
> 
> Currently, the zipcode implementation does not support fake patterns (for default patterns).
> 
> This is planned for #3233 / #3176
> 
> Until then, you have to cross-product them yourself.

Oof that's embarrising 🤦 Just a heads up it will be away for a while before I can come back to this :)

@ST-DDT what do you mean by 'cross-product this myself?' Is it that:
* Or do you mean I have to do it by "hand", as in change the content to `'CF# #??', 'CF## #??',` etc and remove `postcode_prefix`?
* update `/src/definitions/location.ts`?

I pushed a WIP commit of doing the latter, it seems to work. WIP as I would like to update some tests check that the other locales work as well and add documentation.

Also whats the easiest way to run faker locally? via testing or starting `pnpm run docs:dev` and then using the browser console?

**Comment by matthewmayer** [comment]
Yes, do it by hand like

`['CF# #??', 'CF## #??', 'CH# #??', 'CH## #??'...]`

**Comment by matthewmayer** [comment]
In the long run we would like to move to fake patterns and drop use of replaceSymbols completely, but that will potentially be a breaking change and impacts many locales, so the "by hand" trick is the easiest for now.

**Comment by ST-DDT** [comment]
> Also whats the easiest way to run faker locally? via testing or starting `pnpm run docs:dev` and then using the browser console?

There are 4 ways. In descending order of my personal preference.

- pnpm tsx my-test.ts (easy to repeat and debug, looking into complex values may require JSON.stringify, no need to switch windows)
- pnpm run docs:dev and then via browser console (easy to look into array/complex values , you may need to run it/generate:apidocs repeatedly)
- clone and use the playground or create your own (you may need to build/install repeatedly)
- write a (temporary) test, (vitest may add noise to your logs or run more tests than you intended)

@faker-js/members Should we document that somewhere or prepare a commamd for that like `pnpm run testxecute/texecute`? Currently only docs:dev ist documented.

**Comment by Shinigami92** [comment]
IMO this is a really niche case. However you could store scripts in `_local` folder, because this folder will be ignored by the `clean` script.
https://github.com/faker-js/faker/blob/8b4e2b13b6cec4acf89debdde5c80ed313f02694/package.json#L6

**Comment by matthewmayer** [comment]
> > Also whats the easiest way to run faker locally? via testing or starting `pnpm run docs:dev` and then using the browser console?
> 
> 
> 
> There are 4 ways. In descending order of my personal preference.
> 
> 
> 
> - pnpm tsx my-test.ts (easy to repeat and debug, looking into complex values may require JSON.stringify, no need to switch windows)
> 
> - pnpm run docs:dev and then via browser console (easy to look into array/complex values , you may need to run it/generate:apidocs repeatedly)
> 
> - clone and use the playground or create your own (you may need to build/install repeatedly)
> 
> - write a (temporary) test, (vitest may add noise to your logs or run more tests than you intended)
> 
> 
> 
> @faker-js/members Should we document that somewhere or prepare a commamd for that like `pnpm run testxecute/texecute`? Currently only docs:dev ist documented.
> 
> 
> 
> 

Personally I do

```
pnpm run build
node
```

Now you are in node REPL so can do like:

```js
const {fakerCY}=require(".")
fakerCY.location.zipCode()
```

ST-DDT:
I created a separate issue to discuss dev verification tooling and documentation:

- #3854

ST-DDT:
>  Just a heads up it will be away for a while before I can come back to this :)

Thanks for the heads up.

> @ST-DDT what do you mean by 'cross-product this myself?' Is it that:
> 
>     * Or do you mean I have to do it by "hand", as in change the content to `'CF# #??', 'CF## #??',` etc and remove `postcode_prefix`?
> 
>     * update `/src/definitions/location.ts`?
> 
> 
> I pushed a WIP commit of doing the latter, it seems to work. WIP as I would like to update some tests check that the other locales work as well and add documentation.

Sorry, for not being more clear.

I didn't mean add `fake` support.
I meant:

````ts
const suffixes = ['`## ##?', ...]
const prefixes = ['CY', ...]

export prefixes.flatMap(prefix => suffixes.map(suffix => prefix+suffix))
````

Either like this or only the result of it.

IestynGage:
Thanks for the feedback & knowledge ❤️  After your first comment I was using `pnpm run docs:dev`

🤦 I can't belive it didn't occur to me to just combine two arrays together with `map`/`flatmap`. It's much clearer now :)

ST-DDT (approved):

ST-DDT:
I'm glad I was able to help.

Shinigami92 (approved):

PR #3857, opened 2026-05-23
feat: transform core modules to standalone module functions

### LINKED ISSUE #1118: Export individual API modules on their own

### LINKED ISSUE #2667: Proposal: Design for standalone module functions

ST-DDT:
Split from #3748 

- #3748

---

Transforms the core modules to standalone modular functions.

These core modules are interdependent and thus cannot be transformed on their own.

- DatatypeModule
- HelpersModule
- NumberModule
- StringModule

Most of the conversion steps are done automatically, but some changes have been done manually.

Please note that I renamed some util files to `_`, to distinguish files that export an API SMF from those that don't.

### 1. Compare automated steps vs PR

- checkout PR
- hard reset to before the automated scripts
- run scripts
- compare with commit

### 2. Verify no diff in modules to #3748 and look at the manual transform commits

````sh
#checkout PR
git diff origin/feat/standalone-module-functions -- src/modules/<ModuleName>
#No Diff expected
````

Check the individual module transform commits:

- Datatype (No manual changes)
- [`Helpers` (#3748)](https://github.com/faker-js/faker/pull/3748/changes/397222c459249637c46ac0a117a7843cbffe5eff)
- [`Number` (#3748)](https://github.com/faker-js/faker/pull/3748/changes/6401cb17000ed1b5e5ba0ab8dec239ef3de7f70e)
- [`String` (#3748)](https://github.com/faker-js/faker/pull/3748/changes/d7c2fa1f1bd37dac42ba21719db18cc6e104555c)
- [`Generate Module Tree` (#3748)](https://github.com/faker-js/faker/pull/3748/changes/6ddd95bdc5c4271533051ea4429af510068095cb#diff-3f6b4fc424ba2cedcbf5f612ba256d248f0c05f46314eed1dfa81590bcd429fd) (Technically automated, but might be interesting anyway)
- [`Final Fixes and Cleanup` (#3748)](https://github.com/faker-js/faker/pull/3748/changes/5daaa57270365319665f9bfcc10d7be4d3ec67ae)

---

Please do not merge any PRs changing anything in the modules after this PR until all modules are transformed.

### CODE REVIEW

**Comment by Shinigami92** [comment]
discussable arguable opinionated feedback



this feels quite messy
the problem I have with it, is that you mentally need to know the entire faker structure in your head instead of exploring it by e.g. namespace or one import

this is right now just internal code, I would never like to ship something like this for customers
customers should be able to use e.g. `import { datatypeBoolean, helpersArrayElement } from '@faker-js/faker'` and so on

maybe we can have something like this also internally (and still don't 🦆-up 🌲-🫨)

**Comment by ST-DDT** [comment]
Just for clarification:

Would you like for us to use qualified method names from the start?
E.g. export function helpersArrayElement
Instead of the rename during import (if the name collides).
Or something else?

This is kind of relevant for the later public API: Will we only export from root or also export via the module paths e.g. `@faker-js/faker/module/helpers`? Or will we adapt our own naming scheme such as `fakeArrayElement` similar to `useRouter`?

**Comment by Shinigami92** [comment]
> Just for clarification:
> 
> Would you like for us to use qualified method names from the start? E.g. export function helpersArrayElement Instead of the rename during import (if the name collides). Or something else?

Partly that was my thought regarding yes, but I did not fully thought that through yet.
My greater problem is more that we e.g. could use internally imports like:

```ts
// instead of
import { arrayElement as helpersArrayElement } from './array-element';
import { arrayElements as helpersArrayElements } from './array-elements';
import { enumValue as helpersEnumValue } from './enum-value';

// something like
import { arrayElement as helpersArrayElement, arrayElements as helpersArrayElements, enumValue as helpersEnumValue } from '../helpers';
```

but `../helpers` is equals `./index` and this is the file we are just in... so I understand that problem

thinking further...

maybe we move the module class into a separate file, so we can import from index... 🦆 might not work either, because we would need to import the module class in index to pass further to top again 🤔 ...

best what we could try to achieve would be something like:

```ts
import * as helpers from './?whatever?';
import * as helpers from './?whatever?';
import * as string from '../string';

export class SimpleHelpersModule extends SimpleModuleBase {
  slugify(string: string = ''): string {
    return helpers.slugify(this.faker.fakerCore, string);
  }

  // string.alphanumeric
```

I know that what I start to propose here might need fundamental movements of files, and this is why I raising this right now (I'm sorry 🥲)

I can also later (not in company-working-hours try to thinker with claude about this, maybe it has some enlightening ideas)

Edit:
I'm raising this right now, because I think this is how we maintainers, contributors and even AIs will need to live and work in the future codebase of Faker. This is such a fundamental shaping. And I did not delivered the solution yet, just uncovered the "problem" (mental load).

ST-DDT:
IMO moving the module to a separate file (and not including it in the index) would work for me too.

Maybe prepare a proposal so that we can talk about it later. Though I dont have time this weekend.

Note: I ammended my previous comment. Not sure whether you have seen it.

Shinigami92:
> IMO moving the module to a separate file (and not including it in the index) would work for me too.

👍 
 
> Maybe prepare a proposal so that we can talk about it later. Though I dont have time this weekend.

🤝 I will also hide myself this weekend deeply into Gothic 1 Remake 😇

> Note: I ammended my previous comment. Not sure whether you have seen it.

THX! I have NOT seen it 👍 

It is kinda a sibling topic, because this is then how we structure the external API
I would be fine maybe even ship and allow both: `{ arrayElements } from @faker-js/faker/module/helpers` AND `{ helpersArrayElements } from @faker-js/faker`

But maybe we should outsource this discussion to a separate place

ST-DDT:
Alternatively, split the files as proposed here, then later refactor the imports if needed. 
Since that is nonbreaking, we can still do so later without issues.

Shinigami92:
I now had a talk with Claude (`Opus4.8[1m]`), and the simple solution is just to move the Module class to a `src/modules/<module>/module.ts`.
In `src/modules/<module>/index.ts` we then just to `import * from './module'`.

In this PR we can then extend it by adding the standalone functions by just doing this:

```ts
// ./array-element.ts
import { ... } from '../<other-module>'
import { ... } from './<other-same-module-standalone-fn>'

export function arrayElement

// ./module.ts
import { arrayElement } from './array-element'

export class SimpleHelpersModule

// ./index.ts
import * from './array-element'
import * from './module'
```

I will spin up a PR for the moving, and we can move on there and decide if this is the wanted solution or if it contains any other problems I have not seen yet.

ST-DDT:
Ready for review again.

---

The PR may be easier to review if you look at the individual commits.

The auto generated commits can be recreated with:

````sh
pnpm tsx scripts/temp-tranform-once.ts 
pnpm tsx scripts/generate-module-tree.ts 
````

File: src/modules/helpers/_eval.ts
Comment (Shinigami92) [inline src/modules/helpers/_eval.ts:1]: **nitpick (non-blocking):** not really relevant for this PR and we can outsource it into a follow-up PR, but I'm not sure if I like the _ prefix for `internal` in our source folder 🤔  
maybe we could just add an internal folder in every module folder? this resolves also potential naming collisions.

File: src/modules/string/nanoid.ts [hunk 1]
```diff
@@ -0,0 +1,62 @@
+import type { FakerCore } from '../../core';
+import { arrayElement } from '../helpers/array-element';
+import { rangeToNumber } from '../helpers/range-to-number';
+import { weightedArrayElement } from '../helpers/weighted-array-element';
```
Shinigami92, line 5: **suggestion:** here is the point where I would like to see that imports are used from their module/index


```suggestion
import { arrayElement, rangeToNumber, weightedArrayElement } from '../helpers';
```

or does this not work? 🤔

src/modules/string/nanoid.ts
```diff
@@ -0,0 +1,62 @@
...
+import { alphanumeric } from '../string/alphanumeric';
```
Shinigami92: **suggestion:** and inside a module I would expect an import like this

```suggestion
import { alphanumeric } from './alphanumeric';
```

Shinigami92 (commented):
I stopped reviewing here and first would like to get the commented points discussed

src/modules/string/nanoid.ts
ST-DDT: That does work, at the risk of including any side effects the index contains.

src/modules/string/nanoid.ts
ST-DDT: Will apply when, we have discussed the other parts.

src/modules/helpers/_eval.ts
ST-DDT: This affects final files:

- 3x color
- 1x date
- 1x finance
- 2x helpers
- 1x internet
- 1x person
- 1x string
- 1x word

You can see them all [here](https://github.com/faker-js/faker/pull/3748/changes) using the `_` file filter on the left side.

ST-DDT: I don't like folders that only contain a single file.
Also note: The internals are sometimes "faker global"-internals and not "module-exclusive".

The main reason they are prefixed with `_` to visually separate them from files that contain SMFs.
We can remove the `_` prefix, no collisions will occur, as I renamed the files that are likely to collide beforehand.

Shinigami92: @xDivisionByZerox what's your opinion?

src/modules/string/nanoid.ts
Shinigami92: I'm not sure or aware that we plan to include side-effects in index 😅 I would even call that an anti-pattern if there would be side-effects in index files

src/modules/helpers/_eval.ts
Shinigami92: just another soft-contra for `_` prefixed: we have e.g. a _local folder in root which is git-ignored, but excluded from `pnpm run clean`  
so we are introducing different meanings for a `_` file/folder

src/modules/string/nanoid.ts
ST-DDT: The faker main index has some kind of side effects, mainly regarding the pre-created faker instances.
I think I mentioned that somewhen/somewhere related to PURE annoations.

Shinigami92: A root index (or often also a main.ts in other context) is a bit differently handled as index files in sub directories which just reexport their folder content upwards.

ST-DDT:
Random fun fact: I just searched and found the issues which laid the foundation for this/these changes.

- #751 (Merged in v7.0.0 / May 2022)
- #1250

src/modules/string/nanoid.ts
ST-DDT: @xDivisionByZerox Which variant do you prefer for foreign module method imports?

- A) Import from the exact file that exports the method
- B) Import from the foreign module index instead

?

ST-DDT: Currently, most of our imports import directly from the actual source file, not from a parent index:

- https://github.com/faker-js/faker/blob/acd5fdaf099bda7cf90c76bf0ae0a33349a54ce7/src/modules/airline/module.ts#L1-L2
- https://github.com/faker-js/faker/blob/acd5fdaf099bda7cf90c76bf0ae0a33349a54ce7/src/modules/helpers/module.ts#L1-L4
- https://github.com/faker-js/faker/blob/acd5fdaf099bda7cf90c76bf0ae0a33349a54ce7/src/modules/internet/module.ts#L2-L5

(excluding the new module imports)

src/modules/string/nanoid.ts
ST-DDT: Fixed.

ST-DDT:
Should the SMFs be annotated with `@experminetal` for now?

src/modules/helpers/_eval.ts
Shinigami92: We are moving away from `_` prefixed folders with https://github.com/faker-js/faker/pull/3998

ST-DDT: I consider the naming conflict resolved then.
The general discussion whether we want to use `_` prefixes for internal module utils remains though.

ST-DDT:
I wait with this until the main bulk of dependency and infra PRs to have passed.
Then I would like to get this merged, as this gathers merge conflicts quite frequently.

Shinigami92:
> I wait with this until the main bulk of dependency and infra PRs to have passed. Then I would like to get this merged, as this gathers merge conflicts quite frequently.

Yes, absolutely. I hope this will already be in a few days. I'm pushing hard right now 😅 
I also want the medical module land in very near future after the dependency and infra PRs have settled.

ST-DDT:
Ready for review.

I reused the existing commits as much as possible.

Shinigami92 (dismissed):
For the sake of moving forward, I'm approving this. But there might be some follow-ups like open/unresolved discussion about e.g. using `internal` folders instead of `_` file prefixes, how imports are handled, and I ignored that `tranform` should be `transform`.

Shinigami92 (dismissed):

src/modules/string/nanoid.ts
xDivisionByZerox: > @xDivisionByZerox Which variant do you prefer for foreign module method imports?
> 
> * A) Import from the exact file that exports the method
> * B) Import from the foreign module index instead

Definitely import from the exact file. Barrel files are known to cause trouble when it comes to side effects (as already pointed out in this discussion; for example creation of faker instances) as well affect tree shaking (think back at the fix in #2790). So making everything explicit should save us some potential trouble in the long run.

An additional problem I see (but didn't verify if it exists) is "circular imports" when using on entry file per module for the use of external modules: 
Hypothetical Example: Some String-Util requires something from Number so it imports from '../number'. Some Number-Util requires something from String so it imports from '../string'. While the dependency resolution should work under these conditions, the graph itself looks messy and not easy to undertsand.

<table>
<tr>
<th> import via barrel file </th>
<th> direct import </th>
</tr>

<tr>
<td>


```mermaid
flowchart TD
    String --> String-Util-A
    String --> String-Util-B

    Number --> Number-Util-A
    Number --> Number-Util-B

    String-Util-A --> Number
    Number-Util-A --> String
```

_Notice the cyclic graph between String => String-Util-A => Number => Number-Util-A => String_

</td>

<td>

```mermaid
flowchart TD
    String --> String-Util-A
    String ---> String-Util-B

    Number --> Number-Util-A
    Number ---> Number-Util-B

    String-Util-A --> Number-Util-B
    Number-Util-A --> String-Util-B
```

_Notice no cycles in the graph_

</td>
</tr>
</table>

_Use "<--->"-Icon to view graph in full screen_

scripts/temp-transform-once.ts
```diff
@@ -0,0 +1,308 @@
...
+  return (
+    value.substring(0, 1).toLowerCase() +
+    value.substring(1) +
+    more.map(toPascalCase).join('')
+  );
+}
+
+function toPascalCase(value: string): string {
+  return value.substring(0, 1).toUpperCase() + value.substring(1);
+}
```
xDivisionByZerox: These already exist in `scripts/shared/character-case.ts`. Not sure if that's an oversight or if you wanted the transform-once script to be as self contained as possible.

test/modules/number.spec.ts
xDivisionByZerox: Quick question because I was confused for a second: Why do we need to spy on the int method at all? Can we simply test `faker.number.romanNumeral({ min: value, max: value})` and expect the same outcome without having to spy on the implementing function?

src/modules/helpers/array-element.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param fakerCore The FakerCore to use.
+ * @param array The array to pick the value from.
+ *
+ * @throws {FakerError} If the given array is empty.
+ *
+ * @example
+ * arrayElement(fakerCore, ['cat', 'dog', 'mouse']) // 'dog'
+ *
+ * @since 6.3.0
+ */
```
xDivisionByZerox, line 21: Question: How do we intend the keep the JSDocs in sync between SMF and Module implementation? I might have not noticed it in the generation script. Regardless, the docs would need to be updated accordingly, so contributors know which docs to change (module, SMF, both?).

xDivisionByZerox (commented):
In between review. I mostly checked all the non generated files first. I'll do a second review round validating the contents of the generated scripts later this day.

src/modules/helpers/array-element.ts
ST-DDT: The generate-module-tree commmand takes care of it to an extend.
The SMF docs are the source of truth, they (and the implememtation) are slightly processed to replace the lines in the module file. I also considered adding auto generated comments to the files, but havent done so yet. If you think, those are immediately important, I can add them.

Since our current automated tooling does not support all required cleanup tasks, it cannot be covered by CI/running the script creates a diff (in the imports).

I do plan to add module/method parameters later to reduce the spread of things to clean up. And I hope that at some point, that through either improved tooling or an improved script that no manual tadks need to be done.
Though it is likely to slow for inclusion in preflight.

test/modules/number.spec.ts
ST-DDT: Yes likely, though I would change it outside of this PR (Either before or after).

scripts/temp-transform-once.ts
ST-DDT: I need to check (later).

ST-DDT: Fixed. At the transform script creation commit the char-case file doesn't exist.
I need to take that into consideration for the next rebase (if any).

src/modules/helpers/enum-value.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * enum Direction { North = 'North', South = 'South'}
+ * enumValue(fakerCore, Direction) // 'South'
+ *
+ * enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }
+ * enumValue(fakerCore, HttpStatus) // 200 (Ok)
+ *
+ * @since 8.0.0
+ */
+export function enumValue<T extends Record<string | number, string | number>>(
```
xDivisionByZerox, line 28: This lost a comment from the original source code. Don't know if we need it, but I wanted to mention it.

https://github.com/faker-js/faker/blob/bbd169385a438ece2ed7e588945bc68838ed144c/src/modules/helpers/module.ts#L1081-L1083

src/modules/helpers/replace-credit-card-symbols.ts
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @example
+ * legacyReplaceSymbolWithNumber(fakerCore) // ''
+ * legacyReplaceSymbolWithNumber(fakerCore, '#####') // '04812'
+ * legacyReplaceSymbolWithNumber(fakerCore, '!####') // '27378'
+ * legacyReplaceSymbolWithNumber(fakerCore, 'Your pin is: !####') // '29841'
+ *
+ * @since 8.4.0
+ */
+export function legacyReplaceSymbolWithNumber(
```
xDivisionByZerox: So far we have only one exported function per file, so far. This file would break that pattern by exporting the additional helper "legacyReplaceSymbolWithNumber" as well.
Why did you choose to not export this one from a standalone "_*" file instead?

src/modules/string/sample.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param length.max The maximum length of the string to generate.
+ *
+ * @example
+ * sample(fakerCore) // 'Zo!.:*e>wR'
+ * sample(fakerCore, 5) // '6Bye8'
+ * sample(fakerCore, { min: 5, max: 10 }) // 'FeKunG'
+ *
+ * @since 8.0.0
+ */
+export function sample(
```
xDivisionByZerox: This method name could lead to confusions in my opinion. While it was located under the ""faker.string." namespace it was given contextual. Now it is not clear that "sample" is returning a string-samples. Additionally, there will be naming conflicts with "faker.word.sample"

src/modules/helpers/array-element.ts
xDivisionByZerox: > I also considered adding auto generated comments to the files, but havent done so yet. If you think, those are immediately important, I can add them.

That was the missing piece for me. Having the comments in the module files would make it more clear who owns the source of truth in these cases. I would appreciate if an "autogenerated comment" could be added to the module files immediately, since the PR would be merged as is into the code base and is open for contributors to interact with

> Since our current automated tooling does not support all required cleanup tasks, it cannot be covered by CI/running the script creates a diff (in the imports).

This would affect the "auto generated" comment or the general generation of module files?

xDivisionByZerox: Ah, follow up question after you clarified who owns the docs truth:
Is it correct to put the same `@since` tag onto the SMF as the module ones? Technically the implementation/parameter-set of the function did exists since the original `@since` tag. But the function as a standalone entry did not. 🤷‍♀️

src/modules/string/sample.ts
ST-DDT: I can manually rename this if you want.

This may get resolved via the public api though.

Import stringSample from faker
Or
Import sample from faker/modules/string
Or similar

src/modules/helpers/replace-credit-card-symbols.ts
ST-DDT: Because of the legacy part in the name and theoretically it is a helper specifically for the credit card number function.
I would like to not export it, but IIRC it is used by the phone(?) module.

src/modules/helpers/enum-value.ts
ST-DDT: These are lost in the automated process.

src/modules/helpers/array-element.ts
ST-DDT: Should I modify them to `@since v11.0 (v8.4.0)`?

xDivisionByZerox: > Should I modify them to `@since v11.0 (v8.4.0)`?

That's basically what I'm asking. But I'm not sure myself. That's why I was asking an open question. Maybe other @faker-js/members want to give their opinion on that decision?

src/modules/helpers/replace-credit-card-symbols.ts
xDivisionByZerox: But would you not be able to export it from a standalone file with an undescore? That way it's marked as internal, no? Or is the real solution here to migrate the using function to use a more simple implementation or even completely change the replacement algorithm?

src/modules/string/sample.ts
xDivisionByZerox: Interesting. @Shinigami92 since you gave your thumbs up on my comment, do you have any preference here?
I'm currently more leaning towards directly renaming this function to "stringSample". Although, I'm not aware if this causes trouble with generation scripts since they add the module name automatically to the functions name, no? Then we would end up with "stringStringSame"? :thinking:

src/modules/helpers/array-element.ts
ST-DDT: If we are already at it. Should we also flag them as `@experimental`?

src/modules/helpers/replace-credit-card-symbols.ts
ST-DDT: I will move it to a separate file.

src/modules/helpers/array-element.ts
xDivisionByZerox: Do we plan to have this feature being stable in v11? If so I don't think we need it, since we will likely do a pre-release of v11 which is not stable by design. If we are not sure about the stability of the feature, we should mark them as `@experimental` and add docs what users can expect from features being marked as such.

ST-DDT: > This would affect the "auto generated" comment or the general generation of module files?

I will likely ignore the cleanup in the comment.

```ts
// The module content is auto generated.
// Run `pnpm tsx scripts/generate-module-tree.ts <module>` to update the methods.
// To add new methods to the class add a stub `<method>() {}` after implementing the `<method>.ts`.
```

ST-DDT: > Do we plan to have this feature being stable in v11?

I'm not sure about that.
Currently the flag isnt visible to our users, that may change with SMFs getting a public api.
We do have support for it in our api-docs.

I may just add them now and we can remove them later if needed.

test/modules/number.spec.ts
ST-DDT: Created separate issue for tracking.

- #4019

src/modules/string/sample.ts
Shinigami92: The most straightforward would be stringSample

src/modules/helpers/array-element.ts
ST-DDT: `@since 11.0.0 (8.4.0)` causes lint errors (or similar).
I will just use `@since 11.0.0` for the SMFs and re-use the existing `@since` tags when re-generating the module tree.
That should work most of the time.

src/modules/string/sample.ts
ST-DDT: Renamed

src/modules/helpers/replace-credit-card-symbols.ts
ST-DDT: Moved

src/modules/helpers/array-element.ts
ST-DDT: - Using `@since 11.0.0` for SMF.
- Retain original since in modules.
- Add `@experimental` flag
- Add auto generated comment to module
- Adapted `generate:module-tree` script to be repeatable, without any manual cleanup

ST-DDT:
So, I updated this PR with the following improvements:

- Renamed `sample` -> `moduleSample` or do we want to prefix all methods?
- Moved `legacyReplaceSymbolWithNumber` to own file
- Using `@since 11.0.0` for SMF.
- Retain original since in modules.
- Add `@experimental` flag
- Add auto generated comment to module
- Adapted `generate:module-tree` script to be repeatable, without any manual cleanup
  - Not yet part of our CI/preflight
- Improved scripts to only add required imports simplifying the post script cleanup.
- Added ability to run `pnpm run generate:module-tree [module]` optionally limited to a single module.

The scripts can be run via:

- `pnpm tsx scripts/temp-transform-once.ts`
- `pnpm run generate:module-tree [module]`


This PR contains the following checkpoints (preflight green)

- ✔️ `refactor: _prefix module helpers`
- ✔️ `feat: introduce module to standalone transform script`
- ✔️ `chore: limit script to ALLOWED_MODULES`
- ❌ `chore: run transform script (dirty)`
- ❌ `refactor: transform helpers module`
- ❌ `refactor: transform number module`
- ❌ `refactor: transform string module`
- ✔️ `refactor: update phone module imports`
- ✔️ `infra: introduce generate-module-tree script`
- ✔️ `chore: limit script to ALLOWED_MODULES`
- ✔️ `chore: apply generate-module-tree script (dirty)`
- ✔️ `chore: remove temp exports`

If you don't want to review everything from scratch, you can use the following command (diff to previous version):

````sh
git diff  c8e57c4 -- src/modules/
````

- Also the cleanup commits should be now be smaller and fewer.

🤣 Fun fact: To apply the changes, I only had to rebase exactly 30 times (28 times for the big PR, 2 times for this PR).

File: src/modules/string/_types.ts [hunk 8]
```diff
@@ -0,0 +1,78 @@
+export const UPPER_CHARS: ReadonlyArray<string> = [
+  ...'ABCDEFGHIJKLMNOPQRSTUVWXYZ',
+];
+export const LOWER_CHARS: ReadonlyArray<string> = [
+  ...'abcdefghijklmnopqrstuvwxyz',
+];
+export const DIGIT_CHARS: ReadonlyArray<string> = [...'0123456789'];
```
xDivisionByZerox, line 7: **nitpick:** I generally agree that these lists should be in a centralized location (I think this is new to my previous review), but I have to points to add:

1. The name of the files does not match the content anymore. Something like `_characters.ts` would be more fiting.
2. Each respective list should probably be typed as the actual character set it represents (`UPPER_CHARS: ReadonlyArray<UpperAlphaChar> = ...`, etc)

xDivisionByZerox (approved):
For the sake of moving forward, I approve this PR.

I do have opinions regarding the generation script(s); mostly architectural, which probably comes down to preference. The current script works, and can be more easily refactored in the future when an agreement is reached how the final product (the autogenerated modules) should look like.

However, I have to compliment the great work on the general structure and continues work on the feature. I do not only appreciate the work on the code itself, but the entire process around it: draft of the feature and the initial communication of the general intend/goal behind it; full fledged PR descriptions that not only explain the general context of the change a reviewer can expect as well as giving a "guide" on the review process if needed; the continues rework/updates inside the comments of the PR to clarify the work you have done so a bystander has an easier time to understand the change/timeline without having to grasp every detail inside the code.
Good job! :+1:

src/modules/string/_types.ts
ST-DDT: > 1. The name of the files does not match the content anymore.

The three hardest problems in software engineering.

- Naming
- Naming
- Naming
- Of by one errors

# 🤣

I think my script relies/relied on the file being named _types.ts to auto derive imports from it.
But that can be improved in future iterations.

After this PR is merged, I can check whether I can rename the file, without breaking the generator script.

ST-DDT:
> I do have opinions regarding the generation script(s); mostly architectural, which probably comes down to preference. The current script works, and can be more easily refactored in the future when an agreement is reached how the final product (the autogenerated modules) should look like.

I'm also not super happy with it's current structure (and speed).
This is mainly caused by missing exact target specifications/growing with the requirements/being the first version.
As you have mentioned, we can adjust, improve and expand it as we go.

> However, I have to compliment the great work on the general structure and continues work on the feature.

Thank you for your kind words. I really enjoyed working on this, the feedback and discussions.

Shinigami92 (approved):
I had a BIG review with claude, but it only found stuff that is negotiable. I handed that via Discord PM to @ST-DDT and we both agreed that the PR is fine as is 🚀

PR #3866, opened 2026-05-29
chore: fix FakerModule test type

ST-DDT:
Update `FakerModule` test type to latest version.


test/support/seeded-runs.ts
```diff
@@ -28,7 +28,7 @@ type MethodOf<TObjectType, TSignature extends Callable = Callable> = {
 type FakerModule = {
   [Key in keyof Faker]: Faker[Key] extends Callable | string | number | number[]
     ? never
-    : Key extends 'definitions' | 'locales'
+    : Key extends 'definitions' | 'rawDefinitions' | 'fakerCore'
```
Shinigami92: **suggestion (if-minor):** have you considered something like this?

```ts
/**
 * Identity type whose constraint forces every member to be a real key of `Faker`.
 * This makes a stale entry in the {@link FakerModule} exclusion list fail to compile.
 */
type FakerKeys<T extends keyof Faker> = T;

/**
 * A type allowing only the names of faker modules.
 */
type FakerModule = {
  [Key in keyof Faker]: Faker[Key] extends Callable | string | number | number[]
    ? never
    : Key extends FakerKeys<'definitions' | 'rawDefinitions' | 'fakerCore'>
      ? never
      : Key;
}[keyof Faker];
```

ST-DDT: What would be the benefit of that?
Do detect invalidly banned/non-existing keys?
Since it does not detect invalidly allowed keys, I'm kind of neutral about this.

`@`others any preference?

Shinigami92: > detect invalidly banned/non-existing keys

exactly

Shinigami92 (approved):

PR #3882, opened 2026-06-04
feat(helpers): resolve primitive fake property chains

### LINKED ISSUE #3465: helpers.fake - patterns after root expression fail when referencing instance members
### Pre-Checks

#### What I am doing

I am trying to call instance methods of returned generated values from within a `helpers.fake` expression, for example `'{{number.int.toPrecision}}'`. The call failed, which led me down a rabbit hole of trying to figure out why. It turns out that, in many cases, referencing instance/prototype functions of the returned expression value in a pattern fails because the function is not setting `this` when being called.

What's more, the way it fails depends on how the expression is formatted. If the expression uses function shorthand syntax (`'number.int.toPrecision'`), the fake will fail due to the aforementioned bug of `this` not getting set. However, if the expression uses parameterized syntax (`'number.int(100).toPrecision'`), the fake will fail due to the expression not being resolvable. It's also worth noting that in this latter case, properties will also fail. (i.e. `'string.alpha.length'` will succeed, but `'string.alpha(10).length'` will fail.)

#### What I expected

Referencing an instance methods/prototype functions in an expression pattern to further modify the generated value.

#### What actually happened

Referencing an instance methods/prototype functions in an expression pattern results in an error. (The specific error depends on the type of the instance involved and format of the expression. See minimum reproduction.)

### Minimal reproduction code

These expressions will all result in various errors that essentially boil down to "`this` is undefined":

```js
const expNotation = faker.helpers.fake('{{number.int.toPrecision(2)}}'); // Number.prototype.toPrecision requires that 'this' be a Number
const trimmed = faker.helpers.fake('{{string.alpha.trim}}'); // String.prototype.trim called on null or undefined
const isoTimestamp = faker.helpers.fake('{{date.anytime.toISOString}}'); // Method Date.prototype.toISOString called on incompatible receiver undefined
```

These expressions will all result in "Cannot resolve expression '<expression>'":

```js
const expNotation = faker.helpers.fake('{{number.int(100).toPrecision(2)}}');
const trimmed = faker.helpers.fake('{{string.alpha(5).trim}}');
const isoTimestamp = faker.helpers.fake('{{date.anytime({"refDate": 1735714800000}).toISOString}}');
const stringLength = faker.helpers.fake('{{string.alpha(10).length}}');
```

### Additional Context

The impetus for this ticket was that I was trying to convert a `Date` object returned from a date expression (e.g. `date.anytime`) to an ISO timestamp within the resulting string using `'{{date.anytime.toISOString}}'`. The reason I cannot use alternative approaches (such as using `helpers.mustache` and passing the timestamp in as a parameter) is because I am using a utility I created that uses a JSON template object with properties defined using expression strings to generate entire complex objects.

### Environment Info

```shell
System:
    OS: macOS 15.4
    CPU: (14) arm64 Apple M3 Max
    Memory: 1.52 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.asdf/installs/nodejs/22.13.1/bin/node
    npm: 10.9.2 - ~/.asdf/installs/nodejs/22.13.1/bin/npm
    pnpm: 10.5.2 - ~/.asdf/installs/nodejs/22.13.1/bin/pnpm
  Browsers:
    Chrome: 135.0.7049.41
    Safari: 18.4
```

- [ ] CJS
- [x] ESM

### Used Package Manager

pnpm

puneetdixit200:
## Summary
- resolve properties on primitive values returned while evaluating helpers.fake/fakeEval chains
- bind resolved prototype methods to their source value so chained calls keep the correct receiver
- add regressions for implicit and explicit generated primitive property chains

Closes #3465

## Verification
- `corepack pnpm exec vitest run test/modules/helpers-eval.spec.ts test/modules/helpers.spec.ts`
- `corepack pnpm exec prettier --check src/modules/helpers/eval.ts test/modules/helpers-eval.spec.ts test/modules/helpers.spec.ts`
- `corepack pnpm exec eslint src/modules/helpers/eval.ts test/modules/helpers-eval.spec.ts test/modules/helpers.spec.ts`
- `git diff --check`
- `corepack pnpm exec vitest run test/scripts/shared/markdown.spec.ts`
- `corepack pnpm run ts-check`
- `corepack pnpm run test`

### DIFF
--- a/src/modules/helpers/eval.ts
+++ b/src/modules/helpers/eval.ts
@@ -217,15 +217,41 @@ function resolveProperty(entrypoint: unknown, key: string): unknown {
         return undefined;
       }
 
-      return entrypoint?.[key as keyof typeof entrypoint];
+      return getProperty(entrypoint, key);
     }
 
     case 'object': {
-      return entrypoint?.[key as keyof typeof entrypoint];
+      return getProperty(entrypoint, key);
+    }
+
+    case 'bigint':
+    case 'boolean':
+    case 'number':
+    case 'string': {
+      return getProperty(entrypoint, key);
     }
 
     default: {
       return undefined;
     }
   }
+}
+
+type Callable = (this: unknown, ...args: unknown[]) => unknown;
+
+/**
+ * Resolves the given property and binds methods to their source value.
+ *
+ * @param entrypoint The entrypoint to resolve the property on.
+ * @param key The property name to resolve.
+ */
+function getProperty(entrypoint: unknown, key: string): unknown {
+  if (entrypoint == null) {
+    return undefined;
+  }
+
+  const value = (new Object(entrypoint) as Record<string, unknown>)[key];
+  return typeof value === 'function'
+    ? (value as Callable).bind(entrypoint)
+    : value;
 }
--- a/test/modules/helpers-eval.spec.ts
+++ b/test/modules/helpers-eval.spec.ts
@@ -115,6 +115,17 @@ describe('fakeEval()', () => {
     ).toContain(actual);
   });
 
+  it('supports prototype methods after generated primitive values', () => {
+    const entrypoint = { value: () => 1234 };
+
+    expect(fakeEval('value.toPrecision(3)', faker, [entrypoint])).toBe(
+      '1.23e+3'
+    );
+    expect(fakeEval('value().toPrecision(3)', faker, [entrypoint])).toBe(
+      '1.23e+3'
+    );
+  });
+
   it('requires a dot after a function call', () => {
     expect(() => fakeEval('airline.airline()iataCode', faker)).toThrow(
       new FakerError(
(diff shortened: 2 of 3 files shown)


ST-DDT:
Thanks for contributing to faker. ❤️ 

Please note: This may cause a merge conflict with #3857 

- #3857

So this may take far longer than usual to merge.

ST-DDT:
Please update the relevant jsdoc as well e.g.

https://github.com/faker-js/faker/blob/70994dbf77c9aea4edcf2dffbb46c777376a7814/src/modules/helpers/index.ts#L1249

ST-DDT:
There seems to be a merge conflict. Please resolve it and the prior comment, then mark the PR as ready for review again.

ST-DDT:
Closing PR as abandonned. Feel free to ping, if you wish to update.

PR #3883, opened 2026-06-04
docs: escape refreshable code results

ST-DDT:
Based on https://github.com/faker-js/faker/pull/3879#issuecomment-4620819009

- https://github.com/faker-js/faker/pull/3879#issuecomment-4620819009

---

To eliminate any risk from refreshable code examples, escape the result of refreshable code before inserting it.
This adds to the existing defense layers:

- code review of examples
- log only error type 
- no html like output


Shinigami92 (approved):

xDivisionByZerox (approved):
Damn, you are fast.

PR #3885, opened 2026-06-06
docs: handle experimental flag

ST-DDT:
- Add support for `@experimental` jsdoc flag.
  - Currently, the flag does not support additional text, as experimental is self explanatory (IMO)
- Flag get/setDefaultRefDate SMF as `@experimental`
- FakerCore is currently not flagged
- Update the experimental banner copy to explicitly mention semantic versioning and clarify that **future** changes may not follow SemVer.

### Preview

- <a href="https://deploy-preview-3885.fakerjs.dev/api/utils.html#getdefaultrefdate">getDefaultRefDate</a>
- <a href="https://deploy-preview-3885.fakerjs.dev/api/utils.html#setdefaultrefdate">setDefaultRefDate</a>

### DIFF
--- a/docs/.vitepress/components/api-docs/method.vue
+++ b/docs/.vitepress/components/api-docs/method.vue
@@ -8,6 +8,7 @@ import RefreshableCode from './refreshable-code.vue';
 const { method } = defineProps<{ method: ApiDocsMethod }>();
 const {
   deprecated,
+  experimental,
   description,
   remark,
   since,
@@ -38,6 +39,11 @@ function seeAlsoToUrl(see: string): string {
       <p class="custom-block-title">Deprecated</p>
       <p>This method is deprecated and will be removed in a future version.</p>
       <span v-html="deprecated" />
+    </div>
+
+    <div v-if="experimental" class="warning custom-block">
+      <p class="custom-block-title">Experimental</p>
+      <p>This method is experimental and may change in future versions.</p>
     </div>
 
     <div v-html="description"></div>
--- a/docs/.vitepress/components/api-docs/method.vue
+++ b/docs/.vitepress/components/api-docs/method.vue
@@ -43,7 +43,7 @@ function seeAlsoToUrl(see: string): string {
 
     <div v-if="experimental" class="warning custom-block">
       <p class="custom-block-title">Experimental</p>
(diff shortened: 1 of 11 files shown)


scripts/apidocs/processing/jsdocs.ts
```diff
@@ -21,6 +21,10 @@ export function getDeprecated(jsdocs: JSDoc): string | undefined {
   return getOptionalTagFromJSDoc(jsdocs, 'deprecated');
 }
 
+export function getExperimental(jsdocs: JSDoc): true | undefined {
```
Shinigami92, line 24: **question (non-blocking):** is the return type correct?
`getExperimental` is a bit weird to return `true | undefined`

Shinigami92 (dismissed):

scripts/apidocs/processing/jsdocs.ts
ST-DDT: I considered `boolean | undefined`, but I use `undefined` instead of `false` to avoid writing it to/parsing it from JSON.

grafik
grafik

matthewmayer:
Is this supposed to imply that this method is not subject to semver rules?

ST-DDT:
Yes. Should we explicitly mention that in the banner or is the current text sufficient?

matthewmayer:
I think it would be good to mention explicitly.

ST-DDT:
Changed to:

> This method is experimental and changes may not follow semantic versioning.

Feel free to suggest alternatives.

docs/.vitepress/components/api-docs/method.vue
```diff
@@ -40,6 +41,14 @@ function seeAlsoToUrl(see: string): string {
       <span v-html="deprecated" />
     </div>
 
+    <div v-if="experimental" class="warning custom-block">
+      <p class="custom-block-title">Experimental</p>
+      <p>
+        This method is experimental and changes may not follow semantic
```
matthewmayer: Maybe "future changes" instead of "changes"?

ST-DDT: @copilot apply suggestion and then run pnpm run format.

Shinigami92 (approved):

PR #3907, opened 2026-06-24
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here.

The install instructions list npm, pnpm, and yarn, but not Deno. Since Deno is a drop-in replacement for npm these days, I'd like to add a Deno line so Deno users have a copy-pasteable command, in parity with the others:

```sh
deno add -D @faker-js/faker
```

(As of Deno 2.8 there's no `npm:` prefix needed, it's the same command shape as `npm install @faker-js/faker`.)

Totally fine to close this if you'd rather keep the list short, no hard feelings. Happy to adjust wording, placement, or move it elsewhere.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -39,6 +39,10 @@ pnpm add @faker-js/faker --save-dev
 yarn add @faker-js/faker --dev
 ```
 
+```shell [deno]
+deno add --dev @faker-js/faker
+```
+
 :::
 
 ## Faker Modules


Shinigami92 (approved):
> 👋 Bartek from the Deno team here.

Hello @bartlomieju 😲 I'm pleasantly honored.

docs/guide/index.md
```diff
@@ -39,6 +39,10 @@ pnpm add @faker-js/faker --save-dev
 yarn add @faker-js/faker --dev
 ```
 
+```shell [deno]
+deno add --dev @faker-js/faker
```
xDivisionByZerox, line 43: **nitpick:** All other commands list the "--dev" flag as the last parameter. Mostly, because some might want to use faker for things other than tests. Removing the last argument (instead of navigating and deleting) after copy'n'pasting should improve the UX ever so slightly.
Is it possible to provide the "--dev" argument last on deno as well? Or does the add command ignore everything after the default argument/package name?

xDivisionByZerox (approved):
Thanks for the PR. I do not see any problems with providing examples for deno commands in general.

docs/guide/index.md
bartlomieju: It's the latter - if `--dev` is last then it's treated as a package name.

ST-DDT (approved):

ST-DDT:
Thanks for contributing to faker ❤️

docs/guide/index.md
xDivisionByZerox: Alright. In that case there is nothing we can do. I'll close this conversation to unblock a merge :)

PR #3916, opened 2026-06-24
infra!: drop support for node v20

Shinigami92:
- Drop support for Node v20 (EOL)
- Drop support for es2023 (new target is es2024 as minimum fully covered by Node v22.13)


ST-DDT:
FYI: https://github.com/faker-js/faker/blob/e0223a7aea577337e2ea30db324933730b0e7a24/test/scripts/apidocs/verify-jsdoc-tags.spec.ts#L81 

- #3914

matthewmayer:
Can i suggest we use a similar pattern for what we did in v9 and where there are breaking changes queued up for v11 we add something like

`docs/guide/upgrading_v11/3916.md`

with a snippet explaining the breaking change. This will make writing the v11 upgrading guide easier.

xDivisionByZerox:
You don't need to suggest it. You can simply point at our own contributions guidelines:

- https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md#L202

Shinigami92:
> Can i suggest we use a similar pattern for what we did in v9 and where there are breaking changes queued up for v11 we add something like
> 
> `docs/guide/upgrading_v11/3916.md`
> 
> with a snippet explaining the breaking change. This will make writing the v11 upgrading guide easier.

Thanks for the reminder 👍

docs/guide/upgrading_v11/3916.md
```diff
@@ -0,0 +1,3 @@
+### Node v20 No Longer Supported
+
+Support for Node.js v20 has been discontinued, as this version has reached its [end-of-life](https://github.com/nodejs/Release). Faker.js v11 requires a minimum of Node.js v22.13.0, v23.5.0, or v24.0.0.
```
xDivisionByZerox:
```suggestion
Support for Node.js v20 has been discontinued, as this version has reached its [end-of-life](https://github.com/nodejs/Release). Faker v11 requires a minimum of Node.js v22.13.0, v23.5.0, or v24.0.0.
```

See: https://fakerjs.dev/about/announcements/2022-01-14.html#is-it-called-faker-or-faker-js
I also really don't like the dot-style, as it was the name of the old package. I would prefer the FakerJS style if the JS part is required in your opinion.

xDivisionByZerox (commented):
Just saying, but technically the PR title (and thus the resulting commit message) is misleading since dropping support for node v22 does not include the automatic availability of node v26. Technically, Node 26 support could already be provided in v10.No breaking change required.

docs/guide/upgrading_v11/3916.md
Shinigami92: I copied that from https://fakerjs.dev/guide/upgrading.html#node-v18-no-longer-supported, but I'm ok with your suggestion

Shinigami92:
> Just saying, but technically the PR title (and thus the resulting commit message) is misleading since dropping support for node v22 does not include the automatic availability of node v26. Technically, Node 26 support could already be provided in v10.No breaking change required.

Should I explicitly split it into two PRs? Or will we not think back on it in just a few days? 😅

xDivisionByZerox:
> Should I explicitly split it into two PRs? Or will we not think back on it in just a few days? 😅

**I** would prefer to have two PRs (and thus two dedicated commit messages/CHANGELOG entries). One for dropping Node 20 and one for supporting Node 26.
@faker-js/members opinions?

Shinigami92:
> > Should I explicitly split it into two PRs? Or will we not think back on it in just a few days? 😅
> 
> **I** would prefer to have two PRs (and thus two dedicated commit messages/CHANGELOG entries). One for dropping Node 20 and one for supporting Node 26. @faker-js/members opinions?

ok, will do so in a moment

xDivisionByZerox (approved):

PR #3922, opened 2026-06-28
feat(locale): add 060 Japanese mobile phone number formats

noritaka1166:
- Add Japanese mobile phone number formats for the new `060-1` through `060-9` ranges to `phone_number.format.mobile`.

  Source: https://www.soumu.go.jp/main_sosiki/joho_tsusin/top/tel_number/060keitai.html


matthewmayer:
Don't think it's necessary to touch the legacy cell_phone definitions since these will get removed in the next major version.

Shinigami92 (dismissed):

matthewmayer:
I think the test is not really necessary either.

Shinigami92:
I'm ok with letting this into our last planned v10.6 before we start v11.

matthewmayer (approved):

Shinigami92 (approved):

noritaka1166:
> Don't think it's necessary to touch the legacy cell_phone definitions since these will get removed in the next major version.

Good point, thanks. I removed the changes to the legacy `cell_phone` definitions and kept this scoped to `phone_number.format.mobile`.

noritaka1166:
> I think the test is not really necessary either.

Makes sense. I removed the test as well and kept the PR to just the locale data update.

ST-DDT (approved):
AFAICT there aren't valid yet, but may become valid starting July 2026.

PR #3926, opened 2026-06-29
chore: consistent type import paths

ST-DDT:
I noticed that the helpers module imports it's types directly from the root index, while all other modules import the faker instance from the faker file directly. This (minimally) speeds up build performance for partial builds, as a smaller set of files need to be parsed for the build process to complete.

I searched for: `import (type )?\{ .*(\.|/)'`
in: `src`
excluding: `src/locales`

`src/locales` always imports the definitions from the root index.
I plan to address that once this PR is merged, as it requires changes to the `generate:locales` script.

### DIFF
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1,6 +1,7 @@
-import type { Faker, SimpleFaker } from '../..';
 import { FakerError } from '../../errors/faker-error';
+import type { Faker } from '../../faker';
 import { SimpleModuleBase } from '../../internal/module-base';
+import type { SimpleFaker } from '../../simple-faker';
 import type { NumberRange } from '../../utils/types';
 import { fakeEval } from './eval';
 import { luhnCheckValue } from './luhn-check';
--- a/src/modules/string/uuid.ts
+++ b/src/modules/string/uuid.ts
@@ -1,4 +1,4 @@
-import type { SimpleFaker } from '../../';
+import type { SimpleFaker } from '../../simple-faker';
 
 /**
  * Returns a UUID v4 ([Universally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)).
--- a/src/utils/merge-locales.ts
+++ b/src/utils/merge-locales.ts
@@ -1,4 +1,4 @@
-import type { LocaleDefinition } from '..';
+import type { LocaleDefinition } from '../definitions';
 
 /**
  * Merges the given locales into one locale.


xDivisionByZerox (approved):

Shinigami92 (approved):
Would it be possible to use a lint rule for this?

ST-DDT:
> Would it be possible to use a lint rule for this?

Maybe, though I'm not sure what to search for.

Shinigami92:
> > Would it be possible to use a lint rule for this?
> 
> 
> 
> Maybe, though I'm not sure what to search for.

Literally paste the url of this PR and tell your AI to make a WebFetch. It is 2026.

Shinigami92:
@ST-DDT I gave claude this request:
> Is there a lint rule to prevent further cases like https://github.com/faker-js/faker/pull/3926 ?

<details>
<summary>Claude Response</summary>

I've read the PR. The maintainer (Shinigami92) asked exactly this, and ST-DDT wasn't sure what to search for. Let me check what's already in faker's ESLint setup so I can give you something that drops in cleanly.Let me check faker's current ESLint config directly to tailor this.Yes — and `import/no-restricted-paths` is the rule that actually fits, because unlike the plain `no-restricted-imports` it *resolves* the import to a file before checking. That matters here: the offending imports were relative (`from '.'` / `from '../..'`), and a string-based rule can't reliably catch "this relative path happens to resolve to the root barrel." `no-restricted-paths` can, via its `zones` config (`target` = files that are restricted, `from` = what they may not import, plus `except`).

Tailored to what the PR describes — everything under `src` should import from the concrete source file, never from the root barrel `src/index.ts`, with the `src/locales` carve-out ST-DDT mentioned:

```js
// eslint.config.ts (flat config)
{
  files: ['src/**/*.ts'],
  rules: {
    'import/no-restricted-paths': ['error', {
      zones: [
        {
          target: './src',
          from: './src/index.ts',
          // locales still import definitions from the root index until
          // the generate:locales script is reworked (the follow-up PR)
          except: ['./src/locales'],
          message:
            'Import from the specific source file (e.g. ./faker or the concrete definitions/types file), not the root barrel.',
        },
      ],
    }],
  },
}
```

The index file re-exporting from its own modules is fine — it's the *source* of the barrel, not an importer of it. And `no-restricted-paths` deliberately doesn't autofix, so it'll flag rather than rewrite.

A few caveats worth flagging to the maintainers:

`eslint-plugin-import`'s flat-config support has historically been finicky, and faker is ESM-only on flat config now. If it isn't already a dependency, `eslint-plugin-import-x` is the actively maintained, flat-config-native fork with the same `no-restricted-paths` rule and identical config shape — usually the smoother choice for a setup like theirs.

If the real intent is broader than "the root barrel" — i.e. ST-DDT's search regex `import (type )?\{ .*(\.|/)'` was catching *any* import that resolves to a directory/index rather than an explicit file — then `no-restricted-paths` only covers the specific barrel you name. For the general "no barrel/directory imports anywhere" rule, either `eslint-plugin-barrel-files` (rule `avoid-importing-barrel-files`, purpose-built for this) or `import/extensions: ['error', 'always']` (which forces explicit file paths and so inherently rejects `from '.'`) generalizes it, at the cost of more noise.

One thing the plain core `no-restricted-imports` / `@typescript-eslint/no-restricted-imports` *can* do that this can't is `allowTypeImports` — but that's the opposite of what this PR wants, since here the goal is to forbid the type imports from the barrel too, so it's the wrong tool for this particular case.

If you want, I can sketch the matching `except`/zone setup for the locales follow-up, or draft a short reply you could drop straight into the PR thread.

</details>

So maybe we could investigate further with https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-restricted-paths.md

PR #3935, opened 2026-07-02
docs(datatype): drop the two-decimal-places note from boolean()

spokodev:
`faker.datatype.boolean()` compares against the raw float, so the JSDoc claim that the probability is limited to two decimal places was never honored. Per the discussion below, this removes the note from the JSDoc so the documentation matches the behavior, instead of changing the implementation.

### DIFF
--- a/src/modules/datatype/module.ts
+++ b/src/modules/datatype/module.ts
@@ -15,7 +15,6 @@ export class DatatypeModule extends SimpleModuleBase {
    * A probability of `0.75` results in `true` being returned `75%` of the calls; likewise `0.3` => `30%`.
    * If the probability is `<= 0.0`, it will always return `false`.
    * If the probability is `>= 1.0`, it will always return `true`.
-   * The probability is limited to two decimal places.
    *
    * @param options The optional options object or the probability (`[0.00, 1.00]`) of returning `true`.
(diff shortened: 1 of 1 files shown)


ST-DDT:
Nicely spotted.
The two decimal places are a limitation of the original float implementation.

https://github.com/faker-js/faker/blob/98ff809bea1adcd9b2872b9a5a864a59f3df6b86/src/modules/datatype/index.ts#L122-L127

However, I would rather strip the limitation from the jsdocs than forcing the current implememtation to follow it.

spokodev:
That works for me. I can switch the PR to remove the two-decimal note from the jsdoc and drop the behavior change, so the documentation matches the current implementation instead of the other way around. Let me know and I will update it.

matthewmayer:
Agree prefer to change the docs rather than the code.

spokodev:
Updated. Switched to a docs change: dropped the two-decimal-places note from the boolean() JSDoc so it matches the raw-float behavior, reverted the code and test, and rebased onto next. Title and description updated to match.

xDivisionByZerox (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #3938, opened 2026-07-05
fix(locale): use only common surnames for zh_CN last names

### LINKED ISSUE #3466: using "common last_name" when generate chinese name

MahinAnowar:
Closes #3466

### Problem

`zh_CN` last names are drawn uniformly from a ~1000-entry list in which the large majority are ultra-rare surnames (不, 丑, 九, 业, 之, …) that many Chinese speakers never encounter in real life. As the issue reporter showed, most generated names look unnatural at first glance because the surname is almost always one of these rare ones.

### Fix (data-only, per maintainer guidance)

@ST-DDT confirmed in the thread that "both weighting and/or reducing the number of rare last names in that list are valid solutions". With the weighting PoC (#3467) closed, this PR takes the **reduction** route: replace `last_name.generic` with the curated list of the most common surnames posted in the issue by @yyz945947732 — 150 single-character surnames plus 12 common compound surnames (欧阳, 诸葛, 司马, …), covering ~85%+ of the population. This mirrors the scope of the existing `zh_TW` list (~100 common surnames). No code changes; `last_name_pattern` is untouched.

The file was normalized with `pnpm run generate:locales`, and the `locale-data` character-inventory snapshot updated accordingly (one line: `zh_CN`).

### Verification

Added a regression test in `test/modules/person.spec.ts` asserting the `zh_CN` list is the curated common set (contains 王/李/张/刘/陈/欧阳, excludes the rare 不/丑/九/业/之, length < 200). It **fails on the old data and passes with this change**. Full runs green locally: `person.spec` (214), `locale-data` (2635), `locale-imports` (152), `all-functional` (36k), `ts-check`, lint.

Credit for the curated list goes to @yyz945947732 in the issue.


test/modules/person.spec.ts
ST-DDT, line 1: IMO we dont need a test for that.

ST-DDT (commented):
Thanks for contributing. ❤️

test/modules/person.spec.ts
MahinAnowar: Fair enough for a data-only change — removed in b511eb1. Thanks for the quick triage!

matthewmayer (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #3939, opened 2026-07-06
refactor: extract shared NumberOrRange type

Shinigami92:
related to #1443 

Preview:

- https://deploy-preview-3939.fakerjs.dev/api/string.html#alpha

### DIFF
--- a/scripts/apidocs/processing/type.ts
+++ b/scripts/apidocs/processing/type.ts
@@ -95,18 +95,36 @@ export function getTypeText(
         ...type.getAliasTypeArguments(),
       ];
 
-      if (name === 'LiteralUnion') {
-        const displayType = getTypeText(typeArguments[0], options);
-        const baseType = typeArguments[1]
-          ? getTypeText(typeArguments[1], options)
-          : newSimpleType('string');
+      switch (name) {
+        case 'LiteralUnion': {
+          const displayType = getTypeText(typeArguments[0], options);
+          const baseType = typeArguments[1]
+            ? getTypeText(typeArguments[1], options)
+            : newSimpleType('string');
 
-        return newUnionType([displayType, baseType]);
-      } else if (name === 'NumberRange') {
-        return newObjectType([
-          { name: 'min', type: newSimpleType('number') },
-          { name: 'max', type: newSimpleType('number') },
-        ]);
+          return newUnionType([displayType, baseType]);
+        }
+
+        case 'NumberRange': {
+          return newObjectType([
+            { name: 'min', type: newSimpleType('number') },
+            { name: 'max', type: newSimpleType('number') },
+          ]);
+        }
+
+        case 'NumberOrRange': {
+          return newUnionType([
+            newSimpleType('number'),
+            newObjectType([
+              { name: 'min', type: newSimpleType('number') },
+              { name: 'max', type: newSimpleType('number') },
+            ]),
+          ]);
+        }
+
+        default: {
+          break;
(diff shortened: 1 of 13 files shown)


src/utils/types.ts
```diff
@@ -20,3 +20,8 @@ export interface NumberRange {
    */
   max: number;
 }
+
+/**
+ * Either an exact number or a range of numbers with an inclusive minimum and maximum.
+ */
+export type NumberOrRange = number | NumberRange;
```
ST-DDT, line 27: We have to add docs handling for the type similar to NumberRange.

Shinigami92: yes, I hoped that you could take over from here again, would you be up for this?

ST-DDT: Just look at the last PR, it is very easy.

scripts/apidocs/processing/type.ts
```diff
@@ -84,15 +84,27 @@ export function getTypeText(
...
+
+          return newUnionType([displayType, baseType]);
+        }
+
+        case 'NumberRange': {
+          return newSimpleType('{ min: number; max: number }');
+        }
+
+        case 'NumberOrRange': {
+          return newSimpleType('number | { min: number; max: number }');
```
ST-DDT: Maybe use `newUnionType()`

ST-DDT (commented):
I added a preview link to the description.
If possible/relevant, please add a docs link for easy verification yourself.

ST-DDT (dismissed):

ST-DDT:
Note: this is no longer detected as option like if used as standalone parametern.

https://fakerjs.dev/api/string.html#sample

grafik

https://deploy-preview-3939.fakerjs.dev/api/string.html#sample

grafik

Should we remove jsdoc `@param length.min/max` in that case?

Shinigami92:
> Note: this is no longer detected as option like if used as standalone parametern.
> 
> [fakerjs.dev/api/string.html#sample](https://fakerjs.dev/api/string.html#sample)
> 
> grafik
> [deploy-preview-3939.fakerjs.dev/api/string.html#sample](https://deploy-preview-3939.fakerjs.dev/api/string.html#sample)
> 
> grafik
> Should we remove jsdoc `@param length.min/max` in that case?

I not really sure yet, while working on the doc script implementation, I was already thinking why there is only `newUnionType` and `newSimpleType`, and I felt like I miss somehow a `newObjectType` 🤔 And maybe this here is the thing uncovered it 🤔

ST-DDT:
If you want to restore the previous behavior, you may need to look at this:

https://github.com/faker-js/faker/blob/58394eb2610ae8fe15ab9b1b2a17c108fd8e9749/scripts/apidocs/processing/parameter.ts#L173

Though I'm not sure the parameter is an object type in the first place.

Stepping away from the previous behavior for this case, may be an option as well.

Shinigami92:
> If you want to restore the previous behavior, you may need to look at this:
> 
> https://github.com/faker-js/faker/blob/58394eb2610ae8fe15ab9b1b2a17c108fd8e9749/scripts/apidocs/processing/parameter.ts#L173
> 
> Though I'm not sure the parameter is an object type in the first place.
> 
> Stepping away from the previous behavior for this case, may be an option as well.

Thanks for the context, I won't have time today to work on that, earliest tomorrow evening.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3940, opened 2026-07-06
refactor: extract shared LengthStrategy type

Shinigami92:
related to #1443 

Preview:

- https://deploy-preview-3940.fakerjs.dev/api/lorem.html#word

### DIFF
--- a/docs/.vitepress/components/api-docs/method-parameters.vue
+++ b/docs/.vitepress/components/api-docs/method-parameters.vue
@@ -29,7 +29,7 @@ const { parameters } = defineProps<{ parameters: ApiDocsMethodParameter[] }>();
           >
             {{ name }}
           </td>
-          <td v-html="type"></td>
+          <td class="type" v-html="type"></td>
           <td>
             <code v-if="def">{{ def }}</code>
           </td>
@@ -44,12 +44,14 @@ const { parameters } = defineProps<{ parameters: ApiDocsMethodParameter[] }>();
 td.deprecated {
   text-decoration: line-through;
 }
-</style>
 
-<!-- Unscoped: targets markup injected via v-html (shadow type popovers). -->
-<style>
-.shadow-type-value {
+.type {
   font-family: var(--vp-font-family-mono);
+}
+
+/* `:deep` reaches the popover markup injected into the type column via v-html. */
+.type :deep(.shadow-type-value) {
+  font-family: inherit;
   font-size: inherit;
   color: var(--vp-c-brand-1);
   text-decoration-line: underline;
@@ -58,20 +60,9 @@ td.deprecated {
   cursor: help;
 }
 
-/*
- * Author-controlled display: the UA rule that hides a closed popover has the
- * lowest priority and is overridden by VitePress' `.vp-doc` block styling,
- * which would otherwise render every popover inline. Reassert it explicitly.
- */
-.shadow-type-popover {
+/* `display: none` overwrites VitePress' `.vp-doc` default of `inline`. */
+.type :deep(.shadow-type-popover) {
   display: none;
-}
-
-.shadow-type-popover:popover-open {
-  display: block;
-}
-
-.shadow-type-popover {
   max-width: 20rem;
   margin: 0;
   padding: 0.5rem 0.75rem;
@@ -79,14 +70,19 @@ td.deprecated {
   border-radius: 8px;
   background-color: var(--vp-c-bg-elv);
   color: var(--vp-c-text-1);
+  font-family: var(--vp-font-family-base);
   font-size: 0.875rem;
   line-height: 1.5;
   box-shadow: var(--vp-shadow-3);
 }
 
+.type :deep(.shadow-type-popover:popover-open) {
+  display: block;
+}
+
 /* Position next to the trigger where anchor positioning is supported. */
 @supports (anchor-name: --x) {
-  .shadow-type-popover {
+  .type :deep(.shadow-type-popover) {
     position: absolute;
     position-area: bottom span-right;
     margin-top: 4px;
--- a/docs/.vitepress/components/api-docs/method-parameters.vue
+++ b/docs/.vitepress/components/api-docs/method-parameters.vue
@@ -63,8 +63,12 @@ td.deprecated {
  * lowest priority and is overridden by VitePress' `.vp-doc` block styling,
  * which would otherwise render every popover inline. Reassert it explicitly.
  */
-.shadow-type-popover:not(:popover-open) {
+.shadow-type-popover {
   display: none;
+}
+
+.shadow-type-popover:popover-open {
+  display: block;
 }
 
 .shadow-type-popover {
--- a/docs/.vitepress/components/api-docs/method-parameters.vue
+++ b/docs/.vitepress/components/api-docs/method-parameters.vue
@@ -29,7 +29,7 @@ const { parameters } = defineProps<{ parameters: ApiDocsMethodParameter[] }>();
           >
             {{ name }}
           </td>
-          <td>{{ type }}</td>
+          <td v-html="type"></td>
           <td>
             <code v-if="def">{{ def }}</code>
           </td>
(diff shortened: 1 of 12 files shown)


src/utils/types.ts
```diff
@@ -20,3 +20,21 @@ export interface NumberRange {
...
+ *
+ * Available error handling strategies:
+ *
+ * - `fail`: Throws an error if no words with the given length are found.
+ * - `shortest`: Returns any of the shortest words.
+ * - `closest`: Returns any of the words closest to the given length.
+ * - `longest`: Returns any of the longest words.
+ * - `any-length`: Returns a word with any length.
+ */
+export type LengthStrategy =
```
ST-DDT: Should we use enums+types here? Like SexType

Shinigami92: This might be a really good idea 👍 I will do that

Shinigami92:
I will add the docs here when #3939 got merged, to not fall into to many merge conflicts

ST-DDT:
> I will add the docs here when #3939 got merged, to not fall into to many merge conflicts

I have added a preview link for ease of verifcation.
AFAICT this PR does not need a docs upgrade, since the union type is already covered.

ST-DDT (dismissed):

xDivisionByZerox (commented):
We would loose the information describing each strategy itself with this PR:

| Before | After |
| ----- | ----- |
|  |  |

Shinigami92:
> We would loose the information describing each strategy itself with this PR:
> 
> Before	After
> 	

That is a blocker for me 🙁  
I can throw Claude on this again in next few days

ST-DDT:
Should we add a Types page?

Shinigami92:
> Should we add a Types page?

If so, it needs clickable links from the method docs. But then you might loose the context / back history and get lost. How about hover tooltips? But these might introduce problems with mobile device docs. 🤔 Do you have further ideas?

ST-DDT:
Use the shadow type system and add the description as hoverable text.
Aka use the detected types to do actual work.

Shinigami92:
image

popover works on desktop and mobile (the popover api is stable since 2024)

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -45,3 +45,51 @@ td.deprecated {
...
+ */
+.shadow-type-popover {
+  display: none;
+}
+
+.shadow-type-popover:popover-open {
+  display: block;
+}
+
+.shadow-type-popover {
```
ST-DDT: This block is the same as above. Maybe merge them into one?

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -45,3 +45,51 @@ td.deprecated {
...
+  text-decoration-style: dotted;
+  text-underline-offset: 2px;
+  cursor: help;
+}
+
+/*
+ * Author-controlled display: the UA rule that hides a closed popover has the
+ * lowest priority and is overridden by VitePress' `.vp-doc` block styling,
+ * which would otherwise render every popover inline. Reassert it explicitly.
+ */
```
ST-DDT: Not sure whether we need this text.

If you think it is needed maybe shorten it to:

> Overwrites VitePress' `.vp-doc` default of `inline`

scripts/apidocs/output/page.ts
```diff
@@ -247,3 +249,76 @@ export async function toRefreshFunction(
...
+ * Turns an arbitrary string into a slug safe for use in html ids and css idents.
+ *
+ * @param value The string to slugify.
+ */
+function toIdSlug(value: string): string {
+  const slug = value
+    .replaceAll(/[^a-zA-Z0-9]+/g, '-')
+    .replaceAll(/^-+|-+$/g, '');
+
+  return slug.length === 0 ? 'x' : slug;
```
ST-DDT: IMO this can never be empty as method, parameter and value aren't empty.

If they are empty, then that is likely an error.

scripts/apidocs/processing/type.ts
```diff
@@ -11,6 +11,11 @@ export type RawApiDocsType =
 interface RawApiDocsBaseType {
   type: string;
   text: string;
+  /**
+   * The description of this type, e.g. the JSDoc of the enum member backing a
+   * shadow type value. Rendered as hoverable/clickable popover text in the docs.
+   */
+  description?: string;
```
ST-DDT, line 19: Is this the raw data or already html?

scripts/apidocs/processing/type.ts
```diff
@@ -194,6 +199,104 @@ export function isRangeType(type: Type): boolean {
...
+
+/**
+ * Attaches the given per-value descriptions to the matching members of a type.
+ *
+ * @param type The type to enrich (mutated in place).
+ * @param descriptions The per-value descriptions, see {@link getShadowTypeDescriptions}.
+ *
+ * @returns The enriched type, for convenience.
+ */
+export function attachShadowTypeDescriptions(
```
ST-DDT, line 288: Instead of manually attaching the description after the fact.
Is it possible to get access to it when the union type is created?

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -45,3 +45,51 @@ td.deprecated {
   text-decoration: line-through;
 }
 </style>
+
+<!-- Unscoped: targets markup injected via v-html (shadow type popovers). -->
```
ST-DDT: Have you tried scoped with `.type :deep(.shadow-type-value)`?

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -45,3 +45,51 @@ td.deprecated {
...
+<style>
+.shadow-type-value {
+  font-family: var(--vp-font-family-mono);
```
ST-DDT, line 49: If we go for mono here, then it should likely apply to all type texts.

scripts/apidocs/processing/type.ts
```diff
@@ -11,6 +11,12 @@ export type RawApiDocsType =
 interface RawApiDocsBaseType {
   type: string;
   text: string;
+  /**
+   * The raw (non-HTML) description of this type, e.g. the JSDoc of the enum
+   * member backing a shadow type value. Escaped and rendered as hoverable/
```
ST-DDT:
```suggestion
   * member backing a shadow type value. To be rendered as hoverable/
```

scripts/apidocs/processing/type.ts
ST-DDT: AFAICT you didn't change anything here. Did you mark this as resolved this by accident?

docs/.vitepress/components/api-docs/method-parameters.vue
```diff
@@ -44,4 +44,48 @@ const { parameters } = defineProps<{ parameters: ApiDocsMethodParameter[] }>();
 td.deprecated {
   text-decoration: line-through;
 }
+
+.type {
+  font-family: var(--vp-font-family-mono);
```
ST-DDT, line 49: This (font-family) looks quite different from before (pre PR).
Which isn't nessesaryly good or bad, I just want the other reviewers to pay attention to this.

| Old | New |
| --- | --- |
| Image | Image |

ST-DDT (commented):
This is a good addition. I'm not sure about the font-family and how the descriptions gets added to the types, but the rest looks good to me.

During review I noticed, that this applies to LengthStrategy and SexType, but not ColorFormats. Which made me think what would be better and consistent UX.
Keep the named type, then show the values+descriptions on hover/popover or explode the enum/union types (as is).

Image

docs/.vitepress/components/api-docs/method-parameters.vue
Shinigami92: I like to have types as monospace, so I made that decision and accepted your suggestion from above

scripts/apidocs/processing/type.ts
Shinigami92: will look into the claude session and check what it did... tbh I just would like to get this over the table and further improvements can be made by other follow up PRs. This PR is right now doing so much, but the requested bringing-back of the descriptions is done.

Give me some minutes, I'm working on 3 projects in parallel right now...

docs/.vitepress/components/api-docs/method-parameters.vue
ST-DDT: We should replace `{ ... }` with `{ … }` (explicit ellipsis) then though.

scripts/apidocs/processing/type.ts
Shinigami92: 🤖 
the description lives on a different source than the union, which is why it's a second pass.

`getTypeText` builds the union purely from the resolved semantic `Type`. For a shadow type like `` LengthStrategyType = `${LengthStrategy}` ``, TypeScript has already collapsed the template literal into a plain `'fail' | 'shortest' | …` string-literal union — the link back to the `LengthStrategy` enum members and their JSDoc is gone from the `Type`. `getShadowTypeDescriptions` recovers that JSDoc by walking the syntactic AST node (alias decl → enum decl → member JSDoc), which is unrelated to the `Type` recursion.

It *is* technically possible to do it at union-creation time: inside `getTypeText` I could pull the alias declaration off `type.getAliasSymbol()?.getDeclarations()?.[0]` and enrich inline (no extra parameter needed). The cost is coupling the generic core renderer to enum-backed aliases — the current two-pass keeps `getTypeText` type-only and isolates all shadow handling to two small functions.

Since this PR is already doing a lot and the requested bringing-back of the descriptions is done, I'd prefer to leave the split as-is and revisit inlining in a follow-up PR if we decide we want it.
/🤖

ST-DDT (dismissed):
Since the remaining things will be handled in a separate PR, I approve this.

- [Discussion](https://github.com/faker-js/faker/pull/3940#pullrequestreview-4728223330): Do we show the enum type values directly or do we show an popover with more details. aka do we give the user a hint that the wrapper type + actual enum exist or should they read that from the signature?
- [Inline](https://github.com/faker-js/faker/pull/3940#discussion_r3606648127): Read the enum value description at type creation time, instead of retroactively attaching it.

ST-DDT:

ST-DDT (approved):
-

xDivisionByZerox (approved):
Finally got around to review this. I'm approving but there is one thing that bothers me inside this PR. It introduces two dedicated "features":

1. The LengthStrategy type that is obvious from the PR title (and thus the resulting commit description)
2. The ability for the documentation to handle popovers ("feature" in a broader sense)

The length strategy part was easy enough to review and I could have given my approval way earlier. The "hard" part to review was the change for the documentation, which put me off of reviewing this PR multiple times (since I'm not a vue person and out doc generation has a lot of knowledge depth in general). 

I would appreciate if we could split those kinds of topics in the future to make it easier for all participants. For me/us the reviewer/s and you the person wanting their change approved/merged.

PR #3943, opened 2026-07-08
fix(word): pick nearest length for closest strategy

spokodev:
The `closest` resolution strategy picked its offset from the shortest and longest words in the list rather than the lengths nearest to the requested range. When the nearest available length sat between those extremes, the method returned words farther from the target than needed. On the default `en` locale, `faker.word.noun({ length: 17, strategy: 'closest' })` returned words of length 15 or 19 even though length-16 nouns exist.

This derives the offset from the closest length below the range and the closest length above it, so the returned words are the actual nearest ones. Behavior for ranges fully outside the available lengths is unchanged. Adds a unit test covering an interior gap.

### DIFF
--- a/test/modules/word.spec.ts
+++ b/test/modules/word.spec.ts
@@ -106,6 +106,17 @@ describe('word', () => {
       expect(result).toEqual(['very-long', 'almostRight']);
     });
 
+    it('returns the nearest words when the closest length is not the shortest or longest word', () => {
+      // Lengths present: 1, 4, 5. The closest length to 3 is 4 (distance 1),
+      // not the shortest (1) or longest (5) word (distance 2).
+      const result = filterWordListByLength({
+        wordList: ['a', 'wxyz', 'abcde'],
+        length: 3,
+        strategy: 'closest',
+      });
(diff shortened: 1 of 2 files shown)


test/modules/word.spec.ts
```diff
@@ -106,6 +106,17 @@ describe('word', () => {
       expect(result).toEqual(['very-long', 'almostRight']);
     });
 
+    it('returns the nearest words when the closest length is not the shortest or longest word', () => {
+      // Lengths present: 1, 4, 5. The closest length to 3 is 4 (distance 1),
+      // not the shortest (1) or longest (5) word (distance 2).
+      const result = filterWordListByLength({
+        wordList: ['a', 'wxyz', 'abcde'],
```
ST-DDT, line 113: Maybe replace the random letters with numbers representing the length:

```suggestion
        wordList: ['1', '4444', '55555'],
```

That way the brain can more easily "count" them.

ST-DDT (approved):

Shinigami92 (approved):

PR #3944, opened 2026-07-08
docs(lorem): document the fail default strategy for word

spokodev:
`faker.lorem.word` documents an `any-length` default resolution strategy, returning a word of any length when none matches the requested length. The change that switched the shared word length filter to the `fail` strategy (#3560) targeted only the word module, but `lorem.word` forwards its options to that filter without a strategy, so it started throwing `No words found that match the given length.` The upgrade guide only lists `faker.word.*` methods for this breaking change, and `lorem.word`'s JSDoc still documents `any-length`.

This passes `any-length` as the default while still letting a caller-provided strategy override it, restoring the documented behavior. Adds a regression test.


src/modules/lorem/module.ts
ST-DDT: IMO this default is incorrect.

ST-DDT: Same here.

ST-DDT (changes requested):
Nicely spottled.

AFAICT the implementation was changed as part of a major change.
While not exlicitly in the breaking changes list, it has been like this for the entirety of v10, so IMO we should change the jsdocs. Not the implementation.
I consider the current implementation to be correct.

spokodev:
Makes sense — reverted the implementation change and fixed the `word()` JSDoc instead (it documented the default as `any-length`, when it's actually `fail`). The PR is docs-only now.

src/modules/lorem/module.ts
ST-DDT: Can you remove this default as well?

xDivisionByZerox: I agreed. There is no default for `length`. If it is not provided, a completely random word from the given locale list.

spokodev: Removed the `@default 1` on `length`: there's no default, and if it's omitted a word of any length is returned, so I documented that instead. Left `@default 'fail'` on `strategy`, since that one is the real default (`strategy = 'fail'` in `filterWordListByLength`, which the new test covers).

ST-DDT (dismissed):

Shinigami92 (dismissed):

ST-DDT (approved):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #3945, opened 2026-07-08
docs(helpers): clarify fake() does not execute javascript

Shinigami92:
Follow-up to the published security advisory [GHSA-qxc2-j82w-r537](https://github.com/faker-js/faker/security/advisories/GHSA-qxc2-j82w-r537) (arbitrary code execution via `helpers.fake()`).

The underlying vulnerability was already fixed in `10.5.0` by #3852, which made `resolveProperty()` invoke functions and recurse instead of reading a property off the function object — so a `.constructor` chain can never reach the `Function` constructor. That fix was shipped as an incidental change, was never framed as security, and had **no regression test** guarding it. It also left behind JSDoc that overstated the guarantee.

This PR:

- **Corrects the JSDoc claims** that the advisory highlighted:
  - `helpers.fake()` (all three overloads) previously stated *"It is also NOT possible to use any non-faker methods or plain javascript in such patterns."* — reworded to describe the actual behaviour (the pattern is not evaluated as JavaScript; params are parsed as JSON or plain strings) **and** to add a caveat that the pattern should still be treated as trusted input.
  - The *"string coming from a user"* wording is replaced with *"a template authored by a developer or stored in a database or file"*, so the docs no longer imply `fake()` is safe to feed unsanitized external input.
  - `fakeEval()` JSDoc *"It is NOT possible to access any values not passed as entrypoints."* → reworded to the guarantee we actually enforce: no arbitrary JavaScript execution.
- **Adds a regression test** (`test/modules/helpers-eval.spec.ts`) that reproduces the original PoC and several `.constructor`-chain escalations through both `fakeEval()` and `helpers.fake()`, asserting that no injected global side effect ever runs and that the `Function` constructor is never returned. The test was verified to **fail** against the pre-#3852 (vulnerable) implementation and **pass** against current code.

No runtime behaviour changes — documentation and test only.

## Notes

- The "treat as trusted input" caveat is deliberately conservative: faker makes no guarantees about the cost of arbitrary, user-controlled parameters, so callers should sanitize external input regardless.

### DIFF
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1201,8 +1201,10 @@ export class HelpersModule extends SimpleHelpersModule {
    * ```
    *
    * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
-   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
-   * as trusted input and not pass unsanitized values from untrusted or external sources.
+   * parameters are parsed as JSON or plain strings. Nevertheless, it is possible for certain
+   * maliciously crafted patterns to use large amounts of memory or CPU time, so the pattern
+   * itself must always be from trusted input. Do not evaluate patterns provided by untrusted
+   * user input or external sources.
    *
    * @param pattern The pattern string that will get interpolated.
    *
@@ -1255,8 +1257,10 @@ export class HelpersModule extends SimpleHelpersModule {
    * ```
    *
    * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
-   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
-   * as trusted input and not pass unsanitized values from untrusted or external sources.
+   * parameters are parsed as JSON or plain strings. Nevertheless, it is possible for certain
+   * maliciously crafted patterns to use large amounts of memory or CPU time, so the pattern
+   * itself must always be from trusted input. Do not evaluate patterns provided by untrusted
+   * user input or external sources.
    *
    * @param patterns The array to select a pattern from, that will then get interpolated. Must not be empty.
    *
@@ -1300,8 +1304,10 @@ export class HelpersModule extends SimpleHelpersModule {
    * ```
    *
    * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
-   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
-   * as trusted input and not pass unsanitized values from untrusted or external sources.
+   * parameters are parsed as JSON or plain strings. Nevertheless, it is possible for certain
+   * maliciously crafted patterns to use large amounts of memory or CPU time, so the pattern
+   * itself must always be from trusted input. Do not evaluate patterns provided by untrusted
+   * user input or external sources.
    *
    * @param pattern The pattern string that will get interpolated. If an array is passed, a random element will be picked and interpolated.
    *
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1177,7 +1177,7 @@ export class HelpersModule extends SimpleHelpersModule {
    * e.g. ``const address = `${faker.location.zipCode()} ${faker.location.city()}`;``
    *
    * This method is useful if you have to build a random string from a static, non-executable source
-   * (e.g. string coming from a user, stored in a database or a file).
+   * that you control (e.g. a template authored by a developer or stored in a database or file).
    *
    * It checks the given string for placeholders and replaces them by calling faker methods:
    *
@@ -1200,7 +1200,9 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
    * ```
    *
-   * It is also NOT possible to use any non-faker methods or plain javascript in such patterns.
+   * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
+   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
+   * as trusted input and not pass unsanitized values from untrusted or external sources.
    *
    * @param pattern The pattern string that will get interpolated.
    *
@@ -1226,7 +1228,7 @@ export class HelpersModule extends SimpleHelpersModule {
    * e.g. ``const address = `${faker.location.zipCode()} ${faker.location.city()}`;``
    *
    * This method is useful if you have to build a random string from a static, non-executable source
-   * (e.g. string coming from a user, stored in a database or a file).
+   * that you control (e.g. a template authored by a developer or stored in a database or file).
    *
    * It checks the given string for placeholders and replaces them by calling faker methods:
    *
@@ -1252,7 +1254,9 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake(['Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.']);
    * ```
    *
-   * It is also NOT possible to use any non-faker methods or plain javascript in such patterns.
+   * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
+   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
+   * as trusted input and not pass unsanitized values from untrusted or external sources.
    *
    * @param patterns The array to select a pattern from, that will then get interpolated. Must not be empty.
    *
@@ -1272,7 +1276,7 @@ export class HelpersModule extends SimpleHelpersModule {
    * e.g. ``const address = `${faker.location.zipCode()} ${faker.location.city()}`;``
    *
    * This method is useful if you have to build a random string from a static, non-executable source
-   * (e.g. string coming from a user, stored in a database or a file).
+   * that you control (e.g. a template authored by a developer or stored in a database or file).
    *
    * It checks the given string for placeholders and replaces them by calling faker methods:
    *
@@ -1295,7 +1299,9 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
    * ```
    *
-   * It is also NOT possible to use any non-faker methods or plain javascript in such patterns.
+   * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
+   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
+   * as trusted input and not pass unsanitized values from untrusted or external sources.
    *
    * @param pattern The pattern string that will get interpolated. If an array is passed, a random element will be picked and interpolated.
    *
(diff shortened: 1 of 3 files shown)


ST-DDT (dismissed):

src/modules/helpers/module.ts
```diff
@@ -1200,7 +1200,9 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
    * ```
    *
-   * It is also NOT possible to use any non-faker methods or plain javascript in such patterns.
+   * The pattern is not evaluated as JavaScript: only faker methods can be called, and any
+   * parameters are parsed as JSON or plain strings. Nevertheless, you should treat the pattern
+   * as trusted input and not pass unsanitized values from untrusted or external sources.
```
matthewmayer: i find "treat the pattern as trusted input" a little confusing. i think we could be more explicit.

> The pattern is not evaluated as JavaScript: only faker methods can be called, and any parameters are parsed as JSON or plain strings. Nevertheless, it is possible for certain maliciously crafted patterns to use large amounts of memory or CPU time, so the pattern itself must always be from trusted input. Do not evaluate patterns provided by untrusted user input or external sources."

ST-DDT (dismissed):

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,60 @@ describe('fakeEval()', () => {
...
+        for (const attack of attacks) {
+          let result: unknown;
+          try {
+            result = fakeEval(attack, faker);
+          } catch (error) {
+            // Rejecting the expression is a safe outcome, but it must fail with the library's own error type, never a raw JS engine error.
+            expect(error, attack).toBeInstanceOf(FakerError);
+          }
+
+          expect(globalWithFlag.__fakerPwned, attack).toBe(false);
```
xDivisionByZerox: All of the attacks do not change the flag, do they? 
I've testet this behavior in v10.4.0 and 10.1.0.

<details>
<summary> Logs from test </summary>

```txt
Attack: evil.constructor('return 1')
  { resultFakeEval: undefined }
  { globalProperty: false }

--------------------------

Attack: evil().constructor('return 1')
  {
    error: FakerError: Cannot resolve expression 'evil().constructor('return 1')'
        at fakeEval (/workspaces/faker/src/modules/helpers/eval.ts:100:11)
        at <anonymous> (/workspaces/faker/_local/_test.ts:24:23)
  }
  { globalProperty: false }

--------------------------

Attack: evil.constructor.constructor('globalThis.__fakerPwned = true')
  { resultFakeEval: undefined }
  { globalProperty: false }

--------------------------

Attack: evil.constructor.constructor('globalThis.__fakerPwned = true')()
  {
    error: SyntaxError: Unexpected token ')'
        at Function (<anonymous>)
        at <anonymous> (/workspaces/faker/src/modules/helpers/eval.ts:136:42)
        at Array.map (<anonymous>)
        at evalProcessFunction (/workspaces/faker/src/modules/helpers/eval.ts:135:17)
        at fakeEval (/workspaces/faker/src/modules/helpers/eval.ts:84:26)
        at <anonymous> (/workspaces/faker/_local/_test.ts:24:23)
  }
  { globalProperty: false }

--------------------------

Attack: string.constructor.constructor('globalThis.__fakerPwned = true')()
  {
    error: FakerError: Cannot resolve expression 'string.constructor.constructor('globalThis.__fakerPwned = true')()'
        at fakeEval (/workspaces/faker/src/modules/helpers/eval.ts:100:11)
        at <anonymous> (/workspaces/faker/_local/_test.ts:24:23)
  }
  { globalProperty: false }
```

</details>

Please check if my test setup has an error:

<details>
<summary> test file in `_local/_test.ts` </summary>

```ts
import { fakeEval } from '../src/modules/helpers/eval';

const attacks = [
  "evil.constructor('return 1')",
  "evil().constructor('return 1')",
  "evil.constructor.constructor('globalThis.__fakerPwned = true')",
  "evil.constructor.constructor('globalThis.__fakerPwned = true')()",
  "string.constructor.constructor('globalThis.__fakerPwned = true')()",
];

const noop = (): void => undefined;
const functionReturningFunction = (): (() => void) => noop;

(globalThis as any).__fakerPwned = false;
// A hostile definition resolving to a function returning a function, mirroring the original proof-of-concept.
(faker.rawDefinitions as Record<string, unknown>).evil =
  functionReturningFunction;

for (const atk of attacks) {
  console.group('Attack: ' + atk);
  try {
    console.log({
      resultFakeEval: fakeEval(atk, faker),
    });
  } catch (e) {
    console.log({ error: e });
  }

  console.log({ globalProperty: (globalThis as any).__fakerPwned });
  console.groupEnd();

  (globalThis as any).__fakerPwned = false;

  console.log();
  console.log('--------------------------');
  console.log();
}
```

</details>

If thats actually the case, the test is not adding any value.

ST-DDT: Diff to fix:

````ts
const attacks = [
  'evil.constructor(return 1)',
  'evil().constructor(return 1)',
  'evil.constructor.constructor(globalThis.__fakerPwned = true)',
  'evil.constructor.constructor(globalThis.__fakerPwned = true)()',
  'string.constructor.constructor(globalThis.__fakerPwned = true)()',
];
````

Or

````ts
const attacks = [
  'evil.constructor("return 1")',
  'evil().constructor("return 1")',
  'evil.constructor.constructor("globalThis.__fakerPwned = true")',
  'evil.constructor.constructor("globalThis.__fakerPwned = true")()',
  'string.constructor.constructor("globalThis.__fakerPwned = true")()',
];
````

File: test/modules/helpers-eval.spec.ts [hunk 3]
```diff
@@ -167,4 +172,62 @@ describe('fakeEval()', () => {
...
+      globalWithFlag.__fakerPwned = false;
+      // A hostile definition resolving to a function returning a function, mirroring the original proof-of-concept.
+      (faker.rawDefinitions as Record<string, unknown>).evil =
+        functionReturningFunction;
+
+      try {
+        for (const attack of attacks) {
+          let result: unknown;
+          try {
+            result = fakeEval(attack, faker);
```
ST-DDT: I think we expect all of them to fail with an error.

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,62 @@ describe('fakeEval()', () => {
...
+      'evil().constructor(return 1)',
+      'evil.constructor.constructor(globalThis.__fakerPwned = true)',
+      'evil.constructor.constructor(globalThis.__fakerPwned = true)()',
+      'string.constructor.constructor(globalThis.__fakerPwned = true)()',
+    ];
+
+    it('via fakeEval()', () => {
+      globalWithFlag.__fakerPwned = false;
+      // A hostile definition resolving to a function returning a function, mirroring the original proof-of-concept.
+      (faker.rawDefinitions as Record<string, unknown>).evil =
```
ST-DDT: You could drop the cast and just use

`faker.definitions.raw.do = {evil: () => () => {}}`

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,62 @@ describe('fakeEval()', () => {
...
   });
+
+  // Regression test for GHSA-qxc2-j82w-r537: fake()/fakeEval() must never evaluate the expression as JavaScript.
+  // In particular, a (hostile) definition that resolves to a function must not expose the `Function` constructor via a `.constructor` chain, which would allow arbitrary code execution.
+  // Functions are always invoked (never property-accessed), so the constructor chain can never reach `Function`.
+  describe('does not allow arbitrary code execution (GHSA-qxc2-j82w-r537)', () => {
+    const globalWithFlag = globalThis as Record<string, unknown>;
+
+    const attacks = [
+      'evil.constructor(return 1)',
```
ST-DDT: Maybe use the pwned global here as well, since with a simple return you dont notice if it gets executed and then throws.

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,62 @@ describe('fakeEval()', () => {
...
+    ];
+
+    it('via fakeEval()', () => {
+      globalWithFlag.__fakerPwned = false;
+      // A hostile definition resolving to a function returning a function, mirroring the original proof-of-concept.
+      (faker.rawDefinitions as Record<string, unknown>).evil =
+        functionReturningFunction;
+
+      try {
+        for (const attack of attacks) {
```
ST-DDT: Maybe convert to it.each

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,62 @@ describe('fakeEval()', () => {
...
+            result = fakeEval(attack, faker);
+          } catch (error) {
+            // Rejecting the expression is a safe outcome, but it must fail with the library's own error type, never a raw JS engine error.
+            expect(error, attack).toBeInstanceOf(FakerError);
+          }
+
+          expect(globalWithFlag.__fakerPwned, attack).toBe(false);
+          expect(result, attack).not.toBe(Function);
+        }
+      } finally {
```
ST-DDT: Maybe convert to afterEach

test/modules/helpers-eval.spec.ts
```diff
@@ -167,4 +172,46 @@ describe('fakeEval()', () => {
...
+      expect(() => fakeEval(attack, faker)).toThrow(FakerError);
+      expect(globalWithFlag.__fakerPwned, attack).toBe(false);
+    });
+
+    it('rejects the proof-of-concept via helpers.fake()', () => {
+      expect(() =>
+        faker.helpers.fake(
+          '{{do.evil.constructor.constructor(globalThis.__fakerPwned = true)()}}'
+        )
+      ).toThrow(FakerError);
```
ST-DDT: Maybe repeat the same checks/run here as above:

````ts
expect(() => faker.helpers.fake("{{"+attack+"}}")).toThrow(FakerError);
````

**Review by ST-DDT — DISMISSED** [review]

File: test/modules/helpers-eval.spec.ts [hunk 9]
```diff
@@ -167,4 +172,42 @@ describe('fakeEval()', () => {
...
+
+    // Every payload is an observable side effect (setting the global flag), so we
+    // notice if it ever executes instead of a silent expression like `return 1`.
+    const attacks = [
+      'do.evil.constructor(globalThis.__fakerPwned = true)()',
+      'do.evil().constructor(globalThis.__fakerPwned = true)()',
+      'do.evil.constructor.constructor(globalThis.__fakerPwned = true)',
+      'do.evil.constructor.constructor(globalThis.__fakerPwned = true)()',
+      'string.constructor.constructor(globalThis.__fakerPwned = true)()',
+    ];
```
xDivisionByZerox, line 190: Just to clarify: the 2nd and 5th attack would not have worked in versions where GHSA-qxc2-j82w-r537 was active. But I guess this is just for exhausting all possible variants an attacker could think of? If that's the case, I'm fine with keeping them. If the goal was to explicitly have regression tests for working attack patterns, I'd be in favor of removing them.

xDivisionByZerox (approved):

test/modules/helpers-eval.spec.ts
Shinigami92: That was for exhausting all possible variants. I think this is good that way.

ST-DDT (approved):

Issue #768, opened 2022-04-03
Reimplement internet.password() with more generation options


ST-DDT:
### Clear and concise description of the problem

The passwords generated by internet.password() are not really secure.

### Suggested solution

It would be nice if there would be various generation ways one can choose from.

- secure (at least one of each: alpha, ALPHA, 0-9, special)
- memorable (vowel and consonant alternating)
- simple (any of alpha, ALPHA, 0-9)
- words (a combination of random words)

### Alternative

Remove the default pattern, that severely limits the password security

xDivisionByZerox:
This looks interesting, I might provide a PR for this in the near future.

xDivisionByZerox:
The suggestions you made, would they all be implemented in a separate function? Or would there only be a single parameter like `faker.internet.password({ mode: 'secure' })`?

ST-DDT:
IMO it should be accessed via a single method with options.
Whether we use internal helper methods for each generation strategy doesn't really matter to me.

xDivisionByZerox:
For parameter options, I would like something like https://passwordsgenerator.net/plus is doing. This would not be as simple with a single method and a `mode`.

The signature could also look something like this:
```ts
type PasswordMode = 'secure' | 'memorable' | 'simple';

password(
  opt:
    | PasswordMode
    | {
        mode?: PasswordMode;
      }
    | {
        length?: number;
        includeSymbols?: boolean;
        includeNumbers?: boolean;
        includeLowercaseChars?: boolean;
        includeUppercaseChars?: boolean;
        noSimilarChars?: boolean;
        noDuplicateChars?: boolean;
        noSequentialChars?: boolean;
        bannedChars?: Array<string>;
      }
) {
  // implementation
}
```

xDivisionByZerox:
Based on my last comment:
If you provide all the groups (symbol, lower, upper, number) included and set a length less than 4, what would the expected behavior be? Throwing an error? Returning the string with one of each group?

ST-DDT:
This depends on the description you are going to add to the parameters.
IMO include != required -> so you can have even long passwords without any of them. It's just a matter of chance.
If you are going for require, then I would throw.

xDivisionByZerox:
I have read the include properties as required ("at least one of") since if you don't want to have any specific chars (don't include them at all) you can provide them via the banned chars array. But in this case, it would be best to name the properties appropriately (include => require).

xDivisionByZerox:
[...]

PR #3946, opened 2026-07-08
fix(internet): generate long passwords without stack overflow

Shinigami92:
`internet.password()` generated the password by recursing once per character. For large `length` values this exhausted the JavaScript call stack and threw a raw `RangeError: Maximum call stack size exceeded` (reproducible from around `length: 5000` upwards).

This PR rewrites the internal generator as a plain `while` loop, so the length is bounded only by normal time/memory rather than the call-stack depth.

## Changes

- `internet.password()`: replaced the recursive `_password()` helper with an iterative loop. The character generation, `memorable` handling, `pattern` matching, and `prefix` behaviour are unchanged.
- Added a regression test that generates a 100,000-character password and asserts it succeeds (this threw a `RangeError` before).

## Notes

- No behavioural change: the random-number-generator call order and count are identical, so all seeded snapshots are unchanged.

### DIFF
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -895,7 +895,6 @@ export class InternetModule extends ModuleBase {
       prefix = '',
     } = options;
 
-    // A plain loop (rather than recursion) so that large lengths cannot overflow the call stack.
     let currentPattern = pattern;
     let result = prefix;
     while (result.length < length) {
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -887,32 +887,6 @@ export class InternetModule extends ModuleBase {
      */
     const vowel = /[aeiouAEIOU]$/;
     const consonant = /[bcdfghjklmnpqrstvwxyzBCDFGHJKLMNPQRSTVWXYZ]$/;
-    const _password = (
-      length: number,
-      memorable: boolean,
-      pattern: RegExp,
-      prefix: string
-    ): string => {
-      if (prefix.length >= length) {
-        return prefix;
-      }
-
-      if (memorable) {
-        pattern = consonant.test(prefix) ? vowel : consonant;
-      }
-
-      const n = this.faker.number.int(94) + 33;
-      let char = String.fromCodePoint(n);
-      if (memorable) {
-        char = char.toLowerCase();
-      }
-
-      if (!pattern.test(char)) {
-        return _password(length, memorable, pattern, prefix);
-      }
-
-      return _password(length, memorable, pattern, prefix + char);
-    };
 
     const {
       length = 15,
@@ -921,7 +895,27 @@ export class InternetModule extends ModuleBase {
       prefix = '',
     } = options;
 
-    return _password(length, memorable, pattern, prefix);
+    // A plain loop (rather than recursion) so that large lengths cannot
+    // overflow the call stack.
+    let currentPattern = pattern;
+    let result = prefix;
+    while (result.length < length) {
+      if (memorable) {
+        currentPattern = consonant.test(result) ? vowel : consonant;
+      }
+
+      const n = this.faker.number.int(94) + 33;
+      let char = String.fromCodePoint(n);
+      if (memorable) {
+        char = char.toLowerCase();
+      }
+
+      if (currentPattern.test(char)) {
+        result += char;
+      }
+    }
+
+    return result;
   }
 
   /**
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -895,8 +895,7 @@ export class InternetModule extends ModuleBase {
       prefix = '',
     } = options;
 
-    // A plain loop (rather than recursion) so that large lengths cannot
-    // overflow the call stack.
+    // A plain loop (rather than recursion) so that large lengths cannot overflow the call stack.
     let currentPattern = pattern;
     let result = prefix;
     while (result.length < length) {
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -897,6 +897,7 @@ export class InternetModule extends ModuleBase {
 
     let currentPattern = pattern;
     let result = prefix;
+    // TODO @Shinigami92 2026-07-09: This loop never terminates if the pattern can never match a generated char (e.g. `/°/`), blocking the event loop. To be resolved by the password rewrite in https://github.com/faker-js/faker/issues/768.
     while (result.length < length) {
       if (memorable) {
         currentPattern = consonant.test(result) ? vowel : consonant;
--- a/test/modules/internet.spec.ts
+++ b/test/modules/internet.spec.ts
@@ -872,11 +872,8 @@ describe('internet', () => {
         it('should generate long passwords without overflowing the call stack', () => {
           // Regression test: password generation used to recurse once per character, throwing a RangeError for lengths of ~5000 and above.
           const length = 100_000;
-          const password = faker.internet.password({ length });
 
-          expect(password).toBeTypeOf('string');
-          expect(password).toHaveLength(length);
-          expect(password).toMatch(/^\w+$/);
+          expect(() => faker.internet.password({ length })).not.toThrow();
         });
       });
 
--- a/test/modules/internet.spec.ts
+++ b/test/modules/internet.spec.ts
@@ -870,8 +870,7 @@ describe('internet', () => {
         });
 
         it('should generate long passwords without overflowing the call stack', () => {
-          // Regression test: password generation used to recurse once per
-          // character, throwing a RangeError for lengths of ~5000 and above.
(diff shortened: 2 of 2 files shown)


test/modules/internet.spec.ts
```diff
@@ -868,6 +868,17 @@ describe('internet', () => {
           expect(password).toStartWith('a!G6');
           expect(password).toSatisfy(isStrongPassword);
         });
+
+        it('should generate long passwords without overflowing the call stack', () => {
+          // Regression test: password generation used to recurse once per
+          // character, throwing a RangeError for lengths of ~5000 and above.
```
Shinigami92: **suggestion:** 

```suggestion
          // Regression test: password generation used to recurse once per character, throwing a RangeError for lengths of ~5000 and above.
```

src/modules/internet/module.ts
```diff
@@ -887,18 +887,21 @@ export class InternetModule extends ModuleBase {
...
 
+    const {
+      length = 15,
+      memorable = false,
+      pattern = /\w/,
+      prefix = '',
+    } = options;
+
+    // A plain loop (rather than recursion) so that large lengths cannot
+    // overflow the call stack.
```
Shinigami92: **suggestion:** 

```suggestion
    // A plain loop (rather than recursion) so that large lengths cannot overflow the call stack.
```

src/modules/internet/module.ts
```diff
@@ -887,18 +887,20 @@ export class InternetModule extends ModuleBase {
...
-      }
 
+    const {
+      length = 15,
+      memorable = false,
+      pattern = /\w/,
+      prefix = '',
+    } = options;
+
+    // A plain loop (rather than recursion) so that large lengths cannot overflow the call stack.
```
ST-DDT: IMO we don't need this comment.

ST-DDT (dismissed):

ST-DDT (dismissed):

test/modules/internet.spec.ts
```diff
@@ -868,6 +868,16 @@ describe('internet', () => {
...
+
+        it('should generate long passwords without overflowing the call stack', () => {
+          // Regression test: password generation used to recurse once per character, throwing a RangeError for lengths of ~5000 and above.
+          const length = 100_000;
+          const password = faker.internet.password({ length });
+
+          expect(password).toBeTypeOf('string');
+          expect(password).toHaveLength(length);
+          expect(password).toMatch(/^\w+$/);
+        });
```
xDivisionByZerox, line 877: Since our only concern for this test is that no range error occurs, wouldn't it be better to actually check for that?


```suggestion
        it('should generate long passwords without overflowing the call stack', () => {
          // Regression test: password generation used to recurse once per character, throwing a RangeError for lengths of ~5000 and above.
          const length = 100_000;
          expect(() => faker.internet.password({ length })).not.toThrow(RangeError);
        });
```

src/modules/internet/module.ts
```diff
@@ -887,18 +887,19 @@ export class InternetModule extends ModuleBase {
...
+    const {
+      length = 15,
+      memorable = false,
+      pattern = /\w/,
+      prefix = '',
+    } = options;
+
+    let currentPattern = pattern;
+    let result = prefix;
+    while (result.length < length) {
```
xDivisionByZerox, line 901: While the `while` loop fixes the problem with the infinite stack overflow, do we not run into the problem that the loop might never resolve? 

```
// expect password to contain the degree sign (°)  which the underlying algorythm will never generate
faker.internet.password({ length: 100_000, pattern: /°/ })
```

Technically, that was true for the previous implementation as well but was "caught" by throwing a range error. Now it will truly block the event loop - which might be even worse IMO.

_You can test this in the deploy preview. After enabling faker and pasting my POC in the console, the entire faker website will become unresponsive because the password function is still trying to generate a password._

xDivisionByZerox (changes requested):

test/modules/internet.spec.ts
ST-DDT: I would probably check for any thrown exception.

src/modules/internet/module.ts
ST-DDT: We have a separate issue to re-implement the password method entirely:

- #768

xDivisionByZerox:
This PR was created due to a security advisory we received. Wouldn't it be proper etiquette to:

1. Label this PR with https://github.com/faker-js/faker/labels/c%3A%20security?
2. Mention the vulnerability report (even if not visible for the public right now)?
3. Give credit to the author of the report?

I never had to work on an actual "live" vulnerability/disclosure process myself so these are honest questions I'm asking.

Shinigami92:
> This PR was created due to a security advisory we received. Wouldn't it be proper etiquette to:
> 
> 1. Label this PR with 
>    
>    [
>      c: security](https://github.com/faker-js/faker/labels/c%3A%20security)
>      Indicates a vulnerability or a potentially (security) risky change.
>    ?
> 2. Mention the vulnerability report (even if not visible for the public right now)?
> 3. Give credit to the author of the report?
> 
> I never had to work on an actual "live" vulnerability/disclosure process myself so these are honest questions I'm asking.

Actually I omitted all this by design, so e.g. AIs listening don't get that we are actually working on a fix for that 🫠 Now you exposed something that was privately gated behind the scenes.

xDivisionByZerox:
>  Now you exposed something that was privately gated behind the scenes.

The existence of this PR is sufficient information for anybody that is reading through the repo (your AI friends). I am not exposing anything that is not clear by looking at the patch you provided here itself. 🤷‍♀️

src/modules/internet/module.ts
xDivisionByZerox: That does not answer the question I'm having. The new implementation is worse IMO because you can actually block the event loop.  I do not understand how this isn't worse for every scenario in which the previous potential uncatched RangeError would have terminated the process.

test/modules/internet.spec.ts
xDivisionByZerox: Since the comment is explicitly mentioning I thought we might want to be strict, but a general catch works as well for me.

ST-DDT: Theoretically a good idea, but we basically want:
`not throw RangeError && not throw any other error` and then the later part swallows the first part.

src/modules/internet/module.ts
ST-DDT: Lets call both the released and this updated implementation doomed.^^

Shinigami92:
> >  Now you exposed something that was privately gated behind the scenes.
> 
> 
> 
> The existence of this PR is sufficient information for anybody that is reading through the repo (your AI friends). I am not exposing anything that is not clear by looking at the patch you provided here itself. 🤷‍♀️ 

Sorry, I didn't want to be mean or toxic but a bit more funny while also kicking a thought process on your side.

For me it makes a difference if someone can search GitHub repos via api by security label and find unfixed bugs, or if they scan the repository with AI, because in the later one they at least took the effort (at least now) to spend money on the investigation.

But I think we can clearly all agree that open source code is open and therefore scannable. I just saw a difference on non obvious vs obvious.

But I'm thankful for the reviews and the hint to e.g. problem with the while loop, I will try to tackle this this evening after work. 🤝

src/modules/internet/module.ts
Shinigami92: Added TODO comment

test/modules/internet.spec.ts
Shinigami92: done

src/modules/internet/module.ts
xDivisionByZerox: I'm fine with moving forward if the majority of the team is in favor of it, BUT I want it to be known that I'm against shipping this implementation over the previous one, as it does not solve the underlying issue at had. It would only bring the satisfaction of "ticking a checkbox" by being able to resolve a concrete reported issue.

xDivisionByZerox (approved):

ST-DDT (commented):
IMO we should deprecate a lot of the legacy options from this method immediately and remove them in v11.0.

Maybe even simplify the implememtation to length only and then expand it if demand exists using the generator type pattern outlined in the issue.

As for the current fix implementation:
The method pools the characters from a very limited set. Instead of verifying that the chosen character matches the pattern, we could pre filter all characters and then pass only the matching to fromCharacters.
If no character match, then that method should throw an error. This will likely also improve the performance of the method since it will no longer attempt to generate banned characters.

````ts
const vowels = [...].filter(regex.test)
const consonants = [...].filter(regex.test)
let useVowels = boolean();
let result = ''
for (i...length) {
  result += fromCharacters(useVowels ? vovels : consonants);
  useVowels = !useVowels;
}
return result;
````

Shinigami92:
This is approved and we agree all that we need to anyway rework this method later on, so the current implementation is also not much worse. I will merge it for now, as it at least solve the recursion problem.

PR #3947, opened 2026-07-08
docs: add object type, expand NumberRange param rows

Shinigami92:
🤖 _this PR was fully driven by Claude Opus 4.8_ 🤖 

Introduces a structured object type to the API-docs pipeline and uses it to restore the per-member parameter rows (`.min`/`.max`) for named range types such as `NumberRange`.

## Background

When a method takes a parameter whose type is an **inline anonymous** object (e.g. `{ min: number; max: number }`), the docs generator expands its members into individual parameter rows (`length.min`, `length.max`, …).

Since `NumberRange` was introduced (#3838) as a **named** `interface`, `isOptionsLikeType()` — which requires `type.isAnonymous()` — returns `false` for it. As a result, standalone range parameters (`length: number | NumberRange`, `wordCount`, …) stopped expanding, and the `@param length.min` / `@param length.max` descriptions no longer rendered. This was noticed on `string.sample` during the review of #3939.

## Changes

- **`scripts/apidocs/processing/type.ts`**
  - Add a structured `RawApiDocsObjectType` (`type: 'object'` with `members`) to the `RawApiDocsType` union, alongside a `newObjectType()` builder (mirrors `newUnionType`/`newSimpleType`; `text` stays `{ min: number; max: number }`).
  - The `NumberRange` case now returns `newObjectType([min, max])` instead of a flat `newSimpleType('{ min: number; max: number }')`.
  - Add an exported `isRangeLikeType(type)` predicate so the range-type detection lives next to the existing type handling.
- **`scripts/apidocs/processing/parameter.ts`**
  - Thread the method's `@param` tags into `processComplexParameter` / `processComplexParameterProperty`.
  - Expand members for range-like types in addition to anonymous options. For range types, each member row's description is sourced from the method-level `@param name.member` tag (keeping the method-specific wording and its `Defaults to …` hint), falling back to the interface member's own JSDoc when no tag is present.
  - The anonymous-options expansion path is unchanged, so other modules (`number`, `date`, `location`, `commerce`, `finance`, …) are unaffected.

## Notes / scope

- Member descriptions intentionally keep the **per-method wording** (e.g. "The minimum length of the string to generate.") rather than the generic interface wording — it reads better for users.
- Nested range members (e.g. `options.length.min` where `length?: NumberRange` sits inside an options object) remain unexpanded; that would require the property processor to recurse and is out of scope here (matches current behavior).
- `NumberOrRange` is not part of this PR. Once this is merged, #3939 will be rebased on top and re-wire `NumberOrRange` to reuse `newObjectType` via `newUnionType([number, <object type>])`.

## Verification

- `pnpm exec vitest run test/scripts/apidocs/` — all apidocs tests pass (incl. `verify-jsdoc-tags`), no type errors.
- `pnpm run generate:api-docs` — `string.sample` / `nanoid` / `symbol`, the `lorem.*` count methods, and `helpers.rangeToNumber` again show their `.min`/`.max` rows with per-method descriptions.
- `pnpm run lint` — clean.

### DIFF
--- a/scripts/apidocs/output/page.ts
+++ b/scripts/apidocs/output/page.ts
@@ -11,6 +11,10 @@ import {
 import { FILE_PATH_API_DOCS } from '../../shared/paths';
 import { toRefreshableCode } from '../../shared/refreshable-code';
 import type { RawApiDocsPage } from '../processing/class';
+import {
+  extractSummaryDefault,
+  stripSummaryDefault,
+} from '../processing/jsdocs';
 import type { RawApiDocsMethod } from '../processing/method';
 import { required } from '../utils/value-checks';
 import { SCRIPT_COMMAND } from './constants';
@@ -159,8 +163,6 @@ async function writePageData(
   );
 }
 
-const defaultCommentRegex = /\s+Defaults to `([^`]+)`\..*/;
-
 async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
   const { name, signatures, source } = method;
   const signatureData = required(signatures.at(-1), 'method signature');
@@ -220,9 +222,7 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
         ...param,
         type: param.type.text,
         default: param.default ?? extractSummaryDefault(param.description),
-        description: await mdToHtml(
-          param.description.replace(defaultCommentRegex, '')
-        ),
+        description: await mdToHtml(stripSummaryDefault(param.description)),
       }))
     ),
     since,
@@ -239,10 +239,6 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
       seeAlsos.map((seeAlso) => mdToHtml(seeAlso, true))
     ),
   };
-}
-
-export function extractSummaryDefault(description: string): string | undefined {
-  return defaultCommentRegex.exec(description)?.[1];
 }
 
 export async function toRefreshFunction(
--- a/scripts/apidocs/output/page.ts
+++ b/scripts/apidocs/output/page.ts
@@ -11,10 +11,6 @@ import {
 import { FILE_PATH_API_DOCS } from '../../shared/paths';
 import { toRefreshableCode } from '../../shared/refreshable-code';
 import type { RawApiDocsPage } from '../processing/class';
-import {
-  extractSummaryDefault,
-  stripSummaryDefault,
-} from '../processing/jsdocs';
 import type { RawApiDocsMethod } from '../processing/method';
 import { required } from '../utils/value-checks';
 import { SCRIPT_COMMAND } from './constants';
@@ -221,8 +217,7 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
       parameters.map(async (param) => ({
         ...param,
         type: param.type.text,
-        default: param.default ?? extractSummaryDefault(param.description),
-        description: await mdToHtml(stripSummaryDefault(param.description)),
+        description: await mdToHtml(param.description),
       }))
     ),
     since,
--- a/scripts/apidocs/processing/parameter.ts
+++ b/scripts/apidocs/processing/parameter.ts
@@ -10,19 +10,21 @@ import type {
 import { exactlyOne, valueForKey } from '../utils/value-checks';
 import { newProcessingError } from './error';
 import {
+  extractSummaryDefault,
   getDefault,
   getDeprecated,
   getDescription,
   getJsDocs,
   getParameterTags,
   getTypeParameterTags,
+  stripSummaryDefault,
 } from './jsdocs';
 import type { RawApiDocsType } from './type';
 import {
   getNameSuffix,
   getTypeText,
   isOptionsLikeType,
-  isRangeLikeType,
+  isRangeType,
 } from './type';
 
 /**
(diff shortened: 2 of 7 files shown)


scripts/apidocs/processing/parameter.ts
```diff
@@ -194,14 +210,21 @@ function processComplexParameter(
...
+) {
   const declaration = exactlyOne(
     parameter.getDeclarations(),
     'property declaration'
   ) as PropertySignature;
   const propertyType = declaration.getType();
   const jsdocs = getJsDocs(declaration);
   const deprecated = getDeprecated(jsdocs);
+  // Prefer the method-level `@param name.member` tag (per-method wording and its embedded default hint),
+  // falling back to the property's own JSDoc.
```
ST-DDT: IMO this comment is confusing as it does not "prefer" and "fallback" + AFAICT it does not use the "embedded default hint".

Maybe something like

> Use `@param name.member` if available, otherwise use the elements jsdocs.

scripts/apidocs/processing/type.ts
```diff
@@ -26,6 +27,16 @@ export interface RawApiDocsUnionType extends RawApiDocsBaseType {
...
 
+export interface RawApiDocsObjectMember {
+  name: string;
+  type: RawApiDocsType;
+}
+
+export interface RawApiDocsObjectType extends RawApiDocsBaseType {
+  type: 'object';
+  members: RawApiDocsObjectMember[];
+}
```
ST-DDT, line 33: Can you swap these? `RawApiDocsObjectType` is the "entrypoint".

scripts/apidocs/processing/type.ts
```diff
@@ -169,6 +183,17 @@ export function isOptionsLikeType(type: Type): boolean {
   );
 }
 
+/**
+ * Checks whether the given type is a named range type (e.g. `NumberRange`)
+ * whose members should be expanded into individual parameter rows in the docs.
+ *
+ * @param type The type to check.
+ */
+export function isRangeLikeType(type: Type): boolean {
```
ST-DDT: You actually check for specific types not "like" types.

```suggestion
export function isRangeType(type: Type): boolean {
```

test/scripts/apidocs/__snapshots__/method.spec.ts.snap
```diff
@@ -999,7 +999,41 @@ exports[`method > processMethodLike(numberRangeParamMethod) 1`] = `
...
+                },
+              },
+            ],
             "text": "{ min: number; max: number }",
+            "type": "object",
+          },
+        },
+        {
+          "default": undefined,
+          "description": "The maximum value. Defaults to \`10\`.",
```
ST-DDT:
```suggestion
          "default": 10,
          "description": "The maximum value.",
```

ST-DDT (commented):
Diff Preview:

- https://deploy-preview-3947.fakerjs.dev/api/string.html#nanoid
- https://next.fakerjs.dev/api/string.html#nanoid

scripts/apidocs/output/page.ts
```diff
@@ -220,9 +222,7 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
         ...param,
         type: param.type.text,
         default: param.default ?? extractSummaryDefault(param.description),
-        description: await mdToHtml(
-          param.description.replace(defaultCommentRegex, '')
-        ),
+        description: await mdToHtml(stripSummaryDefault(param.description)),
```
ST-DDT: Maybe move the stripping and default extraction entirely out of here and only into the "raw" data.

ST-DDT (commented):
May be able to detect affected methods using `pnpm run generate:api-docs && pnpm run docs:diff https://deploy-preview-3943.fakerjs.dev/api-diff-index.json` (or plain old `diff -r docs/api docs2/api -x '*.json' -x '*.md'`)

scripts/apidocs/output/page.ts
Shinigami92: I started on this but hit a conflict I'd like your take on before committing to a direction.

Moving the **stripping** into the raw layer means the `default` field gets populated and the ``Defaults to `X`.`` text is removed from the raw `description` for **all** params (not just the range members). Extraction alone is fine — the stripping is what causes trouble: it breaks the `verify default value` test in `verify-jsdoc-tags.spec.ts`:

```ts
const commentDefault = extractSummaryDefault(description);
if (paramDefault) {
  ...
  } else if (
    !name.includes('.') &&
    (commentDefault ||
      (!description.includes('Defaults to') && !paramDefault.includes('(')))
  ) {
    expect(commentDefault).toBe(paramDefault);
  }
}
```

That test reads the **raw** `description` and relies on the ``Defaults to `X`.`` text still being there to assert that the JSDoc-documented default matches the implementation/signature default (e.g. `string.sample`'s `length` → `10`). Once we strip it in the raw layer, `commentDefault` becomes `undefined` while `paramDefault` is still set, so the assertion fails for top-level params. Member rows (names containing `.`) are exempt from this check, which is why the earlier `value.max` change didn't trip it.

So "move everything into raw" also removes the source this guardrail depends on. A few ways forward:

1. **Move extraction + stripping into raw, and relocate the guardrail into processing.** Keep the "impl/signature default must equal the summary default" check, but do it in `parameter.ts` at build time (throw a processing error on mismatch) instead of in the test. Cleanest end state, keeps the guarantee, and `page.ts` just renders. Requires reworking the `verify default value` test (its assertion becomes redundant).
2. **Move extraction + stripping into raw, drop the default-consistency assertion.** Smallest change, but we lose the check that catches a stale `Defaults to X` that no longer matches the real default.
3. **Extraction into raw, keep stripping in `page.ts`.** `default` is populated in the raw data, but the raw `description` keeps `Defaults to X` so the existing test is untouched. Doesn't fully move it "out of here", but least disruptive.

I'd lean towards **(1)** — it keeps the guardrail and gives `page.ts` a clean "render only" role — but it's a bit more scope. Which do you prefer? Or is there a fourth option you had in mind for where the consistency check should live?

ST-DDT: Good catch. I would also suggest (1). We may need to check how this impacts the preflight process though.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #3948, opened 2026-07-11
fix(internet): exclude reserved port 0 from internet.port()

### LINKED ISSUE #3937: faker.internet.port() can return 0, a reserved port per RFC 6335
### Problem
`faker.internet.port()` returns integers in `[0, 65535]`, but port `0` is reserved (RFC 6335 §6, IANA) and must not appear on the wire. Test code that feeds the value into a real socket (`net.connect({ port: 0 })`, `new URL(...)`) gets OS-assigned ephemeral behavior instead of the seeded value, producing non-deterministic tests.

### Steps
```js
import { faker } from '@faker-js/faker'; // 10.5.0
faker.seed(1);
for (let i = 0; i < 100000; i++) {
  if (faker.internet.port() === 0) { console.log('got 0 at', i); break; }
}
```

### Expected
A port in the valid usable range `[1, 65535]` (or a documented option to include 0).

### Actual
`0` is returned. Source: `src/modules/internet/module.ts:727` — `return this.faker.number.int(65535);` which inclusively includes 0. The existing test at `test/modules/internet.spec.ts:733` also asserts `toBeGreaterThanOrEqual(0)`, codifying the defect.

### Environment
- `@faker-js/faker` 10.5.0 (next branch, commit on 2026-07-03)
- Node 20+

Thanks for maintaining faker-js/faker!

xianjianlf2:
Closes #3937

`faker.internet.port()` calls `this.faker.number.int(65535)`, whose range is inclusive of `0`. Port `0` is listed as `Reserved` by IANA (RFC 6335 §6) and is never a usable service port, so seeded values fed into a real socket (`net.connect({ port: 0 })`) trigger OS-assigned ephemeral behavior instead of the seeded value.

This restricts the range to `1..65535`:

```diff
-    return this.faker.number.int(65535);
+    return this.faker.number.int({ min: 1, max: 65535 });
```

- Tightened the existing test assertion (`toBeGreaterThanOrEqual(0)` → `1`).
- Regenerated the single affected seeded snapshot (`internet > 42 > port`); no other snapshots changed.
- `pnpm exec vitest run test/modules/internet.spec.ts` → 1125 passed, no type errors.

### DIFF
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -750,7 +750,7 @@ export class InternetModule extends ModuleBase {
    * @since 5.4.0
    */
   port(): number {
-    return this.faker.number.int(65535);
+    return this.faker.number.int({ min: 1, max: 65535 });
   }
 
   /**
--- a/test/modules/__snapshots__/internet.spec.ts.snap
+++ b/test/modules/__snapshots__/internet.spec.ts.snap
@@ -90,7 +90,7 @@ exports[`internet > 42 > password > with pattern option 1`] = `"522319233860266"
 
 exports[`internet > 42 > password > with prefix option 1`] = `"testDfYsZdp522R"`;
 
-exports[`internet > 42 > port 1`] = `24545`;
+exports[`internet > 42 > port 1`] = `24546`;
 
 exports[`internet > 42 > protocol 1`] = `"http"`;
 
--- a/test/modules/internet.spec.ts
+++ b/test/modules/internet.spec.ts
@@ -766,7 +766,7 @@ describe('internet', () => {
           const port = faker.internet.port();
 
           expect(port).toBeTypeOf('number');
-          expect(port).toBeGreaterThanOrEqual(0);
+          expect(port).toBeGreaterThanOrEqual(1);
           expect(port).toBeLessThanOrEqual(65535);
           expect(String(port)).toSatisfy(isPort);
         });


xDivisionByZerox:
As by comment https://github.com/faker-js/faker/issues/3937#issuecomment-4881310791 we are currently not sure how to handle this. Since port `0` is not the only reserved port, flat out excluding it without all other reserved ones could lead to more problems than simply returning it for now.

ST-DDT:
@xianjianlf2 Can you comment your thoughts on the linked issues?

xianjianlf2:
Replied on the linked issue with my thoughts on bug-vs-feature classification and the reserved-port scope: https://github.com/faker-js/faker/issues/3937#issuecomment-4953908511

Short version: I've retracted the inaccurate "must not appear on the wire" wording (updated both descriptions) — the authoritative basis is simply that IANA lists port `0` as `Reserved`. `0` is a special case (never a usable service port), not a slippery slope toward 1023/1024, so I'd keep this PR as the minimal, non-breaking `min: 1` fix and treat configurable ranges/flags as a separate feature. Happy to fold it all into a broader feature PR instead if that's the preferred direction.

ST-DDT (approved):

ST-DDT (approved):

Shinigami92 (approved):

PR #3950, opened 2026-07-13
fix(locale): remove invalid building/flat pattern from pl buildingNumber

### LINKED ISSUE #2103: Building number returns incorrect data for PL
### Pre-Checks

### Describe the bug

The implementation of `buildingNumber` for PL is incorrect because it can return i.e. 4/2 which is not a valid building number in Poland.
https://github.com/faker-js/faker/blob/next/src/locales/pl/location/building_number.ts

4/2 in Poland means:
- 4 is the home number
- 2 is the flat number

I think this applies also to most countries in Europe.

### Additional Context

We should remove `'#/#'` from `buildingNumber` patterns for PL.

### Environment Info

```shell
N/A
```

- [ ] CJS
- [X] ESM

### Used Package Manager

npm

xianjianlf2:
## Problem
`faker.location.buildingNumber()` for the `pl` locale returned values like `4/2`. In Poland `4/2` means "building 4, flat 2" — not a valid standalone building number (#2103).

## Fix
Remove the `#/#` pattern from the `pl` `building_number` data, as requested in the issue. As noted by a native speaker and agreed by the maintainer, this pattern should not be part of `buildingNumber`.

## Testing
Added an assertion in `test/modules/location.spec.ts` that PL building numbers never contain `/` (red before, green after). `pnpm run generate:locales` produced no extra diff; full `location.spec.ts` passes; eslint clean.

Closes #2103


test/modules/location.spec.ts
```diff
@@ -280,6 +281,17 @@ describe('location', () => {
           const buildingNumber = faker.location.buildingNumber();
           expect(buildingNumber).not.toStartWith('0');
         });
+
+        it('does not return an invalid building/flat combination for pl', () => {
```
ST-DDT: IMO we dont need a test for this.

xianjianlf2: Good point — removed the test. Thanks for the review!

ST-DDT:
What about the things mentioned here:

- https://github.com/faker-js/faker/issues/2103#issuecomment-1535875003

Can you share some insights?

xianjianlf2:
Thanks for the pointer, that comment is the key insight I missed.

You're right to flag it. Re-reading [#2103 (comment)](https://github.com/faker-js/faker/issues/2103#issuecomment-1535875003), the native speaker explicitly states that `#/#` (and `#/#/#`) **are** valid Polish building numbers, not just building/flat combinations — the Polish Radio building is a well-known example. So the premise of this PR is wrong: `4/2` is a legitimate standalone building number, and removing the pattern would make the data *less* accurate.

That same comment also concluded the issue could be closed, since the earlier PR had already added the missing letter patterns (`##a`, `##b`, `##c`). So #2103 was effectively already resolved and no change is needed here.

I'll close this PR. Sorry for the noise, and thanks for catching it.

PR #3954, opened 2026-07-16
fix(vehicle): calculate VIN check digit

### LINKED ISSUE #3952: faker.vehicle.vin(): position 9 is random, so generated VINs fail check-digit validation
Since #2765, `vin()` has the correct length and excludes I/O/Q, but position 9 — the check digit under ISO 3779 / 49 CFR §565 — is generated as a free alphanumeric: it is often not even in the legal `[0-9X]` set for that position, and never computed. As a result, any test that feeds a generated VIN into a validator has to hardcode fixtures like `WBADT43452G012345` instead of using faker.

**Proposal:** compute position 9 with the standard algorithm (transliteration table + weights, sum mod 11, value 10 → `X`). Since position 9 carries weight 0, the digit can be computed over the VIN generated exactly as today and spliced in — the number of random draws stays identical, so only `vin()`'s own seeded output changes, not subsequent values in a seeded sequence.

**Precedents:** faker already computes check digits where validators exist — Luhn for card numbers, mod-97 for IBAN — and faker-ruby merged this exact fix (faker-ruby/faker#2633). Prior VIN accuracy fixes here: #319, #540 → #2765.

**Regional note:** the check digit is mandatory under the North American rule and regional under ISO 3779, so a computed digit is also a plausible position-9 value elsewhere — this strictly increases realism.

Seeded `vin()` snapshots would change (as #2765 did). Happy to open the PR (~40 lines + tests) if this is welcome.

tianrking:
## Summary

- calculate the ISO 3779 check digit for generated Vehicle Identification Numbers (VINs)
- preserve the existing random draws and replace only position 9 in the completed VIN
- add checksum coverage for generated VINs, the `X` modulo-11 case, and update the affected seeded snapshots

## Root cause

`vehicle.vin()` generated the ninth character as an unconstrained alphanumeric value. That position is the VIN check digit in the North American VIN standard, so generated values commonly failed validation or contained an invalid letter there.

The calculation uses the standard transliteration table and positional weights. Its zero weight for position 9 means the existing generated VIN can be used as input, preserving the random sequence while replacing only the check digit.

Fixes #3952.

## Validation

- `corepack pnpm@10.34.4 run build`
- `corepack pnpm@10.34.4 exec vitest run test/modules/vehicle.spec.ts`
- `corepack pnpm@10.34.4 exec eslint src/modules/vehicle/module.ts test/modules/vehicle.spec.ts`
- `corepack pnpm@10.34.4 exec prettier --check src/modules/vehicle/module.ts test/modules/vehicle.spec.ts`
- `corepack pnpm@10.34.4 run ts-check`
- `corepack pnpm@10.34.4 run test` *(54,317 passing; one existing docs-version-links failure in `test/docs/versions.spec.ts`, unrelated to this change)*
- `git diff --check`


src/modules/vehicle/module.ts
```diff
@@ -1,5 +1,48 @@
 import { ModuleBase } from '../../internal/module-base';
 
+const vinWeights = [8, 7, 6, 5, 4, 3, 2, 10, 0, 9, 8, 7, 6, 5, 4, 3, 2];
```
matthewmayer, line 6: Mayb add a comment with a link to any official source for the the weights and transliteration.

test/modules/vehicle.spec.ts
```diff
@@ -5,6 +5,44 @@ import { times } from '../support/times';
...
+  T: 3,
+  U: 4,
+  V: 5,
+  W: 6,
+  X: 7,
+  Y: 8,
+  Z: 9,
+};
+
+function calculateVinCheckDigit(vin: string): string {
```
ST-DDT: This is the exact same code as the implementation. Maybe reuse the code from the module/add export.

test/modules/vehicle.spec.ts
```diff
@@ -102,6 +140,15 @@ describe('vehicle', () => {
             /^([A-HJ-NPR-Z0-9]{10}[A-HJ-NPR-Z0-9]{1}[A-HJ-NPR-Z0-9]{1}\d{5})$/
           );
         });
+
+        it('should return a VIN with a valid check digit', () => {
+          expect(calculateVinCheckDigit('1M8GDM9AXKP042788')).toBe('X');
+
+          for (let index = 0; index < 100; index++) {
+            const vin = faker.vehicle.vin();
+            expect(vin[8]).toBe(calculateVinCheckDigit(vin));
```
ST-DDT: Please split this test into two.
One that computes the check digit for 5 hardcoded vins.
The other (no-loop) confirming that the vin was added at the correct place.

ST-DDT (approved):

Shinigami92 (dismissed):

src/modules/vehicle/module.ts
```diff
@@ -1,5 +1,51 @@
 import { ModuleBase } from '../../internal/module-base';
 
+// NHTSA 49 CFR § 565.15(c), Tables III and IV define the transliteration
+// values and position weights used here:
+// https://www.govinfo.gov/content/pkg/CFR-2025-title49-vol6/pdf/CFR-2025-title49-vol6-sec565-15.pdf
```
xDivisionByZerox: I might be stupid, but this link does not lead to any valid PDF for me 🤷‍♀️

ST-DDT (changes requested):
Can you please address the above comment?

src/modules/vehicle/module.ts
ST-DDT:
```suggestion
// https://www.govinfo.gov/content/pkg/CFR-2024-title49-vol6/pdf/CFR-2024-title49-vol6-sec565-15.pdf
```

ST-DDT (approved):

Shinigami92 (approved):

PR #3955, opened 2026-07-16
fix(color): correct LCH hue range and HSLA alpha formatting

### LINKED ISSUE #3956: fix(color): LCH hue range capped at 230 instead of 360, HSLA alpha incorrectly scaled
### Pre-Checks

### Describe the bug

Two bugs in the color module:

### Minimal reproduction code

```ts

faker.seed(42);
  if (hue > 230) {
    break;
  }
}

// Bug 2: HSLA alpha is wrong
```

### Expected behavior

### Environment

- `@faker-js/faker` 10.5.0
- Node 20+

I have a PR ready: #3955

pouyashahrdami:
## Summary

- **LCH hue capped at 230 instead of 360:** The `lch()` method used a loop to generate both chroma and hue with `max: 230`. Chroma's upper bound of 230 is correct (as documented), but hue is an angle in degrees and should range 0–360. This meant ~36% of the hue circle (blues, purples, magentas above 230°) could never be generated.

- **HSLA CSS alpha scaled incorrectly:** The `toCSS()` helper passed the HSLA alpha value through `toPercentage()`, which multiplies by 100. An alpha of `0.75` was output as `75` instead of `0.75` in the CSS string. The JSDoc example shows the correct format (`hsl(0deg 100% 50% / 0.5)`) but the code produced `hsl(0deg 100% 50% / 50)`. Compare with RGBA (line 127) which correctly passes alpha through as-is.

## Test plan

- [x] Updated LCH decimal-format tests to assert per-component bounds (chroma ≤ 230, hue ≤ 360) instead of a blanket ≤ 230 loop
- [x] Updated seeded snapshots to reflect new LCH hue values
- [x] All 236 color module tests pass


test/modules/__snapshots__/color.spec.ts.snap
```diff
@@ -51,7 +51,7 @@ exports[`color > 42 > lch 1`] = `
 [
   0.37454,
   218.7,
-  168.4,
+  263.5,
```
ST-DDT, line 54: I wonder why hsla doesnt show up as changed.

test/modules/color.spec.ts
ST-DDT, line 1: Please add test for hsla format.

ST-DDT:
This fixes two unrelated bugs in two unrelated methods. For ease of review+changelog, please only fix related bugs within the same PR.

ST-DDT:
Closing as stale. If you are interested in updating this/adressing the feedback. Please let us know.

PR #3957, opened 2026-07-19
docs(guide): document native ESM support for Jest 30.4.2+

### LINKED ISSUE #3606: Faker v10 ESM-only breaks compatibility with Jest

MahinAnowar:
Adds the Jest section you suggested in #3606 — that with Jest >=30.4.2 the transformIgnorePatterns workaround isn't needed anymore, just `NODE_OPTIONS="--experimental-vm-modules"`, with a link to Jest's ESM docs.

One detail I hit while double-checking your recipe before writing it down: for require-based CJS test files this also needs Node >=24.9 — on Node 22, Jest 30.4.2 fails with "Jest's require(ESM) requires Node v24.9+ for synchronous vm module APIs". Since Faker v10 still supports Node 20/22 per engines, I added a sentence about that so people on supported-but-older Node versions aren't left confused. Verified both ways in a scratch project: Node 24.18 + the flag passes a plain `require('@faker-js/faker')` test with zero transform config, Node 22 + the flag fails with that exact message, and no flag at all gives the classic SyntaxError.

Left the ts-jest workaround section untouched for older Jest versions. Prettier passes on the file.

Closes #3606 if you feel this covers it.


ST-DDT (dismissed):
I think after adressing the version comment this is good to go.

Shinigami92 (dismissed):

xDivisionByZerox (dismissed):
Works for me. @faker-js/members can we can an implementation of this jn the playground to verify that this actually works (and keeps working) before we merge this?

MahinAnowar:
Done, changed it to v24.9.0.

Shinigami92 (approved):

ST-DDT (approved):

PR #3959, opened 2026-07-20
feat: add exclude option to faker.helpers.enumValue

### LINKED ISSUE #2994: Add exclude on enum faker.helpers.enumValue

I wanted to seed some data to the database and one of the tables needs image type and an image so while seeding I couldn't add a file since it needs to be uploaded to the server so for the seed purpose only I wanted to remove the file value from the enum or exclude but there is no option for that 

### Suggested solution

```
enum ImageType {
  file
  link
  emoji
  constant
}

const imageType = faker.helpers.enumValue(ImageType, {exclude: [ImageType.file]});
```

hazrid93:
Closes #2994.

Add an optional `exclude` option (`enumValue(enumObject, { exclude: [...] })`): values in `exclude` are not eligible to be picked. If exclude removes every enum value, a FakerError is thrown. The default behavior (no exclude / empty exclude) is unchanged.

### DIFF
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1070,6 +1070,10 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    * @template T Type of generic enums, automatically inferred by TypeScript.
    *
    * @param enumObject Enum to pick the value from.
+   * @param options The options to use.
+   * @param options.exclude The values to exclude when picking.
+   *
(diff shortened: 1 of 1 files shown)


ST-DDT (dismissed):

ST-DDT:
Could you please address the feedback provided by copilot?

ST-DDT:
This PR appears to be abandoned. If you would like to continue with this please comment, so we can reopen this.

PR #3960, opened 2026-07-20
feat: casing for random words (lorem.word / lorem.words casing option)

### LINKED ISSUE #2165: casing for random words

hazrid93:
Closes #2165.

Add a `casing?: 'lower' | 'upper' | 'capital' | 'mixed'` option to lorem.word (and a matching options param to lorem.words that threads it to each generated word). 'mixed' (the default) leaves the word's original casing unchanged, so existing default behavior (and snapshots) are unaffected. 'lower'/'upper'/'capital' apply lower/upper/capitalize.

### DIFF
--- a/src/modules/lorem/module.ts
+++ b/src/modules/lorem/module.ts
@@ -2,6 +2,37 @@ import { ModuleBase } from '../../internal/module-base';
 import type { NumberRange } from '../../utils/types';
 import { filterWordListByLength } from '../word/filter-word-list-by-length';
 
+type LoremWordCasing = 'lower' | 'upper' | 'capital' | 'mixed';
+
+/**
+ * Applies the requested casing to a word.
+ *
+ * @param word The word to transform.
+ * @param casing The casing to apply. Defaults to `'mixed'`.
+ */
+function applyCasing(word: string, casing?: LoremWordCasing): string {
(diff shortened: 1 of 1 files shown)


ST-DDT (changes requested):

ST-DDT:
This PR appears to be abbandonned. If you are still interested in it. Please let us know so we can reopen it.

PR #3961, opened 2026-07-20
feat: faker.system.directoryPath depth option

### LINKED ISSUE #3785: `faker.system.directoryPath` should have a "depth" option

When generating a path using `faker.system.directoryPath` there should be an option to specify how "deep" of a. path should be generated.

### Suggested solution

Add parameters to the function that allows the user to specify how deep of a directory path they want generated.

Depth examples:
- 0: `/`
- 1: `/etc`, or `/usr`
- 2: `/etc/mail` or `/usr/lib`
- 3: `/usr/lib/log`

Depth value could be passed either as a parameter, or as part of an "options" object

```ts
const asParameter = faker.system.directoryPath(3);

const asOptionsObject = faker.system.directoryPath({ depth: 3 });
```

If no depth parameter is included, function would behave as it does today.

### Alternative

_No response_

### Additional context

This functionality would be extremely useful when generating test data for parent/child path structures. For example, if I'm generating something with a path of `/x/y/z`, and the structure includes a "parent" property, it would be easier to generate a value that could be used to set both if the depth of the directory structure could be set.

hazrid93:
Closes #3785.

Add an optional `{ depth, root }` to directoryPath: when `depth` is omitted, behavior is unchanged (a random path from the built in directory_path definitions). When `depth` is provided, a path with exactly that many segments is generated from a built in pool of common directory names (bin, etc, lib, usr, ...). depth 0 to "/"; root:false produces a relative path.

### DIFF
--- a/src/modules/system/module.ts
+++ b/src/modules/system/module.ts
@@ -4,6 +4,31 @@ import type { NumberRange } from '../../utils/types';
 
 const commonFileTypes = ['video', 'audio', 'image', 'text', 'application'];
 
+// Common directory-name segments used when generating a path of a given depth.
+const directoryNamePool = [
+  'bin',
+  'boot',
+  'dev',
+  'etc',
+  'home',
+  'lib',
+  'lib64',
+  'log',
+  'mail',
+  'mnt',
+  'opt',
+  'proc',
+  'root',
+  'run',
+  'sbin',
+  'srv',
+  'src',
+  'sys',
+  'tmp',
+  'usr',
+  'var',
+];
+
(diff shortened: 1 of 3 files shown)


ST-DDT (changes requested):

matthewmayer:
the associated issue only has two upvotes, i dont think there is consensus this is needed yet.

ST-DDT:
I agree. Currently the definition of this method is also pretty vague.
Is it a windows path or a linux one? Is it relative or absolute. IMO we should strengthen the definition before adding features to it.

/vote close

PR #3962, opened 2026-07-20
feat: option to exclude emoji from person.bio()

### LINKED ISSUE #2623: Add option to exclude emoji

hazrid93:
Closes #2623.

Add an optional `{ excludeEmoji?: boolean }` to person.bio: when true, the generated bio is stripped of emoji (Extended_Pictographic), zero width joiners, variation selectors, and regional indicator flag pairs, with resulting whitespace collapsed. If stripping would leave an empty string, the original bio is returned. Default (no option or excludeEmoji:false) behavior is unchanged.

### DIFF
--- a/src/modules/person/module.ts
+++ b/src/modules/person/module.ts
@@ -3,6 +3,24 @@ import type { Faker } from '../../faker';
 import { ModuleBase } from '../../internal/module-base';
 
 /**
+ * Strips emoji + related zero-width / variation / flag code points from a
+ * string, collapsing any resulting whitespace. Returns the original string if
+ * the result would be empty.
+ *
+ * @param text The text to strip emoji from.
+ */
+function stripEmoji(text: string): string {
+  const stripped = text
+    .replaceAll(/\p{Extended_Pictographic}/gu, '')
+    .replaceAll('\u200D', '')
+    .replaceAll(/[\uFE0E\uFE0F]/g, '')
+    .replaceAll(/[\u{1F1E6}-\u{1F1FF}]/gu, '')
+    .replaceAll(/\s+/g, ' ')
+    .trim();
+  return stripped.length > 0 ? stripped : text;
+}
+
+/**
  * The enum for values corresponding to a person's sex.
  */
 export enum Sex {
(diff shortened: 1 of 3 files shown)


ST-DDT (changes requested):
Also not sure about the approach.

src/modules/person/module.ts
```diff
@@ -304,15 +322,32 @@ export class PersonModule extends ModuleBase {
...
-    return this.faker.helpers.fake(this.faker.definitions.person.bio_pattern);
+  bio(options?: {
+    /**
+     * When `true`, strip emoji and related code points from the result.
+     */
+    excludesEmoji?: boolean;
+    /**
+     * Alias of `excludesEmoji`. When `true`, strip emoji and related code points from the result.
+     */
+    excludeEmoji?: boolean;
```
ST-DDT, line 344: IMO we only need one of the options.

I prefer `excludeEmoji`

src/modules/person/module.ts
```diff
@@ -304,15 +322,32 @@ export class PersonModule extends ModuleBase {
...
+  }): string {
+    const bio = this.faker.helpers.fake(
+      this.faker.definitions.person.bio_pattern
+    );
+    const exclude = options?.excludesEmoji ?? options?.excludeEmoji ?? false;
```
ST-DDT, line 349: We usually use deconstructor patterns for options.

````ts
const { excludeEmoji = false } = options;
````

File: src/modules/person/module.ts [hunk 3]
```diff
@@ -304,15 +322,32 @@ export class PersonModule extends ModuleBase {
...
+     * When `true`, strip emoji and related code points from the result.
+     */
+    excludesEmoji?: boolean;
+    /**
+     * Alias of `excludesEmoji`. When `true`, strip emoji and related code points from the result.
+     */
+    excludeEmoji?: boolean;
+  }): string {
+    const bio = this.faker.helpers.fake(
+      this.faker.definitions.person.bio_pattern
```
ST-DDT, line 347: Instead of removing the emoji after adding it.
It might be best to either use no-emoji-patterns from the start (preffered) 
or just remove the {{emoji}} patterns prior to evaluating the patterns.

src/modules/person/module.ts
```diff
@@ -304,15 +322,32 @@ export class PersonModule extends ModuleBase {
...
    */
-  bio(): string {
-    return this.faker.helpers.fake(this.faker.definitions.person.bio_pattern);
+  bio(options?: {
+    /**
+     * When `true`, strip emoji and related code points from the result.
+     */
+    excludesEmoji?: boolean;
+    /**
+     * Alias of `excludesEmoji`. When `true`, strip emoji and related code points from the result.
```
ST-DDT, line 342: Please also include a `@default false` in the jsdocs.

src/modules/person/module.ts
```diff
@@ -304,15 +322,32 @@ export class PersonModule extends ModuleBase {
   }
 
   /**
-   * Returns a random short biography
+   * Returns a random short biography.
+   *
+   * @param options The optional options object.
+   * @param options.excludesEmoji When `true`, strip emoji and related code points from the result.
+   * @param options.excludeEmoji Alias of `excludesEmoji`. When `true`, strip emoji and related code points from the result.
```
ST-DDT, line 329:
```suggestion
   * @param options.excludeEmoji When `true` omit emojis from the result. Defaults to `false`.
```

Not sure about the wording other than ``Defaults to `false`.``

ST-DDT (changes requested):
Sorry, I only had time to review it in more detail now.

Though I'm not sure whether we actually need this.
The feature currently only has 5 upvotes.

ST-DDT:
This PR appears to be abandoned. If you would like to continue with it, please address the requested changes.

PR #3963, opened 2026-07-20
feat: allow a value returning function in helpers.weightedArrayElement

### LINKED ISSUE #3442: Allow value-returning function in weightedArrayElement

I want to generate a weighted distribution where each member of the distribution is itself a randomly generated item with specific configuration.

For example, I want to generate a set of fake movies where the release date is random dates weighted into decades (generate more recent movies than older movies).

### Suggested solution

Allow `faker.helpers.weightedArrayElement` to take a `() => T` in addition to just `T` for value:

```ts
faker.helpers.weightedArrayElement([
  {weight: 1, value: () => faker.date.between({from: "1980-01-01", to: "1989-12-31"})},
  {weight: 2, value: () => faker.date.between({from: "1990-01-01", to: "1999-12-31"})},
  {weight: 3, value: () => faker.date.between({from: "2000-01-01", to: "2009-12-31"})},
  {weight: 4, value: () => faker.date.between({from: "2010-01-01", to: "2010-12-31"})},
])
```

### Alternative

Add a new separate method for this kind of weighted randomized distribution, but the code change to support this would be minimal.

### Additional context

_No response_

hazrid93:
Closes #3442.

Allow the `value` of each entry to be a `() => T` function; when the entry is selected, the function is invoked to produce the value. Plain values are returned directly (default behavior unchanged). Only the selected entry's function is invoked (laziness).

### DIFF
--- a/src/modules/helpers/module.ts
+++ b/src/modules/helpers/module.ts
@@ -1003,13 +1003,16 @@ export class SimpleHelpersModule extends SimpleModuleBase {
     for (const { weight, value } of array) {
       current += weight;
       if (random < current) {
-        return value;
+        // Allow a value-returning function so callers can lazily generate the
+        // selected item (e.g. another faker call). See issue #3442.
+        return typeof value === 'function' ? (value as () => T)() : value;
       }
     }
 
     // In case of rounding errors, return the last element
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    return array.at(-1)!.value;
+    const last = array.at(-1)!.value;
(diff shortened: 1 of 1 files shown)


ST-DDT (changes requested):
This will conflict if you actually need to select/return a function.

ST-DDT:
This will be better resolved by a new (set of) function that are designed specifically for that purpose.

See:

- #3176 

---

The linked issue hasnt gathered sufficient community support and the implementation has serious issues/side effects -> Closing

PR #3965, opened 2026-07-20
feat: weight genders in person.gender()

### LINKED ISSUE #1730: Weight genders in `faker.person.gender()`

`faker.person.gender()` pulls from a list of genders at https://github.com/faker-js/faker/blob/next/src/locales/en/person/gender.ts

This is a very inclusive list, however I feel it makes the function less realistic to have all values returned with equal probability. In reality, even in fairly LGBTQ-friendly countries, the percentage of people who define as non-cis-gender is fairly low https://www.bbc.com/news/uk-64184736

### Suggested solution

Now that we have `faker.helpers.weightedArrayElement` i think we could change this so for example it returns "Man" 45% of the time, "Woman" 45% of the time, and one of the other genders 10% of the time (divided equally between the other 73 options)

### Alternative

Keep current behavior (unrealistic)
Only return Man or Woman (non-inclusive)

### Additional context

_No response_

hazrid93:
Closes #1730.

Add an optional `{ weighting?: 'female' | 'male' | 'mixed' }`: when 'mixed' or omitted, behavior is unchanged (uniform arrayElement). When 'female'/'male', every entry whose name contains that gender as a whole word (case insensitive, word boundary regex: so 'male' does not match inside 'female') gets a much higher weight (list.length) via weightedArrayElement, biasing the result toward female/male identifying genders while still allowing other identities.

### DIFF
--- a/src/modules/person/module.ts
+++ b/src/modules/person/module.ts
@@ -242,10 +242,20 @@ export class PersonModule extends ModuleBase {
    *
    * @since 8.0.0
    */
-  gender(): string {
-    return this.faker.helpers.arrayElement(
-      this.faker.definitions.person.gender
-    );
+  gender(options?: { weighting?: 'female' | 'male' | 'mixed' }): string {
+    const list = this.faker.definitions.person.gender;
+    if (!options?.weighting || options.weighting === 'mixed') {
+      return this.faker.helpers.arrayElement(list);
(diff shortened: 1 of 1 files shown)


ST-DDT:
This does not follow the proposed solution of the issue.

PR #3966, opened 2026-07-20
feat(person): support sexType option for sex()

### LINKED ISSUE #3567: Person Module > Add sexType parameter to sex function

You currently cannot generate a **localized** sex and first name for a single person and guarantee that they match.

### Suggested solution

Allow faker.person.sex() to also receive the sexType parameter ensuring the given sex but effectively localizing it.

### Alternative

Alternatively there could also be a "localizer" function that takes a sexType and returns the same output the faker.person.sex function returns for the given sexType, but I personally would prefer the prior.

### Additional context

_No response_

hazrid93:
Closes #3567.

Add an optional `{ sexType?: SexType }`: when omitted, behavior is unchanged (uniform arrayElement). When given ('female'/'male'), returns the matching localized sex string by canonical index (female=0, male=1 in the en source list), so the returned string matches the requested sex even in non en locales. 'generic' (no dedicated list entry) falls back to a valid sex from the list.

### DIFF
--- a/src/modules/person/module.ts
+++ b/src/modules/person/module.ts
@@ -254,6 +254,9 @@ export class PersonModule extends ModuleBase {
    * Output of this method is localised, so it should not be used to fill the parameter `sex`
    * available in some other modules for example `faker.person.firstName()`.
    *
+   * @param options The optional options object.
+   * @param options.sexType When set, return the localized sex string matching the given sex type instead of a random one.
+   *
    * @see faker.person.gender(): For generating a gender related value.
    * @see faker.person.sexType(): For generating a sex value to be used as a parameter.
    *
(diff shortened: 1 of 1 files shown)


ST-DDT (changes requested):

If we define indexes 0/1 as female/male then we have to document that in the definitions.

src/modules/person/module.ts
```diff
@@ -262,8 +265,31 @@ export class PersonModule extends ModuleBase {
...
+      options.sexType === (Sex.Female as SexType)
+        ? 0
+        : options.sexType === (Sex.Male as SexType)
+          ? 1
+          : -1;
+    if (index >= 0 && index < list.length) {
+      return list[index];
+    }
+
+    return this.faker.helpers.arrayElement(list);
```
ST-DDT, line 292: IMO this method should use:

````ts
selectDefinition(
        this.faker,
        sex,
        this.faker.definitions.person.sex
      )
````

similar to the other methods e.g. firstName/lastName.

ST-DDT (changes requested):

ST-DDT:
This PR appears to be abandoned. If you would like to continue with it, please address the requested changes.

PR #3967, opened 2026-07-20
feat: only generate IANA example domains

### LINKED ISSUE #2228: Support only generating IANA example domains

From my understanding, Faker currently concatenates random words to generate domains (eg for internet.domainName, internet.url, etc). When using faker for random seeding, these domains may wind up liable to be actually followed, intentionally or not. In order to avoid directing requests to an actual website, it would be great if there could be a method to only generate domains using domains reserved by IANA for examples

### Suggested solution

Following the example of `exampleEmail`, you could imagine creating `exampleDomainName` and `exampleUrl`, which could use `<domainWord>.example.<com|net|org>`. 

### Alternative

* Providing options to the existing domainName/url functions
* A global configuration property to always generate domains with example domains - this could also be useful for email as well. I would arguable that this would be desirable as it would be more robust if you want to ensure all usages of these functions perform identically, but I can also see a desire for wanting control on a more granular level. This could be a separate scope

### Additional context

_No response_

hazrid93:
Closes #2228.

Add an optional `{ allowOnlyExamples?: boolean }`: when true, returns <word>.example.<com|org|net>: a 3 label subdomain under the IANA reserved example domains (RFC 2606), which never resolve. Default (no option / false) behavior is unchanged.

### DIFF
--- a/src/modules/internet/module.ts
+++ b/src/modules/internet/module.ts
@@ -540,12 +540,27 @@ export class InternetModule extends ModuleBase {
   /**
    * Generates a random domain name.
    *
+   * @param options The optional options object.
+   * @param options.allowOnlyExamples When `true`, generate a domain under the IANA-reserved example domains (example.com / example.org / example.net) so the result never resolves to a real host. See issue #2228.
+   *
    * @example
    * faker.internet.domainName() // 'slow-timer.info'
+   * faker.internet.domainName({ allowOnlyExamples: true }) // 'slow-timer.example.com'
    *
    * @since 2.0.1
    */
-  domainName(): string {
+  domainName(options?: {
+    /**
+     * When `true`, generate a domain under the IANA-reserved example domains (example.com / example.org / example.net) so the result never resolves to a real host.
+     */
+    allowOnlyExamples?: boolean;
+  }): string {
+    if (options?.allowOnlyExamples) {
+      const word = this.domainWord();
+      const tld = this.faker.helpers.arrayElement(['com', 'org', 'net']);
+      return `${word}.example.${tld}`;
+    }
+
     return `${this.domainWord()}.${this.domainSuffix()}`;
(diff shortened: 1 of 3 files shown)


ST-DDT (dismissed):

ST-DDT:
I'm unsure what the best way to achieve this is.

Implementation is mostly fine, but I won't comment on the remaining todos until we decide on the solution.

src/modules/internet/module.ts
```diff
@@ -540,12 +540,27 @@ export class InternetModule extends ModuleBase {
...
+    /**
+     * When `true`, generate a domain under the IANA-reserved example domains (example.com / example.org / example.net) so the result never resolves to a real host.
+     */
+    allowOnlyExamples?: boolean;
+  }): string {
+    if (options?.allowOnlyExamples) {
+      const word = this.domainWord();
+      const tld = this.faker.helpers.arrayElement(['com', 'org', 'net']);
+      return `${word}.example.${tld}`;
+    }
```
ST-DDT, line 562: Maybe this could be replaced by a more flexible `baseDomain`/`parentDomain` parameter.
`example.com` resolves to a real host, so the current claim is not really true. 
We could still mention the three example domains in the jsdocs of the new parameter.

What do you think?

matthewmayer: I like 

    internet.domainName({baseDomain:"example.com"}) 

As it unlocks more use cases like

    internet.domainName({baseDomain:"co.uk"}) 

    internet.domainName({baseDomain:"yourdomain.com"}) 

    internet.domainName({baseDomain:"example"})

ST-DDT:
What do you think about the suggested alternative?

ST-DDT:
This PR appears to be abandonned.
Please feel free to reply, if you want to continue this.

PR #3970, opened 2026-07-27
feat(helpers): resolve functions from prototypes in fake

### LINKED ISSUE #3465: helpers.fake - patterns after root expression fail when referencing instance members

hugosmoreira:
## Summary

- resolve properties on primitive values returned by `fakeEval`
- bind generated object and primitive methods to their source value
- document member chaining for `fakeEval` and `helpers.fake`
- add regression coverage for implicit and explicit calls, properties, prototype methods, and constructor-chain safety

## Root cause

The expression resolver only followed properties on objects. When a Faker method returned a primitive, the chain stopped; when it returned an object method such as `Date.prototype.toISOString`, the method was later invoked without its receiver.

The resolver now boxes supported primitives for property lookup and binds callable properties to the generated value. Functions returned by functions still use the existing recursive resolution path, preserving the constructor-chain protection added for GHSA-qxc2-j82w-r537.

## Validation

- `pnpm run preflight`
  - 55 test files passed
  - 52,740 tests passed
  - TypeScript checks passed
- targeted helpers suites: 862 tests passed

Closes #3465

This contribution was developed and validated with OpenAI Codex assistance.


src/modules/helpers/module.ts
```diff
@@ -1197,6 +1197,13 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
    * ```
    *
+   * Properties and prototype methods on generated values can also be resolved:
+   *
+   * ```js
+   * const length = faker.helpers.fake('{{string.alpha(10).length}}'); // '10'
+   * const timestamp = faker.helpers.fake('{{date.anytime.toISOString}}'); // '2025-01-01T00:00:00.000Z'
+   * ```
+   *
```
ST-DDT, line 1206: Please repeat this for the other signatures of fake.

src/modules/helpers/eval.ts
```diff
@@ -222,11 +225,37 @@ function resolveProperty(entrypoint: unknown, key: string): unknown {
...
+      return getProperty(entrypoint, key);
     }
 
     default: {
       return undefined;
     }
   }
 }
+
+type Callable = (this: unknown, ...args: unknown[]) => unknown;
```
ST-DDT: Maybe

 (this: unknown, ...args: never[]) => unknown;

src/modules/helpers/module.ts
hugosmoreira: Done in 3dffad926: the member-chaining documentation and examples are now included on the array and combined overloads as well.

src/modules/helpers/eval.ts
hugosmoreira: Updated Callable to (this: unknown, ...args: never[]) => unknown in 3dffad926. The full preflight passes.

src/modules/helpers/eval.ts
```diff
@@ -222,11 +225,37 @@ function resolveProperty(entrypoint: unknown, key: string): unknown {
...
+ *
+ * @param entrypoint The entrypoint to resolve the property on.
+ * @param key The property name to resolve.
+ */
+function getProperty(entrypoint: unknown, key: string): unknown {
+  if (entrypoint == null) {
+    return undefined;
+  }
+
+  const value = (new Object(entrypoint) as Record<string, unknown>)[key];
```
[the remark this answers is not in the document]
hugosmoreira: I tested this form, but the configured unicorn/new-for-builtins rule rejects Object(entrypoint) and requires new Object(entrypoint). I kept the current form so the project lint passes.

src/modules/helpers/module.ts
```diff
@@ -1307,6 +1314,13 @@ export class HelpersModule extends SimpleHelpersModule {
    * const message = faker.helpers.fake('Your pin is {{string.numeric(4, {"allowLeadingZeros": true})}}.');
    * ```
    *
+   * Properties and prototype methods on generated values can also be resolved:
+   *
+   * ```js
+   * const length = faker.helpers.fake('{{string.alpha(10).length}}'); // '10'
```
matthewmayer: can we think of a more realistic/useful example for this? The second example is more helpful.

hugosmoreira: Updated all three `fake` overload examples in d73e8edc to use the practical `airline.airline.iataCode` property lookup while keeping `date.anytime.toISOString` as the prototype-method example. Full preflight passes.

src/modules/helpers/eval.ts
ST-DDT, line 66:
```suggestion
 * fakeEval('number.int(9999)', faker) // 4834
 * fakeEval('date.anytime.toISOString', faker); // '2025-01-01T00:00:00.000Z'
```

src/modules/helpers/eval.ts
```diff
@@ -222,11 +225,37 @@ function resolveProperty(entrypoint: unknown, key: string): unknown {
     }
 
     case 'object': {
-      return entrypoint?.[key as keyof typeof entrypoint];
+      return getProperty(entrypoint, key);
```
ST-DDT: Should we merge this case to the cases below?

src/modules/helpers/module.ts
ST-DDT, line 1224: Please add a similar example to the other signatures as well.

```suggestion
   * faker.helpers.fake('Your PIN number is: {{string.numeric(4, {"exclude": ["0"]})}}') // 'Your PIN number is: 4834'
   * faker.helpers.fake('Expires on {{date.future.toISOString.substring(0,10)}}')  // 'Expires on 2026-09-09'
```

src/modules/helpers/eval.ts
hugosmoreira: Added the chained date.anytime.toISOString example to the fakeEval documentation in 831c6ce3. Full preflight passes.

src/modules/helpers/eval.ts
hugosmoreira: Done in 831c6ce3. The object and primitive cases now share the same getProperty branch, while getProperty continues to handle null safely. Full preflight passes.

src/modules/helpers/module.ts
hugosmoreira: Added the PIN and expiry examples to the string, array, and combined fake overloads in 831c6ce3. Full preflight passes.

ST-DDT (dismissed):

Shinigami92 (dismissed):

ST-DDT:
Could you please fix them?

hugosmoreira:
Thanks @ST-DDT! I've merged the current `next` into this branch in d4882ba1, preserving the existing implementation and regression tests through the upstream evaluator rename. GitHub now reports the PR as mergeable.

Validated locally on Windows with Node 22.20.0 and the pinned pnpm 11.25.0: package and documentation builds, code generation, formatting, lint, TypeScript checks, and the full unit/type-test suite with coverage (54,211 passing tests; no unexpected failures). The helper suites also passed all 862 tests. A temporary negative control using the old resolver failed the 11 targeted regression cases as expected; the restored implementation passed.

The new CI workflows are awaiting maintainer authorization (`action_required`). Could you approve those runs when you have a chance? Thanks again!

ST-DDT (approved):

Shinigami92 (approved):

ST-DDT:
@xDivisionByZerox Could you please review this PR. Especially, regarding the usecase/scope.

ST-DDT:
FYI: This is still in internal discussion.
Not sure when we can decide on it. It may take a while though.
Sorry for the slow progress.

PR #3971, opened 2026-07-27
chore: replace import from dist

ST-DDT:
Replaces the import

````ts
import type { PersonEntryDefinition } from '../dist';
````

with one from `'../src/definitions'`.


ST-DDT:
IMO this issue so is so minor, we don't need to wait for it to be merged.

Shinigami92 (approved):

PR #3972, opened 2026-08-01
fix(string): generate valid ULIDs for negative reference dates

hiSandog:
`faker.string.ulid({ refDate: -1 })` currently uses negative Base32 indices while encoding the timestamp, which inserts `undefined` into the result and produces an invalid ULID.

This normalizes pre-epoch timestamps to the Unix epoch before Base32 encoding, matching the existing UUID v7 behavior. A public API regression test verifies that negative reference dates now produce valid ULIDs with a zeroed timestamp prefix.

Validation:

- `pnpm exec vitest run test/internal/base32.spec.ts test/modules/string.spec.ts` (1175 passed, no type errors)
- `pnpm exec prettier --check src/internal/base32.ts test/modules/string.spec.ts`
- `pnpm exec eslint src/internal/base32.ts test/modules/string.spec.ts`

### DIFF
--- a/src/internal/base32.ts
+++ b/src/internal/base32.ts
@@ -9,7 +9,7 @@ export const CROCKFORDS_BASE32 = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
  * @param date The Date to encode.
  */
 export function dateToBase32(date: Date): string {
-  let value = date.valueOf();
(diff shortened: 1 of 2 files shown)


ST-DDT:
Sorry, these PRs seem to have slipped my attention entirely. Thanks for contributing anyway!

PR #3976, opened 2026-08-01
fix(apidocs): scope code group tab identifiers

### LINKED ISSUE #3847: apidocs: Tab-Group name collision between different methods' examples
### Pre-Checks

### Describe the bug

If one page contains multiple examples with the same tab group names, when switching it on a later one, the browser scrolls to the respective examples of the first method that defines that name.

### Minimal reproduction code

- Go to https://deploy-preview-3818.fakerjs.dev/api/utils.html#setdefaultrefdate
- Switch to a different tab group
- Get pushed to that example of `getDefaultRefDate`

### Additional Context

Currently, only visible in a PR. Since faker.defaultRefDate does not have examples (yet).

- https://github.com/faker-js/faker/pull/3818

Not sure whether we can fix it on our side or whether it needs fixing upstream.

### Environment Info

```shell
Firefox 150.0.2 (64-Bit)
Windows 11
```

- [ ] CJS
- [x] ESM

### Used Package Manager

pnpm

UlikGames:
Fixes #3847

The API docs render each method's examples separately. VitePress resets its code group counter on every render, so pages with more than one tabbed example reuse values such as `group-0` and `tab-1`. Clicking a tab in a later example can then affect or jump to the first group.

This adds a stable prefix for each module and method code group. The replacement only touches the generated radio inputs and labels, so matching text inside code examples remains unchanged. Invalid prefixes are rejected as well.

I added regression tests and ran the full test suite, type checking, linting, formatting checks, API docs generation, and the VitePress build.


test/scripts/shared/markdown.spec.ts
```diff
@@ -11,23 +11,55 @@ describe('markdown', () => {
...
+        ['// First title\nconst a = 1;', '// Second title\nconst b = 2;'],
+        'example'
+      );
 
       expect(html).toMatchSnapshot();
     });
 
+    it('scopes tab controls to their code group', async () => {
+      const codes = [
+        '// First title\nconst attributes = \'id="tab-custom" name="group-custom" for="tab-custom"\';',
```
ST-DDT, line 30: Is the later part of this test intended to verify that ids/names/for contained in the code snippet are kept as is?

If yes, maybe add a comment before line 43 stating that.

scripts/apidocs/output/page.ts
```diff
@@ -226,7 +232,7 @@ async function toMethodData(method: RawApiDocsMethod): Promise<ApiDocsMethod> {
       throws.length === 0 ? undefined : await mdToHtml(throws.join('\n'), true),
     returns: returns.text,
     signature: await codeToHtml(formattedSignature),
-    examples: await codeGroupToHtml(examples),
+    examples: await codeGroupToHtml(examples, codeGroupId),
```
ST-DDT: <details><summary>Thoughts</summary>

I'm not sure whether we need globally unique ids here (especially ones with method indexes in the name).

e.g. 

- `utils-module-tab-1`
- `utils-method-2-tab-3`

IMO using the respective method name is a sufficient prefix.

- `utils-module-1` or `utils-module-tab-1` or `utils-module-example-1` or `example-1` or `example-tab-1`
- `get-default-ref-date-3` or `get-default-ref-date-tab-3` or `get-default-ref-date-example-3` or `getdefaultrefdate-...` 

I'm not exactly sure which one would "work" best.
Since these aren't user facing, we should keep the logic simple.

</details>

```suggestion
    examples: await codeGroupToHtml(examples, name),
```

That way, we don't need the extra parameter.

Or am I missing some potential collisions here?

ST-DDT (commented):
Nice implementation. It works well and I consider this ready to be merged.

Thanks for contributing! ❤️

test/scripts/shared/markdown.spec.ts
UlikGames: Yes, that's what the last three assertions are for. Those attributes sit inside the code example itself, so they have to survive untouched, and the `[^>]*` in the patterns can never cross a `>` to reach them.

Added a comment to separate the two halves.

scripts/apidocs/output/page.ts
UlikGames: You're right, the index isn't buying anything. Ids only have to be unique inside one page, and method names already are, since they're the keys of the `Object.fromEntries` map in `writePageData`. The module group is `${camelTitle}-module` and method names can't contain a `-`, so those can't clash either.

Applied the suggestion and dropped the parameter. Ids now read like `getDefaultRefDate-tab-1`. I regenerated the docs and checked every page for duplicates: `utils.ts` was the one that used to collide, two methods both rendering `tab-1..3`, and it's clean now.

ST-DDT (approved):

Shinigami92 (approved):

PR #3977, opened 2026-08-03
fix(color): give LCH hue the full 360 range and stop scaling HSLA alpha

### LINKED ISSUE #3956: fix(color): LCH hue range capped at 230 instead of 360, HSLA alpha incorrectly scaled

UlikGames:
Fixes #3956

Two bugs in the color module are now fixed:

1. The `lch()` function once capped hue at 230, reusing chroma's limit. But hue is an angle and should go from 0 to 360. Before, random values never passed 230. Now, hue uses the full 0–360 range (old values are adjusted by multiplying by 360/230). Lightness and chroma stay unchanged.

2. In `toCSS` for `hsla`, alpha was incorrectly converted to a percentage. For example, `0.6` turned into `60`. But `hsl()` needs the decimal value directly (like `/ 0.6`). This made it differ from `rgba`. Now, alpha stays as a decimal.

Tests are updated. Old tests limited all `lch()` parts to 230, which reflected the bug. Now, each component has proper bounds, plus a check that hue can go above 230. The `hsla` test keeps its regex but now also checks that alpha stays between 0 and 1, since `\d\.?\d` accepted both `60` and `0.6`.

All preflight steps passed: locale sync, docs, formatting, linting, build, full test suite, and type check. One test still fails - `test/docs/versions.spec.ts` - because it reads git tags and I used `--depth 1` when cloning. It fails the same way even without my changes. Total passing tests: 52,738.


ST-DDT:
Please handle the two separate issues in two separate PRs. This simplifies reviews and changelogs.

Issue #1749, opened 2023-01-17
Run seeded tests for all locales


ST-DDT:
### Clear and concise description of the problem

Currently the seeded tests only run for the `en` locale.

### Suggested solution

Run the seeded tests for all locales.

Pro:

- Generate visible examples for all locales

Con:

- Might cause a review overload due to the amount of generated test snapshots

### Alternative

Do nothing extra

### Additional context

The discussion originally came up in https://github.com/faker-js/faker/pull/1735#discussion_r1071751912

ST-DDT:
For now, I'm kind of against adding this as this reduces the probability that someone actually reviews the test snapshots.

ST-DDT:
**Team decision**

We will change the seeded tests to run all tests for all faker instances and verify, but only the main faker instance will write test snapshots.
This will render the all_functional.spec.ts obsolete.

ST-DDT:
**Team decision**

We will change the seeded tests to run all tests for all faker instances and verify, but only the main faker instance will write test snapshots.
This will render the all_functional.spec.ts obsolete.

xDivisionByZerox:
Is this required for v8? I could be implemented in v8-x as well, no?

ST-DDT:
I think its just nice to have.

jackspiece:
Hi, I would like to work on this. I checked the current `next` branch: `seededTests` still exercises only the passed Faker instance, while `test/all-functional.spec.ts` covers the locale instances separately.

My plan is to have each generated seeded case exercise every non-base entry in `allFakers`, keep snapshots only for the passed Faker instance so the snapshot set does not balloon, and remove `all-functional.spec.ts` once the coverage is equivalent. Is that still the direction you want?

jackspiece:
I ran a local spike against the current suite. Applying every parameterized seeded case to every locale creates false failures for valid locale differences, for example four-character word filters in `zh_CN` and abbreviated date forms where a locale does not define abbreviations.

A narrower replacement is green locally: run every no-argument seeded case against all non-base Faker instances, keep snapshots on the passed instance, and retain the existing location/person non-applicable combinations as typed exceptions. Parameterized cases stay on the snapshot instance. This preserves the meaningful coverage from `all-functional.spec.ts` while allowing that 37k-test file to be removed.

Does that scope match what you want today?

ST-DDT:
>  This preserves the meaningful coverage from all-functional.spec.ts while allowing that 37k-test file to be removed.

[...]

PR #3978, opened 2026-08-03
fix(color): give the LCH hue its full 360 degree range

### LINKED ISSUE #3956: fix(color): LCH hue range capped at 230 instead of 360, HSLA alpha incorrectly scaled

UlikGames:
Part of #3956 - this one tackles the first of two bugs mentioned, while the alpha fix is in another PR, just like asked.

So here’s the deal: the `lch()` function used to cap hue at 230, borrowing chroma’s limit. But hue’s an angle - it should spin the full 0 to 360. Before, you’d never get random hues past 230. Now? Full circle. Old values are scaled up by 360/230 to keep things consistent. Lightness and chroma? No changes there - they’re still playing by the same rules.

Tests got a refresh too. The old ones capped everything at 230, which actually mirrored the bug. Now each component has its proper range, and we’ve added a check to make sure hue can definitely go above 230 - no more holding back.

Three seeded snapshots changed, but only in hue - all updated using that same 360/230 multiplier. Draw order? Still totally untouched.

All preflight checks passed: locale sync, docs, formatting, linting, build, full test run, type checks - you name it. Only one test still flakes: `test/docs/versions.spec.ts`. It fails trying to read git tags because I cloned with `--depth 1`, but guess what? It fails the exact same way even without my changes. So yeah, 52,733 tests passing - solid.


src/modules/color/module.ts
```diff
@@ -897,10 +897,13 @@ export class ColorModule extends ModuleBase {
   }): string | number[];
   lch(options: { format?: ColorFormat } = {}): string | number[] {
     const { format = 'decimal' } = options;
-    const lch = [this.faker.number.float({ multipleOf: 0.000001 })];
-    for (let i = 0; i < 2; i++) {
-      lch.push(this.faker.number.float({ max: 230, multipleOf: 0.1 }));
-    }
+    const lch = [
+      this.faker.number.float({ multipleOf: 0.000001 }),
```
ST-DDT, line 902: Might as well add a comment to this one.

test/modules/color.spec.ts
```diff
@@ -323,25 +323,32 @@ describe('color', () => {
...
+          expect(chroma).toBeGreaterThanOrEqual(0);
+          expect(chroma).toBeLessThanOrEqual(230);
+          expect(hue).toBeGreaterThanOrEqual(0);
+          expect(hue).toBeLessThanOrEqual(360);
+        });
+
+        it('should be able to return a hue above the chroma bound', () => {
+          const hues = Array.from({ length: 1000 }, () => faker.color.lch()[2]);
+          expect(Math.max(...hues)).toBeGreaterThan(230);
         });
```
ST-DDT: @faker-js/members What do you think? Do we need this test?

src/modules/color/module.ts
UlikGames: Added.

test/modules/color.spec.ts
UlikGames: I've added this test because the earlier version only checked that hue was <= 360, which would still be true if hue was mistakenly capped at 230 again. Only the snapshot tests would catch that regression, and those are much more likely to be accidentally updated (for example, when rebasing). This test uses a probabilistic check: it samples 1000 values, and for each of them, checks that it's not capped. The odds that all 1000 would randomly fall below 230 by accident is (230/360)^1000, about 1e-195, so we don’t need to worry about false alarms. If you prefer, I can remove the probabilistic check, of course.

ST-DDT: My "problem" with the test is, that is just checks for being bigger than an "arbritatary" number.
e.g. updating the implementation and using 240 or 300 as the max will pass the test as well.

<details>
<summary>Side-Note</summary>

This kind of plays into the following issue/although it is kind of the oppsosite as well:

- #1749

Instead of checking it for all locales, we would also check for the boundries as well. aka generate all min values and generate all max values

</details>

ST-DDT (approved):

test/modules/color.spec.ts
UlikGames: Great catch, I updated the test so that it would pass either 240 or 300, but prefer it this way as it's closer to the boundaries. The test now follows the same pattern as the number's lower and upper value tests in number.spec.ts. To achieve that, I pushed the randomizer to the edges of 0 and MERSENNE_MAX_VALUE and fixed the expected values for lch(). This means we should now see [0, 0, 0] at 0, and roughly [1, 230, 360] at the maximum value. As a sanity check, I made sure that it fails when max is incorrectly set to 300.

This also addresses your comment in #1749 about testing the full min/max value generation. Though right now it's only lch(), but at least it's a solid start.

Shinigami92 (approved):

PR #3980, opened 2026-08-04
fix(locale): Remove deprecated phone number formats for nl_BE

ikbenwasmachine:
Belgian phonenumbers are 8 digits with prefix:
+32 OR +324
Removed other formats


matthewmayer:
Can you review the other format files in https://github.com/ikbenwasmachine/faker/tree/next/src/locales/nl_BE/phone_number/format and see if any of them need corresponding changes too?

ST-DDT (changes requested):
CI seems to be failing. Please fix it using `pnpm run preflight`.

PR #3984, opened 2026-08-08
infra: ensure markdown rendered is only ever initialized once

ST-DDT:
Implement suggestion from https://github.com/shikijs/shiki/discussions/1289

- https://github.com/shikijs/shiki/discussions/1289
- https://github.com/faker-js/faker/actions/runs/27549969244/job/81433308467

---

Store the promise to ensure we only ever initialize the markdown rendered backend once.
This is an attempt to fix the flaky markdown rendered test.

While I don't think that it is the cause for the failure, it does fix the unintended side effects of calling `initMarkdownRenderer()` multiple times.

### DIFF
--- a/scripts/shared/markdown.ts
+++ b/scripts/shared/markdown.ts
@@ -4,15 +4,18 @@ import { createMarkdownRenderer } from 'vitepress';
 import vitepressConfig from '../../docs/.vitepress/config';
 import { FILE_PATH_API_DOCS } from './paths';
 
+let markdownPromise: Promise<MarkdownRenderer>;
 let markdown: MarkdownRenderer;
 
 export async function initMarkdownRenderer(): Promise<void> {
   // eslint-disable-next-line unicorn/no-top-level-assignment-in-function
-  markdown ??= await createMarkdownRenderer(
+  markdownPromise ??= createMarkdownRenderer(
     FILE_PATH_API_DOCS,
     vitepressConfig.markdown,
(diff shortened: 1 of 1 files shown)


scripts/shared/markdown.ts
```diff
@@ -4,15 +4,18 @@ import { createMarkdownRenderer } from 'vitepress';
 import vitepressConfig from '../../docs/.vitepress/config';
 import { FILE_PATH_API_DOCS } from './paths';
 
+let markdownPromise: Promise<MarkdownRenderer>;
 let markdown: MarkdownRenderer;
 
 export async function initMarkdownRenderer(): Promise<void> {
   // eslint-disable-next-line unicorn/no-top-level-assignment-in-function
-  markdown ??= await createMarkdownRenderer(
```
Shinigami92, line 11: **question:** isn't `??=` short circuit? so if I'm not wrong, this change you made does nothing different, does it?

ST-DDT: Yes, but you can create multiple promises while the first one awaits.

````txt
async createMarkdownRenderer
async createMarkdownRenderer
async createMarkdownRenderer
set markdown from any awaited
createMarkdownRenderer is never called again
````

If you think the change is negligible, we can close it.

Shinigami92 (approved):

xDivisionByZerox (approved):

Issue #2518, opened 2023-10-31
CIF generator


randreu28:
### Clear and concise description of the problem

I need a CIF generator. [CIF](https://es.wikipedia.org/wiki/C%C3%B3digo_de_identificaci%C3%B3n_fiscal) is the tax identification code for businesses in spain. Similar to the [LEI identifier](https://www.gleif.org/en/about-lei/introducing-the-legal-entity-identifier-lei), but particulary for spanish businesses only.

### Suggested solution

Create a method similar to this:

```ts
const cif = faker.company.cif();
console.log(cif); //F19262039
```

Some real examples of CIF validators: https://www.generador-de-dni.com/generador-de-dni

matthewmayer:
This could potentially be localizable for example have

faker.company.number()

And then for the es locale it gives a CIF, for the en_GB locale it gives a UK Company number etc.

xDivisionByZerox:
> [...] Similar to the LEI identifier, but particulary for spanish businesses only. [...]

We support over 60 locales. Please tell me how other locales would use this. What is the expected behavior, for example, for `fakerEN.company.cif()`? `en` is our default locale. So it must not error on functions with an "not applicable" error.

matthewmayer:
A generic en locale could probably return a DUNS number. That's fairly standard internationally.

matthewmayer:
Ah there's actually an open issue for DUNS number also https://github.com/faker-js/faker/issues/1202

matthewmayer:
Something like `faker.helpers.fromRegExp("[ABCDEFGHJNPQRSUVW][0-9]{7}[0-9A-J]")` should work as a workaround for making CIF numbers in the meantime.

Issue #1202, opened 2022-07-27
Add DUNS number generation


import-brain:
### Clear and concise description of the problem

[DUNS (Data Universal Numbering System)](https://en.wikipedia.org/wiki/Data_Universal_Numbering_System) numbers are unique, nine-digit codes that helps identify businesses. Users of the system include Apple, the US government, the UN, and more. These numbers are standardized worldwide, and it is standard for businesses to have them.

We already have generation for SWIFT codes (banks), IBAN codes (bank accounts), vehicle identification numbers, mongodb object IDs, etc., we should add a company ID generation method. DUNS numbers would be perfect for this, because they are widely used as a public business entity identifier.

I can submit a PR for this issue if it is accepted.

### Suggested solution

As DUNS numbers are nine-digit randomly generated numbers, this method can be written as a wrapper around `random.numeric`.

In the `company` module, the method should return as follows:

`faker.company.duns() // "060704780"`
`faker.company.duns() // "059501224"`

xDivisionByZerox:
What options could be provided for this method? 

I was thinking about:

## format
Type: `'basic' | 'DUNS+4'`
Default: `'basic'`
Description: Defines the results format.

Description for `'basic'`: A basic 9-digit DUNS string.
Example for `'basic'`: `'192837465'`.

Description for `'DUNS+4'`: A 9-digit DUNS string with four extra alphanumeric characters at the end.
Example for `'DUNS+4'`: `'192837465LJ20'`

## readable
Type: `boolean`
Default: `false`
Description: Defines if the results should be easier readable for humans.

Example for `false`: `'192837465'`.
Example for `true`: `'19-283-7465'`

ST-DDT:
Waiting for user interest.

import-brain:
> What options could be provided for this method?
> 
> I was thinking about:
> 
> ## format
> Type: `'basic' | 'DUNS+4'` Default: `'basic'` Description: Defines the results format.
> 
> Description for `'basic'`: A basic 9-digit DUNS string. Example for `'basic'`: `'192837465'`.
> 
> Description for `'DUNS+4'`: A 9-digit DUNS string with four extra alphanumeric characters at the end. Example for `'DUNS+4'`: `'192837465LJ20'`
> 
> ## readable
> Type: `boolean` Default: `false` Description: Defines if the results should be easier readable for humans.
> 
> Example for `false`: `'192837465'`. Example for `true`: `'19-283-7465'`

[...]

PR #3985, opened 2026-08-09
feat(finance): add vatNumber()

rodrigobnogueira:
Adds `faker.finance.vatNumber()`, generating VAT identification numbers for the EU member states.

## Why

Two open issues ask for pieces of this, and neither has an implementation:

- **#2518 — CIF generator** (7 👍). Spain's CIF *is* its VAT identification number, so `vatNumber({ countryCode: 'ES' })` is exactly what that issue asks for. @matthewmayer suggested there that the answer should be one localizable method rather than a Spain-only `cif()`, and @xDivisionByZerox' constraint on that thread — *"We support over 60 locales… `en` is our default locale. So it must not error on functions with a 'not applicable' error"* — is what drove the design below.
- **#1202 — DUNS** — where @camilleterol [asked specifically for VAT numbers](https://github.com/faker-js/faker/issues/1202#issuecomment-3377605446) in October and got no reply: *"many countries issue government/regulatory IDs, and it would be great (and quite useful to us) if Faker could generate those as well (SIREN/SIRET in France for example, VAT number, …)"*.

This PR covers the government **tax** identifier half of that. It deliberately does **not** touch DUNS, which is a Dun & Bradstreet commercial identifier rather than a government one — conflating the two is what makes "what should this return for DE?" unanswerable.

Concretely, the gap today is that a test fixture needing a VAT ID has to hardcode `DE999999999` or fill the field with `faker.string.sample()` noise.

## Design — it's `iban()` with a different table

The method is deliberately a near-copy of the existing `iban()`, so there is no new pattern to review:

```ts
const vatFormat = countryCode
  ? vatNumberFormats.find((f) => f.country === countryCode)
  : this.faker.helpers.arrayElement(vatNumberFormats);

if (!vatFormat) {
  throw new FakerError(`Country code ${countryCode} not supported.`);
}

return `${vatFormat.country}${this.faker.helpers.replaceSymbols(vatFormat.format)}`;
```

That choice answers the 60-locale constraint by construction: the method never reads `faker.locale` or `faker.definitions`, so it behaves identically in every locale, needs **no locale data**, and cannot raise a "not applicable" error. `fakerEN`, `fakerDE` and `fakerJA` all work. No `BROKEN_LOCALE_METHODS` entry is needed — `all-functional.spec.ts` passes across all locales unchanged.

Formatting reuses the existing `faker.helpers.replaceSymbols()` (`#` digit, `?` letter, literals verbatim). I did not use `replaceCreditCardSymbols()` because its `L` placeholder hard-codes Luhn, which fits almost no VAT scheme.

```ts
faker.finance.vatNumber() // 'SK4318759382'
faker.finance.vatNumber({ countryCode: 'DE' }) // 'DE644073457'
faker.finance.vatNumber({ countryCode: 'NL' }) // 'NL840351580B96'
```

## Scope: structure only, no check digits

Each entry models length, character classes and mandated literals (Austria's `U`, the Netherlands' `B`, Belgium's leading `0`). Where a country's real scheme defines a check digit, it is random here, and the JSDoc says so plainly.

That matches the prior art — faker-ruby's `Faker::Finance.vat_number` is likewise template-driven with no checksums — and it matches what the ecosystem actually verifies. I measured this rather than assumed it: 200 generated values per country through `validator`'s `isVAT` gives **100% for 26 of the 27 member states**, because `isVAT` is a structural check for all of them.

The exception is **Portugal (7%)**, which really is checksum-verified, so it is left out rather than shipped knowingly broken. Same for non-EU schemes that are checksum-verified — `CH` (9%) and `AU` (2%). Those want real check-digit computation, which can arrive later behind an option without changing this signature or its default behaviour.

Greece is included as `EL`, its VAT prefix, not the `GR` ISO code.

## Tests

`test/modules/finance-vat-number.spec.ts` mirrors `finance-iban.spec.ts`: every country in the table is generated and asserted against `validator`'s `isVAT`. The oracle is third-party, and since this PR ships no checksum code of its own there is nothing for the test to be circular with. Plus seeded snapshots in `finance.spec.ts` (noArgs, a numeric country, a country with letters), an unsupported-country `FakerError` case, and a guard that Greece stays on `EL`.

`pnpm run preflight` passes: 53,039 tests, no type errors.

## Cost

~70 lines of table, zero algorithm code, no locale files, no definitions change, no new test-matrix carve-outs — smaller than the `iban()` feature it is modelled on.

One note on timing: I see #3857 is migrating modules to standalone functions. Finance's source isn't converted yet, so this follows the current class shape; happy to rebase onto the new structure if that lands first, or to hold this until it does.


test/modules/finance-vat-number.spec.ts
ST-DDT, line 1: IMO you can move this to the finance spec file.

src/modules/finance/vat-number.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Only the structure of each number is modelled: lengths, character classes and
+ * the literal characters that a given country mandates. Check digits, where a
+ * country's real algorithm defines one, are filled with random data.
+ *
+ * Portugal is deliberately absent: its check digit is verified by common
+ * validators, so a structurally-correct-but-random Portuguese number would be
+ * rejected by them. The same applies to non-EU schemes such as `CH` and `AU`.
+ * Those belong to a follow-up that computes real check digits.
+ */
+const vatNumberFormats: VatNumberFormat[] = [
```
ST-DDT: Maybe we should use Record<Country, Format>, as these are easier to look up.
We can keep the name as comment.

src/modules/finance/module.ts
```diff
@@ -841,6 +842,49 @@ export class FinanceModule extends ModuleBase {
...
+      ? vatNumberFormats.find((f) => f.country === countryCode)
+      : this.faker.helpers.arrayElement(vatNumberFormats);
+
+    if (!vatFormat) {
+      throw new FakerError(`Country code ${countryCode} not supported.`);
+    }
+
+    // The country prefix is always included: France is the only member state whose
+    // number is not also valid without it, so omitting it would break FR alone.
+    return `${vatFormat.country}${this.faker.helpers.replaceSymbols(vatFormat.format)}`;
```
ST-DDT: If you push the country code into the format pattern, then we could create an alias/entry for GR.
This would also help clarifying which countryCode you have to use aka ISO.
Or is there a better name for which kind of countryCode to use?

src/modules/finance/module.ts
```diff
@@ -841,6 +842,49 @@ export class FinanceModule extends ModuleBase {
...
+   *
+   * @since 10.6.0
+   */
+  vatNumber(
+    options: {
+      /**
+       * The country code from which you want to generate a VAT number,
+       * if none is provided a random country will be used.
+       */
+      countryCode?: string;
```
ST-DDT: I'm thinking whether we should use `string`/`LiteralUnion<VatCountryCode>`/`VatCountryCode` as parameter type here.

src/modules/finance/module.ts
```diff
@@ -841,6 +842,49 @@ export class FinanceModule extends ModuleBase {
...
+   * faker.finance.vatNumber({ countryCode: 'DE' }) // 'DE644073457'
+   * faker.finance.vatNumber({ countryCode: 'NL' }) // 'NL840351580B96'
+   *
+   * @since 10.6.0
+   */
+  vatNumber(
+    options: {
+      /**
+       * The country code from which you want to generate a VAT number,
+       * if none is provided a random country will be used.
```
matthewmayer: i think it would be good to explicitly list all supported country codes, so that consumers reading the docs dont have to go hunting in the code to see if a country code is supported.

src/modules/finance/vat-number.ts
```diff
@@ -0,0 +1,109 @@
...
+} satisfies Record<string, string>;
+
+/**
+ * ISO 3166-1 alpha-2 codes that identify a country whose VAT numbers carry a
+ * different prefix, mapped to the entry in {@link vatNumberFormats} that
+ * produces them.
+ */
+export const vatNumberAliases = {
+  GR: 'EL',
+} satisfies Record<string, keyof typeof vatNumberFormats>;
```
ST-DDT: IMO we can inline this into the main record.

src/index.ts
```diff
@@ -66,6 +66,7 @@ export type {
   BitcoinAddressFamilyType,
   BitcoinNetworkType,
 } from './modules/finance/bitcoin';
+export type { VatNumberCountryCode } from './modules/finance/vat-number';
```
ST-DDT: Please re-export  from the module index.

test/modules/finance.spec.ts
```diff
@@ -605,6 +613,64 @@ describe('finance', () => {
...
+        );
+
+        it('should return a structurally valid Spanish NIF/CIF', () => {
+          const actual = faker.finance.vatNumber({ countryCode: 'ES' });
+
+          // The leading character encodes the legal form (I, K, L, M, O, T and
+          // X-Z are not assigned to one) and the control character is a digit
+          // for national entities or a letter A-J for the rest.
+          expect(actual).toMatch(/^ES[ABCDEFGHJNPQRSUVW]\d{7}[0-9A-J]$/);
+        });
```
ST-DDT: IMO we don't need this test, as it tests using the same pattern as the input and only for one country.

Maybe run a loop over all country codes to ensure all can be generated (and match their respective patterns).
Or extend the above loop.

test/modules/finance.spec.ts
```diff
@@ -605,6 +613,64 @@ describe('finance', () => {
...
+
+        it('should only use characters the country actually issues', () => {
+          expect(faker.finance.vatNumber({ countryCode: 'SE' })).toMatch(/01$/);
+          expect(faker.finance.vatNumber({ countryCode: 'SI' })[2]).not.toBe(
+            '0'
+          );
+          expect(faker.finance.vatNumber({ countryCode: 'RO' })[2]).not.toBe(
+            '0'
+          );
+        });
```
ST-DDT: Same here

test/modules/finance.spec.ts
```diff
@@ -605,6 +613,64 @@ describe('finance', () => {
         });
       });
 
+      describe('vatNumber()', () => {
+        // Spain is checked separately: `validator` requires a letter in the last
+        // position, but the Spanish tax agency assigns a digit there to national
+        // legal entities, so it rejects real numbers such as `ESA28015865`.
```
ST-DDT: Could you please report this issue to validatorjs?
Please link the respective issue or PR here for ease of tracking.

test/modules/finance.spec.ts
```diff
@@ -78,6 +80,12 @@ describe('finance', () => {
         });
     });
 
+    t.describe('vatNumber', (t) => {
+      t.it('noArgs')
+        .it('with countryCode option', { countryCode: 'DE' })
+        .it('with countryCode option using letters', { countryCode: 'ES' });
```
ST-DDT: IMO this second case is redundant as it is basically the same as the one above.

src/modules/finance/vat-number.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Each pattern is written for `faker.helpers.fromRegExp()`. Only the structure
+ * of a number is modelled: its length, the characters each position may hold,
+ * and the literals a country mandates. Check digits, where a country's real
+ * algorithm defines one, are filled with random data from the permitted
+ * character set.
+ *
+ * Portugal is deliberately absent: its check digit is verified by common
+ * validators, so a structurally-correct-but-random Portuguese number would be
+ * rejected by them. The same applies to non-EU schemes such as `CH` and `AU`.
+ * Those belong to a follow-up that computes real check digits.
```
ST-DDT: In the first section you claim that check digits are replaced with random data. The next part claims that the countries are skipped because they would require check digits.

IMO we should either skip them (for now) or include them with randomized check digits.
Not sure which one to prever right now.

src/modules/finance/module.ts
```diff
@@ -841,6 +844,67 @@ export class FinanceModule extends ModuleBase {
...
+   * Please note that only the structure of the number is generated: its length, the characters each
+   * position may hold, and any literal characters the country mandates. Where a country's real
+   * numbering scheme defines a check digit, that digit is random here, so the result is not
+   * guaranteed to be a valid registration.
+   *
+   * The supported country codes are
+   * `AT`, `BE`, `BG`, `CY`, `CZ`, `DE`, `DK`, `EE`, `EL` (or `GR`), `ES`, `FI`, `FR`, `HR`, `HU`,
+   * `IE`, `IT`, `LT`, `LU`, `LV`, `MT`, `NL`, `PL`, `RO`, `SE`, `SI` and `SK`.
+   * Portugal is not supported, because a Portuguese number is only recognisable with a real check
+   * digit, which this method does not compute.
```
ST-DDT: In the first section you claim that check digits are replaced with random data. The next part claims that the countries are skipped because they would require check digits.

IMO we should either skip them (for now) or include them with randomized check digits.
Not sure which one to prever right now.

src/modules/finance/module.ts
```diff
@@ -841,6 +844,67 @@ export class FinanceModule extends ModuleBase {
...
+  vatNumber(
+    options: {
+      /**
+       * The ISO 3166-1 alpha-2 code of the country you want a VAT number for,
+       * if none is provided a random supported country will be used.
+       */
+      countryCode?: LiteralUnion<VatNumberCountryCode>;
+    } = {}
+  ): string {
+    const { countryCode } = options;
```
ST-DDT:
```suggestion
    const { countryCode = this.faker.helpers.objectKey(vatNumberFormats) } = options;
```

src/modules/finance/module.ts
```diff
@@ -841,6 +844,67 @@ export class FinanceModule extends ModuleBase {
...
+   *
+   * @since 10.6.0
+   */
+  vatNumber(
+    options: {
+      /**
+       * The ISO 3166-1 alpha-2 code of the country you want a VAT number for,
+       * if none is provided a random supported country will be used.
+       */
+      countryCode?: LiteralUnion<VatNumberCountryCode>;
```
ST-DDT: Since we always throw if countryCode is anything but `VatNumberCountryCode` it may be worth limiting it by more strict types. Not entirely sure yet though.

rodrigobnogueira:
Fixed in af4fc99 and d1a8555: a prototype-chain hole (`countryCode: 'toString'` returned `''`), `GR` moved back out to an alias map because inlining it doubled Greece's odds on a no-arg call, `countryCode` strictly typed to match `system.networkInterface()`, and four countries that were emitting unissuable numbers (ES, CY, NL, PT).

Spanish false negative reported upstream as validatorjs/validator.js#2846.

ST-DDT:
Can you please turn down the verbosity of your responses?

src/modules/finance/vat-number.ts
rodrigobnogueira: Tried that, but it made Greece come out twice as often as any other country on a no-arg call, since objectKey draws over the table's keys, so GR is a separate alias map. Let me know if you prefer it inlined.

ST-DDT:

Shinigami92 (dismissed):

src/modules/finance/_vat-number.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * ISO 3166-1 alpha-2 codes that are accepted as input but are not themselves a
+ * key above, because the country's numbers carry a different prefix. Kept out
+ * of {@link vatNumberFormats} so that a call without a country code draws
+ * uniformly across countries rather than twice as often for Greece.
+ */
+export const vatNumberCountryCodeAliases = {
+  GR: 'EL',
+} as const satisfies Record<string, keyof typeof vatNumberFormats>;
```
ST-DDT: As mentioned before we should add this to the main record.
We can avoid the EL/GR over-representation by using

````ts
export const vatNumberCountryCodes = Object.keys(vatNumberFormats).filter((e) => e != 'GR') as VatNumberCountryCode[];

const { countryCode = arrayElement(vatNumberCountryCodes) } = options;
````

File: src/modules/finance/_vat-number.ts [hunk 16]
```diff
@@ -0,0 +1,130 @@
...
+ * occurs in reality. Only the structure of a number is
+ * modelled: its length, the characters each position may hold, and the
+ * literals a country mandates. Check characters, where a country's real
+ * algorithm defines one — a digit in most, a letter in Cyprus, Ireland and
+ * part of Spain — are filled with random data from the permitted character
+ * set, so a validator that recomputes them rejects nearly all of this output.
+ */
+export const vatNumberFormats = {
+  /** UID-Nummer. */
+  AT: 'ATU[0-9]{8}',
```
ST-DDT, line 11: @faker-js/members Should we explicitly convert this to regex instances to make it more clear what this is?

src/modules/finance/_vat-number.ts
```diff
@@ -0,0 +1,130 @@
+/**
+ * The VAT identification number patterns of the EU member states, keyed by the
+ * two-letter code the numbers themselves carry — the ISO 3166-1 alpha-2 code
+ * everywhere except Greece, whose numbers use `EL` rather than `GR`. `GR` is
+ * still accepted as input, through {@link vatNumberCountryCodeAliases}.
```
ST-DDT:
```suggestion
 * everywhere except Greece, whose numbers use `EL` rather than `GR`.
```

src/modules/finance/_vat-number.ts
```diff
@@ -0,0 +1,130 @@
...
+ * shapes are listed separately, it is because one character class would pair
+ * independent positions freely and emit combinations the country never issues.
+ * Rarer forms are left unmodelled and the per-country comment says so; within a
+ * country the listed shapes are drawn evenly, which is not how often each
+ * occurs in reality. Only the structure of a number is
+ * modelled: its length, the characters each position may hold, and the
+ * literals a country mandates. Check characters, where a country's real
+ * algorithm defines one — a digit in most, a letter in Cyprus, Ireland and
+ * part of Spain — are filled with random data from the permitted character
+ * set, so a validator that recomputes them rejects nearly all of this output.
```
ST-DDT: IMO this is very verbose.

Maybe like this?

```suggestion
 * Each pattern is written for `faker.helpers.fromRegExp()`.
 * Currently, all values are generated randomly, so parts with intend such as check digits will likely produce invalid values.
```

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
...
+   * numbering scheme defines a check digit, that digit is random here, so the result is not
+   * guaranteed to be a valid registration.
+   *
+   * The supported country codes are the EU member states, using the two-letter code each
+   * country's numbers carry:
+   * `AT`, `BE`, `BG`, `CY`, `CZ`, `DE`, `DK`, `EE`, `EL` (or `GR`), `ES`, `FI`, `FR`, `HR`, `HU`,
+   * `IE`, `IT`, `LT`, `LU`, `LV`, `MT`, `NL`, `PL`, `PT`, `RO`, `SE`, `SI` and `SK`.
+   *
+   * @param options An options object.
+   * @param options.countryCode The two-letter code of the country you want a VAT number for. Greece may be given as either `GR` or `EL`; the generated number always carries the `EL` prefix that Greek VAT numbers use. Defaults to a random supported country.
```
ST-DDT:
```suggestion
   * @param options.countryCode The two-letter code of the country you want a VAT number for.
   * Greece may be given as either `GR` or `EL`.
   * Defaults to a random supported country.
```

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
...
+   * faker.finance.vatNumber({ countryCode: 'GR' }) // 'EL892156043'
+   *
+   * @since 11.0.0
+   */
+  vatNumber(
+    options: {
+      /**
+       * The two-letter code of the country you want a VAT number for. Greece
+       * may be given as either `GR` or `EL`; the generated number always
+       * carries the `EL` prefix that Greek VAT numbers use.
```
ST-DDT:
```suggestion
       * The two-letter code of the country you want a VAT number for.
       * Greece may be given as either `GR` or `EL`.
```

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
...
+       *
+       * @default faker.helpers.objectKey(vatNumberFormats)
+       */
+      countryCode?: VatNumberCountryCode;
+    } = {}
+  ): string {
+    // Defaulted here rather than checked for truthiness later, so that an
+    // explicitly passed empty string is reported as unsupported instead of
+    // silently behaving like an omitted option.
```
ST-DDT: Too verbose

```suggestion
```

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
...
+    const { countryCode = this.faker.helpers.objectKey(vatNumberFormats) } =
+      options;
+
+    // Both lookups go through `Object.hasOwn`, because a plain index resolves
+    // an inherited key such as `toString` to `Object.prototype.toString`,
+    // which is truthy and would pass a `!= null` guard. `countryCode` is a
+    // closed union, so only JavaScript callers (or a cast) get there.
```
ST-DDT: Too verbose


```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
         });
       });
 
+      describe('vatNumber()', () => {
+        // Every entry of `vatNumberFormats` is covered by the pattern loop
+        // below. These two are additionally excluded from `validator` (13.15),
+        // which rejects real Spanish numbers whose control character is a
+        // digit, such as `ESA28015865` (reported as
+        // validatorjs/validator.js#2846), and recomputes the Portuguese check
+        // digit, which is random here.
```
```suggestion
        // Exclude Spanish due to validatorjs/validator.js#2846
        // Exclude Portuguese because it verifies the check digit, which is random here.
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+
+        it.each(CHECKED_BY_VALIDATOR)(
+          'should return a valid VAT number for %s',
+          (country) => {
+            const actual = faker.finance.vatNumber({ countryCode: country });
+
+            // Not redundant: nearly every validator matcher makes the prefix
+            // optional — FR is the exception — so `isVAT('123456789', 'DE')`
+            // is true on its own.
+            expect(actual).toStartWith(country);
```
```suggestion
            expect(actual).toStartWith(country); // Validator treats this as optional
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+            expect(actual).toSatisfy((value: string) =>
+              isVAT(value, country as VATCountryCode)
+            );
+          }
+        );
+
+        // `fromRegExp` understands a subset of regex and copies anything
+        // outside it into the output verbatim. This loop proves every pattern
+        // stays inside that subset; it says nothing about a pattern being
+        // right for its country, since it checks each pattern against itself.
```
ST-DDT: Already covered by test name.


```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+        it.each(Object.entries(vatNumberFormats))(
+          'should expand the %s pattern rather than emit it literally',
+          (countryCode, patterns) => {
+            const alternatives = [patterns].flat();
+            // Enough draws that every alternative of a multi-shape country is
+            // exercised, not just whichever one the first draw happens to pick.
+            const actuals = times(50).map(() =>
+              faker.finance.vatNumber({
+                countryCode: countryCode as VatNumberCountryCode,
+              })
```
ST-DDT: I see most of the calls use an explicit cast.

@faker-js/members Should we allow LiteralUnions here, like the first implementation did?

LiteralUnion is easier to write, but may throw if an invalid country code is passed.
But blindly adding a cast to each call will not prevent throwing either, it is just more verbose.

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+        // `fromRegExp` understands a subset of regex and copies anything
+        // outside it into the output verbatim. This loop proves every pattern
+        // stays inside that subset; it says nothing about a pattern being
+        // right for its country, since it checks each pattern against itself.
+        it.each(Object.entries(vatNumberFormats))(
+          'should expand the %s pattern rather than emit it literally',
+          (countryCode, patterns) => {
+            const alternatives = [patterns].flat();
+            // Enough draws that every alternative of a multi-shape country is
+            // exercised, not just whichever one the first draw happens to pick.
```
```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+            const actuals = times(50).map(() =>
```
ST-DDT: Instead of creating 50 and then looping for them, we could just use a for i..max loop.
Also I think 10 are enough. We have I think two patterns at most, so it should be sufficient to hit every pattern.

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+        // country's pattern would otherwise give that country double weight.
+        it('should key every format by the prefix its numbers carry', () => {
+          for (const [countryCode, patterns] of Object.entries(
+            vatNumberFormats
+          )) {
+            for (const pattern of [patterns].flat()) {
+              expect(pattern).toStartWith(countryCode);
+            }
+          }
+        });
```
ST-DDT: IMO this test doesn't add much safety.

```suggestion
```

If you disagree, just omit the comment.

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+
+        it('should return a VAT number of a supported country', () => {
+          const actual = faker.finance.vatNumber();
+          const country = actual.slice(0, 2);
+
+          expect(Object.keys(vatNumberFormats)).toContain(country);
+        });
```
ST-DDT, line 668: This one is kind of redundant with the one above. Keep either this one or the one above. The one above is stronger.

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+
+        it('should accept the GR ISO code and emit the EL prefix Greek numbers use', () => {
+          expect(faker.finance.vatNumber({ countryCode: 'GR' })).toStartWith(
+            'EL'
+          );
+        });
+
+        // `toString` guards the inherited-key path, which is reachable from
+        // JavaScript and used to resolve to `Object.prototype.toString`,
+        // returning an empty string instead of throwing.
```
```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+        // NL as \d{9}B\d{2}, RO as \d{2,10}, SE as \d{12} and SI as \d{8},
+        // and is unusable for ES and PT per the note above. AT and DE are the
+        // only two it checks as tightly as the table does, so they are not
+        // repeated here.
+        it.each([
+          ['BE', /^BE[01]\d{9}$/],
+          ['CY', /^CY[0134569]\d{7}[A-Z]$/],
+          // The entity class decides the control character, so the two are not
+          // independent: a Spanish legal entity takes a digit, a foreign
+          // entity or public body takes a letter.
```
```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+          }
+        );
+
+        // Stated independently of the source patterns, and only where
+        // `validator` is weaker than they are: it checks BE as \d{10}, CY as
+        // \w{9}, FR as [A-Z0-9]{2}\d{9}, IE as \d{7}\w(W)?, LT as \d{9,12},
+        // NL as \d{9}B\d{2}, RO as \d{2,10}, SE as \d{12} and SI as \d{8},
+        // and is unusable for ES and PT per the note above. AT and DE are the
+        // only two it checks as tightly as the table does, so they are not
+        // repeated here.
```
ST-DDT: If the validation inside validatorjs is lacking, then please report/fix it there, so that in the future everyone can use the tightened validation. IMO these regexp are a fairly low barrier, so they should be fine to add upstream.
Please link the upstream issue/PR here, once you have created them.

```suggestion
```

test/modules/finance.spec.ts
```diff
@@ -602,6 +611,163 @@ describe('finance', () => {
...
+
+            expect(
+              actuals.filter((actual) => secondShape.test(actual))
+            ).not.toEqual([]);
+            expect(
+              actuals.filter((actual) => !secondShape.test(actual))
+            ).not.toEqual([]);
+          }
+        );
+      });
```
ST-DDT: IMO this code only tests what is currently there. It does not check for new entries with multiple shapes or IMO we don't need a test for our data structure, so this test can be removed.

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
...
+    const resolvedCode = Object.hasOwn(vatNumberCountryCodeAliases, countryCode)
+      ? vatNumberCountryCodeAliases[
+          countryCode as keyof typeof vatNumberCountryCodeAliases
+        ]
+      : countryCode;
+
+    if (!Object.hasOwn(vatNumberFormats, resolvedCode)) {
```
ST-DDT: @faker-js/members Should we use this check (here and elsewhere) or is the normal pattern != null check sufficient?

src/modules/finance/module.ts
```diff
@@ -841,6 +845,74 @@ export class FinanceModule extends ModuleBase {
     return `${bankIdentifier}${countryCode}${locationCode}${branchCode}`;
   }
 
+  /**
+   * Generates a random VAT identification number for one of the EU member states.
+   *
+   * Please note that only the structure of the number is generated: its length, the characters each
+   * position may hold, and any literal characters the country mandates. Where a country's real
+   * numbering scheme defines a check digit, that digit is random here, so the result is not
+   * guaranteed to be a valid registration.
```
ST-DDT: Maybe shorten it like this:

```suggestion
   * Please note that this currently only generates the structure of the respective country's VAT identification.
   * But it will return random values for digits with intend such as check digits, so the result is likely to be invalid.
```

Maybe add this as an `@remark`.

ST-DDT (changes requested):

test/modules/finance.spec.ts
rodrigobnogueira: Upstream PR: validatorjs/validator.js#2876

ST-DDT: Thanks. Since the validation is now moved upstream, we can remove it here.
Or would you like to wait until is merged upstream?

ST-DDT (commented):
Could you please add this test:

`finance.spec-d.ts`

````ts
import type { VATCountryCode } from 'validator';
import { describe, expectTypeOf, it } from 'vitest';
import type { VatNumberCountryCode } from '../../src/modules/finance/_vat-number';

describe('finance', () => {
  describe('vatNumber', () => {
    it('VatNumberCountryCode', () => {
      expectTypeOf<VatNumberCountryCode>().toExtend<VATCountryCode | 'GR'>();
      // Currently the reverse is not true
    });
  });
});
````

The rest looks good to me.

I hope we get some feedback from the team for these points soon:

- data/optional: https://github.com/faker-js/faker/pull/3985#discussion_r3931661834
- public api: https://github.com/faker-js/faker/pull/3985#discussion_r3931767979
- impl consistency: https://github.com/faker-js/faker/pull/3985#discussion_r3931891934

File: test/modules/finance.spec.ts (hunk 33 above)
Comment (rodrigobnogueira) [reply test/modules/finance.spec.ts]: It's your call 🙂 Happy to drop them. Though the pinned 13.15.35 has none of it yet: [#2876](https://github.com/validatorjs/validator.js/pull/2876) isn't merged, and [#2849 ](https://github.com/validatorjs/validator.js/pull/2849)has been merged since August but is still unreleased.

PT and FR would need to stay either way: validator recomputes the PT check digit, and upstream keeps [A-Z0-9]{2} for FR.

Comment (ST-DDT) [reply test/modules/finance.spec.ts]: I think it not being released is not an issue.
I dont expect our implementation to change again soon.
The regexp are basically the same those in our data so they are kind of redundant.
Maybe taking a snapshot of each country would serve as protection against unindented changes in the patterns.

Lets drop the regexp checks then.
As for PT I think a Todo comment next to the exclusion to be checked with validatorjs is sufficient. Aka "implement check digit calculation for PT".
As for FR, if the arguments for checking more strictly arent enough for inclusion in validatorjs, then we dont need to include them in our validation either. Keeping them for generation is perfectly valid.

**Review by ST-DDT — DISMISSED** [review]
Looks good to me. See also my previous comment regarding points for discussion.

**Review by Shinigami92 — DISMISSED** [review]

File: test/modules/finance.spec.ts [hunk 37]
```diff
@@ -95,6 +107,12 @@ describe('finance', () => {
     });
   });
 
+  // A diff here means a pattern changed; check the country's numbering
+  // rules before accepting it.
+  it('should have stable VAT number patterns', () => {
+    expect(vatNumberFormats).toMatchSnapshot();
```
ST-DDT, line 113: I meant taking a snapshot of a generated value per country. Sorry for not being clear enough.

ST-DDT (dismissed):

Shinigami92 (dismissed):

src/modules/finance/module.ts
```diff
@@ -849,6 +853,60 @@ export class FinanceModule extends ModuleBase {
...
+   */
+  vatNumber(
+    options: {
+      /**
+       * The two-letter code of the country you want a VAT number for.
+       * Greece may be given as either `GR` or `EL`.
+       *
+       * @default faker.helpers.arrayElement(vatNumberCountryCodes)
+       */
+      countryCode?: VatNumberCountryCode;
```
xDivisionByZerox: **nitpick:** Should this not preferably use the locale of the current faker instance? For example: I would expect `fakerDE.finance.vatNumber()` to already use `countryCode: 'DE'` by default. If the current locale is an unsupported locale for the feature, it can still fall back to a random one.

src/modules/finance/_vat-number.ts
xDivisionByZerox: I would say no for two reasons:

1. It is only feed into `faker.string.fromRegex` which internally converts a regex to a representing string. The only feature that you would gain from real regex are flags, which are not used in this feature.
2. Since the only used for `fromRegex` AND this function explicitly states that it does NOT support all regex features, it could be misleading to directly place regexes in this map as this could give the impression that all features from regex are supported.

src/modules/finance/module.ts
xDivisionByZerox: **nitpick:** Should this not preferably use the locale of the current faker instance? For example: I would expect `fakerDE.finance.vatNumber()` to already use `countryCode: 'DE'` by default. If the current locale is an unsupported locale for the feature, it can still fall back to a random one.

test/modules/finance.spec.ts
xDivisionByZerox: I would not allow any string. We usually design APIs with a TS-first mindset. Allowing o to widen the type to an arbitrary string here feels wrong in regards t o the fundamental design principles we adhere to.

src/modules/finance/module.ts
xDivisionByZerox: How exactly would the countryCode variable be nullish at this point in the first place? I'm surprised the linter does not raise an error for the check being unnecessary (example [`ts-eslint/no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/)).

src/modules/finance/module.ts
xDivisionByZerox, line 890: **nitpick:** Should this not preferably use the locale of the current faker instance? For example: I would expect `fakerDE.finance.vatNumber()` to already use `countryCode: 'DE'` by default. If the current locale is an unsupported locale for the feature, it can still fall back to a random one.

ST-DDT: Should we do that via locale data e.g. `faker.definitions.finance.default_vat_country_code` or similar?
Or should we put that into `fakerCore.config`?

src/modules/finance/module.ts
ST-DDT: It is not about `countryCode` being `null` but `vatNumberFormats[countryCode]` being null.

Which happens when `countryCode` is not a valid `VatCountryCode`. Which is not possible in a pure TS based world, but we do check for sometimes (float, zipCode).

---

Current version of the code/PR for easier reference:

https://github.com/faker-js/faker/blob/edfcba13dcb68dbcdc5b50d0c177e2c1d20acb80/src/modules/finance/module.ts#L893-L907

xDivisionByZerox: I would say that these checks are not required at all since we usually design APIs TS first (as said before).
But for the sake consistency I'd say we should keep the check similar which would be the `!= null` check.

https://github.com/faker-js/faker/blob/edfcba13dcb68dbcdc5b50d0c177e2c1d20acb80/src/modules/location/zip-code.ts#L59-L61

src/modules/finance/module.ts
xDivisionByZerox: Ah, my fault. I assumed these codes where the same as we have in PreBuiltMetadataDefinition#code, which is not the case:

https://github.com/faker-js/faker/blob/edfcba13dcb68dbcdc5b50d0c177e2c1d20acb80/src/definitions/metadata.ts#L11-L14

In that case, ignore my comment. We can think about changing that in the future.

src/modules/finance/module.ts
ST-DDT: I would like to use this weekend to complete/rebase the remaining transformation PRs, so I implemented these changes myself.

ST-DDT (approved):

Shinigami92 (approved):

ST-DDT:

PR #3987, opened 2026-08-14
infra: add support for node v26

Shinigami92:
Add support for Node v26

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node_version: [22, 24]
+        node_version: [22, 24, 26]
       fail-fast: false
     timeout-minutes: 10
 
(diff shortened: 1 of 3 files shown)


xDivisionByZerox (changes requested):
You probably want to re-add the node v26 possibility to the migration sample:
https://github.com/faker-js/faker/blob/83a999a8f831a4def0a93308f2cf0a46e924678e/docs/guide/upgrading_v11/3916.md?plain=1#L3

Shinigami92:
> You probably want to re-add the node v26 possibility to the migration sample:
> 
> https://github.com/faker-js/faker/blob/83a999a8f831a4def0a93308f2cf0a46e924678e/docs/guide/upgrading_v11/3916.md?plain=1#L3

there was never a mention for 26 in migration guide, cause nothing important for "migration"

xDivisionByZerox:
> there was never a mention for 26 in migration guide, cause nothing important for "migration"

In that case the mention of v24 also does not really make sense, does it?
Eighter you argue that you mention all the newly supported versions (then v26 should be in there) or you argue that you list realistic migration versions, the next major version, from v20 (then v24 wouldn't make sense here).

ST-DDT (approved):
IMO mentioning v24.0.0 as the first full node that is supported is fine.

xDivisionByZerox (approved):

PR #3995, opened 2026-08-15
chore: gitignore claude

Shinigami92:
see title

extracted from https://github.com/faker-js/faker/pull/3773

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,4 @@ cypress/videos
 .playwright-mcp/
 
 _local/
+.claude/
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "10.6.0",
   "description": "Generate massive amounts of fake contextual data",
   "scripts": {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -113,3 +113,4 @@ cypress/videos
 .playwright-mcp/
 
 _local/
```
ST-DDT, line 115: Maybe we should rename this to .local to stay consistent with the other folders in root.

ST-DDT (approved):

.gitignore
Shinigami92: at work we even use `/local` and `/tmp` folder now for month already
local is for more long-living but scratch files
tmp is for that claude does not access system tmp at root 😅 (minor security enhancement, claude is deny-permissioned to access root/system tmp)

xDivisionByZerox (approved):
🤷‍♀️

.gitignore
Shinigami92: this is something that can be done in another PR

PR #3998, opened 2026-08-16
infra: rename _local to .local

ST-DDT:
Rename the `_local` folder to `.local` for consistency with the other folders in the project root.

According to the original PR: The folder may be used to store more long lived local scripts or files that may help during development.

- https://github.com/faker-js/faker/pull/3541

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,5 +112,5 @@ cypress/screenshots
 cypress/videos
 .playwright-mcp/
 
-_local/
+.local/
 .claude/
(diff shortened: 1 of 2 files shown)


Shinigami92 (approved):

Shinigami92:
For the record: I like this change, because `_`-prefixed folders were kinda influenced by `_`-prefixed variables. But they have two different meanings. _vars are e.g. named parameters to mark an unused var. But . folders are established linux hidden folders and therefor even a bit more clear to users that it is gitignored.

xDivisionByZerox (approved):

PR #4002, opened 2026-08-16
docs: improve pnpm version docs

### LINKED ISSUE #3997: Clearer instructions for upgrading to pnpm 11
I noticed after #3895 was merged, if you run `pnpm i` with an older version of pnpm (e.g. 9.12.1) you get an error:

` ERROR  packages field missing or empty`

this is because a `pnpm-workspace.yaml` file exists but pnpm 9 doesn't know to handle a workspace file with no defined packagesd. But it's not very intuituve for someone who hits this error to realise "i need to upgrade pnpm".

Could we add some check to detect this and give a more useful error?

matthewmayer:
fix #3997

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,8 @@ A lot of effort has been put into `Faker` to create a useful and handy library.
 There are still a lot of things to be done, so all contributions are welcome!
 If you want to make `Faker` a better place, please read the following contribution guide.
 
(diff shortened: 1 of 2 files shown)


ST-DDT (approved):
Should we mention corepack as well?

matthewmayer:
I think no. 

I feel if you know what corepack is, it just works. 

If you don't know what it is, you would have to install it first, which is just as much work as installing pnpm.

Shinigami92 (approved):

ST-DDT:
> If you don't know what it is, you would have to install it first, which is just as much work as installing pnpm.

With the sole difference, that you only have to do it once, with pnpm you have to to it for every (?) version.

PR #4006, opened 2026-08-19
fix(locale): remove stray whitespace from five locale entries

MahinAnowar:
Five locale entries carry a leading or trailing space, so faker hands the space back to the caller:

| locale | definition | entry | of |
|---|---|---|---|
| `ar` | `location.state` | `" بابل"` | 110 |
| `bn_BD` | `location.state` | `"রংপুর "` | 8 |
| `cy` | `book.format` | `"E-lyfr "` | 4 |
| `hu` | `location.city_name` | `"Siófok "` | 3151 |
| `fa` | `lorem.word` | `" و زمان"` | 89 |

Each is the only untrimmed entry in its array, and none of the five are affixes, so the space isn't doing anything. `faker.location.state()` on `ar` can return `" بابل"`, and `hu` city names feed `location.city()`, where the trailing space ends up in the middle of a composed string.

I found these by walking `allLocales`, so I added the check to `locale-data.spec.ts` rather than leaving it as five one-off edits. It follows the `ignoredCharacterData` pattern already in that file, because three definitions legitimately hold whitespace:

```ts
const ignoredWhitespaceData = new Set([
  '.location.building_number',
  '.location.city_suffix',
  '.location.street_suffix',
]);
```

Those are concatenated onto another value, so `" aan de IJssel"` in `nl` and `" s/n."` in `es` are data, not mistakes. I left them alone. `nl.location.city_suffix` also contains an empty string, which reads as "sometimes no suffix" and looks deliberate, so I didn't touch that either.

The five locales fail the new check on `next` and every other locale passes, which is what makes me reasonably confident the ignore list is right rather than just tuned until green. `locale-data.spec.ts` is green at 2763, and location, book, lorem plus `all-functional` come to 40045 with no snapshot changes. oxlint and oxfmt are clean.

I can't read Arabic, Bengali, Welsh, Hungarian or Persian, so I've only removed whitespace and changed nothing else about any of these words. Worth a native speaker's glance if that's cheap for you.


matthewmayer (dismissed):

Shinigami92 (dismissed):

test/locale-data.spec.ts
```diff
@@ -39,6 +39,32 @@ const ignoredCharacterData = new Set([
   '.system.mime_type',
 ]);
 
+/**
+ * A set of locale data paths whose entries are affixes that are concatenated onto another value,
+ * so the surrounding whitespace is part of the data rather than a mistake.
+ */
+const ignoredWhitespaceData = new Set([
+  '.location.building_number',
```
ST-DDT, line 47: Not sure about this part. IMO the `building_number` should not have a whitespace.
(only es + es_MX have whitespace there and not for all entries)
I don't know whether that is correct or not.

It is ok to fix in a separate PR/analyze independently.

test/locale-data.spec.ts
```diff
@@ -39,6 +39,32 @@ const ignoredCharacterData = new Set([
...
+  '.location.city_suffix',
+  '.location.street_suffix',
+]);
+
+function untrimmedEntries(data: unknown, path: string = ''): string[] {
+  if (ignoredWhitespaceData.has(path)) {
+    return [];
+  } else if (Array.isArray(data)) {
+    return data.flatMap((e) => untrimmedEntries(e, `${path}[]`));
```
ST-DDT: Idea: This could theoretically include the array index.

```suggestion
    return data.flatMap((e, i) => untrimmedEntries(e, `${path}[${i}]`));
```

<table>
<tr><th>Current</th><th>Proposed</th></td>
<tr><td><pre lang="diff">- []
+ [
+   ".location.building_number[]: \" s/n.\"",
+   ".location.building_number[]: \" #\"",
+   ".location.building_number[]: \" ##\"",
+   ".location.building_number[]: \" ###\"",
+   ".location.building_number[]: \" ####\"",
+ ]</pre>
</td><td><pre lang="diff">- []
+ [
+   ".location.building_number[0]: \" s/n.\"",
+   ".location.building_number[3]: \" #\"",
+   ".location.building_number[4]: \" ##\"",
+   ".location.building_number[5]: \" ###\"",
+   ".location.building_number[6]: \" ####\"",
+ ]</pre></td></tr>
</table>

ST-DDT (dismissed):

test/locale-data.spec.ts
MahinAnowar: Had a look at that one. In `es` the whitespace is load-bearing right now: `street_address` is `'{{location.street}}{{location.buildingNumber}}'` with nothing between the two, so the separator has to live in the data. And it can't just be hoisted into the template either, because the entries don't agree on what the separator is — `' s/n.'` and `' #'` want a space, `', #'` and `', ##'` want a comma. Same shape in `es_MX`.

So it's not a stray space like the five I removed, but you're right that it's odd — calling `faker.location.buildingNumber()` on its own hands you `" s/n."` with the leading space, which is junk from the caller's point of view. Fixing it properly means reworking the template so the separator is chosen there, and that changes generated output for both locales. Happy to do it as a follow-up if you want, but it felt like too much to smuggle into a whitespace cleanup.

test/locale-data.spec.ts
MahinAnowar: Good call, taken. Makes a real difference on the big arrays — reintroducing the `hu` space now reports `.location.city_name[2407]: "Siófok "` instead of just `.location.city_name[]`, and that array has 3151 entries, so without the index you're grepping for it.

MahinAnowar:
Took the array-index suggestion and answered the `building_number` question inline. Heads up that the push dismissed the three approvals, sorry about that — the only change since is the one-line `(e, i)` in the test helper. Also merged `next` in since it had moved.

ST-DDT (approved):

Shinigami92 (approved):

PR #4007, opened 2026-08-19
chore: sort imports

Shinigami92:
The automatic import sort that was missed by the oxc tooling migration


ST-DDT (dismissed):
Sorting works, but it doesn't remove extraneous imports.

test/all-functional.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import type { Faker, allLocales } from '../src';
+import type { allLocales, Faker } from '../src';
```
xDivisionByZerox: Before we switched to oxlint/format the import ordering was case sensitiv as well. Why exactly do we need to change this now? I would prefer the case sensitivity, as it reflects the natural ordering due to naming conventions as well: classes (usually PascalCase) first; utility functions (usually camelCase) later.

For old refs that used case sensitive ordering you can check the current v10 branch (2cb04231a6ace91a59ebe577c653f4ec66478ca3):
https://github.com/faker-js/faker/blob/2cb04231a6ace91a59ebe577c653f4ec66478ca3/test/all-functional.spec.ts#L2

oxlint.config.ts
```diff
@@ -32,25 +32,29 @@ export default defineConfig({
...
-    'no-useless-backreference': 'error',
     'no-shadow': 'off',
-    'no-new': 'off',
     'no-underscore-dangle': 'off',
+    'no-useless-backreference': 'error',
     'no-useless-constructor': 'off', // oxlint doesn't skip constructors that only declare parameter properties
+    'sort-imports': [
+      'error',
+      { ignoreCase: true, ignoreDeclarationSort: true },
+    ],
```
xDivisionByZerox: Another question: Why do we need to define this "rule" in the formatter as well as in the linter? Wouldn't there be the risk of them desyncing their configs and expecting different sorting states? What benefit would I have from the linter telling me that the imports are not sorted if the formatter tells me the same?

xDivisionByZerox (commented):
No strong pinions, but two questions to understand the overall goal of this PR better.

test/all-functional.spec.ts
Shinigami92: > Why exactly do we need to change this now?

Because I couldn't find a configuration for the `sort-imports` lint rule to be compatible with what we had before. Feel free to spend additional time into it, but I'm satisfied with the result that it is compatible to TypeScript's `Organize Imports`.

oxlint.config.ts
Shinigami92: > Why do we need to define this "rule" in the formatter as well as in the linter?

The **oxfmt** orders the import lines, the **oxlint** rule orders the imports inside the line.

I hope in the future the oxc tooling finds a much better way for this, until then I'm satisfied with to make it as much as possible to TypeScript's `Organize Imports`.

test/all-functional.spec.ts
xDivisionByZerox: By removing the "ignoreCase: true" I was able to mostly revert the state from most imports to the old state. But (as you can see by the diff) some imports were not sorted accordingly pre-migration as well. Although, less changes would be applied by this sorting than with case sensitive sorting active.
Feel free to revert 49885f1d if you do not like this ordering.

oxlint.config.ts
xDivisionByZerox: Thanks, that helps to understand this. Although, I have to agree that having two tools sorting imports in different ways is somewhat unfortunate.

test/all-functional.spec.ts
ST-DDT: Is it now sorted case sensitive or "not sorted"?

xDivisionByZerox: Yes, right now it is sorted case sensitive.

Shinigami92: Tested in `faker/test/all-functional.spec.ts`, TypeScript's `Organize Imports` in VSCode seems to work with AND without `ignoreCase: true`. Maybe TS relaxed it in TS 6 and I missed the update? But when it works I'm satisfied as my only requirement is that it does not interfer/conflict with TypeScript's `Organize Imports`.

PS: I cannot approve, you both need to approve it, cause I'm the author of the PR.

oxlint.config.ts
ST-DDT: https://github.com/typescript-language-server/typescript-language-server/blob/master/docs/configuration.md

According to that, you can configure the behavior in the language server settings.
But those have to be set via IDE settings.

````jsonc
// .vscode/settings.json
"js/ts.preferences.organizeImports": {
  caseSensitivity: "auto" | "caseInsensitive" | "caseSensitive",
  typeOrder: "auto" | "last" | "inline" | "first",
  unicodeCollation: "ordinal" | "unicode",
  locale: string,
  numericCollation: boolean,
  accentCollation: boolean,
  caseFirst: "default" | "upper" | "lower"
}
````

(Not verified)

Not sure whether we want to do this.

ST-DDT (approved):

xDivisionByZerox (approved):

oxlint.config.ts
Shinigami92: I'm against setting VScode specific things and the CI should be compatible to default behavior.
I would more like to report inconsistencies to oxc tooling. But (for other things like package.json sorting) I already did that and got rejected due to to many issues overload on their side. They told my I should try to come back 26Q4 earliest.

PR #4011, opened 2026-08-24
test(seed): make random seed test deterministic

hiSandog:
The seed unit test compared two real Math.random results and assumed they would differ. A collision is valid, so the assertion introduced an unnecessary nondeterministic failure path without checking how randomSeed transforms its input.

This mocks Math.random with a fixed value, asserts the exact seed calculation, and restores the spy after the assertion.

## Validation

- ./node_modules/.bin/vitest run test/internal/seed.spec.ts — 1 test passed, no type errors
- git diff --check — passed

Full local preflight was not run because the existing dependency tree predates the upstream Oxfmt/Oxlint migration; remote CI will run with the current lockfile.

### DIFF
--- a/test/internal/seed.spec.ts
+++ b/test/internal/seed.spec.ts
@@ -1,11 +1,12 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { randomSeed } from '../../src/internal/seed';
 
 describe('seed', () => {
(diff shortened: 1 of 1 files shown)


ST-DDT (changes requested):
The point of this test to verify that the method returns a random value each time.
It does not make any assumptions about the implementation. Not sure how useful the test is overall.

Did you run into issues with this test?

PR #4013, opened 2026-08-27
feat(helpers): add arraySamples for sampling with replacement

### LINKED ISSUE #3222: Add method/option to pick elements from an array multiple times

The current arrayElements method only takes elements from an array once.
The multiple method only uses a function as input.

I need a method to pick random elements from an array multiple times.

### Suggested solution

Either as an additional parameter for the existing arrayElements `repeatedElements: boolean = false` or a new method (not sure what to call it)

---

Here a pseudo implementations with some potential performance improvements (safes a few invocations to faker.number.int by stacking them):

````ts
function multipleArrayElements(values: T[], count: numberOrRange) {

  const valuesLength = values.length:
  const divisions = Math.floor(Math.log(Number.MAX_SAFE_INTEGER) / Math.log(valuesLength));
  const base = valuesLength ** divisions;
  const repeats = Math.floor(count / divisions);
  const remains = count % divisions

  const result = [];
  for(let i = 0; i<repeats; i++) {
    let random = faker.number.int(base);
    for (let j=0; j<divisions; j++) {
	  const index = random % valuesLength;
      random /= valuesLength;
      result.push(values[index]);
    }
  }
  let random = faker.number.int(base);
  for (let j=0; j<remains; j++) {
	const index = random % valuesLength;
    random /= valuesLength;
    result.push(values[index]);
  }
  return result;
}
````

The performance part needs actual benchmarks though.

### Alternative

Keep using the existing code.

### Additional context

This method can be used to improve the performance of the string module and similar modules.

And also reduce code duplication a bit, namely code like this:

https://github.com/faker-js/faker/blob/af1dbcdbf363a12c448834dff031a698c8925e2e/src/modules/string/index.ts#L148-L151

Due to "improvements" you would need to call faker.number.int() only once per 11 random elements of 26 characters.

### PR DESCRIPTION
**Pull request by shreyanshu12345**
Closes #3222

## Summary
Added `faker.helpers.arraySamples()` — samples n elements from an array with replacement, allowing the same element to be picked multiple times and count to exceed array length.

## Changes
- Added `arraySamples` in `src/modules/helpers/array-samples.ts`
- Written testcases for the new helper

### CODE REVIEW

**Review by ST-DDT — DISMISSED** [review]
CI appears to be failing. Please fix using `pnpm run preflight`.

**Comment by ST-DDT** [comment]
Blocked by #3857

- #3857

**Comment by matthewmayer** [comment]
i would prefer an options object to allow for future expansion

i think allowRepeatedElements would be more consistent with other params like allowLeadingZeros and allowSpecialCharacters in other methods.
```js
   faker.helpers.arrayElements([1, 2, 3], 5, {allowRepeatedElements: true}) // [1, 3, 1, 2, 1]
```

src/modules/helpers/module.ts
```diff
@@ -1020,22 +1020,31 @@ export class SimpleHelpersModule extends SimpleModuleBase {
    * @param count Number or range of elements to pick.
    *    When not provided, random number of elements will be picked.
    *    When value exceeds array boundaries, it will be limited to stay inside.
```
matthewmayer, line 433: should we note that this is only the case when allowRepeatedElements is false?

matthewmayer (changes requested):

src/modules/helpers/module.ts
shreyanshu12345: Yes, when allowRepeatedElements is true, my implementation of the function allows it to exceed the length of the array. Well, Since repetition is allowed we can pick more elements than present in the array right ??

shreyanshu12345: should i update the function description to this:
* @param count Number or range of elements to pick.
*    When not provided, random number of elements will be picked.
*    When value exceeds array boundaries and `allowRepeatedElements` is false,
*    it will be limited to stay inside.

src/modules/helpers/module.ts
```diff
@@ -1020,15 +1020,31 @@ export class SimpleHelpersModule extends SimpleModuleBase {
...
+    array: ReadonlyArray<T>,
+    count?: NumberOrRange,
+    options?: {
+      /**
+       * If true, elements can be picked more than once.
+       *
+       * @default false
+       */
+      allowRepeatedElements?: boolean;
+    }
```
Shinigami92: **question (blocking):** shouldn't this collapse to a `countOrOptions?: NumberOrRange | {count:NumberOrRange, allowRepeatedElements?: boolean}`?

please check how other methods are already implemented

ST-DDT:
Please resolve the merge conflicts and update to our new small modular functions.
The signatures in the helper module get updated using `pnpm run generate:module-tree`.

matthewmayer (commented):
I think there is a strong argument that an option is the wrong approach to handle this.

Sampling without replacement, and sampling with replacement, are very different things, and other libraries and languages often treat them seperately

e.g. in Python: Distinguishes between `random.sample(population, k)` (without replacement) and `random.choices(population, k=k)`

The checks and possible errors for the two modes are quite different.

I would propose `faker.helpers.arraySample(array, countOrOptions)` as a better name that avoids overloading arrayElements

```
faker.helpers.arraySample(['heads', 'tails'], 10);
// e.g. ['heads', 'tails', 'heads', 'heads', 'tails', ...]
```

ST-DDT:
Maybe `arraySamples` to hint at returning multiple?

shreyanshu12345:
I've refactored the approach based on the feedback. Instead of adding an allowRepeatedElements option to arrayElements, I've added a separate arraySamples function that samples with replacement, similar to Python's random.choices().

I've made count required for now. Should it have a default of 1 like random.choices(population, k=1), or keep it required?

ST-DDT:
Please move the method implementation to a separate file, like the others do.

matthewmayer:
Please rename the PR to reflect the new approach

Suggest no default

src/modules/helpers/array-samples.ts
```diff
@@ -0,0 +1,43 @@
...
+  const result: T[] = [];
+
+  if (array.length === 0 || count === 0) {
+    return result;
+  }
+
+  const numElements = rangeToNumber(fakerCore, count);
+
+  for (let i = 0; i < numElements; i++) {
+    result.push(array[int(fakerCore, { min: 0, max: array.length - 1 })]);
```
ST-DDT: You can use arrayElement instead.

src/modules/helpers/array-samples.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export function arraySamples<const T>(
+  fakerCore: FakerCore,
+  array: ReadonlyArray<T>,
+  count: NumberOrRange
+): T[] {
+  const result: T[] = [];
+
+  if (array.length === 0 || count === 0) {
+    return result;
```
ST-DDT, line 44: TBD: Return empty or throw on empty array.
ArrayElement does throw.

test/modules/helpers.spec.ts
```diff
@@ -472,6 +483,227 @@ describe('helpers', () => {
...
+          const samples = faker.helpers.arraySamples(testArray, 5);
+
+          expect(samples).toHaveLength(5);
+          for (const item of samples) {
+            expect(testArray).toContain(item);
+          }
+        });
+
+        it('should not mutate the input array', () => {
+          const testArray = ['hello', 'to', 'you', 'my', 'friend'];
```
ST-DDT: Freezing the input array, can also uncover more hidden changes.

matthewmayer:
The PR summary also needs to be updated.

src/modules/helpers/array-samples.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Elements may appear more than once.
+ *
+ * @template T The type of the elements to pick from.
+ *
+ * @param fakerCore The FakerCore to use.
+ * @param array Array to pick the value from.
+ * @param count Number or range of elements to pick.
+ *
+ * @example
+ * arraySamples(fakerCore, ['cat', 'dog', 'mouse'], 1) // ['mouse']
```
matthewmayer: Include an example where n > array.count

For example five coin tosses.

src/modules/helpers/array-samples.ts
matthewmayer: Could we just wrap arrayElement?

src/modules/helpers/array-samples.ts
shreyanshu12345: the arrayElements method returns an empty array when numElements is 0 or when an empty array is passed. My new commit does the same.

src/modules/helpers/array-samples.ts
```diff
@@ -17,6 +17,7 @@ import { rangeToNumber } from './range-to-number';
  * arraySamples(fakerCore, ['cat', 'dog', 'mouse'], 1) // ['mouse']
```
matthewmayer: i feel the first example in the docs should always be very indicative of common usage. So i would make n>2 and have at least one repeat. Looking at this example i can't tell whether or not its with replacement.

test/modules/helpers.spec.ts
```diff
@@ -472,6 +483,233 @@ describe('helpers', () => {
...
+        });
+
+        it('should return an empty array when receiving an empty array and count = 0', () => {
+          const result = faker.helpers.arraySamples([], 0);
+
+          expect(result).toHaveLength(0);
+          expect(result).toEqual([]);
+        });
+
+        it('should return an empty array when receiving an empty array and count > 0', () => {
```
matthewmayer: i feel this should throw.

src/modules/helpers/array-samples.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param array Array to pick the value from.
+ *      when an empty array is provided, an empty array is returned.
+ * @param count Number or range of elements to pick.
+ *      when count is 0 or less, an empty array is returned.
+ *
+ * @example
+ * arraySamples(fakerCore, ['cat', 'dog', 'mouse'], 1) // ['mouse']
+ * arraySamples(fakerCore, [1, 2, 3, 4, 5], 2) // [4, 2]
+ * arraySamples(fakerCore, [1, 2, 3, 4, 5], { min: 2, max: 4 }) // [3, 5, 5]
+ * arraySamples(fakerCore, ["Heads", "Tails"], 4) // ["Heads", "Tails", "Tails", "Heads"]
```
ST-DDT: IMO this should have 2 examples (one for each variant). Maybe a third if it adds something important to it.

1. The head and tails one with n>2 (4 is good)
2. One with a variable count, `[1, 2, 3], { min: 2, max: 5 }` -> `[2, 1, 3, 3, 1]`

test/modules/helpers.spec.ts
```diff
@@ -472,6 +483,233 @@ describe('helpers', () => {
...
+            'hello',
+            'to',
+            'you',
+            'my',
+            'friend',
+          ]);
+          const result = faker.helpers.arraySamples(testArray, 3);
+
+          expect(result).toHaveLength(3);
+          expect(result).not.toBe(testArray);
```
ST-DDT: IMO this can never be true, because the input has length = 5.

test/modules/helpers.spec.ts
```diff
@@ -472,6 +483,233 @@ describe('helpers', () => {
...
+          result.forEach((el) => expect(testArray).toContain(el));
```
ST-DDT: Please keep the validation style consistent => for-loop

test/modules/helpers.spec.ts
```diff
@@ -472,6 +483,233 @@ describe('helpers', () => {
...
+        });
+
+        it('should not throw with a frozen array', () => {
```
ST-DDT: This test is kind of redundant with the one above.

src/modules/helpers/array-samples.ts
```diff
@@ -0,0 +1,50 @@
+import type { FakerCore } from '../../core';
+import { FakerError } from '../../errors/faker-error';
+import type { NumberOrRange } from '../../utils/types';
+import { arrayElement } from './array-element';
+import { rangeToNumber } from './range-to-number';
+
+/**
+ * Returns an array of elements sampled from the given array with replacement.
+ * Elements may appear more than once.
```
ST-DDT: Maybe:

```suggestion
 * Elements may therefore appear more than once.
```

Otherwise the second sentence sounds redundant. But I kind of feel it is important, if you don't know what "sample with replacement" is.

ST-DDT (dismissed):

matthewmayer (commented):
Can you add @see cross-links between arraySamples and arrayElements?

ST-DDT (approved):

src/modules/helpers/module.ts
ST-DDT: IMO this is now redundant.

Shinigami92 (approved):

ST-DDT:
@matthewmayer Could you please re-review this PR (or dismiss your stale review)?

matthewmayer (approved):

PR #4022, opened 2026-08-31
infra: sort type imports first

ST-DDT:
Follow up on #4007 

- #4007

---

Ensure, that type imports are ordered before their non-type files.

**Good**

````ts
import type { AlphaNumericChar } from './_types';
import { DIGIT_CHARS, LOWER_CHARS, UPPER_CHARS } from './_types';
````

**Bad**

````ts
import { DIGIT_CHARS, LOWER_CHARS, UPPER_CHARS } from './_types';
import type { AlphaNumericChar } from './_types';
````

### DIFF
--- a/oxfmt.config.ts
+++ b/oxfmt.config.ts
@@ -16,7 +16,9 @@ export const formatOptions: FormatConfig = {
     groups: [
       ['builtin', 'external', 'internal'],
       { newlinesBetween: false },
-      ['parent', 'sibling', 'index'],
+      ['type-parent', 'value-parent'],
+      ['type-sibling', 'value-sibling'],
+      ['type-index', 'value-index'],
     ],
     newlinesBetween: false,
     order: 'asc',
--- a/oxfmt.config.ts
+++ b/oxfmt.config.ts
@@ -14,7 +14,9 @@ export const formatOptions: FormatConfig = {
   printWidth: 80,
   sortImports: {
     groups: [
-      ['builtin', 'external', 'internal'],
+      ['type-builtin', 'value-builtin'],
+      ['type-external', 'value-external'],
+      ['type-internal', 'value-internal'],
       ['type-parent', 'value-parent'],
       ['type-sibling', 'value-sibling'],
       ['type-index', 'value-index'],
--- a/oxfmt.config.ts
+++ b/oxfmt.config.ts
@@ -15,7 +15,6 @@ export const formatOptions: FormatConfig = {
   sortImports: {
     groups: [
       ['builtin', 'external', 'internal'],
-      { newlinesBetween: false },
       ['type-parent', 'value-parent'],
       ['type-sibling', 'value-sibling'],
       ['type-index', 'value-index'],

### CODE REVIEW

File: oxfmt.config.ts [hunk 1]
```diff
@@ -15,8 +15,9 @@ export const formatOptions: FormatConfig = {
   sortImports: {
     groups: [
       ['builtin', 'external', 'internal'],
```
ST-DDT: These could theoretically be changed to:

````ts
      ['type-builtin', 'value-builtin'],
      ['type-external', 'value-external'],
      ['type-internal', 'value-internal'],
````

but that doesn't have any impact on this project, so not sure if it is usefull.

xDivisionByZerox: I would probably do this for the sake of consistency inside the config :thinking:

ST-DDT: Done

xDivisionByZerox (approved):

Shinigami92:
Tested the new config in company repo and it is still not compatible to "Organize Imports" 🙁  
it has different sorting for imports `from 'node:…'`, and also with a `import type {…} from './'`

But I'm on wrong Mac right now and cannot debug it for Faker. Not sure if I will find time tomorrow or if I finally take time for myself after a looooong week.

ST-DDT:
> Tested the new config in company repo and it is still not compatible to "Organize Imports" 🙁

Is it closer to or farther away from the goal?

If it is, I revert the last commit.
If not we could merge this as an iterative improvement.

EDIT: Without a test set it is hard for me to deliver the optimal result.

Shinigami92:
> > Tested the new config in company repo and it is still not compatible to "Organize Imports" 🙁
> 
> Is it closer to or farther away from the goal?
> 
> If it is, I revert the last commit. If not we could merge this as an iterative improvement.

I need to check that tomorrow, I'm near out the door on my way to Vue Meetup Hamburg

ST-DDT:
> on my way to Vue Meetup Hamburg

No worries. Have fun.

Shinigami92:
I checked some samples and this was the only diff I could produce:


so somehow in Faker it works, but the config does not work in some of my other repos (e.g. at company 🤷)

so iterative process...

Shinigami92 (approved):

ST-DDT:
My config doesn't seem to work (anymore) for faker as well (or at lest not reliably).

`['type-builtin', 'value-builtin'],` may be equivalent to `builtin`.
Not sure why it fixed my test-case during the tests.

I may have confused type-sibling with value-parent.
But the screenshot I took of it, is clearly sibling only. 🤷‍♂️
I hope it wasn't my IDE trolling me by caching the eslint config+plugins until restart or something.
So in a sense it is an improvement, that siblings (and their types) are grouped together independently from parents.
But the type sorting is still garbage...

---

As a workaround, you can use this regexp to search for incorrectly sorted type imports via IDE:

`import \{[^}]*\} from '([^']+)';\r?\nimport type \{[^}]*\} from '\1';`

You may even be able to use this for replacements:

`(import \{[^}]*\} from '([^']+)';)\n(import type \{[^}]*\} from '\2';)` -> `$3\n$1`

PR #4023, opened 2026-08-31
docs(string): document numeric errors

### LINKED ISSUE #4018: string.numeric does not mention the possibility to throw in it's docs
# Description

Our coding guidelines for JSDocs state that a faker function that is able throw should that that in their JSDocs.

https://github.com/faker-js/faker/blob/bbd169385a438ece2ed7e588945bc68838ed144c/CONTRIBUTING.md?plain=1#L362

While reviewing #3857 I noticed that `faker.string.numeric` does not have any `throw` comments, while clearly having possible errors being thrown during runtime:

https://github.com/faker-js/faker/blob/bbd169385a438ece2ed7e588945bc68838ed144c/src/modules/string/module.ts#L559-L561

## Task

Add JSDocs stating that the function can throw as described [in our guidelines.](https://github.com/faker-js/faker/blob/next/CONTRIBUTING.md) 

In this specific case I would expect two separate throw statements as a comment:

- No digits at all are allowed by `allowedDigits`
- Leading zeros are disallowed by the only allowed digit is `0`

deshrajvermay9517-png:
Closes #4018

Adds the two missing `@throws {FakerError}` entries to `faker.string.numeric()` documentation.

Checks:
- pnpm run format
- pnpm run lint
- pnpm run build
- pnpm run ts-check
- pnpm run preflight

### DIFF
--- a/src/modules/string/module.ts
+++ b/src/modules/string/module.ts
@@ -492,8 +492,7 @@ export class StringModule extends SimpleModuleBase {
    * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
    * @param options.exclude An array of digits which should be excluded in the generated string. Defaults to `[]`.
    *
-   * @throws {FakerError} If no digits are available after applying `exclude`.
-   * @throws {FakerError} If `allowLeadingZeros` is `false` and `0` is the only available digit.
+   * @throws {FakerError} If no digits are available after applying `exclude` and `allowLeadingZeros`.
    *
    * @see faker.number.int(): For generating a number (within a range).
    *
--- a/src/modules/string/module.ts
+++ b/src/modules/string/module.ts
@@ -492,6 +492,9 @@ export class StringModule extends SimpleModuleBase {
    * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
    * @param options.exclude An array of digits which should be excluded in the generated string. Defaults to `[]`.
    *
+   * @throws {FakerError} If no digits are available after applying `exclude`.
+   * @throws {FakerError} If `allowLeadingZeros` is `false` and `0` is the only available digit.
+   *
    * @see faker.number.int(): For generating a number (within a range).
    *
    * @example


src/modules/string/module.ts
```diff
@@ -492,6 +492,9 @@ export class StringModule extends SimpleModuleBase {
    * @param options.allowLeadingZeros Whether leading zeros are allowed or not. Defaults to `true`.
    * @param options.exclude An array of digits which should be excluded in the generated string. Defaults to `[]`.
    *
+   * @throws {FakerError} If no digits are available after applying `exclude`.
+   * @throws {FakerError} If `allowLeadingZeros` is `false` and `0` is the only available digit.
```
ST-DDT: I think the second case is just a different variation of the first case, so we dont need to mention it separately, do we?
@xDivisionByZerox

deshrajvermay9517-png: I followed the issue description, which requested two separate @throws statements. I can update it if you prefer a single statement.

xDivisionByZerox: In my opinion these are two dedicated errors that can pop up. I do not see the second case being a different variation asthe first (in the sense that it's the same error).
If you provide leading zeros with false and exclude all digits other than zero, there is technically still available digits when following the logic of the first throw statement. The available digits are simply not applicable. Because of that I suggested two dedicated errors messages.

Do you agree with me after that explanation? Otherwise we might need a third party of @faker-js/members to break the tie.

ST-DDT: Maybe like this?

````jsdoc
@throws {FakerError} If no digits are available after applying `exclude` and `allowLeadingZeros`.
````

EDIT: Backticks are somehow invisible/removed from the code block...

xDivisionByZerox: That would work for me.

deshrajvermay9517-png: Updated as suggested. Thanks.

ST-DDT (approved):

xDivisionByZerox (approved):

PR #4036, opened 2026-09-04
feat(locale): add airline data for az

jamalkamaladdin:
Adds the `airline` module to `az`, which previously had 9 of 23 modules.

- `airport` — 46 entries: all 11 Azerbaijani airports that have an assigned IATA code, plus 35 international airports weighted towards destinations with direct flights from Baku
- `airline` — 32 entries: 3 Azerbaijani carriers and 29 international ones

`airplane` is intentionally omitted. Aircraft model names ("Airbus A320", "Boeing 737 MAX 8") are written identically in Azerbaijani, which uses the Latin script, so copying the `en` list would add no value — the `en` fallback already returns the correct value. `fa` and `zh_CN` translate that file because their scripts differ.

### Sources

Cross-checked against several independent sources rather than copied from one:

- caa.gov.az — State Civil Aviation Agency
- azal.az, airport.az — carrier and airport sites
- az.wikipedia.org and en.wikipedia.org — airport and airline pages
- IATA member list

### Notes on the data

- Ganja airport is `GNJ`; the older `KVD` code, from the city's former name, is no longer in use.
- Buta Airways is excluded — it merged into AZAL in October 2023 and has no independent IATA code.
- Airports with no assigned IATA code (Balakən, Xankəndi, Naftalan, Şəki, Ağdam and others) are excluded.
- International airline names are kept in their original Latin form rather than transliterated. This is how Azerbaijani media writes them, and Azerbaijani already uses the Latin alphabet.

### Character snapshot

This adds `I`, `W` and `w` to the `az` character set. `W` is not a letter of the Azerbaijani alphabet, but it appears in brand names kept in their original form — including Silk Way West Airlines, an Azerbaijani carrier.

### Checks

- Ran `pnpm run preflight`; the full test suite passes — 54201 tests, 0 failures, no type errors
- No duplicate names or IATA codes


src/locales/az/airline/airline.ts
```diff
@@ -0,0 +1,37 @@
...
+  { name: 'Air China', iataCode: 'CA' },
+  { name: 'China Southern', iataCode: 'CZ' },
+  { name: 'Emirates', iataCode: 'EK' },
+  { name: 'Etihad Airways', iataCode: 'EY' },
+  { name: 'flydubai', iataCode: 'FZ' },
+  { name: 'Air Arabia', iataCode: 'G9' },
+  { name: 'Uzbekistan Airways', iataCode: 'HY' },
+  { name: 'Iran Air', iataCode: 'IR' },
+  {
+    name: 'Azərbaycan Hava Yolları (AZAL)',
```
ST-DDT: Is AZAL really part of the name or is it just a common abbreviation?

jamalkamaladdin: The official Azerbaijani name is Azərbaycan Hava Yolları (Azerbaijan Airlines in English); AZAL is the short form, also used as the airline's international call sign.

ST-DDT: Do you expect to see the abbreviation to be there in a list of names.
If yes, that is fine. If not, please remove.

jamalkamaladdin: Pushed edit

ST-DDT (approved):

Shinigami92 (approved):

jamalkamaladdin:
Copilot's both comments incorrect )

PR #4039, opened 2026-09-04
infra: revert oxfmt config changes

ST-DDT:
Cleanup oxfmt config, because the previous changes did not work as expected.

Elements inside nested arrays are treated as the same level, thus splitting type imports from value imports doesn't have an impact.

### DIFF
--- a/oxfmt.config.ts
+++ b/oxfmt.config.ts
@@ -14,12 +14,9 @@ export const formatOptions: FormatConfig = {
   printWidth: 80,
   sortImports: {
     groups: [
-      ['type-builtin', 'value-builtin'],
-      ['type-external', 'value-external'],
-      ['type-internal', 'value-internal'],
-      ['type-parent', 'value-parent'],
-      ['type-sibling', 'value-sibling'],
-      ['type-index', 'value-index'],
+      ['builtin', 'external', 'internal'],
+      { newlinesBetween: false },
+      ['parent', 'sibling', 'index'],
     ],
     newlinesBetween: false,
     order: 'asc',


oxfmt.config.ts
```diff
@@ -13,14 +13,7 @@ export const formatOptions: FormatConfig = {
...
   sortImports: {
-    groups: [
-      ['type-builtin', 'value-builtin'],
-      ['type-external', 'value-external'],
-      ['type-internal', 'value-internal'],
-      ['type-parent', 'value-parent'],
-      ['type-sibling', 'value-sibling'],
-      ['type-index', 'value-index'],
-    ],
+    groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
```
Shinigami92: **question (if-minor):** we are now nearly but no equally back to my initial suggested config in https://github.com/faker-js/faker/pull/4007/changes#diff-c50d6e1331becdc062a0dd24e260d98fb466027ca833047ba96ac501615aa5b2R17-R19

what does it make currently different and is it the final state you think is working? if not, would you maybe open an experimental repo or something to experiment there? otherwise I fear we are moving back and forth with multiple MRs in faker repo 🫤
and I feel like I get told by the other team members (not only you @ST-DDT) that the faker codebase is not a good place for experimenting. we have a similar thing running over in https://github.com/faker-js/faker/pull/4001#issuecomment-5533460161

ST-DDT: This is not the final working version. It cleans up the previous change.

I checked with oxlint/oxfmt and the type sort isn't supported at all right now.
The workaround mentioned is using the prettier-organize-imports plugin.
But I don't plan to do/PR that.

Shinigami92: what if we do just a `revert: 136cef0`?

ST-DDT: Sure, we could 

- revert that commit
- or  revert  and additionally remove the seemingly redundant `newlinesBetween` block
- or also delete the entire `sortImports.group` block
 
that would still result in the same import order.

Which variant would you prefer?

For now, I just removed the entire block.

Shinigami92: Fully removing the `groups` is worse, because this could result in more merge conflicts.
The config I provided worked fine and is at least compatible to `Organize Imports` (afaik).
You then tried to enhance it with the `type-*` prefixes, but found out that your IDE / caching tricked you and we found out it is not working as expected.
So I suggest reverting it back to the config I provided and is working in several of my repos already.

All in all I fully agree that this whole oxfmt should provide a 100% compatibility to TS native language server, but @camc314 and @Boshen are suffer by low capacity over in oxc tooling repo. Maybe @overlookmotel can help us?

camc314: cc @leaysgur

ST-DDT:
@Shinigami92 IMO that those are a bunch of non arguments.

I wasted enough of my time with this formatting/lint changes. So here you go.

Sorry for bothering.

Shinigami92 (approved):

Shinigami92:
> @Shinigami92 IMO that those are a bunch of non arguments.

I disagree

> I wasted enough of my time with this formatting/lint changes. So here you go.

Thanks for accepting the revert suggestion, however your comment sound emotional to me and I would like to tell that nobody asked you to set time into that. #4022 came up on free will from your side.  
Also oxfmt vs prettier is nowadays a non-negotiable think for me, cause oxfmt is sooooo damn much faster, and therefore giving also you time back instead of waiting for local tinkering and CI/CD waiting.

But as you see, we might now get help in the long run from @leaysgur maybe, which I don't know yet 👀 
@leaysgur feel free to ping me in the oxc Discord if you want fast communication 🙂 I send you also a friend inv there.

PR #4043, opened 2026-09-04
refactor: allow Faker to be created without locale data

ST-DDT:
Since `faker.helpers.fake` may become one of the main reasons for using the full `Faker` instance, it is reasonable to create `Faker` instances without locale data. While technically, creating it with an empty locale array is weird, throwing only for that cause is inconsistent with the other cases.

| Case | Old | New |
| --- | --- | --- |
| absent | ✔️ | ✔️ |
| empty-data | ✔️ | ✔️ |
| locale-data | ✔️ | ✔️ |
| empty-array | ❌ | ✔️ |
| locale-array | ✔️ | ✔️ |
| absent-via-core | ✔️ | ✔️ |
| empty-data-via-core | ✔️ | ✔️ |
| locale-data-via-core | ✔️ | ✔️ |
| empty-array-via-core | ✔️ | ✔️ |
| locale-array-via-core | ✔️ | ✔️ |

---

You see one TODO comment, then you end up with 7 new PRs...


matthewmayer (approved):

xDivisionByZerox (approved):

Shinigami92:
One maintainer approval is enough here, please merge 🙃

ST-DDT:
> One maintainer approval is enough here

Am I missing something here?

> please merge 🙃

Sure, will do. You could have merged it yourself too though 🤷

PR #4035, opened 2026-09-04
feat(locale): add science data for az

jamalkamaladdin:
Adds the `science` module to `az`, which previously had 8 of 23 modules.

- `chemical_element` — all 118 elements
- `unit` — 29 SI units

`symbol` and `atomicNumber` are taken unchanged from the `en` locale; only `name` is localised. Both sequences were verified to match `en` exactly.

### Sources

Names were cross-checked against several independent Azerbaijani sources rather than copied from a single one:

- [az.wikipedia.org](https://az.wikipedia.org/wiki/Kimyəvi_element) — element table and individual element pages
- [ens.az](https://ens.az/az/kimyevi-element) — Azerbaijani National Encyclopedia
- [e-qanun.az](https://frameworks.e-qanun.az/21/f_21432.html) — official state table of SI units
- nkpi.az / dim.gov.az — chemistry course material

Where Azerbaijani has its own established name, it is used instead of the Latin form, consistent with how such elements are named in Azerbaijani: Kükürd (S), Dəmir (Fe), Mis (Cu), Gümüş (Ag), Qalay (Sn), Sürmə (Sb), Qızıl (Au), Civə (Hg), Qurğuşun (Pb).

Units follow the official state table: `coul` (J), `hers` (Hz), `vatt` (W), `om` (Ω), `zivert` (Sv), `dərəcə Selsi` (°C). Unit names derived from scientists' names are written in lower case, as Azerbaijani requires.

### Checks

- `pnpm run preflight` passes (exit 0)
- 54199 tests passed, 0 failed; no type errors
- Character snapshot for `az` unchanged — no new characters introduced
- No duplicate entries

PR #4051, opened 2026-09-06
infra(lint): enable jsdoc-js/require-throws

### LINKED ISSUE #4021: Use oxlint to detect other places where functions throw but don't have a throws JSDoc
To catch more issues like

https://github.com/faker-js/faker/issues/4015
https://github.com/faker-js/faker/issues/4018

There's a oxlint rule

`jsdoc-js/require-throws`

If you set this to `warn` it finds 13 more places which can be investigated.

## All 13 warnings

| File | Line | Function |
|---|---|---|
| `src/internal/locale-proxy.ts` | 104 | `assertLocaleData` |
| `src/modules/commerce/module.ts` | 289 | `isbn` |
| `src/modules/helpers/eval.ts` | 67 | `fakeEval` |
| `src/modules/helpers/eval.ts` | 114 | `evalProcessFunction` |
| `src/modules/helpers/eval.ts` | 147 | `findParams` |
| `src/modules/helpers/eval.ts` | 183 | `evalProcessExpression` |
| `src/modules/helpers/module.ts` | 26 | `getRepetitionsBasedOnQuantifierParameters` |
| `src/modules/location/module.ts` | 232 | `zipCode` |
| `src/modules/phone/module.ts` | 30 | `number` |
| `src/modules/string/module.ts` | 123 | `fromCharacters` |
| `src/modules/string/module.ts` | 507 | `numeric` |
| `src/modules/system/module.ts` | 165 | `fileExt` |
| `src/modules/word/module.ts` | 340 | `sample` |

**Public API methods (8):** `isbn`, `fakeEval`, `zipCode`, `number`, `fromCharacters`, `numeric`, `fileExt`, `sample`

**Internal/private helpers (5):** `assertLocaleData`, `evalProcessFunction`, `findParams`, `evalProcessExpression`, `getRepetitionsBasedOnQuantifierParameters`

MahinAnowar:
Closes #4021.

Turning on `jsdoc-js/require-throws` flagged 14 functions that can throw without documenting it. I went through each one and wrote the `@throws` from the actual throw site rather than guessing at it, so the conditions match what the code really does:

| | |
|---|---|
| ~~`Faker` constructor~~ | ~~empty `locale` option~~ |
| `assertLocaleData` | locale data missing or not applicable |
| `commerce.isbn` | no registrant length for the generated group |
| `fakeEval` | expression empty or unresolvable |
| `evalProcessFunction` | unexpected character after a call |
| `findParams` | missing closing parenthesis |
| `evalProcessExpression` | empty part, or a dot with no property name |
| `getRepetitionsBasedOnQuantifierParameters` | unknown quantifier symbol |
| `location.zipCode` | no zip definition for the given state |
| `phone.number` | no definitions for the style in this locale |
| `fromCharacters` | no characters to select from |
| `system.fileExt` | unsupported mime type |
| `word.sample` | no matching word data for the locale |
| `seeded-runs` `it` helper | method not found on the module |

The rule is set to `error` rather than `warn`, per @ST-DDT's question on the issue, so new ones can't creep back in.

Two of these throw a plain `Error` rather than a `FakerError` — `phone.number()` and the seeded-runs helper. I documented them as `{Error}` instead of quietly converting them, since changing the error type is a behaviour change and belongs in its own PR if you want it.

One note on scope: CONTRIBUTING asks to keep a PR to a single module where possible, and this one can't be — the rule only goes to `error` once every site is documented, so it's all of them or none. Happy to split it up if you'd rather review it module by module.

`pnpm lint` is clean with the rule at `error`, `oxfmt --check` passes, and the touched modules' specs run 3067 passed. (`pnpm ts-check` fails on `test/simple-faker.spec.ts` with `Cannot find module '..'` both with and without this change — it needs a build first, so it's not from this.)

### DIFF
--- a/src/faker.ts
+++ b/src/faker.ts
@@ -113,6 +113,8 @@ export class Faker extends SimpleFaker {
    * Refer to the `seed()` method for more information.
    * Defaults to a random seed.
    *
+   * @throws {FakerError} If the `locale` option does not contain at least one locale definition.
+   *
    * @example
    * import { Faker, es } from '@faker-js/faker';
    * // const { Faker, es } = require('@faker-js/faker');
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -292,6 +292,7 @@ export default defineConfig({
     'jsdoc-js/require-returns': 'off',
     'jsdoc-js/require-returns-check': 'error',
     'jsdoc-js/require-returns-description': 'error',
+    'jsdoc-js/require-throws': 'error',
     'jsdoc-js/require-throws-type': 'error',
     'jsdoc-js/require-yields': 'error',
     'jsdoc-js/require-yields-check': 'error',
--- a/src/internal/locale-proxy.ts
+++ b/src/internal/locale-proxy.ts
@@ -100,6 +100,8 @@ export function createLocaleProxy(
  *
  * @param value The value to check.
  * @param path The path to the locale data.
+ *
+ * @throws {FakerError} If the locale data are missing or not applicable to this locale.
  */
 export function assertLocaleData<T>(
   value: T,
--- a/src/modules/commerce/module.ts
+++ b/src/modules/commerce/module.ts
@@ -276,6 +276,8 @@ export class CommerceModule extends ModuleBase {
    * or `13` (13-digit format). Defaults to `13`.
    * @param options.separator The separator to use in the format. Defaults to `'-'`.
    *
+   * @throws {FakerError} If no registrant length is defined for the generated group.
+   *
    * @example
    * faker.commerce.isbn() // '978-0-692-82459-7'
    * faker.commerce.isbn(10) // '1-155-36404-X'
--- a/src/modules/helpers/_eval.ts
+++ b/src/modules/helpers/_eval.ts
@@ -55,6 +55,8 @@ const REGEX_DOT_OR_BRACKET = /\.|\(/;
  * @param faker The faker instance to resolve array elements.
  * @param entrypoints The entrypoints to use when evaluating the expression.
  *
+ * @throws {FakerError} If the expression is empty or cannot be resolved.
+ *
  * @see faker.helpers.fake() If you wish to have a string with multiple expressions.
  *
  * @example
@@ -110,6 +112,8 @@ export function fakeEval(
  *
  * @param input The input string to parse.
  * @param entrypoints The entrypoints to attempt the call on.
+ *
+ * @throws {FakerError} If the function call is not followed by a dot, an opening parenthesis, or the end of the expression.
  */
 function evalProcessFunction(
   input: string,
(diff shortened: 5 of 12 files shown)


ST-DDT:
Blocked by/Conflict with #4043

MahinAnowar:
Good catch, and you're right — but only on one of the 14.

The overlap is just `src/faker.ts`. #4043 deletes the constructor's only throw, so once it lands the `@throws` I added there is both wrong and unnecessary: the rule won't ask for one any more. The other 13 sites don't touch anything you're changing.

Simplest is for #4043 to go first and me to rebase on top, dropping that one line. I can't sensibly drop it ahead of time, since `require-throws` at `error` would then fail on the constructor while the throw is still there on `next`.

Happy to do it the other way round if you'd rather — say the word and I'll pull `src/faker.ts` out entirely and leave the rule at `warn` in this PR, then flip it to `error` in a small follow-up after #4043. Whichever makes your review easier.

Separately, the red Codecov check doesn't look like it's about coverage — the upload sat for three minutes and exited 1, and a sibling fork PR (#4035) uploaded fine. I'll leave it; the rebase will re-run it anyway.

Shinigami92:
> Good catch, and you're right — but only on one of the 14.
> 
> The overlap is just `src/faker.ts`. #4043 deletes the constructor's only throw, so once it lands the `@throws` I added there is both wrong and unnecessary: the rule won't ask for one any more. The other 13 sites don't touch anything you're changing.
> 
> Simplest is for #4043 to go first and me to rebase on top, dropping that one line. I can't sensibly drop it ahead of time, since `require-throws` at `error` would then fail on the constructor while the throw is still there on `next`.
> 
> Happy to do it the other way round if you'd rather — say the word and I'll pull `src/faker.ts` out entirely and leave the rule at `warn` in this PR, then flip it to `error` in a small follow-up after #4043. Whichever makes your review easier.
> 
> Separately, the red Codecov check doesn't look like it's about coverage — the upload sat for three minutes and exited 1, and a sibling fork PR (#4035) uploaded fine. I'll leave it; the rebase will re-run it anyway.

What an AI slop comment 🫠 I read until the `—` and already knew that this is a so hard AI generated answer. Also we know that already, that is the reason why @ST-DDT made his comment.

matthewmayer:
> Blocked by/Conflict with #4043

Good catch, and you're right — probably because you're the damn maintainer.

ST-DDT:
> Good catch, and you're right — probably because you're the damn maintainer.

Right yes, maintainer no, just a frequent contributor/the author of that PR. 🤣

ST-DDT:
Still blocked, by #4070

- #4070

src/faker.ts
```diff
@@ -112,6 +112,8 @@ export class Faker extends SimpleFaker {
    * Refer to the `seed()` method for more information.
    * Defaults to a random seed.
    *
+   * @throws {FakerError} If the `locale` option does not contain at least one locale definition.
```
ST-DDT, line 115: This is now obsolete.

ST-DDT (changes requested):
Could you please update the PR, run `pnpm run generate:module-tree` and `pnpm run preflight`?

ST-DDT:
FYI: I fixed the remaining issues to proceed with the PR.

ST-DDT (dismissed):

ST-DDT (approved):

Shinigami92 (approved):

PR #4063, opened 2026-09-06
feat(location): transform to standalone modular functions

ST-DDT:
Continuation of #4060

- #4060

---

Transforms the location module to standalone modular functions.

---

This PR can be recreated using the following steps:

1. Checkout next/previous SMF PR
2. Run `pnpm tsx scripts/temp-transform-once.ts location`
4. Cherry-pick `refactor: transform location module`
5. Cherry-pick `chore: allow module`
6. Run `pnpm run generate:module-tree`
7. ~~Cherry-pick `chore: remove temp exports`~~
8. Cherry-pick `refactor: rename nearbyGPSCoordinate filename`

PR #4064, opened 2026-09-06
feat(lorem): transform to standalone modular functions

ST-DDT:
Continuation of #4063

- #4063

---

Transforms the lorem module to standalone modular functions.

---

This PR can be recreated using the following steps:

1. Checkout next/previous SMF PR
2. Run `pnpm tsx scripts/temp-transform-once.ts lorem`
4. Cherry-pick `refactor: transform lorem module`
5. Cherry-pick `chore: allow module`
6. Run `pnpm run generate:module-tree`
7. ~~Cherry-pick `chore: remove temp exports`~~


Shinigami92 (dismissed):

src/modules/lorem/module.ts
```diff
@@ -218,17 +214,7 @@ export class LoremModule extends ModuleBase {
...
-      'sentence',
-      'sentences',
-      'paragraph',
-      'paragraphs',
-      'lines',
-    ];
-
-    const method = this.faker.helpers.arrayElement(methods);
-
-    return this[method]();
```
xDivisionByZerox, line 231: **Note:** The implementation of loremText is technically different than before, as now a function ref directly picked inside arrayElement (instead of a method key).  I'd argue that this is an even better way to express the implementation, tho.
Just wanted to point out that I noticed this change.

xDivisionByZerox (dismissed):

Shinigami92 (approved):

xDivisionByZerox (approved):

PR #4079, opened 2026-09-08
refactor: replace internal groupBy with native Object.groupBy

### LINKED ISSUE #3856: Replace internal group-by.ts with native method
This is a reminder that `groupBy` is natively available in Node 22+.
So our workaround can be removed:

https://github.com/faker-js/faker/blob/next/src/internal/group-by.ts

We can do that when dropping support for Node 20 as part of v11.

Currently on hold, since we aren't ready for v11 yet.

YuEfSaEDU:
Replaces the internal `groupBy` helper (`src/internal/group-by.ts`) with the native `Object.groupBy`, which is available in every runtime supported by faker's engine range (Node `^22.13 || ^23.5 || >=24`). The native method produces an equivalent null-prototype, string-keyed record, so prototype-property keys such as `__proto__` and `toString` continue to group correctly; the sole value-mapping use site in the api-docs page index was folded into its tree-building step.

The two `as Record<...>` casts compensate for the standard library's conservative `Partial<Record<K, T[]>>` return type, which marks values optional even though grouping always populates them. Removes roughly 100 lines of internal helper code and its dedicated spec while keeping the full unit suite (54,207 passing tests), typecheck, and api-docs generation green.

Closes #3856


ST-DDT:
> while keeping the full unit suite (54,207 passing tests), typecheck, and api-docs generation green.

CI is red though. Please fix.

ST-DDT (changes requested):
CI is red. Could you please fix that?
If not, this PR will be closed as abandoned.

YuEfSaEDU:
Fixed — the failure was the formatting check, not lint itself: `oxfmt --check` flagged `scripts/apidocs/processing/method.ts` (one of the files this PR touches), and running the formatter resolved it (c8b74eab).

For transparency on the two commits in between: I initially mistook the `.todo` warning in `finance.spec.ts` for the failure and tried converting it to `it.skip`, but that trips the `no-disabled-tests` rule, so it's reverted (fce9733e). That `it.todo` comes from #4088 on `next` and is only a warning — it doesn't fail `oxlint` (as `next`'s own CI run at f8c23f9 shows), so I've left it alone as out of scope here.

Both `pnpm lint` and `pnpm oxfmt --check .` are clean locally now; CI should be green on this push.

ST-DDT (approved):

ST-DDT:

Shinigami92 (approved):
I approve it, but I'm slightly afraid that this could backfire in the weird envs like the react natives, where we also couldn't use Intl. But lets see :3

PR #4080, opened 2026-09-09
fix(git): replace all spaces in branch names

ML642:
`faker.git.branch()` replaces only the first space in each hacker noun and verb. With the Dutch locale and seed 30, it returns `programma-in-stukjes snijden`, which `git check-ref-format --branch` rejects. Replacing all spaces produces the valid name `programma-in-stukjes-snijden`.

Use `replaceAll` for both components and add three deterministic regression cases using custom locale data: multiple spaces in the noun, in the verb, and in both.

Validation:
- All three regression cases failed before the fix.
- Git module tests: 89 passed after the fix.
- Full `pnpm run preflight`: passed (54,213 tests passed; formatting, lint, build and TypeScript checks passed).

Prepared against `next`. The same logic is being moved in #4056; if that lands first, the fix will need to follow it into `src/modules/git/branch.ts`.

### Reproduction

The issue can be reproduced deterministically with the Dutch locale and seed `30`.

Create a small script:

```js
import { fakerNL } from '@faker-js/faker';

fakerNL.seed(30);

const branch = fakerNL.git.branch();

console.log(branch); //programma-in-stukjes snijden -> not a valid branch name since it contains spaces 
```

note: I didn't create an issue since the bug is pretty straight and small.

### DIFF
--- a/src/modules/git/module.ts
+++ b/src/modules/git/module.ts
@@ -19,8 +19,8 @@ export class GitModule extends ModuleBase {
    * @since 5.0.0
    */
   branch(): string {
-    const noun = this.faker.hacker.noun().replace(' ', '-');
-    const verb = this.faker.hacker.verb().replace(' ', '-');
+    const noun = this.faker.hacker.noun().replaceAll(' ', '-');
+    const verb = this.faker.hacker.verb().replaceAll(' ', '-');
     return `${noun}-${verb}`;
   }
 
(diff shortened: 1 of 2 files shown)


Shinigami92 (approved):

ST-DDT (approved):

matthewmayer (approved):

ST-DDT:

PR #4081, opened 2026-09-09
refactor!: only use FakerCore in modules

ST-DDT:
Continuation of #3748

- #3748

---

- Replace all references to `Faker` and `SimpleFaker` inside modules with `FakerCore`.
  - The `HelpersModule` is the only module, that retains a reference to `Faker` as it needs it for the module tree.
- The modules are now initialized inside the `Faker` constructor, as they require a reference to the `FakerCore` instance, that is passed as constructor parameter.
- This is breaking for JS users, that previously could access the `Faker` instances via `faker.airline.faker`.

### DIFF
--- a/scripts/module-tree/generate-module-tree.ts
+++ b/scripts/module-tree/generate-module-tree.ts
@@ -194,7 +194,7 @@ export async function generateModuleTree(onlyModule?: string): Promise<void> {
               .map((param) => param.getName());
 
             child.setBodyText(
-              `return ${toCamelCase(moduleName, methodName)}(this.faker.fakerCore, ${params.join(', ')});`
+              `return ${toCamelCase(moduleName, methodName)}(this.fakerCore, ${params.join(', ')});`
             );
           }
 
--- a/scripts/module-tree/generate-module-tree.ts
+++ b/scripts/module-tree/generate-module-tree.ts
@@ -26,10 +26,15 @@ function patchModuleImports(moduleName: string, importHelper: ImportHelper) {
     }
     case 'helpers': {
       importHelper.addImports('./_eval', 'fakeEval');
+      importHelper.addTypeImports('../../core', 'FakerCore');
       break;
     }
     case 'image': {
       importHelper.addTypeImports('../person', 'SexType');
+      importHelper.removeImports('../../faker', 'Faker');
+      break;
+    }
+    case 'location': {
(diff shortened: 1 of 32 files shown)


Shinigami92 (approved):
I talked to my AI and we agree: this is an acceptable breaking change and would not affect an end-user. The only affected users by this breaking change are meta-framework contributors which extends for one of the abstract classes, but even then these are not made public through `dist/index.ts` and therefore would require a (theoretically not officially supported) deep import.

xDivisionByZerox:
> I talked to my AI and we agree: this is an acceptable breaking change and would not affect an end-user. The only affected users by this breaking change are meta-framework contributors which extends for one of the abstract classes, but even then these are not made public through `dist/index.ts` and therefore would require a (theoretically not officially supported) deep import.

It is a breaking change because the faker property has an access level of protected. Simply extending any module (or the module base) and accessing faker (which is normal to do prior to SMF) would now break.

```ts
const AirlineModuleCtor = faker.airline.constructor.prototype as { new (): AirlineModule }
class MyCustomAirlineModule extends AirlineModuleCtor {
  abc() {
    return this.faker.string.sample(); // error: faker is not defined
  }
}

Shinigami92:
> It is a breaking change because the faker property has an access level of protected. Simply extending any module (or the module base) and accessing faker (which is normal to do prior to SMF) would now break.
> 
> ```ts
> const AirlineModuleCtor = faker.airline.constructor.prototype as { new (): AirlineModule }
> class MyCustomAirlineModule extends AirlineModuleCtor {
>   abc() {
>     return this.faker.string.sample(); // error: faker is not defined
>   }
> }
> ```

You don't even need to go that far, already `myModule instanceof SimpleModuleBase` is not possible anymore

ST-DDT:
Should we write a migration guide for this?
If yes, what do we say in there. "Dear JS user, if you relied on an internal/non-api field field or constructor, then that is no longer possible/it has been renamed to fakerCore." (Just more polite)?

Shinigami92:
> Should we write a migration guide for this? If yes, what do we say in there. "Dear JS user, if you relied on an internal/non-api field field or constructor, then that is no longer possible/it has been renamed to fakerCore." (Just more polite)?

If @matthewmayer says no to that, I think we are safe to merge without a migration guide here.

PR #4084, opened 2026-09-09
feat(helpers)!: forward fakerCore parameter in uniqueArray

ST-DDT:
Continuation of #4085

- #4085

---

Forward `fakerCore` parameter to generator function in `uniqueArray`.

This can be considered a breaking change, because any previously passed method with only optional parameters, will now receive a `FakerCore` as first parameter.

### DIFF
--- a/docs/guide/upgrading_v11/4084.md
+++ b/docs/guide/upgrading_v11/4084.md
@@ -1,0 +1,19 @@
+## Change in signature of uniqueArray
+
+`faker.helpers.uniqueArray` no passes the `fakerCore` to the embedded function.
+
+So functions that previously had only optional parameters are now incompatible as the first parameter is the `FakerCore`.
+
+This change allows users to pass other small modular functions directly to the `uniqueArray` method.
+
+```diff
+// Needs to be migrated
+-faker.helpers.uniqueArray(anyNoRequiredArgMethod, 3);
+
+// Works like before
+faker.helpers.uniqueArray(() => anyNoRequiredArgMethod(), 3);
+
+// New usages
++uniqueArray(fakerCore, () => anyNoRequiredArgMethod(), 3)
++uniqueArray(fakerCore, loremWord, 3)
+```
--- a/docs/guide/upgrading_v11/4084.md
+++ b/docs/guide/upgrading_v11/4084.md
@@ -1,6 +1,6 @@
 ## Change in signature of uniqueArray
 
-`faker.helpers.uniqueArray` no passes the `fakerCore` to the embedded function.
+`faker.helpers.uniqueArray` now passes the `fakerCore` to the embedded function.
 
 So functions that previously had only optional parameters are now incompatible as the first parameter is the `FakerCore`.
 
--- a/test/modules/helpers.spec.ts
+++ b/test/modules/helpers.spec.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import { FakerError, faker } from '../../src';
 import { luhnCheck } from '../../src/modules/helpers/_luhn-check';
+import { word as loremWord } from '../../src/modules/lorem/word';
 import { stringSample } from '../../src/modules/string/sample';
 import { seededTests } from '../support/seeded-runs';
 import { times } from '../support/times';
@@ -814,13 +815,19 @@ describe('helpers', () => {
           expect(unique).toHaveLength(length);
         });
 
-        it('function returns unique array', () => {
+        it('function returns unique array (independent function)', () => {
           const length = faker.number.int({ min: 1, max: 6 });
-          // TODO @ST-DDT 2026-09-09: Fix after word module has been migrated
           const unique = faker.helpers.uniqueArray(
             () => faker.lorem.word(),
             length
           );
+          expect(unique).not.toContainDuplicates();
+          expect(unique).toHaveLength(length);
+        });
+
+        it('function returns unique array (SMF)', () => {
+          const length = faker.number.int({ min: 1, max: 6 });
+          const unique = faker.helpers.uniqueArray(loremWord, length);
           expect(unique).not.toContainDuplicates();
           expect(unique).toHaveLength(length);
         });
--- a/test/modules/helpers.spec.ts
+++ b/test/modules/helpers.spec.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import { FakerError, faker } from '../../src';
 import { luhnCheck } from '../../src/modules/helpers/_luhn-check';
+import { stringSample } from '../../src/modules/string/sample';
 import { seededTests } from '../support/seeded-runs';
 import { times } from '../support/times';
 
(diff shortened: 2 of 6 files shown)


test/modules/helpers.spec.ts
```diff
@@ -811,7 +816,11 @@ describe('helpers', () => {
 
         it('function returns unique array', () => {
           const length = faker.number.int({ min: 1, max: 6 });
-          const unique = faker.helpers.uniqueArray(faker.lorem.word, length);
+          // TODO @ST-DDT 2026-09-09: Fix after word module has been migrated
```
ST-DDT, line 819: Waiting for migration of the word module.

docs/guide/upgrading_v11/4084.md
```diff
@@ -0,0 +1,19 @@
+## Change in signature of uniqueArray
+
+`faker.helpers.uniqueArray` no passes the `fakerCore` to the embedded function.
```
matthewmayer: No should be "now"

matthewmayer:
To clarify does this mean that the code in

https://fakerjs.dev/guide/unique

faker.helpers.uniqueArray(faker.internet.email, 1000);

Would break?

docs/guide/upgrading_v11/4084.md
ST-DDT:
```suggestion
`faker.helpers.uniqueArray` now passes the `fakerCore` to the embedded function.
```

ST-DDT:
Yes, that one will break type wise. Afaict it will not break implementation wise, others will though.
I have to update this example and others as well.

matthewmayer:
Here's a worse example

```
    faker.helpers.uniqueArray(faker.string.sample, 10);
```

previously output say
```
['UwrD]3>{+{', '`#SJMWg?:6', '(go;]`Q8)G', 'b_sP2OeVZ?', 'CY?KUD,wJk', 'yWrkS{0Cw=', 'lt^Fkb*6Hp', `*('rV="Z.h`, 'vg;4Mu]W92', 'q7I<H}L)iP']
```

now...

1. uniqueArray calls faker.string.sample(fakerCore) (methods are bound, so this is fine).
1. sample treats that arg as length: NumberOrRange and passes it to rangeToNumber.
1. rangeToNumber sees a non-number, so it does int(fakerCore, fakerCore).
1. int destructures options; FakerCore has no min/max, so you get:

`min = 0
max = Number.MAX_SAFE_INTEGER (9007199254740991)`
So sample picks a random length in that range (often trillions) and loops appending chars until it reaches it.

=> OOM

test/modules/helpers.spec.ts
```diff
@@ -1,6 +1,8 @@
 import { describe, expect, it } from 'vitest';
 import { FakerError, faker } from '../../src';
 import { luhnCheck } from '../../src/modules/helpers/_luhn-check';
+import { word as loremWord } from '../../src/modules/lorem/word';
+import { stringSample } from '../../src/modules/string/sample';
```
Shinigami92, line 5: **question (non-blocking):** 
```ts
import { word as loremWord } from '../../src/modules/lorem/word';
import { stringSample } from '../../src/modules/string/sample';
```

why is `loremWord` an alias, but `stringSample` does not need that? 🤔 
this is not a blocking review but a question because I might not have catched up to some decisions or something else I missed 😇

Shinigami92 (dismissed):

ST-DDT:
> Here's a worse example
> 
> ```
>     faker.helpers.uniqueArray(faker.string.sample, 10);
> ```
> 
> previously output say
> 
> ```
> ['UwrD]3>{+{', '`#SJMWg?:6', '(go;]`Q8)G', 'b_sP2OeVZ?', 'CY?KUD,wJk', 'yWrkS{0Cw=', 'lt^Fkb*6Hp', `*('rV="Z.h`, 'vg;4Mu]W92', 'q7I<H}L)iP']
> ```
> 
> now...
> 
>     1. uniqueArray calls faker.string.sample(fakerCore) (methods are bound, so this is fine).
> 
>     2. sample treats that arg as length: NumberOrRange and passes it to rangeToNumber.
> 
>     3. rangeToNumber sees a non-number, so it does int(fakerCore, fakerCore).
> 
>     4. int destructures options; FakerCore has no min/max, so you get:
> 
> 
> `min = 0 max = Number.MAX_SAFE_INTEGER (9007199254740991)` So sample picks a random length in that range (often trillions) and loops appending chars until it reaches it.
> 
> => OOM

And typescript will tell you not to do it.

test/modules/helpers.spec.ts
ST-DDT: Because stringSample is already prefixed with string.
After thinking about the public SMF api i noticed that unprefixed methods cause issues in a lot of places.

E.g. const firstName = firstName();

Another point is documentation.
How do we document the methods, if the method name changes based on the location you import it from?

E.g.
import personFirstName from fakerjs
import firstName from fakerjs/module/person.

In this case, I prefixed it to make it clear which word I meant. The word word or the lorem word.

Shinigami92: while I was reading through the stacked PRs, I assumed that all are prefixed except the helpers 🤔

matthewmayer:
I still feel it's a bit of a footgun as we were encouraging passing a  function directly in the docs

I think including something more similar to how we handled it for the V9 upgrading guide would be good for JS users. 

https://v9.fakerjs.dev/guide/upgrading#stricter-checking-for-function-signature-passed-to-faker-helpers-multiple-method

docs/guide/upgrading_v11/4084.md
matthewmayer, line 1: I tried rewriting this so it's more in the style of previous upgrading guides (what might break in your existing code and how to fix it)

## `faker.helpers.uniqueArray` now passes `FakerCore` to the callback

The `faker.helpers.uniqueArray` method takes a function reference (or an array) as its first parameter. Previously you may have written code like this to generate unique values:

```ts
faker.helpers.uniqueArray(faker.internet.email, 1000);
faker.helpers.uniqueArray(faker.string.sample, 10);
```

`uniqueArray` now calls that function with a `FakerCore` as its first argument. Methods that previously worked with no required parameters can therefore receive an unexpected value:

- In TypeScript, this is now a compile-time error.
- In JavaScript, the extra argument can cause unexpected behavior. For example, `faker.string.sample` could try to generate enormous strings and lead to out-of-memory errors.

Instead, wrap the call in a lambda:

```ts
faker.helpers.uniqueArray(() => faker.internet.email(), 1000);
faker.helpers.uniqueArray(() => faker.string.sample(), 10);
```

**Bad**

```ts
faker.helpers.uniqueArray(faker.string.sample, 10); // ❗
// In Typescript, this is now a compile time error
// Argument of type '(length?: NumberOrRange | undefined) => string'
// is not assignable to parameter of type 'readonly string[] | ((fakerCore: FakerCore) => string)'.
```

**Good**

```ts
faker.helpers.uniqueArray(() => faker.string.sample(), 10); // ✔
```

ST-DDT: Looks good to me.
Could you please push it for me?

matthewmayer: done

Shinigami92 (approved):

PR #4088, opened 2026-09-10
fix(finance): avoid hanging on empty currency symbols

### LINKED ISSUE #4087: bug(finance): currencySymbol hangs when all currency symbols are empty
### Pre-Checks

### Describe the bug

faker.finance.currencySymbol() can enter an infinite loop when a custom locale contains currencies with empty symbol values only.

### Minimal reproduction code
```js
import { Faker, base } from './dist/index.js';

const customLocale = {
  ...base,
  finance: {
    ...base.finance,
    currency: [
      {
        name: 'Test Currency',
        code: 'TST',
        symbol: '',
        numericCode: '000',
      },
    ],
  },
};

const faker = new Faker({ locale: customLocale });

console.log('Calling currencySymbol()...'); -> infinite loading
console.log(faker.finance.currencySymbol()); 
````
### Additional Context

While the documentation does not prohibit empty strings for currency symbols, custom locales and overrides are documented and supported.

### Environment Info

```shell
System:
  OS: Windows 11 10.0.26200
  CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1334U
  Memory: 2.36 GB / 15.69 GB
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  pnpm: 11.25.0
Browsers:
  Chrome: 152.0.7977.83
  Edge: Chromium 152.0.4191.66
  Firefox: 155.0.1
npmPackages:
  @faker-js/faker: 10.6.0
```

- [ ] CJS
- [x] ESM

### Used Package Manager

npm

ML642:
Proposed changes
Fixes an infinite loop in faker.finance.currencySymbol() when a custom locale provides currencies with empty symbol values only.
The method now throws a FakerError when no currency symbol is available, while preserving its existing behavior for locale data that includes at least one non-empty symbol.
A regression test covers the all-empty custom locale case.
closes #4087


src/modules/finance/module.ts
```diff
@@ -348,12 +348,24 @@ export class FinanceModule extends ModuleBase {
...
+        (currency) => currency.symbol.length > 0
+      )
+    ) {
+      throw new FakerError(
+        'Cannot get currency symbol from dataset with no currency symbols.'
+      );
+    }
+
     let symbol: string;
     do {
```
ST-DDT, line 370: We should do something about this loop.

First of all: Is not having a currency code a thing in the real world or is it a data problem.
   - I think we had a discussion about it previously somewhere, but I dont have time to check right now.

If it is a thing, the we should either add a requireSymbol option to currency or inline its implementation with an array.filter.
If it is a data problem, we can just drop the loop and fix the broken datasets we ship.

ST-DDT (changes requested):

ML642:
I found the previous discussion in #2435: https://github.com/faker-js/faker/pull/2435#discussion_r1348082771 
Although it was in 2023 , I hope is not outdated ,
 where xDivisionByZerox stated

```I had to look up while this is even getting looped. We should make the check more percise since the symbol itself always exists (as an empty string). Alternatively we could update the Currency definition to allow optional or null values.```

Based on that, I treated an empty string as the supported representation of a missing symbol and used the inline array.This required updating three snapshots because the method now makes a single random selection from the filtered list instead of repeatedly selecting from the full list until it finds a non-empty symbol.
@ST-DDT

File: src/modules/finance/module.ts [hunk 2]
```diff
@@ -348,18 +348,19 @@ export class FinanceModule extends ModuleBase {
...
-    let symbol: string;
-    do {
-      symbol = this.currency().symbol;
-    } while (symbol.length === 0);
-
-    return symbol;
+    return this.faker.helpers.arrayElement(
+      this.faker.definitions.finance.currency.filter(
+        (currency) => currency.symbol.length > 0
+      )
```
ST-DDT: Maybe store the filtered data in an array and throw a dedicated error message like before:

````ts
const filtered = this.faker.definitions.finance.currency.filter(
     (currency) => currency.symbol.length > 0
);

if (filtered.length === 0)
    throw new FakerError(
        'Cannot get currency symbol from dataset with no currency symbols.'
      );

return faker.helpers.arrayElement(filtered);
````

ML642: Thanks,  updated as suggested(only one thing arrayElement(...) as far as i know returns currency object and the method should return a string, so arrayElement(filtered).Symbol). The filtered currencies are now stored in a variable, and currencySymbol() throws a dedicated FakerError when no symbols are available. I kept .symbol on the selected currency so the method continues to return a string. pnpm run preflight passes.

ST-DDT: > Thanks, updated as suggested

Did you forget to push?

ML642: app, sorry about that , I didn't notice it got rejected because of the merge , now everything should be fine

ST-DDT (approved):

ST-DDT:
Thanks for your contribution and fast reactions. ❤️

ST-DDT:
For the other reviewers: The change of behavior can be verified using our docs preview:

````ts
await enableFaker();
const customLocale = {finance:{currency:[{name:"Test Currency",code:"TST",symbol:"",numericCode:"000"}]}};
const faker = new Faker({ locale: customLocale });
faker.finance.currencySymbol();
````

| Old | New
| --- | --- |
| [Website](https://fakerjs.dev/) | [Website](https://deploy-preview-4088.fakerjs.dev/) |
| Hangs forever | Throws error immediately |

Shinigami92 (approved):
thx 🙏

PR #4093, opened 2026-09-13
infra: always export using qualified name

ST-DDT:
Change all small modular functions to be prefixed with their module name.

This will likely match their public API naming.

You can review it commit by commit:

- `infra: always export using qualified name`
- `infra: temp adapt script to rename all methods`
- ~~`refactor: rename all methods (scripted)`~~ or `pnpm run generate:module-tree`
- `infra: revert temp script changes for rename` (clean revert of step 2)
- ~~`refactor: regenerate module-tree (scripted)`~~ or `pnpm run generate:module-tree`
- ~~`chore: format`~~ or `pnpm run format`
- `postfix modules`
- `fix: api-docs generation + jsdoc verification`
- `fix: jsdoc see references`

Doing it commit by commit (ignoring scripted ones) makes the review process way easier.

Splitting this into smaller PRs is near impossible/too much work, so I won't do it this time.

PR #4094, opened 2026-09-13
infra: fix examples with locale core

ST-DDT:
Fix issues from transformation step, that were missed by the script.

e.g. `fakerDE.location.city()` -> `city(fakerCoreDE)`

The change is only in the function files, as the modules should retain the existing code.

Apidocs and Tests are currently not affected, since they use the module code (This will likely change in a future PR, since the function files are now the source of truth/implementation).

---

> [!NOTE]
> To actually use the adjusted examples, we have to create exports for localized `FakerCore`s -> PUBLIC API.


src/modules/location/city.ts
```diff
@@ -8,7 +8,7 @@ import { Faker } from '../../faker';
  *
  * @example
  * city(fakerCore) // 'East Jarretmouth'
```
Shinigami92, line 10: **note (non-blocking):** unrelated to this PR, but should these be e.g. `locationCity(fakerCore)`?

Shinigami92 (approved):

src/modules/location/city.ts
ST-DDT: You mean like this:

- https://github.com/faker-js/faker/pull/4093

PR #4097, opened 2026-09-16
feat(locale): add missing common surnames to en person last names

AmritD:
### Problem

`en` is the default locale for anyone who doesn't pick one, but its `person.last_name` list has poor coverage of the most common English surnames.

**40 of England's 60 most common surnames are missing**, including `Taylor` (rank 3), `Williams` (5), `Wilson` (6), `Davies` (8), `Evans` (13), `Hall` (18), `Thomas` (19), `Clarke` (20), `Lewis` (35) and `Baker` (36).

The list does carry a long tail of very rare names inherited from the original faker.js. `Altenwerth`, for instance, has around 66 bearers worldwide, all of them in Germany, and ranks roughly 2,352,516th globally. It is in the English surname list; `Williams` is not.

So `faker.person.lastName()` seldom returns a surname that reads as ordinary.

### Changes

Data only, additive, one locale file: `src/locales/en/person/last_name.ts`, 473 to 534 entries. Nothing removed or renamed.

61 surnames that are common across English speaking countries. This closes the England top-60 gap, and also fills gaps in common US surnames: `Miller` was already there, but `Nelson`, `Stewart`, `Peterson`, `Reed` and `Brooks` were not.

<details>
<summary>The 61 names</summary>

Allen, Baker, Begum, Bell, Bennett, Brooks, Butler, Campbell, Clark, Clarke, Coleman, Cook, Cooper, Cox, Davies, Edwards, Evans, Foster, Gray, Hall, Hamilton, Harrison, Henderson, Hill, Howard, Hughes, Jackson, James, Kelly, Khan, Lee, Lewis, Long, Martin, Morgan, Morris, Nelson, Patel, Patterson, Perry, Peterson, Phillips, Powell, Reed, Richardson, Robinson, Rogers, Sanders, Scott, Shaw, Simmons, Singh, Stewart, Taylor, Thomas, Watson, Williams, Wilson, Wood, Wright, Young

</details>

### Sourcing

No list was copied from a single source. Candidates were drawn from general knowledge of English surnames, then each was checked against Forebears' name-distribution data, and the common-name ranking was cross-checked against its England frequency table.

### Verification

`pnpm run preflight` passes: 54,260 tests green, lint clean, no type errors.

`generate:locales` produces no further diff, so the added entries already match the project's own normalisation, including the sort order and the double-quoted apostrophe names.

`test/locale-data.spec.ts` passes, and the character-inventory snapshot is unchanged, since every added name is plain ASCII and that inventory already covers a-z and A-Z.

Six seeded snapshot files update: person, internet, git, company, finance and location. Every change is a name substitution caused by the array length changing. For example `person > 42 > lastName > with sex` goes from `Wiegand` to `Wilkinson`.

### Notes

An earlier revision of this PR also added 38 Yorkshire and Lancashire surnames. Per review those belong in `en_GB` rather than the generic `en` locale, so they have been removed here. They will follow in a separate PR that gives `en_GB` a surname list of its own, since a partial list there would shadow `en` entirely.

Seeded output for the default instance changes, which per CONTRIBUTING is a locale-data change and needs no migration-guide entry.


ST-DDT (dismissed):

matthewmayer:
In general its an improvement but i think there's an argument that the regional names like Shufflebotham belong in the en_GB locale rather than the generic en locale. 

`en` should probably be names which are reasonably common across all English speaking countries e.g. UK, USA, India, Canada, Australia etc.

xDivisionByZerox (changes requested):
> en should probably be names which are reasonably common across all English speaking countries e.g. UK, USA, India, Canada, Australia etc.

@AmritD could you remove the "38 regional surnames" (as stated by your PR decription) from the default `en` locale and put them in the `en_GB` locale instead. The "61 common surname" can stay in the `en` locale.

AmritD:
> > en should probably be names which are reasonably common across all English speaking countries e.g. UK, USA, India, Canada, Australia etc.
> 
> @AmritD could you remove the "38 regional surnames" (as stated by your PR decription) from the default `en` locale and put them in the `en_GB` locale instead. The "61 common surname" can stay in the `en` locale.

@xDivisionByZerox Happy to move them. One thing worth checking first though.

`en_GB` has no `person/last_name.ts` at the moment, so it currently falls through to `en`. Because `mergeLocales` replaces a definition key rather than concatenating it, adding a file there with only the 38 regional names makes those the entire `en_GB` surname pool. I tried it locally to be sure: the pool drops from 572 names to 38, and `Smith` and `Taylor` become unreachable for `en_GB`. The existing `en_GB` `last_name_pattern` hyphenates a pair one time in ten, so you also get `Haigh-Shufflebotham` and `Winterbottom-Hardcastle`, both real outputs from that build.

So I think the move implies giving `en_GB` a surname list of its own: the 38 regional names plus a sourced set of common British surnames, somewhere around 200 entries to sit in line with `en_AU` (279) and `en_ZA` (236). Several of the names in this PR would fit better there than in `en` anyway, `Davies`, `Evans` and `Griffiths` in particular.

I can build that here if you want it in one go. If you would rather keep this PR small, I can drop the 38 from `en` so the 61 common names merge on their own, and open `en_GB` as a follow-up.

matthewmayer:
Yes. Suggest just add the globally common names in this PR for en

Then create a separate PR with names suitable for across the UK for en_GB.

AmritD:
@xDivisionByZerox @matthewmayer 
Done, thanks both. The 38 regional names are out of `en`, leaving the 61 that are common across English speaking countries, so the file goes from 473 to 534 entries. Description updated to match, and preflight is green.

The `en_GB` list will follow as a separate PR. I'll make that a complete list rather than a partial one, so it doesn't shadow `en`.

ST-DDT (approved):

matthewmayer (approved):

matthewmayer:
LGTM. Will review the en_gb PR seperately.

I'd also be happy with another PR which strips out the especially obscure last names from `en` like Altenwerth, Gusikowski or Wisozk.

Shinigami92 (approved):

ST-DDT:
@xDivisionByZerox Could you please re-review this PR (or dismiss your stale review)?

PR #4104, opened 2026-09-22
test: assert public api contents

ST-DDT:
In preparation of upcoming changes to the codebase, we need to ensure that we don't accidentally add or remove values or types from the public api.

This PR adds a test that creates a snapshot of the types and values exported from the main index.ts.

The test likely needs to be amended when we export the SMFs, but it is a foundation.


Shinigami92:
_Random idea I have not proofed yet if possible/valid_: could we use `.spec-d.ts` / https://vitest.dev/guide/testing-types.html for this instead? 🤔 That way we wont test only that something exists, but also that it realistically exists and is exported as correctly typed 🤔

ST-DDT:
AFAICT it does not support snapshots, so you would have to manually adjust them every time.
The problem with correctly typed is, that the check itself relies on the value being checked.

expectTypeOf<VatCoutyCode>().toBeTypeOf<VatCountryCode>();

But unless you have a static copy of the type, both side will change at the same time.
If you have an idea to work around that, please let me know.

Shinigami92:
> AFAICT it does not support snapshots, so you would have to manually adjust them every time. The problem with correctly typed is, that the check itself relies on the value being checked.
> 
> expectTypeOf().toBeTypeOf();
> 
> But unless you have a static copy of the type, both side will change at the same time. If you have an idea to work around that, please let me know.

That makes so much sense 🤦 I haven't thought it from that angle 👍 Thanks for clearing that.
I will hopefully do a more review this evening, but cannot promise it, cause my Steam Frame arrives today 🤓

